PR #2695, opened 2025-06-13
Adds ability to set platform height independent of object height for terrain

jtigue-bdai:
# Description
Allow to set platform height independent of the object height.

## Type of change

## Checklist

Credit: @fjenelten-bdai

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -46,6 +46,7 @@ Guidelines for modifications:
 * Bikram Pandit
 * Bingjie Tang
 * Brayden Zhang
+* Brian Bingham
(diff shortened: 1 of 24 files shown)


ooctipus:
having consistent mesh terrain height for repeatedObjects seems to be a nice feature.
Can you remind me if without `platform_height`, what issue will the original implementation run into?

jtigue-bdai:
> having consistent mesh terrain height for repeatedObjects seems to be a nice feature. Can you remind me if without `platform_height`, what issue will the original implementation run into?

Main issue is that the starting platform would be required to be the same height as the objects. This allows for them to be decoupled.

Issue #2702, opened 2025-06-15
[Bug Report] Commit c984604 (optimization of getters) broke functionality with manager based RL workflow


ozhanozen:
### Describe the bug

The recent commit `c984604` (#2118), which optimises data getters inside asset classes, appears to break functionality in my custom training workflows. Learning outcomes now differ significantly (and mostly degrade). After investigation I traced the issue to `articulation_data.py`: reverting only the changes introduced by `c984604` in this file restores the previous training-workflow performance.

**Are the changes in `c984604` intended to affect anything beyond efficiency?** If not, those modifications may introduce problems for other users as well. 

While it’s hard to isolate a single culprit—my setup uses numerous custom MDP elements for events, observations, and curriculum—the likely pattern is that certain data buffers receive new timestamps before their underlying data are fresh, feeding stale values into the learning pipeline.

Example: `body_state_w` depends on `body_link_pose_w` and `body_com_vel_w`. If any function directly modifies the buffers of `_body_link_pose_w.data` *or* `_body_com_vel_w.data` after `body_state_w` has already been accessed, `body_state_w` will not capture this fresh information when it is called a second time. Any MDP element relying on `body_state_w` then might use out-of-date data for that timestep. I suspect such a case can slip through unnoticed, especially in event elements that randomise object or robot poses. In my environment I randomise the robot pose so that its end-effector starts near an object, whose pose is also randomized. With the recent changes, the function that randomises the end-effector may now read the object’s *old* pose, leading to incorrect initialisation.

To mitigate this in my own elements I invalidate the timestamp whenever stale data is possible (e.g. `robot.data._body_link_pose_w.timestamp = -1`). This alters the learning curve slightly but still fails to recover the performance seen before `c984604`. **Do any built-in IsaacLab components (e.g. `FrameTransformers`, standard MDP elements, etc.) call these getters in a way that could also surface stale values?**

### Steps to reproduce

Revert the changes from `c984604` (at least in **`articulation_data.py`**), then re-run a training workflow that previously converged correctly. You might observe markedly different learning behaviour between the reverted and current versions.

### System Info

- Commit: ba2a7dc
- Isaac Sim Version: 4.5
- OS: Ubuntu 22.04
- GPU: 4090 RTX
- CUDA: 12.2
- GPU Driver: 535.129.03

### Additional context

The issue might still relate to how I use the getters in my custom scripts, but ideally the data flow should be robust to such usage patterns.

### Checklist

[...]

PR #2736, opened 2025-06-18
Fixes inconsistent data reading in body, link, com for RigidObject, RigidObjectCollection and Articulation

### LINKED ISSUE #2534: [Bug Report] RIgidObject failed to invalidate timestamp-sensitive data when direct modifying root_pose and root_link_state
### Describe the bug

Correct Behavior
When calling method `isaaclab.assets.Articulation.write_root_link_pose_to_sim`, 
self._data.root_link_state_w[env_ids, :7] = root_pose.clone()    <-----  modified
self._data._body_state_w.timestamp = -1.0    <-----  timestamped invalidated
self._data._body_link_state_w.timestamp = -1.0    <-----  timestamped invalidated
self._data._body_com_state_w.timestamp = -1.0    <-----  timestamped invalidated

Bug Behavior
When calling method `isaaclab.assets.rigid_object.RigidObject.write_root_pose_to_sim`
self._data.root_state_w[env_ids, :7] = root_pose.clone()    <-----  modified (OK)
BUT MISSING Timestamp Invalidation   <-----  (BUGGY)

One potential bug this leads that when user calls `obs = env.reset()`, if obs contains reading of values whose timestamp hasn't properly invalidated, the observation would contain outdated values. A concrete example will be that the pose value of rigid Object can be inconsistent depending on whether user is accessing root_pose or root_link_state, the issue is that one is one step late than the other due to timestamp not properly invalided. After fix, these value will be consistent.

I'd note both RIgidObject and RigidObjectCollection have the issue

### Checklist

- [X] I have checked that there is no similar issue in the repo (**required**)
- [X] I have checked that the issue is not in running Isaac Sim itself and is related to the repo

### Acceptance Criteria

- [ ] time stamp of related data is corrected invalidated

ooctipus:
# Description
When WriteState, WriteLink, WriteCOM, WriteJoint are invoked, there is a inconsistency when reading values of ReadState, ReadLink, ReadCOM. The Source of the bug is because of missing timestamp invalidation of relative data or missing update to the related data within the write function. Below I list the all functions that is problematics

RigitObject:
write_root_link_pose_to_sim
write_root_com_velocity_to_sim

RigitObjectCollection:
write_object_link_pose_to_sim
write_object_com_velocity_to_sim

Articulation:
write_joint_state_to_sim

The bug if fixed by invalidating the relevant data timestamps in write_joint_state_to_sim function for articulation, and added direct update to the dependent data in write_(state|link|com)_to_sim of RigitObject and RigitObjectCollection.

I have added the tests cases that checks the consistency among ReadState, ReadLink, ReadCOM when either WriteState, WriteLink, WriteCOM, WriteJoint is called and passed all tests.

Fixes #2534 #2702 

## Type of change

Bug fix (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.40.7"
+version = "0.40.8"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.40.6"
+version = "0.40.7"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,19 @@
 Changelog
 ---------
 
+0.40.7 (2025-06-18)
+~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed data inconsistency between read_body, read_link, read_com when write_body, write_com, write_joint performed, in
+ :class:`~isaaclab.assets.Articulation`, :class:`~isaaclab.assets.RigidObject`, and
+ :class:`~isaaclab.assets.RigidObjectCollection`
+* added pytest that check against these data consistencies
+
+
+
 0.40.6 (2025-06-12)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab/isaaclab/assets/articulation/articulation.py
+++ b/source/isaaclab/isaaclab/assets/articulation/articulation.py
@@ -517,6 +517,12 @@ class Articulation(AssetBase):
         # set into internal buffers
         self._data.joint_pos[env_ids, joint_ids] = position
         # Need to invalidate the buffer to trigger the update with the new root pose.
+        self._data._body_com_vel_w.timestamp = -1.0
+        self._data._body_link_vel_w.timestamp = -1.0
+        self._data._body_com_pose_b.timestamp = -1.0
+        self._data._body_com_pose_w.timestamp = -1.0
+        self._data._body_link_pose_w.timestamp = -1.0
+
         self._data._body_state_w.timestamp = -1.0
         self._data._body_link_state_w.timestamp = -1.0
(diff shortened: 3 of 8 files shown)


kellyguo11:
@Mayankm96 @ozhanozen could you please help review this one to make sure it resolves the reported issues?

ozhanozen:
> @Mayankm96 @ozhanozen could you please help review this one to make sure it resolves the reported issues?

Hi @kellyguo11, I’d be happy to help. 

Quick thought: for pure-derived properties like `body_state_w`, `root_com_pose_w`, etc., which only concatenate or apply simple math and don’t call PhysX directly, could we drop their `timestamp` guards and recompute them on access instead? This would remove a lot of boilerplate and potential timestamp bugs with a negligible perf hit. What do you think?

ooctipus:
Hi @ozhanozen 
Your concern is right, I am also not very sure what will be the best design-intent here, @Mayankm96 may have more thoughts on this. What I did is to write tests to ensure the data consistency so that the issue you saw(I also experienced the same issue in RigidBody.write) is tested before future code release. I think the test should be complete, in english, they are testing that for RigidBodies, Articulation, RigidBodiesCollection, root_(state|pos|quat|vel|ang), bodylink_(state|pos|quat|vel|ang), com_(state|pos|quat|vel|ang) must agree after any write operation.

 @kellyguo11, feel free to add more input on this matter.

kellyguo11:
I agree it'd be less error-prone to remove the lazy updates if there are no perf hits, though it would be good to profile the operations to make sure there is indeed no impact on the performance before we make that change. Some of the torch operations could still be costly if called many times within a step.

jtigue-bdai:
@kellyguo11 did we get a feel for the speed performance from this?

kellyguo11:
IIRC the initial change for optimizing the APIs showed slight performance improvement, but not too significant. @ooctipus did you get to do some performance analysis with the latest changes here?

jtigue-bdai:
Ok that works. I know these buffers have been monkeyed around with a lot in the last year. I think consistent beets performance

ooctipus:
I didn't do any formal analysis with code. But I think it is pretty easy to reason it, 
last version were invalidating the timestamp where as the mayank's newer version transitions from some timestamp-invalidation to direct buffer modification upon write. This fix is matching the new style.

we basically have 3 ways to do about it:
 
1: get from physics:
pro: more direct
con: incur expense from physx everytime

2: invalidate time step
pro: save compute time, if that variable is never called
con: not aligning with current code how buffer is directly modified

3: direct update to all buffers (current)
pro: 1 time cost, align with direct buffer modification style in current code.
con: this 1 time cost is incurred every time write is invoked

I think all ways makes sense, and the performance gain is probably very minimum, if that is true, then I think style match can be important factor.

PR #2760, opened 2025-06-23
Updates `NoiseModelWithAdditiveBias` to apply per-feature bias sampling

### LINKED ISSUE #2759: [Proposal] Add per-feature bias sampling in `NoiseModelWithAdditiveBias`
### Proposal

The current `NoiseModelWithAdditiveBias` samples a single scalar bias per episode and applies it uniformly across all feature dimensions (i.e., axis 1 of a (`num_env`, `feature_dim`) tensor). I propose changing this behavior to sample a separate bias value for each feature dimension instead.

### Motivation

Applying the same bias value across all feature dimensions is overly restrictive, and often unrealistic, for structured inputs such as position vectors, velocities, or multi-DOF actions, where independent per-dimension drift is expected. Sampling per-dimension bias provides a more accurate and flexible representation of real-world noise in robotics and sim-to-real workflows.

As this is a minor change, I will create the PR already, which you can review.

### Checklist

### Acceptance Criteria

- [ ] `NoiseModelWithAdditiveBias` has independent sampling for all feature dimensions.

ozhanozen:
# Description

This PR updates the `NoiseModelWithAdditiveBias` to apply per-feature bias sampling.

Previously, the model sampled a single scalar bias per episode and applied it uniformly across all feature dimensions (i.e., axis 1 of a (`num_env`, `feature_dim`) tensor). This PR changes the behavior to instead sample a separate bias value for each feature dimension, making the model more suitable for structured inputs such as positions, velocities, or multi-DOF actions.

### Notes

Structured inputs typically contain semantically distinct components, like [x, y, z] coordinates, where applying the same bias across all components introduces unrealistic, fully correlated noise. Independent per-dimension bias sampling leads to more realistic and robust policy training, especially for sim-to-real transfer.

I’ve replaced the previous behavior with this new default, as I believe the original implementation could be misleading and not well-suited for many practical scenarios. However, if desired, I can make both behaviors available via a configuration flag (e.g., `per_feature_bias=True`), to retain backward compatibility.

Fixes #2759.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/mdp/events.py
+++ b/source/isaaclab/isaaclab/envs/mdp/events.py
@@ -1157,7 +1157,7 @@ def reset_scene_to_default(env: ManagerBasedEnv, env_ids: torch.Tensor):
         default_joint_vel = articulation_asset.data.default_joint_vel[env_ids].clone()
         # set into the physics simulation
         articulation_asset.set_joint_position_target(default_joint_pos, env_ids=env_ids)
-        articulation_asset.set_joint_velocity_target(default_joint_pos, env_ids=env_ids)
+        articulation_asset.set_joint_velocity_target(default_joint_vel, env_ids=env_ids)
         articulation_asset.write_joint_state_to_sim(default_joint_pos, default_joint_vel, env_ids=env_ids)
     # deformable objects
     for deformable_object in env.scene.deformable_objects.values():


ooctipus:
Thank you for PR, this is a very nice feature to have. 

One small question as I review the code, if you relie on self.reset() to re-sample that expanded bias in-place, does that mean multi-dim intput will have a different bias value every step? My intuition tells me the bias per feature should at least stay consistent throughout the episode and reset at next episode, If I am not wrong, the code is currently resample-per-episode for 1D bias, yet resample-per-step for N-D bias.

ozhanozen:
> Thank you for PR, this is a very nice feature to have.
> 
> One small question as I review the code, if you relie on self.reset() to re-sample that expanded bias in-place, does that mean multi-dim intput will have a different bias value every step? My intuition tells me the bias per feature should at least stay consistent throughout the episode and reset at next episode, If I am not wrong, the code is currently resample-per-episode for 1D bias, yet resample-per-step for N-D bias.

Hi @ooctipus,

The code section I wrote:
```
if self._feature_dim is None:
   *_, self._feature_dim = data.shape
   self._bias = self._bias.repeat(1, self._feature_dim)
   self.reset()
```
only ever happens once, on the very first `apply`. At that moment:
	1.	`self._feature_dim` goes from None → some integer.
	2.	We expand the bias from shape (N,1) → (N,D).
	3.	We invoke `self.reset()`, which re-samples that full (N,D) bias in place.

After that first call, `self._feature_dim` is no longer None, so we never enter that branch again. `apply` jumps straight to `return super().apply(data) + self._bias`. Subsequent steps are as they were before. No extra calls to `reset()` during an episode: no per-step re-sampling.

kellyguo11:
I think having an option to enable the backwards compatible approach will still be useful. Do you mind rebasing the changes on top of the other PR for noise model changes and update the extension versions?

ozhanozen:
> I think having an option to enable the backwards compatible approach will still be useful. Do you mind rebasing the changes on top of the other PR for noise model changes and update the extension versions?

I have added the `sample_bias_per_component` flag (default `True`) to enable/disable backwards compatibility, updated the PR with the latest changes, and updated extension version & changelog. PR is ready to be merged from my side.

PR #2797, opened 2025-06-27
Fixes the implementation of `quat_inv()`

### LINKED ISSUE #1263: [Bug Report] Potential implementation error in `quat_inv()`
Hello everyone,

The current implementation of the `quat_inv()` is as the following:

https://github.com/isaac-sim/IsaacLab/blob/a4118d7f9be827f8755ee71ea922928bb6c4a314/source/extensions/omni.isaac.lab/omni/isaac/lab/utils/math.py#L238-L247
where `normalize()` is:
https://github.com/isaac-sim/IsaacLab/blob/a4118d7f9be827f8755ee71ea922928bb6c4a314/source/extensions/omni.isaac.lab/omni/isaac/lab/utils/math.py#L81-L91

I believe the correct quaternion inverse should include a division with the norm-squared, not the norm. If the input quaternion is not a unit quaternion, the current implementation would give a wrong result. If it is assumed to be a unit quaternion, then we could simply return `quat_conjugate(q)` without normalizing it again. 

I suggest we change the implementation to something like:

```
def quat_inv(q: torch.Tensor, eps: float = 1e-9) -> torch.Tensor:
    """Computes the inverse of a quaternion.

    Args:
        q (torch.Tensor): The quaternion orientation in (w, x, y, z). Shape is (N, 4).
        eps (float): A small value to avoid division by zero. Defaults to 1e-9.

    Returns:
        torch.Tensor: Inverse quaternion in (w, x, y, z). Shape is (N, 4).
    """
    return quat_conjugate(q) / q.pow(2).sum(dim=-1, keepdim=True).clamp(min=eps)
```

### Checklist

ozhanozen:
# Description

Corrects `quat_inv()` in utils/math.py: the inverse is now computed as:

`quat_conjugate(q) / q.pow(2).sum(dim=-1, keepdim=True).clamp(min=eps)`

ensuring correct results for **non-unit** quaternions.  

Fixes #1263 (see discussion for details). Also updated `CHANGELOG.rst` and corrected a few minor typos within.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.40.10"
+version = "0.40.11"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -16,7 +16,7 @@ Fixed
 Fixed
 ^^^^^
 
-* Fixed :meth:`omni.isaac.lab.utils.dict.update_class_from_dict` preventing setting flat Iterables with different lengths.
+* Fixed :func:`~isaaclab.utils.dict.update_class_from_dict` preventing setting flat Iterables with different lengths.
 
 
 0.40.9 (2025-06-25)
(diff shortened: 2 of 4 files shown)


jtigue-bdai:
Can you add a test in `isaaclab/test/utils/test_math.py` that verifies the effectiveness of this change.

ozhanozen:
> Can you add a test in `isaaclab/test/utils/test_math.py` that verifies the effectiveness of this change.

Hi @jtigue-bdai, I have added a unit test that checks for the following:

$$
q \otimes q^{-1} = q^{-1} \otimes q = (1,0,0,0)
$$

The old implementation fails at this; the new one passes.

jtigue-bdai (approved):

PR #2826, opened 2025-06-30
Adds record at close functionality to the RecorderManager

jtigue-bdai:
# Description

Introducing a close function to the recorder manager which exports the data to file when the environment is closed and closes the recorder terms.

Fixes # (issue)

## Type of change

## Checklist


ooctipus (approved):
LGTM!

kellyguo11:
@peterd-NV do you mind taking a quick look at this change to make sure it's ok for mimic?

peterd-NV:
Hi @kellyguo11 @jtigue-bdai, I just have a couple of questions to make sure that everything will work with Mimic workflows: 

1. The new recorder manager close function will always be called during env.close(). This will result in whatever existing episode data for each env to be written to the dataset file. This will not overwrite any previously written data/episodes to the dataset file, correct?
2. The current episode when recording manager close is called may not be a complete episode. For example,  when doing multi-env data generation, there may be envs in the middle of an episode when the script is shutdown. In these cases, we do not want these episodes written to file. Can we continue to prevent the close function from writing the remaining incomplete episodes to file by using the `episode_succeeded` mechanism in the `export_epsiodes` function of the recorder manager?

jtigue-bdai:
> Hi @kellyguo11 @jtigue-bdai, I just have a couple of questions to make sure that everything will work with Mimic workflows:
> 
> 1. The new recorder manager close function will always be called during env.close(). This will result in whatever existing episode data for each env to be written to the dataset file. This will not overwrite any previously written data/episodes to the dataset file, correct?
> 2. The current episode when recording manager close is called may not be a complete episode. For example,  when doing multi-env data generation, there may be envs in the middle of an episode when the script is shutdown. In these cases, we do not want these episodes written to file. Can we continue to prevent the close function from writing the remaining incomplete episodes to file by using the `episode_succeeded` mechanism in the `export_epsiodes` function of the recorder manager?

Thanks @peterd-NV 
1. Correct this shouldn't affect any previous episode data. 
2. I believe it should still follow the same rules for recording as the original so the `episode_succeeded` should still function but what I don't know is if the `success_results` will be set if not entering the `record_pre_reset`. We could add a cfg parameter `export_in_close` to be able to toggle this though. Similar to the `export_in_record_pre_reset`.

peterd-NV:
Thanks for the answers @jtigue-bdai. 

Yes, if we could add a cfg parameter to toggle if export occurs in close that would be great. Ideally, if we could have this parameter default to be `False` that may be the safest option for existing Mimic workflow users. I believe most Mimic users are running parallel data generation/recording so having export in close default to `False` will guarantee their generated datasets don't contain unwanted incomplete demos.

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+0.41.1 (2025-07-25)
```
jtigue-bdai:
```suggestion
0.41.2 (2025-07-25)
```

jtigue-bdai:
@peterd-NV @kellyguo11 i have added the configuration parameter to this. Let me know what you think. Also i can't seem to get the recorder manager test to run locally for main or this branch. Anyone else having this issue?

source/isaaclab/isaaclab/managers/recorder_manager.py
peterd-NV, line 560: The new `export_in_close` config param should be added here otherwise it will be processed as a RecorderTerm.

jtigue-bdai: thanks for the catch.

jtigue-bdai: @peterd-NV I have added the param to the check.

peterd-NV: thanks

jtigue-bdai:
@kellyguo11 is anyone else seeing timeouts for: 
/workspace/isaaclab/source/isaaclab_rl/test/test_skrl_wrapper.py 

I can't seem to replicate locally

kellyguo11:
hmm I haven't been noticing, let me trigger it again

kellyguo11:
I can't repro locally as well, likely not relevant to this PR.

PR #2838, opened 2025-07-02
Relax upper version pin for torch for Blackwell GPU support.

### LINKED ISSUE #2431: [Proposal] Torch dependency for 50 series GPUs
### Describe the bug

The following packages have strict dependency on ```torch==2.5.1```: isaaclab, isaaclab_rl and isaaclab_tasks
This is a bit in contrast to the "Attention" section [here](https://isaac-sim.github.io/IsaacLab/main/source/setup/installation/isaaclab_pip_installation.html) where you are advised to work with the latest PyTorch version and CUDA 12.8

My point is - in the ```setup.py``` file of the different packages, the pytorch dependency should be ```torch>=2.5.1``` (you can have an upper-bound if you want). This will reduce dependency conflicts with other packages and GPUs, and as far as I tested it works (and documented).

### System Info

- Isaac Sim Version:4.5
- OS: Ubuntu 24.04
- GPU: B200
- CUDA: 12.8
- GPU Driver: 570.133.20

### Checklist

### Acceptance Criteria

- [ ] IsaacLab can be installed with pytorch>=2.5.1 dependencies

kwlzn:
# Description

This relaxes the maximum upper version for `torch` to allow for version 2.7 and above, which supports Blackwell GPUs. User reports in #1888 indicate that some folks are already operating successfully in a similar environment. This also updates the docs accordingly to use the released version of torch with this support vs the pre-release.

Fixes #2431
Relates to #1888

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/setup.py
+++ b/source/isaaclab/setup.py
@@ -20,7 +20,7 @@ EXTENSION_TOML_DATA = toml.load(os.path.join(EXTENSION_PATH, "config", "extensio
 INSTALL_REQUIRES = [
     # generic
     "numpy<2",
-    "torch==2.5.1",
+    "torch>=2.5.1",
     "onnx==1.16.1",  # 1.16.2 throws access violation on Windows
     "prettytable==3.3.0",
     "toml",
--- a/docs/source/setup/installation/binaries_installation.rst
+++ b/docs/source/setup/installation/binaries_installation.rst
@@ -385,7 +385,7 @@ Installation
 
 .. attention::
 
-   For 50 series GPUs, please use the latest PyTorch nightly build instead of PyTorch 2.5.1, which comes with Isaac Sim:
+   For 50 series (Blackwell) GPUs, please use PyTorch >=2.7 (`release notes <https://pytorch.org/get-started/previous-versions/#v270>`__) instead of PyTorch 2.5.1, which comes with Isaac Sim:
 
    .. tab-set::
       :sync-group: os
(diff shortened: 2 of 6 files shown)


kwlzn:
noting two flaky test failures in the CI job that pass locally for me, so I think this is green.

source/isaaclab/setup.py
```diff
@@ -20,7 +20,7 @@
 INSTALL_REQUIRES = [
     # generic
     "numpy<2",
-    "torch==2.5.1",
+    "torch>=2.5.1",
```
kellyguo11, line 23: this unfortunately will break the install on windows and end up pulling a non-cuda enabled torch build that causes downstream errors. we have moved the torch install step into isaaclab.sh/bat in the feature/isaacsim_5_0 branch to always use torch 2.7.1, which is also the version that Isaac Sim 5.0 will have.

kwlzn: what is it about the `>=` here that causes it to break? I was under the impression that 2.7.1 is the latest avail, so presumably it would resolve that on the upper bounds whether hard-coded to 2.7.1 or open ended like this?

but that's great news about torch 2.7.1 and IsaacSim 5.0 - we'd love to try that out as soon as it's avail.

kellyguo11: I think it's likely some way that pypi is setup on windows for the torch package that makes it pull in only the CPU build despite having a cuda build on the system.

Isaac Sim 5.0 has been open sourced and accessible at https://github.com/isaac-sim/IsaacSim. There is also a development branch for Isaac Lab that's compatible with Isaac Sim 5.0 at https://github.com/isaac-sim/IsaacLab/tree/feature/isaacsim_5_0

Issue #2632, opened 2025-06-06
[Proposal] Improve Current Ray Training Scripts


binw666:
### Proposal

This proposal aims to enhance the current Ray-based training scripts in the `isaaclab` project, with a focus on improving usability and flexibility for distributed training workflows. The goal is to address several limitations related to custom module support, file management, resource specification, and configuration handling.

### Motivation

The existing Ray scripts located in the `scripts/reinforcement_learning/ray` directory exhibit several usability issues that hinder effective deployment and integration in practical scenarios:

- **Lack of support for `py_modules`** prevents users from distributing custom Python packages or third-party libraries to Ray workers, limiting extensibility and customization.
- **Model files are written to local or relative directories**, such as `logs/`, which makes it difficult to access trained models—especially in remote or cloud environments where local filesystems are not directly accessible.
- **Resource allocation is not intuitive**—users often encounter confusion or make mistakes when specifying CPU, GPU, or memory requirements.
- **Heavy reliance on command-line arguments only** results in long, complex, and error-prone commands, reducing readability and maintainability.

I'm often frustrated when I need to manually patch the Ray scripts just to load a custom library or retrieve trained models without having to SSH into a remote machine.

### Alternatives

Some potential workarounds include:

- Modifying the Ray scripts manually to add support for `py_modules`.
- Using external tools (e.g., `scp`, or cloud CLI utilities) to move files after training.
- Writing custom Ray submission scripts to better manage job configuration, resource allocation, and post-training file handling.

While these alternatives can provide temporary relief, they increase complexity and maintenance burden, and do not offer a sustainable or user-friendly solution in the long run.

### Additional context

Currently, all training outputs (e.g., model checkpoints and logs) are saved to local working directories used by Ray. This setup works reasonably well in local environments but becomes problematic in remote or cluster-based executions. To improve accessibility and streamline deployment, there should be a built-in way to move or copy these files to mounted storage locations (e.g., network drives, cloud buckets like OSS) after training completes.

Additionally, while the current scripts do allow some level of resource specification, the interface is not intuitive. 

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

### Acceptance Criteria

- [ ] Support for `py_modules` to allow distribution of custom Python packages  
- [ ] Ability to move or copy training output files (e.g., models, logs) to specified directories or mounted storage (e.g., OSS)
[...]

PR #2346, opened 2025-04-21
Fixes ray lazy metric reporting and hanging processes

ozhanozen:
# Description

The step() function of ray/tuner.py has some issues preventing one from having an uninterrupted ray hyperparameter tuning session. Please refer to #2328 for details.

Fixes #2328.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #2847, opened 2025-07-03
✨ feat(Ray): Enhance Ray

binw666:
# Description

This PR:
- Add task_runner.py to support specifying resources, py_modules, and pip.

Fixes [# (issue)](https://github.com/isaac-sim/IsaacLab/issues/2632)

## Type of change

## Checklist

### DIFF
--- a/docs/source/features/ray.rst
+++ b/docs/source/features/ray.rst
@@ -46,7 +46,7 @@ specifying the ``--num_workers`` argument for resource-wrapped jobs, or ``--num_
 for tuning jobs, which is especially critical for parallel aggregate
 job processing on local/virtual multi-GPU machines. Tuning jobs assume homogeneous node resource composition for nodes with GPUs.
 
-The two following files contain the core functionality of the Ray integration.
+The three following files contain the core functionality of the Ray integration.
 
 .. dropdown:: scripts/reinforcement_learning/ray/wrap_resources.py
   :icon: code
@@ -62,6 +62,12 @@ The two following files contain the core functionality of the Ray integration.
     :language: python
     :emphasize-lines: 18-53
 
+.. dropdown:: scripts/reinforcement_learning/ray/task_runner.py
+  :icon: code
+
+  .. literalinclude:: ../../../scripts/reinforcement_learning/ray/task_runner.py
+    :language: python
+    :emphasize-lines: 9-55
 
 The following script can be used to submit aggregate
 jobs to one or more Ray cluster(s), which can be used for
@@ -73,7 +79,7 @@ resource requirements.
 
   .. literalinclude:: ../../../scripts/reinforcement_learning/ray/submit_job.py
     :language: python
-    :emphasize-lines: 12-53
+    :emphasize-lines: 13-59
 
 The following script can be used to extract KubeRay cluster information for aggregate job submission.
 
@@ -150,6 +156,15 @@ Submitting resource-wrapped individual jobs instead of automatic tuning runs is 
   .. literalinclude:: ../../../scripts/reinforcement_learning/ray/wrap_resources.py
     :language: python
     :emphasize-lines: 14-66
+
+Supports specifying per-task resources and setting ``py_modules`` and ``pip`` packages for each run.
+
+.. dropdown:: scripts/reinforcement_learning/ray/task_runner.py
+  :icon: code
+
+  .. literalinclude:: ../../../scripts/reinforcement_learning/ray/task_runner.py
+    :language: python
+    :emphasize-lines: 9-55
 
 Transferring files from the running container can be done as follows.
 
--- a/scripts/reinforcement_learning/ray/task_runner.py
+++ b/scripts/reinforcement_learning/ray/task_runner.py
@@ -1,0 +1,176 @@
+import yaml
+import ray
+import sys
+import argparse
+import subprocess
+import threading
+from enum import Enum
+
+"""
+This script dispatches one or more user-defined Python tasks to workers in a Ray cluster.
+Each task, with its resource requirements and execution parameters, is described in a YAML configuration file.
+You may specify the desired number of CPUs, GPUs, and memory allocation for each task in the config file.
+
+Key features:
+- Flexible resource management per task via config fields (`num_gpus`, `num_cpus`, `memory`).
+- Real-time output streaming (stdout/stderr) for each task.
+- Parallel execution of multiple tasks across cluster resources.
+
+Tasks are distributed and scheduled according to Ray’s built-in resource manager.
+
+Typical usage:
+---------------
+
+.. code-block:: bash
+
+    # Print help and argument details:
+    python task_runner.py -h
+
+    # Submit tasks defined in a YAML file to the Ray cluster (auto-detects Ray head address):
+    python task_runner.py --task_cfg /path/to/tasks.yaml
+
+YAML configuration example:
+---------------------------
+.. code-block:: yaml
+    pip: ["xxx"]
+    py_modules: ["my_package/my_package"]
+    tasks:
+      - name: "task1"
+        py_args: "-m torch.distributed.run --nnodes=1 ..."
+        num_gpus: 2
+        num_cpus: 10
+        memory: 10737418240
+      - name: "task2"
+        py_args: "script.py --option arg"
+        num_gpus: 0
+        num_cpus: 1
+        memory: 10*1024*1024*1024
+
+- `pip`: List of pip packages to install.
+- `py_args`: Arguments passed to the Python executable for this task.
+- `num_gpus`, `num_cpus`: Number of GPUs/CPUs to allocate. Can be integer or a string like `"2*2"`.
+- `memory`: Amount of memory (bytes) to allocate. Can be integer or a string like `"10*1024*1024*1024"`.
+
+To stop all tasks early, press Ctrl+C; the script will cancel all running Ray tasks.
+"""
+
+class OutputType(str, Enum):
+    STDOUT = "stdout"
+    STDERR = "stderr"
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Run tasks from a YAML config file.")
+    parser.add_argument("--task_cfg", type=str, required=True, help="Path to the YAML task file.")
+    parser.add_argument("--ray_address", type=str, default="auto", help="the Ray address.")
+    return parser.parse_args()
+    
+@ray.remote
+def task_wrapper(task):
+    task_name = task["name"]
+    task_py_args = task["py_args"]
+
+    # build command
+    cmd = [sys.executable, *task_py_args.split()]
+    print(f"[INFO]: {task_name} run: {' '.join(cmd)}")
+    def handle_stream(stream, output_type):
+        for line in iter(stream.readline, ''):
+            stripped_line = line.rstrip('\n')
+            if output_type == OutputType.STDOUT:
+                print(stripped_line)
+            elif output_type == OutputType.STDERR:
+                print(stripped_line, file=sys.stderr)
+        stream.close()
+    try:
+        process = subprocess.Popen(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+            bufsize=1  # None for best performance and 1 for realtime output
+        )
+
+        # start tow threads to read stdout and stderr
+        stdout_thread = threading.Thread(
+            target=handle_stream, args=(process.stdout, OutputType.STDOUT)
+        )
+        stderr_thread = threading.Thread(
+            target=handle_stream, args=(process.stderr, OutputType.STDERR)
+        )
+        stdout_thread.start()
+        stderr_thread.start()
+        # wait for process to finish
+        process.wait()
+        # wait for threads to finish
+        stdout_thread.join()
+        stderr_thread.join()
+
+        returncode = process.returncode
+    except Exception as e:
+        print(f"[ERROR]: error while running task {task_name}: {str(e)}" )
+        raise e
+
+    print(f"[INFO]: task {task_name} finished with return code {returncode}")
+    return True
+
+
+def submit_tasks(ray_address,pip,py_modules,tasks):
+    if not tasks:
+        print("[WARNING]: no tasks to submit")
+        return
+
+    if not ray.is_initialized():
+        try:
+            ray.init(address=ray_address, log_to_driver=True, runtime_env={
+                "pip": pip,
+                "py_modules": py_modules,
+            })
+        except Exception as e:
+            raise RuntimeError(f"initialize ray failed: {str(e)}")
+    task_results = []
+    for  task in tasks:
+        num_gpus = eval(task["num_gpus"]) if isinstance(task["num_gpus"], str) else task["num_gpus"]
+        num_cpus = eval(task["num_cpus"]) if isinstance(task["num_cpus"], str) else task["num_cpus"]
+        memory = eval(task["memory"]) if isinstance(task["memory"], str) else task["memory"]
+        print(f"[INFO]: submitting task {task['name']} with num_gpus={num_gpus}, num_cpus={num_cpus}, memory={memory}")
+        task_results.append(task_wrapper.options(
+            num_gpus=num_gpus,
+            num_cpus=num_cpus,
+            memory=memory,
+        ).remote(task))
+    
+    try:
+        results = ray.get(task_results)
+        for i, _ in enumerate(results):
+            print(f"[INFO]: Task {tasks[i]['name']} finished")
+        print("[INFO]: all tasks completed.")
+    except KeyboardInterrupt:
+        print("[INFO]: dealing with keyboard interrupt")
+        for future in task_results:
+            ray.cancel(future,force=True)
+        print("[INFO]: all tasks cancelled.")
+        sys.exit(1)
+    except Exception as e:
+        print(f"[ERROR]: error while running tasks: {str(e)}")
+        raise e
+
+
+def main():
+    args = parse_args()
+    try:
+        with open(args.task_cfg, 'r') as f:
+            config = yaml.safe_load(f)
+    except Exception as e:
+        raise SystemExit(f"error while loading task config: {str(e)}")
+    tasks = config["tasks"]
+    py_modules = config.get("py_modules",None)
+    pip = config.get("pip",None)
+    submit_tasks(
+            ray_address=args.ray_address,
+            pip=pip,
+            py_modules=py_modules,
+            tasks=tasks,
+        )
+
+if __name__ == "__main__":
+    main()
+
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -125,6 +125,7 @@ Guidelines for modifications:
 * Ziqi Fan
 * Zoe McCarthy
 * David Leon
+* Song Yi
 
 ## Acknowledgements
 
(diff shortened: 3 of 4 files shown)


binw666:
@ozhanozen @garylvov I have added task_runner.py as an additional method for submitting jobs. This method supports specifying resources, py_modules, and pip dependencies. If you have time, could you please help review the code? Thank you!

garylvov:
Hi I will review by the end of this weekend please ping me if no activity within 1 week

docs/source/features/ray.rst
```diff
@@ -73,7 +79,7 @@ resource requirements.
 
   .. literalinclude:: ../../../scripts/reinforcement_learning/ray/submit_job.py
     :language: python
-    :emphasize-lines: 12-53
+    :emphasize-lines: 13-59
```
garylvov: mind showing a preview of this highlight in the docs to make sure it's ok?

docs/source/features/ray.rst
```diff
@@ -62,6 +62,12 @@ The two following files contain the core functionality of the Ray integration.
     :language: python
     :emphasize-lines: 18-53
 
+.. dropdown:: scripts/reinforcement_learning/ray/task_runner.py
+  :icon: code
+
+  .. literalinclude:: ../../../scripts/reinforcement_learning/ray/task_runner.py
+    :language: python
+    :emphasize-lines: 9-55
```
garylvov: mind showing a preview of this highlight in the docs to make sure it's ok? Check out docs instructions for how to build locally

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,176 @@
...
+
+    # Print help and argument details:
+    python task_runner.py -h
+
+    # Submit tasks defined in a YAML file to the Ray cluster (auto-detects Ray head address):
+    python task_runner.py --task_cfg /path/to/tasks.yaml
+
+YAML configuration example:
+---------------------------
+.. code-block:: yaml
```
garylvov, line 55: Nit: can you show a real example in a config file as well if possible? Maybe will make this example a bit more clear

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,176 @@
...
+    STDERR = "stderr"
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="Run tasks from a YAML config file.")
+    parser.add_argument("--task_cfg", type=str, required=True, help="Path to the YAML task file.")
+    parser.add_argument("--ray_address", type=str, default="auto", help="the Ray address.")
+    return parser.parse_args()
+    
+@ray.remote
+def task_wrapper(task):
```
garylvov: I think you almost def may not want to handle the stdout and stderr logic here. We already have a piece of code that is similar to this, and it is already kinda brittle; Please see https://github.com/isaac-sim/IsaacLab/pull/2346

garylvov: Check out https://github.com/binw666/IsaacLab/blob/752be19bade88c1f1e2a06a1ca6519baafbba216/scripts/reinforcement_learning/ray/util.py#L115

garylvov: and https://github.com/binw666/IsaacLab/blob/752be19bade88c1f1e2a06a1ca6519baafbba216/scripts/reinforcement_learning/ray/util.py#L100

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,176 @@
...
+        returncode = process.returncode
+    except Exception as e:
+        print(f"[ERROR]: error while running task {task_name}: {str(e)}" )
+        raise e
+
+    print(f"[INFO]: task {task_name} finished with return code {returncode}")
+    return True
+
+
+def submit_tasks(ray_address,pip,py_modules,tasks):
```
garylvov: Similar to https://github.com/binw666/IsaacLab/blob/752be19bade88c1f1e2a06a1ca6519baafbba216/scripts/reinforcement_learning/ray/submit_job.py#L79 seems like duplicate logic that could be unified

garylvov:
Hi @binw666 thank you for the PR please let me know if you have any questions about my comments I'm happy to help. Once you address these issues just rerequest me for the review or @ me and I'll check it out again.

Thanks!

docs/source/features/ray.rst
binw666: The document has been updated, and the highlight is okay.
image
image

docs/source/features/ray.rst
binw666: The document has been updated, and the highlight is okay.
image
image

scripts/reinforcement_learning/ray/task_runner.py
binw666: Thank you for pointing out the execute_job function. I've removed the task_wrapper in later commits and replaced it with execute_job. Additionally, I set log_all_output=True by default to enable real-time output logging through Ray, making it easier to monitor job execution.

scripts/reinforcement_learning/ray/task_runner.py
binw666: This does resemble the submit_job function in submit_job.py, but it serves a different purpose. Here, the function calls remote to execute, whereas submit_job submits tasks via client.submit_job. Do you have any better suggestions?

scripts/reinforcement_learning/ray/task_runner.py
binw666: In the most recent commit, I have updated the comments and included an official example from IsaacLab for multi-GPU training. This may help clarify the usage.

binw666:
@garylvov Thank you very much for your review. I have updated the code and documentation to address the above issues.

ozhanozen:
@binw666, I don't have a HW currently that I can use to test the functionality of the PR, but I try to give feedback on other stuff as much as possible.

I noticed that some of the highlighted line numbers of the file `ray.rst` are outdated (some before your changes). Would it be possible for you to go over the file and update them as the last step once everything else is done?

ozhanozen:
@garylvov @binw666

I’m not sure if this PR is the right place to bring this up or if it should be addressed in a separate one, but based on my experience, I believe the current Ray integration is missing a key feature: support for a custom `CLIReporter`.

Right now, the integration defaults to the standard `CLIReporter`, which often displays metrics that aren’t particularly relevant. Similar to how we allow users to specify a `--cfg_class` (e.g., `CartpoleTheiaJobCfgE`), I suggest we let them optionally provide a custom `CLIReporter` class (that goes to the `RunConfig` as an argument). If none is provided, we can fall back to the default.

What do you think about this?

garylvov:
I think a custom CLIReporter could be good. As for including it in this PR, I am personally ok with it, but it's up to @binw666 . I could see why @binw666 may want to limit scope here, in which case we can follow up with another PR after this one is merged.

garylvov:
That being said, it's probably quickest to include that functionality in this PR in terms of getting things merged into main

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,176 @@
...
+    args = parse_args()
+    try:
+        with open(args.task_cfg, 'r') as f:
+            config = yaml.safe_load(f)
+    except Exception as e:
+        raise SystemExit(f"error while loading task config: {str(e)}")
+    tasks = config["tasks"]
+    py_modules = config.get("py_modules",None)
+    pip = config.get("pip",None)
+    submit_tasks(
```
garylvov: I think it would be good to link to these two comments here : https://github.com/isaac-sim/IsaacLab/issues/2632#issuecomment-2958392424 and  https://github.com/isaac-sim/IsaacLab/issues/2632#issuecomment-2959343978

scripts/reinforcement_learning/ray/task_runner.py
garylvov: Oh sorry I accidententally linked the wrong thing. I think ``wrap_resources_to_jobs`` and ``submit_tasks`` should be unified into a single method

wrap resources to jobs: https://github.com/binw666/IsaacLab/blob/752be19bade88c1f1e2a06a1ca6519baafbba216/scripts/reinforcement_learning/ray/wrap_resources.py#L68

garylvov: I know that the way you define your arguments is slightly different. I'm ok with modifying ``wrap_resources_to_jobs`` to fit your py modules and pip capability, as long as it's backwards compatible. 

Maybe an easy way to do this would to be having a single underlying function that we put into utilities.

Let's call this like `submit_wrapped_jobs`. Then, both ``submit_jobs`` and ``wrap_resources_to_jobs`` can call this shared utility.

I think having a single unified structure here will make maintenance significantly easier, similar to my previous comment about retrieving the logs themselves

docs/source/features/ray.rst
```diff
@@ -151,6 +157,15 @@ Submitting resource-wrapped individual jobs instead of automatic tuning runs is
     :language: python
     :emphasize-lines: 14-66
 
+Supports specifying per-task resources and setting ``py_modules`` and ``pip`` packages for each run.
```
garylvov: I'd also contrast this relative to how this is different from ``submit_job.py``. It should be clear to the user that these scripts have similar capability, and yours has extra benefits. Maybe, your script should be mentioned directly after ``submit_job.py`` script as well to group the similar packages together instead of separating them by other scripts.

I'd write something along the lines of: 

Similar to ``submit_job.py``, the following script submits aggregate jobs to a Ray cluster while supporting YAML-based resource specification (instead of command line arguments) and py_modules/pip modules through Ray.

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,159 @@
...
+    # Submit tasks defined in a YAML file to the Ray cluster (auto-detects Ray head address):
+    python task_runner.py --task_cfg /path/to/tasks.yaml
+
+YAML configuration example:
+---------------------------
+.. code-block:: yaml
+    pip: ["xxx"]
+    py_modules: ["my_package/my_package"]
+    tasks:
+      - name: "task1"
```
garylvov: I'd also add a comment here showing what a task1 could be

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,159 @@
...
+        num_gpus: 2
+        num_cpus: 10
+        memory: 10737418240
+      - name: "task2"
+        py_args: "script.py --option arg"
+        num_gpus: 0
+        num_cpus: 1
+        memory: 10*1024*1024*1024
+
+- `pip`: List of pip packages to install.
```
garylvov: A little pedantic, but I'd say something like an empty list means no additional packages above what is included in the base.

garylvov:
Hi @binw666 sorry for my inactivity, please feel free to ping me in the future if I'm inactive. I left another pass on the PR, please see my comments.

I think if the current comments are addressed, then this PR should be good to merge after it's nearly there!

Thank you for your hard work. 

It's up to you if you want to try adding https://github.com/isaac-sim/IsaacLab/pull/2847#issuecomment-3048613738, if not we can always follow it up in another PR.

scripts/reinforcement_learning/ray/task_runner.py
binw666: Great suggestion! I have extracted the task submission parts from `task_runner.py` and `wrap_resources.py` into `util.py`, and also enhanced `wrap_resources.py` to support `py_modules`.

scripts/reinforcement_learning/ray/task_runner.py
binw666: Currently, `py_modules` are supported in both `task_runner.py` and `wrap_resources.py`, but `pip` is only supported in `task_runner.py`. This is because `task_runner.py` tasks specify `py_args` without an explicit Python executable. When scheduled to Ray Nodes, `task_runner.py` retrieves the current Python executable and constructs the full command. In contrast, `wrap_resources.py` executes commands directly.

docs/source/features/ray.rst
binw666: Thank you for your suggestions. I have updated the document with more detailed usage instructions.
https://github.com/isaac-sim/IsaacLab/pull/2847/commits/26b365a32ce6d4e8e414fd19077278d29d360eb5#diff-35ad41931ab8268ec184eeac0be4f51a9036ece77aa318e8819b2ca3b1d58981R160

scripts/reinforcement_learning/ray/task_runner.py
binw666: Great suggestion! I have modified the `name` according to the task to better reflect its purpose.

scripts/reinforcement_learning/ray/task_runner.py
binw666: Haha, I emphasized in the latest commit that these are extra pip packages and python modules. This seems to be related to Ray, as empty lists are not allowed.

binw666:
@garylvov @ozhanozen Thank you very much for your review. I have updated the code and documentation to address the above comments.

Regarding support for a custom CLIReporter, I’m not very familiar with this part yet. If possible, I’d be glad to add this feature in a future PR.

scripts/reinforcement_learning/ray/task_runner.py
garylvov: Thank you!

garylvov:
Hi @binw666 , 
Thank you for the updates!
Sorry for the slow turn around time.

This PR looks like it's in pretty good shape. 

I plan to pull/test locally on my machine with your cartpole example later today, if it works I'll give it my approve stamp!

Nice work with getting PyTorch distributed to run with Ray tune I don't believe this was previously possible when I tried it super sick!

garylvov:
Editing PR title to fit style of other feature addition titles ("Adds ____")

garylvov:
Ok, I've confirmed that this works, super cool stuff thank you @binw666 

I tested this on my local 4 gpu machine. Here are my replication steps:

```
# Build the base image, but we don't need to run it
python3 docker/container.py start && python3 docker/container.py stop
# Build the tuning image with extra deps
docker build -t isaacray -f scripts/reinforcement_learning/ray/cluster_configs/Dockerfile .
# Start the tuning image - symlink so that changes in the source folder show up in the container
docker run -v $(pwd)/source:/workspace/isaaclab/source -it --gpus all --net=host --entrypoint /bin/bash isaacray
# Start the Ray server within the tuning image
echo "import ray; ray.init(); import time; [time.sleep(10) for _ in iter(int, 1)]" | ./isaaclab.sh -p
```


```
# In a new terminal (don't close the above) , enter the image with a new shell.
docker container ps
docker exec -it <ISAAC_RAY_IMAGE_ID_FROM_CONTAINER_PS> /bin/bash
```
^ from docs https://isaac-sim.github.io/IsaacLab/main/source/features/ray.html


New stuff:

```
cat > scripts/reinforcement_learning/ray/tasks.yaml <<EOF
concurrent: true
tasks:
  - name: "Isaac-Cartpole-job-1"
    py_args: >
      -m torch.distributed.run
      --nproc_per_node=2
      --nnodes=1
      --rdzv_id=job1
      --rdzv_backend=c10d
      --rdzv_endpoint=localhost:5555
      /workspace/isaaclab/scripts/reinforcement_learning/rsl_rl/train.py
      --task=Isaac-Cartpole-v0
      --headless
      --distributed
      --max_iterations 1000
    num_gpus: 2
    num_cpus: 10
    memory: 10*1024*1024*1024

  - name: "Isaac-Cartpole-job-2"
    py_args: >
      -m torch.distributed.run
      --nproc_per_node=2
      --nnodes=1
      --rdzv_id=job2
      --rdzv_backend=c10d
      --rdzv_endpoint=localhost:5556
      /workspace/isaaclab/scripts/reinforcement_learning/rsl_rl/train.py
      --task=Isaac-Cartpole-v0
      --headless
      --distributed
      --max_iterations 1000
    num_gpus: 2
    num_cpus: 10
    memory: 10*1024*1024*1024
EOF
```

```
echo -e "name: minerva address: http://127.0.0.1:8265" > /root/.cluster_config
python3 scripts/reinforcement_learning/ray/submit_job.py --aggregate_jobs task_runner.py --task_cfg tasks.yaml
```

image


However, if the commands are not well formatted in the .yaml configuration file, it seems to return early without throwing any obvious errors. This isn't a huge deal as the error handling is hard, or maybe I missed the logs. I think the user should always test the commands before putting them into the yaml to make sure it works.

garylvov (approved):
@kellyguo11 @jtigue-bdai I've reviewed and tested this PR. It looks good to me, and my local test ran successfully.

It only touches strictly Ray functionality.

~~Can you please review/merge into main when you get a chance?~~ Edit: need to lint 

Thank you!

garylvov:
@binw666 can you please fix the linting errors? After that we should be good to go

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,187 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+import argparse
+import sys
+import yaml
+from datetime import datetime
+
+import util
+
+"""
```
Mayankm96, line 6: NIT: Could this be moved to the top of the script (under the license). This then stays consistent with how we document modules as well.

scripts/reinforcement_learning/ray/task_runner.py
```diff
@@ -0,0 +1,187 @@
...
+        memory: 10*1024*1024*1024
+        node:
+          specific: "hostname"
+          hostname: "xxx"
+
+To stop all tasks early, press Ctrl+C; the script will cancel all running Ray tasks.
+"""
+
+
+def parse_args():
```
Mayankm96: Missing annotations and docstrings in the functions added.

binw666:
Sorry for the late reply — I’ve been caught up with some other tasks recently.

Many thanks to @garylvov for helping test the new functionality in the PR. Regarding your question “can you please fix the linting errors? After that we should be good to go”, I wasn’t entirely sure which tool you prefer for linting — I’ve been using ./isaaclab.sh -f for that.

Also, thanks to @Mayankm96 for pointing out the missing annotations and docstrings in the newly added functions. I’ve addressed these issues in my latest commit.

scripts/reinforcement_learning/ray/task_runner.py
binw666: Thank you. I have added the missing annotations and docstrings to the newly introduced functions in my latest commit.

scripts/reinforcement_learning/ray/task_runner.py
binw666: Thank you. I have moved the import to the top of the script (under the license). Could you please confirm if my understanding is correct?

garylvov:
Hi @binw666 thanks for getting back to me, I see that locally, pre-commit does pass, but [not on CI ](https://github.com/isaac-sim/IsaacLab/actions/runs/16925753085/job/48012346745?pr=2847)

I frankly have no idea what's going on here I'll just patch it manually

scripts/reinforcement_learning/ray/task_runner.py
garylvov: This is correct LGTM

garylvov:
@jtigue-bdai @kellyguo11 I think we're good to go on this PR; it only touches Ray functionality. 
Thanks!

garylvov:
Hi @kellyguo11  can we please merge up this PR? It's been unattended for a while

kellyguo11 (approved):

garylvov:
Thank you!!

PR #2885, opened 2025-07-09
Fixes ObservationManager history buffer corrupted by external calls to ObservationManager.compute

### LINKED ISSUE #2884: [Bug Report] ObservationManager history buffer corrupted by external calls to ObservationManager.compute
### Describe the bug

When `RecorderManager` and `RslRlVecEnvWrapper` calls `ObservationManager.compute`, it also appends the current observation to the history even though environment is not stepping. This caused corrupted history buffer.

With `RecorderManager` and observation history buffer length greater than zero, it records inaccurate observation.

### Checklist

patrickhaoy:
# Description

When observation group has history length greater than zero, calling `ObservationManager.compute` modifies history state by appending current observation to history. This creates history corruption when non-`ManagerBasedEnv` classes invoke `ObservationManager.compute`. This PR introduces `update_history` flag (default to `False`) and only `ManagerBasedEnv` has the privilege to run `ObservationManager.compute` with `update_history=True`. If `update_history=False` and the history buffer is `None`, a copy of history is returned instead of the original. 

I have added test cases to verify this fix is effective.

Fixes #2884 

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/managers/observation_manager.py
+++ b/source/isaaclab/isaaclab/managers/observation_manager.py
@@ -245,7 +245,7 @@ class ObservationManager(ManagerBase):
         # nothing to log here
         return {}
 
-    def compute(self) -> dict[str, torch.Tensor | dict[str, torch.Tensor]]:
+    def compute(self, update_history: bool = False) -> dict[str, torch.Tensor | dict[str, torch.Tensor]]:
         """Compute the observations per group for all groups.
 
         The method computes the observations for all the groups handled by the observation manager.
@@ -260,14 +260,14 @@ class ObservationManager(ManagerBase):
         obs_buffer = dict()
         # iterate over all the terms in each group
         for group_name in self._group_obs_term_names:
-            obs_buffer[group_name] = self.compute_group(group_name)
+            obs_buffer[group_name] = self.compute_group(group_name, update_history=update_history)
         # otherwise return a dict with observations of all groups
 
         # Cache the observations.
         self._obs_buffer = obs_buffer
         return obs_buffer
 
-    def compute_group(self, group_name: str) -> torch.Tensor | dict[str, torch.Tensor]:
+    def compute_group(self, group_name: str, update_history: bool = False) -> torch.Tensor | dict[str, torch.Tensor]:
         """Computes the observations for a given group.
 
         The observations for a given group are computed by calling the registered functions for each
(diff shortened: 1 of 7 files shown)


ooctipus (approved):
Thanks for the pull request, it seems to be an bug we definitely need to fix soon.

source/isaaclab/isaaclab/managers/observation_manager.py
```diff
@@ -260,14 +260,14 @@ def compute(self) -> dict[str, torch.Tensor | dict[str, torch.Tensor]]:
...
-            obs_buffer[group_name] = self.compute_group(group_name)
+            obs_buffer[group_name] = self.compute_group(group_name, update_history=update_history)
         # otherwise return a dict with observations of all groups
 
         # Cache the observations.
         self._obs_buffer = obs_buffer
         return obs_buffer
 
-    def compute_group(self, group_name: str) -> torch.Tensor | dict[str, torch.Tensor]:
+    def compute_group(self, group_name: str, update_history: bool = False) -> torch.Tensor | dict[str, torch.Tensor]:
```
kellyguo11, line 275: could we also add this to the docstring to explain what the parameter does?

source/isaaclab/isaaclab/managers/observation_manager.py
```diff
@@ -330,13 +330,23 @@ def compute_group(self, group_name: str) -> torch.Tensor | dict[str, torch.Tenso
...
+                    circular_buffer.append(obs)
+                elif circular_buffer._buffer is None:
+                    # because circular buffer only exits after the simulation steps,
+                    # this guards history buffer from corruption by external calls before simulation start
+                    circular_buffer = CircularBuffer(
+                        max_len=circular_buffer.max_length,
+                        batch_size=circular_buffer.batch_size,
+                        device=circular_buffer.device,
                     )
+                    circular_buffer.append(obs)
```
kellyguo11, line 352: should this be called only if `update_history=True`?

source/isaaclab/isaaclab/managers/observation_manager.py
```diff
@@ -330,13 +330,23 @@ def compute_group(self, group_name: str) -> torch.Tensor | dict[str, torch.Tenso
...
             # Update the history buffer if observation term has history enabled
             if term_cfg.history_length > 0:
-                self._group_obs_term_history_buffer[group_name][term_name].append(obs)
-                if term_cfg.flatten_history_dim:
-                    group_obs[term_name] = self._group_obs_term_history_buffer[group_name][term_name].buffer.reshape(
-                        self._env.num_envs, -1
+                circular_buffer = self._group_obs_term_history_buffer[group_name][term_name]
+                if update_history:
+                    circular_buffer.append(obs)
+                elif circular_buffer._buffer is None:
```
kellyguo11, line 344: can `circular_buffer` be None in the above case when `if update_history` evaluates to True?

source/isaaclab/isaaclab/managers/observation_manager.py
```diff
@@ -330,13 +330,23 @@ def compute_group(self, group_name: str) -> torch.Tensor | dict[str, torch.Tenso
                 obs = obs.mul_(term_cfg.scale)
             # Update the history buffer if observation term has history enabled
             if term_cfg.history_length > 0:
```
kellyguo11, line 340: what is the difference between the new `update_history` argument and the `history_length > 0`?

source/isaaclab/isaaclab/managers/observation_manager.py
ooctipus: The idea is that the only difference between update_history=True and update_history=False is that, one modifies internal state and the other doesn't. Beside this difference user should expect the current obs been returned. (no surprising values, such as dummy values)

if we don't append, when user calls observation_manager.compute() before first reset, they will get empty buffer filled with dummy value most likely, that can sometime violate the clip specified in each observation term. So the safest option is to make a empty circular buffer copy, and append obs to that copy, and return.

source/isaaclab/isaaclab/managers/observation_manager.py
ooctipus: while the circular_buffer._buffer could be None, The CircularBuffer class is not None. 

if update_history is true, and circular_buffer._buffer is None, circular_buffer.append(obs) will make sure to initialize the _buffer within the append function.

source/isaaclab/isaaclab/managers/observation_manager.py
ooctipus: history_length > 0 means means if which term is a circular buffer.

update_history means if calling `compute()` will add the current obs to this buffer. Say if `ManagerBasedEnv.step()` calls `observation_manager.compute()` we want `update_history=True` because we want to append the current obs to internal history buffer.

But if say `recorder_term` call  `observation_manager.compute()`, we want `update_history=False` because otherwise the history_buffer will be corrupted by repeated obs.

source/isaaclab/isaaclab/managers/observation_manager.py
ooctipus: added XD

PR #2027, opened 2025-03-06
Adds Anymal-C navigation rsl_rl example with obstacle and camera in the scene

KingsleyLiu-NV:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

This PR adds the anymal-c navigation rsl_rl example with obstacle and camera in the scene:

- Extends the rsl_rl libraries to support convolution networks and ppo runner
- Extends the anymal-c navigation environment to have rigid object obstacle and tiled camera sensor
  - Adds the event reset function to sample the position in a sector of ring area
  - Adds the observation function for the flattened normalized rgb data
  - Adds the pose3 command resample function to determine the target based on the obstacle position
- Tune the RL reward terms to make the training converge

## Type of change

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

It can be run with the following command:
```
./isaaclab.sh -p scripts/reinforcement_learning/rsl_rl/train.py --task Isaac-Navigation-Flat-Obstacle-Anymal-C-v0 --headless --num_envs 4096 --max_iterations 1501 --enable_cameras

./isaaclab.sh -p scripts/reinforcement_learning/rsl_rl/play.py --task Isaac-Navigation-Flat-Obstacle-Anymal-C-Play-v0 --headless --num_envs 16 --video
```

## Screenshots

The training results are as below:
image

Memory footprint: (num_envs=4096, image_input_shape=(3, 64, 64))
image

Training speed:
image

Play with `model_1300.pt`:

https://github.com/user-attachments/assets/b1eed6b5-f570-4eaf-9b62-27e1bcfbafb3

## Checklist

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
  - [x] isaaclab changelog & extension.toml
  - [x] isaaclab_rl changelog & extension.toml
  - [x] isaaclab_tasks changelog & extension.toml
  - [x] environments.rst
  - [x] ??? More
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #2962, opened 2025-07-16
Adds changes for rsl_rl 3.0.1

ClemensSchwarke:
# Description

This PR adds the necessary changes to work with the new version of RSL RL.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
+++ b/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
@@ -148,6 +148,12 @@ class RslRlBaseRunnerCfg:
     max_iterations: int = MISSING
     """The maximum number of iterations."""
 
+    empirical_normalization: bool = None
+    """This parameter is deprecated and will be removed in the future.
+
+    Use `actor_obs_normalization` and `critic_obs_normalization` instead.
+    """
+
     obs_groups: dict[str, list[str]] = MISSING
     """A mapping from observation groups to observation sets.
 
--- a/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
+++ b/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
@@ -32,11 +32,11 @@ class RslRlPpoActorCriticCfg:
     noise_std_type: Literal["scalar", "log"] = "scalar"
     """The type of noise standard deviation for the policy. Default is scalar."""
 
-    actor_obs_normalization: bool = False
-    """Whether to normalize the observation for the actor network. Default is False."""
+    actor_obs_normalization: bool = MISSING
+    """Whether to normalize the observation for the actor network."""
 
-    critic_obs_normalization: bool = False
-    """Whether to normalize the observation for the critic network. Default is False."""
+    critic_obs_normalization: bool = MISSING
+    """Whether to normalize the observation for the critic network."""
 
     actor_hidden_dims: list[int] = MISSING
     """The hidden dimensions of the actor network."""
--- a/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
+++ b/source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
@@ -32,6 +32,12 @@ class RslRlPpoActorCriticCfg:
     noise_std_type: Literal["scalar", "log"] = "scalar"
     """The type of noise standard deviation for the policy. Default is scalar."""
 
+    actor_obs_normalization: bool = False
+    """Whether to normalize the observation for the actor network. Default is False."""
+
+    critic_obs_normalization: bool = False
+    """Whether to normalize the observation for the critic network. Default is False."""
+
     actor_hidden_dims: list[int] = MISSING
     """The hidden dimensions of the actor network."""
 
@@ -114,13 +120,11 @@ class RslRlPpoAlgorithmCfg:
     Otherwise, the advantage is normalized over the entire collected trajectories.
     """
 
+    rnd_cfg: RslRlRndCfg | None = None
+    """The RND configuration. Default is None, in which case RND is not used."""
+
     symmetry_cfg: RslRlSymmetryCfg | None = None
     """The symmetry configuration. Default is None, in which case symmetry is not used."""
-
-    rnd_cfg: RslRlRndCfg | None = None
-    """The configuration for the Random Network Distillation (RND) module. Default is None,
-    in which case RND is not used.
-    """
 
 
 #########################
@@ -129,8 +133,8 @@ class RslRlPpoAlgorithmCfg:
 
 
 @configclass
-class RslRlOnPolicyRunnerCfg:
-    """Configuration of the runner for on-policy algorithms."""
+class RslRlBaseRunnerCfg:
+    """Base configuration of the runner."""
 
     seed: int = 42
     """The seed for the experiment. Default is 42."""
@@ -144,17 +148,30 @@ class RslRlOnPolicyRunnerCfg:
     max_iterations: int = MISSING
     """The maximum number of iterations."""
 
-    empirical_normalization: bool = MISSING
-    """Whether to use empirical normalization."""
+    obs_groups: dict[str, list[str]] = MISSING
+    """A mapping from observation groups to observation sets.
 
-    policy: RslRlPpoActorCriticCfg | RslRlDistillationStudentTeacherCfg = MISSING
-    """The policy configuration."""
+    The keys of the dictionary are predefined observation sets used by the underlying algorithm
+    and values are lists of observation groups provided by the environment.
 
-    algorithm: RslRlPpoAlgorithmCfg | RslRlDistillationAlgorithmCfg = MISSING
-    """The algorithm configuration."""
+    For instance, if the environment provides a dictionary of observations with groups "policy", "images",
+    and "privileged", these can be mapped to algorithmic observation sets as follows:
+
+    .. code-block:: python
+
+        obs_groups = {
+            "policy": ["policy", "images"],
+            "critic": ["policy", "privileged"],
+        }
+
+    This way, the policy will receive the "policy" and "images" observations, and the critic will
+    receive the "policy" and "privileged" observations.
+
+    For more details, please check ``vec_env.py`` in the rsl_rl library.
+    """
 
     clip_actions: float | None = None
-    """The clipping value for actions. If ``None``, then no clipping is done.
+    """The clipping value for actions. If None, then no clipping is done. Defaults to None.
 
     .. note::
         This clipping is performed inside the :class:`RslRlVecEnvWrapper` wrapper.
@@ -184,7 +201,10 @@ class RslRlOnPolicyRunnerCfg:
     """The wandb project name. Default is "isaaclab"."""
 
     resume: bool = False
-    """Whether to resume. Default is False."""
+    """Whether to resume a previous training. Default is False.
+
+    This flag will be ignored for distillation.
+    """
 
     load_run: str = ".*"
     """The run directory to load. Default is ".*" (all).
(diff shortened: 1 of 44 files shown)


kellyguo11:
Thanks for the changes, this is awesome! Looks like it could have some breaking changes for existing cfgs, should we include a check for the RSL RL version to force a version update if users have older versions installed?

ClemensSchwarke:
> Thanks for the changes, this is awesome! Looks like it could have some breaking changes for existing cfgs, should we include a check for the RSL RL version to force a version update if users have older versions installed?

Ah yes that would make a lot of sense I think. We could alternatively just keep all the configs unmodified as the new rsl rl version is backwards compatible, but I thought it would be nicer to show the updated way of handling normalization in the example configs. Any opinions on this?

kellyguo11:
> > Thanks for the changes, this is awesome! Looks like it could have some breaking changes for existing cfgs, should we include a check for the RSL RL version to force a version update if users have older versions installed?
> 
> Ah yes that would make a lot of sense I think. We could alternatively just keep all the configs unmodified as the new rsl rl version is backwards compatible, but I thought it would be nicer to show the updated way of handling normalization in the example configs. Any opinions on this?

I think showing the new features would be nice as well. @Mayankm96 @ooctipus @pascal-roth @jtigue-bdai what are your thoughts?

source/isaaclab_rl/isaaclab_rl/rsl_rl/vecenv_wrapper.py
```diff
@@ -162,30 +136,28 @@ def episode_length_buf(self, value: torch.Tensor):
     def seed(self, seed: int = -1) -> int:  # noqa: D102
         return self.unwrapped.seed(seed)
 
-    def reset(self) -> tuple[torch.Tensor, dict]:  # noqa: D102
```
pascal-roth, line 165: Whats the reason that the reset function is no longer needed?

source/isaaclab_tasks/isaaclab_tasks/direct/allegro_hand/agents/rsl_rl_ppo_cfg.py
```diff
@@ -14,9 +14,10 @@ class AllegroHandPPORunnerCfg(RslRlOnPolicyRunnerCfg):
     max_iterations = 10000
     save_interval = 250
     experiment_name = "allegro_hand"
-    empirical_normalization = True
```
pascal-roth, line 17: It is not backwards compatible here, the key is missing in the new confit class (the key is missing in the config)

pascal-roth (changes requested):
Thanks for pushing that out. 

I agree we should update to the new normalisation definition but currently it is not backward compatible, should have that fixed

ClemensSchwarke:
> Thanks for pushing that out.
> 
> I agree we should update to the new normalisation definition but currently it is not backward compatible, should have that fixed

@pascal-roth So you are saying forcing the new rsl rl version is not the way to go?

source/isaaclab_rl/isaaclab_rl/rsl_rl/vecenv_wrapper.py
ClemensSchwarke: It was never used/needed.

pascal-roth:
> > Thanks for pushing that out.
> 
> > 
> 
> > I agree we should update to the new normalisation definition but currently it is not backward compatible, should have that fixed
> 
> 
> 
> @pascal-roth So you are saying forcing the new rsl rl version is not the way to go?

No, we should update. 

But in the RunnerCfg, you still need to have an attribute called: Empiricial Normalization, which throws a depreciation warning and sets the new actor and critic normalisation

source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
```diff
@@ -144,17 +148,36 @@ class RslRlOnPolicyRunnerCfg:
     max_iterations: int = MISSING
     """The maximum number of iterations."""
 
-    empirical_normalization: bool = MISSING
-    """Whether to use empirical normalization."""
+    empirical_normalization: bool = None
```
pascal-roth:
```suggestion
    empirical_normalization: bool | None = None
```

kellyguo11:
@ClemensSchwarke looks like we might be hitting some package dependency errors in CI, are these expected?

```
#14 171.1 ERROR: Cannot install None, isaaclab-rl[all]==0.1.8 and isaaclab_rl because these package versions have conflicting dependencies.
#14 171.1 
#14 171.1 The conflict is caused by:
#14 171.1     isaaclab-rl 0.1.8 depends on torch==2.5.1
#14 171.1     isaaclab-rl[all] 0.1.8 depends on torch==2.5.1
#14 171.1     rsl-rl-lib 3.0.0 depends on torch>=2.6.0
#14 171.1 
#14 171.1 To fix this you could try to:
#14 171.1 1. loosen the range of package versions you've specified
#14 171.1 2. remove package versions to allow pip attempt to solve the dependency conflict
#14 171.1 
#14 171.1 ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

KingsleyLiu-NV:
Hi @ClemensSchwarke @kellyguo11, does rsl-rl 3.0 support convolution networks? If not, it still requires some customization and two PRs (one for isaaclab and one for rsl-rl) to enable the tiled-camera navigation example in [PR 2027](https://github.com/isaac-sim/IsaacLab/pull/2027).

ClemensSchwarke:
> @ClemensSchwarke looks like we might be hitting some package dependency errors in CI, are these expected?
> 
> ```
> #14 171.1 ERROR: Cannot install None, isaaclab-rl[all]==0.1.8 and isaaclab_rl because these package versions have conflicting dependencies.
> #14 171.1 
> #14 171.1 The conflict is caused by:
> #14 171.1     isaaclab-rl 0.1.8 depends on torch==2.5.1
> #14 171.1     isaaclab-rl[all] 0.1.8 depends on torch==2.5.1
> #14 171.1     rsl-rl-lib 3.0.0 depends on torch>=2.6.0
> #14 171.1 
> #14 171.1 To fix this you could try to:
> #14 171.1 1. loosen the range of package versions you've specified
> #14 171.1 2. remove package versions to allow pip attempt to solve the dependency conflict
> #14 171.1 
> #14 171.1 ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
> ```

@kellyguo11 I think we need 2.6 for tensordict. Do you think it is feasible to switch to torch 2.6 for isaac lab?

kellyguo11:
Yes, maybe we should cherry pick the changes we had internally to update torch to 2.7.0. The issue with only updating it in setup.py is that on windows, it'll end up installing the CPU build, which doesn't work for most of our workflows. I can put together a PR for main to update the torch installation.

@ClemensSchwarke could you also help comment on the CNN question from @KingsleyLiu-NV? Are we ok with accepting his PR for the RSL RL repo?

ClemensSchwarke:
@kellyguo11 Ah I see, thanks a lot! I answered on slack, I think the CNN topic doesn't relate to this PR too much.

kellyguo11:
@ClemensSchwarke could you take a look at test_rsl_rl_wrapper.py to see why it's timing out?

ooctipus (approved):
I have tested locally myself, looks pretty good!

ooctipus:
I looked into the test a bit,

the test `test_random_actions` fails because it seems like wrapper has removed the `reset`. While this doesn't introduce any error in pure rsl-rl training pipeline this might break some user's workflow if they customized older rsl_rl and depended on the reset to construct their own environment.

The test is relying on `obs, extras = wrapper.reset` to test both observation and extra. 
But in newer wrapper code, you can't get observation and extra at the same time in any alternative without stepping physics. while you can get observation from `wrapper.get_observation`, extra is only accessible by calling `wrapper.step`. (Maybe we should call step rather than reset during the test?, or maybe the test doesn't reflect the new design philosophy and should be rewrite.)

I am unsure what the proper fix should be, probably rsl team has better idea, hope this provides some quick preview of source of test failure :))

ClemensSchwarke:
Thanks a lot for checking @ooctipus!!! I think we shouldn't keep the reset function just for the case that someone used it in a custom environment, given that older rsl_rl versions will not work with the new play/train scripts in general. Thus, someone with customized rsl_rl code would need to port that or use older lab versions anyways. However, if people think the reset function is important and can be useful, we can also add it again.

ooctipus:
I think keeping `reset` or not keeping `reset` all make sense. If we do not keep `reset`, then maybe can you help on fixing the test case so it is testing the new rsl-rl without `reset` instead testing the old one : ))

ClemensSchwarke:
I added the reset function again to the wrapper but not to rsl_rl, this way rsl_rl stays cleaner but the functionality is there. I also fixed the test to work with TensorDict. Let me know if that makes sense :)

KingsleyLiu-NV:
Hi @ClemensSchwarke @kellyguo11 , since tensordict as input is a key feature of rsl-rl 3.0, why the rsl-rl exporter does not have a conditional branch handling that: [exporter.py](https://github.com/isaac-sim/IsaacLab/blob/feature/rsl_rl_3.0_update/source/isaaclab_rl/isaaclab_rl/rsl_rl/exporter.py#L141-L171)?

ClemensSchwarke:
Hi @KingsleyLiu-NV, I am not sure if I understand you correctly. The exporter works for the current modules as it uses the dimension of the first layer of the network to infer the observation size. If you implement a different architecture, e.g., a CNN for images, the current exporter would not work. As we cannot cover all possible architectures users implement, the exporter needs to be adapted by the users in that case as well. However, we are thinking about moving the exporter inside rsl_rl to the corresponding modules. That would make it more clear that implementing a new module also necessitates implementing the corresponding export function. Does this answer your question?

kellyguo11:
hmm not quite sure why but CI doesn't seem to like this change...something is causing `ModuleNotFoundError: No module named 'pip._vendor.packaging._structures'` from pip

ClemensSchwarke:
> hmm not quite sure why but CI doesn't seem to like this change...something is causing `ModuleNotFoundError: No module named 'pip._vendor.packaging._structures'` from pip

That is super weird, not sure what is causing this. But I saw that the merges from main broke something, so I have to go through this PR again and see what changed. Maybe that fixes the issue somehow :sweat_smile:

ClemensSchwarke:
It was only a minor change, I don't think this will change the outcome of CI :/

ClemensSchwarke:
I tried to debug the failing CI, but I can't really make the connection to this PR. It seems like pip is breaking because the packaging package is uninstalled, due to skrl requiring a newer version. But this would not be related to rsl_rl so maybe not the right direction.

ooctipus:
this is very strange indeed, I also don't think its anything related to your code, Maybe do you think using a new PR can fix? 😅

kellyguo11:
yup quite strange. @ooctipus @ClemensSchwarke could one of you try to run the same pipeline from docker similar to what CI is doing and see if you can repro it locally?

ooctipus:
Local test running !

ooctipus:
all tests pass except for

`Job rsl_rl:Isaac-Velocity-Rough-Anymal-C-v0 failed to meet success criteria`

I ran it three times and it consistenly fails in my local test

@ClemensSchwarke would you mind take look a bit at this one
you should be able to run by 

`./isaaclab.sh -p -m pytest source/isaaclab_tasks/test/benchmarking/test_environments_training.py::test_train_environments[rsl_rl-task_spec94]`

ClemensSchwarke:
The test passes for me without any issues. What is failing for you @ooctipus ?

```
source/isaaclab_tasks/test/benchmarking/test_environments_training.py .                                                                                                                     [100%]{
  "rsl_rl:Isaac-Velocity-Rough-Anymal-C-v0": {
    "success": true,
    "msg": "",
    "reward": 15.3572,
    "reward_normalized": 2.1938857142857144,
    "reward_threshold": 7,
    "episode_length": 963.23,
    "episode_length_threshold": 900,
    "duration": 542.6443,
    "duration_threshold": 1800,
    "max_iterations": 300
  },
  "overall": {
    "totals": {
      "rsl_rl": 1
    },
    "successes": {
      "rsl_rl": 1
    },
    "failures_did_not_finish": {
      "rsl_rl": 0
    },
    "failures_did_not_pass_thresholds": {
      "rsl_rl": 0
    },
    "timestamp": "2025-08-25T11:41:08.402576",
    "tag": ""
  }
}


================================================================================== 1 passed in 608.53s (0:10:08) ==================================================================================
```

PR #2842, opened 2025-07-02
Adds contact point location reporting to ContactSensor

jtigue-bdai:
# Description

This PR: 

- Adds ContactSensorCfg.track_contact_points to toggle tracking of contact point locations between sensor bodies and filtered bodies.
- Adds ContactSensorCfg.max_contact_data_per_prim to configure the maximum amount of contacts per sensor body.
- Adds ContactSensorData.contact_pos_w data field for tracking contact point locations.

Fixes # (issue)

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #2983, opened 2025-07-19
Improves various points in contact sensor tests

ooctipus:
# Description

the logic in original contact sensor tests was a bit convoluted, 
1. disable_contact_processing was parameterized over many tests, even expensive ones. What is trying to test is just that creating scene will disable able it. This PR created one separate test just for that, and removed all other disable_contact_processing pytest parameterization

2. what suppose to be pytest parameters were hidden within a deeply nested helper function, and using multiple for loop to test each case. This can undermine the testing suites clarity when bug shows up as you don't know which hyparameter was causing the issue. I brought the for loop from deeper level to more surface level, whether or not to embed it into pytest parameter needs to confirm with original author and check for any oversight

3. added robust multi-envs, random contact location to stress test the collision. This tests are currently skipped due to it will fail. Need to work with sim and physx team to make sure they pass, and they will be serve for rigorous regression test.

4. refactored a bit of code, so all testing for contact sensors (testing air/contact time, testing contact positions, testing contact forces) can be configured nicely(avoid too many if else) 

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/test/sensors/test_contact_sensor.py
+++ b/source/isaaclab/test/sensors/test_contact_sensor.py
@@ -44,6 +44,15 @@ class ContactTestMode(Enum):
     """Enum to test the condition where the test object is not in contact with the ground plane (air time)."""
 
 
+class SensorTestMode(Enum):
+
+    JUST_CONTACT = 0
+
+    CONTACT_TIME_AIR_TIME = 1
+
+    CONTACT_POSITION = 2
+
+
 @configclass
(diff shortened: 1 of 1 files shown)


ooctipus:
This PR requires #2842 to be merged first, because it improves upon the test written for contact positions which doesn't exist in main yet

kellyguo11:
looks like this could also use a linter run

PR #3027, opened 2025-07-25
Internal template generator warning for case when user pip installed isaaclab

ooctipus:
# Description

User may install isaaclab with pip but also want to create an internal template in cloned isaaclab that doesn't sit in miniconda's python site-packages, and of course this internal template will not be detect by python, but we also don't want to create templates in site-package neither, nobody do development in site-packages

The proper fix is to support both internal and external template regardless the method user pick to install isaaclab, but for now, we will add the warning to documentation that internal template with pip isaaclab package is not supported

Fixes # (issue)

## Type of change

## Screenshots

Screenshot from 2025-07-25 13-42-25

## Checklist

### DIFF
--- a/docs/source/overview/developer-guide/template.rst
+++ b/docs/source/overview/developer-guide/template.rst
@@ -22,6 +22,12 @@ The template generator enables you to create an:
 * **Internal task**: A task that is part of the Isaac Lab repository. This approach should only be used to create
   new tasks within the Isaac Lab repository in order to contribute to it.
 
+  .. warning::
+   
+    If you installed Isaac Lab via pip, any internal task generated outside of the pip-installed environment may not
+    be discovered properly. We are working on better support, but please prefer external projects when using
+    isaac lab pip installation.
+
 Running the template generator
 ------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/source/overview/developer-guide/template.rst
```diff
@@ -22,6 +22,12 @@ The template generator enables you to create an:
 * **Internal task**: A task that is part of the Isaac Lab repository. This approach should only be used to create
   new tasks within the Isaac Lab repository in order to contribute to it.
 
+  .. warning::
+
+    If you installed Isaac Lab via pip, any internal task generated outside of the pip-installed environment may not
```
Mayankm96: what does it mean to make an internal task outside the environment?

ooctipus: ahhh I need to change a wording to be more clear. Maybe calling `internal template` instead of `internal task`? Called `internal task` because it is what the bullet point above is called.

so someone pip installed isaaclab in conda, and wanted to generate internal template. What the person did was to git clone the dummy isaaclab just to use ./isaaclab.sh and /scripts/, and ran ./isaaclab -n, creating an internal template. 

This internal template created this way sits in cloned dummy isaaclab, not in conda site-packages and the tasks in template will not get discovered by `import isaaclab_tasks`

PR #2836, opened 2025-07-01
Add example usage of RslRlSymmetryCfg in velocity environment

T-K-233:
## //Before Merge//

I would like to know if this is the proper way to implement this functionality. Please let me know if there are better ways to structure the code.

# Description

This PR creates `SymmetryCfg` and `SymmetryTermCfg` config classes that can be used to set symmetric configuration to the RSL-RL framework. An example data augmentation function, `symmetry_data_augmentation_function`, is also provided to implement corresponding symmetry terms.

This PR answers Issue #2835.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #3057, opened 2025-07-30
Adds RSL-RL symmetry example for locomotion training

Mayankm96:
# Description

This MR introduces the following:

* An `agent` argument to all scripts to allow selecting different entry points (each then get resolved to their respective settings file).
* Symmetry function for ANYmal robot for the locomotion task

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/scripts/reinforcement_learning/rl_games/play.py
+++ b/scripts/reinforcement_learning/rl_games/play.py
@@ -21,6 +21,9 @@ parser.add_argument(
 )
 parser.add_argument("--num_envs", type=int, default=None, help="Number of environments to simulate.")
 parser.add_argument("--task", type=str, default=None, help="Name of the task.")
+parser.add_argument(
+    "--agent", type=str, default="rl_games_cfg_entry_point", help="Name of the RL agent configuration entry point."
+)
 parser.add_argument("--checkpoint", type=str, default=None, help="Path to model checkpoint.")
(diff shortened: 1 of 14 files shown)


Mayankm96:
@Toni-SM for skrl it seems you have users specify the agent/algorithm name. I didn't want to change the behavior there but if you think above solution is generic (or also not - we can discuss), let me know how to proceed for skrl with it.

Mayankm96:
@T-K-233 , I see you were working on something similar in https://github.com/isaac-sim/IsaacLab/pull/2836. Let me know your thoughts on the approach in this MR.

Toni-SM:
> @Toni-SM for skrl it seems you have users specify the agent/algorithm name. I didn't want to change the behavior there but if you think above solution is generic (or also not - we can discuss), let me know how to proceed for skrl with it.

While this solution is more generic and doesn't require any code update if a new agent/algorithm is added... at the same time is more tedious to type:

`--agent rl_games_cfg_entry_point` vs `--agent PPO` 

Up to you... I am fine with the proposed change

Mayankm96:
> Why do you create a new entry point instead of a new task?

To avoid having a lot of duplications in the gym registry for every setting. If there are different algorithms (SAC, PPO etc) I think allowing users to pass entry point cfg makes it easier to scale.

ooctipus:
Do we need to update CHANGELOG or extension.toml for this one?

PR #3058, opened 2025-07-31
Adds basic validation tests for scale-based randomization ranges

louislelay:
# Description

While working on a task I made a tiny typo in the scale randomization range for the stiffness-gain parameter. The robot still spawned, but its behaviour was utterly bizarre. It only took a minute to spot the mistake, yet it made me realize we have no guard-rails for this sort of edge case.

This PR introduces a lightweight check that verifies, when operation=="scale", the lower bound is non-negative and the upper bound is not smaller than the lower one. Right now I cover the most common parameters (stiffness, damping, mass, tendon gains, etc.), basically anything that must stay positive to make physical sense.

If you’d like the same safeguard applied to other parameter types just let me know and I’ll happily extend the patch.

## Type of change

## Checklist


ooctipus:
Hi @louislelay,

This is amazing! thank you so much on moving toward this direction, please let me know if you need any my help : )). More than happy to support it!

ooctipus (approved):

louislelay:
Hey @ooctipus, thanks for the kind message haha!

I was thinking we should add this kind of safeguard to the other parameters in events as well. For some parameters, negative values should be allowed, but whenever there’s both a lower and an upper limit, we need to verify that one is greater than the other, not just for `scale`, but for every operation. Should we merge this PR as it is and open a new one for the rest, or do you think we should handle everything here?

ooctipus:
@louislelay 
I think we can merge this PR as it is, it does deal with all scale cases, so I think it is pretty standalone. 
But if you think the guard in other places are almost identical to this one, then maybe we put them into one PR. I am happy either way : ))

louislelay:
Well, I think we should merge this one in this case, @ooctipus! I’ll open another PR later this week once I have a bit more time to work on it :) 
I’ll mention you when I open the other PR so you can review it!

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -35,6 +35,30 @@
     from isaaclab.envs import ManagerBasedEnv
 
 
+def _validate_scale_range(
```
Mayankm96: Could you please move the private functions to the end of the file?

Similar to here: https://github.com/isaac-sim/IsaacLab/blob/4b6e27582620b6a3c57fad4b0b93469f0be1893d/source/isaaclab/isaaclab/envs/mdp/events.py#L1445-L1450

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -35,6 +35,30 @@
...
+    params: tuple[float, float] | None,
+    name: str,
+    *,
+    allow_negative: bool = False,
+    allow_zero: bool = True,
+) -> None:
+    """Sanity–check (low, high) tuples used with `operation=="scale"`.
```
Mayankm96: Would be good to clarify all the args properly and the purpose of this function in the docstrings.

Mayankm96:
Is there a perf cost for these sanity checks? I wonder if they should only happen "once" on construction to not lose performance because of the introduced branching checks.

ooctipus:
My understanding is that these operations are for `mode: startup` use case only, if it is `mode: reset`, then it is definitely a problem...

louislelay:
Hey @Mayankm96 and @ooctipus, actually, there are a few of these that are used [sometimes as startup](https://github.com/isaac-sim/IsaacLab/blob/19b24c780ea2279ea42bcd8a4c8a8dd8bd75955e/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/inhand/inhand_env_cfg.py#L198) and [sometimes as reset](https://github.com/isaac-sim/IsaacLab/blob/19b24c780ea2279ea42bcd8a4c8a8dd8bd75955e/source/isaaclab_tasks/isaaclab_tasks/direct/shadow_hand_over/shadow_hand_over_env_cfg.py#L39), so that’s definitely an issue. These current changes then don’t really make sense anymore, so I should probably look as Mayank suggested for another way to check at construction.

If you have any pointers about that, I’d gladly take them!

ooctipus:
Hey @louislelay Thank you for checking that there are examples of using them as `mode:reset`, I was under impression that only `mode:startup` was the usage in isaaclab. But I was wrong.

For making a sanity check at construction time rather than call time
please take a look at here. [example of sanity check in constructor](https://github.com/isaac-sim/IsaacLab/blob/19b24c780ea2279ea42bcd8a4c8a8dd8bd75955e/source/isaaclab/isaaclab/envs/mdp/events.py#L176-L234)

You can move all the check into the `__init__` and check `cfg.params` passed in from __init__ input arguments

louislelay:
Thanks @ooctipus, that’s exactly the kind of pointer I was looking for :)

I’ll take the time to do it tomorrow -- I don’t have any GPU at hand right now!

louislelay:
I've updated a few things but haven’t tested every function during use yet. I’ll do that later tonight, so it’s not ready for review for now!

louislelay:
Hey @Mayankm96 and @ooctipus, sorry for the delay, it’s been a really busy week. I tried the OpenAI Shadow Hand environment since it uses all the changed events. I realized parameters shouldn’t be checked if they’re not used, so I updated the tests accordingly and added one to ensure the user can only choose scale, add, or abs.

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -592,97 +648,126 @@ def randomize_joint_parameters(
...
+            ValueError: If the upper bound is less than the lower bound.
+        """
+        super().__init__(cfg, env)
+
+        # extract the used quantities (to enable type-hinting)
+        self.asset_cfg: SceneEntityCfg = cfg.params["asset_cfg"]
+        self.asset: RigidObject | Articulation = env.scene[self.asset_cfg.name]
+        # check for valid operation
+        if cfg.params["operation"] == "scale":
+            if "stiffness_distribution_params" in cfg.params:
```
ooctipus:
```suggestion
            if "friction_distribution_params" in cfg.params:
```

ooctipus: Sorry for reviewing on this PR late, there seems to be a small typo can you please verify? @louislelay . Everything else looks pretty good lets get this PR merged!

louislelay: Good catch @ooctipus, thanks!

ooctipus:
Everything looks good!! 
tests all passing, last thing
@louislelay Could you please update CHANGELOG and extension.toml in isaaclab?

louislelay:
Hey @ooctipus, thank you! I just updated those :)

ooctipus (approved):

ooctipus (approved):

PR #3062, opened 2025-07-31
Simply remove print statements. 

constant5:
If we want these to display consider using logging.info.


source/isaaclab_tasks/isaaclab_tasks/direct/automate/automate_algo_utils.py
```diff
@@ -11,15 +11,14 @@
...
-print("Python Executable:", sys.executable)
-print("Python Path:", sys.path)
-
 from scipy.stats import norm
 
 from sklearn.gaussian_process import GaussianProcessRegressor
 from sklearn.mixture import GaussianMixture
 
-base_dir = os.path.abspath(os.path.join(os.path.dirname(os.path.abspath(__file__)), "."))
+base_dir = os.path.abspath(
```
Mayankm96, line 19: Not sure why but there a bunch of these lint changes. Is this intended?

ooctipus:
@constant5 can you please not include line breaking in file change if you are only removing print statements?

Mayankm96:
We have switched to logging module. Thank you!

PR #3067, opened 2025-08-01
Skip operation to set velocity if kinematic is enabled for the rigido…

ooctipus:
Reset velocity is not a valid operation for RigidObject if kinematic enabled flag is True

This PR makes set-velocity an no-op if detected kinematic_enabled is True

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
+++ b/source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
@@ -56,6 +56,8 @@ class RigidObject(AssetBase):
             cfg: A configuration instance.
         """
         super().__init__(cfg)
+        prim = sim_utils.find_matching_prims(self.cfg.prim_path)
+        self.is_kinamtic_enabled = prim[0].GetAttribute("physics:kinematicEnabled").Get()
 
     """
     Properties
@@ -292,6 +294,8 @@ class RigidObject(AssetBase):
             root_velocity: Root center of mass velocities in simulation world frame. Shape is (len(env_ids), 6).
             env_ids: Environment indices. If None, then all indices are used.
         """
+        if self.is_kinamtic_enabled:
+            return
         self.write_root_com_velocity_to_sim(root_velocity=root_velocity, env_ids=env_ids)
 
     def write_root_com_velocity_to_sim(self, root_velocity: torch.Tensor, env_ids: Sequence[int] | None = None):
@@ -304,6 +308,8 @@ class RigidObject(AssetBase):
             root_velocity: Root center of mass velocities in simulation world frame. Shape is (len(env_ids), 6).
             env_ids: Environment indices. If None, then all indices are used.
         """
+        if self.is_kinamtic_enabled:
+            return
         # resolve all indices
         physx_env_ids = env_ids
         if env_ids is None:
@@ -335,6 +341,8 @@ class RigidObject(AssetBase):
             root_velocity: Root frame velocities in simulation world frame. Shape is (len(env_ids), 6).
             env_ids: Environment indices. If None, then all indices are used.
         """
+        if self.is_kinamtic_enabled:
+            return
         # resolve all indices
         if env_ids is None:
             local_env_ids = slice(env_ids)
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.41.3"
+version = "0.41.5"
 
 # Description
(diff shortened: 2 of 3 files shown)


source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, cfg: RigidObjectCfg):
             cfg: A configuration instance.
         """
         super().__init__(cfg)
+        prim = sim_utils.find_matching_prims(self.cfg.prim_path)
+        self.is_kinamtic_enabled = prim[0].GetAttribute("physics:kinematicEnabled").Get()
```
Mayankm96, line 60: This check should happen inside `initialize_impl` function. It could be a user creates the constructor and switches the behavior someplace else (eg: inside spawn events).

source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
```diff
@@ -292,6 +294,8 @@ def write_root_velocity_to_sim(self, root_velocity: torch.Tensor, env_ids: Seque
             root_velocity: Root center of mass velocities in simulation world frame. Shape is (len(env_ids), 6).
             env_ids: Environment indices. If None, then all indices are used.
         """
+        if self.is_kinamtic_enabled:
```
Mayankm96, line 297: Calling the variable just "is_kinematic" is sufficiently concise. Don't need to say enabled.

source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
```diff
@@ -304,6 +308,8 @@ def write_root_com_velocity_to_sim(self, root_velocity: torch.Tensor, env_ids: S
...
```
Mayankm96, line 311: We can't handle the case where some bodies in the view are kinematic and some aren't? If so, maybe the docstrings should clarify this with a note.

Mayankm96 (changes requested):
The behavior also needs to be propagated to `RigidObjectCollection` class. Some tests would be nice.

In general, I feel instead of doing a No-Op here, the user should NOT call the function to set velocities at all. I think if they do call set velocity accidentally, we should throw a warning inside our classes complaining that this isn't good and they should ideally handle this case themselves properly. Otherwise, we might introduce unnecessary branching in our interfaces for such usecases.

source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
Mayankm96: Also this can be a "data" attribute? You should be able to get it from the PhysX view for individual prims.

ooctipus:
Thank you for reviewing, very insightful as always. I think your point is valid, I had the same concern with you, and I agree the most proper fix should be preventing velocity to be set at all if rigidbody is kinematic.

I think the main issue regarding preventing user to call function is that some of our mdp will break.
For example, the `reset_root_state_uniform` were supposedly working for both Articulation and RigidBody, Now we will have to check if it is Articulation or RigidBody, only if it is RigidBody should we check again if it's kinematic. I didn't really like that level of if-else branch either, so didn't after that route initially, instead went for quick and easy NO-OP.

If we are going to do this more properly, I think we need to break up some of our mdps to better reflect pose and velocity should be treated differently depending on the property of asset user supplied. For example, breaking up `reset_root_state_uniform` into two parts, 1. `reset_root_pose_uniform` and 2.`reset_root_vel_uniform` and user should not choose `reset_root_vel_uniform` if Articulation fixed_root or RigidObject is kinematic.

Say if we assume user should be responsible to choose right operation, then in the case if user accidently choose wrong operation:
 
| 1. Are you thinking no-op + warning, then I can keep develop this branch continuing NO-OP to RigidObjectCollection and writing test cases.

| 2. Or are you thinking of  RigidObject-Class should still do what user instruct anyway instead of NO-OP, then throw a warning

If 1 is the case, I can keep propagating No-Op to RigidObjectCollection, but if 2 is the case, maybe No-Op is not needed, This PR will  just add warning.

Another PR will be opened for changes coming from mdp side rather than changing RigidObject, RigidObjectCollection.

Let me know your thought : ))

source/isaaclab/isaaclab/assets/rigid_object/rigid_object.py
ooctipus: Yes I don't think we can, I tested multi-asset-spawner, where some cfg has is_kinematic_enabled = True and some has  is_kinematic_enabled = False, I think it will just apply the is_kinematic_enabled argument from first object to entire view. I wonder does RigidObjectView support mixed kinematic? if it does, we should definitely make a pr to support it.

PR #3071, opened 2025-08-01
Fixes '!python_exe!' is not recognized as an internal or external command

Chunde:
# Description

This pull request fixes an issue where the isaaclab.bat script would fail with the error '!python_exe!' is not recognized as an internal or external command, operable program or batch file. when running commands that require the Python executable. The root cause was that the python_exe variable, set in the :extract_python_exe subroutine, was not correctly propagated back to the main context with delayed variable expansion enabled.

## Motivation and context:
This bug prevented users from running commands such as` isaaclab.bat --new` and other commands that rely on the Python executable, breaking core functionality of the batch script on Windows.

Fixes # (issue)

## Summary of the change:

Added setlocal EnableDelayedExpansion at the start of the :extract_python_exe subroutine.
Used endlocal & set "python_exe=%python_exe%" at the end of the subroutine to propagate the variable back to the parent context.
This ensures that !python_exe! is correctly expanded and available after calling :extract_python_exe.

## Screenshots

Before the fix, the script will error out:
Screenshot 2025-08-01 002050

After the fix, the script run correctly:

Screenshot 2025-08-01 002125

### DIFF
--- a/isaaclab.bat
+++ b/isaaclab.bat
@@ -95,8 +95,6 @@ if not exist "%python_exe%" (
     echo %tab%2. Python executable is not available at the default path: %ISAACLAB_PATH%\_isaac_sim\python.bat
     exit /b 1
 )
-echo [DEBUG] python_exe is set to: !python_exe!
-
 goto :eof
 
 
@@ -317,7 +315,7 @@ goto :eof
 
 rem Main
 :main
-setlocal EnableDelayedExpansion
+
 rem check argument provided
 if "%~1"=="" (
     echo [Error] No arguments provided.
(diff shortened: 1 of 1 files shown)


kellyguo11:
Thanks for catching the issue! This fix seems to have some side effects with commands like `isaaclab.bat -i`, causing 
```
[INFO] Installing extensions inside the Isaac Lab repository...
'-m' is not recognized as an internal or external command,
operable program or batch file.
```

I believe the proper fix is to use `call` instead. We will incorporate these fixes into the official Isaac Lab 2.2 release.

Issue #3070, opened 2025-08-01
[Proposal] Enhanced Asset Selection Strategy for Multi-USD Multi-Agent Environments


Kyu3224:
### Proposal

Enhanced Asset Selection Strategy for Multi-USD Multi-Agent Environments

### Motivation

I am using multi-agent robots that require multiple robot assets, and I utilize the multi-USD spawning feature. Currently, only "random" or "sequential" (e.g., 0, 1, 2, 0, 1, 2) selection methods are supported, which limits the pool of robot combinations that can be spawned. For instance, if I want to spawn 2000 environments using two multi-USD assets, each consisting of 4 different robots, using the sequential method would result in repeated patterns like (0, 1, 2, 0, 1, 2), which leads to environment compositions such as (0, 0), (1, 1), (2, 2), and so on.

Moreover, with the random method, I cannot control the distribution ratio of the robot types being spawned. To address this, I added a new "split" option in addition to "sequential", which allows assets to be spawned in grouped segments, such as (0, 0, 1, 1, 2, 2). This way, by combining "sequential" and "split" methods, we can deterministically structure the environment setup—for example: (0, 1, 2, 0, 1, 2) + (0, 0, 1, 1, 2, 2) results in compositions like (0, 0), (1, 0), (2, 1), (0, 1), (1, 2), (2, 2), which allows for more structured and controlled spawning across environments.

### Alternatives

I will submit this as a pull request.

### Checklist

- [X] I have checked that there is no similar issue in the repo (**required**)

### Acceptance Criteria

Add the criteria for which this task is considered **done**. If not known at issue creation time, you can add this once the issue is assigned.

- [ ] Criteria 1
- [ ] Criteria 2

Mayankm96:
More generically, we can have a function in the config that users can set to decide how the choices are made?

Kyu3224:
Hi @Mayankm96, thank you for your comment!

In my project, I initially implemented a simple method to switch between two selection modes. 


<img width="1429" height="244" alt="Image" src="https://github.com/user-attachments/assets/a2654ed4-240b-4b79-a221-3b42eff09ede" />

But after considering your suggestion, I agree that allowing a function in the config would be a more flexible and extensible approach.

For example, the current "sequential" mode could be expressed as a simple function like:

```
def choice_fn(x, num_assets):
    return x % num_assets
```

And more advanced strategies — such as controlling the distribution of assets — could also be supported via user-defined functions.

[...]

PR #3083, opened 2025-08-03
Added method for spawning multi-asset

Kyu3224:
# Description
Regarding issue #3070, I added support for spawning objects in each environment within a multi-USD (or asset) setup using various methods. Users can define custom functions in wrappers/utils.py or any other desired location, specify them in the config, and have them dynamically loaded using hydra.utils.get_method. The loaded choice_fn will then iterate through all prim_paths and assign asset indices based on the defined rules.

Following @Mayankm96 's advice, this design allows for user-defined logic to determine how assets are distributed across environments, providing both built-in methods (e.g., sequential, split) and the flexibility to extend with custom strategies.

I aimed to align this as closely as possible with the Isaac Lab code style and direction, but I'm not sure if it worked well, so I would appreciate your feedback. Basically, I have implemented sequential and split methods, so users can select between these two by default, and they can also add and use additional methods as needed.

## Type of change

## Checklist

- [X] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [X] I have made corresponding changes to the documentation
- [X] My changes generate no new warnings
- [X] I have added my name to the `CONTRIBUTORS.md` or my name already exists there


source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py
```diff
@@ -62,6 +85,29 @@ class MultiUsdFileCfg(UsdFileCfg):
     random_choice: bool = True
```
ooctipus: Why don't we make random choice an option in choice_method?

ooctipus (approved):
@Kyu3224 Thanks for the PR, this is a very interesting feature
Overall looks good to me.

My only edit comment is that random choice can also be an option of choices., that we can remove extra branches to check if random_asset is true or false.

On another scope:

I am thinking about this problem from a broader picture: what you are currently doing is to make sure multi-asset spawner are spawned following a rule that independent env_id ends up with right combination.

but maybe an alternative question to ask is that, how do we spawn several multi-assets such that asset finds the right pairs and avoid undesired pairs. Maybe what we are missing is to support spawning rigidBodyCollection within multi asset, then you can even more freely to configure what collection combinations are allow or not allowed when sampling multi-asset object. Curious of your thought

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py
Kyu3224: Good idea! Adding random as an option in choice_method makes the code cleaner.

Kyu3224:
Hi @ooctipus , I think the suggestion about simplifying with random as an option is a great idea.

Regarding the idea about spawning rigidBodyCollection within multi-asset, could you give an example of how that would work? Currently, I’m focusing only on spawning articulation systems using different robots, but I think it could be extended to a more generalized approach.

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py
```diff
@@ -34,11 +34,25 @@ class MultiAssetSpawnerCfg(RigidObjectSpawnerCfg, DeformableObjectSpawnerCfg):
     assets_cfg: list[SpawnerCfg] = MISSING
     """List of asset configurations to spawn."""
 
-    random_choice: bool = True
-    """Whether to randomly select an asset configuration. Default is True.
+    choice_method: str = "random_choice"
```
Mayankm96: Instead of using a string here, what do you think about allowing this to be a function that users can customize externally? We could still support both strings and callables, but making it callable would be more flexible and future-proof. That way, if new choice methods are added later, users won’t need to modify IsaacLab’s source code to use them.

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers.py
```diff
@@ -5,14 +5,15 @@
...
-import random
 import re
 from typing import TYPE_CHECKING
 
 import carb
 import isaacsim.core.utils.prims as prim_utils
 import isaacsim.core.utils.stage as stage_utils
 from isaacsim.core.utils.stage import get_current_stage
+
+from hydra.utils import get_method
```
Mayankm96: Would avoid using hydra inside the core base. Hydra is an optional dependency used only for env/agent configuration.

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py
Mayankm96: What if user wants to pass custom args to this function. For example, 10% of one asset, 20% of other, 50% of next one and so on? We should have a mechanism for it.

source/isaaclab/isaaclab/sim/spawners/wrappers/utils.py
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
```
Mayankm96: Module level docs are nice but I suggest moving args behavoir to per-function implementation. Most editors would then show the function docs directly if you hover over it.

Mayankm96: Also would be good to avoid yet another generic utils.py name. Maybe `choices.py` is better for following.

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers_cfg.py
Kyu3224: Hi, I’ve added support for both `str` and `callable`. Now, `choice_method` can be specified either as a string or as a function (after importing it). 

Also, `random_choice` adjusted the sampling probabilities using `random.choices` with weights, but it resampled every iteration. With `deterministic_choice`, the env_idx is now selected strictly according to the given weights via `kwargs`, ensuring the specified ratios are followed.

source/isaaclab/isaaclab/sim/spawners/wrappers/wrappers.py
Kyu3224: Thanks for pointing that out! I’ve replaced it with Python’s built-in `importlib`.

source/isaaclab/isaaclab/sim/spawners/wrappers/utils.py
Kyu3224: I renamed the file from `utils.py` to `choices.py` and added docstrings to each function as suggested.

Kyu3224:
image

I tried to incorporate all the suggested comments as much as possible. 
Please let me know if there are any further improvements that can be made.
Additionally, I added a section in the initialization to show the distribution of agents for each asset (30 environments with 4 assets).

ooctipus:
Thank you @Kyu3224 this is very nice of you, and thank you @Mayankm96 for the nice suggestion!

ooctipus (approved):

ooctipus:
@Kyu3224 

Big Big sorry!!!!!

We are planning to do a big refactor on spawner + cloner that would 1. drop the requirement for using isaac_sim 2. allow more flexible combination of assets. 3. faster cloning time.

which means this PR will probably not be applicable soon. 

Sorry again for all the work, I really appreciate reviewing your PR and discussions.

Feel free to close it, or I can help close it when the refactor comes in : )))

PR #3087, opened 2025-08-04
RL libraries training performance

Toni-SM:
# Description

This PR updates the agent configuration (to be as similar as possible) for the `Isaac-Humanoid-v0` task to ensure a more accurate comparison of the RL libraries when generating the [Training Performance](https://isaac-sim.github.io/IsaacLab/main/source/overview/reinforcement-learning/rl_frameworks.html) table.

To this end:
1. A common *Training time* info (e.g: `Training time: 278.38 seconds`) is printed when running existing `train.py` scripts. Currently the RL libraries output training information in different formats and extends.
2. A note is added to involved agent configurations to notify/ensure that any modification should be propagated to the other agent configuration files.
3. The commands used to benchmark the RL libraries is added to docs, for clearness and repro.

---

With the updated config, the following times are obtained, with an NVIDIA GeForce RTX 4090

```
+--------------------+-----------------+
| RL Library         | Time in seconds |
+====================+=================+
| RL-Games           | 195             |
+--------------------+-----------------+
| RSL RL             | 192             |
+--------------------+-----------------+
| Stable-Baselines3  | 278             |
+--------------------+-----------------+
| SKRL               | 192             |
+--------------------+-----------------+
```

@kellyguo11 please update table for RTX PRO 6000 GPU, since I don't have access to such GPU

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/rl_games_ppo_cfg.yaml
+++ b/source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/rl_games_ppo_cfg.yaml
@@ -1,3 +1,7 @@
+# This file defines the agent configuration used to generate the "Training Performance" table in
+# https://isaac-sim.github.io/IsaacLab/main/source/overview/reinforcement-learning/rl_frameworks.html.
+# Ensure that the configurations for the other RL libraries are updated if this one is modified.
+
 params:
   seed: 42
 
--- a/source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/rsl_rl_ppo_cfg.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/rsl_rl_ppo_cfg.py
@@ -3,6 +3,13 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+"""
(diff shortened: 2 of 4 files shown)


source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/rl_games_ppo_cfg.yaml
```diff
@@ -45,13 +49,13 @@ params:
     device_name: 'cuda:0'
     multi_gpu: False
     ppo: True
-    mixed_precision: True
+    mixed_precision: False
```
Mayankm96, line 52: This configuration initially came from IsaacGymEnvs. Wondering if we should modify it like this if still plan to compare IsaacLab to isaacGym implementation of this environment.

Toni-SM: Mmm,

Enabling `mixed_precision`, for rl_games and skrl, reduces training times...making comparison unfair/inaccurate.  

```
rl_games: 190
skrl: 188
```

If it is desired comparison with IsaacGym, possible solutions are:
* Update IsaacGym config to match IsaacLab
* Use another task (e.g.: `Isaac-Humanoid-Direct-v0`) for RL libraries training performance in IsaacLab

ooctipus:
@Toni-SM This seems very useful can you resolve conflict and we can try merge it in : ))

Issue #3088, opened 2025-08-04
[Question] IMU Sensor Requires RigidBodyAPI on Target Prim


GiulioRomualdi:
I'm not sure if this is expected behavior or a bug.

I'm trying to attach an IMU sensor to a frame that does **not** have a `RigidBodyAPI`. Specifically, the prim path is:

```
/ergoCub/root_link/torso_1/waist_imu_0
```

As shown in the image below, the prim exists and has valid transform data, but no `RigidBodyAPI` component:

<img width="1390" height="1433" alt="Image" src="https://github.com/user-attachments/assets/07fe25c7-05c9-4fb4-b6ba-df5a01beecb3" />

When I try to create the IMU sensor using:

```python
imu = ImuCfg(prim_path="{ENV_REGEX_NS}/Robot/root_link/torso_1/waist_imu_0", debug_vis=False)
```

I get the following runtime error:

```
RuntimeError: Failed to find a RigidBodyAPI for the prim paths: /World/envs/env_.*/Robot/root_link/torso_1/waist_imu_0
```

This error originates from this section of the code:
https://github.com/isaac-sim/IsaacLab/blob/77a649844d180fc7706ad6a78fe0617b311dc6df/source/isaaclab/isaaclab/sensors/imu/imu.py#L137-L140

### Question

Is it **required** that the target prim for an IMU sensor has `RigidBodyAPI`?
I was expecting it to be possible to attach IMUs to arbitrary (static or articulated) frames, not just rigid bodies.

I'm using Isaac Lab v2.1.1 and IsaacSim 4.5.0

GiulioRomualdi:
I tried to dig a bit into the problem and I realized that the IMU class calls `create_rigid_body_view()` and then, every frame, uses that view’s `get_transforms()`, `get_velocities()`, and `get_coms()` methods. All three calls rely on the prim owning a `UsdPhysics.RigidBodyAPI` handle. Indeed, the key point is that `PhysX` only tracks pose, velocity, and COM buffers for rigid bodies. If the prim pointed is just an `Xform` (as in my case) it has no rigid-body data it is not possible to call that function.

TLDR: the sensor grabs its data straight from PhysX’s rigid-body buffers, so it can only attach to prims that are rigid bodies. Anything else fails because the required physics data isn’t there.



---


I'm currently modifing the imu sensor to give the user the possibility to set a prim that is just an `Xform`.

Mayankm96:
[...]

PR #3094, opened 2025-08-05
Update the IMU sensor implementation to support general prims

GiulioRomualdi:
# Description

Refactored IMU sensor to support non-rigid prims by finding the nearest rigid ancestor and computing the fixed transform. Maintains compatibility with existing code.

Fixes https://github.com/isaac-sim/IsaacLab/issues/3088

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/sensors/imu/imu.py
+++ b/source/isaaclab/isaaclab/sensors/imu/imu.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING
 
 import isaacsim.core.utils.stage as stage_utils
 from isaacsim.core.simulation_manager import SimulationManager
-from pxr import Usd, UsdGeom, UsdPhysics, Gf, Sdf
+from pxr import Gf, Sdf, Usd, UsdGeom, UsdPhysics
 
 import isaaclab.sim as sim_utils
 import isaaclab.utils.math as math_utils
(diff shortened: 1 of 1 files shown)


louislelay:
Hi @GiulioRomualdi, nice work!

Do you think it could be generalized to handle not only Xforms but also Spheres? I tested your PR on a model converted from MJCF and applied it to the sites, but the transform was applied to the Sphere instead of the Xform, so no offset was computed.

**See below:**
image

ooctipus:
Thank you for the PR @GiulioRomualdi , this is a very nice observation!

After thinking about this feature for long time, I have a bit mix feeling on this.

This definitely making sensor attachment a lot more easier which is a great bonus, though I am a bit curious how clear to user this feature is. What if there is no rigid ancestor, will users complain because they were under impression they can attach sensor anywhere? Also, if we are going to move forward with this, we better make sure it can work beyond just xforms, and works with other sensors as well, etc `ray caster`, `camera` ...., maybe this should be a SensorBase implementation?

I am happy to move forward and help on this direction if we figure out a way to apply this feature too all sensors, and we can attach to any prim underneath the rigid object. But it seems right now it is still in a early exploitative phase. Anyway, thank you for bring inspiring and interesting idea and actively contribute to IsaacLab. I am very excited to review and help :))

Mayankm96:
Closing this MR as it has been merged in #3864.

Thank you for contributing the fix; it was very useful!

Issue #2977, opened 2025-07-18
[Bug Report] Termination Overcounting Caused by Missing Log Buffer Reset in manager_based_rl_env.py


Kyu3224:
### Describe the bug

In the current implementation of `manager_based_rl_env.py`, the `self.extras['log']` buffer retains termination statistics from previous timesteps if no environment reset occurs during a rollout. Since rsl_rl’s logging system averages these values per rollout, this leads to duplicated counting of terminations and inflated termination ratios.

Specifically, even when `reset_env_ids` is empty, the termination logs are not cleared or updated appropriately, causing multiple counting of the same termination event across subsequent timesteps.

This issue is demonstrated in the attached logs and visualizations, where termination counts exceed the expected values based on actual resets.

### Steps to reproduce

I conducted experiments using an environment that logs 2 types of rewards and 3 types of terminations, excluding the timeout termination. The setup uses the `rsl_rl` framework with PPO, with 200 environments and `num_steps_per_env = 7`.

Unfortunately, I’m unable to share the specific environment due to it being part of a proprietary research platform, but I believe the issue is not limited to this particular environment.

`self.extras['log']` is passed to the [def log](https://github.com/leggedrobotics/rsl_rl/blob/main/rsl_rl/runners/on_policy_runner.py) function in `on_policy_runner.py` of `rsl_rl`, where the logged values are averaged and printed. However, if the environment does not reset, it reuses the previous termination and log values in the averaging process. This can lead to inaccurate logging.

I will attach a log visualization comparing the original code with the version I modified for debugging, to better illustrate this behavior.

[Current code](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab/isaaclab/envs/manager_based_rl_env.py#L216-231)

```python
        # -- reset envs that terminated/timed-out and log the episode information
        reset_env_ids = self.reset_buf.nonzero(as_tuple=False).squeeze(-1)
        if len(reset_env_ids) > 0:
            # trigger recorder terms for pre-reset calls
            self.recorder_manager.record_pre_reset(reset_env_ids)

            self._reset_idx(reset_env_ids)
            # update articulation kinematics
            self.scene.write_data_to_sim()
            self.sim.forward()

            # if sensors are added to the scene, make sure we render to reflect changes in reset
            if self.sim.has_rtx_sensors() and self.cfg.rerender_on_reset:
                self.sim.render()

            # trigger recorder terms for post-reset calls
[...]

PR #3107, opened 2025-08-07
Fixed Termination Manager logging to report aggregated percentage of environments done due to each term. 

ooctipus:
# Description

Currently Termination Manager write current step's done count for each term if reset is detected. This leads to two problem. 
1. User sees different counts just by varying num_envs
2. the count can be over-count or under-count depending on when reset happens, as pointed out in #2977 (Thanks, @Kyu3224)

The cause of the bug is because we are reporting current step status into a buffer that suppose to record episodic done. So instead of write the entire buffer base on current value, we ask the update to respect the non-reseting environment's old value, and instead of reporting count, we report percentage of environment that was done due to the particular term.

Test on Isaac-Velocity-Rough-Anymal-C-v0

Before fix:
Screenshot from 2025-08-06 22-16-20
Red: num_envs = 4096, Orange: num_envs = 1024

After fix:

Screenshot from 2025-08-06 22-16-12
Red: num_envs = 4096, Orange: num_envs = 1024

Note that curve of the same color ran on same seed, and curves matched exactly, the only difference is the data gets reported in termination.  The percentage version is a lot more clear in conveying how agent currently fails, and how much percentage of agent fails, and shows that increasing num_envs to 4096 helps improve agent avoiding termination by `base_contact` much quicker than num_envs=1024. Such message is a bit hard to tell in first image.

## Checklist


source/isaaclab/isaaclab/managers/termination_manager.py
```diff
@@ -182,7 +182,7 @@ def get_term(self, name: str) -> torch.Tensor:
         Returns:
             The corresponding termination term value. Shape is (num_envs,).
         """
-        return self._term_dones[name]
+        return self._term_dones[name, self._term_names.index(name)]
```
Mayankm96: This indexing here is not great. We use the value from here to give termination rewards in different environments. Doing the `.index` every time for reward computation may lead to slow downs (as it searches over the list in O(n) fashion).

source/isaaclab/isaaclab/managers/termination_manager.py
```diff
@@ -61,9 +61,7 @@ def __init__(self, cfg: object, env: ManagerBasedRLEnv):
         # call the base class constructor (this will parse the terms config)
         super().__init__(cfg, env)
         # prepare extra info to store individual termination term information
-        self._term_dones = dict()
-        for term_name in self._term_names:
-            self._term_dones[term_name] = torch.zeros(self.num_envs, device=self.device, dtype=torch.bool)
+        self._term_dones = torch.zeros((self.num_envs, len(self._term_names)), device=self.device, dtype=torch.bool)
```
Mayankm96, line 65: Is there a reason to change this as a dict of tensors to a single tensor?

source/isaaclab/isaaclab/managers/termination_manager.py
```diff
@@ -139,9 +137,10 @@ def reset(self, env_ids: Sequence[int] | None = None) -> dict[str, torch.Tensor]
             env_ids = slice(None)
         # add to episode dict
         extras = {}
-        for key in self._term_dones.keys():
+        last_episode_done_stats = self._term_dones.float().mean(dim=0)
+        for i, key in enumerate(self._term_names):
             # store information
-            extras["Episode_Termination/" + key] = torch.count_nonzero(self._term_dones[key][env_ids]).item()
+            extras["Episode_Termination/" + key] = last_episode_done_stats[i].item()
```
Mayankm96, line 144: If you want the ratio, isn't that simply?

```python
self._term_dones[key][env_ids].sum() / len(env_ids)
```

ooctipus: Thanks for reveiwing!
I guess, doing with env_ids will be viewing ratio of resetting environments. I thought maybe report stats of all environment can be a bit more nicer as user can verify from the graph that all terms sum up to 1. Of course you can do `self._term_dones[key].sum() / self.env.num_envs` as well

But it seems like if this approach is what I after, do it in one tensor operation seems quite nice, both speed wise and memory utility wise.

source/isaaclab/isaaclab/managers/termination_manager.py
ooctipus: Ahhh if you like the single tensor approach I can also store one more name -> idx dict, and make change here O(1), I wasn't aware of termination rewards, (my bad!!) and thought this function may be used infrequently. I should do a global search next time rather than assume!

source/isaaclab/isaaclab/managers/termination_manager.py
ooctipus: I thought it `last_episode_done_stats = self._term_dones.float().mean(dim=0)` this operation is very nice and optimized, thats why I did it, but if you think dict is more clear I can revert back : ))

ooctipus:
@Mayankm96 
I did my best to optimize the operations in compute and reset,
index all changed from O(n) to O(1) -> this part not captured in benchmark below

benchmarked task: velocity rough anymal c, 4096 envs, 1000 steps

before change :
one step total: 79.038
```
| termination.compute     |       0.229 ms|
| termination.reset       |       0.007 ms|
```
after change:

one step total: 80.187 ms
```
| termination.compute     |       0.274 ms|  slower due to one extra operation
| termination.reset       |       0.004 ms|  twice as fast
```

the cost for this PR is about 0.04 ms / 80, 0.05%, mostly due to compute needs modify other terms's done to correctly update the done buffer. I think this is reasonable

PR #3110, opened 2025-08-07
Disallows string value written in sb3_ppo_cfg.yaml get evaluated in process_sb3_cfg

ooctipus:
# Description

This PR adds stricter interpretation rules to value specified in sb3_ppo_cfg.yaml, disallowing eval on any dict, which my contain arbitrary code that makes program vulnerable.

Now, `eval` is got rid of, only str that start with `nn.`  can be used to only import the module from torch.nn. That seems to cover all usage for lab so far, I can make more accommodations if there are more cases but it seems like it is currently sufficient. 

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_rl/docs/CHANGELOG.rst
+++ b/source/isaaclab_rl/docs/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+0.2.4 (2025-08-07)
+~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Disallowing string value written in sb3_ppo_cfg.yaml gets eval() in :meth:`~isaaclab_rl.sb3.process_sb3_cfg`.
+
+
 0.2.3 (2025-06-29)
 ~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab_rl/docs/CHANGELOG.rst
+++ b/source/isaaclab_rl/docs/CHANGELOG.rst
@@ -7,8 +7,8 @@ Changelog
 Fixed
 ^^^^^
 
-* Disallowed string values in ``sb3_ppo_cfg.yaml`` from being passed to ``eval()`` in 
-  :meth:`~isaaclab_rl.sb3.process_sb3_cfg`. This change prevents accidental or malicious 
+* Disallowed string values in ``sb3_ppo_cfg.yaml`` from being passed to ``eval()`` in
+  :meth:`~isaaclab_rl.sb3.process_sb3_cfg`. This change prevents accidental or malicious
   code execution when loading configuration files, improving overall security and reliability.
 
 
--- a/source/isaaclab_rl/docs/CHANGELOG.rst
+++ b/source/isaaclab_rl/docs/CHANGELOG.rst
@@ -7,7 +7,9 @@ Changelog
 Fixed
 ^^^^^
 
-* Disallowing string value written in sb3_ppo_cfg.yaml gets eval() in :meth:`~isaaclab_rl.sb3.process_sb3_cfg`.
+* Disallowed string values in ``sb3_ppo_cfg.yaml`` from being passed to ``eval()`` in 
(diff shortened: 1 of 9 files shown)


source/isaaclab_rl/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+0.1.9 (2025-08-07)
+~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Disallowing string value written in sb3_ppo_cfg.yaml gets eval() in :meth:`~isaaclab_rl.sb3.process_sb3_cfg`.
```
Mayankm96: NIT: Would be nice to stick the format of writing the changelog (Past tense and a minor reason to know why the change was made).

source/isaaclab_rl/isaaclab_rl/sb3.py
```diff
@@ -81,7 +82,7 @@ def update_dict(hyperparams: dict[str, Any]) -> dict[str, Any]:
         return hyperparams
 
     # parse agent configuration and convert to classes
-    return update_dict(cfg)
+    return update_dict(cfg, 0)
```
Mayankm96: NIT: Normally it is nicer to write the arg name if it isn't very clear what the thing is. For example, here `0` means what? This just improves the code readability.

```suggestion
    return update_dict(cfg, depth=0)
```

source/isaaclab_tasks/isaaclab_tasks/manager_based/classic/humanoid/agents/sb3_ppo_cfg.yaml
```diff
@@ -14,9 +14,9 @@ n_epochs: 5
...
 max_grad_norm: 1.0
 vf_coef: 0.5
-policy_kwargs: "dict(
-  activation_fn=nn.ELU,
-  net_arch=[400, 200, 100],
-  optimizer_kwargs=dict(eps=1e-8),
-  ortho_init=False,
-  )"
+policy_kwargs:
+  activation_fn: 'nn.ELU'
```
Mayankm96, line 23: What happens if user does `torch.nn.ELU`? Is this something SB3 doesn't support? Why not instead check if there is "_fn" argument and parse the string for it?

ooctipus: Thanks for reviewing
It will be very nice if function input can be less strict,
as I am thinking about your suggestion, I am wondering how do you check if the argument is `_fn`? and if it is how do you verify that this function is not malicious?

I guess there has to be some presumption what has been imported and what has not, if user start to write something like `import torch; torch.nn.ELU` because they didn't know torch is imported, that feels we give away too much freedom to user, and our system can be vulnerable interpreting the all code that user imports, processes, and returns. What do you think? Another alternative is make sb3 `@configclass` typed, and use to_dict(). But I guess that will be a big breaking change.

source/isaaclab_rl/isaaclab_rl/sb3.py
ooctipus: good suggestion!

source/isaaclab_rl/docs/CHANGELOG.rst
ooctipus: Thank you for letting me know, I wasn't aware of the format! I edited my changelog will stick to it in future!

ooctipus:
After reviewing this again, 
I wasn't able to find a solution that securely parse the function,
the best I currently think we could to is getattr, since getattr doesn't allow function parsing while still enables some flexibility within the module namespace.

And depending on the usage, we could add the supported namespace, but I am very unsure of security of evaluation any string that contains unknown function.

Happy to hear other idea that will be at same time flexible and secure

kellyguo11 (approved):

Issue #1362, opened 2024-11-02
[Bug Report] Final observation in step


EladSharony:
### Describe the bug

In the [`step`](https://github.com/isaac-sim/IsaacLab/blob/f7350c72b3e7ac310eda00cbc325b8f80f31e66a/source/extensions/omni.isaac.lab/omni/isaac/lab/envs/direct_rl_env.py#L283) method, the `next_obs` is overwritten by the `reset_obs`.

Assume we have reached a terminal/truncated state, i.e., `self.reset_buf.any() == True`. Currently, these are the steps performed in `step`:

> 1. Pre-process the actions before stepping through the physics.
> 2. Apply the actions to the simulator and step through the physics in a decimated manner.
> 3. Compute the reward and done signals.
> **-> If we call `self._get_observations()` here, we will get the actual `next_obs`.**
> 4. Reset environments that have terminated or reached the maximum episode length.
> **-> When we call `self._get_observations()` here, we get the `reset_obs`.**
> 5. Apply interval events if they are enabled.
> 6. Compute observations. **Here is where `self._get_observations()` is actually called.**

Following the same logic as in [Gymnasium v0.29.0](https://github.com/Farama-Foundation/Gymnasium/blob/v0.29.0/gymnasium/vector/sync_vector_env.py), I suggest adding a `final_observation` key in the `info` dictionary:

``` python
# ... 
# -- reset envs that terminated/timed-out and log the episode information
reset_env_ids = self.reset_buf.nonzero(as_tuple=False).squeeze(-1)
if len(reset_env_ids) > 0:
    for key in self.obs_buf:
        # store the final observation for terminated/timed-out environments
        self.extras.update({"final_observation": {key: self.obs_buf[key][reset_env_ids]}})

    self._reset_idx(reset_env_ids)
    # if sensors are added to the scene, make sure we render to reflect changes in reset
    if self.sim.has_rtx_sensors() and self.cfg.rerender_on_reset:
        self.sim.render()
elif "final_observation" in self.extras:
    # remove the final observation if no envs have terminated/timed-out
    self.extras.pop("final_observation")
# ... 
```

And in the training loop, we can then handle it as follows:

``` python 
obs, infos = env.reset()

obs = torch.as_tensor(obs['policy'])
for _ in range(100):
    action = torch.as_tensor(env.action_space.sample())
    next_obs, rewards, terminations, truncations, infos = env.step(action)

    next_obs = next_obs['policy']
[...]

PR #3136, opened 2025-08-09
Updates environments to use `gymnasium.vector.VectorEnv`

### LINKED ISSUE #3095: [Proposal] All RL environment should use `gymnasium.VectorEnv` rather than `gymnasium.Env`
### Proposal

The RL environments (e.g., `DirectRLEnv`) should all inherit from `gymnasium.VectorEnv` rather than `gymnasium.Env`. 

### Motivation

The `DirectRLEnv` docstring say 

> This class implements the core functionality for reinforcement learning (RL)
> environments. It is designed to be used with any RL library. The class is designed
> to be used with vectorized environments, i.e., the environment is expected to be run
> in parallel with multiple sub-environments.
> 
> While the environment itself is implemented as a vectorized environment, we do not
> inherit from :class:`gym.vector.VectorEnv`. This is mainly because the class adds
> various methods (for wait and asynchronous updates) which are not required.
> Additionally, each RL library typically has its own definition for a vectorized
> environment. Thus, to reduce complexity, we directly use the :class:`gym.Env` over
> here and leave it up to library-defined wrappers to take care of wrapping this
> environment for their agents.

This description is true for Gym and Gymnasium < 1.0, however, in v1.0, the problematic functions described (wait and asynchronous updates) were removed from the base class. 

Updating to VectorEnv should make it clearer to users that the environments are vectorised and remove the necessity for libraries to implement their own converter functions. 
Further, users will be able to use vector wrappers ()with the environments, which isn't possible with `gymnasium.Env` implementations.

Looking at the code, I believe that minimal code changes will be necessary, as the library already requires Gymnasium v1.2 and we should be able to simplify the implementation to remove components already included in `VectorEnv` but not in `Env`

### Alternatives

Keep the current solution

### Checklist

pseudo-rnd-thoughts:
# Description

Currently, the environments use `gymnasium.Env` even though the implementations are all vectorised-based. This PR updates the implementations to use `gymnasium.vector.VectorEnv`.

Fixes https://github.com/isaac-sim/IsaacLab/issues/3095

## Type of change

## Checklist

### DIFF
--- a/docs/source/migration/migrating_from_isaacgymenvs.rst
+++ b/docs/source/migration/migrating_from_isaacgymenvs.rst
@@ -484,7 +484,7 @@ Each environment in Isaac Lab should be in its own directory following this stru
 
    gym.register(
        id="Isaac-Cartpole-Direct-v0",
-       entry_point="isaaclab_tasks.direct_workflow.cartpole:CartpoleEnv",
+       vector_entry_point="isaaclab_tasks.direct_workflow.cartpole:CartpoleEnv",
        disable_env_checker=True,
        kwargs={
            "env_cfg_entry_point": CartpoleEnvCfg,
--- a/docs/source/migration/migrating_from_omniisaacgymenvs.rst
+++ b/docs/source/migration/migrating_from_omniisaacgymenvs.rst
@@ -401,7 +401,7 @@ Each environment in Isaac Lab should be in its own directory following this stru
 
     gym.register(
         id="Isaac-Cartpole-Direct-v0",
-        entry_point="isaaclab_tasks.direct_workflow.cartpole:CartpoleEnv",
+        vector_entry_point="isaaclab_tasks.direct_workflow.cartpole:CartpoleEnv",
         disable_env_checker=True,
         kwargs={
             "env_cfg_entry_point": CartpoleEnvCfg,
--- a/docs/source/refs/snippets/tutorial_modify_direct_rl_env.py
+++ b/docs/source/refs/snippets/tutorial_modify_direct_rl_env.py
@@ -12,7 +12,7 @@ from .h1_env import H1Env, H1EnvCfg
 # [start-init-register]
 gym.register(
     id="Isaac-H1-Direct-v0",
-    entry_point="isaaclab_tasks.direct.humanoid:H1Env",
+    vector_entry_point="isaaclab_tasks.direct.humanoid:H1Env",
     disable_env_checker=True,
     kwargs={
         "env_cfg_entry_point": H1EnvCfg,
--- a/docs/source/setup/quickstart.rst
+++ b/docs/source/setup/quickstart.rst
@@ -188,7 +188,7 @@ generator, you will find at least one ``__init__.py`` file with something that l
 
     gym.register(
         id="Template-isaaclabtutorial_env-v0",
-        entry_point=f"{__name__}.isaaclabtutorial_env:IsaaclabtutorialEnv",
+        vector_entry_point=f"{__name__}.isaaclabtutorial_env:IsaaclabtutorialEnv",
         disable_env_checker=True,
         kwargs={
             "env_cfg_entry_point": f"{__name__}.isaaclabtutorial_env_cfg:IsaaclabtutorialEnvCfg",
--- a/source/isaaclab/isaaclab/envs/direct_marl_env.py
+++ b/source/isaaclab/isaaclab/envs/direct_marl_env.py
@@ -36,7 +36,7 @@ from .ui import ViewportCameraController
 from .utils.spaces import sample_space, spec_to_gym_space
 
 
-class DirectMARLEnv(gym.Env):
+class DirectMARLEnv(gym.vector.VectorEnv):
     """The superclass for the direct workflow to design multi-agent environments.
 
     This class implements the core functionality for multi-agent reinforcement learning (MARL)
@@ -45,9 +45,6 @@ class DirectMARLEnv(gym.Env):
     in parallel with multiple sub-environments.
 
     The design of this class is based on the PettingZoo Parallel API.
-    While the environment itself is implemented as a vectorized environment, we do not
-    inherit from :class:`pettingzoo.ParallelEnv` or :class:`gym.vector.VectorEnv`. This is mainly
-    because the class adds various attributes and methods that are inconsistent with them.
 
     Note:
         For vectorized environments, it is recommended to **only** call the :meth:`reset`
--- a/source/isaaclab/isaaclab/envs/direct_rl_env.py
+++ b/source/isaaclab/isaaclab/envs/direct_rl_env.py
@@ -37,21 +37,13 @@ from .ui import ViewportCameraController
 from .utils.spaces import sample_space, spec_to_gym_space
 
 
-class DirectRLEnv(gym.Env):
+class DirectRLEnv(gym.vector.VectorEnv):
     """The superclass for the direct workflow to design environments.
 
     This class implements the core functionality for reinforcement learning (RL)
     environments. It is designed to be used with any RL library. The class is designed
     to be used with vectorized environments, i.e., the environment is expected to be run
     in parallel with multiple sub-environments.
-
-    While the environment itself is implemented as a vectorized environment, we do not
-    inherit from :class:`gym.vector.VectorEnv`. This is mainly because the class adds
-    various methods (for wait and asynchronous updates) which are not required.
-    Additionally, each RL library typically has its own definition for a vectorized
-    environment. Thus, to reduce complexity, we directly use the :class:`gym.Env` over
-    here and leave it up to library-defined wrappers to take care of wrapping this
-    environment for their agents.
 
     Note:
         For vectorized environments, it is recommended to **only** call the :meth:`reset`
@@ -62,8 +54,6 @@ class DirectRLEnv(gym.Env):
 
     """
 
-    is_vector_env: ClassVar[bool] = True
-    """Whether the environment is a vectorized environment."""
     metadata: ClassVar[dict[str, Any]] = {
         "render_modes": [None, "human", "rgb_array"],
         "isaac_sim_version": get_version(),
--- a/source/isaaclab/isaaclab/envs/manager_based_rl_env.py
+++ b/source/isaaclab/isaaclab/envs/manager_based_rl_env.py
@@ -23,7 +23,7 @@ from .manager_based_env import ManagerBasedEnv
 from .manager_based_rl_env_cfg import ManagerBasedRLEnvCfg
 
 
-class ManagerBasedRLEnv(ManagerBasedEnv, gym.Env):
+class ManagerBasedRLEnv(ManagerBasedEnv, gym.vector.VectorEnv):
     """The superclass for the manager-based workflow reinforcement learning-based environments.
 
     This class inherits from :class:`ManagerBasedEnv` and implements the core functionality for
@@ -36,14 +36,6 @@ class ManagerBasedRLEnv(ManagerBasedEnv, gym.Env):
     environments. The method :meth:`step` is also expected to receive a batch of actions
     for each sub-environment.
 
-    While the environment itself is implemented as a vectorized environment, we do not
-    inherit from :class:`gym.vector.VectorEnv`. This is mainly because the class adds
-    various methods (for wait and asynchronous updates) which are not required.
-    Additionally, each RL library typically has its own definition for a vectorized
-    environment. Thus, to reduce complexity, we directly use the :class:`gym.Env` over
-    here and leave it up to library-defined wrappers to take care of wrapping this
-    environment for their agents.
-
     Note:
         For vectorized environments, it is recommended to **only** call the :meth:`reset`
         method once before the first call to :meth:`step`, i.e. after the environment is created.
@@ -53,8 +45,6 @@ class ManagerBasedRLEnv(ManagerBasedEnv, gym.Env):
 
     """
 
-    is_vector_env: ClassVar[bool] = True
-    """Whether the environment is a vectorized environment."""
     metadata: ClassVar[dict[str, Any]] = {
         "render_modes": [None, "human", "rgb_array"],
         "isaac_sim_version": get_version(),
--- a/source/isaaclab_mimic/isaaclab_mimic/envs/__init__.py
+++ b/source/isaaclab_mimic/isaaclab_mimic/envs/__init__.py
@@ -21,7 +21,7 @@ from .franka_stack_ik_rel_visuomotor_mimic_env_cfg import FrankaCubeStackIKRelVi
 
 gym.register(
     id="Isaac-Stack-Cube-Franka-IK-Rel-Mimic-v0",
-    entry_point="isaaclab_mimic.envs:FrankaCubeStackIKRelMimicEnv",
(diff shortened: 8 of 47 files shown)


kellyguo11:
Thanks for the PR. Will we require to switch to `gym.make_vec` instead of `gym.make` with this change?

pseudo-rnd-thoughts:
@kellyguo11 Yes and the `gym.register` should use `vector_entry_point` for `make_vec`

Do you know what autoreset mode IsaacLab uses? https://farama.org/Vector-Autoreset-Mode

ooctipus:
@pseudo-rnd-thoughts 
Thanks for submitting this PR,
I think IsaacLab uses same-step reset mode!

ooctipus:
@pseudo-rnd-thoughts Let me know if there is something I can help, it will be nice to update to gymnasium up-to-date standard : ))

Mayankm96:
image

We don't provide `terminal_obs` in the info dictionary. So it is "Disabled" mode.

ooctipus:
ohhh thats right, sorry for the miss, @Mayankm96 Maybe do you think it makes sense to let user choose which mode to enable in IsaacLab side? and we can default to `Disabled` mode for RL applications, I think if user can change the mode from `Disabled` to `Same-Step` or `Next-Step`, then we could rewrite Recorder Manager not inside ManagerBasedRLEnv, one reason it had to be there was because `Disabled` Mode doesn't provide terminal_obs, so the only way they can record pre-reset observation is to put in ManagerBasedRLEnv in reset logic.

pseudo-rnd-thoughts:
> We don't provide terminal_obs in the info dictionary. So it is "Disabled" mode.

@Mayankm96 Are you sure? The [manager_based_rl_env](https://github.com/isaac-sim/IsaacLab/blob/f20d74c59d3e20fc822c4e4c5bf8535a48c5aa0b/source/isaaclab/isaaclab/envs/manager_based_rl_env.py#L47) has the following note

>     Note:
>        For vectorized environments, it is recommended to **only** call the :meth:`reset`
>        method once before the first call to :meth:`step`, i.e. after the environment is created.
>        After that, the :meth:`step` function handles the reset of terminated sub-environments.
>        This is because the simulator does not support resetting individual sub-environments
>        in a vectorized environment. 

For the `disabled` mode, users must call `reset` to reset each sub-environment. 
Would you happen to know where the auto-reset code is implemented? 

@ooctipus Could you approve the github actions? 
I'm unfamiliar with the codebase so I suspect that I've missed a necessary code change and I haven't been able to get the tests to run locally.

Mayankm96:
@pseudo-rnd-thoughts : I didn't see the `reset(masks)` is a user-defined call. In that case, the closest we have to what is described is the Same-Step reset. Although we don't provide the `final_obs` as mentioned above.

pseudo-rnd-thoughts:
> I didn't see the reset(masks) is a user-defined call. In that case, the closest we have to what is described is the Same-Step reset. Although we don't provide the final_obs as mentioned above.

Yes, checking the code in `DirectRLEnv.step`, it looks like you've implemented the same-step autoreset mode as I've described it in that article; however, you don't return the `final_obs`, as you mentioned. 

Is there a reason the implementation doesn't include the episode's final observation, as currently, you get the next episode's first observation?
To my understanding, it's necessary to correctly implement most RL algorithms as in the case of truncation (i.e., time-outs), you shouldn't have an episode's final observation. https://farama.org/Gymnasium-Terminated-Truncated-Step-API

ooctipus:
| /workspace/isaaclab/source/isaaclab_rl/test/test_rl_games_wrapper.py                   |  FAILED |     0.08 |   0/1   |
| /workspace/isaaclab/source/isaaclab_rl/test/test_skrl_wrapper.py                       |  FAILED |     0.09 |   0/1   |
| /workspace/isaaclab/source/isaaclab_rl/test/test_sb3_wrapper.py                        |  FAILED |     0.08 |   0/1   |
| /workspace/isaaclab/source/isaaclab_rl/test/test_rsl_rl_wrapper.py                     |  FAILED |     0.16 |   0/2   |

seems like few tests are failing, I am letting it run again, if fails again I ll take a look at it

Mayankm96:
> Is there a reason the implementation doesn't include the episode's final observation, as currently, you get the next episode's first observation?

I checked the code for RSL-RL (RL-Games and skrl also do something similar): It seems they use the state before the termination ($s_{T-1}$ instead of $s_T$) for computing the bootstrapped values at truncations.

* SKRL: https://github.com/Toni-SM/skrl/blob/main/skrl/agents/torch/ppo/ppo.py#L297-L299
* RSL-RL: https://github.com/leggedrobotics/rsl_rl/blob/main/rsl_rl/algorithms/ppo.py#L161-L164

While I agree, the `final_obs` is technically the right thing to do. Empirically, it does not make a difference. Using the values for states (computed in the previous algorithm step) saves an additional forward call of the critic network and provides some performance gain when handling a large batch of environments.

Tagging @Toni-SM since he has a better idea for algorithms other than PPO.

Mayankm96:
Do we need to switch from `gym.make` to `gym.make_vec` calls in the scripts? @ooctipus

kellyguo11:
> Do we need to switch from `gym.make` to `gym.make_vec` calls in the scripts? @ooctipus

I think this could become a breaking change for user environments, not sure if we get much benefit in return?

ooctipus:
the largest benefit will probably be for implementing a new rl-lib wrapper, the process could be easier by viewing ManagerBasedEnv or DirectEnv as VectorEnv. 
```
Updating to VectorEnv should make it clearer to users that the environments are vectorised and remove the necessity for libraries to implement their own converter functions.
Further, users will be able to use vector wrappers ()with the environments, which isn't possible with gymnasium.Env implementations.
```
right now we get around it by wraping IsaacLabEnvs(not gym env) with RLLib-Wrapper , where the rl-libs can directly assume it is interacting with ManagerBasedEnv, or DirectEnv. We basically wrote `own converter functions`, but even switch to VectorEnv, customized converter functions may still needed because each rllib also have some special requirements that a generic VectorEnv is likely not enough to satisfy.

I guess maybe the largest breaking change is because `entry_point` -> `vector_entry_point`?  If we applied this change, a user implemented `entry_point` for their environment will be potentially not get registered properly until they also change it to vector_entry_point. 
I am also weighing if change `entry_point` -> `vector_entry_point` is needed at all. Since in current IsaacLab implementation, even if num_envs == 1, it is still vectorized as (n, observation_shape), (n, action_shape), n = 1

My general feeling is that if we introduce very minimum breaking change, I think there is still value in adapting to gym's best practice for long term.

noahfarr:
Do I understand correctly that currently there is no way to access the true final obs of an episode? Why not just add it to the info dict as you compute it anyway in step?

Mayankm96:
@noahfarr We don't compute it in the step. It's only computed after the reset is performed for all the environments together. We don't support computing the observations for only a sub-set of environment indices in the vectorized environment.

From what batched RL libraries do: the agent can use the last observations of the previous step as final obs instead of having the environment return this twice. At least for continuous control problems, I think the two obs can be considered fairly close to use an approximation, and I would rather avoid calling compute obs twice for performance reasons.

We have two ways of designing environments. Users who design their environments can compute `final_obs` in the step function if they need it. But the examples we provide don't do this right now. Very likely, we won't due to reasons above.

mschuck-rai:
A big +1 for moving to gymnasium's vector env and providing a clear `AutoresetMode`. Some frameworks work around this issue (RSL_RL for sure, SKRL at least for PPO), but others (e.g. torchRL) do not. It would be great if IsaacLab had less footguns when switching learning frameworks.

george-nehma:
@Mayankm96 how would you propose we handle this when we have environments in which the terminal state of the environment gives a significant contribution to the reward function? For example in my work, when a lander lands it receives a large positive or negative reward for landing or crashing, but this is only conditioned on the terminal observation so would this not affect the learning algorithm?

tobiabir:
> @pseudo-rnd-thoughts : I didn't see the `reset(masks)` is a user-defined call. In that case, the closest we have to what is described is the Same-Step reset. Although we don't provide the `final_obs` as mentioned above.

There is two open issues for this bug of the missing final observation (#1362, #4369).

> A big +1 for moving to gymnasium's vector env and providing a clear `AutoresetMode`. Some frameworks work around this issue (RSL_RL for sure, SKRL at least for PPO), but others (e.g. torchRL) do not. It would be great if IsaacLab had less footguns when switching learning frameworks.

Another big +1 for providing a clear `AutoresetMode`. Gymnasium has those since version v1.1 (https://github.com/Farama-Foundation/Gymnasium/pull/1227, https://github.com/Farama-Foundation/Gymnasium/releases/tag/v1.1.0).

> @noahfarr We don't compute it in the step. It's only computed after the reset is performed for all the environments together. We don't support computing the observations for only a sub-set of environment indices in the vectorized environment.
> 
> From what batched RL libraries do: the agent can use the last observations of the previous step as final obs instead of having the environment return this twice. At least for continuous control problems, I think the two obs can be considered fairly close to use an approximation, and I would rather avoid calling compute obs twice for performance reasons.
> 
> We have two ways of designing environments. Users who design their environments can compute `final_obs` in the step function if they need it. But the examples we provide don't do this right now. Very likely, we won't due to reasons above.

@Mayankm96 If the `recorder_manager` has active terms the observations are already computed before and after the resets.
I strongly dislike the idea of intentionally using the wrong observation.
- The second to last observation might not be close at all.
- Even if close this will introduce noise with unknown effects on the performance.
- Major frameworks (like Stable-Baselines3 and TorchRL) will rely on the gymnasium interface (with AutoresetMode) and not support this work around.

Also what is the purpose of an example implementation that is demonstrating wrong behavior?

@vmoens fyi, could be interesting for TorchRL

vmoens:
Thanks for the ping! Big +1 for moving to gymnasium.vector.VectorEnv with explicit AutoresetMode support.
From TorchRL's perspective:

- We need the final observation of each trajectory (N+1 obs, N actions). This is not optional -- it's needed for value bootstrapping / GAE, and using the second-to-last observation as an approximation introduces bias with unknown effects on training. I strongly agree with @noahfarr on this point.

- Our preferred mode is `AutoresetMode.DISABLED`. With this mode, the step just returns the actual final observation (no extra compute), and we call `reset(options={"reset_mask": ...})` ourselves on the envs that are done. This completely avoids the "double observation compute" concern -- the env only computes observations once per step, and the reset computes new observations only for the terminated subset.

- We also support `AutoresetMode.SAME_STEP` (with `final_obs` in info), but `DISABLED` is preferred as it's cleaner and doesn't require any extra observation computation.

- We explicitly reject `AutoresetMode.NEXT_STEP` -- it loses the final observation entirely.

So if IsaacLab supports `DISABLED` mode (which just means: don't auto-reset, let the caller do it with a mask), TorchRL will work out of the box with zero overhead and no double observation computation. We can accommodate other reset modes as long as the last obs is available (eg, we did that for envpool, it's not beautiful but it works!)

pseudo-rnd-thoughts:
Returning to this PR, I think there are two different requests at play

1. Update IsaacLab's environments to use `gymnasium.vector.VectorEnv` rather than `gymnasium.Env` for the simple reason that the environments are vector environments. This should require only minor updates to the environments themselves, I believe, to match Gymnasium API expectations. The largest change is that environment should be created using `gymnasium.make_vec` rather than `gymnasium.make`, and would require users to update downstream. 
2. Add support for Gymnasium Autoreset API (see this blog post I made on the particular details https://farama.org/Vector-Autoreset-Mode). This is more nuanced, particularly on end-users. IsaacLab implements a variant of the SameStep mode, however, it does not compute the final observation. I suspect it should be possible to implement the "correct" version of the SameStep mode along with NextStep and Disabled mode. 

If the IssacLab maintainers are happy with these changes, I would suggest that these requests be completed in different PRs. The first in this PR and the second in a different PR

@vmoens you note 
> We explicitly reject AutoresetMode.NEXT_STEP -- it loses the final observation entirely.

Could you make an issue on Gymnasium, as checking the code, the current implementation does return the final observation, so I don't understand what you mean

StoneT2000:
> Thanks for the ping! Big +1 for moving to gymnasium.vector.VectorEnv with explicit AutoresetMode support. From TorchRL's perspective:
> 
> * We need the final observation of each trajectory (N+1 obs, N actions). This is not optional -- it's needed for value bootstrapping / GAE, and using the second-to-last observation as an approximation introduces bias with unknown effects on training. I strongly agree with @noahfarr on this point.
> * Our preferred mode is `AutoresetMode.DISABLED`. With this mode, the step just returns the actual final observation (no extra compute), and we call `reset(options={"reset_mask": ...})` ourselves on the envs that are done. This completely avoids the "double observation compute" concern -- the env only computes observations once per step, and the reset computes new observations only for the terminated subset.
> * We also support `AutoresetMode.SAME_STEP` (with `final_obs` in info), but `DISABLED` is preferred as it's cleaner and doesn't require any extra observation computation.
> * We explicitly reject `AutoresetMode.NEXT_STEP` -- it loses the final observation entirely.
> 
> So if IsaacLab supports `DISABLED` mode (which just means: don't auto-reset, let the caller do it with a mask), TorchRL will work out of the box with zero overhead and no double observation computation. We can accommodate other reset modes as long as the last obs is available (eg, we did that for envpool, it's not beautiful but it works!)

Just adding my support for almost every point here. I personally do not like the disabled autoreset mode but it gives individual RL libraries more agency to handle it however they wish. SAME_STEP mode is from what I observe the current most used mode in most RL (research) codebases.

Final observation is also an absolute must, especially for manipulation problems (or problems of that nature in terms of how resets happen). There are a lot of parts of isaac lab and it's various supported RL libraries that are optimized just for locomotion problems. Not handling final observations is likely one of them (implicitly).

In my own codebase I end up modifying `DirectRLEnv` and replacing the whole `step` and `reset` functions to handle various aspects of RL workflows to make RL faster and easier to debug.

And finally, I strongly agree with moving to a true `VectorEnv` interface. The Maniskill project does this too but via a wrapper approach around the base environment (https://github.com/haosulab/ManiSkill/blob/main/mani_skill/vector/wrappers/gymnasium.py). Strongly recommend, the feedback I've gotten on this has been very positive given a lot of researchers (in manipulation / other areas of robotics) tend to use heavily modified RL code / their own libraries, most of which assume `VectorEnv` APIs.

fl1ps1de5:
@StoneT2000  Are you able to provide an example `DirectRLEnv` with modified `step` and `reset` functions.

wzm2256:
Any update on this?

ooctipus:
https://github.com/isaac-sim/IsaacLab/pull/6219 we addressed it here, IsaacLab will now use SAME-STEP mechanics (default has no final_obs, but has a optional flag to turn on returning final obs)

AntoineRichard:
Hi @pseudo-rnd-thoughts — thank you so much for opening this PR, and for taking the time to contribute to Isaac Lab! 🙏

| | |
|---|---|
| **Opened** | 2025-08-09 (about a year ago) |
| **Last commit on the branch** | 2025-08-19 |
| **Last activity from the author** | about a year ago |
| **Review status** | Never reviewed by a maintainer — nobody on the team got to it. Sorry about that. |
| **Size** | 3 commit(s), 47 file(s) changed, +120 / -142 |

It was picked up by the sweep because it targets `main` instead of `develop`, and it has been open for about a year. It was then put in the "close" bucket because the author has been silent for about a year — which is the signal we used to tell apart pull requests that are still being worked on from ones that have genuinely been set aside.

---

PR #3139, opened 2025-08-10
Fixs isaaclab.sh to detect isaacsim_version accurately 4.5 or >= 5.0

ooctipus:
# Description

Current is_isaacsim_version_4_5() sometimes mis-detects 5.0 when running 4.5 (e.g., Docker/symlinked installs), causing the conda bootstrap to pick Python 3.11 instead of 3.10.

This pr ensures determining the isaacsim version following below steps

1. Read ${ISAACLAB_PATH}/_isaac_sim/VERSION (symlink/binary install).

2. If unknown, import isaacsim and read ../../VERSION next to the package. (pip installation)

3. If still unknown, fall back to importlib.metadata.version("isaacsim").

I tested locally to work with 4.5 pip or binary install.

## Screenshots

## Checklist

### DIFF
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -43,33 +43,48 @@ install_system_deps() {
     fi
 }
 
+# Returns success (exit code 0 / "true") if the detected Isaac Sim version starts with 4.5,
+# otherwise returns non-zero ("false"). Works with both symlinked binary installs and pip installs.
 is_isaacsim_version_4_5() {
+    local version=""
     local python_exe
     python_exe=$(extract_python_exe)
 
-    # 1) Try the VERSION file
-    local sim_file
-    sim_file=$("${python_exe}" -c "import isaacsim; print(isaacsim.__file__)" 2>/dev/null) || return 1
-    local version_path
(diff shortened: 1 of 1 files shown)


isaaclab.sh
```diff
@@ -43,33 +43,48 @@ install_system_deps() {
     fi
 }
 
+# Returns success (exit code 0 / "true") if the detected Isaac Sim version starts with 4.5,
```
Mayankm96, line 46: Thanks for the fix. Does the isaaclab.bat file also need to be adapted?

Mayankm96: Maybe we should convert these shell/bat scripts to a python file to simplify the maintainance work between different OS platforms.

ooctipus: Yeah, I think people generally also understand python more than bash.
Though I wonder if using python, could it be a lot more verbose than writing bash.

ooctipus: I checked the isaaclab.bat, its checking workflow is good, and also worked with windows, so I think no need for now

Issue #547, opened 2024-06-21
[Bug Report] [Docs] Wrong RL Framework suggestion in Docs


EricVoll:
### Describe the bug

The [setup instructions](https://github.com/isaac-sim/IsaacLab/blob/main/docs/source/setup/installation/binaries_installation.rst) list the possible rl frameworks (e.g., rsl_rl) but the script takes `rsl-rl` as an input. 

### Steps to reproduce

Call `isaaclab.sh -i rsl_rl` as explained in the docs.
The script won't be able to find rsl_rl, since it only knows rsl-rl

### Additional context

This can either be a bug in the docs, or in the script.

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)
- [x] I have checked that the issue is not in running Isaac Sim itself and is related to the repo

PR #3149, opened 2025-08-11
Enhance Pink IK controller with null-space posture control and improv…

michaellin6:
# Description

Enhance Pink IK Controller with Null Space Posture Control

This PR improves the Pink IK controller integration for better humanoid robot control and more natural postures.

**Note**: Original this PR was staged in the internal repo (#547). It has been moved here due to new Github workflow.

### New Null Space Posture Task
- Added NullSpacePostureTask to enforce postural constraints on shoulder/waist joints while prioritizing end-effector tasks
- Maintains natural robot poses during manipulation
### Controller Improvements
- Tuned low level PD controller gains
- Support mixed task types (FrameTask + NullSpacePostureTask)
### Testing & Environment Updates
- Redesigned pink controller test script to use JSON-based configurations to program test motions.
- Updated all environments (PickPlace, NutPour, ExhaustPipe) with null space control, damping tasks, and improved tracking
- Added `Isaac-PickPlace-GR1T2-WaistEnabled-Abs-v0` env that is identical to `Isaac-PickPlace-GR1T2-Abs-v0` but enables the Waist DOFs.
- Added target_eef_link_names mapping for clearer link specification

Fixes # (issue)
These changes help fix the following problems from [VDR feedback](https://docs.google.com/document/d/1saB1QA5r_WlD1l17q7C04WWNltnBW-K0ydI2UB8jxAs/edit?tab=t.0)
- [Enable Waist DOF](https://nvbugspro.nvidia.com/bug/5235527)
- Discourage elbow flare
- Make controller low-latency and low-jerk. **We improved the unit test for the pink controller and reduced our position and rotation accuracy tolerance from 30 mm, 10 degrees to 1 mm, 1 degree.**
- Develop metric for controller performance
- Added a flag to disable failure due to joint limits. Previously, any commanded pose that ended in joint limit violation would result in no solution and the controlled robot freezing in place. This change gets the solver to still provide a solution and instead issue a warning for joint limit violations.

## Screenshots

These controller changes have been tested through the Mimic pipeline (teleop_se3_agent.py, record/replay_demos.py). Here are videos showing teleoperation of all three environments working.

### PickPlace-GR1T2-Abs
image

### NutPour-GR1T2
image

### ExhaustPipe-GR1T2
image

### Successfully Trained Robomimic Model Rollout on PickPlace task
image

## Checklist


yanchangNvidia:
Nice work. Overall it looks good to me.

peterd-NV:
Do we have numbers to validate Mimic datagen/Robomimic policy success rate on the existing teleoperation/imitation learning workflows? [Demo 1](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-1-data-generation-and-policy-training-for-a-humanoid-robot) and [Demo 2](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-2-visuomotor-policy-for-a-humanoid-robot) here. @michaellin6

michaellin6:
> Do we have numbers to validate Mimic datagen/Robomimic policy success rate on the existing teleoperation/imitation learning workflows? [Demo 1](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-1-data-generation-and-policy-training-for-a-humanoid-robot) and [Demo 2](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-2-visuomotor-policy-for-a-humanoid-robot) here. @michaellin6

I am working on getting these metrics

source/isaaclab/isaaclab/utils/string.py
```diff
@@ -272,7 +272,10 @@ def resolve_matching_names(
 
 
 def resolve_matching_names_values(
-    data: dict[str, Any], list_of_strings: Sequence[str], preserve_order: bool = False
+    data: dict[str, Any],
+    list_of_strings: Sequence[str],
+    preserve_order: bool = False,
+    require_all_keys_matched: bool = True,
```
Mayankm96:
```suggestion
    strict: bool = True,
```

Would rename this to be `strict` to prevent long arg names.

source/isaaclab/isaaclab/utils/string.py
```diff
@@ -293,14 +296,15 @@ def resolve_matching_names_values(
         data: A dictionary of regular expressions and values to match the strings in the list.
         list_of_strings: A list of strings to match.
         preserve_order: Whether to preserve the order of the query keys in the returned values. Defaults to False.
+        require_all_keys_matched: Whether to require that all keys in the dictionary are matched. Defaults to True.
```
Mayankm96: The wording here seems off. Is it that all variables in `list_of_strings` have a match, or all keys in the dictionary have a string to which they got matched to?

source/isaaclab/test/controllers/test_pink_ik.py
```diff
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 """Launch Isaac Sim Simulator first."""
+import json
```
Mayankm96: Would move this down after simulation app was made. The top import are only for launching simulation app related things.

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -10,6 +10,8 @@
 from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
+from pink.tasks import FrameTask
```
Mayankm96, line 13: Does that mean pinocchio and pink are now mandatory dependencies? I thought there were optional?

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -56,7 +58,11 @@ def __init__(self, cfg: pink_actions_cfg.PinkInverseKinematicsActionCfg, env: Ma
         assert env.num_envs > 0, "Number of environments specified are less than 1."
         self._ik_controllers = []
         for _ in range(env.num_envs):
-            self._ik_controllers.append(PinkIKController(cfg=copy.deepcopy(self.cfg.controller), device=self.device))
+            self._ik_controllers.append(
+                PinkIKController(
+                    cfg=copy.deepcopy(self.cfg.controller), robot_cfg=env.scene.cfg.robot, device=self.device
```
Mayankm96: Deepcopy doesn't work all the time with nested configclass instances. Using `.copy()` should be safer.

```suggestion
                    cfg=self.cfg.controller.copy(), robot_cfg=env.scene.cfg.robot, device=self.device
```

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
```diff
@@ -0,0 +1,133 @@
...
+        anchor_rot=(1.0, 0.0, 0.0, 0.0),
+    )
+
+    # Temporary directory for URDF files
+    temp_urdf_dir = tempfile.gettempdir()
+
+    # Idle action to hold robot in default pose
+    # Action format: [left arm pos (3), left arm quat (4), right arm pos (3), right arm quat (4),
+    #                 left hand joint pos (11), right hand joint pos (11)]
+    idle_action = torch.tensor([
```
Mayankm96: Maybe you can rearrange the array to make it easier to look at and disable the linter for that part. Right now this list looks long and hard to figure out :)

Mayankm96 (changes requested):
I did a high-level pass of the code. Most of it seems fine. Since I'm not familiar with PINK and mimic related tasks, I don't have any comments on the implementation side. It would be great if someone from the Mimic team also goes over it.

From high-level, it seems there are some mimic tasks in `isaaclab_mimic` and some in `isaaclab_tasks`. It would be great to keep them all together as much as possible. We can discuss this offline as it's more about restructuring the code to make it easier to navigate through it.

source/isaaclab/test/utils/test_string.py
```diff
@@ -162,6 +162,14 @@ def test_resolve_matching_names_values_with_basic_strings():
     query_names = {"a|c": 1, "b": 0, "f": 2}
     with pytest.raises(ValueError):
         _ = string_utils.resolve_matching_names_values(query_names, target_names)
+    # test require_all_keys_matched=False
```
Mayankm96: Could this be made its own test fixture?

source/isaaclab/isaaclab/controllers/null_space_posture_task.py
```diff
@@ -0,0 +1,217 @@
...
+
+import numpy as np
+
+import pinocchio as pin
+from pink.configuration import Configuration
+from pink.tasks import Task
+from pink.utils import get_root_joint_dim
+
+
+class NullSpacePostureTask(Task):
```
Mayankm96: Do you plan to add more of these tasks in the future? If so, I suggest making a separate folder called "pink_tasks" and have it in there. Then later on, other tasks can be added easily.

Mayankm96: Also please update the API docs to include these in the doc generation: https://github.com/michaellin6/IsaacLab/blob/1bd63bffc5f54bfadb1c927f25bb4eaa760d71bf/docs/source/api/lab/isaaclab.controllers.rst

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
huihuaNvidia2023: @Mayankm96 can you please elaborate more about why `.copy()` is safer in this case?

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,28 @@
...
+Changed
+^^^^^
+
+* Updated pink inverse kinematics controller configuration for the following tasks (Isaac-PickPlace-GR1T2, Isaac-NutPour-GR1T2, Isaac-ExhaustPipe-GR1T2)
+  to increase end-effector tracking accuracy and speed. Also added a null-space regularizer that enables turning on of waist degrees-of-freedom.
+* Improved the test_pink_ik script to more comprehensive test on controller accuracy. Also, migrated to use pytest. With the current IK controller
+  improvements, our unit tests pass position and orientation accuracy test within **(1 mm, 1 degree)**. Previously, the position accuracy tolerances
+  were set to **(30 mm, 10 degrees)**.
+* Including a new config parameter :attr:`fail_on_ik_error` to :class:`~isaaclab.controllers.pink_ik.PinkIKControllerCfg`
+  to control whether the IK controller should fail if robot joint limits are exceeded. This adds to stability of the controller and avoids
```
huihuaNvidia2023: Can you elaborate more about what does `fail` mean? Will the app crash?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
```diff
@@ -0,0 +1,133 @@
...
+        """Post initialization."""
+        # general settings
+        self.decimation = 6
+        self.episode_length_s = 20.0
+        # simulation settings
+        self.sim.dt = 1 / 120  # 120Hz
+        self.sim.render_interval = 2
+
+        # Add waist joint to pink_ik_cfg
+        self.actions.pink_ik_cfg.pink_controlled_joint_names.append("waist_yaw_joint")
```
huihuaNvidia2023: nit: this is nicer and less prone to error.
```
waist_joint_names = ["waist_yaw_joint", "waist_pitch_joint", "waist_roll_joint"]
for joint_name in waist_joint_names
     self.actions.pink_ik_cfg.pink_controlled_joint_names.append(joint_name)
     self.actions.pink_ik_cfg.ik_urdf_fixed_joint_names.remove(joint_name)
```

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
```diff
@@ -0,0 +1,133 @@
...
+        self.actions.pink_ik_cfg.controller.mesh_path = temp_urdf_meshes_output_path
+
+        self.teleop_devices = DevicesCfg(
+            devices={
+                "handtracking": OpenXRDeviceCfg(
+                    retargeters=[
+                        GR1T2RetargeterCfg(
+                            enable_visualization=True,
+                            # OpenXR hand tracking has 26 joints per hand
+                            num_open_xr_hand_joints=2 * 26,
```
huihuaNvidia2023: nit: can we define a constant for this number?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_env_cfg.py
```diff
@@ -209,20 +261,52 @@ class ActionsCfg:
...
+                            "waist_roll_joint",
+                        ],
+                        "GR1T2_fourier_hand_6dof_right_hand_pitch_link": [
+                            "right_shoulder_pitch_joint",
+                            "right_shoulder_roll_joint",
+                            "right_shoulder_yaw_joint",
+                            "right_wrist_yaw_joint",
+                            "right_wrist_roll_joint",
+                            "right_wrist_pitch_joint",
+                            "waist_yaw_joint",
```
huihuaNvidia2023: add a placeholder here. If we define joint groups this way, it will mean we have to compute the Jacobian for the waist joints twice.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/exhaustpipe_gr1t2_base_env_cfg.py
```diff
@@ -134,6 +135,52 @@ class ObjectTableSceneCfg(InteractiveSceneCfg):
             },
             joint_vel={".*": 0.0},
         ),
+        actuators={
```
huihuaNvidia2023: maybe this is also a question for @Mayankm96 and @kellyguo11, I see a lot of code duplication when defining the robot. Could we have a separate file named `gr1t2_robot_cfg.py` which contains all the variations of the same robot?

source/isaaclab_tasks/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,24 @@
...
+
+
+Changed
+^^^^^
+
+* Updated pink inverse kinematics controller configuration for the following tasks (``Isaac-PickPlace-GR1T2``, ``Isaac-NutPour-GR1T2``, ``Isaac-ExhaustPipe-GR1T2``)
+  to increase end-effector tracking accuracy and speed. Also added a null-space regularizer that enables turning on of waist degrees-of-freedom without
+  the robot control drifting to a bending posture.
+* Tuned the pink inverse kinematics controller and joint PD controllers for the following tasks (``Isaac-PickPlace-GR1T2``, ``Isaac-NutPour-GR1T2``, ``Isaac-ExhaustPipe-GR1T2``)
+  to improve the end-effector tracking accuracy and speed. Achieving position and orientation accuracy test within **(2 mm, 1 degree)**.
```
huihuaNvidia2023, line 20: should we be consistent here? It was mentioned that the accuracy was improved to (1mm, 1 degree), but here states (2 mm, 1 degree).

source/isaaclab_tasks/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,24 @@
 Changelog
 ---------
 
+0.10.46 (2025-07-25)
+~~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+* New ``Isaac-PickPlace-GR1T2-WaistEnabled-Abs-v0`` environment that enables the waist degrees-of-freedom for the GR1T2 robot.
```
huihuaNvidia2023, line 10: what does `Abs` mean here? The env cfg is defined as `PickPlaceGR1T2WaistEnabledEnvCfg` which does not have `Abs` in it?

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env_cfg.py
```diff
@@ -0,0 +1,111 @@
...
+
+@configclass
+class PickPlaceGR1T2WaistEnabledMimicEnvCfg(PickPlaceGR1T2WaistEnabledEnvCfg, MimicEnvCfg):
+
+    def __post_init__(self):
+        # Calling post init of parents
+        super().__post_init__()
+
+        # Override the existing values
+        self.datagen_config.name = "demo_src_gr1t2_demo_task_D0"
```
huihuaNvidia2023: Curious about the datagen_config name convention here? Do you need to point out the difference between with or without waist?  Also FYI @peterd-NV

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Takes a target pose and gripper action for the end effector controller and returns an action
+        (usually a normalized delta pose action) to try and achieve that target pose.
+        Noise is added to the target pose action if specified.
+
+        Args:
+            target_eef_pose_dict: Dictionary of 4x4 target eef pose for each end-effector.
+            gripper_action_dict: Dictionary of gripper actions for each end-effector.
+            noise: Noise to add to the action. If None, no noise is added.
+            env_id: Environment index to get the action for.
```
huihuaNvidia2023: not used

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env.py
```diff
@@ -0,0 +1,129 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from collections.abc import Sequence
+
+import isaaclab.utils.math as PoseUtils
+from isaaclab.envs import ManagerBasedRLMimicEnv
+
+
+class PickPlaceGR1T2WaistEnabledMimicEnv(ManagerBasedRLMimicEnv):
```
huihuaNvidia2023: Can these four functions be a common utility for the existing MimicEnv?

huihuaNvidia2023: Also, adding a docstring for the class would be helpful for the user

source/isaaclab/isaaclab/controllers/pink_ik_cfg.py
```diff
@@ -57,3 +57,6 @@ class PinkIKControllerCfg:
 
     show_ik_warnings: bool = True
     """Show warning if IK solver fails to find a solution."""
+
+    fail_on_joint_limit_violation: bool = True
+    """Fail the IK solver if a joint limit is violated."""
```
huihuaNvidia2023: nit: maybe add more details what is expected with IK solver failed.

source/isaaclab/isaaclab/utils/string.py
michaellin6: It is to require that all keys in the dictionary get a match. I updated the working slightly to hopefully make that more clear

source/isaaclab/test/controllers/test_pink_ik.py
michaellin6: done

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: If you are using any pink components for inverse kinematics then pink and pinocchio are required. I don't think the changes in this PR change the requirements.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
michaellin6: I'm removing this tensor as it is not needed

source/isaaclab/test/utils/test_string.py
michaellin6: moved the test case to its own test function

source/isaaclab/isaaclab/controllers/null_space_posture_task.py
michaellin6: done

source/isaaclab/docs/CHANGELOG.rst
michaellin6: clarified as "Including a new config parameter :attr:`fail_on_ik_error` to :class:`~isaaclab.controllers.pink_ik.PinkIKControllerCfg` to control whether the IK controller raise an exception if robot joint limits are exceeded. In the case of an exception, the controller will hold the last joint position."

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_waist_enabled_env_cfg.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_env_cfg.py
michaellin6: I think that is normal. On both tasks there are shared joints. As long as we cast through the null-space of each task, there shouldn't be a problem with having the same joints on multiple tasks.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/exhaustpipe_gr1t2_base_env_cfg.py
michaellin6: There is [fourier.py](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_assets/isaaclab_assets/robots/fourier.py) I will move the actuator definitions to a common class to avoid repeating in each env.

source/isaaclab_tasks/docs/CHANGELOG.rst
michaellin6: I think this is a naming convention that seems to have started from here https://isaac-sim.github.io/IsaacLab/main/source/overview/environments.html#manipulation where Abs and Relative are how the inputs commands are specified.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env_cfg.py
michaellin6: naming updated

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env.py
michaellin6: It is needed to conform to the mimic interface. I will add a comment to clarify why its not used.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env.py
michaellin6: I realized that this file was not needed. All four environments already share the common class [PickPlaceGR1T2MimicEnv](https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_mimic_env.py). They are defined here https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/__init__.py

source/isaaclab/isaaclab/controllers/pink_ik_cfg.py
michaellin6: done

source/isaaclab/isaaclab/utils/string.py
michaellin6: done

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: done

source/isaaclab_tasks/docs/CHANGELOG.rst
michaellin6: Actually, there are some tests that were added (forward waist bending) that are getting 1.5 mm error when reaching forward, so I've had to increase the test threshold. I think is safer to say that we can achieve 2 mm error for the tests that we have.

michaellin6:
> > Do we have numbers to validate Mimic datagen/Robomimic policy success rate on the existing teleoperation/imitation learning workflows? [Demo 1](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-1-data-generation-and-policy-training-for-a-humanoid-robot) and [Demo 2](https://isaac-sim.github.io/IsaacLab/main/source/overview/teleop_imitation.html#demo-2-visuomotor-policy-for-a-humanoid-robot) here. @michaellin6
> 
> I am working on getting these metrics

@peterd-NV  Values are updated in the PR description. We will need to update the datasets provided.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -0,0 +1,206 @@
...
+    Multiple tasks are resolved through weighted optimization, formulating a quadratic program that minimizes
+    weighted task errors while respecting joint velocity limits.
+
+    It supports user defined tasks, and we have provided a NullSpacePostureTask for maintaining desired joint configurations.
+
+    Reference:
+        Pink IK Solver: https://github.com/stephane-caron/pink
+    """
+
+    def __init__(self, cfg: PinkIKControllerCfg, robot_cfg: ArticulationCfg, device: str):
```
ooctipus, line 50: If you are sure that dictionary from expression to default joint position is only thing needed from ArticulationCfg, maybe more clear to just ask for that rather than entire ArticulationCfg. Some ArticulationCfg doesn't provide default and will instead get the default from usd, and some articulation over specify default, which you in case with strict = false, you may sometime omit some joint? I guess those case will break the code because the init_joint_pos coming from robot_cfg is incomplete. What do you think? Feel free to disagree : ))

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -0,0 +1,206 @@
...
+        Args:
+            curr_joint_pos: The current joint positions of shape (num_joints,).
+            dt: The time step for computing joint position changes in seconds.
+
+        Returns:
+            The target joint positions as a tensor of shape (num_joints,) on the specified device.
+            If the IK solver fails, returns the current joint positions unchanged to maintain stability.
+        """
+        # Initialize joint positions for Pink, including the root and universal joints
+        joint_positions_pink = np.array(self._reorder_array(curr_joint_pos, self.isaac_lab_to_pink_ordering))
```
ooctipus: Using list comprehension to reorder array that gets used computed doesn't seems the most optimum way to reorder, doesn't curr_joint_pos and self.isaac_lab_to_pink_ordering work with np indices mechanism?

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -0,0 +1,206 @@
...
+                    f" positions.\nError: {e}"
+                )
+            return torch.tensor(curr_joint_pos, device=self.device, dtype=torch.float32)
+
+        # Discard the first 6 values (for root and universal joints)
+        pink_joint_angle_changes = Delta_q
+
+        # Reorder the joint angle changes back to Isaac Lab conventions
+        joint_vel_isaac_lab = torch.tensor(
+            self._reorder_array(pink_joint_angle_changes, self.pink_to_isaac_lab_ordering),
```
ooctipus: similar indices suggestion as above

source/isaaclab/isaaclab/envs/mdp/actions/pink_actions_cfg.py
```diff
@@ -5,7 +5,7 @@
 
 from dataclasses import MISSING
 
-from isaaclab.controllers.pink_ik_cfg import PinkIKControllerCfg
+from isaaclab.controllers.pink_ik import PinkIKControllerCfg
```
ooctipus, line 8: maybe you can just import from `isaaclab.controllers` ?

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/__init__.py
```diff
@@ -11,6 +11,7 @@
 from .nutpour_gr1t2_mimic_env_cfg import NutPourGR1T2MimicEnvCfg
 from .pickplace_gr1t2_mimic_env import PickPlaceGR1T2MimicEnv
 from .pickplace_gr1t2_mimic_env_cfg import PickPlaceGR1T2MimicEnvCfg
+from .pickplace_gr1t2_waist_enabled_mimic_env_cfg import PickPlaceGR1T2WaistEnabledMimicEnvCfg
```
ooctipus, line 14: by the way for registration, if we explicitly write out import, these module gets imported even if they are not the environment users are going to run.
like other il envs registration you can do something like below to minimize this overhead:

```
"env_cfg_entry_point": f"{__name__}.pickplace_gr1t2_waist_enabled_mimic_env_cfg:PickPlaceGR1T2WaistEnabledMimicEnvCfg"
```
then you don't need to import anything : ))

source/isaaclab_tasks/docs/CHANGELOG.rst
ooctipus: maybe you meant IK-Abs?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/__init__.py
```diff
@@ -37,3 +43,13 @@
     },
     disable_env_checker=True,
 )
+
+gym.register(
+    id="Isaac-PickPlace-GR1T2-WaistEnabled-Abs-v0",
```
ooctipus, line 48: similarily for import here

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/nutpour_gr1t2_base_env_cfg.py
```diff
@@ -28,7 +28,7 @@
 
 from . import mdp
 
-from isaaclab_assets.robots.fourier import GR1T2_CFG  # isort: skip
+from isaaclab_assets.robots.fourier import GR1T2_PICK_PLACE_CFG  # isort: skip
```
ooctipus: It might not make much sense to have a robot name to be pick_place. Maybe you meant GR1T2_HIGH_PD_CFG?
we also have FRANKA_PANDA_CFG for rl purpose and FRANKA_PANDA_HIGH_PD_CFG for teloperation purpose, robot naming directly reflecting attribute can be more clear reflecting the property it represents.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: If your suggestion is to pass in the joint position dictionary instead of the entire ArticulationCfg, I think we can do that and its a good suggestion. But if you are asking for the users to compose the regex not from the ArticulationCfg, I think that is unnecessary. The default for this dictionary is `joint_pos: dict[str, float] = {".*": 0.0}` as specified in `articulation_cfg.py`. I am not sure in what scenario it would get this config from USD, usually it is the other way around, the USD is updated based on this joint_position dictionary.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: I'll change the re-indexing to use numpy's mechanism which O(1) runtime. Good suggestion!

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: done

source/isaaclab/isaaclab/envs/mdp/actions/pink_actions_cfg.py
michaellin6: I want to be explicit about what is imported. `isaaclab.controllers` has other things like OSC. Code is easier the way it is.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/__init__.py
michaellin6: Is this the suggested method for importing on environments? I don't know if it is in the scope of this PR, but I don't want to go changing all the environment imports as this is not the only place this happens. We can make a separate PR if you want to make that optimization.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/__init__.py
michaellin6: Same as my comment above.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/nutpour_gr1t2_base_env_cfg.py
michaellin6: done

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
ooctipus: oh I am saying that some articulation cfg underspecifies joint_pos: dict[str, float], in those case we get the default joint_pos directly from usd. So it maybe possible that articulation cfg misses some part of joint pos, and all default joint pos become fully specified(filled with USD default) after spawning is complete. (If articulation is specified, and differs from usd default, then we override for sure)

Then you could read the fully specified `joint_pos: dict[str, float]` key and value from `env.scene['robot'].data.joint_names`. `env.scene['robot'].data.default_joint_pos`. If there is a senario you don't want all keys, then you could do, `ik_robot_cfg = SceneEntityCfg('robot', joint_names=[<rexpression1>, <expression2>])`, then you can get the order by `env.scene['robot'].data.default_joint_pos[:, ik_robot_cfg.joint_ids]`

If above doesn't sounds something that fit you need, then I think your point 1 also work, `pass in the joint position dictionary instead of the entire ArticulationCfg` : ))

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/__init__.py
ooctipus: yes, it is not the scope of this PR, It will be nice to have a separate pr that make environment registration consistent and performant : ))

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
ooctipus: the .copy() is a method in @configclass that will make sure return a full copy of the dataclass with exactly all method attached.  

it is kind of like tensor.clone() for tensor object, instead of using deepcopy(tensor).

source/isaaclab_tasks/docs/CHANGELOG.rst
huihuaNvidia2023: sounds good.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_waist_enabled_mimic_env.py
huihuaNvidia2023: SG

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
```diff
@@ -0,0 +1,219 @@
...
+        _, root_nv = get_root_joint_dim(configuration.model)
+
+        # Compute configuration difference
+        return pin.difference(
+            configuration.model,
+            self.target_q,
+            configuration.q,
+        )[root_nv:]
+
+    def compute_jacobian(self, configuration: Configuration) -> np.ndarray:
```
ooctipus, line 201: is compute jacobian used anywhere? I see it is currently not by any module or task, in terms of implementation detail,
is there any chance that self.frame_task_controlled_joints is dynamic during simulation? if it is dynamic then building `frame_names` and `J_frame_task` every time seems necessary, but if it is static, I guess we could pre-cache, and just use self.J_frame_tasks

```
        if not self.frame_task_controlled_joints:
            print("No frame tasks defined")
            # Return identity matrix if no frame tasks defined
            n_joints = len(configuration.model.names) - 1  # Exclude root joint
            return np.eye(n_joints)

        # Get frame names
        frame_names = list(self.frame_task_controlled_joints.keys())

        # Get Jacobians for all frame tasks
        J_frame_tasks = [configuration.get_frame_jacobian(frame_name) for frame_name in frame_names]
```

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
```diff
@@ -0,0 +1,219 @@
...
+        # Initialize null space task matrix
+        null_space_task = np.eye(self._jacobian_dim)
+
+        # Create mask for all controlled joints
+        mask = np.zeros(self._jacobian_dim, dtype=bool)
+        for selected_indices in self._selected_joint_indices or []:
+            mask[selected_indices] = True
+
+        # Compute null space projectors efficiently
+        for J_frame_task, selected_indices in zip(J_frame_tasks, self._selected_joint_indices or []):
```
ooctipus: How many elements are there usually in J_frame_tasks? if it is more than 4, or 5. Do you think this code block is vectorizable? to vectorize you might need to build ter-tasks index selection mask, it maybe some overhead to write that code.

If your application never foresee extensive amount of J_frame_tasks, maybe it is ok to do for loop. But if it is possible in furturre, I suggest loop into pre allocate mask and memory for full size batched frames and indices, and use batched operations to process them. It will also pay off when num_envs scales as well

ooctipus (commented):
Most of the things look a lot better now, I just review the null_space_posture_task, should be the last one

source/isaaclab_tasks/docs/CHANGELOG.rst
huihuaNvidia2023: Thanks for the pointer, then it's fine to me.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/pickplace_gr1t2_env_cfg.py
huihuaNvidia2023: This is related to the discussion above with the null space task. I am closing this one.

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def compute_jacobian(self, configuration: Configuration) -> np.ndarray:
+        r"""Computes the product of left and right null space projectors
+        :math:`N(q) = I - J^+J`, where :math:`J^+` is the pseudoinverse of the
+        corresponding end-effector Jacobian. The null space projector ensures
+        joint velocities in the null space produce zero end-effector velocity
+        (:math:`J \cdot \dot{q}_{\text{null}} = 0`), allowing posture control
+        without affecting end-effector motion.
+
+            The final Jacobian returned is:
```
huihuaNvidia2023: I think the formulation still needs some thought.

- We are dealing with multi-task IK. It can be formulated either as a hierarchical problem or a simultaneous problem. These two formulation will yield the same result if there are no conflict in the tasks. Otherwise, the solution will be different.

Given two velocity tasks 1 and 2 as
```
V1 = J1 qdot_full
V2 = J2 qdot_full
```

For hierarchical one, the null space would be
```
N2 = N1 @ (I - J2_proj_pinv @ J2_proj), where
N1 = I - J1_pinv * J1 and J2_proj = J2 @ N1
```
Therefore, it will always make sure task 1 is satisfied.

For simultaneous formulation, the null space would be
```
J_cat = [J1, J2]
N = I - J_cat_pinv * J_cat
```
It will try to find a balance between the two tasks if there is conflict. 

For our case, both arms depend on the waist and there will be conflict. Since we don't have a priority which arm should be more success, we should use the simultaneous formulation.

- Another point I don't follow is why we need the mask matrix. Since we are solving for the `qdot_full`, all the jacobian should be based on `qdot_full`. For the left arm task, the null space will cover right arm by default. Therefore, we don't need to do anything special.

Finally, a unittest is needed for this task. Even though we have tried some buggy formulation, the existing tests still passed fine. A dedicated test for this task would be needed. Something as simple as Nx = 0 would be helpful.

huihuaNvidia2023 (commented):
Had some discussion about the null space calculation. PTAL.

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
ooctipus: Ok I was under impression that some task may only require fewer joint in calculation which a mask may be required. If it doesn't require a mask, it will make the implementation simpler which is great. Yes if there is priority in task, then it needs to be careful about the ordering.

 I am not against for loop if there no other way around and is the best we can do(I am also not super familiar how pink exactly solves for each task), then go for it, if in future this part maybe become a bottleneck we may revisit back and see what else could be done to optimize here. : ))

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: That is a good point. We can move some of the initialization to the init function.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: I will go with point 1 to keep it simple :)

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: @huihuaNvidia2023 I am working on this. You are right that we should be using the simultaneous formulation. I'm working on this now and adding the unit test. Thanks!

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: done

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: @huihuaNvidia2023 I have updated the null space formulation to use the simultaneous or same frame task priority formulation. Also, added a unit test that makes sure that the projected joint velocities indeed cause a zero linear velocity. I still left the joint masking method and put it into the error computation function instead (idea is basically that we don't apply posture correction objective to those joints). The reason I am still leaving the mask is because some joints typically experience much more motion relative to normal posture, specifically that is wrist joints, and when jointly optimizing for all the joint velocities they have a scaling effect on all other joints. Basically, the nullspace task works well if the wrist joint errors are not controlled.

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: The code for this has changed already so this does not apply anymore.

huihuaNvidia2023 (approved):
LGTM, thanks for the fixes.

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,28 @@
...
+
+Changed
+^^^^^^^
+
+* Updated pink inverse kinematics controller configuration for the following tasks (Isaac-PickPlace-GR1T2, Isaac-NutPour-GR1T2, Isaac-ExhaustPipe-GR1T2)
+  to increase end-effector tracking accuracy and speed. Also added a null-space regularizer that enables turning on of waist degrees-of-freedom.
+* Improved the test_pink_ik script to more comprehensive test on controller accuracy. Also, migrated to use pytest. With the current IK controller
+  improvements, our unit tests pass position and orientation accuracy test within **(1 mm, 1 degree)**. Previously, the position accuracy tolerances
+  were set to **(30 mm, 10 degrees)**.
+* Including a new config parameter :attr:`fail_on_ik_error` to :class:`~isaaclab.controllers.pink_ik.PinkIKControllerCfg`
```
Mayankm96:
```suggestion
* Included a new config parameter :attr:`fail_on_ik_error` to :class:`~isaaclab.controllers.pink_ik.PinkIKControllerCfg`
```

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
Mayankm96: If this function is a private method, could we start it with an underscore and move it to the end of the class (similar to asset classes)?

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
...
+        "left-hand": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "L_.*",
+            ],
+            stiffness=None,
+            damping=None,
+        ),
+    },
+)
+"""Configuration for the GR1T2 Humanoid robot configured for pick-place manipulation tasks."""
```
Mayankm96:
```suggestion
"""Configuration for the GR1T2 Humanoid robot configured with high PD gains for manipulation tasks."""
```

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
     },
 )
 """Configuration for the GR1T2 Humanoid robot."""
+
+
+GR1T2_HIGH_PD_CFG = GR1T2_CFG.replace(
+    actuators={
+        "trunk": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "waist_.*",
```
Mayankm96:
```suggestion
                "waist_.*"
```

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
...
+            velocity_limit=None,
+            stiffness=4400,
+            damping=40.0,
+            armature=0.01,
+        ),
+        "right-arm": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "right_shoulder_.*",
+                "right_elbow_.*",
+                "right_wrist_.*",
```
Mayankm96:
```suggestion
                "right_wrist_.*"
```

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
...
+            ],
+            stiffness=4400.0,
+            damping=40.0,
+            armature=0.01,
+        ),
+        "left-arm": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "left_shoulder_.*",
+                "left_elbow_.*",
+                "left_wrist_.*",
```
Mayankm96:
```suggestion
                "left_wrist_.*"
```

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
...
+            ],
+            stiffness=4400.0,
+            damping=40.0,
+            armature=0.01,
+        ),
+        "right-hand": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "R_.*",
```
Mayankm96:
```suggestion
                "R_.*"
```

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
```diff
@@ -123,3 +124,54 @@
...
+            ],
+            stiffness=None,
+            damping=None,
+        ),
+        "left-hand": ImplicitActuatorCfg(
+            joint_names_expr=[
+                "L_.*",
```
Mayankm96:
```suggestion
                "L_.*"
```

source/isaaclab_tasks/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,25 @@
 Changelog
 ---------
 
+0.10.47 (2025-07-25)
+~~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+
+* New ``Isaac-PickPlace-GR1T2-WaistEnabled-Abs-v0`` environment that enables the waist degrees-of-freedom for the GR1T2 robot.
```
Mayankm96, line 10:
```suggestion
* Created ``Isaac-PickPlace-GR1T2-WaistEnabled-Abs-v0`` environment that enables the waist degrees-of-freedom for the GR1T2 robot.
```

CONTRIBUTORS.md
```diff
@@ -129,6 +131,7 @@ Guidelines for modifications:
 * Wei Yang
 * Xavier Nal
 * Xinpeng Liu
+* Yan Chang
```
Mayankm96: NIT: We keep leadership in "Acknowledgement" section

huihuaNvidia2023:
[like]  Huihua Zhao reacted to your message:
________________________________

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -223,7 +230,7 @@ def apply_actions(self):
         all_envs_joint_pos_des = torch.cat((all_envs_joint_pos_des, self._target_hand_joint_positions), dim=1)
         self._processed_actions = all_envs_joint_pos_des
 
-        self._asset.set_joint_position_target(self._processed_actions, self._joint_ids)
+        self._asset.set_joint_position_target(all_envs_joint_pos_des, self._joint_ids)
```
peterd-NV: I think it's better to leave this self._processed_actions, they are functionally the same but we now record processed actions in the HDF5 dataset for use with downstream policies. This recorded value lets joint space policies such as GNX to get the same joint targets that an IK controller ultimately sets to the robot. By using self._processed_actions here, we can better ensure that the recording always matches what the action is setting to the joints.

source/isaaclab/isaaclab/controllers/pink_ik/null_space_posture_task.py
michaellin6: we have moved all the initialization needed for this class into _build_joint_mapping already.

source/isaaclab_tasks/docs/CHANGELOG.rst
michaellin6: done

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_mimic_env.py
```diff
@@ -49,8 +49,8 @@ def target_eef_pose_to_action(
         Args:
             target_eef_pose_dict: Dictionary of 4x4 target eef pose for each end-effector.
             gripper_action_dict: Dictionary of gripper actions for each end-effector.
-            noise: Noise to add to the action. If None, no noise is added.
-            env_id: Environment index to get the action for.
+            action_noise_dict: Noise to add to the action. If None, no noise is added.
+            env_id: Environment index to get the action for. (Unused in this implementation, present for interface compatibility.)
```
peterd-NV: This shouldn't be commented as "unused in this implementation." The action_noise_dict should be used as all of our Mimic envs add noise to the eef actions.

CONTRIBUTORS.md
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab_assets/isaaclab_assets/robots/fourier.py
michaellin6: done

source/isaaclab/docs/CHANGELOG.rst
michaellin6: done

peterd-NV (approved):

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: done

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/pickplace_gr1t2_mimic_env.py
michaellin6: done

huihuaNvidia2023:
[celebrate]     Huihua Zhao reacted to your message:
________________________________

PR #3150, opened 2025-08-12
Implement Locomanipulation with Improved IK Control, G1 teleoperation and Locomotion policy

michaellin6:
# Description

Contributors: @huihuaNvidia2023, @rwiltz, @life1ess, @peterd-NV, @yanchangNvidia

These changes introduce new environments: `Isaac-PickPlace-Locomanipulation-G1-Abs-v0` and `Isaac-PickPlace-FixedBaseUpperBodyIK-G1-Abs-v0`. These environments, specially the locomanipulation one, feature:
1. The G1 robot with the retargeter for 3-finger dexterous hand to operator hands tracked by Apple Vision Pro
2. A locomotion policy trained with RL that enables G1 to walk and follow a input velocity command (vx, vy, wz, h) linear velocity in two axes, angular velocity around its pelvis' z-axis, and height distance of the pelvis from the feet.
3. Integration with upper body controller that is driven by Inverse Kinematics.

Below is a diagram showing the high level controller composition. The environments introduced have the waist controlled by IK, but it is possible to configure it such that waist DOFs are controlled by the lower body policy as well.
image

This code has been tested through the [Mimic pipeline](https://laughing-dollop-wom7611.pages.github.io/devel/source/overview/teleop_imitation.html), and the output dataset was used to [train a robomimic policy](https://laughing-dollop-wom7611.pages.github.io/devel/source/overview/teleop_imitation.html#training-an-agent) that achieves a success rate of **92% from 50 policy rollouts**.

Fixes # (issue)

Some fixes included in these changes:
1. Previously, Inverse kinematics relied on an input URDF of the robot to load a pinocchio kinematics model. This URDF was exported from the robot asset USD through USD to URDF export extension, however, if the extension is not setup properly, the joint ordering can change causing unexpected behavior due to mismatch of joint ordering between the IK controller and that of Isaac Lab. For the G1 asset, we are for now using a URDF asset directly instead of relying on the URDF exporter.
2. Previously, the Inverse Kinematics controller assumed that for its kinematics model, only the joints that are being controlled by IK need to be updated, but when combined with other controllers on other joints this assumption breaks. Added a new kinematics configuration `class:PinkKinematicsConfiguration` that allows for all joints of an articulation to be updated while only a subset are controlled by IK.

## Type of change

## Screenshots
Locomanipulation Teacher Policy - Robomimic Policy

## Checklist


rwiltz (commented):
LGTM from the teleop side of things.

rwiltz (approved):

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
```diff
@@ -0,0 +1,143 @@
...
+from isaaclab.envs.mdp.actions.joint_actions import JointPositionAction
+from isaaclab.managers.action_manager import ActionTerm
+
+if TYPE_CHECKING:
+    from isaaclab.envs import ManagerBasedEnv
+
+    from .configs.action_cfg import JointPositionPolicyActionCfg, LowerBodyActionCfg
+
+
+class LowerBodyAction(ActionTerm):
```
peterd-NV: Do you see this action term only being for G1 and the locomanip tasks defined in this directory? If it's only for G1 and this particular WBC policy I think it would be good to change the name to be more specific. Otherwise if it's generalizable then I think we could move it to source/isaaclab and be with the other action terms instead of source/isaaclab_tasks

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
```diff
@@ -0,0 +1,143 @@
...
+            history_length = 1
+        repeated_commands = base_command.unsqueeze(1).repeat(1, history_length, 1).reshape(base_command.shape[0], -1)
+        policy_input = torch.cat([repeated_commands, obs_tensor], dim=-1)
+
+        joint_actions = self._policy.forward(policy_input)
+
+        self._raw_actions[:] = joint_actions
+
+        # Apply scaling and offset to the raw actions from the policy
+        self._processed_actions = joint_actions * self._scale + self._offset
```
peterd-NV: The scaling/offset/clipping is done the same way for joint position actions. Not sure if we actually need this here since the LowerBody action space is velocity and height. 

If it's not a use case that we're using or think users can benefit from, it's better to just remove it imo, otherwise it adds confusion.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
```diff
@@ -0,0 +1,143 @@
...
+            actions: The input actions tensor.
+        """
+        # Get the cached observation from the environment
+        obs_dict = self._env.obs_buf
+        # Generate new actions from policy
+        actions = self._policy(obs_dict["policy"])
+        self._raw_actions[:] = actions
+
+        # Apply scaling and offset to the raw actions from the policy
+        self._processed_actions = self._raw_actions * self._scale + self._offset
```
peterd-NV: I believe the purpose of this action term is to extract the previous step's action from the observation term to use for the WBC? In that case is there any instance we would want to scale and add offset to the observation since that would alter what the actual output of the policy was. Same as above, if we don't ever see this being used, I think it should be removed to prevent misuse/confusion.

scripts/tools/replay_demos.py
```diff
@@ -149,6 +149,9 @@ def main():
     # create environment from loaded config
     env = gym.make(args_cli.task, cfg=env_cfg).unwrapped
 
+    # Set up camera view
+    env.sim.set_camera_view(eye=[2.5, 2.5, 2.5], target=[0.0, 0.0, 1.0])
```
peterd-NV: Can we make this an optional CLI parameter instead? Env can placement can get pretty arbitrary so any nice viewpoint we setup here might not work well for others. If we're changing it from the Lab/Sim default I think we should highlight this as a parameter otherwise users will need to go hunting for it in the script to change.

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
```diff
@@ -381,3 +382,174 @@
 
 This configuration removes most collision meshes to speed up simulation.
 """
+
+
+G1_LOCOMANIPULATION_ROBOT_CFG = ArticulationCfg(
+    spawn=sim_utils.UsdFileCfg(
+        usd_path="omniverse://isaac-dev.ov.nvidia.com/Projects/agile/Robots/Collected_g1/g1_collision_geom_simplified_bigger_offset_with_hand_collision.usd",
```
peterd-NV: Will this Nucleus link be accessible to public users? Isaac OV drive links are usually NV private and the assets need to be hosted on public S3 for users to access.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/locomanipulation_g1_mimic_env.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Extracts the gripper actuation part from a sequence of env actions (compatible with env.step).
+
+        Args:
+            actions: environment actions. The shape is (num_envs, num steps in a demo, action_dim).
+
+        Returns:
+            A dictionary of torch.Tensor gripper actions. Key to each dict is an eef_name.
+        """
+        return {"left": actions[:, 14:21], "right": actions[:, 21:]}
```
peterd-NV, line 129: I believe the action term for locomanipulation includes the lower body commands? If so then we should be careful to make sure those commands are not accidentally being included in the right gripper action here since it is not explicitly giving an end index.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/mdp/observations.py
```diff
@@ -46,59 +48,52 @@ def object_obs(
     )
 
 
-def get_left_eef_pos(
-    env: ManagerBasedRLEnv,
-) -> torch.Tensor:
+def get_left_eef_pos(env: ManagerBasedRLEnv, link_name: str) -> torch.Tensor:
```
huihuaNvidia2023: Now you have the `link_name` input, you can unify the `get_left_eef_pos` and `get_right_eef_pos` to have a `get_eef_pos` or even just `get_link_pos`. Same for the quat.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/mdp/observations.py
```diff
@@ -46,59 +48,52 @@ def object_obs(
...
     return hand_joint_states
 
 
-def get_head_state(
-    env: ManagerBasedRLEnv,
-) -> torch.Tensor:
+def get_head_state(env: ManagerBasedRLEnv, head_joint_names: list[str]) -> torch.Tensor:
     robot_joint_names = env.scene["robot"].data.joint_names
-    head_joint_names = ["head_pitch_joint", "head_roll_joint", "head_yaw_joint"]
     indexes = torch.tensor([robot_joint_names.index(name) for name in head_joint_names], dtype=torch.long)
```
huihuaNvidia2023: I feel like the `get_hand_state` and `get_head_state` can also be unified for the same reason. Also, let's stick to use the `find_joints` to get indexes.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/observations.py
```diff
@@ -0,0 +1,32 @@
...
+    joint_pos_target = asset.data.joint_pos_target
+
+    # Use joint_names from asset_cfg to find indices
+    joint_names = asset_cfg.joint_names if hasattr(asset_cfg, "joint_names") else None
+    if joint_names is None:
+        raise ValueError("asset_cfg must have 'joint_names' attribute for upper_body_last_action.")
+
+    # Find joint indices matching the provided joint_names (supports regex)
+    joint_indices, _ = asset.find_joints(joint_names)
+    joint_indices = torch.tensor(joint_indices, dtype=torch.long)
```
huihuaNvidia2023, line 27: nit: is this step necessary?

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
huihuaNvidia2023: This is used by the locomotion policy.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
huihuaNvidia2023: The design is to have it generalizable so that it can used for different robot and different polices. The GEAR's HOMIE policy can also use this setup.

I agree that we could think about moving it to a different place. The question is do we do it now or wait until we have another use case. @michaellin6 can decide.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
```diff
@@ -0,0 +1,236 @@
...
+        object_rot = ObsTerm(func=base_mdp.root_quat_w, params={"asset_cfg": SceneEntityCfg("object")})
+        robot_links_state = ObsTerm(func=manip_mdp.get_all_robot_link_state)
+
+        left_eef_pos = ObsTerm(func=manip_mdp.get_left_eef_pos, params={"link_name": "left_wrist_yaw_link"})
+        left_eef_quat = ObsTerm(func=manip_mdp.get_left_eef_quat, params={"link_name": "left_wrist_yaw_link"})
+        right_eef_pos = ObsTerm(func=manip_mdp.get_right_eef_pos, params={"link_name": "right_wrist_yaw_link"})
+        right_eef_quat = ObsTerm(func=manip_mdp.get_right_eef_quat, params={"link_name": "right_wrist_yaw_link"})
+
+        hand_joint_state = ObsTerm(func=manip_mdp.get_hand_state, params={"hand_joint_names": [".*_hand.*"]})
+        # head_joint_state = ObsTerm(func=manip_mdp.get_head_state, params={"head_joint_names": []})
```
huihuaNvidia2023: nit: remove

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
```diff
@@ -0,0 +1,236 @@
...
+            params={"left_eef_link_name": "left_wrist_yaw_link", "right_eef_link_name": "right_wrist_yaw_link"},
+        )
+
+        def __post_init__(self):
+            self.enable_corruption = False
+            self.concatenate_terms = False
+
+    # observation groups
+    policy: PolicyCfg = PolicyCfg()
+    # lower_body_policy: AgileObservationsWithHistoryCfg = AgileObservationsWithHistoryCfg()
```
huihuaNvidia2023: nit: remove

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
```diff
@@ -0,0 +1,236 @@
...
+        asset_name="robot",
+        joint_names=[
+            # "waist_.*_joint",
+            ".*_hip_.*_joint",
+            ".*_knee_joint",
+            ".*_ankle_.*_joint",
+        ],
+        scale=0.25,
+        obs_group_name="lower_body_policy",  # need to be the same name as the on in ObservationCfg
+        # policy_path=Path(__file__).parent.parent / "policy/standing_g1" / "policy_with_waist.pt",
```
huihuaNvidia2023: nit:rm

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/fixed_base_upper_body_ik_g1_env_cfg.py
```diff
@@ -0,0 +1,215 @@
...
+        """Post initialization."""
+        # general settings
+        self.decimation = 4
+        self.episode_length_s = 20.0
+        # simulation settings
+        self.sim.dt = 1 / 200  # 200Hz
+        self.sim.render_interval = 2
+
+        # Set the URDF and mesh paths for the IK controller
+        urdf_omniverse_path = "omniverse://isaac-dev.ov.nvidia.com/Projects/agile/Robots/urdf/g1/g1_minimal_with_leg_hand_collision_corrected.urdf"
```
huihuaNvidia2023: Is this accessible by the external users (likely not)? We need to come up a plan on how to expose this.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/fixed_base_upper_body_ik_g1_env_cfg.py
```diff
@@ -0,0 +1,215 @@
...
+        self.actions.upper_body_ik.controller.mesh_path = retrieve_file_path(mesh_omniverse_path)
+
+        self.teleop_devices = DevicesCfg(
+            devices={
+                "handtracking": OpenXRDeviceCfg(
+                    retargeters=[
+                        G1UpperBodyRetargeterCfg(
+                            enable_visualization=True,
+                            # OpenXR hand tracking has 26 joints per hand
+                            num_open_xr_hand_joints=2 * 26,
```
huihuaNvidia2023, line 206: Is this number irrelevant to the actual number of fingers of the robot?

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/configs/agile_locomotion_observation_cfg.py
```diff
@@ -0,0 +1,196 @@
...
+from isaaclab.managers import ObservationGroupCfg as ObsGroup
+from isaaclab.managers import ObservationTermCfg as ObsTerm
+from isaaclab.managers import SceneEntityCfg
+from isaaclab.utils import configclass
+
+import isaaclab_tasks.manager_based.locomanipulation.pick_place.mdp.observations as obs_mdp
+
+
+@configclass
+class AgileObservationsCfg(ObsGroup):
```
huihuaNvidia2023: since we have switched to a new policy and corresponding observation space, should we just remove this one?

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/locomanipulation_g1_mimic_env.py
huihuaNvidia2023: Echo on this point, same for the GR1 even though there is no lower body policy. I feel like we could use a MimicAction term to make everything explicit.

Not sure we will have time for this MR, but we can at least start the conversation.

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
huihuaNvidia2023: yeah, likely not. Same for the other place.

source/isaaclab/test/controllers/test_ik_configs/README.md
```diff
@@ -0,0 +1,124 @@
...
+    "allowed_steps_per_motion": 100,
+    "repeat": 2,
+    "requires_waist_bending": false
+}
+```
+
+## Pose Guidelines
+
+### Safe Workspace
+- **X**: -0.3 to 0.3 meters (left to right)
```
huihuaNvidia2023: curious whether this safe workspace is enforced or just a suggestion

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_lower_body_standing.py
```diff
@@ -0,0 +1,27 @@
...
+from dataclasses import dataclass
+
+from isaaclab.devices.retargeter_base import RetargeterBase, RetargeterCfg
+
+
+@dataclass
+class G1LowerBodyStandingRetargeterCfg(RetargeterCfg):
+    """Configuration for the G1 lower body standing retargeter."""
+
+    hip_height: float = 0.72
```
huihuaNvidia2023, line 16: is there any documentation about how is this `hip_height` parameter used by the retargeter?

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_lower_body_retargeter.py
```diff
@@ -0,0 +1,85 @@
...
+
+import torch
+from dataclasses import dataclass
+
+from isaaclab.devices import OpenXRDevice
+from isaaclab.devices.retargeter_base import RetargeterBase, RetargeterCfg
+
+
+@dataclass
+class G1LowerBodyRetargeterCfg(RetargeterCfg):
```
huihuaNvidia2023: is the lower body retargeting required by 2.3?

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_dex_retargeting_utils.py
```diff
@@ -0,0 +1,247 @@
...
+])
+
+_OPERATOR2MANO_LEFT = np.array([
+    [0, 0, 1],
+    [1, 0, 0],
+    [0, 1, 0],
+])
+
+# G1 robot hand joint names - 2 fingers and 1 thumb configuration
+_LEFT_HAND_JOINT_NAMES = [
```
huihuaNvidia2023: given we already have hand joint names defined, can use the names to get `_HAND_JOINTS_INDEX`

source/isaaclab/isaaclab/controllers/pink_ik_cfg.py
```diff
@@ -45,10 +45,18 @@ class PinkIKControllerCfg:
...
-    joint_names: list[str] | None = None
+    controlled_joint_names: list[str] | None = None
+    """A list of joint names in the USD asset controlled by the Pink IK controller. This is required because the joint naming conventions differ between USD and URDF files.
+    This value is currently designed to be automatically populated by the action term in a manager based environment."""
+
+    all_joint_names: list[str] | None = None
     """A list of joint names in the USD asset. This is required because the joint naming conventions differ between USD and URDF files.
     This value is currently designed to be automatically populated by the action term in a manager based environment."""
 
+    controlled_joint_indices: list[int] | None = None
```
huihuaNvidia2023: do we need to define both `controlled_joint_names` and `controlled_joint_indices`?

.gitattributes
```diff
@@ -10,3 +10,4 @@
 *.pt filter=lfs diff=lfs merge=lfs -text
 *.jit filter=lfs diff=lfs merge=lfs -text
 *.hdf5 filter=lfs diff=lfs merge=lfs -text
+*.npz filter=lfs diff=lfs merge=lfs -text
```
kellyguo11: which file is this targeted for?

michaellin6:
Just rebased this branch to ToT. Will address all the previous review comments shortly.

.gitattributes
michaellin6: I will be removing this file, so will update this attribute

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik_cfg.py
```diff
@@ -45,10 +45,18 @@ class PinkIKControllerCfg:
...
-    joint_names: list[str] | None = None
+    controlled_joint_names: list[str] | None = None
+    """A list of joint names in the USD asset controlled by the Pink IK controller. This is required because the joint naming conventions differ between USD and URDF files.
+    This value is currently designed to be automatically populated by the action term in a manager based environment."""
+
+    all_joint_names: list[str] | None = None
     """A list of joint names in the USD asset. This is required because the joint naming conventions differ between USD and URDF files.
     This value is currently designed to be automatically populated by the action term in a manager based environment."""
 
+    controlled_joint_indices: list[int] | None = None
```
huihuaNvidia2023: can you please explain why both `controlled_joint_names` and `controlled_joint_indices ` are needed? This can potentially introduce inconsistency.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/mdp/observations.py
huihuaNvidia2023: Further question, there are still two separate functions for left and right. Can we further unify them?

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -162,75 +191,177 @@ def IO_descriptor(self) -> GenericActionIODescriptor:
...
+            pos_start = task_index * self.pose_dim
+            pos_end = pos_start + self.position_dim
+            quat_start = pos_end
+            quat_end = (task_index + 1) * self.pose_dim
+
+            position = actions[:, pos_start:pos_end]
+            quaternion = actions[:, quat_start:quat_end]
+
+            # Create pose matrix directly into pre-allocated tensor
+            self._controlled_frame_poses[task_index] = math_utils.make_pose(
```
huihuaNvidia2023, line 258: maybe have a try, catch here so that we can print a useful message to the user if the orders are not correct.

We now have strong assumption about the order of these information in the action tensor. We could use the rotation property (orthogonality ) , acting as one check.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
michaellin6: @peterd-NV @huihuaNvidia2023 I think for now, this action term is Agile specific as it loads the observation tensor that is required by Agile. In the long run we can split it out into a more general action term like PolicyAction or something like that. I will change the name to reflect it being more specific, that should address @peterd-NV comment.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
michaellin6: I am a bit confused about this question. The scaling and offset are being applied to the joint space actions, not to the higher level root velocity and height command.

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -162,75 +191,177 @@ def IO_descriptor(self) -> GenericActionIODescriptor:
...
+        all_joint_positions = torch.cat((ik_joint_positions, self._target_hand_joint_positions), dim=1)
+        self._processed_actions = all_joint_positions
+
+        # Apply gravity compensation to arm joints
+        self._apply_gravity_compensation()
+
+        # Apply joint position targets
+        self._asset.set_joint_position_target(self._processed_actions, self._controlled_joint_ids)
+
+    def _apply_gravity_compensation(self) -> None:
```
huihuaNvidia2023, line 323: should we add a note or a config parameter some where to tell the user about this feature? Real robots may not have this feature so the user can decide whether they want it in sim.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
michaellin6: I've removed this action term as we are using the LowerBodyAction term. This was mostly an example provided for the user to load a N-joint policy (for either upper body or full body or any combination of action space in joint space). However, I agree that since it is not being used it can be confusing.

scripts/tools/replay_demos.py
michaellin6: I've removed this as it is arbitrary location. The replay may not happen at the specific target for example.

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -46,53 +46,82 @@ def __init__(self, cfg: pink_actions_cfg.PinkInverseKinematicsActionCfg, env: Ma
...
-        # order to transform the desired pose of the controlled_frame to be with respect to the base_link_frame
-        # Shape of env.scene[self.cfg.articulation_name].data.body_link_state_w is (num_instances, num_bodies, 13)
-        base_link_frame_in_world_origin = env.scene[self.cfg.controller.articulation_name].data.body_link_state_w[
-            :,
-            env.scene[self.cfg.controller.articulation_name].data.body_names.index(self.cfg.controller.base_link_name),
-            :7,
-        ]
+        # Cache base link index to avoid string lookup every time
+        articulation_data = self._env.scene[self.cfg.controller.articulation_name].data
+        self._base_link_idx = articulation_data.body_names.index(self.cfg.controller.base_link_name)
```
huihuaNvidia2023, line 107: similarly, can we do
`self._controlled_joint_ids = articulation_data.body_names.index(self._controlled_joint_names)`

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
huihuaNvidia2023: this is minor. I am open to hear your thoughts

source/isaaclab/test/controllers/test_ik_configs/pink_ik_g1_test_configs.json
```diff
@@ -0,0 +1,107 @@
+{
+    "tolerances": {
+        "position": 0.003,
```
huihuaNvidia2023, line 3: so G1 has more tracking error than GR1? Curious about the reason

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
michaellin6: This will be moved to Nucleus link.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/locomanipulation_g1_mimic_env.py
michaellin6: Yes, I agree with @huihuaNvidia2023 we should have a code refactoring in the future to make use of ActionTerms and term.action_dim to handle these indexes as they can be a source of error.

source/isaaclab/test/controllers/test_pink_ik.py
```diff
@@ -178,27 +243,51 @@ def run_movement_test(test_setup, test_config, test_cfg, aux_function=None):
             setpoint_poses = np.concatenate([left_hand_poses[curr_pose_idx], right_hand_poses[curr_pose_idx]])
             actions = np.concatenate([setpoint_poses, np.zeros(num_joints_in_robot_hands)])
             actions = torch.tensor(actions, device=env.device, dtype=torch.float32)
+            # Append base command for Locomanipulation environments with fixed height
+            if test_setup["env_cfg"].__class__.__name__ == "LocomanipulationG1EnvCfg":
+                # Use a named variable for base height for clarity and maintainability
+                BASE_HEIGHT = 0.7
```
huihuaNvidia2023: nit: this number is 0.72

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/mdp/observations.py
michaellin6: Done

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/pick_place/mdp/observations.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/observations.py
michaellin6: yes, otherwise find_joints returns a list

source/isaaclab/test/controllers/test_ik_configs/README.md
michaellin6: It was a recommendation but I realized that it changes depending on the robot, so I have removed it.

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_lower_body_retargeter.py
michaellin6: not required so I've removed it

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_lower_body_standing.py
michaellin6: added some code documentation.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik_cfg.py
michaellin6: It is a variable that is used for passing variables from the action term pink_task_space_actions to pink_ik controller, but actually this is not a clean implementation. I will removed it and pass the indices through pink ik constructor instead.

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_dex_retargeting_utils.py
michaellin6: Let make this code re-factoring a separate future PR as I am not a subject expert on dexpilot. I think @rwiltz will have more expertise here.

source/isaaclab/test/controllers/test_ik_configs/pink_ik_g1_test_configs.json
michaellin6: Mainly due to adding the lower body locomotion. As the hip sways, the upper body has to track fast enough to keep the end-effector fixed on fixed frame.

michaellin6: If we were setting the end-effector targets fixed to the pelvis, then this would probably be much more accurate.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
peterd-NV: Yes so the action space is `AgileBasedLowerBodyAction` which means that the user does not directly interface with the joint positions right? My comment was mainly regarding if we need to have scaling and offset as parameters for this action term since the action is `[vx, vy, wz, hip_height]` which don't need any scaling. So even though the scaling/offset here is applied to the joint positions, I wonder if that's something that's actually necessary for this particular action term. If a user doesn't dig into the process_actions function, they might think that the scale/offset is being applied to the `[vx, vy, wz, hip_height]` since that's the user provided env action.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/configs/agile_locomotion_observation_cfg.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/fixed_base_upper_body_ik_g1_env_cfg.py
michaellin6: It is the number of finger joints in the human operator which is modeled as 26 joints by OpenXR.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
michaellin6: done

source/isaaclab/isaaclab/controllers/pink_ik_cfg.py
michaellin6: it was redundant so I have removed indices

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
michaellin6: Ok, I understand your point now, and I can see the possible confusion for some users but I think for the more advanced user that would take this action term and use it with their own trained Agile policy in the future it would be useful to have these parameters. Do you think we can change the naming of the scale and offset to be policy_output_offset and policy_output_scale would be a good solution to avoid this confusion?

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: Good point, I will add a flag parameter in the config file to enable/disable this

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: I think using find joints for _controlled_joint_ids is better as it can work through the array of regex exps. In here we assume that the `base_link_name` is matching exactly.

source/isaaclab/test/controllers/test_pink_ik.py
michaellin6: done

source/isaaclab/isaaclab/devices/openxr/retargeters/humanoid/unitree/g1_dex_retargeting_utils.py
huihuaNvidia2023: sounds good.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -56,14 +56,25 @@ def __init__(self, cfg: PinkIKControllerCfg, robot_cfg: ArticulationCfg, device:
...
-        self.pink_configuration = Configuration(
-            self.robot_wrapper.model, self.robot_wrapper.data, self.robot_wrapper.q0
+        if cfg.controlled_joint_names is None:
+            raise ValueError("controlled_joint_names must be provided in the configuration")
+        if cfg.all_joint_names is None:
+            raise ValueError("all_joint_names must be provided in the configuration")
+
+        self.cfg = cfg
+        self.device = device
+        self.controlled_joint_indices = controlled_joint_indices
```
huihuaNvidia2023, line 71: I am still a little bit confused here. `controlled_joint_indices` is passed in as an argument, while at the same time the `cfg` also have `controlled_joint_names`. Do we need to worry about the inconsistency between them?

huihuaNvidia2023: I see you added doc string about `controlled_joint_names`. Both of them are USD based.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/configs/agile_locomotion_observation_cfg.py
```diff
@@ -0,0 +1,79 @@
...
+
+from isaaclab.envs import mdp
+from isaaclab.managers import ObservationGroupCfg as ObsGroup
+from isaaclab.managers import ObservationTermCfg as ObsTerm
+from isaaclab.managers import SceneEntityCfg
+from isaaclab.utils import configclass
+
+
+@configclass
+class AgileTeacherPolicyObservationsCfg(ObsGroup):
```
huihuaNvidia2023, line 14: This can be a little bit confusing as this is only partial observation. There is no good way to tackle this at the point given the workflow. It would be nice to have a note here explaining that we don't use command term in the Mimic workflow. If a policy depends on command, the user would have to manually assemble it in the action.

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
...
 ---------
 
+0.45.10 (2025-08-27)
+~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+
+* Added teleoperation environments for Unitree G1. This includes an environment with lower body fixed and upper body
+  controller by IK, and an environment with the lower body controlled by a policy and the upper body controlled by IK.
```
```suggestion
  controlled by IK, and an environment with the lower body controlled by a policy and the upper body controlled by IK.
```

source/isaaclab/isaaclab/controllers/local_frame_task.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import numpy as np
+from collections.abc import Sequence
+
+import pinocchio as pin
```
kellyguo11: will this get called for all workflows? would be good to make sure this works on windows.

source/isaaclab/isaaclab/controllers/local_frame_task.py
```diff
@@ -0,0 +1,102 @@
...
+        base_link_frame_name: str,
+        position_cost: float | Sequence[float],
+        orientation_cost: float | Sequence[float],
+        lm_damping: float = 0.0,
+        gain: float = 1.0,
+    ):
+        """
+        Initialize the LocalFrameTask with configuration.
+
+        Args:
```
kellyguo11: please add docstrings for all arguments

huihuaNvidia2023 (commented):
LTGM with two minor comments. Happy to discuss alternatives.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -56,14 +56,25 @@ def __init__(self, cfg: PinkIKControllerCfg, robot_cfg: ArticulationCfg, device:
...
 
         Raises:
             KeyError: When Pink joint names cannot be matched to robot configuration joint positions.
         """
-        # Initialize the robot model from URDF and mesh files
-        self.robot_wrapper = RobotWrapper.BuildFromURDF(cfg.urdf_path, cfg.mesh_path, root_joint=None)
-        self.pink_configuration = Configuration(
-            self.robot_wrapper.model, self.robot_wrapper.data, self.robot_wrapper.q0
+        if cfg.controlled_joint_names is None:
+            raise ValueError("controlled_joint_names must be provided in the configuration")
```
kellyguo11: could we include the ValueErrors in the docstring as well?

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik_cfg.py
```diff
@@ -45,7 +45,11 @@ class PinkIKControllerCfg:
     For more details, visit: https://github.com/stephane-caron/pink
     """
 
-    joint_names: list[str] | None = None
```
kellyguo11, line 48: are we able to keep this name for backwards compatibility to avoid breaking changes?

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
```diff
@@ -56,14 +56,25 @@ def __init__(self, cfg: PinkIKControllerCfg, robot_cfg: ArticulationCfg, device:
...
+            controlled_joint_indices: A list of joint indices in the USD asset controlled by the Pink IK controller.
 
         Raises:
             KeyError: When Pink joint names cannot be matched to robot configuration joint positions.
         """
-        # Initialize the robot model from URDF and mesh files
-        self.robot_wrapper = RobotWrapper.BuildFromURDF(cfg.urdf_path, cfg.mesh_path, root_joint=None)
-        self.pink_configuration = Configuration(
-            self.robot_wrapper.model, self.robot_wrapper.data, self.robot_wrapper.q0
+        if cfg.controlled_joint_names is None:
```
kellyguo11: any chance these can be backwards compatible?

source/isaaclab/isaaclab/controllers/local_frame_task.py
```diff
@@ -0,0 +1,102 @@
...
+import numpy as np
+from collections.abc import Sequence
+
+import pinocchio as pin
+from pink.tasks.frame_task import FrameTask
+
+from isaaclab.controllers.pink_kinematics_configuration import PinkKinematicsConfiguration
+
+
+class LocalFrameTask(FrameTask):
```
kellyguo11: I'm not too sure if this should belong as a Controller, it doesn't seem to follow the other interfaces with `reset`, `set_command`, `compute`

source/isaaclab/isaaclab/controllers/pink_kinematics_configuration.py
```diff
@@ -0,0 +1,178 @@
...
+
+import numpy as np
+
+import pinocchio as pin
+from pink.configuration import Configuration
+from pink.exceptions import FrameNotFound
+from pinocchio.robot_wrapper import RobotWrapper
+
+
+class PinkKinematicsConfiguration(Configuration):
```
kellyguo11: why does this not belong in the pink_ik folder?

source/isaaclab/isaaclab/controllers/utils.py
```diff
@@ -98,3 +100,67 @@ def change_revolute_to_fixed(urdf_path: str, fixed_joints: list[str], verbose: b
...
+                omni.log.warn(f"Replacing {joint} with fixed joint")
+                omni.log.warn(old_str)
+                omni.log.warn(new_str)
+            content = content.replace(old_str, new_str)
+
+    with open(urdf_path, "w") as file:
+        file.write(content)
+
+
+def load_torchscript_model(model_path: str, device: str = "cpu") -> torch.nn.Module:
```
kellyguo11: this feels a bit strange under controllers. looks like it's used in the tasks, would it make sense to move it there instead?

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -46,53 +46,78 @@ def __init__(self, cfg: pink_actions_cfg.PinkInverseKinematicsActionCfg, env: Ma
...
+        self._initialize_helper_tensors()
+
+    def _initialize_joint_info(self) -> None:
+        """Initialize joint IDs and names based on configuration."""
+        # Resolve pink controlled joints
+        self._isaaclab_controlled_joint_ids, self._isaaclab_controlled_joint_names = self._asset.find_joints(
             self.cfg.pink_controlled_joint_names
         )
-        self.cfg.controller.joint_names = self._pink_controlled_joint_names
+        self.cfg.controller.controlled_joint_names = self._isaaclab_controlled_joint_names
```
kellyguo11: are we able to keep the previous names for backwards compatibility?

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: added a checker for quaternion norm == 1

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: Do you think we can just add a consistency check to ensure that the joint indices and the joint names are the same?

Cdfghglz:
Hi, will the agile_locomotion training code be published in any form?

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
huihuaNvidia2023: Yeah, works for me as you mentioned this will affect this xr code.

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/configs/agile_locomotion_observation_cfg.py
michaellin6: I've added explanation

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/mdp/actions.py
michaellin6: I've updated the naming of both to:
policy_output_scale and policy_output_offset

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
kellyguo11: have we moved this to the corresponding isaac sim folder already on nucleus?

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
```diff
@@ -381,3 +382,174 @@
 
 This configuration removes most collision meshes to speed up simulation.
 """
+
+
+G1_LOCOMANIPULATION_ROBOT_CFG = ArticulationCfg(
```
kellyguo11: are these parameters general enough for other tasks? can we name it something like G1_29DOF_CFG?

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
kellyguo11: why do we have a new locomanipulation folder here? how does it differ from the existing loco_manipulation one?

kellyguo11: I think it would actually be best to move all of the mimic environments under isaaclab_mimic/envs instead, so that we can differentiate between IL vs. RL environments

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
```diff
@@ -0,0 +1,229 @@
...
+    lower_body_joint_pos = AgileBasedLowerBodyActionCfg(
+        asset_name="robot",
+        joint_names=[
+            ".*_hip_.*_joint",
+            ".*_knee_joint",
+            ".*_ankle_.*_joint",
+        ],
+        policy_output_scale=0.25,
+        obs_group_name="lower_body_policy",  # need to be the same name as the on in ObservationCfg
+        policy_path="omniverse://isaac-dev.ov.nvidia.com/Projects/agile/policy_checkpoints/agile_locomotion.pt",
```
kellyguo11, line 103: let's move this to the proper folder on nucleus to where we'd want to release it

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
```diff
@@ -0,0 +1,229 @@
...
+        # general settings
+        self.decimation = 4
+        self.episode_length_s = 20.0
+        # simulation settings
+        self.sim.dt = 1 / 200  # 200Hz
+        self.sim.render_interval = 2
+
+        # Set the URDF and mesh paths for the IK controller
+        urdf_omniverse_path = (
+            "omniverse://isaac-dev.ov.nvidia.com/Projects/agile/Robots/urdf/g1/g1_29dof_with_hand_exported.urdf"
```
kellyguo11: this should be moved to the isaac lab folder as well, we can make these public already given we have the licenses for them.

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: done

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik.py
michaellin6: I will change the naming to joint_names to keep backwards compatibility.

source/isaaclab/isaaclab/controllers/local_frame_task.py
michaellin6: will check

source/isaaclab/isaaclab/controllers/local_frame_task.py
michaellin6: This is not a controller but a component of pink controller. I forgot to move it to within pink_ik directory to avoid confusion. Would that solve your concern?

source/isaaclab/isaaclab/controllers/local_frame_task.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
michaellin6: will do

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/__init__.py
michaellin6: I agree with @kellyguo11 that it would be better to move all the IL envs to mimic module. @peterd-NV what do you think we should do?
1. should we move all the mimic envs in this PR?
2. move all on a separate PR? 
3. Or I can also just move the locomanipulation ens that are added in this PR and we open a new PR to move the rest. 
4. What do you all think is the best option?

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
michaellin6: Good suggestion. I changed the naming.

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
michaellin6: In progress once Renato signs off on it.

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
michaellin6: will do

source/isaaclab/isaaclab/controllers/utils.py
michaellin6: This utility function will mostly be used to load control policy checkpoints, that was why I put it in the controller utils. If I put it in tasks and then it is used in mimic later when we migrate some tasks, it will also be placed in a weird place. I could also move it into: `source/isaaclab/isaaclab/utils/io` if you think that would be more appropriate.

source/isaaclab/isaaclab/controllers/pink_kinematics_configuration.py
michaellin6: moved

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
michaellin6: done

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/pick_place/locomanipulation_g1_env_cfg.py
michaellin6: done

source/isaaclab/isaaclab/controllers/pink_ik/pink_ik_cfg.py
michaellin6: done

source/isaaclab/isaaclab/controllers/local_frame_task.py
kellyguo11: sounds good, that would make sense

source/isaaclab/isaaclab/controllers/utils.py
kellyguo11: yes, io sounds like a better place for it

source/isaaclab_tasks/isaaclab_tasks/manager_based/locomanipulation/__init__.py
kellyguo11: not sure which will be easier. we can create a new PR for the existing environments and merge that one first, and then update this one to follow that?

michaellin6: follow up on our conversation: we will open a follow-up PR to move the mimic related envs to a sub-directory in `source/isaaclab_tasks/isaaclab_tasks/manager_based/imitation`

source/isaaclab/isaaclab/controllers/local_frame_task.py
michaellin6: training tasks and a few of the tutorial scripts run without issue on Windows. Checked on a horde machine on Windows 11.

source/isaaclab/isaaclab/devices/openxr/retargeters/__init__.py
```diff
@@ -5,6 +5,8 @@
 """Retargeters for mapping input device data to robot commands."""
 
 from .humanoid.fourier.gr1t2_retargeter import GR1T2Retargeter, GR1T2RetargeterCfg
+from .humanoid.unitree.g1_lower_body_standing import G1LowerBodyStandingRetargeter, G1LowerBodyStandingRetargeterCfg
+from .humanoid.unitree.g1_upper_body_retargeter import G1UpperBodyRetargeter, G1UpperBodyRetargeterCfg
```
ooctipus: Maybe this is a scope of another PR but I am not in favor of swamping robot specific implementation into core. Maybe the right place for them is isaaclab_task, or mimic.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/locomanipulation_g1_mimic_env.py
```diff
@@ -0,0 +1,129 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+from collections.abc import Sequence
+
+import isaaclab.utils.math as PoseUtils
+from isaaclab.envs import ManagerBasedRLMimicEnv
+
+
+class LocomanipulationG1MimicEnv(ManagerBasedRLMimicEnv):
```
ooctipus, line 13: Again, if this inheritance is only meaningful to specifc robot with specific workflow in mind it might better just implement them in the task, will we start to have locomanipulation_gr1_mimic_env.py, etc in future?

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/__init__.py
```diff
@@ -44,3 +46,10 @@
     kwargs={"env_cfg_entry_point": exhaustpipe_gr1t2_mimic_env_cfg.ExhaustPipeGR1T2MimicEnvCfg},
     disable_env_checker=True,
 )
+
+gym.register(
+    id="Isaac-Locomanipulation-G1-Abs-Mimic-v0",
+    entry_point="isaaclab_mimic.envs.pinocchio_envs:LocomanipulationG1MimicEnv",
+    kwargs={"env_cfg_entry_point": locomanipulation_g1_mimic_env_cfg.LocomanipulationG1MimicEnvCfg},
```
ooctipus, line 53: Maybe after this, can we make a PR to change this to string module rather than direct import?

michaellin6: Sounds good to me

ooctipus (approved):
Implementation looks good to me, 
Most concern were legacy issues regarding where to put robot - tasks specific files and specification on gym module registration. It might not be scope of this PR, but will be good to start address them earily

source/isaaclab/isaaclab/devices/openxr/retargeters/__init__.py
rwiltz: @ooctipus We can think about moving in follow up MR, we should move all the retageters (GR1T2, Franka, etc) to keep it consistent.

source/isaaclab_mimic/isaaclab_mimic/envs/pinocchio_envs/locomanipulation_g1_mimic_env.py
michaellin6: As we already discussed, this is a good idea for code refactoring but we will leave it for future work as this is a design that was already in place and does not pertain to these changes.

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
michaellin6: Updated to nucleus asset.

source/isaaclab/isaaclab/controllers/utils.py
kellyguo11: are we still planning on moving this?

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+0.45.13 (2025-08-27)
+~~~~~~~~~~~~~~~~~~~
```
kellyguo11, line 5:
```suggestion
~~~~~~~~~~~~~~~~~~~~
```

kellyguo11:
Seeing timeout in `test_environments.py` and failure in `test_environments_with_stage_in_memory.py` in CI. could you check locally if the tests are failing or whether we need to increase the timeouts for them?

PR #3151, opened 2025-08-12
Changes the expected input of checkpoint in rsl-rl to an absolute path in documentation to avoid ambiguity

fan-ziqi:
# Description

If the user directly enters model_xxx.pt, an error will be reported. The user should be guided to enter a path

## Type of change

## Checklist


ooctipus:
@fan-ziqi Thanks for the PR : ))

PR #3172, opened 2025-08-14
Adds uv support as an alternative to conda in isaaclab.sh

KyleM73:
# Description

This PR adds support for uv as an alternative to conda for managing python dependencies. Uv is significantly faster than conda and has many useful tools.

If users wish to use the uv workflow they will need to have uv installed, but otherwise no additional dependencies are added.

Docs should be updated to describe this option. There may need to be more work done to clean when deactivating the environment. Uv does not support pre and post activation hooks like conda so cleaning up the environment variables is slightly more tricky. I would greatly appreciate feedback to improve this workflow!

## Type of change

## Checklist

- [X] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [X] My changes generate no new warnings
- [X] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

### DIFF
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -130,6 +130,9 @@ extract_python_exe() {
     if ! [[ -z "${CONDA_PREFIX}" ]]; then
         # use conda python
         local python_exe=${CONDA_PREFIX}/bin/python
+    elif ! [[ -z "${VIRTUAL_ENV}" ]]; then
+        # use uv virtual environment python
+        local python_exe=${VIRTUAL_ENV}/bin/python
     else
         # use kit python
         local python_exe=${ISAACLAB_PATH}/_isaac_sim/python.sh
@@ -147,7 +150,7 @@ extract_python_exe() {
     if [ ! -f "${python_exe}" ]; then
         echo -e "[ERROR] Unable to find any Python executable at path: '${python_exe}'" >&2
         echo -e "\tThis could be due to the following reasons:" >&2
-        echo -e "\t1. Conda environment is not activated." >&2
+        echo -e "\t1. Conda or uv environment is not activated." >&2
         echo -e "\t2. Isaac Sim pip package 'isaacsim-rl' is not installed." >&2
         echo -e "\t3. Python executable is not available at the default path: ${ISAACLAB_PATH}/_isaac_sim/python.sh" >&2
         exit 1
@@ -183,10 +186,18 @@ extract_isaacsim_exe() {
 install_isaaclab_extension() {
     # retrieve the python executable
     python_exe=$(extract_python_exe)
+
+    # detect if we're in a uv environment
+    if [ -n "$VIRTUAL_ENV" ] && [ -f "$VIRTUAL_ENV/pyvenv.cfg" ] && grep -q "uv" "$VIRTUAL_ENV/pyvenv.cfg"; then
+        pip_cmd="uv pip install"
+    else
+        pip_cmd="${python_exe} -m pip install"
+    fi
+
     # if the directory contains setup.py then install the python module
     if [ -f "$1/setup.py" ]; then
         echo -e "\t module: $1"
-        ${python_exe} -m pip install --editable $1
+        $pip_cmd --editable "$1"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


isaaclab.sh
```diff
@@ -183,10 +186,18 @@ extract_isaacsim_exe() {
 install_isaaclab_extension() {
     # retrieve the python executable
     python_exe=$(extract_python_exe)
+
+    # detect if we're in a uv environment
+    if [ -n "$VIRTUAL_ENV" ] && [ -f "$VIRTUAL_ENV/pyvenv.cfg" ] && grep -q "uv" "$VIRTUAL_ENV/pyvenv.cfg"; then
+        pip_cmd="uv pip install"
+    else
+        pip_cmd="${python_exe} -m pip install"
+    fi
```
Mayankm96: Nitpicking on bash - We don't have a linter for it, it seems.

```suggestion
    # detect if we're in a uv environment
    if [ -n "${VIRTUAL_ENV}" ] && [ -f "${VIRTUAL_ENV}/pyvenv.cfg" ] && grep -q "uv" "${VIRTUAL_ENV}/pyvenv.cfg"; then
        pip_cmd="uv pip install"
    else
        pip_cmd="${python_exe} -m pip install"
    fi
```

isaaclab.sh
```diff
@@ -307,6 +318,66 @@ setup_conda_env() {
...
 
+# setup uv environment for Isaac Lab
+setup_uv_env() {
+    # get environment name from input
+    local env_name="$1"
+    local python_path="$2"
+
+    # check conda is installed
+    if ! command -v uv &>/dev/null; then
+        echo "[ERROR] uv could not be found. Please install uv and try again."
```
Mayankm96, line 375: Would be helpful to share a link to a resource with isntructions for installing uv.

isaaclab.sh
```diff
@@ -361,23 +433,36 @@ while [[ $# -gt 0 ]]; do
             install_system_deps
             # install the python packages in IsaacLab/source directory
             echo "[INFO] Installing extensions inside the Isaac Lab repository..."
-            python_exe=$(extract_python_exe)
+            # detect if uv is active
+            use_uv=false
+            if [ -n "$VIRTUAL_ENV" ] && [ -f "$VIRTUAL_ENV/pyvenv.cfg" ] && grep -q "uv" "$VIRTUAL_ENV/pyvenv.cfg"; then
```
Mayankm96:
```suggestion
            if [ -n "${VIRTUAL_ENV}" ] && [ -f "${VIRTUAL_ENV}/pyvenv.cfg" ] && grep -q "uv" "${VIRTUAL_ENV}/pyvenv.cfg"; then
```

Mayankm96 (changes requested):
Mainly NITs. Could you also update the setup instructions for UV? :)

It would be good to extend the support to Windows. If you don't have access to a Windows machine, could you please make a ticket so our team can look into it? Thank you!

isaaclab.sh
KyleM73: It's very easy- on linux just `wget -qO- https://astral.sh/uv/install.sh | sh`. Both linux and windows are supported, I can link [here](https://docs.astral.sh/uv/getting-started/installation/#__tabbed_1_1) both in the docs and in a comment. I had considered including the download automatically like I do in this [Makefile](https://github.com/KyleM73/isaac_manager/blob/45435b944962f6402b80156c1b55f8dbe8dcd568/Makefile#L25) I wrote that installs everything but I thought that was too heavy handed.

KyleM73:
I'm working on updating the docs accordingly. I don'y have access to a Windows machine so would greatly appreciate some help setting it up in the `.bat` file. uv is supported on Windows so the setup should be similar.

KyleM73:
Ok I believe the docs have been updated as needed. Only the `.bat` needs to be updated. The uv install should be the same as linux, just the terminal commands for Windows is different (and maybe the method I'm using to alias `isaaclab` when in the venv). Since I can't test it I haven't edited the `.bat` file at all.

Mayankm96:
@kellyguo11 shall we go ahead and merge this? We can make a separate MR on support for Windows.

Toni-SM:
> This PR adds support for uv as an alternative to conda for managing python dependencies.

I want to say that no `conda` nor `uv` is used for managing python dependencies (what it actually does that is `pip`) but just creating virtual Python environments... where you need to install a third-party manager for this purpose :sweat_smile:

KyleM73:
> > This PR adds support for uv as an alternative to conda for managing python dependencies.
> 
> I want to say that no `conda` nor `uv` is used for managing python dependencies (what it actually does that is `pip`) but just creating virtual Python environments... where you need to install a third-party manager for this purpose 😅

For `conda` yes, here I actually use `uv pip` to perform the installs which is why it is so much faster- it is using `uv`'s resolver instead of the default `pip` resolver. `uv pip` isn't just `pip` [under the hood](https://docs.astral.sh/uv/pip/). Maybe the more accurate description of the PR is that it adds `uv` as an alternative to `conda` and `venv` for creating virtual environments and adds `uv pip` as an alternative to `pip` for python package management.

Toni-SM:
> adds uv pip as an alternative to pip for python package management.

If this is the case, then, please also update the following files to use `uv pip` (or `up add`) to install the Python packages on an `uv` setup

* https://github.com/isaac-sim/IsaacLab/blob/main/docs/source/setup/installation/pip_installation.rst
* https://github.com/isaac-sim/IsaacLab/blob/main/docs/source/setup/installation/isaaclab_pip_installation.rst

KyleM73:
Done! I added a note that `uv` users should use its version of `pip`. However, using regular `pip` is also fine too; the `uv` virtual environment still encapsulates it. Regular `pip` will just be slower to install (but otherwise the same).

kellyguo11 (approved):

kellyguo11:
@KyleM73 could you help resolve the conflict and run through the formatter again?

KyleM73:
I merged IsaacLab main into the branch and made sure all formatter tests were passing! Thanks for the help getting this over the line :)

edit: one sec need to push a bug fix, testing all the options in isaaclab.sh
edit 2: fixed!

KyleM73:
Timing comparison:

TLDR:
uv gets **10x speed up** with caching, vs. conda sees virtually no improvement.

uv is **2x as fast** as conda without caching, and with caching, uv is nearly **30x faster** for reinstalls.

uv (no torch cache):
```
uv cache clean
./isaaclab.sh --uv ilab
source ilab/bin/activate
isaaclab -i
```
real    3m3.427s
user    0m47.334s
sys 0m14.178s

uv (torch cached)
```
rm -rf ilab/
./isaaclab.sh --uv ilab
source ilab/bin/activate
isaaclab -i
```
real    0m13.632s
user    0m10.912s
sys 0m2.760s

conda (no torch cache):
```
conda clean --all
./isaaclab.sh --conda ilab
conda activate ilab
isaaclab -i
```
real    6m9.684s
user    5m48.326s
sys 0m8.500s

conda (torch cached):
```
conda env remove --name ilab
./isaaclab.sh --conda ilab
conda activate ilab
isaaclab -i
```
real    6m12.353s
user    5m48.358s
sys 0m8.008s

Note: I have a few custom extensions that were installed as well, but the timing should be representative since both install methods performed the same installations. tests were run on an Ubuntu 22.04 system with Intel 10900K CPU and 3080 12Gb GPU. Clearing caches was not included in the reported times- the times are just for creating the virtual environment and installing IsaacLab with all rl frameworks.

Mayankm96:
Adds issue #3438  to track for windows.

Thanks again for the great work and benchmarking! We should at some point just default to UV for installation and make `conda` a second option for users.

Mayankm96:
Sorry for missing this earlier but can you also add the [license for UV](https://github.com/astral-sh/uv/blob/main/LICENSE-MIT) inside `docs/licenses`. Hopefully that's the last step now and we can merge.

KyleM73:
done!

docs/licenses/dependencies/uv-license.txt
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
Mayankm96:
```suggestion
SOFTWARE.

```

PR #3174, opened 2025-08-14
Adds surface gripper support in manager-based workflow

rebeccazhang0707:
# Description

We only have a surface gripper sample with direct workflow, whereas in manager-based workflow it is not supported yet.
- Add surface gripper as an asset instance to the scene (CPU only)
- Add SurfaceGripperAction and SurfaceGripperActionCfg
- Add two TaskEnvs for testing: 

1. Isaac-Stack-Cube-UR10-Long-Suction-IK-Rel-v0  
2. Isaac-Stack-Cube-UR10-Short-Suction-IK-Rel-v0

You can test recording demos by:
`./isaaclab.sh -p scripts/tools/record_demos.py --task Isaac-Stack-Cube-UR10-Long-Suction-IK-Rel-v0 --teleop_device keyboard --device cpu`

## Type of change

## Screenshot
environments_surface_gripper

## Checklist

### DIFF
--- a/source/isaaclab_tasks/config/extension.toml
+++ b/source/isaaclab_tasks/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.10.48"
+version = "0.10.49"
 
 # Description
 title = "Isaac Lab Environments"


kellyguo11:
@AntoineRichard @ooctipus - could you please help review this one?

source/isaaclab/isaaclab/envs/mdp/actions/binary_joint_actions.py
```diff
@@ -154,3 +155,110 @@ class BinaryJointVelocityAction(BinaryJointAction):
...
+
+    cfg: actions_cfg.SurfaceGripperContinuousActionCfg
+    """The configuration of the action term."""
+
+    def process_actions(self, actions: torch.Tensor):
+        # store the raw actions
+        self._raw_actions[:] = actions
+        # For continuous actions, we directly use the input values
+        # but clamp them to the valid range [-1, 1]
+        self._processed_actions = torch.clamp(actions, min=-1.0, max=1.0)
```
AntoineRichard: It would be better to create a dedicated file for the surface gripper mdps these are technically not really binary joint actions. (Even though in their current implementation they are)

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -552,10 +552,6 @@ def reset_to(
             root_velocity = asset_state["root_velocity"].clone()
             rigid_object.write_root_pose_to_sim(root_pose, env_ids=env_ids)
             rigid_object.write_root_velocity_to_sim(root_velocity, env_ids=env_ids)
-        # surface grippers
-        for asset_name, gripper in self._surface_grippers.items():
-            asset_state = state["gripper"][asset_name]
-            gripper.write_gripper_state_to_sim(asset_state, env_ids=env_ids)
```
AntoineRichard: Is there a specific reason why this is removed? This could break feature parity with other assets?

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -642,6 +638,10 @@ def get_state(self, is_relative: bool = False) -> dict[str, dict[str, dict[str,
                 asset_state["root_pose"][:, :3] -= self.env_origins
             asset_state["root_velocity"] = rigid_object.data.root_vel_w.clone()
             state["rigid_object"][asset_name] = asset_state
+        # # surface grippers
+        # state["gripper"] = dict()
+        # for asset_name, gripper in self._surface_grippers.items():
+        #     state["gripper"][asset_name] = gripper.state.clone()
```
AntoineRichard: Do we want to remove this? This could break feature parity with other assets?

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -3,6 +3,11 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
AntoineRichard: This doesn't look right. 

It needs to be like that:
```python
# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
# All rights reserved.
#
```

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -34,21 +38,86 @@
...
             "shoulder_pan_joint": 0.0,
-            "shoulder_lift_joint": -1.712,
-            "elbow_joint": 1.712,
-            "wrist_1_joint": 0.0,
-            "wrist_2_joint": 0.0,
+            "shoulder_lift_joint": -1.5707,
+            "elbow_joint": 1.5707,
+            "wrist_1_joint": -1.5707,
+            "wrist_2_joint": -1.5707,
             "wrist_3_joint": 0.0,
```
AntoineRichard: I don't think we should change these parameters here. Other users may rely on these defaults. Could these be changed somewhere else in your code? Unless this is a needed change?

source/isaaclab/isaaclab/envs/mdp/actions/binary_joint_actions.py
rebeccazhang0707: Sure, I will modify as you suggest. Thanks!

source/isaaclab/isaaclab/scene/interactive_scene.py
rebeccazhang0707: Thanks for pointing out this. I would like to discuss with you. I removed these lines since there is no implementation of `write_gripper_state_to_sim` in `assets/surface_gripper/surface_gripper.py`. And also in get_state(), nothing is done to get surfacegripper state. So I just ignore all surface_gripper related states in reset_to() and get_state(). I see surface_gripper is added by you, is `write_gripper_state_to_sim` held somewhere I did not find? Thanks.

source/isaaclab/isaaclab/scene/interactive_scene.py
rebeccazhang0707: No, actually, these lines are added by me. I am curious why this is missing.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
rebeccazhang0707: Thanks @AntoineRichard. Sure, we will modify these setup using events in task EnvCfg.

rebeccazhang0707:
@AntoineRichard thanks for your time kindly helping review this. I have fixed all issues you recommended, except for that in `interactive_scene.py`. Please have a look.

source/isaaclab/isaaclab/scene/interactive_scene.py
AntoineRichard: Probably forgot them when drafting the first implementation... Please add them in!

source/isaaclab/isaaclab/scene/interactive_scene.py
AntoineRichard: I think we removed it in one of the iterations. It so you may have to run `set_gripper_command` and then `write_data_to_sim` since we didn't test the managed workflow, we didn't see it.

AntoineRichard: Alternatively, we could bring back the method, but it would likely just contain these two calls. I don't know if it's required strictly speaking. It doesn't help that the current implementation of these grippers is not really tensor friendly...

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
AntoineRichard: Great, I don't think you need to use the events, you could also just overload the robot config and change these when you define the environment.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -72,10 +68,10 @@
...
         joint_pos={
             "shoulder_pan_joint": 0.0,
-            "shoulder_lift_joint": -1.5707,
-            "elbow_joint": 1.5707,
-            "wrist_1_joint": -1.5707,
-            "wrist_2_joint": -1.5707,
+            "shoulder_lift_joint": -1.712,
+            "elbow_joint": 1.712,
+            "wrist_1_joint": 0.0,
+            "wrist_2_joint": 0.0,
```
AntoineRichard: These were new additions from you. So I think here you could keep your own joint configs.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -104,10 +100,10 @@
...
             "wrist_3_joint": 0.0,
```
AntoineRichard: Same here :)

AntoineRichard:
> @AntoineRichard thanks for your time kindly helping review this. I have fixed all issues you recommended, except for that in `interactive_scene.py`. Please have a look.

Thanks for your changes! It looks good, I have a couple more comments on the interactive scene and the robot configs, feel free to look at these when you have the time.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
Gina07: Thanks, done.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
Gina07: Thanks, done.

Gina07:
Hi @AntoineRichard and @kellyguo11， I have modified code given Antoine's suggestion. Please take a review. thanks.

source/isaaclab/isaaclab/envs/mdp/actions/surface_gripper_actions.py
```diff
@@ -0,0 +1,127 @@
...
+    def apply_actions(self):
+        """Apply the processed actions to the surface gripper."""
+        self._asset.set_grippers_command(self._processed_actions.view(-1))
+        self._asset.write_data_to_sim()
+
+    def reset(self, env_ids: Sequence[int] | None = None) -> None:
+        if env_ids is None:
+            self._raw_actions[:] = 0.0
+        else:
+            self._raw_actions[env_ids] = 0.0
```
AntoineRichard: This is intended to be the base surface gripper action? Should we just leave the `process_actions` method not implemented here, and implement it inside `SurfaceGripperBinaryAction`? I think it could be confusing. Can we just create a `SurfaceGripperAction` term? Or does it need to be either `SurfaceGripperBinaryAction` or `SurfaceGripperContinuousAction`? If that's the case, then I'd make sure the `SurfaceGripperAction` cannot be called on its own.

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -554,9 +554,9 @@ def reset_to(
             rigid_object.write_root_pose_to_sim(root_pose, env_ids=env_ids)
             rigid_object.write_root_velocity_to_sim(root_velocity, env_ids=env_ids)
         # surface grippers
-        for asset_name, gripper in self._surface_grippers.items():
+        for asset_name, surface_gripper in self._surface_grippers.items():
             asset_state = state["gripper"][asset_name]
-            gripper.write_gripper_state_to_sim(asset_state, env_ids=env_ids)
+            surface_gripper.set_grippers_command(asset_state)
```
AntoineRichard, line 559: You may want to write to the sim here.

AntoineRichard (changes requested):
Hey, thanks for the changes! I had missed something in the action terms, it's not really clear to me if `SurfaceGripperAction` is a valid term, or if only the Continuous and Binary are. Let's just make sure this is clear so that the workflow is clear for the users!

source/isaaclab/isaaclab/envs/mdp/actions/surface_gripper_actions.py
rebeccazhang0707: hi Antoine, thanks for your kind comments. 
- This is implemented just following `BinaryJointAction` class and `BinaryJointActionCfg` configclass. Since the `SurfaceGripperActionCfg` has no associated action term class, so I believe it cannot be called on its own in manager-based workflow. 
- But anyway you are right, currently the naming is confusing, and `ContinuousAction` is not used at all. So I would prefer to keep `SurfaceGripperBinaryAction` class only. I will revise the code for your review.

source/isaaclab/isaaclab/scene/interactive_scene.py
rebeccazhang0707: ah, I think `write_data_to_sim` is handled by interactiveScene instance all together, see line 462-474.

source/isaaclab/isaaclab/assets/surface_gripper/surface_gripper_cfg.py
```diff
@@ -7,12 +7,15 @@
 
 from isaaclab.utils import configclass
 
+from ..asset_base_cfg import AssetBaseCfg
+from .surface_gripper import SurfaceGripper
+
 
 @configclass
-class SurfaceGripperCfg:
+class SurfaceGripperCfg(AssetBaseCfg):
```
kellyguo11, line 15: @AntoineRichard - was there a specific reason why this wasn't inherited from `AssetBaseCfg`?

source/isaaclab/isaaclab/envs/mdp/actions/surface_gripper_actions.py
Gina07: hi @AntoineRichard I have refactored the code, and kept only the SurfaceGripperBinaryAction class. Could you take time for a review? thanks.

AntoineRichard (approved):
LGTM! Thanks Rebecca!
@kellyguo11 Ideally we'd need another pair of eyes on this PR.

source/isaaclab/isaaclab/assets/surface_gripper/surface_gripper_cfg.py
```diff
@@ -7,12 +7,15 @@
...
     """Configuration parameters for a surface gripper actuator."""
 
-    prim_expr: str = MISSING
+    prim_path: str = MISSING
```
kellyguo11, line 18: will this impact the existing surface gripper demo in scripts/demos/pick_and_place.py? can we please test that and make sure it can still run?

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -51,3 +51,68 @@
     },
 )
 """Configuration of UR-10 arm using implicit actuator models."""
+
+
+"""UR10 with long suction"""
+UR10_LONG_SUCTION_CFG = ArticulationCfg(
```
kellyguo11: is it possible to keep the definition the same as UR10_CFG and just override the variants field?

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
```diff
@@ -51,3 +51,68 @@
...
+            velocity_limit_sim=100.0,
+            effort_limit_sim=2000.0,
+            stiffness=800.0,
+            damping=40.0,
+        ),
+    },
+)
+
+"""UR10 with short suction"""
+UR10_SHORT_SUCTION_CFG = ArticulationCfg(
```
kellyguo11: same comment as above

source/isaaclab/isaaclab/assets/surface_gripper/surface_gripper_cfg.py
rebeccazhang0707: You are right, Kelly! This demo is not working now. @AntoineRichard is it ok for me to update the code from `prim_expr` to `prim_path`? As Kelly's last question, did you have some specific reason for this? thanks.

source/isaaclab_assets/isaaclab_assets/robots/universal_robots.py
rebeccazhang0707: Sure, good point. I will modify given your suggestion. thanks Kelly.

source/isaaclab/isaaclab/assets/surface_gripper/surface_gripper_cfg.py
AntoineRichard: I think it's because initially I had made it an actuator and forgot to add the base class when I switched to asset.

source/isaaclab/isaaclab/assets/surface_gripper/surface_gripper_cfg.py
AntoineRichard: That's perfectly fine if you modify it. I think it's because initially I had made it as an actuator, and they use the expr suffix. Which probably is why I forgot to add the asset_cfg to the config.

kellyguo11 (approved):

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/stack/config/ur10_gripper/__init__.py
```diff
@@ -0,0 +1,35 @@
...
+# Register Gym environments.
+##
+
+
+##
+# Inverse Kinematics - Relative Pose Control
+##
+
+gym.register(
+    id="Isaac-Stack-Cube-UR10-Long-Suction-IK-Rel-v0",
```
kellyguo11, line 20: could we ensure we include these environments in the documentation? https://isaac-sim.github.io/IsaacLab/main/source/overview/environments.html#manipulation and https://isaac-sim.github.io/IsaacLab/main/source/overview/environments.html#manipulation

rebeccazhang0707: Sure, Kelly, just to make sure that this should be updated in the same PR, right?

rebeccazhang0707: @kellyguo11 I updated the docs and figures for your review. But I found a bug `[Error] [omni.physx.fabric.plugin] FabricManager::initializePointInstancer mismatched prototypes on point instancer: /Visuals/FrameTransformer.` that using "--device cpu" in `record_demos.py`, the FrameTransfromer Visualizer seems not working correctly (at origin). 
image

kellyguo11: sry I added the wrong 2nd link, could you please also add the new environments to https://isaac-sim.github.io/IsaacLab/main/source/overview/environments.html#comprehensive-list-of-environments? and let's set the debug viz to False by default to avoid this issue until it's fixed.

ooctipus:
I am still not fully convinced that we need entire different Asset to handle the SurfaceGripper, and am hesitate to make all these changes to core without much review from @Mayankm96 .

(Articulation, RididBody, Deformable, Sensor, SurfaceGripper), SurfaceGripper feels should be a item under a larger umbrella.

I think there is a lot of part we can make sure it is done more nicely. If surface gripper can only support CPU and only used for mimic workflow, then it probably won't work super well with RL workflow. In that case options I am open to options such as create a asset for mimic, or create an other class of asset that is only cpu based

ooctipus (approved):

PR #3251, opened 2025-08-22
Fixes tensor construction warning in `events.py`

louislelay:
# Description

This PR removes a `UserWarning` from PyTorch about using `torch.tensor()` on an existing tensor in `events.py`. It replaces `torch.tensor(actuator.joint_indices, device=asset.device)` with .to(device) to avoid unnecessary copies.

Warning mentionned:

```bash
/home/spring/IsaacLab/source/isaaclab/isaaclab/envs/mdp/events.py:542: 
UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor). actuator_joint_indices = torch.tensor(actuator.joint_indices, device=asset.device)
```

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/mdp/events.py
+++ b/source/isaaclab/isaaclab/envs/mdp/events.py
@@ -597,13 +597,13 @@ class randomize_actuator_gains(ManagerTermBase):
                 if isinstance(actuator.joint_indices, slice):
                     global_indices = slice(None)
                 else:
-                    global_indices = torch.tensor(actuator.joint_indices, device=self.asset.device)
+                    global_indices = actuator_joint_indices = actuator.joint_indices.to(self.asset.device)
             elif isinstance(actuator.joint_indices, slice):
                 # we take the joints defined in the asset config
                 global_indices = actuator_indices = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
             else:
                 # we take the intersection of the actuator joints and the asset config joints
-                actuator_joint_indices = torch.tensor(actuator.joint_indices, device=self.asset.device)
+                actuator_joint_indices = actuator.joint_indices.to(self.asset.device)
                 asset_joint_ids = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
                 # the indices of the joints in the actuator that have to be randomized
                 actuator_indices = torch.nonzero(torch.isin(actuator_joint_indices, asset_joint_ids)).view(-1)


ooctipus:
@louislelay 

So actuator.joint_indices can only be torch.tensor or slice? if that is true maybe add `if isinstance(slice), elif isinstance(torch.Tensor), else raise error` may be a bit more readable?

On a second note, do you think check the type during __call__ time can be too expensive (I know you didn't write the original function, so no pressure to edit), (◎ ◎)ゞ why couldn't we do the check and construct indices during init construction instead?

Would love to know your thought on it. : ))

Mayankm96 (approved):

Mayankm96:
> So actuator.joint_indices can only be torch.tensor or slice? if that is true maybe add if isinstance(slice), elif isinstance(torch.Tensor), else raise error may be a bit more readable?

Yes, i think it makes sense to become more strong-type if the type is known already.

> Why couldn't we do the check and construct indices during init construction instead?

We do the checks for findding joint indices when resolving scene entities. Do you mean the check in the event term itself?

louislelay:
Hey @ooctipus and @Mayankm96, thanks for the feedback!

I’ve updated the code to make the checks type-explicit and raise on unexpected types. ~~I’ll run tests once my current training job finishes and let you know when it’s ready for review again.~~ Edit: it's good to me :)

On the second point: I agree with Octi, moving the joint index resolution to `__init__` makes sense since it stays stable during training and avoids extra work each step.

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -596,14 +596,16 @@ def randomize(data: torch.Tensor, params: tuple[float, float]) -> torch.Tensor:
...
-                    global_indices = torch.tensor(actuator.joint_indices, device=self.asset.device)
+                    raise TypeError("Actuator joint indices must be a slice or a torch.Tensor.")
             elif isinstance(actuator.joint_indices, slice):
                 # we take the joints defined in the asset config
-                global_indices = actuator_indices = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
+                global_indices = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
             else:
                 # we take the intersection of the actuator joints and the asset config joints
-                actuator_joint_indices = torch.tensor(actuator.joint_indices, device=self.asset.device)
+                actuator_joint_indices = actuator.joint_indices.to(self.asset.device)
```
Mayankm96: There isn't a case where these devices should differ IMO
```suggestion
                actuator_joint_indices = actuator.joint_indices
```

louislelay:
I tested it and accepted your changes @Mayankm96, aside from the init vs call thingy the rest is good to me!

ooctipus:
@louislelay 

```
- global_indices = actuator_indices = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
+ global_indices = torch.tensor(self.asset_cfg.joint_ids, device=self.asset.device)
```
                
                
ahh we are now missing actuator_indices and can cause error because it is not defined
I will submit an PR to fix this

PR #3269, opened 2025-08-25
Adds support for custom `ProgressReporter` to Ray integration

### LINKED ISSUE #3268: [Proposal] Custom `ProgressReporter` for Ray integration
### Proposal

I propose allowing users to provide an (optional) custom `ProgressReporter` while using Ray integration for hyperparameter tuning.

### Motivation

Right now, the Ray integration defaults to the standard `CLIReporter`, which often displays metrics that aren’t particularly relevant or at the desired frequency. Similar to how we allow users to specify a --cfg_class (e.g., `CartpoleTheiaJobCfg`), I suggest we let them optionally provide a custom `ProgressReporter` class (that goes to the `RunConfig` as an argument). If none is provided, we can fall back to the default.

### Additional context

Discussed briefly in #2847. I will go ahead and open the PR.

### Checklist

### Acceptance Criteria

- [ ] Support for custom `ProgressReporter` is added

ozhanozen:
# Description

This PR adds support for providing a custom `ProgressReporter` while doing hyperparameter tuning with Ray Integration. 

Without the PR, the Ray integration defaults to the standard `CLIReporter`, which often displays metrics that aren’t particularly relevant or at the desired frequency. Similar to how we allow users to specify a --cfg_class (e.g., `CartpoleTheiaJobCfg`), this PR lets them optionally provide a custom `ProgressReporter` class. If such is not provided, it falls back to the default.

Moreover, I have added an example inside `vision_cartpole_cfg.py` (i.e., `CustomCartpoleProgressReporter`).

One point to highlight is that the new "[context-aware progress reporting](https://github.com/ray-project/ray/issues/36949)" conflicts with custom `ProgressReporter`, so if a custom `ProgressReporter` is provided, the PR disables the context-aware progress reporting.

Fixes #3268.

## Type of change

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -73,6 +73,7 @@ Guidelines for modifications:
 * HoJin Jeon
 * Hongwei Xiong
 * Hongyu Li
+* Hougant Chen
 * Huihua Zhao
 * Iretiayo Akinola
 * Jack Zeng
--- a/docs/source/api/lab/isaaclab.devices.rst
+++ b/docs/source/api/lab/isaaclab.devices.rst
@@ -48,11 +48,13 @@ Game Pad
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 .. autoclass:: Se3Gamepad
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 Keyboard
 --------
@@ -61,11 +63,13 @@ Keyboard
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 .. autoclass:: Se3Keyboard
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 Space Mouse
 -----------
@@ -74,11 +78,13 @@ Space Mouse
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 .. autoclass:: Se3SpaceMouse
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 Haply
 -----
@@ -87,6 +93,7 @@ Haply
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 OpenXR
 ------
@@ -95,6 +102,7 @@ OpenXR
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 Manus + Vive
 ------------
@@ -103,6 +111,7 @@ Manus + Vive
     :members:
     :inherited-members:
     :show-inheritance:
+    :noindex:
 
 Retargeters
 -----------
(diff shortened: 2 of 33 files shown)


ozhanozen:
@garylvov, would love to get your feedback if you have time!

garylvov:
Hi thanks for the PR! (EDIT: didn't mean to mention the issue got this confused with early stopping)

scripts/reinforcement_learning/ray/hyperparameter_tuning/vision_cartpole_cfg.py
```diff
@@ -47,3 +48,18 @@ def __init__(self, cfg: dict = {}):
         cfg = util.populate_isaac_ray_cfg_args(cfg)
         cfg["runner_args"]["--task"] = tune.choice(["Isaac-Cartpole-RGB-TheiaTiny-v0"])
         super().__init__(cfg)
+
+
+class CustomCartpoleProgressReporter(CLIReporter):
+    def __init__(self):
+        super().__init__(
+            metric_columns={
+                "training_iteration": "iter",
```
garylvov, line 59: Do we need to reprint these labels? Could it be cleaner to print the variable name directly to avoid needing to hard-code this for other cases?

garylvov (approved):
LGTM

scripts/reinforcement_learning/ray/hyperparameter_tuning/vision_cartpole_cfg.py
ozhanozen: By default, CLIReporter prints some random metrics. When you want to print a specific one, you need to specify all that needs to be printed, one by one, including training_iteration, etc.

We could use a list rather than a dict and keep the original variable names, but sometimes the original names are very verbose due to hierarchy, not fitting the CLI. I had considered `CustomCartpoleProgressReporter` as an example to highlight what is possible, e.g., using custom names, etc. Ultimately, the exact implementation should depend on the use case. 

What do you think?

garylvov: I see I think what you did already is a pragmatic solution!

scripts/reinforcement_learning/ray/tuner.py
```diff
@@ -399,6 +418,12 @@ def __init__(self, cfg: dict):
         default=MAX_LOG_EXTRACTION_ERRORS,
         help="Max number number of LogExtractionError failures before we abort the whole tuning run.",
     )
+    parser.add_argument(
+        "--progress_reporter",
+        type=str,
+        default=None,
+        help="A progress reporter in cfg_file, must be a ProgressReporter object.",
```
Mayankm96: The help instruction here could be improved by porividing examples of what this string may look like. If there are some fixed choices, I think it's good to mention them and use `choices` in the argument.

Mayankm96: Also updating the docs would help (if there is any for this example).

ozhanozen: Thanks for the review. 

`progress_reporter` would be a custom class, inheriting from `ProgressReporter`, that the user needs to implemented according to the specific needs. `CustomCartpoleProgressReporter` is included as an example for this.

I have updated the cli instructions as:
```
Optional: name of a custom reporter class defined in the cfg_file. Must subclass ray.tune.ProgressReporter (e.g., CustomCartpoleProgressReporter).
```
and added to the` tuner.py `docstrings:

```
# Local with a custom progress reporter
 ./isaaclab.sh -p scripts/reinforcement_learning/ray/tuner.py \
 --cfg_file scripts/reinforcement_learning/ray/hyperparameter_tuning/vision_cartpole_cfg.py \
 --cfg_class CartpoleTheiaJobCfg \
 --progress_reporter CustomCartpoleProgressReporter
```
finally, updated the reference line numbers in the docs that points to these docstrings.

garylvov:
@kellyguo11 pinging on small diff PR that only touches Ray stuff that's been open for some while, can we please merge? Thank you!!!

kellyguo11 (approved):

PR #3273, opened 2025-08-26
Fixes distributed training hanging issue

kellyguo11:
# Description

We have been hunting down a strange issue in distributed training setups with rendering enabled, where often the process would hang midway through training and causes NCCL timeouts. A workaround was discovered to set `app.execution.debug.forceSerial = true`, which forces serialized scheduling of omni graph within the same thread. This appears to have resolved the hanging issue and did not cause performance regressions.

## Type of change

## Checklist

### DIFF
--- a/apps/isaaclab.python.headless.rendering.kit
+++ b/apps/isaaclab.python.headless.rendering.kit
@@ -83,6 +83,9 @@ app.updateOrder.checkForHydraRenderComplete = 1000
 app.renderer.waitIdle=true
 app.hydraEngine.waitIdle=true
 
+# Forces serial processing for omni graph to avoid NCCL timeout hangs in distributed training
+app.execution.debug.forceSerial = true
+
 app.audio.enabled = false
 
 # Enable Vulkan - avoids torch+cu12 error on windows


Mayankm96 (approved):

apps/isaaclab.python.headless.rendering.kit
```diff
@@ -83,6 +83,9 @@ app.updateOrder.checkForHydraRenderComplete = 1000
 app.renderer.waitIdle=true
 app.hydraEngine.waitIdle=true
 
+# Forces serial processing for omni graph to avoid NCCL timeout hangs in distributed training
+app.execution.debug.forceSerial = true
```
Mayankm96, line 87: Should this be propagated to other app files as well? Or it's mainly for tiled rendering?

kellyguo11: we've only noticed this issue with rendering. maybe safer to keep it just for the rendering app for now in case there are any side effects we haven't discovered yet. if it starts to become an issue with non-rendering as well, we can add it into the other app files.

PR #3276, opened 2025-08-26
Adds early stopping support for Ray integration

### LINKED ISSUE #3270: [Proposal] Early stopping while doing hyperparameter tuning with Ray integration
### Proposal

I propose adding an early stopping mechanism for Ray integration.

### Motivation

When doing hyperparameter tuning with Ray integration, the trials are continued until whatever limit is set within the training config. Even if a trial is going badly and hopelessly, Ray waits until it is finished before proceeding to the next one. This is not always the most efficient thing to do, especially when it is possible to predict how the end performance will be based on the current performance.

If we can add an optional early stopping mechanism, e.g., using the Stoppers, this would increase the hyperparameter tuning efficiency by giving up on bad trials early.

### Additional context

I have actually managed to integrate the optional provision of a custom `Stopper` that correctly raises the stop flags based on the performance metrics. However, I did not manage to stop the ongoing trials for some reason; potentially because they are on a different subprocess that runs the simulation, or the simulation being unresponsive to the stop signals from Ray.

Hi @garylvov, if you have any feedback regarding how to manage this, I would be happy to try. I could create a PR draft to show you what I did and for us to work on together if you wish.

### Checklist

### Acceptance Criteria

- [ ] Eary stopping for ray tuning is added.

ozhanozen:
# Description

This PR introduces support for early stopping in Ray integration through the `Stopper` class. It enables trials to end sooner when they are unlikely to yield useful results, reducing wasted compute time and speeding up experimentation.

Previously, when running hyperparameter tuning with Ray integration, all trials would continue until the training configuration’s maximum iterations were reached, even if a trial was clearly underperforming. This wasn’t always efficient, since poor-performing trials could often be identified early on. With this PR, an optional early stopping mechanism is introduced, allowing Ray to terminate unpromising trials sooner and improve the overall efficiency of hyperparameter tuning.

The PR also includes a `CartpoleEarlyStopper` example in `vision_cartpole_cfg.py`. This serves as a reference implementation that halts a trial if the `out_of_bounds` metric doesn’t reduce after a set number of iterations. It’s meant as a usage example: users are encouraged to create their own custom stoppers tailored to their specific use cases.

Fixes #3270.

## Type of change

## Checklist

### DIFF
--- a/scripts/reinforcement_learning/ray/tuner.py
+++ b/scripts/reinforcement_learning/ray/tuner.py
@@ -245,16 +245,16 @@ def invoke_tuning_run(
     print("[WARNING]: Not saving checkpoints, just running experiment...")
     print("[INFO]: Model parameters and metrics will be preserved.")
     print("[WARNING]: For homogeneous cluster resources only...")
+
+    # Initialize Ray
+    util.ray_init(
+        ray_address=args.ray_address,
+        log_to_driver=True,
+    )
+
     # Get available resources
     resources = util.get_gpu_node_resources()
     print(f"[INFO]: Available resources {resources}")
-
-    if not ray.is_initialized():
-        ray.init(
-            address=args.ray_address,
-            log_to_driver=True,
-            num_gpus=len(resources),
-        )
 
     print(f"[INFO]: Using config {cfg}")
 
--- a/scripts/reinforcement_learning/rl_games/train.py
+++ b/scripts/reinforcement_learning/rl_games/train.py
@@ -141,11 +141,10 @@ def main(env_cfg: ManagerBasedRLEnvCfg | DirectRLEnvCfg | DirectMARLEnvCfg, agen
     # specify directory for logging experiments
     config_name = agent_cfg["params"]["config"]["name"]
     log_root_path = os.path.join("logs", "rl_games", config_name)
-    if "pbt" in agent_cfg:
-        if agent_cfg["pbt"]["directory"] == ".":
-            log_root_path = os.path.abspath(log_root_path)
-        else:
-            log_root_path = os.path.join(agent_cfg["pbt"]["directory"], log_root_path)
+    if "pbt" in agent_cfg and agent_cfg["pbt"]["directory"] != ".":
+        log_root_path = os.path.join(agent_cfg["pbt"]["directory"], log_root_path)
+    else:
+        log_root_path = os.path.abspath(log_root_path)
 
     print(f"[INFO] Logging experiment in directory: {log_root_path}")
     # specify directory for logging runs
@@ -160,6 +159,7 @@ def main(env_cfg: ManagerBasedRLEnvCfg | DirectRLEnvCfg | DirectMARLEnvCfg, agen
     # dump the configuration into log-directory
     dump_yaml(os.path.join(log_root_path, log_dir, "params", "env.yaml"), env_cfg)
     dump_yaml(os.path.join(log_root_path, log_dir, "params", "agent.yaml"), agent_cfg)
+    print(f"Exact experiment name requested from command line: {os.path.join(log_root_path, log_dir)}")
 
     # read configurations about the agent-training
     rl_device = agent_cfg["params"]["config"]["device"]
--- a/.github/ISSUE_TEMPLATE/question.md
+++ b/.github/ISSUE_TEMPLATE/question.md
@@ -10,4 +10,4 @@ Basic questions, related to robot learning, that are not bugs or feature request
 
 Advanced/nontrivial questions, especially in areas where documentation is lacking, are very much welcome.
 
-For questions that are related to running and understanding Isaac Sim, please post them at the official [Isaac Sim forums](https://docs.omniverse.nvidia.com/app_isaacsim/app_isaacsim/isaac_sim_forums.html).
+For questions that are related to running and understanding Isaac Sim, please post them at the official [Isaac Sim forums](https://forums.developer.nvidia.com/c/omniverse/simulation/69).
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -1,0 +1,120 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+name: Check Documentation Links
+
+on:
+  # Run on pull requests that modify documentation
+  pull_request:
+    paths:
+      - 'docs/**'
+      - '**.md'
+      - '.github/workflows/check-links.yml'
+  # Run on pushes to main branches
+  push:
+    branches:
+      - main
+      - devel
+      - 'release/**'
+    paths:
+      - 'docs/**'
+      - '**.md'
+      - '.github/workflows/check-links.yml'
+  # Allow manual trigger
+  workflow_dispatch:
+  # Run weekly to catch external links that break over time
+  schedule:
+    - cron: '0 0 * * 0'  # Every Sunday at midnight UTC
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  check-links:
+    name: Check for Broken Links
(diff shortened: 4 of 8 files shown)


ozhanozen:
Hi @garylvov, here is the PR as agreed. 

I have noticed while re-testing that it sometimes works, sometimes doesn't. I think the issue is not the mechanism inside the PR, but rather that Isaac Sim does not always respond in time to the termination signal. When this happens, the next training does not start within `process_response_timeout`, hence halting all the ray main process.

One solution idea could be to check if a subprocess exists within a threshold and kill it, if it doesn't, but I do not know how to do this with stoppers as normally Ray handles this. Alternatively, would it be better to add a mechanism to `execute_job` such that if there is a halted subprocess for Isaac Sim, we kill it before starting a new subprocess no matter what?

ozhanozen:
@garylvov, a small update:

I was wrong to say it sometimes works. By coincidence, the processes were ending anyway shortly after the early stop signal, and that is why I thought it sometimes worked.

I have debugged it further and can confirm that even after a "trial" is marked as completed, the subprocess/training continues to the end. The following trial might fail, e.g., due to a lack of GPU memory.

garylvov:
Hi thanks for your further investigation.

> Alternatively, would it be better to add a mechanism to `execute_job` such that if there is a halted subprocess for Isaac Sim, we kill it before starting a new subprocess no matter what?

I think this could work, but I would be a little worried about it being too "kill happy", and erroneously shutting down processes that were experiencing an ephemeral stalling period. Perhaps we can just wait for a few moments, and if it's still halted, then kill it.

However, I think it may be non-optimal design to have a new ray process try to do cleanup of other processes before starting, as opposed to a ray process doing clean-up on its own process after it finishes. 
> I have debugged it further and can confirm that even after a "trial" is marked as completed, the subprocess/training continues to the end. The following trial might fail, e.g., due to a lack of GPU memory.

I would assume that Ray could do this well enough out of the box to stop the rogue processes, but I guess that's wishful thinking ;)

I will do some testing of this too. I think you may be onto something with some mechanism related to the Ray stopper. Maybe we can override some sort of cleanup method to aggressively SIGKILL the PID recovered by `execute_job`

ozhanozen:
> I think this could work, but I would be a little worried about it being too "kill happy", and erroneously shutting down processes that were experiencing an ephemeral stalling period. Perhaps we can just wait for a few moments, and if it's still halted, then kill it.
> 
> However, I think it may be non-optimal design to have a new ray process try to do cleanup of other processes before starting, as opposed to a ray process doing clean-up on its own process after it finishes.

I agree. Nevertheless, I still wanted to try to kill it the following way:
```
self.proc.terminate()
try:
    self.proc.wait(timeout=20)
except subprocess.TimeoutExpired:
   self.proc.kill()
   self.proc.wait()
```
but it did not work. Did you manage to get a lead on what might work in a robust way?

garylvov:
Hi @ozhanozen I have yet another workaround that may just work ;) 

I think we could add 

```
parser.add_argument("--ray-proc-id", "-rid", 
                    type=int,default=None, help="Automatically configured by Ray integration, otherwise None.") 
```
to the training scripts (RSL RL, RL Games, SKRL, etc)

Then, for each job submitted in util, we can assign an extra PID. I haven't implemented it, but I've tested it manually with the following:

```
./IsaacLab/isaaclab.sh -p IsaacLab/scripts/reinforcement_learning/rl_games/train.py --task Isaac-Cartpole-v0 --headless -rid 31415
```
Then, the following has proven to be effective at shutting down all processes.
```
pkill -9 -f "rid 31415" && sleep 5
```


I think this will be effective, as for example, having a mismatched id like below doesn't kill the process


```
pkill -9 -f "rid 31414"  # does nothing, proves that kill commands wouldn't affect unrelated running training
```

Now we just need to add some sort of exit hook to the trials to kill the assigned ray proc id

garylvov:
We could maybe use https://docs.ray.io/en/latest/tune/api/callbacks.html to have each trial clean itself up

garylvov:
Actually, now that I've played with this more, I think it's important that we SIGKILL instead of SIGINT

Maybe
```
os.kill(self.proc.pid, signal.SIGKILL)
```
would work without the extra rid stuff

garylvov:
I looked into this even more, and I believe the RID is helpful as there are so many related processes for a single isaac training session:

image'

I believe I have a minimal version that kind of works pushing progress now

garylvov:
I think it is better at killing old jobs, but it keeps on running into log extraction errors after terminating processes. Please let me know if you can figure out why!

ozhanozen:
Hi @garylvov, I had just the opportunity to check this; sorry for the delay. 

I had encountered two different issues (one regarding rl_games at #3531, one about ray initialization at #3533), which might be resulting in the log extraction errors you faced. If I combine the fixes on both these PRs with the current PR, I am able to successfully do early stopping without encountering any further errors. Could you also try again with these?

Note: You had also used `self.proc.terminate(9)`, which I believe is an invalid operation, so I reverted this back. Nevertheless, the `self.proc.kill()` (that follows `self.proc.terminate()`) should already send the same kill signal as you wanted.

garylvov:
Awesome, thank you so much for figuring it out!

Sounds good about the ``terminate`` issue, thank you for spotting that.

I'm excited to try this soon!

scripts/reinforcement_learning/ray/tuner.py
```diff
@@ -260,13 +295,14 @@ def invoke_tuning_run(cfg: dict, args: argparse.Namespace) -> None:
...
+            callbacks=[ProcessCleanupCallback(), mlflow_callback],
             checkpoint_config=ray.train.CheckpointConfig(checkpoint_frequency=0, checkpoint_at_end=False),
-            stop=LogExtractionErrorStopper(max_errors=MAX_LOG_EXTRACTION_ERRORS),
+            stop=stoppers,
         )
     else:
         raise ValueError("Unrecognized run mode.")
-
+    # RID isn't optimized as it is sampled from, but useful for cleanup later
+    cfg["runner_args"]["-rid"] = tune.sample_from(lambda _: str(random.randint(int(1e9), int(1e10) - 1)))
```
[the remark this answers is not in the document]
garylvov: bro RID is to prevent resource collisions in high parallelism lol

scripts/reinforcement_learning/rl_games/train.py
```diff
@@ -42,6 +42,9 @@
     help="if toggled, this experiment will be tracked with Weights and Biases",
 )
 parser.add_argument("--export_io_descriptors", action="store_true", default=False, help="Export IO descriptors.")
+parser.add_argument(
+    "--ray-proc-id", "-rid", type=int, default=None, help="Automatically configured by Ray integration, otherwise None."
```
kellyguo11, line 46: what happens if a user sets this by accident? if it's meant to be configured directly by ray, does it have to be exposed as a user-level command line argument?

garylvov: If a user sets it by accident, nothing different will happen than if the user didn't set it. This is a no-op in terms of functionality outside of the Ray stuff.

The reason this is added here is because one single `./isaaclab.sh -p train.py` starts many processes, all with different PIDs. For example, the following shows many PIDs associated with one single isaac run: 
image





When using python's subprocess to recover the PID, only one PID is returned out of the many subprocesses created by a single Isaac train command, and so it's hard to completely kill a run programmatically. This rid is used to correlate what group of subprocesses relate to a single Isaac lab run, so that a single isaac lab run can be `pkill`ed by it's unique RID, without affecting unrelated runs on the same machine with a different RID. This is critical to our early stopping functionality; without it, we couldn't find a good way to make sure the run was killed, resulting in hanging blocking processes on the GPU. For more context, see: https://github.com/isaac-sim/IsaacLab/pull/3276#issuecomment-3253465196 and https://github.com/isaac-sim/IsaacLab/pull/3276#issuecomment-3271655432

I tried to configure this without rid (using the isaac command itself as a unique identifier, or starting the command with RAY_PROC_ID=<id> in a bash like way so that RID isn't exposed to the user), but I couldn't get it to work. Although the RID is not as clean as it could be, I think using another method to ensure runs are killed would require a significant refactor in the training scripts (mentioned in https://github.com/isaac-sim/IsaacLab/issues/3546#issuecomment-3330217352), or significantly more investigation into how to best enable early stopping

PR #3294, opened 2025-08-28
Simplifies cross platform installation setup.py

ooctipus:
# Description

This PR 

1. makes sure(skip if already satisfy, else install) the right torch is installed before and after pip installing isaaclab packages as sometime(rare case) due to flaky setup.py and unknown library dependencies changes pytorch version gets overriden.

2. only install pink and retargeters in linux x86 or amd64 machines

## Type of change

## Screenshots

## Checklist


source/isaaclab/setup.py
```diff
@@ -47,12 +46,12 @@
     "flaky",
 ]
 
-# Additional dependencies that are only available on Linux platforms
-if platform.system() == "Linux":
-    INSTALL_REQUIRES += [
-        "pin-pink==3.1.0",  # required by isaaclab.isaaclab.controllers.pink_ik
-        "dex-retargeting==0.4.6",  # required by isaaclab.devices.openxr.retargeters.humanoid.fourier.gr1_t2_dex_retargeting_utils
-    ]
+# Append Linux x86_64–only deps via PEP 508 markers
```
Mayankm96, line 49: I think it is goot to still keep the docs on where these dependencies are needed.

source/isaaclab/setup.py
```diff
@@ -47,12 +46,12 @@
...
+X64 = "platform_machine in 'x86_64,AMD64'"
+INSTALL_REQUIRES += [
+    f"pin-pink==3.1.0 ; platform_system == 'Linux' and ({X64})",
```
kellyguo11, line 53: I don't think we should remove pin-pink for arm

source/isaaclab/setup.py
ooctipus: done

source/isaaclab/setup.py
ooctipus: done

Mayankm96 (approved):
Looks good overall.

Mostly wondering why are we enforcing torch via the install script and not the pyproject.toml file. I think back then the motivation was to force install a specific torch version (different from one shipped with IsaacSim). But is this still needed?

kellyguo11:
> Looks good overall.
> 
> Mostly wondering why are we enforcing torch via the install script and not the pyproject.toml file. I think back then the motivation was to force install a specific torch version (different from one shipped with IsaacSim). But is this still needed?

I think it'd still be needed if we want to continue support for 4.5. We might also need to do something like this eventually when Isaac Sim removes the pytorch dependency to avoid issues on windows.

PR #3298, opened 2025-08-28
Adds Raycaster with tracking for Dynamic Meshes

renezurbruegg:
# Description

This PR introduces `MultiMeshRayCaster` and `MultiMeshRayCasterCamera`, an extension of the default `RayCaster` with the following enhancements:

1. **Raycasting against multiple target types** : Supports primitive shapes (spheres, cubes, …) as well as arbitrary meshes.
2. **Dynamic mesh tracking** : Keeps track of specified meshes, enabling raycasting against moving parts (e.g., robot links, articulated bodies, or dynamic obstacles).
3. **Memory-efficient caching** : Avoids redundant memory usage by caching and reusing duplicate meshes.

This is joint work with @pascal-roth and @Mayankm96.

The default `RayCaster` was limited to static environments and required manual handling of moving meshes, which restricted its use for robotics scenarios where robots or obstacles move dynamically.  

`MultiMeshRayCaster` addresses these limitations by and now supports  **raycasting against robot parts and other moving entities**.

---
## Usage

For a quick demo, run:

```bash
python scripts/demos/sensors/multi_mesh_raycaster.py --num_envs 16 --asset_type <allegro_hand|anymal_d|multi>
```

demo image

### Drop-in replacement

Example change to migrate from `RayCasterCfg` to `MultiMeshRayCasterCfg`:

```diff
- ray_caster_cfg = RayCasterCfg(
+ ray_caster_cfg = MultiMeshRayCasterCfg(
      prim_path="{ENV_REGEX_NS}/Robot",
-     mesh_prim_paths=[
-        "/World/ground"
-     ],
+     mesh_prim_paths=[
+         "/World/Ground",
+         MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/LF_.*/visuals"),
+         MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/RF_.*/visuals"),
+         MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/LH_.*/visuals"),
+         MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/RH_.*/visuals"),
+         MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/base/visuals"),
+     ],
      pattern_cfg=patterns.GridPatternCfg(resolution=resolution, size=(5.0, 5.0)),
+     track_mesh_transforms=True,
 )
```

---

## Benchmarking & Validation

To benchmark the new raycaster, run:

```bash
python scripts/benchmarks/benchmark_ray_caster.py
```

Then plot the results with:

```bash
python scripts/benchmarks/plot_raycast_results.py
```

This will generate outputs under:
`outputs/benchmarks/raycast_benchmark...`

### Example plots

<table style="border-collapse:collapse; width:100%;">
  
  <tr>
    <td colspan="2" style="text-align:center; padding-bottom:8px;">
      big image
    </td>
  </tr>

  

  <tr>
    <td style="text-align:center; padding-right:8px;">
      left image
    </td>
    <td style="text-align:center;">
      right image
    </td>
  </tr>

  

  <tr>
    <td colspan="2" style="text-align:center; padding-top:8px;">
      bottom image
    </td>
  </tr>
</table>

---

## Type of Change

* [x] New feature (non-breaking change which adds functionality)
* [ ] This change requires a documentation update

## Checklist


source/isaaclab/isaaclab/sensors/ray_caster/ray_caster.py
```diff
@@ -3,14 +3,20 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
pascal-roth:
```suggestion
```

still have to filter the old header

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,441 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
pascal-roth:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+
+        is_global: bool = False
+        """Whether the target prim is a global object or exists for each environment instance. Defaults to False."""
+
+        is_shared: bool = False
+        """Whether the target prim is shared across all environments. Defaults to False.
+        If True, the target prim is assumed to be the same mesh in all environments. In this case, the target prim is only read once
+        and the same warp mesh is used for all environments. This provides a performance boost when the target prim
+        is shared across all environments.
+        """
```
jtigue-bdai: If a target `is_global=True` is it not true that `is_shared==True`? or is there a situation where this isn't the case?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+
+    track_mesh_transforms: bool = False
+    """Whether the meshes transformations should be tracked. Defaults to False.
+
+    Note:
+        Not tracking the mesh transformations is recommended when the meshes are static to increase performance.
+    """
+
+    merge_prim_meshes: bool = True
+    """Whether to merge the meshes under each entry of :attr:`mesh_prim_paths`."""
```
jtigue-bdai: will semantic information (are mesh ids semantic info?) be lost here if merged?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_data.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/utils.py
```diff
@@ -0,0 +1,52 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2024, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera_data.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera_cfg.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,436 @@
...
+from ..utils import compute_world_poses
+from .multi_mesh_ray_caster_data import MultiMeshRayCasterData
+from .ray_caster import RayCaster
+
+if TYPE_CHECKING:
+    from .multi_mesh_ray_caster_cfg import MultiMeshRayCasterCfg
+
+
+class MultiMeshRayCaster(RayCaster):
+    """A ray-casting sensor.
```
jtigue-bdai: might be good to add why to use MultiMeshRayCaster vs RayCaster for new users that do not fully understand the limitations of the RayCaster.

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,436 @@
...
+    def _initialize_warp_meshes(self):
+        """Parse mesh prim expressions, build (or reuse) Warp meshes, and cache per-env mesh IDs.
+
+        High-level steps (per target expression):
+            1. Resolve matching prims by regex/path expression.
+            2. Collect supported mesh child prims; merge into a single mesh if configured.
+            3. Deduplicate identical vertex buffers (exact match) to avoid uploading duplicates to Warp.
+            4. Partition mesh IDs per environment or mark as globally shared.
+            5. Optionally create physics views (articulation / rigid body / fallback XForm) and cache local offsets.
+        """
```
jtigue-bdai, line 162: add exceptions to docstring

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,436 @@
...
+                )
+                if len(mesh_prims) == 0:
+                    error_msg = (
+                        f"No mesh prims found at path: {target_prim.GetPath()} with supported types:"
+                        f" {PRIMITIVE_MESH_TYPES + ['Mesh']}"
+                        " Skipping this target."
+                    )
+                    for prim in sim_utils.get_all_matching_child_prims(target_prim.GetPath(), lambda prim: True):
+                        error_msg += f"\n - Available prim '{prim.GetPath()}' of type '{prim.GetTypeName()}'"
+                    carb.log_warn(error_msg)
```
jtigue-bdai: should this be an `omni.log_error` or should it be `warn_msg`

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,436 @@
...
+
+    def _initialize_rays_impl(self):
+        super()._initialize_rays_impl()
+        if self.cfg.update_mesh_ids:
+            self._data.ray_mesh_ids = torch.zeros(
+                self._num_envs, self.num_rays, 1, device=self.device, dtype=torch.int16
+            )
+
+    def _update_buffers_impl(self, env_ids: Sequence[int]):
+        """Fills the buffers of the sensor data."""
```
jtigue-bdai: NIT: add args to docstring

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,167 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers.
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
jtigue-bdai:
```suggestion
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera_cfg.py
```diff
@@ -0,0 +1,24 @@
...
+from isaaclab.utils import configclass
+
+from .multi_mesh_ray_caster_camera import MultiMeshRayCasterCamera
+from .multi_mesh_ray_caster_cfg import MultiMeshRayCasterCfg
+from .ray_caster_camera_cfg import RayCasterCameraCfg
+
+
+@configclass
+class MultiMeshRayCasterCameraCfg(RayCasterCameraCfg, MultiMeshRayCasterCfg):
+    """Configuration for the ray-cast sensor."""
```
jtigue-bdai:
```suggestion
    """Configuration for the multi-mesh ray-cast camera sensor."""
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+
+from isaaclab.utils import configclass
+
+from .multi_mesh_ray_caster import MultiMeshRayCaster
+from .ray_caster_cfg import RayCasterCfg
+
+
+@configclass
+class MultiMeshRayCasterCfg(RayCasterCfg):
+    """Configuration for the ray-cast sensor."""
```
jtigue-bdai:
```suggestion
    """Configuration for the multi-mesh ray-cast sensor."""
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,167 @@
...
+from .multi_mesh_ray_caster import MultiMeshRayCaster
+from .multi_mesh_ray_caster_camera_data import MultiMeshRayCasterCameraData
+from .ray_caster_camera import RayCasterCamera
+
+if TYPE_CHECKING:
+    from .multi_mesh_ray_caster_camera_cfg import MultiMeshRayCasterCameraCfg
+
+
+class MultiMeshRayCasterCamera(RayCasterCamera, MultiMeshRayCaster):
+    """A ray-casting camera sensor.
```
jtigue-bdai:
```suggestion
    """A multi-mesh ray-casting camera sensor.
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
jtigue-bdai:
```suggestion
    """A multi-mesh ray-casting sensor.
```

source/isaaclab/test/sensors/test_ray_caster.py
jtigue-bdai, line 1: can you convert this to a `pytest` instead of `unittest`

ooctipus:
It looks very good on first pass thank you @pascal-roth @renezurbruegg @Mayankm96 , but I want to spend a bit more time review this carefully a bit later!

source/isaaclab/test/sensors/test_ray_caster.py
ooctipus: same comment

ooctipus (approved):

scripts/benchmarks/plot_raycast_results.py
```diff
@@ -0,0 +1,186 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import matplotlib.pyplot as plt
+
+import pandas as pd
+
+
+def compare_single_vs_multi():
+    df = pd.read_csv("/workspace/isaaclab/outputs/benchmarks/ray_caster_benchmark_single_vs_multi.csv")
```
kellyguo11: should we make this path more customizable? or at least a relative path?

scripts/benchmarks/plot_raycast_results.py
```diff
@@ -0,0 +1,186 @@
...
+    fig.tight_layout(rect=[0, 0.03, 1, 0.97])
+
+    # Save as PNG
+    plt.savefig("outputs/benchmarks/raycast_benchmark_comparison.png")
+    print("Saved plot to outputs/benchmarks/raycast_benchmark_comparison.png")
+
+
+def compare_num_assets_vram_cache():
+    """Plots Average steps per ms vs Num assets and Avg memory vs num assets"""
+    df = pd.read_csv("/workspace/isaaclab/outputs/benchmarks/ray_caster_benchmark_num_assets.csv")
```
kellyguo11: similar comment about path

scripts/benchmarks/plot_raycast_results.py
```diff
@@ -0,0 +1,186 @@
...
+    fig.tight_layout(rect=[0, 0.03, 1, 0.97])
+
+    # Save as PNG
+    plt.savefig("outputs/benchmarks/raycast_benchmark_comparison_num_assets_vram.png")
+    print("Saved plot to outputs/benchmarks/raycast_benchmark_comparison_num_assets_vram.png")
+
+
+def compare_num_assets():
+    """Plots Average steps per ms vs Num assets and Avg memory vs num assets"""
+    df = pd.read_csv("/workspace/isaaclab/outputs/benchmarks/ray_caster_benchmark_num_assets.csv")
```
kellyguo11: same comment about path

scripts/benchmarks/plot_raycast_results.py
```diff
@@ -0,0 +1,186 @@
...
+    fig.tight_layout(rect=[0, 0.03, 1, 0.97])
+
+    # Save as PNG
+    plt.savefig("outputs/benchmarks/raycast_benchmark_comparison_num_assets.png")
+    print("Saved plot to outputs/benchmarks/raycast_benchmark_comparison_num_assets.png")
+
+
+def compare_num_faces():
+    """Plots Average steps per ms vs Num faces and Avg memory vs num faces"""
+    df = pd.read_csv("/workspace/isaaclab/outputs/benchmarks/ray_caster_benchmark_num_faces.csv")
```
kellyguo11: same comment on path

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
Mayankm96: Should this property live inside `RaycastTargetCfg`?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+        is_shared: bool = False
+        """Whether the target prim is shared across all environments. Defaults to False.
+        If True, the target prim is assumed to be the same mesh in all environments. In this case, the target prim is only read once
+        and the same warp mesh is used for all environments. This provides a performance boost when the target prim
+        is shared across all environments.
+        """
+
+    class_type: type = MultiMeshRayCaster
+
+    mesh_prim_paths: list[str | RaycastTargetCfg] = MISSING
```
Mayankm96, line 55: Should explain in the docs that if it is a string, it is casted to raycasttargetcfg with default params. otherwise behavior is unclear.

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+        Not tracking the mesh transformations is recommended when the meshes are static to increase performance.
+    """
+
+    merge_prim_meshes: bool = True
+    """Whether to merge the meshes under each entry of :attr:`mesh_prim_paths`."""
+
+    update_mesh_ids: bool = False
+    """Whether to update the mesh ids of the ray hits in the :attr:`data` container."""
+
+    reference_meshes: bool = True
```
Mayankm96, line 64: How is this different from `is_shared`? Is there a case where this might be true?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,62 @@
...
+        and the same warp mesh is used for all environments. This provides a performance boost when the target prim
+        is shared across all environments.
+        """
+
+    class_type: type = MultiMeshRayCaster
+
+    mesh_prim_paths: list[str | RaycastTargetCfg] = MISSING
+    """The list of mesh primitive paths to ray cast against."""
+
+    track_mesh_transforms: bool = False
```
Mayankm96: Should we be doing it at per-target level? 🤔

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: The case is_global implies that only one global mesh exists (e.g. ground plane). In this case, is_shared won't have any influence. I will make this more specific in the docstrings

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: Yes, the mesh IDs will get "lost/pooled", i.e. a new mesh is created on the fly consisting of the old meshes. Will update the docstring with that information

renezurbruegg: Yes, the mesh IDs will get "lost/pooled", i.e. a new mesh is created on the fly consisting of the old meshes. Will update the docstring with that information

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: There are important differences: when `reference_meshes = true`, the raycaster still parses every environment, but if it encounters duplicated meshes it reuses a reference rather than creating a new copy. This reduces memory usage, but the parsing and duplicate-checking step still has to happen for each environment.

For example, if you have 10,000 environments and each contains the same table mesh, the raycaster will still parse that table mesh 10,000 times, but it will then reference the first copy instead of storing 10,000 duplicates. This saves memory, but the upfront parsing can still take a noticeable amount of time during startup.

By contrast, `is_shared` skips the duplicate-check entirely. It assumes that the first mesh in each environment is identical and directly reuses it. Using the same example with 10,000 environments, the raycaster will only parse the table mesh once and then automatically reuse it everywhere. This makes startup much faster.

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
renezurbruegg: Personally, I prefer warnings over errors, if the issue is recoverable (which it is in this case). Will change the variable name

renezurbruegg:
Thanks a lot for the fast feedback! I believe I should have addressed all the comments that have been raised so far

scripts/demos/sensors/muli_mesh_raycaster.py
```diff
@@ -0,0 +1,290 @@
...
+
+elif args_cli.asset_type == "anymal_d":
+    asset_cfg = ANYMAL_D_CFG.replace(prim_path="{ENV_REGEX_NS}/Robot")
+    ray_caster_cfg = MultiMeshRayCasterCfg(
+        prim_path="{ENV_REGEX_NS}/Robot",
+        update_period=1 / 60,
+        offset=MultiMeshRayCasterCfg.OffsetCfg(pos=(0, -0.1, 0.3)),
+        mesh_prim_paths=[
+            "/World/Ground",
+            MultiMeshRayCasterCfg.RaycastTargetCfg(target_prim_expr="{ENV_REGEX_NS}/Robot/LF_.*/visuals"),
```
Mayankm96: Where is {ENV_REGEX_NS} getting resolved?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+
+@configclass
+class MultiMeshRayCasterCfg(RayCasterCfg):
+    """Configuration for the multi-mesh ray-cast sensor."""
+
+    @configclass
+    class RaycastTargetCfg:
+        """Configuration for different ray-cast targets."""
+
+        target_prim_expr: str = MISSING
```
Mayankm96: NIT: Would it make sense to rename this to `prim_name` since it is clear this is target?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+        """The regex to specify the target prim to ray cast against."""
+
+        is_global: bool = False
+        """Whether the target prim is a global object or exists for each environment instance. Defaults to False."""
```
Mayankm96: As I was writing more explanation, I was wondering if the expression resolves to only 1 prim, is it safe to assume that it is a global target? Otherwise, let's say, user defines this attribute to `False` but passes "/World/ground". Would in this case, this only be used to the first env?

```suggestion
        """Whether the target prim is a global object or exists for each environment instance. Defaults to False.
        
        If True, the target prim is used for ray-casting in all environments.
        Otherwise, it is only used per environment.
        """
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+
+        is_shared: bool = False
+        """Whether the target prim is assumed to be the same mesh across all environments. Defaults to False.
+
+        If True, only the first mesh is read and then reused for all environments, rather than re-parsed.
+        This provides a startup performance boost when there are many environments that all use the same asset. Note,
+        if reference_meshes is False, this flag has no effect.
```
Mayankm96:
```suggestion
        This provides a startup performance boost when there are many environments that all use the same asset.
        
        .. note::
            If :attr:`MultiMeshRayCasterCfg.reference_meshes` is False, this flag has no effect.
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+
+        merge_prim_meshes: bool = True
+        """Whether to merge the meshes when a prim contains multiple meshes. Defaults to True. Note, this will
+        create a new mesh that combines all meshes in the prim. The raycast hits mesh ids will then refer to the single
+        merged mesh."""
```
Mayankm96:
```suggestion
        """Whether to merge the parsed meshes for a prim that contains multiple meshes. Defaults to True.
        
        This will create a new mesh that combines all meshes in the parsed prim. The raycast hits mesh IDs will then refer to the single
        merged mesh.
        """
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+
+        track_mesh_transforms: bool = True
+        """Whether the mesh transformations should be tracked. Defaults to True.
+
+        Note:
```
Mayankm96:
```suggestion
        .. note::
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,66 @@
...
+            Not tracking the mesh transformations is recommended when the meshes are static to increase performance.
+        """
+
+    class_type: type = MultiMeshRayCaster
+
+    mesh_prim_paths: list[str | RaycastTargetCfg] = MISSING
+    """The list of mesh primitive paths to ray cast against. If an entry is a string, it is internally converted to a :class:`RaycastTargetCfg` with global set to true and track_mesh_transforms disabled to ensure backwards compatibility with the default raycaster."""
```
Mayankm96:
```suggestion
    """The list of mesh primitive paths to ray cast against.
    
    If an entry is a string, it is internally converted to :class:`RaycastTargetCfg` with :attr:`~RaycastTargetCfg.is_global` enabled and attr:`~RaycastTargetCfg.track_mesh_transforms` disabled. These settings ensure backwards compatibility with the default raycaster.
    """
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,162 @@
...
+        self._check_supported_data_types(cfg)
+        # initialize base class
+        MultiMeshRayCaster.__init__(self, cfg)
+        # create empty variables for storing output data
+        self._data = MultiMeshRayCasterCameraData()
+
+    def __str__(self) -> str:
+        """Returns: A string containing information about the instance."""
+        return (
+            f"Ray-Caster-Camera @ '{self.cfg.prim_path}': \n"
```
Mayankm96:
```suggestion
            f"Multi-Mesh Ray-Caster-Camera @ '{self.cfg.prim_path}': \n"
```

scripts/demos/sensors/muli_mesh_raycaster.py
renezurbruegg: Here : https://github.com/isaac-sim/IsaacLab/blob/404bc69437bb58b12bb7d94113aaa90e324d919a/source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py#L108
```python
# Resolve regex namespace if set
for cfg in self._raycast_targets_cfg:
	cfg.target_prim_expr = cfg.target_prim_expr.format(ENV_REGEX_NS="/World/envs/env_.*")
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
```diff
@@ -0,0 +1,70 @@
...
+
+        track_mesh_transforms: bool = True
+        """Whether the mesh transformations should be tracked. Defaults to True.
+
+        .. note::
+            Not tracking the mesh transformations is recommended when the meshes are static to increase performance.
+        """
+
+    class_type: type = MultiMeshRayCaster
+
```
Mayankm96, line 54: For cameras, this is just instance segmentation image, right? 🤔 

Just wondering about the overlap of this attribute and the data type. Would make sense to consolidate the two in some manner. but we can keep it for later.

One way is for cameras, if user specifies "instance_segmentation" in data type, we internally enable this flag. I can see soon people wanting to map mesh ids to specific colors but that's for future us :)

source/isaaclab/isaaclab/sensors/utils.py
```diff
@@ -0,0 +1,47 @@
...
+import torch
+
+import omni.physics.tensors.impl.api as physx
+from isaacsim.core.prims import XFormPrim
+
+from isaaclab.utils.math import convert_quat
+
+
+def compute_world_poses(
```
Mayankm96: Instead of opening a new `utils.py`, can we just move these to `sim/utils.py` and rename it to `obtain_world_pose_from_view` or something?

docs/source/api/lab/isaaclab.utils.rst
```diff
@@ -20,6 +20,7 @@
       timer
       types
       warp
+      mesh
```
Mayankm96: Nice. I think we will need to consolidate the terrain mesh utils also in this to just have one place with bunch of stuff. Right now things have become spreaded out...

Future us task for sure.

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: Very good point! I think right now, it just fails tbh.

renezurbruegg: I will remove the flag, and treat a mesh as global if only one instance was matched. Imo, there should never be the case that you have one mesh more in one environment than the others

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -198,7 +201,7 @@ def wrapper(prim_path: str | Sdf.Path, *args, **kwargs):
             # if successful, do not look at children
             # this is based on the physics behavior that nested schemas are not allowed
             if not success:
-                all_prims += child_prim.GetChildren()
+                all_prims += child_prim.GetFilteredChildren(Usd.TraverseInstanceProxies())
```
Mayankm96: Hm apply_nested won't work on instanced proxies so parsing them here is unnecssary. users have to make the asset uninstanced before they can do this function.

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -570,7 +573,72 @@ def make_uninstanceable(prim_path: str | Sdf.Path, stage: Usd.Stage | None = Non
...
+    world_transform = xform.ComputeLocalToWorldTransform(Usd.TimeCode.Default())
+    world_position = [*world_transform.ExtractTranslation()]
+    world_orientation = [
+        world_transform.ExtractRotation().GetQuaternion().GetNormalized().real,
+        *world_transform.ExtractRotation().GetQuaternion().GetNormalized().imaginary,
+    ]
+    return world_position, world_orientation
+
+
+def resolve_relative_pose(
```
Mayankm96: is it cleaner to just keep one of the two  and have optional arg for relative to?

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -570,7 +573,72 @@ def make_uninstanceable(prim_path: str | Sdf.Path, stage: Usd.Stage | None = Non
...
+
+    Returns:
+        A tuple containing the relative position (as a 3D vector) and the relative orientation (as a quaternion, format wxyz).
+    """
+
+    prim_pos, prim_quat = resolve_world_pose(prim)
+    if relative_to is None:
+        return torch.tensor(prim_pos, device=device), torch.tensor(prim_quat, device=device)
+
+    relative_to_pos, relative_to_quat = resolve_world_pose(relative_to)
```
Mayankm96: Device isn't resolved correctly.

Mayankm96: Instead of torch operation, we can also simply use pxr GfMat4 math

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -570,7 +573,72 @@ def make_uninstanceable(prim_path: str | Sdf.Path, stage: Usd.Stage | None = Non
...
+        A tuple containing the world position (as a 3D vector) and the world orientation (as a quaternion, format wxyz).
+    """
+
+    xform = UsdGeom.Xformable(prim)
+    world_transform = xform.ComputeLocalToWorldTransform(Usd.TimeCode.Default())
+    world_position = [*world_transform.ExtractTranslation()]
+    world_orientation = [
+        world_transform.ExtractRotation().GetQuaternion().GetNormalized().real,
+        *world_transform.ExtractRotation().GetQuaternion().GetNormalized().imaginary,
+    ]
```
Mayankm96: In my tests, this returned wrong values. Please check: https://github.com/isaac-sim/IsaacLab/pull/3371

kellyguo11:
`test_multi_mesh_ray_caster_camera.py` tests are failing in CI. could you also run `./isaaclab.sh -f` when the changes are ready pls?

source/isaaclab/isaaclab/utils/mesh.py
```diff
@@ -0,0 +1,132 @@
...
+    # cast into UsdGeomMesh
+    mesh = UsdGeom.Mesh(mesh_prim)
+
+    # read the vertices and faces
+    points = np.asarray(mesh.GetPointsAttr().Get()).copy()
+
+    # Load faces and convert to triangle if needed. (Default is quads)
+    num_vertex_per_face = np.asarray(mesh.GetFaceVertexCountsAttr().Get())
+    indices = np.asarray(mesh.GetFaceVertexIndicesAttr().Get())
+    return points, convert_faces_to_triangles(indices, num_vertex_per_face)
```
Mayankm96: Should return a trimesh mesh instance to match the function name?

source/isaaclab/isaaclab/utils/mesh.py
```diff
@@ -0,0 +1,132 @@
...
+    """Creates a trimesh for a plane primitive."""
+    size = (2e6, 2e6)
+    vertices = np.array([[size[0], size[1], 0], [size[0], 0.0, 0], [0.0, size[1], 0], [0.0, 0.0, 0]]) - np.array(
+        [size[0] / 2.0, size[1] / 2.0, 0.0]
+    )
+    faces = np.array([[1, 0, 2], [2, 3, 1]])
+    return trimesh.Trimesh(vertices=vertices, faces=faces)
+
+
+def _create_cube_trimesh(prim: Usd.Prim) -> trimesh.Trimesh:
```
Mayankm96: Would move all the helper functions to end of the file to keep the public methods visible first.

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: The argument has now been removed: [20772f4](https://github.com/isaac-sim/IsaacLab/pull/3298/commits/20772f47091768237c076426d41633f4d0e55dd1)

If a mesh only occurs once, it is assumed to be global.

scripts/demos/sensors/muli_mesh_raycaster.py
pascal-roth, line 1: spelling mistake of the file

muli --> multi

kellyguo11:
hmm looks like there's still a failure in CI :(
```
| /workspace/isaaclab/source/isaaclab/test/sensors/test_multi_mesh_ray_caster_camera.py  | FAILED |    30.52 |  12/21  |
```

source/isaaclab/isaaclab/sim/utils.py
renezurbruegg: Merging in your PR instead of these changes

source/isaaclab/isaaclab/sim/utils.py
renezurbruegg: Merging in your PR instead of these changes

source/isaaclab/isaaclab/sim/utils.py
renezurbruegg: Merging in your PR instead of these changes

source/isaaclab/isaaclab/sim/utils.py
renezurbruegg: Merged this PR instead.

source/isaaclab/isaaclab/utils/mesh.py
renezurbruegg: c89111203b1ce192346cba41c258a251c6ca744f

source/isaaclab/isaaclab/utils/mesh.py
renezurbruegg: c89111203b1ce192346cba41c258a251c6ca744f

renezurbruegg:
> hmm looks like there's still a failure in CI :(
> 
> ```
> | /workspace/isaaclab/source/isaaclab/test/sensors/test_multi_mesh_ray_caster_camera.py  | FAILED |    30.52 |  12/21  |
> ```

@pascal-roth could you maybe quickly have a look?

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,23 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+Changed
+^^^^^^^
+
+* Added parsing of instanced prims in :meth:`~isaaclab.sim.utils.get_all_matching_child_prims` and :meth:`~isaaclab.sim.utils.get_first_matching_child_prim`.
+  Earlier, instanced prims were skipped since :meth:`Usd.Prim.GetChildren` does not return instanced prims.
+* Added parsing of instanced prims in :meth:`~isaaclab.sim.utils.make_uninstanceable` to make all prims uninstanceable.
+
+
```
jtigue-bdai: Dont forget to add the multimesh raycaster additions

pascal-roth:
when the root prim of the sensor is an Xform prim, then we get into a huge slowdown due to IsaacSim (thanks @ooctipus for fining this issue). It will make sense to replicate the code we use for targets also for the root prim and resolve until we find a physics prim @renezurbruegg

ooctipus:
as an example

some robot has prim_path of `{ENV_REGEX_NS}/Robot` where the ArticulationRootAPI is applied
other robot may have ArticulationRootAPI applied at prim_path `{ENV_REGEX_NS}/Robot/root_joint`

if user mistakenly place the raycasterCfg's prim path to the one that doesn't have ArticulationRootAPI, you will see the slow down.

renezurbruegg:
@pascal-roth if you find some time, maybe you can update the MultiMeshRayCasterCamera to also use to `_update_rays_impl` workflow, which updates the ray directions and sensor origin

ooctipus:
```
def create_primitive_mesh(prim) -> trimesh.Trimesh:
    prim_type = prim.GetTypeName()
    if prim_type == "Cube":
        size = UsdGeom.Cube(prim).GetSizeAttr().Get()
        return trimesh.creation.box(extents=(size, size, size))
    elif prim_type == "Sphere":
        r = UsdGeom.Sphere(prim).GetRadiusAttr().Get()
        return trimesh.creation.icosphere(subdivisions=3, radius=r)
    elif prim_type == "Cylinder":
        c = UsdGeom.Cylinder(prim)
        return trimesh.creation.cylinder(radius=c.GetRadiusAttr().Get(), height=c.GetHeightAttr().Get())
    elif prim_type == "Capsule":
        c = UsdGeom.Capsule(prim)
        tri_mesh = trimesh.creation.capsule(radius=c.GetRadiusAttr().Get(), height=c.GetHeightAttr().Get())
        if c.GetAxisAttr().Get() == "X":
            # rotate −90° about Y to point the length along +X
            R = rotation_matrix(np.radians(-90), [0, 1, 0])
            tri_mesh.apply_transform(R)
        elif c.GetAxisAttr().Get() == "Y":
            # rotate +90° about X to point the length along +Y
            R = rotation_matrix(np.radians(90), [1, 0, 0])
            tri_mesh.apply_transform(R)
        return tri_mesh

    elif prim_type == "Cone":
        c = UsdGeom.Cone(prim)
        radius = c.GetRadiusAttr().Get()
        height = c.GetHeightAttr().Get()
        mesh = trimesh.creation.cone(radius=radius, height=height)
        # shift all vertices down by height/2 for usd / trimesh cone primitive definiton discrepancy
        mesh.apply_translation((0.0, 0.0, -height / 2.0))
        return mesh
    else:
        raise KeyError(f"{prim_type} is not a valid primitive mesh type")
```
@renezurbruegg @pascal-roth 
I also have something in my other code to do primitive to trimesh conversion, I see currently only 
Plane, Cube, Sphere for ray-caster, could you please add the Cone, Capsule, Cylinder as well? Be careful with the descrepancy how trimesh cone and capsule convention might be a bit different from that of USD. I have written out the convention conversion in the code snippet above and tested it worked

if you find this single function nice, I'd also just replace all collapse helpfer functions and constant to just single function.

in my dexsuite environments, this code is also used, https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/dexsuite/mdp/utils.py

once you guys are done here, I will refactor my dexsuite environment to use your guys utility : )))

ooctipus:
I also see that I have a same triangulate face utility https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/dexsuite/mdp/utils.py

```
def _triangulate_faces(prim) -> np.ndarray:
    mesh = UsdGeom.Mesh(prim)
    counts = mesh.GetFaceVertexCountsAttr().Get()
    indices = mesh.GetFaceVertexIndicesAttr().Get()
    faces = []
    it = iter(indices)
    for cnt in counts:
        poly = [next(it) for _ in range(cnt)]
        for k in range(1, cnt - 1):
            faces.append([poly[0], poly[k], poly[k + 1]])
    return np.asarray(faces, dtype=np.int64)
```

your:
```
def convert_faces_to_triangles(faces: np.ndarray, point_counts: np.ndarray) -> np.ndarray:
    # check if the mesh is already triangulated
    if (point_counts == 3).all():
        return faces.reshape(-1, 3)  # already triangulated
    all_faces = []

    vertex_counter = 0
    # Iterates over all triangles of the mesh.
    # could be very slow for large meshes
    for num_points in point_counts:
        if num_points == 3:
            # triangle
            all_faces.append(faces[vertex_counter : vertex_counter + 3])
        elif num_points == 4:
            # quads. Subdivide into two triangles
            f = faces[vertex_counter : vertex_counter + 4]
            first_triangle = f[:3]
            second_triangle = np.array([f[0], f[2], f[3]])
            all_faces.append(first_triangle)
            all_faces.append(second_triangle)
        else:
            raise RuntimeError(f"Invalid number of points per face: {num_points}")

        vertex_counter += num_points
    return np.asarray(all_faces)
```

I wonder maybe you want the code still to work beyond num_points > 4?

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -568,7 +573,93 @@ def make_uninstanceable(prim_path: str | Sdf.Path, stage: Usd.Stage | None = Non
...
+        ref_tf = ref_xform.ComputeLocalToWorldTransform(Usd.TimeCode.Default())
+        # make sure ref tf is orthonormal
+        ref_tf = ref_tf.GetOrthonormalized()
+        # compute relative transform to get prim in ref frame
+        prim_tf = prim_tf * ref_tf.GetInverse()
+
+    # extract position and orientation
+    prim_pos = [*prim_tf.ExtractTranslation()]
+    prim_quat = [prim_tf.ExtractRotationQuat().real, *prim_tf.ExtractRotationQuat().imaginary]
+    return tuple(prim_pos), tuple(prim_quat)
```
ooctipus: does not Gf.Transform(child_xf * root_xf.GetInverse()) give you matrix which you can then extract relative pos, quat and non-shear-scale all at once? (maybe you don't want scale)

checkout my implementation for getting all colliders relative pos, quat, scale relative to the root_rigid_body,
```
            xform_cache = UsdGeom.XformCache()
            root_xf = xform_cache.GetLocalToWorldTransform(prim_utils.get_prim_at_path(prim_paths[i]))
            root_tf = Gf.Transform(root_xf)
            rel_tfs = []
            root_prim_scales.append(torch.tensor(root_tf.GetScale()))
            for prim in coll_prims:
                child_xf = xform_cache.GetLocalToWorldTransform(prim)
                rel_mat_tf = Gf.Transform(child_xf * root_xf.GetInverse())
                rel_quat = rel_mat_tf.GetRotation().GetQuat()
                rel_t = torch.tensor(rel_mat_tf.GetTranslation())
                rel_q = torch.tensor([rel_quat.GetReal(), *rel_quat.GetImaginary()])
                rel_s = torch.tensor(rel_mat_tf.GetScale())
                rel_tfs.append(torch.cat([rel_t, rel_q, rel_s]))
            rel_tfs = torch.cat(rel_tfs)
```

ooctipus: ahhh i see you are already doing that with xf rather than Gf.Transform

then only recommendation is consider use xform_cache,  it should give better performance when repeatedly access transform calculation. `It caches ancestor parent xforms, so that when we query leaf prims under the same parent hierarchy, the lookup retrieves the cached parent xforms. `

source/isaaclab/isaaclab/sim/utils.py
```diff
@@ -1051,3 +1144,42 @@ def get_current_stage_id() -> int:
...
         stage_id = stage_cache.Insert(stage).ToLongInt()
     return stage_id
+
+
+"""
+PhysX prim views utils.
+"""
+
+
+def obtain_world_pose_from_view(
```
ooctipus: I am not a big fan of put this utility in core like sim/utils.py, what about other views like Deformable, SuctionGripper, to put in core, I hope we can show either it is general or it can be factored to multiple usage cases. This util feel more like just for ray_caster's convenience.

Can we make it just a helper function in multi_mesh_ray_caster folder, or even just class level helper. we can add it to sim/utils.  If there is more than one instance usage of this method in future then we can refactor it into core.

pavelacamposp:
Thanks for this feature! I've noticed that the duplicate mesh detection in `MultiMeshRayCaster._initialize_warp_meshes()` (from `multi_mesh_ray_caster.py`) doesn't scale well with the number of meshes.

Currently, `_registered_points_idx()` iterates over all registered meshes and compares their vertices. This operation is roughly O(n * V) for each new mesh (n = number of registered meshes, V = vertex count), leading to an overall O(n² * V) cost during initialization as the list of registered meshes grows (`loaded_vertices`).

A more efficient alternative could be to use a hash-based lookup table to detect mesh duplicates. We can assign a hash key to each mesh based on its vertices and use it for fast duplicate checks. This would reduce the overall complexity to O(n * V) for the entire duplicate mesh detection process.

Example implementation:
```python
def _get_mesh_key(vertices: np.ndarray) -> tuple[int, int, bytes]:
    """Build a key from the shape and data hash of a mesh vertex array."""
    data = np.ascontiguousarray(vertices).view(np.uint8)  # Ensure array is contiguous
    h = hashlib.blake2b(data, digest_size=16)
    return (vertices.shape[0], vertices.shape[1], h.digest())
```

And in `_initialize_warp_meshes()`:
```python
def _initialize_warp_meshes(self):
    ...
    for target_cfg in self._raycast_targets_cfg:
        ...
        registered_meshes: dict[tuple[int, int, bytes], int] = {}  # Maps mesh keys to wp_mesh indices 
        wp_mesh_ids = []

        for target_prim in target_prims:
        	...
            if str(target_prim.GetPath()) in MultiMeshRayCaster.meshes:
                wp_mesh_ids.append(MultiMeshRayCaster.meshes[str(target_prim.GetPath())].id)
                continue
            ...
            mesh_key = _get_mesh_key(trimesh_mesh.vertices)
            registered_idx = registered_meshes.get(mesh_key, -1)
            if registered_idx != -1 and self.cfg.reference_meshes:
                omni.log.info("Found a duplicate mesh, only reference the mesh.")
                wp_mesh_ids.append(wp_mesh_ids[registered_idx])
            else:
                wp_mesh = convert_to_warp_mesh(trimesh_mesh.vertices, trimesh_mesh.faces, device=self.device)
                MultiMeshRayCaster.meshes[str(target_prim.GetPath())] = wp_mesh
                wp_mesh_ids.append(wp_mesh.id)
                registered_meshes[mesh_key] = len(wp_mesh_ids) - 1  # Store wp_mesh idx
```

This approach should reduce initialization time significantly when processing multiple meshes (from O(n² * V) to O(n * V)).

renezurbruegg:
Thanks a lot for the valuable feedback. Hashing the vertices is a great idea to speed up retrieval. I’ll incorporate this, along with an additional check for potential hash collisions, once I find the time.

Also, note that cache lookups can be further accelerated by setting the is_shared flag in the configuration to true. This assumes that all environments share the same meshes, avoiding redundant checking for each one.

Mayankm96:
This is definitely useful feedback @pavelacamposp 

@renezurbruegg I suggest though that we keep this MR to the current limitation and make a separate one with the hashing implementation. Just to not have this MR hanging around for too long :)

renezurbruegg:
> ```
> def create_primitive_mesh(prim) -> trimesh.Trimesh:
>     prim_type = prim.GetTypeName()
>     if prim_type == "Cube":
>         size = UsdGeom.Cube(prim).GetSizeAttr().Get()
>         return trimesh.creation.box(extents=(size, size, size))
>     elif prim_type == "Sphere":
>         r = UsdGeom.Sphere(prim).GetRadiusAttr().Get()
>         return trimesh.creation.icosphere(subdivisions=3, radius=r)
>     elif prim_type == "Cylinder":
>         c = UsdGeom.Cylinder(prim)
>         return trimesh.creation.cylinder(radius=c.GetRadiusAttr().Get(), height=c.GetHeightAttr().Get())
>     elif prim_type == "Capsule":
>         c = UsdGeom.Capsule(prim)
>         tri_mesh = trimesh.creation.capsule(radius=c.GetRadiusAttr().Get(), height=c.GetHeightAttr().Get())
>         if c.GetAxisAttr().Get() == "X":
>             # rotate −90° about Y to point the length along +X
>             R = rotation_matrix(np.radians(-90), [0, 1, 0])
>             tri_mesh.apply_transform(R)
>         elif c.GetAxisAttr().Get() == "Y":
>             # rotate +90° about X to point the length along +Y
>             R = rotation_matrix(np.radians(90), [1, 0, 0])
>             tri_mesh.apply_transform(R)
>         return tri_mesh
> 
>     elif prim_type == "Cone":
>         c = UsdGeom.Cone(prim)
>         radius = c.GetRadiusAttr().Get()
>         height = c.GetHeightAttr().Get()
>         mesh = trimesh.creation.cone(radius=radius, height=height)
>         # shift all vertices down by height/2 for usd / trimesh cone primitive definiton discrepancy
>         mesh.apply_translation((0.0, 0.0, -height / 2.0))
>         return mesh
>     else:
>         raise KeyError(f"{prim_type} is not a valid primitive mesh type")
> ```
> 
> @renezurbruegg @pascal-roth I also have something in my other code to do primitive to trimesh conversion, I see currently only Plane, Cube, Sphere for ray-caster, could you please add the Cone, Capsule, Cylinder as well? Be careful with the descrepancy how trimesh cone and capsule convention might be a bit different from that of USD. I have written out the convention conversion in the code snippet above and tested it worked
> 
> if you find this single function nice, I'd also just replace all collapse helpfer functions and constant to just single function.
> 
> in my dexsuite environments, this code is also used, https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/dexsuite/mdp/utils.py
> 
> once you guys are done here, I will refactor my dexsuite environment to use your guys utility : )))

Thanks a lot for the input. I added all prim types now and verified that they work correctly:
image

renezurbruegg:
> I also see that I have a same triangulate face utility https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/dexsuite/mdp/utils.py
> 
> ```
> def _triangulate_faces(prim) -> np.ndarray:
>     mesh = UsdGeom.Mesh(prim)
>     counts = mesh.GetFaceVertexCountsAttr().Get()
>     indices = mesh.GetFaceVertexIndicesAttr().Get()
>     faces = []
>     it = iter(indices)
>     for cnt in counts:
>         poly = [next(it) for _ in range(cnt)]
>         for k in range(1, cnt - 1):
>             faces.append([poly[0], poly[k], poly[k + 1]])
>     return np.asarray(faces, dtype=np.int64)
> ```
> 
> your:
> 
> ```
> def convert_faces_to_triangles(faces: np.ndarray, point_counts: np.ndarray) -> np.ndarray:
>     # check if the mesh is already triangulated
>     if (point_counts == 3).all():
>         return faces.reshape(-1, 3)  # already triangulated
>     all_faces = []
> 
>     vertex_counter = 0
>     # Iterates over all triangles of the mesh.
>     # could be very slow for large meshes
>     for num_points in point_counts:
>         if num_points == 3:
>             # triangle
>             all_faces.append(faces[vertex_counter : vertex_counter + 3])
>         elif num_points == 4:
>             # quads. Subdivide into two triangles
>             f = faces[vertex_counter : vertex_counter + 4]
>             first_triangle = f[:3]
>             second_triangle = np.array([f[0], f[2], f[3]])
>             all_faces.append(first_triangle)
>             all_faces.append(second_triangle)
>         else:
>             raise RuntimeError(f"Invalid number of points per face: {num_points}")
> 
>         vertex_counter += num_points
>     return np.asarray(all_faces)
> ```
> 
> I wonder maybe you want the code still to work beyond num_points > 4?

Thanks, Updated it with your generic fan triangulation

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
renezurbruegg: I would keep "expr" in there, since it can be a regex pattern. I could rename it to "prim_expr" or "target_expr". What do you think?

Mayankm96:
@renezurbruegg @pascal-roth Should we review this again and then prepare it for merge?

renezurbruegg:
> @renezurbruegg @pascal-roth Should we review this again and then prepare it for merge?

Yes :+1: 

There is one remaining comment from @ooctipus, which ideally would be resolved in this pr https://github.com/isaac-sim/IsaacLab/pull/3371

pascal-roth:
@renezurbruegg we should probably do a quick profiling with tools like Nsight Systems or with Warp's ScopedTimer (https://nvidia.github.io/warp/profiling.html) to know our bottleneck for large resolutions, then we can also think about improving our kernel (maybe also for a next version)

ooctipus:
line 166 in `source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera.py`
```
if not target_cfg.is_global:
```

But I don't see is_global defined in target_cfg?

it should be changed to 

```
target_cfg.track_mesh_transforms
```

am I right?

Mayankm96:
@greptileai

renezurbruegg:
I believe I addressed all relevant comments! 

FYI: While cleaning up some warp kernels, I found a bug in the warp operators for atomic_min https://github.com/NVIDIA/warp/issues/1058. I believe this bug is quite severe and will hopefully be addressed soon. For now I added a fixme to the warp kernel sections.

renezurbruegg:
Bad Bot ;) 
The change should be fine. Each raycaster needs to lead the reference in "_initialize_warp_meshes". The caching and referencing happens properly in L198

kellyguo11:
@renezurbruegg @pascal-roth @Mayankm96 - could we try to get another approval for this PR?

also looks like the doc building job is failing - https://github.com/isaac-sim/IsaacLab/actions/runs/19120760997/job/54640579518?pr=3298

XiaoYi-Wei:
# Simulation startup becomes extremely slow when using the RayCaster sensor with large numbers of environments

## Description
When running the RayCaster sensor branch in IsaacSim, the simulation initialization time increases drastically as the number of environments grows.

The program does not crash, but the simulation hangs for a long time at the message:
[INFO]: Starting the simulation. This may take a few seconds. Please wait...


---

## Steps to Reproduce
1. Launch the simulation using the RayCaster sensor branch.  
2. Set the number of environments to a large value (e.g., 1000, 2000, or more).  
3. Observe the time it takes to start the simulation.

---

## Observed Behavior
- The console shows normal terrain and scene creation messages, then remains at picture


- The program stays in this state for a long time without proceeding.  
- Startup time increases roughly linearly with the number of environments.

| Number of Environments | Approx. Startup Time |
|------------------------:|----------------------:|
| 1                      | Immediate             |
| 1000                   | ~1 minute             |
| 4000                   | Several minutes       |

---

## Expected Behavior
The simulation should start within a reasonable time regardless of the number of environments, consistent with other sensors or configurations.

---

## Environment
- **IsaacSim version:** [your version]  
- **RayCaster branch:** [branch name or commit hash]  
- **GPU / OS:** [your specs]  
- **Number of environments tested:** up to 4000

fe5b1e9eb50aa8db7d162d1476d9356e

XiaoYi-Wei:
````markdown
### Issue Description

I found a potential bug in the file:

`source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_camera.py`  
at **line 178**.

The current code is:

```python
self.ray_hits_w, ray_depth, ray_normal, _, ray_mesh_ids = raycast_dynamic_meshes()
````

However, the first returned value of `raycast_dynamic_meshes()` has the shape:

```
[len(env_ids), num_rays, 3]
```

This causes a **CUDA kernel out-of-bounds access**, because `self.ray_hits_w` expects indexing over all environments, while the returned tensor only corresponds to `len(env_ids)`.
The issue doesn't throw an error immediately because `self.ray_hits_w` is not used later in this path, but the kernel still performs invalid memory access.

### Suggested Fix

It seems the assignment should correctly write only to the selected environments:

```python
self.ray_hits_w[env_ids], ray_depth, ray_normal, _, ray_mesh_ids = raycast_dynamic_meshes()
```

This aligns the shapes and prevents hidden OOB kernel access.

```
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,424 @@
...
+
+class MultiMeshRayCaster(RayCaster):
+    """A multi-mesh ray-casting sensor.
+
+    The ray-caster uses a set of rays to detect collisions with meshes in the scene. The rays are
+    defined in the sensor's local coordinate frame. The sensor can be configured to ray-cast against
+    a set of meshes with a given ray pattern.
+
+    The meshes are parsed from the list of primitive paths provided in the configuration. These are then
+    converted to warp meshes and stored in the :attr:`warp_meshes` list. The ray-caster then ray-casts against
```
pascal-roth:
```suggestion
    converted to warp meshes and stored in the :attr:`meshes` list. The ray-caster then ray-casts against
```

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster_cfg.py
pascal-roth: I agree, lets make that afterwards work

pascal-roth:
@XiaoYi-Wei thanks for the catch, should be fixed now

pascal-roth:
need to merge #3924 first

kellyguo11:
@pascal-roth @renezurbruegg could someone help take a look at the doc build failure?

Mayankm96:
@kellyguo11 I made an MR now to address the comments remaining here. I think Pascal and Rene are OoO now.

PR #3302, opened 2025-08-28
Improves recorder performance and add additional recording capability

peterd-NV:
# Description

This PR adds fixes from LightWheel Labs and additional functionality to the IsaacLab recorder.

Fixes # (issue)

- Fixes performance issue when recording long episode data by replacing the use of torch.cat at every timestep with list append.
- Fixes configclass validation when key is not a string

Adds Functionality

- Adds optional episode meta data to HDF5 recorder
- Adds option to record data pre-physics step
- Adds joint target data to episode data. Joint target data can be optionally recorded by users and replayed to bypass action term controllers and improve replay determinism.

## Type of change

## Screenshots

## Checklist


source/isaaclab/isaaclab/envs/manager_based_rl_env.py
```diff
@@ -188,6 +188,7 @@ def step(self, action: torch.Tensor) -> VecEnvStepReturn:
             self.scene.write_data_to_sim()
             # simulate
             self.sim.step(render=False)
+            self.recorder_manager.record_pre_physics_step()
```
Mayankm96: Isn't this happening now after the physics step?

peterd-NV: The recording right now records the states after a physics step. Here, it's adding functionality to record the states prior to every physics step in the decimation loop which provides more fine grain recording.

peterd-NV: The motivation here is that full determinism during replay is not possible when running parallel envs as the physX buffers cannot be flushed on every IsaacLab reset. This can cause significant divergence when replaying data collected with parallel envs by just using env.step as it has to run through the entire action space which may contain controllers that are sensitive to this nondeterminism. Recording at the decimation level every sim dt can help alleviate this by giving users the option of more fine grained control.

peterd-NV: Note that recording pre physics steps is fully optional and will not enabled by our existing default recorder workflows to limit dataset size. Users can take advantage of it via their own recorder terms if needed (like how Lightwheel is currently doing).

ooctipus: but should this be called record_post_physics_step() or  record_post_physics_decimated_step because it is after ?

peterd-NV: Sorry I misunderstood Mayank's original comment. Yes I agree the naming here is confusing. I have updated to be called "record_post_physics_decimation_step."

source/isaaclab/isaaclab/managers/recorder_manager.py
```diff
@@ -123,6 +123,15 @@ def record_post_step(self) -> tuple[str | None, torch.Tensor | dict | None]:
         """
         return None, None
 
+    def record_post_physics_decimation_step(self) -> tuple[str | None, torch.Tensor | dict | None]:
+        """Record data before the physics step is executed.
```
kellyguo11:
```suggestion
        """Record data after the physics step is executed in the decimation loop.
```

source/isaaclab/isaaclab/managers/recorder_manager.py
```diff
@@ -362,6 +371,16 @@ def record_post_step(self) -> None:
             key, value = term.record_post_step()
             self.add_to_episodes(key, value)
 
+    def record_post_physics_decimation_step(self) -> None:
+        """Trigger recorder terms for pre-physics step functions."""
```
kellyguo11:
```suggestion
        """Trigger recorder terms for post-physics step functions in the decimation loop."""
```

PR #3317, opened 2025-09-01
[Newton] Fix UTs

AntoineRichard:
# Description

Fix some of our UTs

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/cloner/utils.py
+++ b/source/isaaclab/isaaclab/cloner/utils.py
@@ -8,7 +8,6 @@ from typing import Any
 
 import warp as wp
 from newton import AxisType, ModelBuilder
-from newton.utils import parse_usd
 
 from isaaclab.utils.timer import Timer
 
@@ -47,10 +46,8 @@ def replicate_environment(
     with Timer(name="newton_env_builder", msg="Env Builder took:", enable=True, format="ms"):
         builder = ModelBuilder(up_axis=up_axis)
 
-        # first, load everything except the prototype env
-        stage_info = parse_usd(
+        stage_info = builder.add_usd(
             source,
-            builder,
             ignore_paths=[prototype_path],
             **usd_kwargs,
         )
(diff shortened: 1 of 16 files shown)


source/isaaclab/isaaclab/cloner/utils.py
```diff
@@ -80,22 +76,21 @@ def replicate_environment(
             joint_start = builder.joint_count
             articulation_start = builder.articulation_count
 
-            with Timer(name="newton_add_builder", msg="Add builder took:", enable=False, format="ms"):
```
adenzler-nvidia, line 83: is the timer removal intentional?

source/isaaclab/test/assets/test_articulation.py
```diff
@@ -826,7 +854,7 @@ def test_external_force_on_multiple_bodies(sim, num_articulations, device, add_g
         # check condition
         for i in range(num_articulations):
             # since there is a moment applied on the articulation, the articulation should rotate
-            assert articulation.data.root_ang_vel_w[i, 2].item() > 0.1
+            assert articulation.data.root_ang_vel_w[i, 2].item() < -1.0
```
adenzler-nvidia, line 857: can't judge that change from far - did you double check?

source/isaaclab/test/sim/test_simulation_context.py
```diff
@@ -135,8 +135,7 @@ def test_zero_gravity():
     """Test that gravity can be properly disabled."""
     cfg = SimulationCfg(gravity=(0.0, 0.0, 0.0))
 
-    sim = SimulationContext(cfg)
+    SimulationContext(cfg)
```
adenzler-nvidia, line 138: does that line have side-effects I don't know about or is it unnecessary now?

source/isaaclab/isaaclab/cloner/utils.py
AntoineRichard: Yes, I need to rework all of this with a new timing method.

source/isaaclab/test/sim/test_simulation_context.py
AntoineRichard: That's not used in the function anymore, but it's needed to see if the gravity value is properly given to our Newton manager.

source/isaaclab/test/assets/test_articulation.py
AntoineRichard: Should be OK? I'm not sure. It just happens to rotate which is what we want, but the direction is odd.

AntoineRichard:
> Looks good in general and works. Some stuff I can't judge added some comments - I could approve, question is how final all of these changes need to be.
> 
> I'd rather iterate here than having to do another full review when merging to the main dev branch, but I'll let you guys judge that.

Let's merge it into dev then. I don't think its critical ATM

PR #3328, opened 2025-09-02
Update technical_env_design.rst FIX jetbot path

jfelixre:
fix jetbot path


Mayankm96:
Thank you for the fix! It would be great to keep the PR template for us to track things better.

PR #3398, opened 2025-09-09
Fixes the import issues in stacking manipulation task

rebeccazhang0707:
# Description

Fixes the import issues via "path:Module" in below files, which will not import all modules/dependencies unless in use
```
- isaaclab_tasks/manager_based/manipulation/stack/config/galbot/__init__.py
- isaaclab_tasks/manager_based/manipulation/stack/config/ur10_gripper/__init__.py
- isaaclab_mimic/envs/__init__.py
```

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_mimic/isaaclab_mimic/envs/__init__.py
+++ b/source/isaaclab_mimic/isaaclab_mimic/envs/__init__.py
@@ -7,11 +7,13 @@
 
 import gymnasium as gym
 
+from .franka_bin_stack_ik_rel_mimic_env_cfg import FrankaBinStackIKRelMimicEnvCfg
 from .franka_stack_ik_abs_mimic_env import FrankaCubeStackIKAbsMimicEnv
 from .franka_stack_ik_abs_mimic_env_cfg import FrankaCubeStackIKAbsMimicEnvCfg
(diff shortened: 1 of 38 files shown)


source/isaaclab_mimic/isaaclab_mimic/envs/__init__.py
```diff
@@ -16,13 +16,6 @@
 from .franka_stack_ik_rel_skillgen_env_cfg import FrankaCubeStackIKRelSkillgenEnvCfg
 from .franka_stack_ik_rel_visuomotor_cosmos_mimic_env_cfg import FrankaCubeStackIKRelVisuomotorCosmosMimicEnvCfg
 from .franka_stack_ik_rel_visuomotor_mimic_env_cfg import FrankaCubeStackIKRelVisuomotorMimicEnvCfg
```
Mayankm96, line 18: Thank you for the fix! Can we also propagate the changes to these imports here? We shouldn't be importing the env cfgs since they increase the module load time.

ooctipus: We can also @peterd-NV to have mimic team submit a PR to fix them as well

PR #3399, opened 2025-09-09
Adds PBT algorithm to rl games

ooctipus:
# Description

This PR introduces the Population Based Training algorithm originally implemented in

Petrenko, Aleksei, et al. "Dexpbt: Scaling up dexterous manipulation for hand-arm systems with population based training." arXiv preprint arXiv:2305.12127 (2023).

Pbt algorithm offers a alternative to scaling when increasing number of environment has margin effect.
It takes idea in natural selection and stochastic property in rl-training to always keeps the top performing agent while replace weak agent with top performance to overcome the catastrophic failure, and improve the exploration.

Training view, underperformers are rescued by best performers and later surpasses them and become best performers
Screenshot from 2025-09-09 00-55-11

Note:
PBT is still at beta phase and has below limitations:

1. in theory It can work with any rl algorithm but current implementation only works for rl-games
2. The API could be furthur simplified without needing explicitly input num_policies or policy_idx, which allows for dynamic max_population, but it is for future work

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_rl/isaaclab_rl/rl_games/pbt/pbt.py
+++ b/source/isaaclab_rl/isaaclab_rl/rl_games/pbt/pbt.py
@@ -22,6 +22,7 @@ _UNINITIALIZED_VALUE = float(-1e9)
 
 class PbtAlgoObserver(AlgoObserver):
     """rl_games observer that implements Population-Based Training for a single policy process."""
+
     def __init__(self, params, args_cli):
         """Initialize observer, print the mutation table, and allocate the restart flag.
 


Mayankm96:
Thank you for this feature! Would be good to add the images that we are working on for PBT. 

Is PBT just an add-on to the library? Are the assumptions in there specific to RL-Games or we can make it generic enough to use with RSL-RL too?

ooctipus:
Yes PBT should be an addon to the library, currently it is specific design to rl-games by attaching to rl-games module. It is possible to make it generic, but will take more time and work.

kellyguo11 (approved):

PR #3454, opened 2025-09-15
Adds single-shot events

renezurbruegg:
# Description

This PR introduces one-shot events.

A one-shot event is triggered at most once before the environment is reset. This makes it possible to modify the environment at a specific time, for example by lowering the friction after one second or by applying forces at different moments such as 1s, 2s, and 3s.

## Example

The following example schedules different forces to be applied at one-second intervals without being repeated. Without the is_single_shot option, the pull_up event would instead be triggered repeatedly at 1s, 2s, and 3s.

```python

@configclass
class SingleShotEventExample:
    pull_up = EventTerm(
        func=mdp.pull_object,
        mode="interval",
        is_global_time=True,
        interval_range_s=(1.0, 1.0),
        is_single_shot=True,
        params={"direction": (0.0, 0.0, 1.0)},
    )
    
    pull_down = EventTerm(
        func=mdp.pull_object,
        mode="interval",
        is_global_time=True,
        interval_range_s=(2.0, 2.0),
        is_single_shot=True,
        params={"direction": (0.0, 0.0, -1.0)},
    )
    
    pull_left = EventTerm(
        func=mdp.pull_object,
        mode="interval",
        is_global_time=True,
        interval_range_s=(3.0, 3.0),
        is_single_shot=True,
        params={"direction": (1.0, 0.0, 0.0)},
    )
    reset_in_betweeen = EventTerm(
        func=mdp.reset,
        mode="interval",
        is_global_time=True,
        interval_range_s=(1.0, 1.0),
    )
```

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+0.46.2 (2025-09-15)
+~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^^^
+
+* Added argument :attr:`is_single_shot` to :class:`~isaaclab.managers.EventTermCfg` to control whether the event is only applied once.
+
+
 0.46.1 (2025-09-10)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab/isaaclab/managers/event_manager.py
+++ b/source/isaaclab/isaaclab/managers/event_manager.py
@@ -212,8 +212,11 @@ class EventManager(ManagerBase):
                 # note: we compare with a small value to handle floating point errors
                 if term_cfg.is_global_time:
                     if time_left < 1e-6:
-                        lower, upper = term_cfg.interval_range_s
-                        sampled_interval = torch.rand(1) * (upper - lower) + lower
+                        if not term_cfg.is_single_shot:
+                            lower, upper = term_cfg.interval_range_s
+                            sampled_interval = torch.rand(1) * (upper - lower) + lower
+                        else:
+                            sampled_interval = float('inf')
                         self._interval_term_time_left[index][:] = sampled_interval
 
                         # call the event term (with None for env_ids)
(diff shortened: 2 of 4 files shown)


jtigue-bdai (approved):
lgtm

Mayankm96:
Would there be a case where someone says they want N-shot? 🤔

renezurbruegg:
Maybe yes, personally I think it's unlikely. In this case, one could just put N single shot events in the config.

Thinking about it, being able to specify initial delay and interval on the other hand, might be interesting and would be able to also mimic the single shot behavior (infinite interval time with finite inital delay). But I fear this would introduce unwanted breaking changes.

ooctipus:
I am wonder if feature like this is better to be written as custom function to combined with `interval`, rather than allocate a new attribute in EventsManager.

renezurbruegg:
> I am wonder if feature like this is better to be written as custom function to combined with `interval`, rather than allocate a new attribute in EventsManager.

I think there are enough practical use cases to justify a dedicated flag. Requiring a custom functor would force the user to move away from the functional interface and implement a separate class with state, which adds unnecessary complexity for something that could be handled more directly.

ooctipus:
ok between combining `interval` + `is_single_shot` as an extra attribute
vs
maybe giving a new mode called `one_shot_mode`, wondering your opinion, which one do you thinks make more sense?

renezurbruegg:
To me the current implementation is more efficient coding wise whereas `one_shot_mode` would avoid another "this argument is only used if mode is set to `reset|interval`" argument, which I think is nice to avoid. 

Thinking more about it, imo the cleanest solution would be to introduce EventResetTermCfg, EventIntervalTermCfg, EventSingleShotTermCfg, ... and move all "interval|reset only" arguments to these configs. But this would introduce quite some breaking changes. 

What's your opinion? I think removing the argument and having it in the mode as you said is cleaner.

ooctipus:
```
Thinking more about it, imo the cleanest solution would be to introduce EventResetTermCfg, EventIntervalTermCfg, EventSingleShotTermCfg, ... and move all "interval|reset only" arguments to these configs. But this would introduce quite some breaking changes.
```

if this is the future way to go, then having one shot mode will prepare us to that goal and somewhat aligns with that separation, I'd say removing the argument and having it in the mode. 

But no rush at all : ))
Thanks for considering my opinion!

@Mayankm96  would also be great if you can give some input here as well

AntoineRichard:
Hi @renezurbruegg — thank you so much for opening this PR, and for taking the time to contribute to Isaac Lab! 🙏

| | |
|---|---|
| **Opened** | 2025-09-15 (about 11 months ago) |
| **Last commit on the branch** | 2025-09-15 |
| **Last activity from the author** | about 11 months ago |
| **Review status** | Approved at the time (jtigue-bdai), but it was never merged and has since gone stale |
| **Size** | 3 commit(s), 4 file(s) changed, +124 / -4 |

It was picked up by the sweep because it targets `main` instead of `develop`, and it has been open for about 11 months. It was then put in the "close" bucket because the author has been silent for about 11 months — which is the signal we used to tell apart pull requests that are still being worked on from ones that have genuinely been set aside.

---

PR #3464, opened 2025-09-15
Fixes torch nightly version install in arm system

ooctipus:
# Description

This PR improves the robustness in detecting arm system and correctly install torch nightly build for cuda130

Fixes # (issue)

## Screenshots

## Checklist

### DIFF
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -102,37 +102,34 @@ is_arm() {
 }
 
 ensure_cuda_torch() {
-  local py="$1"
-  local -r TORCH_VER="2.7.0"
-  local -r TV_VER="0.22.0"
-  local -r CUDA_TAG="cu128"
-  local -r PYTORCH_INDEX="https://download.pytorch.org/whl/${CUDA_TAG}"
-  local torch_ver
+    local py="$1"
 
-  if "$py" -m pip show torch >/dev/null 2>&1; then
-    torch_ver="$("$py" -m pip show torch 2>/dev/null | awk -F': ' '/^Version/{print $2}')"
-    echo "[INFO] Found PyTorch version ${torch_ver}."
-    if [[ "$torch_ver" != "${TORCH_VER}+${CUDA_TAG}" ]]; then
-      echo "[INFO] Replacing PyTorch ${torch_ver} → ${TORCH_VER}+${CUDA_TAG}..."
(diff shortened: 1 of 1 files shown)


isaaclab.sh
```diff
@@ -102,37 +102,34 @@ is_arm() {
...
-      else
-        "$py" -m pip install "torch==${TORCH_VER}" "torchvision==${TV_VER}" --index-url "${PYTORCH_INDEX}"
-      fi
-    else
-      echo "[INFO] PyTorch ${TORCH_VER}+${CUDA_TAG} already installed."
+    local py="$1"
+
+    # read installed torch version (e.g., "2.7.0+cu128"), empty if missing
+    local v=""
+    if "$py" -m pip show torch >/dev/null 2>&1; then
```
kellyguo11, line 109: do we need to accommodate for the case where no torch is installed?

isaaclab.sh
```diff
@@ -102,37 +102,34 @@ is_arm() {
 }
 
 ensure_cuda_torch() {
```
Mayankm96, line 104: General question: Is there a reason why we are enforcing torch CUDA in our install scripts and not via the `setup.py`?

ooctipus: Ahh good point

IIRC setup.py's wheel sometime gets a bit unrealible, and had to do more explicit install in bash. 

If above point is not true anymore, I think it might be an historical reason?
I was checking the commit history of this file and explicit installation of pytorch is introduced here, 01732521a2323ce9cdac8e0c97e737ef29ddcee4, which seems the intend is to fix the pytorch version docker

The right path forward seems like that we need to ensure we have arm machine running on ci in addition to linux. and whatever changes we make here needs to work on those machine as well. So far I mostly have just been mannually checking if it works in ARM

isaaclab.sh
ooctipus: it works fine when Torch isn’t installed. basically no printing of `Found PyTorch verison` and will execute fresh install in next lines

isaaclab.sh
kellyguo11: ya the setup.py cuda build of torch is very unreliable on windows especially and it will always install the cpu build instead and leads to more errors when we run our workflows :( not sure if uv can do a better job at that though, that might be something we can look into as well.

Issue #3465, opened 2025-09-15
[Proposal] Delayed Observation Term


mikegroom765:
### Proposal

Add DelayedObsTerm (a ManagerTermBase term) that wraps any observation term and returns stale versions using the IsaacLab DelayBuffer class.

- Preserves shape.
- Samples lag per-env in [min_lag, max_lag].
- Causal clamp (lag_t ≤ lag_{t-1}+1) and hold_prob to mimic frame holds.
- Optionally models multi-rate signals via update_period (with optional per-env phase)
- Params nested under "_" (manager requirement).

### Motivation

Real hardware has multi-rate sensors and variable latency. Currently only sensor based observations support a specified update rate, and variable latency is not supported. This term can be used to specify an update rate on non-sensor based observations, and model variable latency (for improved sim2real).  

A feature like this has been discussed a few times, but not yet added as far as I know. #431 #2575

### Alternatives

None considered.

### Additional context

An example config (using, for example, a height scan observation) for testing:

```
from isaaclab.managers.terms.delayed_obs_term import DelayedObsTerm

height_scan = ObservationTermCfg(
    func=DelayedObsTerm,
    params={"_": {
        "func": mdp.height_scan,
        "func_params": {"sensor_cfg": SceneEntityCfg("height_scanner")},
        "min_lag": 1, "max_lag": 6,
        "per_env": True, "hold_prob": 0.66,
        # optional:
        # "update_period": 3, "per_env_phase": True,
    }},
)
```

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

### Acceptance Criteria

Add the criteria for which this task is considered **done**. If not known at issue creation time, you can add this once the issue is assigned.

- [ ] Criteria 1
- [ ] Criteria 2

Mayankm96:
We can make a modifier that handles this. We don't need to add yet another attribute to the observation term class for it :)

mikegroom765:
Thanks for the quick feedback @Mayankm96! :) 

To clarify, my current version is an observation decorator (a `ManagerTermBase` that takes in another terms tensor and returns a delayed version via the `DelayBuffer` class), and doesn't add any new attributes to existing term classes. 

I agree a modifier is a cleaner fit, I'll rework this as a modifier and open a PR once it is ready! Thanks again!

PR #3471, opened 2025-09-16
Delayed observation modifier

mikegroom765:
# Description

Add `DelayedObservation` modifier for stochastic observation latency and multi-rate observation modelling. Closes issue #3465. The `DelayedObservation` modifier is a shape preserving observation modifier that returns a stale version of the input observation using the `DelayBuffer` class. It models:

- Variable latency on observations:  per-env (optional) integer lags sampled in `[min_lag, max_lag]` at each policy step, returning an observation from $$lag_{t}$$ steps in the past.
- Causal progression: $$lag_{t} <= lag_{t+1} + 1$$
- (Optional) Frame holds: either specified through `update_period` (deterministic) to simulate slow sensor updates (e.g., `update_period=2` means the observation is held constant, and refreshed every 2 policy steps), or stochastically with `hold_prob`. With probability `hold_prob`, the modifier will reuse the previous lag so consecutive steps return the same (older) frame, simulating jitter
- Supports applying both `update_period` and `hold_prob` together, to simulate jitter on low frequency observation updates

## Type of change

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -96,6 +96,7 @@ Guidelines for modifications:
 * Manuel Schweiger
 * Masoud Moghani
 * Maurice Rahme
+* Michael Groom
 * Michael Gussert
 * Michael Noseworthy
 * Michael Lin
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+Unreleased
+~~~~~~~~~~
+
+Added
+^^^^^
+* Added :class:`isaaclab.utils.modifiers.DelayedObservation` for stochastic latency and multi-rate observation modeling (per-env lags, ``hold_prob``, ``update_period``, ``per_env_phase``).
+* Added :class:`isaaclab.utils.modifiers.DelayedObservationCfg` configuration class.
+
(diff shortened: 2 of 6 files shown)


kellyguo11:
@mikegroom765 Thanks for the addition! apologies for the delay in getting this merged. Would you be able to rebase these changes onto the latest `develop` branch? Thanks!

krivonogov0:
First of all, thank you for this excellent addition, as the `DelayedObservation` modifier is exactly what is needed for robust Sim-to-Real transfer!
However, I encountered a limitation when using it in a **Curriculum Learning** setup where latency is gradually increased. Currently, the internal `DelayBuffer` is initialized with a fixed size based on the initial `cfg.max_lag`. This means if you start training with `max_lag=0` and try to increase it later via `modify_term_cfg`, the code fails because the buffer capacity remains too small to accommodate the new lag values.

This is problematic because stable training often requires starting with perfect observations and progressively introducing latency as the policy matures. The current implementation forces users to initialize delay with the maximum expected lag from step 0, which can destabilize early learning phases.

I suggest adding a check in the `__call__` method to detect if `cfg.max_lag` has increased since initialization. If so, the `DelayBuffer` should be dynamically resized to support the new range.

ooctipus:
https://github.com/isaac-sim/IsaacLab/pull/7976
please check if this meets what you need!

PR #3502, opened 2025-09-19
Exposes `physxscene:solveArticulationContactLast` flag through PhysxCfg

ooctipus:
# Description

This PR adds api to set physxscene:solveArticulationContactLast through PhysxCfg,
this is only available in sim 5.1

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -5,8 +5,8 @@ Changelog
 ~~~~~~~~~~~~~~~~~~~
 
 * Exposed parameter :attr:`~isaaclab.sim.spawners.PhysxCfg.solve_articulation_contact_last`
-   to configure USD attribute ``physxscene:solveArticulationContactLast``. This parameter may
-   help improve solver stability with grippers, which previously required reducing simulation time-steps.
+  to configure USD attribute ``physxscene:solveArticulationContactLast``. This parameter may
+  help improve solver stability with grippers, which previously required reducing simulation time-steps.
   :class:`~isaaclab.sim.spawners.PhysxCfg`
 
 


source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -160,6 +160,9 @@ class PhysxCfg:
     gpu_max_particle_contacts: int = 2**20
     """Size of particle contacts stream buffer allocated in pinned host memory. Default is 2 ** 20."""
 
+    solve_articulation_contact_last: bool = False
+    """This reorders the solver and could be a potential aid for gripper scenarios, Default is False"""
```
Mayankm96: It's good to give a thorough documentation so users can read more about the parameter from the PhysX docs. Otherwise, it's based on their luck if they find the param.

Also could you move this parameter to above in the config where we have different solver level configuration. The `gpu_` params are only for buffers.

```suggestion
    """Changes the ordering inside the articulation solver. Default is False.
    
    PhysX employs a strict ordering for handling constraints in an articulation. The outcome of 
    each constraint resolution modifies the joint and associated link speeds. However, the default
    ordering may not be ideal for gripping scenarios because the solver favours the contraint
    types that are resolved last. This is particularly true of stiff constraint systems that are hard 
    to resolve without resorting to vanishingly small simulation timesteps.
    
    With dynamic contact resolution being such an important part of gripping, it may make
    more sense to solve dynamic contact towards the end of the solver rather than at the
    beginning. This parameter modifies the default ordering to enable this change.
    
    For more information, please check `here <https://docs.omniverse.nvidia.com/kit/docs/omni_physics/107.3/dev_guide/guides/articulation_stability_guide.html#articulation-solver-order>`__.
    
    .. versionadded:: v2.3
        This parameter is only available with Isaac Sim 5.1.

    """
```

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+0.46.3 (2025-09-17)
+~~~~~~~~~~~~~~~~~~~
+
+Changed
+^^^^^^^
+
+* Exposed physxscene: USD attribute "physxscene:solveArticulationContactLast" to isaaclab
```
Mayankm96:
```suggestion
* Exposed parameter :attr:`~isaaclab.sim.spawners.PhysxCfg.solve_articulation_contact_last` 
   to configure USD attribute ``physxscene:solveArticulationContactLast``. This parameter may
   help improve solver stability with grippers, which previously required reducing simulation time-steps.
```

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -758,6 +758,10 @@ def _set_additional_physx_params(self):
         physx_scene_api.CreateGpuCollisionStackSizeAttr(self.cfg.physx.gpu_collision_stack_size)
         # -- Improved determinism by PhysX
         physx_scene_api.CreateEnableEnhancedDeterminismAttr(self.cfg.physx.enable_enhanced_determinism)
+        # -- Set solve_articulation_contact_last by add attribute to the PhysxScene prim, and add attribute there.
+        physx_prim = physx_scene_api.GetPrim()
+        if self.cfg.physx.solve_articulation_contact_last:
```
Mayankm96: what is parameter is False? Shouldn't that be set as well?

Mayankm96:
Do we have an example showing that this feature helps?

source/isaaclab/isaaclab/sim/simulation_context.py
ooctipus: good point, the small issue with set to false is that we break the backward compatibility, I think such flag does not exist pre isaac-sim 5.1

ooctipus:
this flag was troduced when mimic gripper had the issues iirc, @peterd-NV , do you think we can engage with you to document the incidence of gripper behavior and how adding this flag can make gripper behave well, to illustrate the impact of this flag?

source/isaaclab/isaaclab/sim/simulation_context.py
kellyguo11: let's test to see if it will break 5.0, we may want to add in a check if that's the case

PR #3512, opened 2025-09-20
Adds torch hook to export libgomp.so.1 from installed torch

ooctipus:
# Description

This PR makes the LD_PRELOAD always points to the right python libgomp.so.1 when conda activate is invoked.

The ARM machines doesn't really work with conda unless LD_PRELOAD is point to `torch/lib/libgomp.so.1`

This has been tested to work on linux and arm to work on both machines

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -247,6 +247,45 @@ install_isaaclab_extension() {
     fi
 }
 
+# Resolve Torch-bundled libgomp and prepend to LD_PRELOAD, once per shell session
+write_torch_gomp_hooks() {
+  mkdir -p "${CONDA_PREFIX}/etc/conda/activate.d" "${CONDA_PREFIX}/etc/conda/deactivate.d"
+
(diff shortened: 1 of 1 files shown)


isaaclab.sh
```diff
@@ -247,6 +247,45 @@ install_isaaclab_extension() {
     fi
 }
 
+# Resolve Torch-bundled libgomp and prepend to LD_PRELOAD, once per shell session
+write_torch_gomp_hooks() {
+  mkdir -p "${CONDA_PREFIX}/etc/conda/activate.d" "${CONDA_PREFIX}/etc/conda/deactivate.d"
```
kellyguo11, line 252: is this required? looks like it's already done earlier in the script

ooctipus: this is like a exist_ok check,
so it doesn't assume such file always exist, but if exist, do nothing.

this line makes function a bit more self contained and readable is my judge.

PR #3522, opened 2025-09-22
[Bugfix] use effort_limit from USD asset if not specified in cfg

JuanaDd:
# Description

This PR fixes a bug in actuator initialization where effort limits specified in USD assets were being incorrectly overridden with a very large default value (1.0e9) for explicit actuator models.

Fixes # (issue)

Previously, the ActuatorBase initialization logic would unconditionally fall back to _DEFAULT_MAX_EFFORT_SIM (1.0e9) for explicit actuator models when effort_limit_sim was not explicitly set in the configuration, even when the USD asset contained finite, meaningful effort limit values.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/actuators/actuator_base.py
+++ b/source/isaaclab/isaaclab/actuators/actuator_base.py
@@ -159,7 +159,10 @@ class ActuatorBase(ABC):
         self.joint_property_resolution_table: dict[str, list] = {}
         # For explicit models, we do not want to enforce the effort limit through the solver
         # (unless it is explicitly set)
-        if not self.is_implicit_model and self.cfg.effort_limit_sim is None:
+        effort_limit_is_finite = (
+            effort_limit != torch.inf if isinstance(effort_limit, float) else torch.all(effort_limit != torch.inf)
+        )
+        if not self.is_implicit_model and self.cfg.effort_limit_sim is None and (not effort_limit_is_finite):
             self.cfg.effort_limit_sim = self._DEFAULT_MAX_EFFORT_SIM
 
         # resolve usd, actuator configuration values


source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -159,7 +159,10 @@ def __init__(
         self.joint_property_resolution_table: dict[str, list] = {}
         # For explicit models, we do not want to enforce the effort limit through the solver
         # (unless it is explicitly set)
-        if not self.is_implicit_model and self.cfg.effort_limit_sim is None:
+        effort_limit_is_finite = (
```
Mayankm96: Modifying the code where we set effort_limit might be simpler i think

```python
        # For explicit models, if user does not provide effort limit, we want to default to USD values.
        # For implicit models, we want to enforce the effort limit through the solver.
        if not self.is_implicit_model and self.cfg.effort_limit_sim is None:
            self.effort_limit = self._parse_joint_parameter(self.cfg.effort_limit, effort_limit)
        else:
            self.effort_limit = self._parse_joint_parameter(self.cfg.effort_limit, self.effort_limit_sim)

```

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -1681,7 +1681,7 @@ def _process_actuators_cfg(self):
                 friction=self._data.default_joint_friction_coeff[:, joint_ids],
                 dynamic_friction=self._data.default_joint_dynamic_friction_coeff[:, joint_ids],
                 viscous_friction=self._data.default_joint_viscous_friction_coeff[:, joint_ids],
-                effort_limit=self._data.joint_effort_limits[:, joint_ids],
+                effort_limit=self._data.joint_effort_limits[:, joint_ids].clone(),
```
JuanaDd, line 1716: @Mayankm96   clone is required because effort_limit would be overwritten to effort_limit_sim again [here](https://github.com/isaac-sim/IsaacLab/blob/3a0db9d761982dc65417e6c6d0714cec61ceadb3/source/isaaclab/isaaclab/assets/articulation/articulation.py#L792)

source/isaaclab/isaaclab/actuators/actuator_base.py
ooctipus: I think this is a bit confusing, I reasoned about several time to understand why I was confused XD


I think this is due to the fact that `effort_limit` from USD is unclear if author who wrote the value intended to be motor maximum or simulation solver maximum:

- I guess @JuanaDd is thinking that if   `effort_limit` from USD has a solver limit maximum value, then we shouldn't just override it with   _DEFAULT_MAX_EFFORT_SIM (1.0e9), disrespecting USD's solver maximum value.

- @Mayankm96 in your reply, I guess you are thinking `effort_limit` from USD may be a motor limit and thus
`# For explicit models, if user does not provide effort limit, we want to default to USD values.`  I think the else section (`implicit part`) is not needed because the later code will do exactly else logic.

I think it is more safe to assume `effort_limit` this is simulation solver maximum since it will end up behaving like that.

ooctipus:
I will propose my solution to this @Mayankm96 @JuanaDd  let me know what you think.
1. Regarding `effort_limit` from USD as effort_limit_sim, not motor level limit,
2. Deleting below block so no more assumption about -> explicit actuator should have _DEFAULT_MAX_EFFORT_SIM if ActuatorCfg.effort_limit_sim not specified. We want to use USD default instead for that case.
```
        # For explicit models, we do not want to enforce the effort limit through the solver
        # (unless it is explicitly set)
        if not self.is_implicit_model and self.cfg.effort_limit_sim is None:
            self.cfg.effort_limit_sim = self._DEFAULT_MAX_EFFORT_SIM
```
3. for all places we want _DEFAULT_MAX_EFFORT_SIM as `effort_limit_sim` we mannually define them in ActuatorCfg

This we have the greatest clarity from ActuatorCfg and also minimum code path(carity) for USD-Cfg resolution

source/isaaclab/isaaclab/actuators/actuator_base.py
JuanaDd: Hi @ooctipus  thanks for your comment. My understanding is that, `effort_limit` is used for internal clipping in Isaaclab's explicit actuators but `effort_limit_sim` is passed to PhysX as constraints on joint efforts. 

The isssue is that the USD defaults are not set to the effort_limit when using explicit actuators. Instead it is set to 1e9 in this [line](https://github.com/isaac-sim/IsaacLab/blob/3a0db9d761982dc65417e6c6d0714cec61ceadb3/source/isaaclab/isaaclab/actuators/actuator_base.py#L199)

ooctipus: Let me clarify two different usage of `effort_limit` here

-1.`ActuatorCfg.effort_limit` is used for internal clipping in Isaaclab's explicit actuators

-2. `effort_limit` below is actually USD's default, (we should probably rename it XD)
```
    def __init__(
        self,
        cfg: ActuatorBaseCfg,
        joint_names: list[str],
        joint_ids: slice | torch.Tensor,
        num_envs: int,
        device: str,
        stiffness: torch.Tensor | float = 0.0,
        damping: torch.Tensor | float = 0.0,
        armature: torch.Tensor | float = 0.0,
        friction: torch.Tensor | float = 0.0,
        dynamic_friction: torch.Tensor | float = 0.0,
        viscous_friction: torch.Tensor | float = 0.0,
        effort_limit: torch.Tensor | float = torch.inf,
        velocity_limit: torch.Tensor | float = torch.inf,
    ):
```
TLDR: Treat USD’s effort_limit as the solver/physics limit (effort_limit_sim) and never silently overwrite it. Motor-level clipping stays in ActuatorCfg.effort_limit. If someone wants “effectively unbounded,” they must set ActuatorCfg.effort_limit_sim = 1e9 (or similar) explicitly in config—not via hidden defaults.

Longer:
It is unclear USD's effort_limit is meant to be motor_level limit or Simulation solver limit, it could be either depending on who provided USD.

If the USD's effort_limit is Simulation solver limit, then overriding it to _DEFAULT_MAX_EFFORT_SIM (1.0e9) for no reason is wrong. This is also what you are trying to fix if I understand correctly. (Note you are comparing USD's `effort_limit` with ActuatorCfg's `effort_limit_sim` )

If the USD's effort_limit is motor_level limit, then then overriding it to _DEFAULT_MAX_EFFORT_SIM (1.0e9) actually make sense! which sounds like this is the reason why the overriding line was designed, and also why @Mayankm96 later proposed, `# For explicit models, if user does not provide effort limit, we want to default to USD values.` (Note here is comparing USD's `effort_limit` with ActuatorCfg's `effort_limit` )

What I am saying is that USD's effort_limit is AMBIGUOUS, but if we have to resolve on one definition,  I think treating USD's effort_limit is Simulation solver limit makes sense, because it is what eventually will behave like in simulation. That is to also say, we shouldn't do any smart auto-overriding anymore, if explicit actuator wants very very large effort_limit_sim, we should always write that value out in the ActuatorCfg(Refer my solution in later comment).

ooctipus: @jtigue-bdai also take a look at this discussion and let us know what you think :))

source/isaaclab/test/assets/test_articulation.py
```diff
@@ -1436,6 +1450,9 @@ def test_setting_effort_limit_explicit(sim, num_articulations, device, effort_li
     # Play sim
     sim.reset()
 
+    # usd default is 80
+    usd_default = 80.0
```
jtigue-bdai:
```suggestion
    # usd default effort limit is set to 80
    usd_default_effort_limit = 80.0
```

source/isaaclab/test/assets/test_articulation.py
```diff
@@ -1452,8 +1469,9 @@ def test_setting_effort_limit_explicit(sim, num_articulations, device, effort_li
         # check physx effort limit does not match the one explicit actuator has
         assert not (torch.allclose(actuator_effort_limit, physx_effort_limit))
     else:
-        # check actuator effort_limit is the same as the PhysX default
-        torch.testing.assert_close(actuator_effort_limit, physx_effort_limit)
+        # When effort_limit is None, actuator should use USD default values
+        expected_actuator_effort_limit = torch.full_like(physx_effort_limit, usd_default)
```
jtigue-bdai:
```suggestion
        expected_actuator_effort_limit = torch.full_like(physx_effort_limit, usd_default_effort_limit)
```

jtigue-bdai:
@JuanaDd Thanks for looking into this. I agree with the problem. The effort limit values of the USD articulation (and the USD created from urdf importers) should be passed through as effort_limit. 

I think it would be good to add more of a note to the documentation on effort_limit and effort_limit_sim explaining the pass through.

JuanaDd:
Hi @jtigue-bdai  Could you review my latest commit and check if the documentation is clear enough?

Mayankm96:
Could you pleaes update the changelog and extension.toml as well? Thanks!

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,12 +1,22 @@
...
+* Fixed the bug where effort limits were being overridden in :class:`~isaaclab.actuators.ActuatorBase` when the ``effort_limit`` parameter is set to None.
+* Corrected the unit tests for three effort limit scenarios with proper assertions
+
 0.47.7 (2025-10-31)
 ~~~~~~~~~~~~~~~~~~~
 
 Changed
 ^^^^^^^
 
+
```
```suggestion
```

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,12 +1,22 @@
...
 
+0.47.8 (2025-11-03)
+~~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed the bug where effort limits were being overridden in :class:`~isaaclab.actuators.ActuatorBase` when the ``effort_limit`` parameter is set to None.
+* Corrected the unit tests for three effort limit scenarios with proper assertions
+
```
kellyguo11, line 12:
```suggestion


```

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -1843,6 +1843,7 @@ def _apply_actuator_model(self):
         """
         # process actions per group
         for actuator in self.actuators.values():
+
```
kellyguo11:
```suggestion
```

kellyguo11:
source/isaaclab/test/actuators/test_ideal_pd_actuator.py  seems to be failing now, do we need to update any of the test cases there?

Issue #3559, opened 2025-09-25
[Proposal] Add optional --python arg when --uv is used to create venv


kwang-12:
### Proposal

Add support for specifying a custom Python interpreter when creating a uv virtual environment via isaaclab.sh.
This would extend the existing --uv option to accept an optional --python /path/to/python argument. The --help text would also be updated to reflect the new syntax and provide usage examples.

### Motivation

Currently isaaclab.sh always uses the default python3 in the user’s PATH when creating a uv environment.
On systems with multiple Python versions installed (e.g., Ubuntu 22.04 ships with Python 3.10 but Isaac Sim 5.X requires Python 3.11), the only way to force a specific interpreter is to manually edit the script or run uv venv directly.

Supporting a --python flag would let users do this directly from the script without manual patching.

### Alternatives

1. Manually run uv venv --python /path/to/python and then alias isaaclab.sh yourself (works but bypasses the helper script). or
2. Edit the isaaclab.sh script to hardcode your preferred Python path (not sustainable across updates).

Neither of these alternatives is user-friendly compared to a built-in --python flag.

### Additional context

Example usage after feature implementation:

```
        1.a create env_isaaclab with system python3
        isaaclab.sh --uv
        1.b create myenv with system python3
        isaaclab.sh --uv myenv               
        1.c create myenv with specified Python 3.11
        isaaclab.sh --uv myenv --python ${HOME}/isaacsim50/kit/python/bin/python3
```

### Checklist

- [x] I have checked that there is no similar issue in the repo (**required**)

### Acceptance Criteria

Add the criteria for which this task is considered **done**. If not known at issue creation time, you can add this once the issue is assigned.

- [ ] isaaclab.sh supports an optional --python /path/to/python argument following --uv to override the interpreter used by uv venv.
- [ ] print_help (or equivalent) reflects the new syntax and shows example usage.

Fixes: #3560

RandomOakForest:
Thank you for posting this. The team will review it.

AntoineRichard:
Thanks for this, @kwang-12 — pinning the interpreter is a reasonable ask, and we appreciate that you opened #3560 with an implementation.

[...]

Issue #3524, opened 2025-09-22
[Bug Report] Cannot install IsaacLab with uv


PabloVD:
I'm following the IsaacLab installation instructions building it from source using uv from [the documentation](https://isaac-sim.github.io/IsaacLab/main/source/setup/installation/source_installation.html), but it raises the following error. I'm using Ubuntu 22.

```
     Build targets in project: 100
      WARNING: Project specifies a minimum meson_version '>=1.2.99' but uses features which were added in newer versions:
       * 1.3.0: {'dep 'blas' custom lookup', 'dep 'lapack' custom lookup'}

      NumPy 1.26.4

        User defined options
          Native files: /home/tda/.cache/uv/sdists-v9/pypi/numpy/1.26.4/VbLTxVdsH1TK70hFaVj86/src/.mesonpy-1lf6nsl5/meson-python-native-file.ini
          buildtype   : release
          b_ndebug    : if-release
          b_vscrt     : md

      Found ninja-1.10.1 at /usr/bin/ninja

      [stderr]
      Traceback (most recent call last):
        File "/home/tda/.cache/uv/builds-v0/.tmpfqM4uA/lib/python3.13/site-packages/pyproject_metadata/__init__.py", line 515, in validate
          packaging.utils.canonicalize_name(self.name, validate=True)
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
      TypeError: canonicalize_name() got an unexpected keyword argument 'validate'

      During handling of the above exception, another exception occurred:

      Traceback (most recent call last):
        File "<string>", line 11, in <module>
          wheel_filename = backend.build_wheel("/home/tda/.cache/uv/builds-v0/.tmp26dQol", {}, None)
        File "/home/tda/.cache/uv/builds-v0/.tmpfqM4uA/lib/python3.13/site-packages/mesonpy/__init__.py", line 985, in wrapper
          return func(*args, **kwargs)
        File "/home/tda/.cache/uv/builds-v0/.tmpfqM4uA/lib/python3.13/site-packages/mesonpy/__init__.py", line 1038, in build_wheel
          with _project(config_settings) as project:
               ~~~~~~~~^^^^^^^^^^^^^^^^^
        File "/home/tda/.local/share/uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/contextlib.py", line 141, in __enter__
          return next(self.gen)
[...]

PR #3560, opened 2025-09-25
Adds optional --python arg when --uv is used to create venv

kwang-12:
# Description

This PR directly addresses #3559. 

## Type of change

## Checklist

### DIFF
--- a/docs/source/setup/installation/include/src_clone_isaaclab.rst
+++ b/docs/source/setup/installation/include/src_clone_isaaclab.rst
@@ -53,7 +53,7 @@ respectively that provides utilities to manage extensions.
             -d, --docs           Build the documentation from source using sphinx.
             -n, --new            Create a new external project or internal task from template.
             -c, --conda [NAME]   Create the conda environment for Isaac Lab. Default name is 'env_isaaclab'.
-            -u, --uv [NAME]      Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'.
+            -u, --uv [NAME]      Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'. Use delimiter -- to pass additional arg to uv.
 
    .. tab-item:: :icon:`fa-brands fa-windows` Windows
       :sync: windows
@@ -75,4 +75,3 @@ respectively that provides utilities to manage extensions.
             -d, --docs           Build the documentation from source using sphinx.
             -n, --new            Create a new external project or internal task from template.
             -c, --conda [NAME]   Create the conda environment for Isaac Lab. Default name is 'env_isaaclab'.
-            -u, --uv [NAME]      Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'.
--- a/docs/source/setup/installation/include/src_python_virtual_env.rst
+++ b/docs/source/setup/installation/include/src_python_virtual_env.rst
@@ -50,8 +50,11 @@ instead of *./isaaclab.sh -p* or *isaaclab.bat -p*.
 
                # Option 1: Default environment name 'env_isaaclab'
                ./isaaclab.sh --uv  # or "./isaaclab.sh -u"
-               # Option 2: Custom name
+               # Option 2: Custom environment name
                ./isaaclab.sh --uv my_env  # or "./isaaclab.sh -u my_env"
+               # Option 3: Custom environment name with additional args to uv
+               # the following command set the path to python interpreter and use verbose output
+               ./isaaclab.sh --uv my_env -- --python ${HOME}/isaacsim50/kit/python/bin/python3 --verbose
 
             .. code:: bash
 


rowoflo:
I was blocked by not being able to specify the python path to `uv` as well. I wrote a fix but then found this PR. The diff for my fix is below. You might want to use the portion that has it default to a given Python version based on the Isaac Sim version as stated [here](https://isaac-sim.github.io/IsaacLab/main/source/setup/installation/binaries_installation.html#setting-up-a-python-environment-optional).

```diff
╰─$ git diff
diff --git a/isaaclab.sh b/isaaclab.sh
index 263ff88e8a..92f66eb9d3 100755
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -496,7 +496,8 @@ print_help () {
     echo -e "\t-d, --docs           Build the documentation from source using sphinx."
     echo -e "\t-n, --new            Create a new external project or internal task from template."
     echo -e "\t-c, --conda [NAME]   Create the conda environment for Isaac Lab. Default name is 'env_isaaclab'."
-    echo -e "\t-u, --uv [NAME]      Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'."
+    echo -e "\t-u, --uv [NAME] [PYTHON_PATH]  Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'."
+    echo -e "\t                     Optional PYTHON_PATH: Python interpreter path (e.g., '3.11', 'python3.10', or '/usr/bin/python3.11')."
     echo -e "\n" >&2
 }

@@ -599,8 +600,25 @@ while [[ $# -gt 0 ]]; do
                 uv_env_name=$2
                 shift # past argument
             fi
+
+            # check if python path is provided
+            if [ -n "$2" ] && [[ "$2" != -* ]]; then
+                echo "[INFO] Using python interpreter: $2"
+                python_path="$2"
+                shift # past argument
+            else
+                # use default python version based on Isaac Sim version
+                if is_isaacsim_version_4_5; then
+                    echo "[INFO] Detected Isaac Sim 4.5 → using python 3.10"
+                    python_path="3.10"
+                else
+                    echo "[INFO] Using default python 3.11"
+                    python_path="3.11"
+                fi
+            fi
+
             # setup the uv environment for Isaac Lab
-            setup_uv_env ${uv_env_name}
+            setup_uv_env ${uv_env_name} ${python_path}
             shift # past argument
             ;;
         -f|--format)
```

KyleM73:
I agree with @rowoflo that the default should be the version based on isaacsim (that's what I had intended the default behavior to be). Thanks for this PR! I'm not sure if you need so much context in the help script, but the core change + default fixes the problem.

KyleM73:
@kellyguo11 with a little polish this should solve #3524

kwang-12:
@KyleM73 
The help texts was included to showcase how to use the option.
For my use case, I have multiple python3.11 installed so I need the path to be specific when creating the venv.

isaaclab.sh
```diff
@@ -496,7 +496,18 @@ print_help () {
...
+    echo -e "\t                          Create the uv environment for Isaac Lab. Default name is 'env_isaaclab'."
+    echo -e "\t                          Optionally specify a Python interpreter path to use for the environment."
+    echo -e "\nExamples:"
+    echo -e "\t 1. uv environment creation:"
+    echo -e "\t\t1.a create env_isaaclab with system python3"
+    echo -e "\t\t$(basename "$0") --uv"
+    echo -e "\t\t1.b create myenv with system python3"
+    echo -e "\t\t$(basename "$0") --uv myenv               "
+    echo -e "\t\t1.c create myenv with specified Python 3.11"
+    echo -e "\t\t$(basename "$0") --uv myenv --python ${HOME}/isaacsim50/kit/python/bin/python3"
```
Mayankm96: Maybe this should go to the docs instead of having it here? Or maybe more optional, i.e. printed only when `uv --help` is passed? Otherwise it becomes many things.

isaaclab.sh
```diff
@@ -590,18 +601,20 @@ while [[ $# -gt 0 ]]; do
...
         -u|--uv)
-            # use default name if not provided
-            if [ -z "$2" ]; then
-                echo "[INFO] Using default uv environment name: env_isaaclab"
-                uv_env_name="env_isaaclab"
-            else
-                echo "[INFO] Using uv environment name: $2"
-                uv_env_name=$2
-                shift # past argument
+            uv_env_name="env_isaaclab"
```
Mayankm96, line 605: Is it better to just pass through all arguments from the user following `--uv`? 

This will then include name, python path or other uv-related params that a user may want to pass through?

kwang-12: I have updated the shell for passing additional args to uv.
now it's possible to do something like:
`./IsaacLab/isaaclab.sh --uv custom_env_1 -- --python ${HOME}/isaacsim50/kit/python/bin/python3 --verbose`
additional args to uv is delimited by `--`. I think this is cleanest version yet.
check out 3ff0dcc

KyleM73: Can we just enforce that the first arg is always the env name, second arg (if provided) is always python path, etc.? The extra `--` would be great to remove. By default the python path just needs to be matched to the correct version from isaacsim, 3.11 currently. This is the main bug that needs to be resolved.

kwang-12: 5b0c2d2 includes updated doc
https://github.com/isaac-sim/IsaacLab/blob/5b0c2d2447f870b7a17533592cb7475599511e2c/docs/source/setup/installation/include/src_python_virtual_env.rst?plain=1#L51-L57
anything after -- are forwarded to uv which makes it versatile IMO.

PR #3565, opened 2025-09-25
[Newton] Adds policy transfer script for sim2sim transfer from Newton to physX

Milad-Rakhsha-NV:
# Description

This PR adds a play script to physX-based IsaacLab to make it possible to play a Newton-based trained policy. Tested environments are H1/G1/Anymal-D but other exisiting Newton environment should transfer as well.

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/docs/source/experimental-features/newton-physics-integration/sim-to-sim.rst
+++ b/docs/source/experimental-features/newton-physics-integration/sim-to-sim.rst
@@ -2,38 +2,40 @@
 
 Sim-to-Sim Policy Transfer
 ==========================
-This section provides examples of sim-to-sim policy transfer using the Newton backend. Sim-to-sim transfer is an essential step before real robot deployment because it verifies that policies work across different simulators. Policies that pass sim-to-sim verification are much more likely to succeed on real robots.
+This section provides examples of sim-to-sim policy transfer between PhysX and Newton backends. Sim-to-sim transfer is an essential step before real robot deployment because it verifies that policies work across different simulators. Policies that pass sim-to-sim verification are much more likely to succeed on real robots.
 
 
 Overview
 --------
 
-This guide shows how to run a PhysX-trained policy on the Newton backend. While the method works for any robot and physics engine, it has only been tested with Unitree G1, Unitree H1, and ANYmal-D robots using PhysX-trained policies.
+This guide shows how to transfer policies between PhysX and Newton backends in both directions. The main challenge is that different physics engines may parse the same robot model with different joint and link ordering.
 
-PhysX-trained policies expect joints and links in a specific order determined by how PhysX parses the robot model. However, Newton may parse the same robot with different joint and link ordering.
+Policies trained in one backend expect joints and links in a specific order determined by how that backend parses the robot model. When transferring to another backend, the joint ordering may be different, requiring remapping of observations and actions.
 
 In the future, we plan to solve this using **robot schema** that standardizes joint and link ordering across different backends.
 
-Currently, we solve this by remapping observations and actions using joint mappings defined in YAML files. These files specify joint names in both PhysX order (source) and Newton order (target). During policy execution, we use this mapping to reorder observations and actions so they work correctly with Newton.
+Currently, we solve this by remapping observations and actions using joint mappings defined in YAML files. These files specify joint names in both source and target backend orders. During policy execution, we use this mapping to reorder observations and actions so they work correctly with the target backend.
+
+The method has been tested with Unitree G1, Unitree Go2, Unitree H1, and ANYmal-D robots for both transfer directions.
 
 
 What you need
 ~~~~~~~~~~~~~
 
-- A policy checkpoint trained with PhysX (RSL-RL).
-- A joint mapping YAML for your robot under ``scripts/newton_sim2sim/mappings/``.
-- The provided player script: ``scripts/newton_sim2sim/rsl_rl_transfer.py``.
+- A policy checkpoint trained with either PhysX or Newton (RSL-RL).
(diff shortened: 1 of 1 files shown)


scripts/newton_sim2sim/rsl_rl_transfer.py
```diff
@@ -0,0 +1,239 @@
...
+    # set the environment seed
+    # note: certain randomizations occur in the environment initialization so we set the seed here
+    env_cfg.seed = agent_cfg.seed
+    env_cfg.sim.device = args_cli.device if args_cli.device is not None else env_cfg.sim.device
+
+    # specify directory for logging experiments
+    log_root_path = os.path.join("logs", "rsl_rl", agent_cfg.experiment_name)
+    log_root_path = os.path.abspath(log_root_path)
+    print(f"[INFO] Loading experiment from directory: {log_root_path}")
+    if args_cli.use_pretrained_checkpoint:
```
kellyguo11: is this still applicable for sim2sim in this context? we only provide physx pretrained checkpoints right now, so maybe this only makes sense if we are running newton?

scripts/sim2sim_transfer/rsl_rl_transfer.py
```diff
@@ -0,0 +1,239 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Script to play a checkpoint of an RL agent from RSL-RL with policy transfer capabilities."""
```
kellyguo11, line 6: I think it'll be good to add this into the documentation as well. right now, we only provide commands for running physx trained policies in newton

scripts/newton_sim2sim/policy_mapping.py
```diff
@@ -0,0 +1,74 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Policy transfer utilities for handling joint ordering differences between physics engines.
+
+This module provides functionality to handle policy transfer between different physics engines
+that may have different joint orderings. It supports loading joint mappings from YAML files
+and provides observation and action remapping functions.
+"""
+
+import yaml
```
kellyguo11: in general, we want to keep the scripts folder to only contain python scripts that can be executed directly. For simplicity, maybe we can just move this into the rsl_rl_transfer.py script?

scripts/sim2sim_transfer/config/newton_to_physx_anymal_d.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Joint names in the source physics engine where policy is trained (Newton)
```
kellyguo11, line 6: is this the temporary method for specifying these joint names, eventually I think we'd want to parse this from the robot schema or enforce some ordering between physx and newton?

also, since they specify source and target, maybe in the file name, instead of sim2sim, we specify explicitly newton2physx

scripts/sim2sim_transfer/config/newton_to_physx_anymal_d.yaml
```diff
@@ -0,0 +1,34 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
kellyguo11, line 1: might just be a personal taste, but I'm not a fan of "mappings" as the folder name. I would propose to just go with something simple like config

Milad-Rakhsha-NV: It is a bit tricky now, because I made this consistent with physX-to-Newton (dev/newton branch) and the commands to use this is no part of a blogpost and I am not sure if we can still modify them.

scripts/sim2sim_transfer/config/newton_to_physx_anymal_d.yaml
Milad-Rakhsha-NV: see my comments below, we can change these things but it is a bit tricky now

scripts/newton_sim2sim/policy_mapping.py
Milad-Rakhsha-NV: yep will do that

scripts/sim2sim_transfer/rsl_rl_transfer.py
Milad-Rakhsha-NV: sounds good, let's decide on the structure of these scripts and yaml and whether it should be consistent with dev/newton ones and I will take care of this

kellyguo11:
ok let's fix what we can for now. please also run the formatter to resolve the pre-commit failures

scripts/sim2sim_transfer/config/newton_to_physx_anymal_d.yaml
Milad-Rakhsha-NV: robot schema is goign to be a bit more complex but is doable and we can look into later

Milad-Rakhsha-NV:
@kellyguo11 I updated the documentation as well. 
I updated the feature/dev branch for consistency as well please see this 
https://github.com/isaac-sim/IsaacLab/pull/3581

PR #3598, opened 2025-09-30
Updates path to G1 asset used by locomanipulation task

michaellin6:
# Description

G1 asset in Nucleus was updated breaking the locomanipulation mimic workflow since one of the pick place datasets was recorded before the asset changes. The old G1 asset was included as a copy in a different path, and this PR updates the G1 asset to point to that temporary path. In the long term, a new dataset will be collected but needs to go through testing.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_assets/isaaclab_assets/robots/unitree.py
+++ b/source/isaaclab_assets/isaaclab_assets/robots/unitree.py
@@ -387,7 +387,7 @@ This configuration removes most collision meshes to speed up simulation.
 
 G1_29DOF_CFG = ArticulationCfg(
     spawn=sim_utils.UsdFileCfg(
-        usd_path=f"{ISAAC_NUCLEUS_DIR}/Robots/Unitree/G1/g1.usd",
+        usd_path=f"{ISAAC_NUCLEUS_DIR}/Samples/Groot/Robots/g1_29dof_hand.usd",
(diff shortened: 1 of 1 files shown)


stevfeng:
Looks good to me, we need to make sure ISAAC_NUCLEUS_DIR points to the 5.1 staging server

source/isaaclab_assets/isaaclab_assets/robots/unitree.py
```diff
@@ -387,7 +387,7 @@
 
 G1_29DOF_CFG = ArticulationCfg(
     spawn=sim_utils.UsdFileCfg(
-        usd_path=f"{ISAAC_NUCLEUS_DIR}/Robots/Unitree/G1/g1.usd",
+        usd_path=f"{ISAAC_NUCLEUS_DIR}/Samples/Groot/Robots/g1_29dof_hand.usd",
```
Mayankm96, line 390: This makes the temporary asset global for all the environments in IsaacLab, right? Is that intentional? We want to use the most sim-ready asset at all places. If only Mimic workflow requires the old asset version, please make the change for it over there instead.

michaellin6:
These changes are not necessary now as we have updated the datasets that were broken by the asset change rather than keeping a temporary copy of the old asset.

PR #3716, opened 2025-10-15
Adds preserve order flag to JointPositionToLimitsAction

renezurbruegg:
# Description

Adds `preserve_order` flag to `JointPositionToLimitsActionCfg`

## Type of change

## Screenshots

## Checklist


renezurbruegg:
> The changes looks good to me, though can you share what bug you went into without it?

Without the preserving order flag, the order of the joints in the action term are defined by the internal parsing of physx. 
So e,g, providing `["index_q1", "middle_q1", "ring_q1"]` as joint expressions (with `preserve_order=False`, which is the default), can lead to the action term internally using `["ring_q1",  "middle_q1", "index_q1"]` as joint order.

kellyguo11:
do we want to consider the `EMAJointPositionToLimitsAction` case for this change?

PR #3722, opened 2025-10-15
Adds onnxscript dependency to isaaclab_rl module

ooctipus:
# Description

This PR includes onnxscript in isaaclab_rl dependency. Without explicit pointing, aarch64 platform ignores it and causes downstream onnx exporter to break.

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_rl/setup.py
+++ b/source/isaaclab_rl/setup.py
@@ -46,7 +46,7 @@ EXTRAS_REQUIRE = {
         "rl-games @ git+https://github.com/isaac-sim/rl_games.git@python3.11",
         "gym",
     ],  # rl-games still needs gym :(
-    "rsl-rl": ["rsl-rl-lib==3.0.1", "onnxscript>=0.5"]  # linux aarch 64 requires manual onnxscript installation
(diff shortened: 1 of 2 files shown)


source/isaaclab_rl/setup.py
```diff
@@ -34,6 +34,8 @@
     # make sure this is consistent with isaac sim version
     "pillow==11.3.0",
     "packaging<24",
+    # linux aarch 64 requires manual onnxscript installation
```
Mayankm96: Isn't onnx scripting only needed for RSL-RL? Should we only install it for that case?

ooctipus: the exporter code is written in isaaclab_rl, so it makes sense to include dependency in isaaclab? If exporter is written entirely in rsl_rl, and you can do from rslrl_rl import exporter, then we don't need this dependency in isaaclab.

kellyguo11: I think Mayank might have meant adding it to https://github.com/isaac-sim/IsaacLab/blob/main/source/isaaclab_rl/setup.py#L49 together with the RSL RL dependency, so it doesn't get installed if users don't install RSL RL

ooctipus: Ahhh understood

PR #3724, opened 2025-10-15
Adds enable_cameras=True to pink_ik test

jaybdub:
# Description

Adds enable_cameras=True to pink_ik test to support envs with cameras

## Type of change

- 
## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/test/controllers/test_pink_ik.py
+++ b/source/isaaclab/test/controllers/test_pink_ik.py
@@ -14,7 +14,7 @@ if sys.platform != "win32":
 from isaaclab.app import AppLauncher
 
 # launch omniverse app
-simulation_app = AppLauncher(headless=True).app
+simulation_app = AppLauncher(headless=True, enable_cameras=True).app
 
 """Rest everything follows."""
 


michaellin6 (changes requested):
Instead of adding the camera object to the LocomanipulationG1EnvCfg environment, we should add it at the child class used by the navigation generation G1LocomanipulationSDGSceneCfg since it is already inheriting from LocomanipulationG1EnvCfg and you can just add the camera component there. Adding it to LocomanipulationG1EnvCfg will make even the basic pick place datasets have image data which are not needed for state-only based models

Mayankm96 (approved):
I find it strange that we have test in the core extension that rely on underlying task implementation. This is not how the modules should be structured. The controllers tests should be made independent of underlying task choices.

We need to have this fixed ASAP. Let's have a discussion next week.

@kellyguo11 for vis.

michaellin6:
> I find it strange that we have test in the core extension that rely on underlying task implementation. This is not how the modules should be structured. The controllers tests should be made independent of underlying task choices.
> 
> We need to have this fixed ASAP. Let's have a discussion next week.
> 
> @kellyguo11 for vis.

Is the dependency supposed to be a one way thing? Only isaaclab_tasks depends on isaaclab core? The ik tests I've put span all the humanoid task environments that use pink IK, so its not specific to a particular task environment and mostly serves as a signal to catch deltas in either controller changes or asset changes that break performance requirements. If it is mostly an issue about the location of the tests we can discuss and I'm open to suggestions.

kellyguo11:
closing in favor of #3740. let's follow up on the discussion around the test structure

PR #3728, opened 2025-10-15
Adds test suite for kinematic rigidbody scaling/rendering related issue.

ooctipus:
# Description

This PR adds the test suite that check for kinematic rigid object scaling, and rendering issue. It was introduced in IsaacSim 5.0, caused pain to many users, and was uncaught by any tests. Though it is now fixed in IsaacSim 5.1, this test ensure if anything related happens again we fail fast. 

## Checklist

### DIFF
--- a/source/isaaclab/test/assets/test_kinematic_rigid_object.py
+++ b/source/isaaclab/test/assets/test_kinematic_rigid_object.py
@@ -1,0 +1,185 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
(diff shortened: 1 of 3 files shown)


Mayankm96 (changes requested):
This test seems more concerning cameras than rigid obejct class. I think it should be moved there instead.

source/isaaclab/test/assets/test_kinematic_rigid_object.py
```diff
@@ -0,0 +1,185 @@
...
+from isaaclab.app import AppLauncher
+
+# launch omniverse app
+simulation_app = AppLauncher(headless=True, enable_cameras=True).app
+
+"""Rest everything follows."""
+
+import torch
+
+import isaacsim.core.utils.prims as prim_utils
```
kellyguo11, line 17: do we want to remove isaac sim APIs?

Issue #3720, opened 2025-10-15
[Bug Report] TerminationManager._term_dones not updated per step (stale values during rollout)


Weifan408:
### Describe the bug

Hi, thank you for your excellent work.
I noticed that in the current implementation of `TerminationManager.compute()`, the internal buffer `self._term_dones` is **only updated when a termination actually occurs, but not during the normal running steps** of the environment.

#### Current Behavior
```python
def compute(self) -> torch.Tensor:
        """Computes the termination signal as union of individual terms.

        This function calls each termination term managed by the class and performs a logical OR operation
        to compute the net termination signal.

        Returns:
            The combined termination signal of shape (num_envs,).
        """
        # reset computation
        self._truncated_buf[:] = False
        self._terminated_buf[:] = False
        # iterate over all the termination terms
        for i, term_cfg in enumerate(self._term_cfgs):
            value = term_cfg.func(self._env, **term_cfg.params)
            # store timeout signal separately
            if term_cfg.time_out:
                self._truncated_buf |= value
            else:
                self._terminated_buf |= value
            # add to episode dones
            rows = value.nonzero(as_tuple=True)[0]  # indexing is cheaper than boolean advance indexing
            if rows.numel() > 0: 
                self._term_dones[rows] = False
                self._term_dones[rows, i] = True
        # return combined termination signal
        return self._truncated_buf | self._terminated_buf
```
Here, `self._term_dones` is only modified when `rows.numel() > 0`, i.e., when a specific termination condition is triggered.
During the rest of the rollout (normal running steps), `self._term_dones` keeps **the values from the previous episode**, and is therefore stale.

#### Consequences
For logging (e.g. inside reset()), this behavior is harmless — it still correctly summarizes the last episode via:
```python
        last_episode_done_stats = self._term_dones.float().mean(dim=0)
```
However, for runtime queries like:
```python
get_term(name)
get_active_iterable_terms(env_idx)
```
the returned tensors will not reflect the current state of termination terms, but rather the previous episode’s data.
This can be problematic if a user or module expects real-time termination signals per environment step.

#### Expected Behavior
`self._term_dones` or a new parameter should be continuously updated at each compute() call, `get_term(name)` returns the correct, up-to-date boolean tensor for the current environment step — even when no episode termination occurs.

#### Suggestion
A minimal fix could be to explicitly clear the term buffer before the loop:
```python
self._term_dones[:] = False
```

### Checklist

[...]

PR #3745, opened 2025-10-17
Separates per-step termination and last-episode termination bookkeeping

### LINKED ISSUE #2977: [Bug Report] Termination Overcounting Caused by Missing Log Buffer Reset in manager_based_rl_env.py

ooctipus:
# Description

This PR fixes the issue where get_done_term returned last episode value rather than current step value.

This PR realizes values used for get_term should be different from that used for logging, and mixed useage leads to non-intuitive behavior.

 using per-step value for logging leads to overcounting and undercounting reported in #2977
 using last-episode value for get_term leads to misalignment with expectation reported in #3720 

Fixes #2977 #3720 

---

The logging behavior remains *mostly* the same as #3107, and and also got rid of the weird overwriting behavior(yay).
I get exactly the same termination curve as  #3107 when run on `Isaac-Velocity-Rough-Anymal-C-v0`

Here is a benchmark summary with 1000 steps running `Isaac-Velocity-Rough-Anymal-C-v0 ` with 4096 envs

Before #3107: 

`| termination.compute     |       0.229 ms|`
`| termination.reset       |       0.007 ms|`

PR #3107: 

`| termination.compute     |       0.274 ms|`
`| termination.reset       |       0.004 ms|`

This PR: 

`| termination.compute     |       0.258 ms|`
`| termination.reset       |       0.004 ms|`

We actually see improvement, this is due to the fact that expensive maintenance of last_episode_value is only computed once per compute(#3107 computes last_episode_value for every term)

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.47.8"
+version = "0.47.9"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"


Mayankm96:
Can we add some sort of tests to make sure this behavior does not break in the future? We want to avoid such regressions.

PerFrivik:
thanks!

PR #3818, opened 2025-10-23
Adds parameter to specify number of rerenders after environment reset

peterd-NV:
# Description

Adds a new parameter to ManagerBasedEnv and DirectRLEnv to give users better control over rerender on reset behaviour. The new parameter `num_rerenders_on_reset` allows users to explicitly define the number of re-render steps after an env reset. When using DLSS, this allows for the elimination of artifacts/ghosting that are present after a single rendering step. 

Add a deprecation warning for the old parameter `rerender_on_reset`. Functionality of old parameter is preserved. 

Updates the existing visuomotor envs to use new rerendering API together with DLAA for high quality rendering. 

Fixes # (issue)

Non-DLSS denoising is not supported on aarch64. There are also future plans from the rendering team to disable use of non-DLSS antialiasing for all platforms in the future. This causes an issue for visuomotor envs which suffer from image ghosting/artifacts when using DLSS. The new rerendering API allows for users of visuomotor envs to enable DLSS/DLAA while preserving image integrity.

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.47.8"
+version = "0.47.10"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -2,8 +2,8 @@ Changelog
 ---------
 
 
-0.47.8 (2025-11-04)
-~~~~~~~~~~~~~~~~~~~
+0.47.10 (2025-11-06)
+~~~~~~~~~~~~~~~~~~~~
 
 Added
 ^^^^^
@@ -15,6 +15,27 @@ Changed
 ^^^^^^^
 
 * Added deprecation warning for ``rerender_on_reset`` parameter in ManagerBasedEnv and DirectRLEnv.
+
+
+0.47.9 (2025-11-05)
+~~~~~~~~~~~~~~~~~~~
+
+Changed
+^^^^^^^
+
+* Fixed termination term bookkeeping in :class:`~isaaclab.managers.TerminationManager`:
+  per-step termination and last-episode termination bookkeeping are now separated.
+  last-episode dones are now updated once per step from all term outputs, avoiding per-term overwrites
+  and ensuring Episode_Termination metrics reflect the actual triggering terms.
+
+
+0.47.8 (2025-11-06)
+~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+
+* Added parameter :attr:`~isaaclab.terrains.TerrainImporterCfg.use_terrain_origins` to allow generated sub terrains with grid origins.
 
 
 0.47.7 (2025-10-31)
--- a/source/isaaclab/isaaclab/managers/termination_manager.py
+++ b/source/isaaclab/isaaclab/managers/termination_manager.py
@@ -63,6 +63,8 @@ class TerminationManager(ManagerBase):
         self._term_name_to_term_idx = {name: i for i, name in enumerate(self._term_names)}
         # prepare extra info to store individual termination term information
         self._term_dones = torch.zeros((self.num_envs, len(self._term_names)), device=self.device, dtype=torch.bool)
+        # prepare extra info to store last episode done per termination term information
+        self._last_episode_dones = torch.zeros_like(self._term_dones)
         # create buffer for managing termination per environment
         self._truncated_buf = torch.zeros(self.num_envs, device=self.device, dtype=torch.bool)
         self._terminated_buf = torch.zeros_like(self._truncated_buf)
@@ -138,7 +140,7 @@ class TerminationManager(ManagerBase):
             env_ids = slice(None)
         # add to episode dict
         extras = {}
-        last_episode_done_stats = self._term_dones.float().mean(dim=0)
+        last_episode_done_stats = self._last_episode_dones.float().mean(dim=0)
         for i, key in enumerate(self._term_names):
             # store information
             extras["Episode_Termination/" + key] = last_episode_done_stats[i].item()
@@ -169,15 +171,17 @@ class TerminationManager(ManagerBase):
             else:
                 self._terminated_buf |= value
             # add to episode dones
-            rows = value.nonzero(as_tuple=True)[0]  # indexing is cheaper than boolean advance indexing
-            if rows.numel() > 0:
-                self._term_dones[rows] = False
-                self._term_dones[rows, i] = True
+            self._term_dones[:, i] = value
+        # update last-episode dones once per compute: for any env where a term fired,
+        # reflect exactly which term(s) fired this step and clear others
+        rows = self._term_dones.any(dim=1).nonzero(as_tuple=True)[0]
+        if rows.numel() > 0:
+            self._last_episode_dones[rows] = self._term_dones[rows]
         # return combined termination signal
         return self._truncated_buf | self._terminated_buf
 
     def get_term(self, name: str) -> torch.Tensor:
-        """Returns the termination term with the specified name.
+        """Returns the termination term value at current step with the specified name.
 
         Args:
             name: The name of the termination term.
@@ -190,7 +194,8 @@ class TerminationManager(ManagerBase):
     def get_active_iterable_terms(self, env_idx: int) -> Sequence[tuple[str, Sequence[float]]]:
         """Returns the active terms as iterable sequence of tuples.
 
-        The first element of the tuple is the name of the term and the second element is the raw value(s) of the term.
+        The first element of the tuple is the name of the term and the second element is the raw value(s) of the term
+        recorded at current step.
 
         Args:
             env_idx: The specific environment to pull the active terms from.
--- a/source/isaaclab/isaaclab/terrains/terrain_importer.py
+++ b/source/isaaclab/isaaclab/terrains/terrain_importer.py
@@ -87,8 +87,11 @@ class TerrainImporter:
                 cfg=self.cfg.terrain_generator, device=self.device
             )
             self.import_mesh("terrain", terrain_generator.terrain_mesh)
-            # configure the terrain origins based on the terrain generator
-            self.configure_env_origins(terrain_generator.terrain_origins)
+            if self.cfg.use_terrain_origins:
+                # configure the terrain origins based on the terrain generator
+                self.configure_env_origins(terrain_generator.terrain_origins)
+            else:
+                self.configure_env_origins()
             # refer to the flat patches
             self._terrain_flat_patches = terrain_generator.flat_patches
         elif self.cfg.terrain_type == "usd":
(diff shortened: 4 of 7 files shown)


ooctipus:
If the environment is parallel, then non-resetting environment will also gets re-rendered. If the goal is to remove artifacts or ghosting, this may work around it at the cost of dramatically increase per step cost in high number of environments. RL workflows may  run over 10k envs, and you almost get reset every step,. For this reason I'd prefer this path not exist at all. Should we consult with rendering team before we workaround in this way? If the it is currently only required for mimic related take, why can't we provide it from mimic env?

source/isaaclab/isaaclab/envs/direct_rl_env_cfg.py
```diff
@@ -213,15 +213,26 @@ class DirectRLEnvCfg:
     """
 
     rerender_on_reset: bool = False
-    """Whether a render step is performed again after at least one environment has been reset.
+    """[DEPRECATED] Use num_rerenders_on_reset instead.
```
Mayankm96: Please use the sphinx directive for this everywhere: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

You can add the explanation and clear the remainig docstring so that it is cleaner for users and they all are pointed to the right attribute directly. Something like:

```
.. deprecated: v2.3.0
   Please use :attr:`num_rerenders_on_reset`.

   Setting this flag as True is the same as setting :attr:`num_rerenders_on_reset` to one.
```

source/isaaclab/isaaclab/envs/manager_based_env.py
```diff
@@ -190,6 +190,13 @@ def __init__(self, cfg: ManagerBasedEnvCfg):
         if self.cfg.export_io_descriptors:
             self.export_IO_descriptors()
 
+        # show deprecation message for rerender_on_reset
+        if self.cfg.rerender_on_reset:
```
Mayankm96, line 195: You can throw the warning and then internally set the variable `num_rerenders_on_reset` to 1. This then simplifies your check at other places. Also could you use deprecate warning from python instead of omni logger: https://docs.python.org/3/library/warnings.html

source/isaaclab/isaaclab/envs/direct_rl_env_cfg.py
peterd-NV: Update to use sphinx deprecation directive and following format of other existing deprecations in Isaac Lab.

source/isaaclab/isaaclab/envs/manager_based_env.py
peterd-NV: Thanks for the recommendation. Setting the variable `num_rerenders_on_reset` internally and removed checks elsewhere in the code to simplify.

peterd-NV:
Thanks for the comments and suggestions @Mayankm96. I've updated the PR to use the proper deprecation directives and simplify the code by always setting the new parameter internally and using that throughout the code.

I've also validated policy performance with the new rendering settings on both Franka/GR1 and can confirm that they are expected with the latest iteration of this PR.

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -219,6 +220,16 @@ def __init__(self, cfg: DirectRLEnvCfg, render_mode: str | None = None, **kwargs
         # video matches the simulation
         self.metadata["render_fps"] = 1 / self.step_dt
 
+        # show deprecation message for rerender_on_reset
+        if self.cfg.rerender_on_reset:
+            warnings.warn(
+                "\033[93m\033[1m[DEPRECATION WARNING] DirectRLEnvCfg.rerender_on_reset is deprecated. Use"
```
kellyguo11, line 226: what is the purpose of `\033[93m\033[1m`?

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -219,6 +220,16 @@ def __init__(self, cfg: DirectRLEnvCfg, render_mode: str | None = None, **kwargs
...
+                " DirectRLEnvCfg.num_rerenders_on_reset instead.\033[0m",
+                FutureWarning,
```
kellyguo11, line 231: is there a reason to use this instead of DeprecationWarning?

source/isaaclab/isaaclab/envs/direct_rl_env_cfg.py
```diff
@@ -222,6 +222,22 @@ class DirectRLEnvCfg:
       to reflect the latest states from the reset. This comes at a cost of performance as an additional render
       step will be performed after each time an environment is reset.
 
+    .. deprecated:: 0.47.2
```
kellyguo11: maybe we can use 2.3.1 for the version instead. I'm not sure if 0.47.2 is as obvious to users

kellyguo11:
> If the environment is parallel, then non-resetting environment will also gets re-rendered. If the goal is to remove artifacts or ghosting, this may work around it at the cost of dramatically increase per step cost in high number of environments. RL workflows may run over 10k envs, and you almost get reset every step,. For this reason I'd prefer this path not exist at all. Should we consult with rendering team before we workaround in this way? If the it is currently only required for mimic related take, why can't we provide it from mimic env?

I think I can see this being a valid option for RL environments as well. We should note for sure that it will come at a high performance cost, but if users really need the extra quality for renderer convergence, then the same path could still be applied.

source/isaaclab/isaaclab/envs/direct_rl_env.py
peterd-NV: This is to format the string to be bolded and yellow. Without it, I find the message is very easily lost in the Isaac Lab logs.

source/isaaclab/isaaclab/envs/direct_rl_env.py
peterd-NV: By default in python warnings, DeprecationWarning is not shown to users running the code. Using FutureWarning ensures the warning is printed and shown in the logs.

source/isaaclab/isaaclab/envs/direct_rl_env_cfg.py
peterd-NV: Good call, that makes it easier to understand. I've updated to 2.3.1

kellyguo11:
hmm the linter check is failing with a strange error. are you seeing that locally as well?

peterd-NV:
> hmm the linter check is failing with a strange error. are you seeing that locally as well?

No I’m not seeing it locally when running `isaaclab.sh -f`. I reran the linter job on github earlier and it failed with the same error.

source/isaaclab/isaaclab/envs/direct_rl_env.py
Mayankm96: Can you do it more like: 

```
msg = ( '"""""") # my-multi line thingie
omni.log.warn(msg)
warnings.warn(msg...)
```

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/stack/config/franka/__init__.py
```diff
@@ -65,7 +65,7 @@
     entry_point="isaaclab.envs:ManagerBasedRLEnv",
     kwargs={
         "env_cfg_entry_point": stack_ik_rel_visuomotor_env_cfg.FrankaCubeStackVisuomotorEnvCfg,
-        "robomimic_bc_cfg_entry_point": os.path.join(agents.__path__[0], "robomimic/bc_rnn_image_84.json"),
+        "robomimic_bc_cfg_entry_point": os.path.join(agents.__path__[0], "robomimic/bc_rnn_image_200.json"),
```
Mayankm96: NIT: unrelated changes. could we remove them from this MR?

peterd-NV: Hi Mayank, this change is required for this MR as changing the re-render API and using DLAA requires an update to the visuomotor envs to still get policy success in our franka demo workflow. In my testing, the old 84x84 camera/policy does not work with DLAA and requires a higher resolution. Thus I've updated both the camera res to [200x200](https://github.com/isaac-sim/IsaacLab/pull/3818/files/edfd95f725ac14f654d7b11b29f35bdb09ca5f39#diff-37fe44e049f8f9533ba6fbce1e20cfb2ec97e6825849c48eff9d7a2034533605R207) and included the new training config.

source/isaaclab/isaaclab/envs/direct_rl_env.py
peterd-NV: Thanks, updated to use msg string and use omni.log and warnings together.

source/isaaclab/isaaclab/envs/direct_rl_env.py
peterd-NV: Following Mayank's comments, this has been updated to use a message string followed by both omni.log and warnings. Since omni log already displays the message to users, I've reverted back to using DeprecationWarning.

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -219,6 +220,20 @@ def __init__(self, cfg: DirectRLEnvCfg, render_mode: str | None = None, **kwargs
         # video matches the simulation
         self.metadata["render_fps"] = 1 / self.step_dt
 
+        # show deprecation message for rerender_on_reset
+        if self.cfg.rerender_on_reset:
+            msg = (
+                "[DEPRECATION WARNING] ManagerBasedEnvCfg.rerender_on_reset is deprecated. Use"
+                " ManagerBasedEnvCfg.num_rerenders_on_reset instead."
+            )
+            omni.log.warn(msg)
```
kellyguo11: we are trying to move away from omni.log, we can probably remove this line

peterd-NV: I've removed the spark limitation from the docs and removed omni.log.

source/isaaclab/isaaclab/envs/manager_based_env.py
```diff
@@ -190,6 +191,19 @@ def __init__(self, cfg: ManagerBasedEnvCfg):
...
+            msg = (
+                f"\033[93m\033[1m[DEPRECATION WARNING] ManagerBasedEnvCfg.rerender_on_reset is deprecated. Use"
+                f" ManagerBasedEnvCfg.num_rerenders_on_reset instead.\033[0m"
+            )
+            warnings.warn(
+                msg,
+                FutureWarning,
+                stacklevel=2,
+            )
+            self.cfg.num_rerenders_on_reset = 1
```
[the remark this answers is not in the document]
kellyguo11: this seems like a good suggestion.

kellyguo11:
could you also help take a look at the linter failure?

source/isaaclab/isaaclab/envs/manager_based_env.py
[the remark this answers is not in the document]
peterd-NV: Added in latest commit

peterd-NV:
Thanks @kellyguo11 @Mayankm96 for the reviews. I've addressed all comments and cleared out the lint errors. Should be ready for a final pass :)

PR #3873, opened 2025-10-29
Adds Haply device API with force feedback and teleoperation demo

mingxueg-nv:
# Description
Add Haply haptic device teleoperation support for robotic manipulation with force feedback.

## Type of change

1.  Add haply devices API for teleoperation with force feedback
2. Real-time teleoperation with force feedback demo via Haply Inverse3 handle

[Haply device documentation](https://docs.haply.co/docs/quick-start) and [usage](https://docs.haply.co/inverseSDK/)

**Usage (make sure your Haply device is connected):**
```bash
./isaaclab.sh -p scripts/demos/haply_teleoperation.py --websocket_uri ws://localhost:10001 --pos_sensitivity 1.65
```

## Screenshots
haply-frankal

## Checklist

### DIFF
--- a/docs/source/features/multi_gpu.rst
+++ b/docs/source/features/multi_gpu.rst
@@ -141,14 +141,14 @@ For the master node, use the following command, where ``--nproc_per_node`` repre
 
         .. code-block:: shell
 
-            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=localhost:5555 scripts/reinforcement_learning/rl_games/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/rl_games/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
     .. tab-item:: rsl_rl
         :sync: rsl_rl
 
         .. code-block:: shell
 
-            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=localhost:5555 scripts/reinforcement_learning/rsl_rl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/rsl_rl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
     .. tab-item:: skrl
         :sync: skrl
@@ -160,7 +160,7 @@ For the master node, use the following command, where ``--nproc_per_node`` repre
 
                 .. code-block:: shell
 
-                    python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=localhost:5555 scripts/reinforcement_learning/skrl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+                    python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=0 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/skrl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
             .. tab-item:: JAX
                 :sync: jax
@@ -181,14 +181,14 @@ For non-master nodes, use the following command, replacing ``--node_rank`` with 
 
         .. code-block:: shell
 
-            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=ip_of_master_machine:5555 scripts/reinforcement_learning/rl_games/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/rl_games/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
     .. tab-item:: rsl_rl
         :sync: rsl_rl
 
         .. code-block:: shell
 
-            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=ip_of_master_machine:5555 scripts/reinforcement_learning/rsl_rl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+            python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/rsl_rl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
     .. tab-item:: skrl
         :sync: skrl
@@ -200,7 +200,7 @@ For non-master nodes, use the following command, replacing ``--node_rank`` with 
 
                 .. code-block:: shell
 
-                    python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --rdzv_id=123 --rdzv_backend=c10d --rdzv_endpoint=ip_of_master_machine:5555 scripts/reinforcement_learning/skrl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
+                    python -m torch.distributed.run --nproc_per_node=2 --nnodes=2 --node_rank=1 --master_addr=<ip_of_master> --master_port=5555 scripts/reinforcement_learning/skrl/train.py --task=Isaac-Cartpole-v0 --headless --distributed
 
             .. tab-item:: JAX
                 :sync: jax
--- a/docs/source/setup/installation/index.rst
+++ b/docs/source/setup/installation/index.rst
@@ -89,8 +89,6 @@ Other notable limitations with respect to Isaac Lab include...
 
 #. Livestream and Hub Workstation Cache are not supported on the DGX spark.
 
-#. Multi-node training may require direct connections between Spark machines or additional network configurations.
-
 #. :ref:`Running Cosmos Transfer1 <running-cosmos>` is not currently supported on the DGX Spark.
 
 Troubleshooting
--- a/source/isaaclab_assets/isaaclab_assets/robots/__init__.py
+++ b/source/isaaclab_assets/isaaclab_assets/robots/__init__.py
@@ -7,11 +7,14 @@
 # Configuration for different assets.
 ##
 
+from .agibot import *
+from .agility import *
 from .allegro import *
 from .ant import *
 from .anymal import *
 from .cart_double_pendulum import *
 from .cartpole import *
+from .cassie import *
 from .fourier import *
 from .franka import *
 from .galbot import *


scripts/demos/haply_teleoperation.py
```diff
@@ -0,0 +1,374 @@
...
+
+        jacobian = robot.root_physx_view.get_jacobians()[:, ee_body_idx, :, arm_joint_indices]
+        ik_controller.set_command(command=target_pos_tensor, ee_quat=ee_quat_w)
+        joint_pos_des = ik_controller.compute(ee_pos_w, ee_quat_w, jacobian, current_joint_pos)
+
+        joint_pos_target = robot.data.joint_pos[0].clone()
+
+        # Update the joint
+        joint_pos_target[arm_joint_indices] = joint_pos_des[0]  # arm joints
+        joint_pos_target[6] = ee_rotation_angle  # end-effector rotation
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed, have removed overwriting joint7

scripts/demos/haply_teleoperation.py
```diff
@@ -0,0 +1,374 @@
...
+        count += 1
+
+        if args_cli.enable_force_feedback:
+            left_finger_forces = left_finger_sensor.data.net_forces_w[0, 0]
+            right_finger_forces = right_finger_sensor.data.net_forces_w[0, 0]
+            total_contact_force = (left_finger_forces + right_finger_forces) * 0.5
+            feedback_force = torch.clamp(total_contact_force, -2.0, 2.0)
+            haply_device.set_force_feedback(
+                feedback_force[0].item(), feedback_force[1].item(), feedback_force[2].item()
+            )
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed, default teleoperation with force feedback

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,414 @@
...
+        """
+        if not self.device_ready:
+            # Return zero command if device not ready
+            return torch.zeros(10, dtype=torch.float32, device=self._sim_device)
+
+        # Get latest cached data
+        with self.data_lock:
+            position = self.cached_data["position"].copy()
+            quaternion = self.cached_data["quaternion"].copy()
+            buttons = self.cached_data["buttons"].copy()
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,414 @@
...
+                f"Failed to connect to Haply WebSocket at {self.websocket_uri} "
+                f"within {timeout}s. Is the Haply SDK running?"
+            )
+
+    def __del__(self):
+        """Cleanup on deletion."""
+        self.running = False
+        with self.force_lock:
+            self.feedback_force = {"x": 0.0, "y": 0.0, "z": 0.0}
+        print("[INFO] Haply device disconnected")
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,414 @@
...
+import threading
+import time
+import torch
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any
+
+import websockets
+
+WEBSOCKETS_AVAILABLE = True
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed, added import check

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,438 @@
...
+                - connected: bool indicating if device is connected
+                - data_fresh: bool indicating if data is recent
+        """
+        with self.data_lock:
+            current_data = self.cached_data.copy()
+
+        data_fresh = self._is_data_fresh()
+        device_connected = (
+            current_data.get("inverse3_connected", False) and current_data.get("versegrip_connected", False)
+        ) and data_fresh
```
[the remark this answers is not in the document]
mingxueg-nv: Inverse3 and VerseGrip must both be connected

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,414 @@
...
+                                        "commands": {"set_cursor_force": {"values": current_force}},
+                                    }]
+                                }
+                                await ws.send(json.dumps(request_msg))
+
+                            # Control data rate
+                            await asyncio.sleep(1.0 / self.data_rate)
+
+                        except asyncio.TimeoutError:
+                            continue
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,438 @@
...
+
+        self.running = False
+
+        with self.force_lock:
+            self.feedback_force = {"x": 0.0, "y": 0.0, "z": 0.0}
+
+        # Wait for thread to finish
+        time.sleep(0.2)
+        self.device_ready = False
+        print("[INFO] Haply device disconnected")
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,392 @@
...
+        msg += "\tOutput: [x, y, z, qx, qy, qz, qw, btn_a, btn_b, btn_c]\n"
+        msg += "\tInverse3: Provides position (x, y, z) and force feedback\n"
+        msg += "\tVerseGrip: Provides orientation (quaternion) and buttons (a, b, c)"
+        return msg
+
+    def reset(self):
+        """Reset the device internal state."""
+        with self.force_lock:
+            self.feedback_force = {"x": 0.0, "y": 0.0, "z": 0.0}
+        self._prev_buttons = {"a": False, "b": False, "c": False}
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,392 @@
...
+                            if self.inverse3_device_id:
+                                with self.force_lock:
+                                    current_force = self.feedback_force.copy()
+
+                                request_msg = {
+                                    "inverse3": [{
+                                        "device_id": self.inverse3_device_id,
+                                        "commands": {"set_cursor_force": {"values": current_force}},
+                                    }]
+                                }
```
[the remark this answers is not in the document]
mingxueg-nv: Fixed

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,388 @@
...
+            ],
+            dtype=np.float32,
+        )
+
+        # Construct command tensor: [position(3), quaternion(4), buttons(3)]
+        command = np.concatenate([position, quaternion, button_states])
+
+        return torch.tensor(command, dtype=torch.float32, device=self._sim_device)
+
+    def set_force_feedback(self, force_x: float, force_y: float, force_z: float):
```
rwiltz: It would be nice if we generalized this and added it to the DeviceBase interface, so that this teleop device could be used in the exisiting teleop/record_demos scripts.

mingxueg-nv: @rwiltz thank you for your suggestion, I have added this interface into `DeviceBase`, the input is force from contact sensor, set to fx, fy, fz, and the output is customed to adapt to different force feedback devices.

mingxueg-nv:
Hi, @rwiltz , @kellyguo11, I resolved the most comments, please help review this PR again, thanks🙏

source/isaaclab/isaaclab/devices/device_base.py
```diff
@@ -77,6 +77,22 @@ def add_callback(self, key: Any, func: Callable):
         """
         raise NotImplementedError
 
+    def set_force_feedback(self, force_x: float, force_y: float, force_z: float) -> Any:
```
rwiltz: This is great, but can we make it a little more general since we expect more devices to be coming soon? Maybe we can add something like:
```
    # -----------------------------
    # Optional force/torque feedback
    # -----------------------------
    def push_force(self, forces: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:
        """Push one or more 3D force vectors to the device (optional; default no-op).

        Args:
            forces: Tensor of shape (N, 3) with forces [fx, fy, fz].
            names: Optional labels for each force channel (e.g., ["ee", "finger_1"]).
            frame: Frame of the vectors: "world" (default) or "device".

        Note:
            Devices that support haptics should override this method to forward
            forces to their hardware API. Devices that do not support haptics
            may ignore this call.
        """
        return  # no-op by default

    def push_wrench(self, wrenches: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:
        """Push one or more 6D wrenches to the device (optional; default no-op).

        Args:
            wrenches: Tensor of shape (N, 6) with [fx, fy, fz, tx, ty, tz].
            names: Optional labels for each wrench channel (e.g., ["ee", "finger_1"]).
            frame: Frame of the vectors: "world" (default) or "device".

        Note:
            Devices that support haptics should override this method. Devices that
            do not support haptics may ignore this call.
        """
        return  # no-op by default`
        ```
        
        Does that make sense?

Comment (rwiltz) [reply source/isaaclab/isaaclab/devices/device_base.py]: Also do we need a "clear_force" call to reset all the forces?

Comment (rwiltz) [reply source/isaaclab/isaaclab/devices/device_base.py]: If we just update your function to   ` def push_force(self, forces: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:` we can merge this and add the rest later incrementally

Comment (mingxueg-nv) [reply source/isaaclab/isaaclab/devices/device_base.py]: Hi, @rwiltz, thank you for your good point. I'll use `push_force` and test with office's haply device tomorrow - will push the commit once verified.

Comment (mingxueg-nv) [reply source/isaaclab/isaaclab/devices/device_base.py]: @rwiltz ,I've pushed the commit for the new interface modifications and have tested it with Haply, and it worked well. 
For the point of 'clear_force, I implemented  it in [reset()](https://github.com/isaac-sim/IsaacLab/pull/3873/files#diff-0843fe9192364d4bdb3933506c42cb909c628dbc384394a43402468dbee37c14R178).
If there is any other question, please let me know. Thank you.🙏

File: source/isaaclab/isaaclab/devices/device_base.py [hunk 13]
```diff
@@ -77,6 +77,38 @@ def add_callback(self, key: Any, func: Callable):
         """
         raise NotImplementedError
 
+    # -----------------------------
```
ooctipus: since only haply device is using force, I wonder maybe nicer to not introduce these two functions at device_base level.

scripts/demos/haply_teleoperation.py
```diff
@@ -0,0 +1,361 @@
...
+
+Prerequisites:
+    1. Install websockets package: pip install websockets
+    2. Have Haply SDK running and accessible via WebSocket
+    3. Connect Inverse3 and VerseGrip devices
+"""
+
+"""Launch Isaac Sim Simulator first."""
+
+import argparse
```
ooctipus, line 34: if the goal is to add a demo, 
it will be nice to create a demo entry in https://github.com/isaac-sim/IsaacLab/blob/main/docs/source/overview/showroom.rst

and show case this to broader community about this teleoperation, it will be nice to provide more detailed instruction in how to set it up. It will be nice to show something like this.
https://github.com/isaac-sim/IsaacLab/edit/main/docs/source/how-to/cloudxr_teleoperation.rst

ooctipus (commented):
The code looks good to me in first pass,
I think the main concern is that not many people has this kind of teleportation setup, updating documentation to be more instructive could be very helpful.

mingxueg-nv:
> ## Greptile Overview
> ### Greptile Summary
> This PR adds comprehensive Haply haptic device support for robot teleoperation with bidirectional force feedback. The implementation includes a WebSocket-based device controller (`HaplyDevice`), a demo script for Franka Panda teleoperation, and detailed documentation.
> 
> **Key Changes:**
> 
> * Added `HaplyDevice` class in `se3_haply.py` with 200Hz WebSocket communication for Inverse3 and VerseGrip devices
> * Extended `DeviceBase` with `push_force()` method for haptic feedback support
> * Created comprehensive teleoperation demo with workspace mapping, inverse kinematics, and contact force feedback
> * Added thorough documentation including setup guide, troubleshooting, and usage examples
> 
> **Architecture:** The system uses a background thread for WebSocket communication, thread-safe data caching with locks, and real-time force feedback streaming. Position data from Inverse3 is mapped to robot workspace, while VerseGrip provides orientation and button controls.
> 
> **Critical Issues Found:** Previous review comments identified several important issues that require attention before merging:
> 
> * Race condition in `se3_haply.py:207` where data is accessed outside lock protection (critical)
> * Invalid quaternion multiplication by sensitivity factor that breaks normalization (critical)
> * Demo script IK joint indexing mismatch that skips IK-computed joint values
> * Missing error handling for import failures in `WEBSOCKETS_AVAILABLE` flag
> 
> ### Confidence Score: 2/5
> * Multiple critical race conditions and logic errors need resolution before merging
> * Score of 2 reflects several unresolved critical issues from previous comments: (1) race condition in `advance()` where cached data is accessed after lock release, creating potential for corrupted reads during concurrent updates, (2) invalid quaternion scaling that violates unit quaternion constraint, (3) IK joint assignment logic error in demo that causes computed values to be ignored. While documentation is excellent and architecture is sound, these logical errors affect core functionality and data integrity.
> * `source/isaaclab/isaaclab/devices/haply/se3_haply.py` requires immediate attention for race condition fix (line 207) and quaternion normalization issue. Demo script `scripts/demos/haply_teleoperation.py` needs IK joint indexing correction.
> 
> ### Important Files Changed
> File Analysis
> 
> Filename	Score	Overview
> source/isaaclab/isaaclab/devices/haply/se3_haply.py	2/5	New Haply device controller with WebSocket communication and force feedback. Has critical race condition in `advance()` method (line 207) and invalid quaternion scaling that was flagged in previous comments.
> source/isaaclab/isaaclab/devices/device_base.py	5/5	Added `push_force()` method to base class for force feedback support. Clean implementation with proper default no-op behavior.
> docs/source/how-to/haply_teleoperation.rst	5/5	Comprehensive documentation for Haply device setup and usage. Well-structured with clear examples and troubleshooting section.
> ### Sequence Diagram
> ```mermaid
> sequenceDiagram
>     participant User as Operator
>     participant Haply as Haply Hardware<br/>(Inverse3 + VerseGrip)
>     participant SDK as Haply SDK<br/>(WebSocket Server)
>     participant Device as HaplyDevice<br/>(Python)
>     participant Demo as Demo Script
>     participant IK as IK Controller
>     participant Sim as Isaac Sim
>     participant Robot as Franka Panda
> 
>     User->>Haply: Move Inverse3
>     User->>Haply: Press buttons on VerseGrip
>     
>     Haply->>SDK: Stream position, orientation, buttons (200Hz)
>     SDK->>Device: WebSocket message (JSON)
>     
>     Note over Device: _websocket_loop() thread<br/>receives data
>     Device->>Device: Update cached_data<br/>(with data_lock)
>     
>     Demo->>Device: advance()
>     Device->>Device: Copy position, quaternion, buttons<br/>(with data_lock)
>     Device-->>Demo: torch.Tensor [x,y,z,qx,qy,qz,qw,btn_a,btn_b,btn_c]
>     
>     Demo->>Demo: apply_haply_to_robot_mapping()<br/>(transform workspace)
>     Demo->>IK: compute(target_pos, current_quat)
>     IK-->>Demo: joint_pos_des [6 joints]
>     
>     Demo->>Robot: Set joint positions
>     Demo->>Sim: Step simulation
>     
>     Sim->>Robot: Update physics
>     Robot->>Sim: Contact forces
>     
>     Demo->>Demo: Read ContactSensor data
>     Demo->>Device: push_force(contact_forces)
>     
>     Device->>Device: Update feedback_force<br/>(with force_lock)
>     Device->>SDK: WebSocket send force command
>     SDK->>Haply: Apply force to Inverse3
>     
>     Haply->>User: Feel haptic feedback
> ```
> 5 files reviewed, no comments
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=isaaclab_3873)

All flagged issues were resolved in recent commits, Greptile appears to be reviewing an outdated snapshot.

source/isaaclab/isaaclab/devices/device_base.py
mingxueg-nv: Hi, @ooctipus , following up on previous discussion with @rwiltz , I've removed `push_wrench` and kept the force interface. Currently added Haply device uses force-based feedback. If future devices require torque or other support, we can change it into a more general interface to DeviceBase and implement it in custom subclasses.

scripts/demos/haply_teleoperation.py
mingxueg-nv: Thank you @ooctipus . I have updated the documentation to include instructions on how to set up the Haply device and use the demo. Please let me know if any changes to the docs are needed.🙏

docs/source/how-to/haply_teleoperation.rst
```diff
@@ -0,0 +1,240 @@
...
+* **Isaac Lab Workstation**
+
+  * Ubuntu 22.04 or Ubuntu 24.04
+  * Hardware requirements for 200Hz physics simulation:
+
+    * CPU: 8-Core Intel Core i7 or AMD Ryzen 7 (or higher)
+    * Memory: 32GB RAM (64GB recommended)
+    * GPU: RTX 3090 or higher
+
+  * Network: Same local network as Haply devices for WebSocket communication
```
shanereetz, line 60: Is simply connecting the provided USB-C cable to the host computer sufficient for 200hz physics sim? 

That's the only cable that came with my Inverse3, other than power.

rwiltz (approved):

docs/source/_static/demos/haply_teleop_franka.jpg
ooctipus, line 1: Nice !, can we have the franka seat on the metal bracket? or rotate the table 90 degree?

scripts/demos/haply_teleoperation.py
```diff
@@ -0,0 +1,361 @@
...
+            sim.step()
+
+        scene.update(sim_dt)
+        count += 1
+
+        # get contact forces and apply force feedback
+        left_finger_forces = left_finger_sensor.data.net_forces_w[0, 0]
+        right_finger_forces = right_finger_sensor.data.net_forces_w[0, 0]
+        total_contact_force = (left_finger_forces + right_finger_forces) * 0.5
+        haply_device.push_force(total_contact_force.unsqueeze(0))
```
ooctipus: how does the haply_device understand this force is only on end effect or not on other bodies? say is it possible if I want to push force on link2 and end effector?(maybe link2 hits obstacle) or what if I have bi-mannual and want to apply this force two both end effectors?

If the goal is to add to base class, then we need to make sure its truly general :))

ooctipus (commented):
the docs looks good to me, 
My rest of discussion is around introducing push_force to base class.

My current fear is that this push_force function is only designed to haply case while being introduced to base class, we need to make sure the signature is general if the goal is to modify base class. (see my detailed comment below)

scripts/demos/haply_teleoperation.py
ooctipus: and what if device support fine grained force locations, will this api support sending force reading to device at precise location on the body as well?

rwiltz:
> the docs looks good to me, My rest of discussion is around introducing push_force to base class.
> 
> My current fear is that this push_force function is only designed to haply case while being introduced to base class, we need to make sure the signature is general if the goal is to modify base class. (see my detailed comment below)

@ooctipus I shared your concern as well and the base class function has been updated to:

`def push_force(self, forces: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:` Now there is a tensor with variable count of forces, with optional link names and reference frame for the implementing device to interpret.

ooctipus:
> > the docs looks good to me, My rest of discussion is around introducing push_force to base class.
> > My current fear is that this push_force function is only designed to haply case while being introduced to base class, we need to make sure the signature is general if the goal is to modify base class. (see my detailed comment below)
> 
> @ooctipus I shared your concern as well and the base class function has been updated to:
> 
> `def push_force(self, forces: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:` Now there is a tensor with variable count of forces, with optional link names and reference frame for the implementing device to interpret.

@rwiltz what about force location, some gloves have that right?
If device support fine grained force locations, will this api support sending force reading to device at precise location on the body as well?

below is physx set force api I am comparing for reference
```
    def set_external_force_and_torque(
        self,
        forces: torch.Tensor,
        torques: torch.Tensor,
        positions: torch.Tensor | None = None,
        body_ids: Sequence[int] | slice | None = None,
        env_ids: Sequence[int] | None = None,
        is_global: bool = False,
    ):
```

rwiltz:
> > > the docs looks good to me, My rest of discussion is around introducing push_force to base class.
> > > My current fear is that this push_force function is only designed to haply case while being introduced to base class, we need to make sure the signature is general if the goal is to modify base class. (see my detailed comment below)
> > 
> > 
> > @ooctipus I shared your concern as well and the base class function has been updated to:
> > `def push_force(self, forces: torch.Tensor, names: list[str] | None = None, frame: str = "world") -> None:` Now there is a tensor with variable count of forces, with optional link names and reference frame for the implementing device to interpret.
> 
> what about force location? If device support fine grained force locations, will this api support sending force reading to device at precise location on the body as well?
> 
> below is physx set force api I am comparing for reference
> 
> ```
>     def set_external_force_and_torque(
>         self,
>         forces: torch.Tensor,
>         torques: torch.Tensor,
>         positions: torch.Tensor | None = None,
>         body_ids: Sequence[int] | slice | None = None,
>         env_ids: Sequence[int] | None = None,
>         is_global: bool = False,
>     ):
> ```
That's a good point and thanks for providing the physix API for reference. My thinking was to keep the first force feedback API simple so that additional functions can be added as additional devices need them, but I agree adding an optional position arg might be helpful here. For torque I was thinking using a wrench format (force, torque), do you think we need a precise location there too or should we just add torque to push force and cover them in this function?

ooctipus:
>That's a good point and thanks for providing the physix API for reference. My thinking was to keep the first force feedback API simple so that additional functions can be added as additional devices need them, but I agree adding an optional position arg might be helpful here. For torque I was thinking using a wrench format (force, torque), do you think we need a precise location there too or should we just add torque to push force and cover them in this function?


@rwiltz 
Yeah thats my point, If **simple first example** force api is preferred, it maybe more natural to just introduce it in HaplyDevice, instead of adding to base_device. But if the goal is to modify base_device, then lets make sure it is general.  

I don't think mirroring physx's `set_torque_and_force` api to Device is 100% reasonable, and I am not super familiar with all the  teleoperation details. Though I recommend to be similar to physx ones as those api are proven to be solid with many usecases

rwiltz:
> > That's a good point and thanks for providing the physix API for reference. My thinking was to keep the first force feedback API simple so that additional functions can be added as additional devices need them, but I agree adding an optional position arg might be helpful here. For torque I was thinking using a wrench format (force, torque), do you think we need a precise location there too or should we just add torque to push force and cover them in this function?
> 
> @rwiltz Yeah thats my point, If **simple first example** force api is preferred, it maybe more natural to just introduce it in HaplyDevice, instead of adding to base_device. But if the goal is to modify base_device, then lets make sure it is general.
> 
> I don't think mirroring physx's `set_torque_and_force` api to Device is 100% reasonable, and I am not super familiar with all the teleoperation details. Though I recommend to be similar to physx ones as those api are proven to be solid with many usecases

ACK. For context, the motivation for adding to device base is to prevent special handling of teleop devices in places like teleop_se3_agent.py and record_demos.py. The sooner we converge on the basic building blocks the less tech debt we will need to deal with later (refactoring teleop device specific code out of those scripts was a chore :) ).

For the interim, do you think simply adding a position tensor to the proposed interface to be a sensible initial version for device_base?

ooctipus:
@rwiltz 
Yes I think that could work, you could also add the torque into the signature as well if you'd like: ))

docs/source/how-to/haply_teleoperation.rst
mingxueg-nv: Hi, @shanereetz, because we have implemented a cache data mechanism, so the default USB-C cable is enough.

docs/source/_static/demos/haply_teleop_franka.jpg
mingxueg-nv: @ooctipus thanks for the tip, I've updated the scene with the franka seat on the metal bracket.

mingxueg-nv:
> @rwiltz Yes I think that could work, you could also add the torque into the signature as well if you'd like: ))

Hi, @rwiltz @ooctipus , the specific location for force feedback also depends on where we've placed the contact sensors. In the demo, I added them to left and right fingers on the end-effector. 
So I've incorporated your suggestion by adding a position parameter to specify which sensor's force to feedback. (In the demo script, I'm currently using the default end-effector for now.)

source/isaaclab/isaaclab/devices/device_base.py
```diff
@@ -77,6 +77,32 @@ def add_callback(self, key: Any, func: Callable):
         """
         raise NotImplementedError
 
+    def push_force(
+        self,
+        forces: torch.Tensor | dict[str, torch.Tensor | list[float]],
+        names: list[str] | None = None,
+        frame: str = "world",
+        position: list[str] | None = None,
```
rwiltz: Position should be a tensor

scripts/demos/haply_teleoperation.py
mingxueg-nv: Resolved this in comment below.

source/isaaclab/isaaclab/devices/device_base.py
mingxueg-nv: Done, thanks!

rwiltz (approved):

source/isaaclab/isaaclab/devices/device_base.py
```diff
@@ -77,6 +77,29 @@ def add_callback(self, key: Any, func: Callable):
         """
         raise NotImplementedError
 
+    def push_force(
```
kellyguo11: is this API needed for any other devices at the moment? if not, may be better to place it into the Haply specific implementation?

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,399 @@
...
+    Note: All button logic (e.g., gripper control, reset, mode switching) should be
+    implemented in the application layer using the raw button states from advance().
+
+    Note:
+        Requires the Haply SDK to be running and accessible via WebSocket.
+        Install dependencies: pip install websockets
+
+    """
+
+    def __init__(self, cfg: HaplyDeviceCfg, retargeters=None):
```
kellyguo11: could we add a typing for retargeters?

source/isaaclab/isaaclab/devices/haply/se3_haply.py
```diff
@@ -0,0 +1,399 @@
...
+    ) -> None:
+        """Push force vector to Haply Inverse3 device.
+
+        Overrides DeviceBase.push_force() to provide force feedback for Haply Inverse3.
+        Forces are clipped to [-limit_force, limit_force] range for safety.
+
+        Args:
+            forces: Tensor of shape (N, 3) with forces [fx, fy, fz].
+            names: Optional labels (ignored for Haply Inverse3).
+            frame: Frame of the vectors (currently only "world" is supported).
```
kellyguo11: if only world is supported, do we really need this as an argument?

source/isaaclab/isaaclab/devices/device_base.py
rwiltz: This was my suggestion, I suspect this will be more widely used in the future, and I'm thinking it would be better to capture it in the interface so prevent fragmentation. For example, with this in place any env can push force to any device via the standard scripts (teleop_se3_agent.py, record_demos.py, etc). 

Also happy to be wrong here if you disagree.

kellyguo11: I see, just want to make sure we are 100% confident this will be implemented for other devices. otherwise, it'll be easier for us to maintain if we limit APIs to individual implementations, and we can also move them into base classes when we see the need for other devices.

rwiltz: The alternative is having each device define a different way to push force feedback which I don't think scales. This function should be general enough to scale to additional devices (exoskeleton gloves for example) which we know are on the horizon. If we are wrong, we can always modify the call, replace it with a deprecation, or add another function if it's orthogonal enough.

kellyguo11:
doc building job looks to be failing - https://github.com/isaac-sim/IsaacLab/actions/runs/19074390680/job/54639489929?pr=3873, please take a look

source/isaaclab/isaaclab/devices/device_base.py
```diff
@@ -77,6 +77,29 @@ def add_callback(self, key: Any, func: Callable):
...
+        self,
+        forces: torch.Tensor,
+        names: list[str] | None = None,
+        frame: str = "world",
+        position: torch.Tensor | None = None,
```
ooctipus: the description to position argument seems wrong, it is all the position on the local body not a mask of `which forces to use` , also what if I have 2 hands 5 envs, should it be 3D position tensor (num_envs, num_positions, 3)?

`
If we are wrong, we can always modify the call, replace it with a deprecation, or add another function if it's orthogonal enough.
`

I prefer we don't do this, this creates requirement that people needs to either take care of others design, or hack out new code path, in addition of breaking current users , If we are not certain, make experimental function into child class, If the reason is to scale, refactor them into single function once we have seen more example of use cases.

rwiltz: Seems like the consensus is to keep this out of DeviceBase. Let's revisit at a later time and keep push_force in the haply implementation @mingxueg-nv. Thanks @ooctipus and @kellyguo11 for your input.

source/isaaclab/isaaclab/devices/device_base.py
mingxueg-nv: Resolved in comment below.

source/isaaclab/isaaclab/devices/haply/se3_haply.py
mingxueg-nv: Updated

source/isaaclab/isaaclab/devices/haply/se3_haply.py
mingxueg-nv: Yes, you are right, I've remove `frame` argument.

source/isaaclab/isaaclab/devices/device_base.py
mingxueg-nv: Hi, @kellyguo11 , @rwiltz , @ooctipus , thank you for the discussion and suggestions. 
I have moved `push_force` out of `DeviceBase` and kept it in `HaplyDevice`. 
This is a case for getting started with force feedback. If we later add more devices supporting more types of feedback, we can further refine and extract the most general-purpose feedback interface.

mingxueg-nv:
Thanks @kellyguo11 for helping trigger CI. The new [docs issue](https://github.com/isaac-sim/IsaacLab/actions/runs/19190939447/job/54931553249) has been fixed, and the local docs build test has passed 
```
copying static files... done
copying extra files... done
dumping search index in English (code: en)... done
dumping object inventory... done
build succeeded.

The HTML pages are in _build/current.
```
Could you please help trigger CI again? Thanks so much.

ooctipus (approved):

PR #3920, opened 2025-11-03
Replaces torch_utils `set_seed` method with IsaacLab implementation

pascal-roth:
# Description

Replace torch_utils `set_seed` method with IsaacLab implementation

## Type of change

- Dependency removal

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/utils/seed.py
+++ b/source/isaaclab/isaaclab/utils/seed.py
@@ -3,9 +3,6 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-# This method has been adjusted from
-# https://github.com/isaac-sim/IsaacSim/blob/main/source/extensions/isaacsim.core.utils/python/impl/torch/maths.py
-
 import numpy as np
 import os
 import random


source/isaaclab/isaaclab/utils/seed.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def set_seed(seed: int, torch_deterministic: bool = False) -> int:
+    """set seed across modules"""
+    if seed == -1 and torch_deterministic:
+        seed = 42
+    elif seed == -1:
+        seed = np.random.randint(0, 10000)
+    print(f"Setting seed: {seed}")
+
```
Mayankm96:
```suggestion
    if seed == -1:
        seed = 42 if  torch_deterministic else np.random.randint(0, 10000)
```

source/isaaclab/isaaclab/utils/seed.py
```diff
@@ -0,0 +1,40 @@
...
+import numpy as np
+import os
+import random
+import torch
+
+import warp as wp
+
+
+def set_seed(seed: int, torch_deterministic: bool = False) -> int:
```
Mayankm96: NIT: To me a "set" usually should be combined with "get". Do you want to add something like "get_seed" as well? Or maybe we call this function `configure_seed`?

source/isaaclab/isaaclab/utils/seed.py
```diff
@@ -0,0 +1,40 @@
...
+    """set seed across modules"""
```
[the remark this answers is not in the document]
Mayankm96: Should seed also support `None` type which means it becomes "-1" behavior? I feel it is more inutitive than -1. Also are negative number generators valid? It seems ti is a yes for random and numpy at least.

source/isaaclab/isaaclab/utils/seed.py
pascal-roth: In that case, I would rename it! Only effect is a bigger break as also the name of the function changes

ooctipus:
`source/isaaclab_tasks/test/test_factory_environments.py` and `source/isaaclab_rl/test/test_rl_games_wrapper.py` did not time out locally.

pascal-roth:
@kellyguo11 and @Mayankm96, how do we want to handle the license as the function is taken from IsaacSIm 

https://github.com/isaac-sim/IsaacSim/blob/main/source/extensions/isaacsim.core.utils/python/impl/torch/maths.py

source/isaaclab/isaaclab/utils/seed.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# This method has been adjusted from
+# https://github.com/isaac-sim/IsaacSim/blob/main/source/extensions/isaacsim.core.utils/python/impl/torch/maths.py
+
```
pascal-roth:
```suggestion
```

PR #3923, opened 2025-11-03
Replaces IsaacSim `stage_utils` with IsaacLab `stage_utils`

pascal-roth:
# Description

Remove dependency on IsaacSim `stage_utils` for integration of new simulation engines like `newton`. 

## Type of change

- Dependency removal

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.48.4"
+version = "0.48.5"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"


ooctipus:
Seems like many things are failing :eyes:

source/isaaclab/isaaclab/sim/spawners/materials/visual_materials.py
```diff
@@ -12,7 +12,7 @@
 import omni.log
 from pxr import Usd
 
-from isaaclab.sim.utils import attach_stage_to_usd_context, clone, safe_set_attribute_on_usd_prim
+from isaaclab.sim.utils.stage import attach_stage_to_usd_context, clone, safe_set_attribute_on_usd_prim
```
ooctipus:
```suggestion
from isaaclab.sim.utils import attach_stage_to_usd_context, clone, safe_set_attribute_on_usd_prim
```

ooctipus:
I looked into the failing issue it seems like because we are combining `from utils import *` with cross package import. If you changes `import isaaclab.sim.utils.stage as stage_utils` to `from isaaclab.sim.utils import stage as stage_utils`

it will work again,
Explaination:

`from utils import *` rebinds names on the parent package, and in your case it ends up binding `isaaclab.sim.utils` to the module `isaaclab.sim.utils.utils` instead of the utils package.
`import isaaclab.sim.utils.stage as stage_utils` requires isaaclab.sim.utils to be a package (with __path__) so it can descend to stage, which fails when it’s actually the module.
`from isaaclab.sim.utils import stage as stage_utils` works because it imports the package and then accesses its stage attribute directly, bypassing that package-versus-module mix-up.


Personally I think that the root cause is import * in __init__.py files. It blindly rebinds names and can shadow the package object (isaaclab.sim.utils) with a module (utils.py), breaking dotted imports like import isaaclab.sim.utils.stage as ….

In the new rework we may need to consider to drop * import all together (at least for core)

pascal-roth:
@ooctipus I think for this one we are also ready to go

PR #3924, opened 2025-11-03
Replaces IsaacSim `prim_utils` with IsaacLab `prim_utils`

pascal-roth:
# Description

Remove dependency on IsaacSim `prim_utils` for integration of new simulation engines like `newton`. 

NOTE: this PR depends on #3923 

## Type of change

- Dependency removal

## Checklist


ooctipus:
Looks like tests are failing :eyes:

pascal-roth:
@ooctipus lets get that one merged too

ooctipus:
the package import for prim_utils doesn't work for the same reason as stage_utils, we used too many wild card
`import isaaclab.sim.utils.prims as prim_utils`

need to change to `from isaaclab.sim.utils import prims as prim_utils` so python can correctly recognized it as module

pascal-roth:
> the package import for prim_utils doesn't work for the same reason as stage_utils, we used too many wild card
> 
> `import isaaclab.sim.utils.prims as prim_utils`
> 
> 
> 
> need to change to `from isaaclab.sim.utils import prims as prim_utils` so python can correctly recognized it as module

That's not great, let's split up the utils/utils.py now to enable those imports again

pascal-roth:
behavior change for a util function, has to be checked before merg

source/isaaclab/isaaclab/sim/utils/logger.py
```diff
@@ -0,0 +1,50 @@
...
+
+"""Sub-module with logging utilities."""
+
+from __future__ import annotations
+
+import logging
+import time
+
+# import logger
+logger = logging.getLogger(__name__)
```
kellyguo11, line 14: should this replace _setup_logger in SimulationContext?

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -706,7 +1020,7 @@ def get_first_matching_child_prim(
     if not prim_path.startswith("/"):
         raise ValueError(f"Prim path '{prim_path}' is not global. It must start with '/'.")
     # get prim
-    prim = stage.GetPrimAtPath(prim_path)
+    prim: Usd.Prim = stage.GetPrimAtPath(prim_path)
```
kellyguo11: why is this type hinting needed here?

kellyguo11 (commented):
we are adding a lot of new code, do we have plans to add tests for all the new utils?

source/isaaclab/isaaclab/sim/utils/logger.py
ooctipus: I thought this only defines logger behaviror( ColoredFormatter and RateLimitFilter), but SimulationContext needs to setup the root logger that everyone uses. (I think this is what is happening?)

source/isaaclab/isaaclab/sim/utils/prims.py
ooctipus: no needed!

source/isaaclab/isaaclab/sim/utils/logger.py
kellyguo11: ah ok that makes sense

source/isaaclab/isaaclab/sim/utils/semantics.py
```diff
@@ -0,0 +1,276 @@
...
+    Args:
+        prim_path (str | None): This will check Prim path and its childrens' labels. If None, checks the whole stage.
+
+    Returns:
+        list[str]: Prim paths of meshes with no LabelsAPI applied.
+    """
+    prim_paths = []
+    stage = get_current_stage()
+    if stage is None:
+        carb.log_warn("Invalid stage, skipping label check")
```
kellyguo11: should we replace the carb.log calls with logging as well?

source/isaaclab/isaaclab/sim/utils/stage.py
```diff
@@ -194,7 +193,7 @@ def get_current_stage(fabric: bool = False) -> Usd.Stage:
 
     .. code-block:: python
 
-        >>> from isaaclab.sim.utils import stage as stage_utils
+        >>> import isaaclab.sim.utils.stage as stage_utils
         >>>
         >>> stage_utils.get_current_stage()
         Usd.Stage.Open(rootLayer=Sdf.Find('anon:0x7fba6c04f840:World7.usd'),
```
AntoineRichard, line 199: This needs to be able to return a fabric stage.

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -3,695 +3,1009 @@
...
+from collections.abc import Callable, Sequence
 from typing import TYPE_CHECKING, Any
 
 import omni
 import omni.kit.commands
+import omni.usd
+import usdrt
 from isaacsim.core.cloner import Cloner
 from isaacsim.core.version import get_version
+from omni.usd.commands import DeletePrimsCommand, MovePrimCommand
```
AntoineRichard, line 22: Should still have this here?

Mayankm96:
@pascal-roth Thanks for putting this together. I’m not fully convinced these one-line wrappers add much value, though. Since USD already has solid documentation, I wonder if maintaining parallel Isaac Sim functions might increase overhead without clear benefit. Would it make sense to remove these helpers and reference USD directly instead?

Personally, I’d prefer removing the wrappers altogether, but I’m fine handling that either here or in a follow-up MR depending on what keeps the review simplest.

ooctipus:
I agree is Mayank, my personal opinion is to remove this wrapper in follow up PR, as thorough removal require more thorough refactor, this pr is more likely what will lands in newton for the only purpose of remove kit dependency, and we make sure the main has something similar.

source/isaaclab/isaaclab/sim/utils/prims.py
ooctipus: These will require another follow up pr or lazy import to handle

source/isaaclab/isaaclab/sim/utils/stage.py
ooctipus: we will not able to get fabric stage without isaacsim or kit, usd stage is reliable

kellyguo11:
sounds like we are mostly ok with this PR to go in for now to be in alignment with the newton branch, but we should look into cleaning up all of the utilities in a follow-up PR. I think we should get this merged in then so that we also do not block the multi-mesh raycaster PR.

@Mayankm96 @ooctipus @AntoineRichard are you guys ok to give approval to merge this?

PR #3929, opened 2025-11-03
Pins python version of pre-commmit.yaml workflow

hhansen-bdai:
# Description

pre-commit CI runs have been failing (see [example](https://github.com/isaac-sim/IsaacLab/actions/runs/19017065661/job/54306430424)) due to an incompatibility between our specified flake version and python 3.14, as python3.14 has started to be installed by the setup-python GHA. This PR pins python to 3.12 in order to fix these failures.

## Type of change

### DIFF
--- a/.github/workflows/pre-commit.yaml
+++ b/.github/workflows/pre-commit.yaml
@@ -15,4 +15,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v3
+      with:
+        python-version: "3.12"
     - uses: pre-commit/action@v3.0.0


.github/workflows/pre-commit.yaml
```diff
@@ -15,4 +15,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v3
+      with:
+        python-version: "3.12"
```
Mayankm96, line 19: Does Isaac Sim 5.1 use Python 3.12 or 3.11? Might make sense to keep the version similar to that?

d-kleine: Yeah, IsaacSim 5 requires Python 3.11, other versions don't work:
https://docs.isaacsim.omniverse.nvidia.com/latest/installation/install_python.html

From my point-of-view, this needs to be fixed. Tagging @hhansen-bdai for heads-up

PR #3942, opened 2025-11-04
[SIM-950] Actuator Drive Model Improvements

### LINKED ISSUE #3909: [Proposal] Add support for actuator envelope models
### Proposal

Isaac Sim 5.0 introduced new actuator models that supported additional features. We should expose these in Isaac Lab to provide an interface for users to leverage them.

### Checklist

- [ ] I have checked that there is no similar issue in the repo (**required**)

### Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

bmccann-bdai:
This change implements the necessary changes to the articulation and actuator classes in order to configure the new actuator drive model including velocity and effort dependent constraints on motor actuation.

I have written tests against instantiation, and no other tests fail. I began looking into testing whether the drive model is being applied by physX. Doing so will require getting an isaacsim rather than isaaclab or physx view into the articulation in order to access the measured effort. @jtigue-bdai suggested we are trying to minimize dependence on the isaacsim api, preferring direct access to the physx layer. I decided to PR these changes without testing the physX level behavior. If we decide we want to introduce a dependency on the IsaacSim ArticulationView within the tests, I can either modify this PR, or create a follow on PR. 

Fixes #3909 

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/actuators/actuator_base.py
+++ b/source/isaaclab/isaaclab/actuators/actuator_base.py
@@ -8,13 +8,12 @@ from __future__ import annotations
 import torch
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, ClassVar
+from typing import ClassVar
 
 import isaaclab.utils.string as string_utils
 from isaaclab.utils.types import ArticulationActions
 
-if TYPE_CHECKING:
-    from .actuator_base_cfg import ActuatorBaseCfg
+from .actuator_base_cfg import ActuatorBaseCfg
 
 
 class ActuatorBase(ABC):
@@ -84,6 +83,18 @@ class ActuatorBase(ABC):
     For implicit actuators, the :attr:`velocity_limit` and :attr:`velocity_limit_sim` are the same.
     """
 
+    drive_model: torch.Tensor
+    """Three parameters for each joint/env defining the:
+       (1)  [:,:,0] speed_effort_gradient : float = 1 (default),
+       (2)  [:,:,1] maximum_actuator_velocity : float = torch.inf (default), and
+       (3)  [:,:,2] velocity_dependent_resistance : float = 1 (default)
+
+    which define velocity and effort dependent constraints on the motor's performance.
+
+    This feature is only implemented in IsaacSim v5.0.
+
+    The shape is (num_envs, num_joints, 3)."""
+
     stiffness: torch.Tensor
     """The stiffness (P gain) of the PD controller. Shape is (num_envs, num_joints)."""
 
@@ -124,6 +135,7 @@ class ActuatorBase(ABC):
         viscous_friction: torch.Tensor | float = 0.0,
         effort_limit: torch.Tensor | float = torch.inf,
         velocity_limit: torch.Tensor | float = torch.inf,
+        drive_model: torch.Tensor | tuple[float, float, float] = ActuatorBaseCfg.DriveModelCfg(),
     ):
         """Initialize the actuator.
 
@@ -160,6 +172,9 @@ class ActuatorBase(ABC):
                 If a tensor, then the shape is (num_envs, num_joints).
             velocity_limit: The default velocity limit. Defaults to infinity.
                 If a tensor, then the shape is (num_envs, num_joints).
+            drive_model: Drive model for the actuator including speed_effort_gradient, max_actuator_velocity, and
+                velocity_dependent_resistance in that order. Defaults to (0.0, torch.inf, 0.0).
+                If a tensor then the shape is (num_envs, num_joints, 3).
         """
         # save parameters
         self.cfg = cfg
@@ -187,19 +202,32 @@ class ActuatorBase(ABC):
             ("friction", friction),
             ("dynamic_friction", dynamic_friction),
             ("viscous_friction", viscous_friction),
+            ("drive_model", drive_model, 3),
         ]
-        for param_name, usd_val in to_check:
+        for param_name, usd_val, *tuple_len in to_check:
+            # check if the parameter requires a tuple or a single float
+            if len(tuple_len) > 0:
+                shape = (self._num_envs, self.num_joints, tuple_len[0])
+            else:
+                shape = (self._num_envs, self.num_joints)
+
             cfg_val = getattr(self.cfg, param_name)
-            setattr(self, param_name, self._parse_joint_parameter(cfg_val, usd_val))
+            setattr(self, param_name, self._parse_joint_parameter(cfg_val, usd_val, shape, param_name=param_name))
             new_val = getattr(self, param_name)
 
             allclose = (
-                torch.all(new_val == usd_val) if isinstance(usd_val, (float, int)) else torch.allclose(new_val, usd_val)
+                torch.all(new_val == usd_val)
+                if isinstance(usd_val, (float, int))
+                else (
+                    all([torch.all(new_val[:, :, i] == float(v)) for i, v in enumerate(usd_val)])
+                    if isinstance(usd_val, tuple)
+                    else torch.allclose(new_val, usd_val)
+                )
             )
             if cfg_val is None or not allclose:
                 self._record_actuator_resolution(
                     cfg_val=getattr(self.cfg, param_name),
-                    new_val=new_val[0],  # new val always has the shape of (num_envs, num_joints)
+                    new_val=new_val[0],
                     usd_val=usd_val,
                     joint_names=joint_names,
                     joint_ids=joint_ids,
@@ -303,13 +331,24 @@ class ActuatorBase(ABC):
 
         ids = joint_ids if isinstance(joint_ids, torch.Tensor) else list(range(len(joint_names)))
         for idx, name in enumerate(joint_names):
-            cfg_val_log = "Not Specified" if cfg_val is None else float(new_val[idx])
-            default_usd_val = usd_val if isinstance(usd_val, (float, int)) else float(usd_val[0][idx])
-            applied_val_log = default_usd_val if cfg_val is None else float(new_val[idx])
-            table.append([name, int(ids[idx]), default_usd_val, cfg_val_log, applied_val_log])
+            if len(new_val.shape) == 1:
+                cfg_val_log = "Not Specified" if cfg_val is None else float(new_val[idx])
+                default_usd_val = usd_val if isinstance(usd_val, (float, int)) else float(usd_val[0][idx])
+                applied_val_log = default_usd_val if cfg_val is None else float(new_val[idx])
+                table.append([name, int(ids[idx]), default_usd_val, cfg_val_log, applied_val_log])
+            else:
+                cfg_val_log = "Not Specified" if cfg_val is None else tuple(new_val[idx])
+                default_usd_val = usd_val if isinstance(usd_val, (tuple)) else tuple(usd_val[0][idx][:])
+                applied_val_log = default_usd_val if cfg_val is None else tuple(new_val[idx])
+                table.append([name, int(ids[idx]), default_usd_val, cfg_val_log, applied_val_log])
 
     def _parse_joint_parameter(
-        self, cfg_value: float | dict[str, float] | None, default_value: float | torch.Tensor | None
+        self,
+        cfg_value: tuple[float, ...] | dict[str, tuple[float, ...]] | float | dict[str, float] | None,
+        default_value: tuple[float, ...] | float | torch.Tensor | None,
+        expected_shape: tuple[int, ...] | None = None,
+        *,
+        param_name: str = "No name specified",
     ) -> torch.Tensor:
         """Parse the joint parameter from the configuration.
 
@@ -317,6 +356,10 @@ class ActuatorBase(ABC):
             cfg_value: The parameter value from the configuration. If None, then use the default value.
             default_value: The default value to use if the parameter is None. If it is also None,
                 then an error is raised.
+            expected_shape: The expected shape for the tensor buffer. Usually defaults to (num_envs, num_joints).
+
+        Kwargs:
+            param_name: a string with the parameter name. (Optional used only in exception messages).
 
         Returns:
             The parsed parameter value.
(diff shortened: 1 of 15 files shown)


scripts/get_omni_version.py
jtigue-bdai, line 1: did you add this?

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -76,6 +75,11 @@ class ActuatorBase(ABC):
     For implicit actuators, the :attr:`velocity_limit` and :attr:`velocity_limit_sim` are the same.
     """
 
+    drive_model: torch.Tensor
+    """Three parameters for each joint/env defining the speed_effort_gradient, maximum actuator velocity, and
```
gattra-rai: nitpick but formatting this comment to make it easier to identify the params and shape would be helpful

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -176,27 +184,44 @@ def __init__(
             ("friction", friction),
             ("dynamic_friction", dynamic_friction),
             ("viscous_friction", viscous_friction),
+            ("drive_model", drive_model, 3),
         ]
-        for param_name, usd_val in to_check:
+        for param_name, usd_val, *tuple_len in to_check:
+            # check if the parameter requires a tuple or a single float
+            if len(tuple_len) > 0:
```
gattra-rai, line 209: since the only param with a `tuple_len` in the list is `drive_model`, it might be simpler to just check `if param_name == "drive_model"`

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -176,27 +184,44 @@ def __init__(
             ("friction", friction),
             ("dynamic_friction", dynamic_friction),
             ("viscous_friction", viscous_friction),
+            ("drive_model", drive_model, 3),
```
gattra-rai, line 205: if you choose to create the `DriveModelParam` enum suggested in another comment, you could do `len(DriveModelParam.values)` instead of a hardcoded magic number

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -149,6 +154,9 @@ def __init__(
                 If a tensor, then the shape is (num_envs, num_joints).
             velocity_limit: The default velocity limit. Defaults to infinity.
                 If a tensor, then the shape is (num_envs, num_joints).
+            drive_model: Drive model for the actuator including speed_effort_gradient, max_actuator_velocity, and
```
gattra-rai, line 175: a `DriveModelParam` enum with a mapping of `param -> index` would avoid requiring the user to remember the indices of each drive model param

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -309,38 +349,89 @@ def _parse_joint_parameter(
...
                 param[:] = float(cfg_value)
+            elif isinstance(cfg_value, tuple):
+                # if tuple, ensure we expect a tuple for this parameter
+                if len(expected_shape) < 3:
+                    raise TypeError(
+                        f"Invalid type for parameter value: {type(cfg_value)} for parameter {param_name}"
+                        + f"actuator on joints {self.joint_names}. Expected float or dict, got tuple"
+                    )
+                # ensure the tuple is the correct length, and assign to the last tensor dimensions across all joints
+                if len(cfg_value) is expected_shape[2]:
```
gattra-rai: could invert this `if` and avoid the nesting:

```
if not len(cfg_value) ... :
    raise ValueError(...
    
for i, v in enumerate(cfg_value):
...
```

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -309,38 +349,89 @@ def _parse_joint_parameter(
...
                 raise TypeError(
                     f"Invalid type for parameter value: {type(cfg_value)} for "
-                    + f"actuator on joints {self.joint_names}. Expected float or dict."
+                    + f"actuator on joints {self.joint_names}. Expected tuple, float or dict."
                 )
         elif default_value is not None:
             if isinstance(default_value, (float, int)):
                 # if float, then use the same value for all joints
                 param[:] = float(default_value)
+            elif isinstance(default_value, tuple):
```
gattra-rai, line 428: i might be missing context here, but im seeing a lot of checks for `isinstance(..., tuple)` in this function and elsewhere. is it possible to convert the value to a `tensor` or `tuple` upfront and then reduce the branching code downstream?

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -309,38 +349,89 @@ def _parse_joint_parameter(
...
+                # if tuple, ensure we expect a tuple for this parameter
+                if len(expected_shape) < 3:
+                    raise TypeError(
+                        f"Invalid default type for parameter value: {type(default_value)} for "
+                        + f"actuator on joints {self.joint_names}. Expected float or dict, got tuple"
+                    )
+                # ensure the tuple is the correct length, and assign to the last tensor dimensions across all joints
+                if len(default_value) is expected_shape[2]:
```
gattra-rai: same as above, consider inverting this `if`

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -1594,6 +1633,12 @@ def _create_buffers(self):
         self._data.joint_pos_limits = self._data.default_joint_pos_limits.clone()
         self._data.joint_vel_limits = self.root_physx_view.get_dof_max_velocities().to(self.device).clone()
         self._data.joint_effort_limits = self.root_physx_view.get_dof_max_forces().to(self.device).clone()
+        if int(get_version()[2]) >= 5:
```
gattra-rai: please use a constant instead of magic numbers

gattra-rai (commented):
looks good!

source/isaaclab/isaaclab/actuators/actuator_base.py
bmccann-bdai: I was trying to do this through the NamedTuple class. At least for the ActuatorBaseCfg.DriveModelCfg, they can treat the cfg either as a tuple, or a dataclass with named attributes.

source/isaaclab/isaaclab/actuators/actuator_base.py
bmccann-bdai: There might be a way to extract it from the DriveModelCfg. I know it automatically imposes the tuple[float,float,float] constraint. I agree, the magic number isn't ideal.

source/isaaclab/isaaclab/actuators/actuator_base.py
bmccann-bdai: The goal was to implement a more general pattern which other related parameters can use.

source/isaaclab/isaaclab/actuators/actuator_base.py
bmccann-bdai: Unfortunately not, this is where we parse the configuration file, and a design constraint was that we use simple types in the configuration classes. I originally proposed using tensors directly, but they asked me to stay with vanilla types.

source/isaaclab/isaaclab/assets/articulation/articulation.py
bmccann-bdai: I was mirroring the pattern from line 1621.

source/isaaclab/isaaclab/actuators/actuator_base.py
gattra-rai: oh yeah, not suggesting changing anything about the config class itself. my thinking was, if the output/result of the function is the same, just with branching logic based on the type, it might be simpler to convert the `tuple` to a `tensor` (or vice versa) at the top of the function and then scrap the branching logic.

bmccann-bdai: I see. There's an additional complication which is that here `isinstance(tensor)` is taken as shorthand for the complete tensor (across envs and joints), whereas the tuple is treated as a single entry to be broadcast across envs and joints. So if I were to convert to a tensor, I would still need to branch (maybe with a flag or something) so I can correctly check the shape, and broadcast the values correctly.

What might work is to create a more general base class for tuple parameters. The NamedTuple is a good start, but maybe something that 1) exposes an enum to directly map indices (as you suggest), and 2) has something like a to_tensor() call to support casting to something that can be broadcast like a slice, which can't be done with tuples. At least then the branching logic would be more concise, and look more like a simple assignment than admittedly opaque nested loops.

My concern was creating code that relies on a more complex object, and forces other code to rely on a class when a tuple will do. We could do something like cast the tuple to the tuple parameter class early in the function as you suggest. At least then the exposed api could use either a tuple, or the convenience object depending on the consumer's preference.

bmccann-bdai:
I'm getting a test failure:
```source/isaaclab/test/sim/test_utils.py .....F.                           [100%]

=================================== FAILURES ===================================
____________________________ test_resolve_prim_pose ____________________________
source/isaaclab/test/sim/test_utils.py:254: in test_resolve_prim_pose
    np.testing.assert_allclose(quat, rand_quats[i, 2], atol=1e-3)
/isaac-sim/kit/python/lib/python3.11/contextlib.py:81: in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
E   AssertionError: 
E   Not equal to tolerance rtol=1e-07, atol=0.001
E   
E   Mismatched elements: 4 / 4 (100%)
E   Max absolute difference: 0.00692904
E   Max relative difference: 0.03919505
E    x: array([ 0.615105,  0.438734,  0.169855, -0.632699])
E    y: array([ 0.61288 ,  0.432324,  0.176784, -0.637355])
- generated xml file: /workspace/isaaclab/tests/test-reports-test_utils.py.xml -
=========================== short test summary info ============================
FAILED source/isaaclab/test/sim/test_utils.py::test_resolve_prim_pose - Asser...
```

It seems unrelated to my changes. I will rerun, but if it continues to fail, what is the best action to take? My usual approach would be to mark it as expected to fail or to skip the test. Otherwise, I will have to bypass merge the MR. What's the best way forward?

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -309,38 +356,87 @@ def _parse_joint_parameter(
...
+                        + f" {expected_shape[2]}"
+                    )
+                for i, v in enumerate(cfg_value):
+                    param[:, :, i] = float(v)
             elif isinstance(cfg_value, dict):
                 # if dict, then parse the regular expression
-                indices, _, values = string_utils.resolve_matching_names_values(cfg_value, self.joint_names)
-                # note: need to specify type to be safe (e.g. values are ints, but we want floats)
-                param[:, indices] = torch.tensor(values, dtype=torch.float, device=self._device)
+                indices, j, values = string_utils.resolve_matching_names_values(cfg_value, self.joint_names)
```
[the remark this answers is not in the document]
bmccann-bdai: This is probably a useful data point for the Greptile devs.

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -1758,7 +1805,7 @@ def _process_actuators_cfg(self):
             if int(get_version()[2]) >= 5:
                 self._data.default_joint_dynamic_friction_coeff[:, actuator.joint_indices] = actuator.dynamic_friction
                 self._data.default_joint_viscous_friction_coeff[:, actuator.joint_indices] = actuator.viscous_friction
-
+                self._data.default_joint_drive_model_parameters[:, actuator.joint_indices] = actuator.drive_model
```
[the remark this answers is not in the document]
bmccann-bdai: This is also probably an interesting data point for the greptile devs.

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -309,38 +356,87 @@ def _parse_joint_parameter(
...
+                    # note: need to specify type to be safe (e.g. values are ints, but we want floats)
+                    param[:, indices] = torch.tensor(values, dtype=torch.float, device=self._device)
+                # otherwise, we expect tuples
+                else:
+                    # We can't directly assign tuples to tensors, so iterate through them
+                    for i, v in enumerate(values):
+                        # Raise an exception if the tuple is the incorrect length
+                        if len(v) is not expected_shape[2]:
+                            raise ValueError(
+                                f"Invalid tuple length for parameter {param_name} on joint {j[i]} at index"
```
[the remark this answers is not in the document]
bmccann-bdai: see above.

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -325,38 +368,87 @@ def _parse_joint_parameter(
...
+                        + f"actuator on joints {self.joint_names}. Expected float or dict, got tuple"
+                    )
+                # ensure the tuple is the correct length, and assign to the last tensor dimensions across all joints
+                if not len(default_value) == expected_shape[2]:
+                    raise ValueError(
+                        f"Invalid tuple length for parameter {param_name}, got {len(default_value)}, expected"
+                        + f" {expected_shape[2]}"
+                    )
+                for i, v in enumerate(default_value):
+                    param[:, :, i] = float(v)
```
jtigue-bdai, line 442: I believe you can do 
param[:] = torch.full((n, m, 3), fill_value, dtype=torch.float32)

source/isaaclab/isaaclab/actuators/actuator_cfg.py
jtigue-bdai, line 1: why were these changes necessary?

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -48,6 +50,18 @@ class ActuatorBaseCfg:
 
     """
 
+    """Optional (min v5) settings to the drive model capturing performance envelope velocity-effort dependence.
```
jtigue-bdai:
```suggestion
    """Optional settings to the drive model capturing performance envelope velocity-effort dependence. Requires IsaacSim>5.0.0.
```

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -796,6 +795,47 @@ def write_joint_effort_limit_to_sim(
         # set into simulation
         self.root_physx_view.set_dof_max_forces(self._data.joint_effort_limits.cpu(), indices=physx_env_ids.cpu())
 
+    def write_joint_drive_model_to_sim(
+        self,
+        drive_params: torch.Tensor | tuple[float, float, float] | None = None,
+        joint_ids: Sequence[int] | slice | None = None,
+        env_ids: Sequence[int] | None = None,
+    ):
+        """Write parameters (in additional to max forces) necessary to model motor drive performance envelopes.
```
jtigue-bdai, line 804:
```suggestion
        """Write parameters (in additional to max joint effort) necessary to model motor drive performance envelopes. 

        Maximum actuator effort must also be set with :meth:`Articulation.write_joint_effort_limit_to_sim` to completely parameterize the drive performance envelope. 
```

source/isaaclab/isaaclab/assets/articulation/articulation.py
```diff
@@ -796,6 +795,47 @@ def write_joint_effort_limit_to_sim(
...
+        motor in the articulation. See: https://docs.omniverse.nvidia.com/kit/docs/omni_physics/107.3/_downloads/f44e831b7f29e7c2ec8e3f2c54418430/drivePerformanceEnvelope.pdf
+
+        Args:
+            drive_params: The drive parameters. Shape is (len(env_ids), len(joint_ids), 3). The final three dimensions
+                          reflect the speed-effort gradient, max actuator velocity, and velocity dependent resistance
+                          respectively.
+            joint_ids: The joint indices to set the joint torque limits for. Defaults to None (all joints).
+            env_ids: The environment indices to set the joint torque limits for. Defaults to None (all environments).
+        """
+        if int(get_version()[2]) >= 5:
```
jtigue-bdai: should add a warning output if version is less than 5.

source/isaaclab/isaaclab/assets/articulation/articulation_cfg.py
```diff
@@ -69,3 +70,25 @@ class InitialStateCfg(AssetBaseCfg.InitialStateCfg):
...
+        the necessary attributes exist and are available to be configured by the usual configuration parser logic.
+
+        .. Note: Importantly, configuring the drive model on any actuator in the articulation will enable the api on _all_
+           actuators in the articulation (with some default values). It is not currently possible to enable the api on
+           specific joint indices. This may lead to unexpected impacts on drive behavior or frame rate."""
+
+        if self.actuators is not None:
+            if len(self.actuators.keys()) > 0:
+                for act_cfg in self.actuators.values():
+                    if act_cfg.class_type == actuator_pd.ImplicitActuator:
```
jtigue-bdai, line 88: any actuator should be able to use this right?

source/isaaclab/isaaclab/assets/articulation/articulation_data.py
```diff
@@ -384,6 +388,9 @@ def update(self, dt: float):
     joint_effort_limits: torch.Tensor = None
     """Joint maximum effort provided to the simulation. Shape is (num_instances, num_joints)."""
 
+    joint_drive_model_parameters = None
+    """Joint drive model parameters provided to the simulation. Shape is (num_instances, num_joints, 3)."""
+
```
jtigue-bdai, line 393:
```suggestion
    joint_drive_model_parameters: torch.Tensor = None
    """Joint drive model parameters provided to the simulation. Shape is (num_instances, num_joints, 3). The last indices reflect
       the speed_effort_gradient, the max_actuator_velocity, and the velocity_dependent_resistance respectively."""

```

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -678,7 +680,11 @@ def modify_joint_drive_properties(
     # set into USD API
     for attr_name, attr_value in cfg.items():
         attr_name = cfg_to_usd_map.get(attr_name, attr_name)
-        safe_set_attribute_on_usd_schema(usd_drive_api, attr_name, attr_value, camel_case=True)
+        if attr_name is not None:
+            safe_set_attribute_on_usd_schema(usd_drive_api, attr_name, attr_value, camel_case=True)
+
+    if cfg["enable_physx_perf_env"]:
+        prim.ApplyAPI(PERF_ENV_API, UsdPhysics.Tokens.angular)
```
jtigue-bdai, line 687: have we gotten confirmation on angular vs prismatic?

source/isaaclab/test/actuators/test_drive_model_properties.py
```diff
@@ -0,0 +1,69 @@
...
+
+import pytest
+
+from isaaclab.actuators import IdealPDActuatorCfg
+
+
+@pytest.mark.parametrize("num_envs", [1, 2])
+@pytest.mark.parametrize("num_joints", [1, 2])
+@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
+def test_init_drive_model(num_envs, num_joints, device):
```
jtigue-bdai, line 25: NIT: docstring

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -84,6 +83,18 @@ class ActuatorBase(ABC):
...
+       (1)  [:,:,0] speed_effort_gradient : float = 1 (default),
+       (2)  [:,:,1] maximum_actuator_velocity : float = torch.inf (default), and
+       (3)  [:,:,2] velocity_dependent_resistance : float = 1 (default)
+
+    which define velocity and effort dependent constraints on the motor's performance.
+
+    This feature is only implemented in IsaacSim v5.0.
+
+    The shape is (num_envs, num_joints, 3)."""
+
```
jtigue-bdai, line 97: should we move this to the ImplicitActuatorCfg since it only works with the ImplicitActuator PD gains

scripts/get_omni_version.py
bmccann-bdai: Yea, I had it locally as a hack to get the version string. There's probably a better way, and it didn't seem appropriate to commit.

source/isaaclab/isaaclab/actuators/actuator_cfg.py
bmccann-bdai: It was just a (correct) greptile suggestion which I thought landed with the circular import PR, but must have been committed to this branch I had based on top. It's just a little cleaner, without changing what I was doing.

The original changes were necessary to avoid changing the user facing api when I moved the cfg files around.

source/isaaclab/isaaclab/assets/articulation/articulation_cfg.py
bmccann-bdai: No. It is only available with implicit_pd actuators.

source/isaaclab/isaaclab/sim/schemas/schemas.py
bmccann-bdai: Somewhere in the docs suggested the api can be applied to either. I'm worried I may not be accounting for units correctly in the prismatic case. Nor have I written any tests. We might want to talk about it, and maybe ask the NV devs for clarification.

source/isaaclab/isaaclab/actuators/actuator_base.py
bmccann-bdai: I tend to agree. It seems dishonest to imply the feature is more widely available. We might want to do a larger refactor before landing this.

kellyguo11:
@AntoineRichard @Mayankm96 could you give this one a review pls? We can get this into develop after branching off.

source/isaaclab/isaaclab/assets/articulation/articulation_data.py
AntoineRichard, line 1: A couple of things had changed on the articulation data with the new develop branch. This would have to be addressed.

source/isaaclab/isaaclab/assets/articulation/articulation_cfg.py
AntoineRichard, line 1: That's a wee bit violent here. Is there no way to have this in the articulation itself rather than its config? I even feel this could be in the actuator's config?

AntoineRichard (commented):
In general I'm a bit torn on this one. I don't know if we need a named tuple explicitly. Or if we could do things the same way they are done for the new friction parameters.

AntoineRichard:
Hi @bmccann-bdai — thank you so much for opening this PR, and for taking the time to contribute to Isaac Lab! 🙏

| | |
|---|---|
| **Opened** | 2025-11-04 (about 10 months ago) |
| **Last commit on the branch** | 2026-01-16 |
| **Last activity from the author** | about 7 months ago |
| **Review status** | Reviewed by AntoineRichard, gattra-rai, jtigue-bdai, but it was never brought to a conclusion |
| **Size** | 3 commit(s), 15 file(s) changed, +817 / -77 |

It was picked up by the sweep because it targets `main` instead of `develop`, and it has been open for about 10 months. It was then put in the "close" bucket because the author has been silent for about 7 months — which is the signal we used to tell apart pull requests that are still being worked on from ones that have genuinely been set aside.

---

PR #3975, opened 2025-11-07
Adds disable sleeping global flag for physx

kellyguo11:
# Description

Introducing new PhysxCfg flag disable_sleeping to disable sleeping on a global level. This flag will now override per-object sleeping settings if set to True. If sleeping is desired for any object in the scene, ensure this flag is set to False. We are setting this to True by default to keep a consistent behavior between CPU and GPU simulation, since sleeping is not supported by PhysX on the GPU pipeline. This has also previously caused confusion for some users where the contact forces reported on CPU and GPU simulation were different due to sleeping.

## Type of change

- Documentation update

## Checklist


source/isaaclab/test/sim/test_simulation_context.py
```diff
@@ -146,3 +146,109 @@ def test_zero_gravity():
...
+    # Check that the attribute exists and is set to True
+    disable_sleeping_attr = physics_scene_prim.GetAttribute("physxSceneAPI:disableSleeping")
+    assert disable_sleeping_attr.IsValid()
+    assert disable_sleeping_attr.Get() is True
+
+
+@pytest.mark.isaacsim_ci
+def test_disable_sleeping_false_with_cpu():
+    """Test that disable_sleeping can be set to False when using CPU simulation."""
+    from pxr import PhysxSchema
```
kellyguo11:
```suggestion
```

source/isaaclab/test/sim/test_simulation_context.py
```diff
@@ -146,3 +146,109 @@ def test_zero_gravity():
     gravity_dir, gravity_mag = sim.get_physics_context().get_gravity()
     gravity = np.array(gravity_dir) * gravity_mag
     np.testing.assert_almost_equal(gravity, cfg.gravity)
+
+
+@pytest.mark.isaacsim_ci
+def test_disable_sleeping_setting():
+    """Test that the disable_sleeping PhysX setting is applied correctly."""
+    from pxr import PhysxSchema
```
kellyguo11:
```suggestion
```

ooctipus (approved):

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -180,6 +180,26 @@ class PhysxCfg:
...
+    This flag disables sleeping for all objects in the scene, overriding any sleeping settings on individual bodies.
+    If sleeping is required on any individual body, this flag must be set to False.
+
+    .. warning::
+
+        If :attr:`disable_sleeping` is set to False and the GPU pipeline is enabled, PhysX will issue an error
+        message and fail scene creation as this is not supported.
+    """
+
+    enable_external_forces_every_iteration: bool = False
```
Mayankm96: Doesn't this flag only work when TGS is used? Also if this flag is enabled, then within the asset classes, I wonder if we still need to set forces at simulation steps. Previously those were getting cleared which wasn't a desired effect.

By iteration: do you mean solver iteration or sim.step? The document is clear about it.

Mayankm96: Based on here, this flag is only valid for TGS: https://docs.omniverse.nvidia.com/kit/docs/omni_usd_schema_physics/106.0/class_physx_schema_physx_scene_a_p_i.html#a0a8da8575a2180564cede1ddf4f5ab1a

Mayankm96:
> enable_external_forces_every_iteration is setting for the TGS solver the applies external forces in lockstep with every solver step. This flag was previously not set, which we are keeping as the default behavior with setting it to False. However, this can introduce noisy joint velocities as some users have noticed. To avoid these artifacts, the flag can be set to True at the cost of minor performance hits (we measured around 4% for a scene with 2048 humanoids)

If the values are more correct, I think we should keep this flag as True by default. Most usecases where sim-to-real is concerned, bad joint velocities affects the deployment. In my experience, many users remove joint velocities from observations because of the jitteriness in the deployed behavior.

We should inform the users that if they care about performance, then they can disable these flags to get "approximated" values, but the default should try to remain a "correct" value.

kellyguo11:
Keeping external force flag independent in #3989

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -180,6 +180,18 @@ class PhysxCfg:
 
     """
 
+    disable_sleeping: bool = True
+    """Disable sleeping for all objects in the physics scene on a global level. Default is True.
+
+    This flag disables sleeping for all objects in the scene, overriding any sleeping settings on individual bodies.
```
Mayankm96, line 186: We should update the schema docs that sleep threshold is ignored if this flag is set. Also does this flag impact performance? Could we set it to None in which case this flag is not set and prior behavior is maintained?

AntoineRichard:
Hi @kellyguo11 — thanks for putting this one up! 🙏

We're doing a cleanup pass over the Isaac Lab PR backlog, which had grown past 400 open pull requests, and we're closing out the ones that have gone quiet so the queue is reviewable again.

| | |
|---|---|
| **Opened** | 2025-11-07 (about 10 months ago) |
| **Last commit on the branch** | 2025-11-11 |
| **Last activity from the author** | about 10 months ago |
| **Review status** | Approved at the time (Mayankm96, ooctipus), but it was never merged and has since gone stale |
| **Size** | 9 commit(s), 5 file(s) changed, +97 / -1 |

It was picked up by the sweep because it targets `main` instead of `develop`, and it has been open for about 10 months. It was then put in the "close" bucket because the author has been silent for about 10 months — which is the signal we used to tell apart pull requests that are still being worked on from ones that have genuinely been set aside.

No judgement on the change itself — this is purely backlog hygiene.

**If this is still wanted, please reopen it or re-submit against `develop`.** 💚

---

PR #3989, opened 2025-11-10
Fixes noisy velocities around limits

AntoineRichard:
# Description

Enables a new PhysX flag to help mitigate noisy velocities. 

## Type of change

- Documentation update

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/sim/simulation_context.py
+++ b/source/isaaclab/isaaclab/sim/simulation_context.py
@@ -814,7 +814,7 @@ class SimulationContext(_SimulationContext):
 
         if self.cfg.physx.solver_type == 1:
             if not self.cfg.physx.enable_external_forces_every_iteration:
-                omni.log.warn(
+                logger.warning(
                     "The `enable_external_forces_every_iteration` parameter in the PhysxCfg is set to False. If you are"
                     " experiencing noisy velocities, consider enabling this flag. You may need to slightly increase the"
                     " number of velocity iterations (setting it to 1 or 2 rather than 0), together with this flag, to"


source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -99,6 +99,19 @@ class PhysxCfg:
         Enabling this flag may lead to incorrect contact forces report from the contact sensor.
     """
 
+    enable_external_forces_every_iteration: bool = False
+    """Enable/disable external forces every position iteration in the TGS solver. Default is False.
```
kellyguo11, line 103: I think @Mayankm96's preference was to keep this to True so that by default we always get the most accurate data. If users need better perf, they can set it to False.

I don't feel too strongly either way. My concern was mostly that setting it to True by default might break existing policies if they use joint velocities, but data accuracy is also important.

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -792,6 +792,19 @@ def _set_additional_physx_params(self):
         physx_prim.CreateAttribute("physxScene:solveArticulationContactLast", Sdf.ValueTypeNames.Bool).Set(
             self.cfg.physx.solve_articulation_contact_last
         )
+        # -- Enable external forces every iteration, helps improve the accuracy of velocity updates.
+
+        if self.cfg.physx.solver_type == 1:
+            if not self.cfg.physx.enable_external_forces_every_iteration:
+                omni.log.warn(
```
kellyguo11: should we move away from omni.log?

kellyguo11:
Copying over some comments from Mayank from an earlier PR:

> if this flag is enabled, then within the asset classes, I wonder if we still need to set forces at simulation steps. Previously those were getting cleared which wasn't a desired effect.

source/isaaclab/isaaclab/sim/simulation_cfg.py
AntoineRichard: That was my concern too. It may change the default behavior of the sim, and that may be considered a breaking change? I'd say for this version we keep it this way, and for 3.0 we can make it a default.

source/isaaclab/isaaclab/sim/simulation_context.py
AntoineRichard: Should I wait for Pascal's PR to be in?

AntoineRichard:
I would merge as is and make the breaking change with 3.0 @Mayankm96 would you be ok with this?

AntoineRichard:
> Copying over some comments from Mayank from an earlier PR:
> 
> > if this flag is enabled, then within the asset classes, I wonder if we still need to set forces at simulation steps. Previously those were getting cleared which wasn't a desired effect.

I checked with Gordon, and this flag does NOT allow forces to be conserved between solver steps. What this does is that rather than applying external forces in one go on the first solver **position iteration**, it applies it (divided by the number of solver position iterations) for **each solver position iteration**.

source/isaaclab/isaaclab/sim/simulation_cfg.py
kellyguo11: ok yup that was my thought as well. let's keep it False for now until 3.0.

source/isaaclab/isaaclab/sim/simulation_context.py
kellyguo11: it has landed :)

kellyguo11 (approved):
should be good to go once omni.log is updated

source/isaaclab/isaaclab/sim/simulation_context.py
jtigue-bdai: may want to add the intention to "This feature will be True by default in a future release."

AntoineRichard:
> should be good to go once omni.log is updated

Updated

PR #3992, opened 2025-11-10
Adds APIs to Isaac Lab Mimic for supporting loco-manipulation datagen

peterd-NV:
# Description

This PR adds new APIs to Isaac Lab Mimic to expand support for loco-manipulation data generation.

It adds:

- Processing for body end effector to treat a robot's base as an eef during Mimic data generation. This enables the use of the same Mimic annotation and subtask interface to enable lower body movement.
- An optional way to cleanly add custom recorders for Mimic data generation (useful for when users want to record beyond the action/state data provided by Isaac Lab's default recorder).
- Interface for enabling a navigation p-controller during Mimic data generation for robot's with mobile bases.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,7 @@
 Changelog
 ---------
 
+
 0.48.9 (2025-11-21)
 ~~~~~~~~~~~~~~~~~~~
 
@@ -27,6 +28,7 @@ Changed
 ^^^^^^^
 
 * Changed import from ``isaaclab.sim.utils`` to ``isaaclab.sim.utils.stage`` to properly propagate the Isaac Sim stage context.
+
 
 0.48.6 (2025-11-18)
 ~~~~~~~~~~~~~~~~~~~


source/isaaclab/isaaclab/envs/manager_based_rl_mimic_env.py
```diff
@@ -156,3 +156,18 @@ def serialize(self):
         and used in utils/env_utils.py.
         """
         return dict(env_name=self.spec.id, type=2, env_kwargs=dict())
+
+    def get_navigation_state(self, env_ids: Sequence[int] | None = None) -> dict[str, torch.Tensor]:
```
xyao-nv, line 167: Can you make it as optional, ideally both in doc and code? It will confuse users about why navigation appears in this mimic_env.py. At least at code level, it shall specify it's not needed for tabletop manip. You can define a decorator `@optional_method`.
```
def optional_method(func):
    """
    Decorator to mark a method as optional for subclasses.
    """
    func.__optional__ = True
    return func

class Base:
    @optional_method
    def get_navigation_state(self):
        print("Base implementation")
```

source/isaaclab_mimic/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+
+1.0.16 (2025-11-10)
+
+Added
+^^^^^
+
+* Add body end effector to Mimic data generation to enable loco-manipulation data generation when a navigation p-controller is provided.
```
xyao-nv, line 11: If possible, considering what else needed to go with any feedback controller(i.e. P vs PD vs PID). It sounds to me this Mimic eef interface shall be agnostic to what controller it is.

source/isaaclab_mimic/isaaclab_mimic/datagen/data_generator.py
```diff
@@ -880,8 +887,53 @@ async def generate(  # noqa: C901
                 generated_actions.extend(exec_results["actions"])
                 generated_success = generated_success or exec_results["success"]
 
+            # Get the navigation state
+            if self.env_cfg.datagen_config.use_navigation_p_controller:
+                processed_nav_subtask = False
+                navigation_state = self.env.get_navigation_state(env_id)
+                is_navigating = navigation_state["is_navigating"]
+                navigation_goal_reached = navigation_state["navigation_goal_reached"]
```
xyao-nv, line 898: Can you add assertion (`navigation_state is not None` && keys in the dict) before accessing the keys here?

xyao-nv (approved):

source/isaaclab/isaaclab/envs/manager_based_rl_mimic_env.py
peterd-NV: Added

source/isaaclab_mimic/docs/CHANGELOG.rst
peterd-NV: This is a good point. From the Mimic side it doesn't need any info as it assumes the controller will handle everything and just provide a signal to indicate if its 1) is navigating, and 2) has reached goal. 

However, I've updated the naming in the config files to use_navigation_controller instead of use_navigation_p_controller to make this clear that any controller can be supported.

source/isaaclab_mimic/isaaclab_mimic/datagen/data_generator.py
peterd-NV: Added

source/isaaclab_mimic/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,15 @@
 Changelog
 ---------
 
+
+1.0.16 (2025-11-10)
```
kellyguo11, line 5:
```suggestion
1.0.16 (2025-11-10)
~~~~~~~~~~~~~~~~~~~
```

source/isaaclab_mimic/isaaclab_mimic/datagen/data_generator.py
```diff
@@ -7,10 +7,13 @@
 Base class for data generator.
 """
 import asyncio
+import copy
 import numpy as np
 import torch
 from typing import Any
 
+import omni.log
```
kellyguo11: just a heads up we have removed references of omni.log in our codebase as part of the OV decoupling efforts. might not be as import for mimic if this extension still requires OV dependency, but otherwise, may be good to also move away from omni.log and switch to python's logging module instead.

peterd-NV: thanks for the heads up. I've switched from omni.log to python logging

PR #4008, opened 2025-11-12
Adds greptile.json file for specifying triggers for Greptile on PRs

AmmarYAhmed:
Adds greptile.json file for specifying triggers for Greptile on PRs
This should avoid Greptile from getting triggered on every commit for a PR.
It should still be able to get triggered manually by @ the bot.

### DIFF
--- a/greptile.json
+++ b/greptile.json
@@ -1,0 +1,3 @@
+{
+  "triggerOnUpdates": false
+}


greptile.json
```diff
@@ -0,0 +1,3 @@
+{
```
kellyguo11, line 1: would it work if we put this file under .github/ folder?

PR #4016, opened 2025-11-14
fix(isaaclab.bat) avoid CMD window-title quote rule breaking for /f subprocess

KraHsu:
# Description

This PR fixes an issue in the Windows batch script responsible for detecting the `torch` version.  
Inside `:ensure_cuda_torch`, the `for /f ('...')` construct wraps the inner command with an additional layer, resulting in something like:

```bash
for /f "tokens=2" %%V in ('"!python_exe!" -m pip show torch ^| findstr /B /C:"Version:"') do set "TORCH_CUR=%%V"
==>
cmd /c ""root_path\_isaac_sim\python.bat" -m pip show zipp | find "Version:""
```

This triggers the quote-handling behavior of cmd /c, causing the outer pair of quotes to be stripped. As a result, the actual inner command becomes something like:

```bash
root_path\_isaac_sim\python.bat" -m pip show zipp | find "Version:"
```

This leads to the error: `The filename, directory name, or volume label syntax is incorrect.`

Considering that this fix is extremely small and straightforward, I did not open a separate issue for it (please let me know if this aligns with the contribution guidelines). Also, because the change is minimal and the bug itself is not severe, I did not add my name to CONTRIBUTOR.md.

## Type of change

## Screenshots

Before the fix (my Windows system is in Chinese, so the error message is localized):
Debug - findstr with utf-8_1
After the fix:
Debug - findstr with utf-8_2

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,6 +12,7 @@ docs/
 **/output/*
 **/outputs/*
 **/videos/*
+**/wandb/*
 *.tmp
 # ignore docker
 docker/cluster/exports/
@@ -25,3 +26,5 @@ recordings/
 _isaac_sim?
 # Docker history
 docker/.isaac-lab-docker-history
+# ignore uv environment
+env_isaaclab
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -96,6 +96,7 @@ Guidelines for modifications:
 * Lukas Fröhlich
 * Manuel Schweiger
 * Masoud Moghani
+* Mateo Guaman Castro
 * Maurice Rahme
 * Michael Gussert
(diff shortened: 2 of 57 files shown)


KraHsu:
Apologies, I just realized that the original use of double quotes "" might have been to avoid errors caused by spaces in the path.  Based on my tests, after modifying my PR, it does indeed cause errors when there are spaces in the ISAACLAB_PATH.

However, other for /f statements in the isaaclab.bat script seem to use the version without quotes, so the issue might actually be in line with the “consistent behavior” across the script?

I suspect a better solution might be to clarify in the documentation not to place IsaacLab in a path with spaces (or make more significant changes to ensure there are no errors in such cases).

PR #4017, opened 2025-11-14
Adds gnureadline to base docker image

bmccann-bdai:
# Description
Add gnureadline to the base image. Fix up arrow capture in the python REPL.


bmccann-bdai:
> Do we need to add license for this dependency?

I just want it as a dependency for the python REPL. I don't intend to build code against it. If it's a problem, I can just install it locally.

kellyguo11:
Generally, we need permissive licenses for all the dependencies we install in the docker. gnureadline looks like is GPL license, so I think it's best to avoid it if we can.

PR #4102, opened 2025-11-27
[Newton]Replaces PhysxSchema API call with string Token get and set

ooctipus:
# Description

PhysxSchema requires omni.physx packages which is not available if ov is not available. Since it is not a pxr pacakge, we use string rather than api to avoid dependency on ov.

## Checklist


AntoineRichard:
@ooctipus Could you resolve the merge conflicts? It's not straight forward, and I don't have time to do it. Thanks! :D

ooctipus:
Done : ))

PR #4123, opened 2025-12-02
[Newton] Lots of hackery to get kit-optional paths working with and without OV

kellyguo11:
# Description

A lot of hacking to decouple from OV - so far only tested on Cartpole.

Changes that need to be considered:

- Includes new cloner changes from Octi
- Need a nice way to deal with assets without nucleus and omni.client APIs
- Removed isaac sim SimulationContext and SimulationManager references
- All omni.ui dependencies are lazy loaded
- A lot of hacks around prim and stage utils
- Alternative way for carb settings
- XFormPrim class is re-implemented in Isaac Lab
- omni.kit.commands, semantics are replaced
- omni.timeline callbacks are removed

Known Issues:

- OV visualizer with ground plane doesn't show up properly - material files are not being downloaded with the asset
- Not sure if training is still working


kellyguo11:
Closing, newer PR available at #4165

PR #4168, opened 2025-12-07
Fixes FrameTransformer to handle same body names at different hierarchy levels

### LINKED ISSUE #4167: [Bug Report] FrameTransformer returns identical transforms for bodies with same name but different paths
### Describe the bug

`FrameTransformer` uses only the body name (last component of the prim path) as the unique key to identify and track bodies. This causes target frames with the same body name but different hierarchical paths to collide, resulting in only one body being tracked while all frames with matching body names return identical (incorrect) transforms.

**Example:** When configuring a `FrameTransformer` to track both `Robot/left_hand` and `Robot_1/left_hand`:
- Both resolve to `body_name = "left_hand"` 
- Only the first registered body is actually tracked in the physics view
- Both `left_hand` and `left_hand_1` target frames return identical transforms (the transform of whichever body was registered first)

**Affected use cases:**
- Multi-robot scenarios with identical robot types
- Single articulation with duplicate body names at different hierarchy levels (e.g., `arm/link` vs `leg/link`)

### Steps to reproduce

```python
from isaaclab.sensors import FrameTransformerCfg, OffsetCfg

# Configure FrameTransformer to track same-named bodies from two robots
frame_transformer = FrameTransformerCfg(
    prim_path="{ENV_REGEX_NS}/Robot/torso_base",
    target_frames=[
        # Robot's left hand
        FrameTransformerCfg.FrameCfg(
            name="left_hand",
            prim_path="{ENV_REGEX_NS}/Robot/left_hand",
        ),
        # Robot_1's left hand (same body name, different robot)
        FrameTransformerCfg.FrameCfg(
            name="left_hand_1",
            prim_path="{ENV_REGEX_NS}/Robot_1/left_hand",
        ),
    ],
)

# After running simulation:
# env.scene["frame_transformer"].data.target_pos_source for "left_hand"
# and "left_hand_1" will return IDENTICAL values, even though the robots
# are at different positions.
```

**Root cause in code**:
https://github.com/isaac-sim/IsaacLab/blob/64ecea24f51bd008d2ae75dc2a233db5db515a71/source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py#L205

### System Info

- Commit: 64ecea24f51bd008d2ae75dc2a233db5db515a71
- Isaac Sim Version: 5.1.0.0
- OS: Ubuntu 22.04
- GPU: RTX 3060
- CUDA: 13.0
- GPU Driver: 580.95.05

### Additional context

The bug is in the `_initialize_impl` method of `FrameTransformer` class. The `body_name` variable is used as the key for `body_names_to_frames` dictionary, but it only contains the leaf name of the prim path, not the full relative path.

When a second body with the same leaf name is encountered, it adds the frame name to the existing entry but does NOT update the `prim_path`, causing only the first body to be tracked.

### Checklist

### Acceptance Criteria

- [x] Bodies with the same name but different hierarchical paths are tracked separately
- [x] Each target frame returns the correct transform for its specified body
- [x] Existing functionality for single-robot scenarios remains unchanged
- [x] Unit test added to verify the fix

bikcrum:
# Description

**Bug:** `FrameTransformer` uses only the body name (last component of the prim path) as the unique key to identify bodies. This causes target frames with the same body name but different hierarchical paths to collide, resulting in only one body being tracked.

**Example:** When tracking both `Robot/left_hand` and `Robot_1/left_hand`:
- Both resolve to `body_name = "left_hand"`
- Only the first one registered gets tracked
- Both `left_hand` and `left_hand_1` target frames return identical (incorrect) transforms

**Use cases affected:**
- Multi-robot scenarios (e.g., `Robot/left_hand` vs `Robot_1/left_hand`)
- Single articulation with duplicate body names at different hierarchy levels (e.g., `arm/link` vs `leg/link`)

**Fix:** Use the relative prim path (path after `env_X/`) as the unique body identifier instead of just the body name. For example:
- `/World/envs/env_0/Robot/left_hand` ->  key: `Robot/left_hand`
- `/World/envs/env_0/Robot_1/left_hand` -> key: `Robot_1/left_hand`

Fixes #4167

> Note: While users could work around this by creating separate FrameTransformer instances, this is not an acceptable solution because:
> 1. The current behavior silently returns incorrect data without any warning
> 2. The API contract (specifying explicit prim_paths) implies distinct bodies should be tracked
> 3. The workaround adds unnecessary complexity and memory overhead

## Type of change

## Screenshots

N/A - Internal data structure fix. No visual changes.

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py
+++ b/source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py
@@ -36,29 +36,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-def _get_relative_body_path(prim_path: str) -> str:
-    """Extract a normalized body path from a prim path.
-
-    Removes the environment instance segment `/envs/env_<id>/` to normalize paths
-    across multiple environments, while preserving the `/envs/` prefix to
-    distinguish environment-scoped paths from non-environment paths.
-
-    Examples:
-    - '/World/envs/env_0/Robot/torso' -> '/World/envs/Robot/torso'
-    - '/World/envs/env_123/Robot/left_hand' -> '/World/envs/Robot/left_hand'
-    - '/World/Robot' -> '/World/Robot'
-    - '/World/Robot_2/left_hand' -> '/World/Robot_2/left_hand'
-
-    Args:
-        prim_path: The full prim path.
-
-    Returns:
-        The prim path with `/envs/env_<id>/` removed, preserving `/envs/`.
-    """
-    pattern = re.compile(r"/envs/env_[^/]+/")
-    return pattern.sub("/envs/", prim_path)
-
-
 class FrameTransformer(SensorBase):
     """A sensor for reporting frame transforms.
 
@@ -225,7 +202,7 @@ class FrameTransformer(SensorBase):
                     )
 
                 # Get the name of the body: use relative prim path for unique identification
-                body_name = _get_relative_body_path(matching_prim_path)
+                body_name = self._get_relative_body_path(matching_prim_path)
                 # Use leaf name of prim path if frame name isn't specified by user
                 frame_name = frame if frame is not None else matching_prim_path.rsplit("/", 1)[-1]
 
@@ -315,10 +292,10 @@ class FrameTransformer(SensorBase):
             sorted_prim_paths = [all_prim_paths[index] for index in self._per_env_indices]
 
         # -- target frames: use relative prim path for unique identification
-        self._target_frame_body_names = [_get_relative_body_path(prim_path) for prim_path in sorted_prim_paths]
+        self._target_frame_body_names = [self._get_relative_body_path(prim_path) for prim_path in sorted_prim_paths]
 
         # -- source frame: use relative prim path for unique identification
-        self._source_frame_body_name = _get_relative_body_path(self.cfg.prim_path)
+        self._source_frame_body_name = self._get_relative_body_path(self.cfg.prim_path)
         source_frame_index = self._target_frame_body_names.index(self._source_frame_body_name)
 
         # Only remove source frame from tracked bodies if it is not also a target frame
@@ -550,3 +527,26 @@ class FrameTransformer(SensorBase):
         orientations = quat_from_angle_axis(angle, rotation_axis)
 
         return positions, orientations, lengths
+
+    @staticmethod
+    def _get_relative_body_path(prim_path: str) -> str:
+        """Extract a normalized body path from a prim path.
+
+        Removes the environment instance segment `/envs/env_<id>/` to normalize paths
+        across multiple environments, while preserving the `/envs/` prefix to
+        distinguish environment-scoped paths from non-environment paths.
+
+        Examples:
+        - '/World/envs/env_0/Robot/torso' -> '/World/envs/Robot/torso'
+        - '/World/envs/env_123/Robot/left_hand' -> '/World/envs/Robot/left_hand'
+        - '/World/Robot' -> '/World/Robot'
+        - '/World/Robot_2/left_hand' -> '/World/Robot_2/left_hand'
+
+        Args:
+            prim_path: The full prim path.
+
+        Returns:
+            The prim path with `/envs/env_<id>/` removed, preserving `/envs/`.
+        """
+        pattern = re.compile(r"/envs/env_[^/]+/")
+        return pattern.sub("/envs/", prim_path)


source/isaaclab/test/sensors/test_frame_transformer.py
```diff
@@ -589,3 +589,175 @@ def test_sensor_print(sim):
     sim.reset()
     # print info
     print(scene.sensors["frame_transformer"])
+
+
+@pytest.mark.isaacsim_ci
```
jtigue-bdai, line 593: @kellyguo11 Is this necessary here?

source/isaaclab/test/sensors/test_frame_transformer.py
```diff
@@ -589,3 +589,175 @@ def test_sensor_print(sim):
...
+
+        # Second robot (same type, different prim path)
+        robot_1 = ANYMAL_C_CFG.replace(
+            prim_path="{ENV_REGEX_NS}/Robot_1",
+            init_state=ANYMAL_C_CFG.init_state.replace(pos=(2.0, 0.0, 0.6)),
+        )
+
+        # Frame transformer tracking same-named bodies from both robots
+        frame_transformer: FrameTransformerCfg = FrameTransformerCfg(
+            prim_path="{ENV_REGEX_NS}/Robot/base",
```
jtigue-bdai: Currently this doesn't exactly test that the source frame is found correctly (because Robot is the normal). Can you also (or instead) test that Robot_1/base work as the FrameTransformerCfg.prim_path and that the transforms are correct below?

jtigue-bdai (commented):
Hey thanks for the feature. Some minor stuff with the test.

source/isaaclab/test/sensors/test_frame_transformer.py
bikcrum: Thanks for pointing that out, @jtigue-bdai. I’ve added test coverage to verify the source frame selection (including `Robot_1/base`) when there are duplicate body names.

source/isaaclab/test/sensors/test_frame_transformer.py
```diff
@@ -589,3 +589,207 @@ def test_sensor_print(sim):
...
+    expected_source_robot = "robot" if source_robot == "Robot" else "robot_1"
+    expected_source_base_body_idx = robot_base_body_idx if source_robot == "Robot" else robot_1_base_body_idx
+
+    # Define simulation stepping
+    sim_dt = sim.get_physics_dt()
+
+    # Simulate physics
+    for count in range(50):
+        # Reset periodically
+        if count % 25 == 0:
```
jtigue-bdai:
```suggestion
    for count in range(20):
        # Reset periodically
        if count % 10 == 0:
```

jtigue-bdai: unless you need 50 for this test to work i would suggest running the minimum number of steps to make this test work.

jtigue-bdai (approved):
NIT in the test, but thanks for the added functionality.

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,17 @@
...
 
+0.50.6 (2025-12-07)
+~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed FrameTransformer body name collision when tracking bodies with the same name but different hierarchical paths
+  (e.g., Robot/left_hand vs Robot_1/left_hand). The sensor now uses the relative prim path as the unique body identifier
+  instead of body name.
```
Mayankm96: Is the relative prim path w.r.t. environment namespace? 

Is it somewhat backwards compatible for existing users?

source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py
```diff
@@ -35,6 +35,32 @@
 # import logger
 logger = logging.getLogger(__name__)
 
+# Regex pattern for environment path parsing
+# Matches: env_0/, env_.*/, etc. and captures the path after
+_ENV_PATH_PATTERN = re.compile(r"env_[^/]+/(.*)")
```
Mayankm96: This seems the only occurance of this pattern. Might move to the function itself?

source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py
```diff
@@ -35,6 +35,32 @@
...
+
+
+def _get_relative_body_path(prim_path: str) -> str:
```
Mayankm96: This would fail for use-cases where environments aren't being used? For example if I have a simple use-case with "/World/Robot" and "/World/Robot_2"

source/isaaclab/docs/CHANGELOG.rst
bikcrum: @Mayankm96 

> Is the relative prim path w.r.t. environment namespace?

Not now. I updated it to normalize paths by removing the environment instance segment (env_0, env_1, etc.) while preserving the full path structure. For example, `/World/envs/env_0/Robot/torso` -> `/World/envs/Robot/torso`. This ensures all envs map to the same identifier for indexing. It also helps avoid collisions between env-scoped paths (e.g., `/World/envs/Robot/torso`) and non-env paths (e.g., `/World/Robot/torso`), as they remain distinct. If you have a better suggestion for this approach, let me know.

> Is it somewhat backwards compatible for existing users?

Yes. Only internal body tracking changed. User-facing frame names still default to leaf names when not explicitly provided, so existing code like `target_frame_names.index("left_hand")` continues to work.

I have updated the changelog for clarity.

source/isaaclab/isaaclab/sensors/frame_transformer/frame_transformer.py
bikcrum: @Mayankm96 That's a good point. I've added a change to handle non-env paths. The normalization only affects paths with `env_*` segments. For paths without environments:

`/World/Robot` -> `/World/Robot` (unchanged)
`/World/Robot_2` -> `/World/Robot_2` (unchanged)

These remain distinct, so there's no collision. The function only normalizes env-scoped paths (removing env_0, env_1, etc.) and leaves non-env paths unchanged. This preserves uniqueness for both cases.

I've added test cases to verify this works correctly for both env and non-env scenarios.

Mayankm96: NIT: Can this function be moved to the class itself as a static method? Mainly to keep the class as the main entry point for this file. Thanks!

bikcrum: @Mayankm96 Good point. I moved it to a class.

Mayankm96 (approved):

Issue #4274, opened 2025-12-23
[Bug Report] delayed_buffer and its internal circular_buffer performance


iansseijelly:
### Describe the bug

This is not a correctness bug but rather a performance issue that could be addressed to improve the training throughput. 

During profiling the training process, we found a non-trivial amount of time spent in `_apply_actuator_model`, which tracks down to the `compute` function in delayed pd actuator. This profiling is using the sampling python profiler `py-spy`, so it provides an application-view of where time is spent and is rather coarse-grained. 

<img width="1512" height="859" alt="Image" src="https://github.com/user-attachments/assets/53d6fe09-980a-427e-8c2b-b4491c65d8b0" />

We created a standalone MVP with just the delayed actuators and profiled it with torch profiler. The biggest CPU time slice is spent on excessive calls to the `aten::item`, which is an artifact of storing the `max_length` property, which should be an integer, but is stored as a tensor on the GPU. The second biggest contributing factor is the calls to `torch.any`, which should evaluate to false after the policy warms up and reset should be infrequent. 
Attached is the complete torch profile for an unoptimized run:
```
Time taken: 317.0001810067333 ms
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                                                   Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg     Self CUDA   Self CUDA %    CUDA total  CUDA time avg    # of Calls  
-------------------------------------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                                            aten::copy_         0.60%       1.838ms         1.55%       4.721ms       5.245us       1.489ms        17.98%       1.489ms       1.655us           900  
                         Memcpy DtoD (Device -> Device)         0.00%       0.000us         0.00%       0.000us       0.000us       1.489ms        17.98%       1.489ms       1.655us           900  
                                              aten::any         1.14%       3.469ms         1.76%       5.377ms       8.962us       1.394ms        16.82%       1.394ms       2.323us           600  
void at::native::reduce_kernel<512, 1, at::native::R...         0.00%       0.000us         0.00%       0.000us       0.000us       1.394ms        16.82%       1.394ms       2.323us           600  
                                            aten::index         0.76%       2.318ms         1.38%       4.205ms      14.017us       1.241ms        14.97%       1.241ms       4.135us           300  
[...]

PR #4275, opened 2025-12-23
Optimizes the performance of circular buffer 

iansseijelly:
# Description

This PR addresses issue #4274.

## Type of change

## Screenshots

Training throughput before and after the patch running training on task `[Isaac-Velocity-Flat-Spot-v0]`.

delayed_buffer_perf_fix
## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/actuators/actuator_pd.py
+++ b/source/isaaclab/isaaclab/actuators/actuator_pd.py
@@ -360,9 +360,9 @@ class DelayedPDActuator(IdealPDActuator):
         self, control_action: ArticulationActions, joint_pos: torch.Tensor, joint_vel: torch.Tensor
     ) -> ArticulationActions:
         # apply delay based on the delay the model for all the setpoints
-        control_action.joint_positions = self.positions_delay_buffer.compute(control_action.joint_positions)
-        control_action.joint_velocities = self.velocities_delay_buffer.compute(control_action.joint_velocities)
-        control_action.joint_efforts = self.efforts_delay_buffer.compute(control_action.joint_efforts)
+        control_action.joint_positions[:] = self.positions_delay_buffer.compute(control_action.joint_positions)
+        control_action.joint_velocities[:] = self.velocities_delay_buffer.compute(control_action.joint_velocities)
+        control_action.joint_efforts[:] = self.efforts_delay_buffer.compute(control_action.joint_efforts)
         # compte actuator model
         return super().compute(control_action, joint_pos, joint_vel)
 
--- a/source/isaaclab/isaaclab/actuators/actuator_pd.py
+++ b/source/isaaclab/isaaclab/actuators/actuator_pd.py
@@ -360,9 +360,9 @@ class DelayedPDActuator(IdealPDActuator):
         self, control_action: ArticulationActions, joint_pos: torch.Tensor, joint_vel: torch.Tensor
     ) -> ArticulationActions:
         # apply delay based on the delay the model for all the setpoints
-        control_action.joint_positions[:] = self.positions_delay_buffer.compute(control_action.joint_positions)
-        control_action.joint_velocities[:] = self.velocities_delay_buffer.compute(control_action.joint_velocities)
-        control_action.joint_efforts[:] = self.efforts_delay_buffer.compute(control_action.joint_efforts)
+        control_action.joint_positions = self.positions_delay_buffer.compute(control_action.joint_positions)
+        control_action.joint_velocities = self.velocities_delay_buffer.compute(control_action.joint_velocities)
+        control_action.joint_efforts = self.efforts_delay_buffer.compute(control_action.joint_efforts)
         # compte actuator model
         return super().compute(control_action, joint_pos, joint_vel)
 
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -141,10 +141,8 @@ class CircularBuffer:
         # Only check if we haven't confirmed all batches are reset (avoids unnecessary checks if no reset done)
         if self._need_reset:
             is_first_push = self._num_pushes == 0
-            if is_first_push.any().item():
-                expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
-                expanded_data = data[None].expand_as(self._buffer)
-                torch.where(expanded_mask, expanded_data, self._buffer, out=self._buffer)
+            if torch.any(is_first_push):
+                self._buffer[:, is_first_push] = data[is_first_push]
             # mark all the batches to be available
             self._need_reset = False
         # increment number of number of pushes for all batches
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -50,7 +50,7 @@ class CircularBuffer:
         # note: this is initialized on the first call to :meth:`append`
         self._buffer: torch.Tensor = None  # type: ignore
         # track if all batches have been initialized
-        self._all_initialized: bool = False
+        self._need_reset: bool = True
 
     """
     Properties.
@@ -109,7 +109,7 @@ class CircularBuffer:
         # reset the number of pushes for the specified batch indices
         self._num_pushes[batch_ids] = 0
         # re-initialization is required
-        self._all_initialized = False
+        self._need_reset = True
         if self._buffer is not None:
             # set buffer at batch_id reset indices to 0.0 so that the buffer()
             # getter returns the cleared circular buffer after reset.
@@ -141,13 +141,13 @@ class CircularBuffer:
         self._buffer[self._pointer] = data
         # Check for batches with zero pushes and initialize all values in batch to first append
         # Only check if we haven't confirmed all batches are initialized (avoids GPU sync in hot path)
-        if not self._all_initialized:
+        if self._need_reset:
             is_first_push = self._num_pushes == 0
             if is_first_push.any().item():
                 self._buffer[:, is_first_push] = data[is_first_push]
             else:
                 # All batches now initialized, skip this check in future calls
-                self._all_initialized = True
+                self._need_reset = False
         # increment number of number of pushes for all batches
         self._num_pushes += 1
 
@@ -172,7 +172,7 @@ class CircularBuffer:
         if len(key) != self.batch_size:
             raise ValueError(f"The argument 'key' has length {key.shape[0]}, while expecting {self.batch_size}")
         # check if the buffer is empty
-        if not self._all_initialized:
+        if self._need_reset:
             if self._buffer is None or (self._num_pushes == 0).any().item():
                 raise RuntimeError("Attempting to retrieve data on an empty circular buffer. Please append data first.")
 
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -108,7 +108,7 @@ class CircularBuffer:
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
         self._num_pushes[batch_ids] = 0
-        # re-initialization is required
+        # reset is needed on next update to fill entire buffer with initial data
         self._need_reset = True
         if self._buffer is not None:
             # set buffer at batch_id reset indices to 0.0 so that the buffer()
@@ -140,13 +140,13 @@ class CircularBuffer:
         # add the new data to the last layer
         self._buffer[self._pointer] = data
         # Check for batches with zero pushes and initialize all values in batch to first append
-        # Only check if we haven't confirmed all batches are initialized (avoids GPU sync in hot path)
+        # Only check if we haven't confirmed all batches are reset (avoids unnecessary checks if no reset done)
         if self._need_reset:
             is_first_push = self._num_pushes == 0
             if is_first_push.any().item():
                 self._buffer[:, is_first_push] = data[is_first_push]
             else:
-                # All batches now initialized, skip this check in future calls
+                # mark all the batches to be available
                 self._need_reset = False
         # increment number of number of pushes for all batches
         self._num_pushes += 1
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -107,7 +107,7 @@ class CircularBuffer:
         if batch_ids is None:
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
-        self._num_pushes[batch_ids] = 0 
+        self._num_pushes[batch_ids] = 0
         # re-initialization is required
         self._all_initialized = False
         if self._buffer is not None:
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -38,7 +38,9 @@ class CircularBuffer:
         self._device = device
         self._ALL_INDICES = torch.arange(batch_size, device=device)
 
-        # max length tensor for comparisons
+        # max length integer for cpu comparisons
+        self._max_len_int = max_len
+        # broadcastedmax length tensor for gpu tensor comparisons
         self._max_len = torch.full((batch_size,), max_len, dtype=torch.int, device=device)
         # number of data pushes passed since the last call to :meth:`reset`
         self._num_pushes = torch.zeros(batch_size, dtype=torch.long, device=device)
@@ -47,6 +49,8 @@ class CircularBuffer:
         # the actual buffer for data storage
         # note: this is initialized on the first call to :meth:`append`
         self._buffer: torch.Tensor = None  # type: ignore
+        # track if all batches have been initialized
+        self._all_initialized: bool = False
 
     """
     Properties.
@@ -65,7 +69,7 @@ class CircularBuffer:
     @property
     def max_length(self) -> int:
         """The maximum length of the ring buffer."""
-        return int(self._max_len[0].item())
+        return self._max_len_int
 
     @property
     def current_length(self) -> torch.Tensor:
@@ -103,7 +107,9 @@ class CircularBuffer:
         if batch_ids is None:
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
-        self._num_pushes[batch_ids] = 0
+        self._num_pushes[batch_ids] = 0 
+        # re-initialization is required
+        self._all_initialized = False
         if self._buffer is not None:
             # set buffer at batch_id reset indices to 0.0 so that the buffer()
             # getter returns the cleared circular buffer after reset.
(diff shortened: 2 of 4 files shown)


garylvov:
Hi, this is a cool optimization, I'm excited to try it, thank you!

Two quick qs: 

- Does this only affect configurations that use the delayed PD actuator, or does it apply to all actuator configurations?
- I see in your charts that the FPS is improved. How much of a difference does this make in cumulative training time? Like, does it go from 2 hours to train to 1.5 hours to train? I'd love to see the reward with and without the optimization over time

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -358,9 +358,9 @@ def compute(
         self, control_action: ArticulationActions, joint_pos: torch.Tensor, joint_vel: torch.Tensor
     ) -> ArticulationActions:
         # apply delay based on the delay the model for all the setpoints
-        control_action.joint_positions = self.positions_delay_buffer.compute(control_action.joint_positions)
-        control_action.joint_velocities = self.velocities_delay_buffer.compute(control_action.joint_velocities)
-        control_action.joint_efforts = self.efforts_delay_buffer.compute(control_action.joint_efforts)
+        control_action.joint_positions[:] = self.positions_delay_buffer.compute(control_action.joint_positions)
```
Mayankm96: Is the assignment operation here needed? 🤔

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -100,6 +104,8 @@ def reset(self, batch_ids: Sequence[int] | None = None):
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
         self._num_pushes[batch_ids] = 0
+        # re-initialization is required
+        self._all_initialized = False
```
Mayankm96: NIT: Instead of "all_initialized", a better name might be "_has_reset"?

Syntax-wise, initialization sounds more like creation of the buffers, while here it is more resetting them.

source/isaaclab/isaaclab/actuators/actuator_pd.py
T-K-233: This uses slice assignment to keep the original tensor storage (`control_action.joint_positions` etc.) and overwrite its contents in place. Without this, compute() may return a new tensor, causing buffer replacement and additional allocation or copy overhead.

T-K-233:
> Hi, this is a cool optimization, I'm excited to try it, thank you!
> 
> Two quick qs:
> 
> * Does this only affect configurations that use the delayed PD actuator, or does it apply to all actuator configurations?

It only affect the delayed PD actuator. Only the `DelayedPDActuator` utilizes the `DelayBuffer` object.

> * I see in your charts that the FPS is improved. How much of a difference does this make in cumulative training time? Like, does it go from 2 hours to train to 1.5 hours to train? I'd love to see the reward with and without the optimization over time

For the `Isaac-Velocity-Flat-Spot-v0` task, the cumulative training time reduces from 45.76 minutes to 42.35 minutes. This is with 2000 iters, and the difference will become bigger for longer training runs. 

The reward and metric of this task with and without this change is identical.

image

The full tensorboard log is also attached for reference:

[spot_training_logs.zip](https://github.com/user-attachments/files/24356993/spot_training_logs.zip)

garylvov:
Awesome, thank you for answering my questions, I really appreciate it!

garylvov (approved):

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
iansseijelly: thanks, I have changed the naming and the boolean conditions accordingly

source/isaaclab/isaaclab/actuators/actuator_pd.py
Mayankm96: The delay buffer anyway returns a copied tensor since the time-lags indexed the torch tensor internally. The operation here will do another copy of that tensor which I don't think is needed. Also it will override the initial command set into the environment (after action processing). This may affect the next decimation consequently.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -100,6 +104,8 @@ def reset(self, batch_ids: Sequence[int] | None = None):
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
         self._num_pushes[batch_ids] = 0
+        # re-initialization is required
```
Mayankm96:
```suggestion
        # reset is needed on next update to fill entire buffer with initial data
```

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -129,9 +135,14 @@ def append(self, data: torch.Tensor):
         # add the new data to the last layer
         self._buffer[self._pointer] = data
         # Check for batches with zero pushes and initialize all values in batch to first append
-        is_first_push = self._num_pushes == 0
-        if torch.any(is_first_push):
-            self._buffer[:, is_first_push] = data[is_first_push]
+        # Only check if we haven't confirmed all batches are initialized (avoids GPU sync in hot path)
```
Mayankm96:
```suggestion
        # Only check if we haven't confirmed all batches are reset (avoids unnecessary checks if no reset done)
```

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -129,9 +135,14 @@ def append(self, data: torch.Tensor):
...
+        if self._need_reset:
+            is_first_push = self._num_pushes == 0
+            if is_first_push.any().item():
+                self._buffer[:, is_first_push] = data[is_first_push]
+            else:
+                # All batches now initialized, skip this check in future calls
```
Mayankm96:
```suggestion
                # mark all the batches to be available
```

iansseijelly: Thanks, I have rephrased all the comments accordingly.

source/isaaclab/isaaclab/actuators/actuator_pd.py
Mayankm96: @iansseijelly and @T-K-233 : Let me know your thoughts on the above. Thanks again!

Mayankm96: Pinging again. @iansseijelly @T-K-233

iansseijelly: Sorry, I missed this comment earlier. 
The intention of line this was just to replace the .clone called from the original compute here to save the cost of one allocation. Indeed, the time-lag index should create a copied tensor, so if the original clone is unuseful, then so is this one. 
Regarding the decimation comment, I do not fully get it. But the decimation semantics should remain unchanged before and after this PR.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -156,8 +167,9 @@ def __getitem__(self, key: torch.Tensor) -> torch.Tensor:
         if len(key) != self.batch_size:
             raise ValueError(f"The argument 'key' has length {key.shape[0]}, while expecting {self.batch_size}")
         # check if the buffer is empty
-        if torch.any(self._num_pushes == 0) or self._buffer is None:
-            raise RuntimeError("Attempting to retrieve data on an empty circular buffer. Please append data first.")
+        if self._need_reset:
+            if self._buffer is None or (self._num_pushes == 0).any().item():
+                raise RuntimeError("Attempting to retrieve data on an empty circular buffer. Please append data first.")
```
AntoineRichard: Mostly a NIT but I'm not sure, about these lines.
If you need a reset, do we need to perform that extra check to raise the exception?
Also what happens if the `self._need_reset` flag is True, but none of the other checks are True. Should we still return the buffer? 

It looks to me that if `_need_reset` is False, then `append` should have made sure none of the other conditions could be True.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -100,6 +104,8 @@ def reset(self, batch_ids: Sequence[int] | None = None):
             batch_ids = slice(None)
         # reset the number of pushes for the specified batch indices
         self._num_pushes[batch_ids] = 0
+        # reset is needed on next update to fill entire buffer with initial data
+        self._need_reset = True
         if self._buffer is not None:
             # set buffer at batch_id reset indices to 0.0 so that the buffer() getter returns the cleared circular buffer after reset.
             self._buffer[:, batch_ids, :] = 0.0
```
AntoineRichard: Consider replacing with:
`self._buffer[:, batch_ids].zero_()`

```
Benchmarking with BATCH_SIZE=4096, NUM_JOINTS=12, HISTORY_LENGTH=6
====================================================================================================
Scenario                  | Method 1 (= 0.0)          | Method 2 (.zero_())       | Best         
----------------------------------------------------------------------------------------------------
1 reset (single index)    | 20.33 ± 1.51 µs           | 19.52 ± 1.70 µs           | Method 2 (1.04x)
10 resets                 | 21.42 ± 1.55 µs           | 21.20 ± 1.67 µs           | Method 2 (1.01x)
100 resets                | 22.24 ± 1.60 µs           | 21.64 ± 1.20 µs           | Method 2 (1.03x)
500 resets                | 21.91 ± 1.03 µs           | 21.57 ± 1.10 µs           | Method 2 (1.02x)
1000 resets               | 22.03 ± 1.30 µs           | 21.79 ± 1.16 µs           | Method 2 (1.01x)
2048 resets (half)        | 22.61 ± 2.16 µs           | 22.31 ± 1.34 µs           | Method 2 (1.01x)
All resets (N)            | 23.23 ± 1.37 µs           | 23.18 ± 0.51 µs           | Method 2 (1.00x)
====================================================================================================
```

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -129,9 +135,14 @@ def append(self, data: torch.Tensor):
...
         self._buffer[self._pointer] = data
         # Check for batches with zero pushes and initialize all values in batch to first append
-        is_first_push = self._num_pushes == 0
-        if torch.any(is_first_push):
-            self._buffer[:, is_first_push] = data[is_first_push]
+        # Only check if we haven't confirmed all batches are reset (avoids unnecessary checks if no reset done)
+        if self._need_reset:
+            is_first_push = self._num_pushes == 0
+            if is_first_push.any().item():
+                self._buffer[:, is_first_push] = data[is_first_push]
```
AntoineRichard, line 145: I would also consider doing:
```python
is_first_push = self._num_pushes == 0
expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
expanded_data = data[None].expand_as(self._buffer)
torch.where(expanded_mask, expanded_data, self._buffer,  out=self._buffer)
```

This should be about 3 times faster than:
```python
is_first_push = self._num_pushes == 0
self._buffer[:, is_first_push] = data[is_first_push]
```

And then it saves you the:
```python
if is_first_push.any().item():
```

which on its own costs about 0.6x the time of:
```
is_first_push = self._num_pushes == 0
expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
expanded_data = data[None].expand_as(circular_buffer)
torch.where(expanded_mask, expanded_data, self._buffer, self._buffer)
```

TLDR; replace:
```
is_first_push = self._num_pushes == 0
            if is_first_push.any().item():
                self._buffer[:, is_first_push] = data[is_first_push]
```
with:
```python
is_first_push = self._num_pushes == 0
if is_first_push.any().item():
    expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
    expanded_data = data[None].expand_as(self._buffer)
    torch.where(expanded_mask, expanded_data, self._buffer, out=self._buffer)
```

```
[INFO] Using python from: /home/antoiner/Documents/IsaacLab-Internal/_isaac_sim/python.sh                                                                                              
Benchmarking with BATCH_SIZE=4096, NUM_JOINTS=12, HISTORY_LENGTH=6
============================================================================================================================================
Scenario             | direct_mask        | where_mask         | Best
--------------------------------------------------------------------------------------------------------------------------------------------
1 reset              |  94.4± 3.6µs |  25.0± 1.1µs | where_mask
10 resets            |  96.8± 4.2µs |  25.3± 0.8µs | where_mask
100 resets           |  99.3± 4.2µs |  25.3± 1.7µs | where_mask
500 resets           |  99.7± 4.5µs |  25.5± 1.5µs | where_mask
1000 resets          |  98.9± 3.4µs |  25.3± 1.0µs | where_mask
2048 resets (half)   |  99.8± 3.4µs |  26.0± 1.4µs | where_mask
All resets (N)       | 100.3± 3.7µs |  26.0± 1.7µs | where_mask
============================================================================================================================================
```

AntoineRichard (commented):
The policies are training well on this branch and they do transfer to main, so it's likely the implementation is correct.

Code-wise it also looks good. I've suggested some modification to further improve the performance, but it's not as substantial as removing the cloning step. If we are in a rush we can get this in as is. These will likely get reworked to support warp anyway.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
iansseijelly: In practice, this is never raised. I wanted to match the semantics of the original code, but guard the check with a scalar value to make it cheaper, so hence the change. The assumption here is that when `_need_reset` is True, the conditions in here better be true, because _need_reset is a shadow boolean state in cpu.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
iansseijelly: I agree with this. Thanks for the detailed microbenchmark.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
iansseijelly: I also agree with this. Looks like it saves a lot of time. 
Probably need a little more comment/documentation on what's happening here since mask programming is less interpretable.

iansseijelly:
Hi, we have just addressed the suggestions that @AntoineRichard provided. 
Thanks for the help, everyone :)

kellyguo11:
@iansseijelly Thanks for the improvements and addressing the comments! As we've shifted our development to the new `develop` branch, could you do a rebase to resolve the conflicts on top of develop?

iansseijelly:
> @iansseijelly Thanks for the improvements and addressing the comments! As we've shifted our development to the new `develop` branch, could you do a rebase to resolve the conflicts on top of develop?

We just rebased atop `develop`. Thanks.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
kellyguo11: could we simplify this to check only for `self._need_reset` then if that's the only value needed?

source/isaaclab/isaaclab/actuators/actuator_pd.py
kellyguo11: I agree it'd be better removing the `[:]` in these lines if they are not necessary.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
iansseijelly: Yeah, sounds good. I have made teh changes accordingly. This should also be safe, since overall we are expected to only call getitem after at least one append.

source/isaaclab/isaaclab/actuators/actuator_pd.py
iansseijelly: Thanks, I have removed them accordingly.

kellyguo11:
hmm github is not giving me the option to update the branch. could you grab the latest changes from tot develop so that the CI tests can run? we added a fix recently that was preventing the installation from working.

iansseijelly:
Hi, I just rebased again atop develop. Can you try again? Thanks for the help.

kellyguo11:
Thanks a lot! seeing a test failure in test_observation_manager.py

```
source/isaaclab/test/managers/test_observation_manager.py ......F....... [ 93%]
.                                                                        [100%]

=================================== FAILURES ===================================
_________________________ test_compute_with_2d_history _________________________
source/isaaclab/test/managers/test_observation_manager.py:462: in test_compute_with_2d_history
    observations = obs_man.compute()
                   ^^^^^^^^^^^^^^^^^
source/isaaclab/isaaclab/managers/observation_manager.py:337: in compute
    obs_buffer[group_name] = self.compute_group(group_name, update_history=update_history)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
source/isaaclab/isaaclab/managers/observation_manager.py:421: in compute_group
    circular_buffer.append(obs)
source/isaaclab/isaaclab/utils/buffers/circular_buffer.py:145: in append
    expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   RuntimeError: The expanded size of the tensor (256) must match the existing size (20) at non-singleton dimension 3.  Target sizes: [5, 20, 128, 256, 1].  Tensor sizes: [1, 20, 1]
```

Dexsuite task in test_environments.py also failing with similar error:

```
________ test_environments[Isaac-Dexsuite-Kuka-Allegro-Lift-v0-32-cuda] ________
source/isaaclab_tasks/test/test_environments.py:38: in test_environments
    _run_environments(task_name, device, num_envs, create_stage_in_memory=False)
source/isaaclab_tasks/test/env_test_utils.py:158: in _run_environments
    _check_random_actions(
source/isaaclab_tasks/test/env_test_utils.py:230: in _check_random_actions
    obs, _ = env.reset()
             ^^^^^^^^^^^
/isaac-sim/kit/python/lib/python3.12/site-packages/gymnasium/wrappers/common.py:400: in reset
    return super().reset(seed=seed, options=options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/isaac-sim/kit/python/lib/python3.12/site-packages/gymnasium/core.py:333: in reset
    return self.env.reset(seed=seed, options=options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
source/isaaclab/isaaclab/envs/manager_based_env.py:384: in reset
    self.obs_buf = self.observation_manager.compute(update_history=True)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
source/isaaclab/isaaclab/managers/observation_manager.py:337: in compute
    obs_buffer[group_name] = self.compute_group(group_name, update_history=update_history)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
source/isaaclab/isaaclab/managers/observation_manager.py:412: in compute_group
    circular_buffer.append(obs)
source/isaaclab/isaaclab/utils/buffers/circular_buffer.py:145: in append
    expanded_mask = is_first_push[None, :, None].expand_as(self._buffer)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E   RuntimeError: The expanded size of the tensor (4) must match the existing size (32) at non-singleton dimension 2.  Target sizes: [5, 32, 4, 3].  Tensor sizes: [1, 32, 1]
```

T-K-233:
Thanks. We have fixed the logic.

I think @AntoineRichard 's solution only applies to the actuator's case. However, Isaac Lab test catches a case where input tensor can be higher dimensional (dim=5 in the failing case).

Additionally, we realized that the suggested line
```python
torch.where(expanded_mask, expanded_data, self._buffer,  out=self._buffer)
```
does not apply any effect.

We are reverting to our initial solution.

StafaH:
Hi @iansseijelly, sorry for the delay in getting back to you here. We are still interested in getting this merged in, are you able to update the MR and resolve the conflicts? If you can also post up-to-date benchmark results in the perf difference that would be appreciated!

ooctipus:
Thanks @iansseijelly (and apologies for the long open time) — the ideas here have since landed on `develop` via the frame-stacking refactor (#5574, #5849): `_max_len_int` avoids the `.item()` sync in `max_length`, and the CPU-side `_need_reset` gate skips the steady-state `torch.any` probe. The final piece this PR targeted — the `torch.any` + masked backfill inside `append` — is removed in #6471 with benchmarks. Closing as superseded; the analysis here was correct and is now fully covered on `develop`.

PR #4286, opened 2025-12-27
Simplifies import of stage and prim utils

Mayankm96:
# Description

This MR removes simplifies usage of prim and stage utillities.

* Removed unused utility functions (some of them are in legacy sub-module)
* Deleted entire nucleus.py module (functions check_server() and get_assets_root_path() were not used anywhere)
* Updated all spawner functions to use USD Stage API directly via stage.GetPrimAtPath() instead of wrapper utilities
* Refactored create_prim() to use stage.DefinePrim() directly with better error handling
* Added stage parameter to various functions for better control
* Updated files across scripts, source code, and tests to use the unified sim_utils import pattern
* Enhanced logger.py with comprehensive docstrings (non-functional change)

Moving forward, it is recommended to use:

```python
```

instead of:

```python
import isaaclab.sim.utils.prims as prim_utils
import isaaclab.sim.utils.stage as stage_utils
```

## Type of change

- Documentation update

## Checklist

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,5 +1,6 @@
 Changelog
 ---------
+
 0.51.0 (2025-12-29)
 ~~~~~~~~~~~~~~~~~~~
 
@@ -27,7 +28,7 @@ Changed
 * Moved ``logger.py`` to the :mod:`isaaclab.utils` module.
 
 
-0.50.6 (2025-12-29)
+0.50.7 (2025-12-29)
 ~~~~~~~~~~~~~~~~~~~
 
 Changed


source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -481,14 +482,6 @@ def get_setting(self, name: str) -> Any:
...
         return self.carb_settings.get(name)
 
-    def forward(self) -> None:
-        """Updates articulation kinematics and fabric for rendering."""
-        if self._fabric_iface is not None:
-            if self.physics_sim_view is not None and self.is_playing():
-                # Update the articulations' link's poses before rendering
-                self.physics_sim_view.update_articulations_kinematic()
-            self._update_fabric(0.0, 0.0)
-
```
ooctipus, line 491: is this ordering change intended?

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -176,225 +157,153 @@ def create_prim(
...
-    if not is_prim_path_valid(prim_path, fabric=fabric):
-        raise Exception(f"A prim does not exist at prim path: {prim_path}")
-    prim = get_prim_at_path(prim_path, fabric=fabric)
-    if fabric:
-        return prim.GetTypeName()
-    else:
-        return prim.GetPrimTypeInfo().GetTypeName()
-
-
-def from_prim_get_type_name(prim: Usd.Prim, fabric: bool = False) -> str:
```
ooctipus, line 360: Even our isaaclab always operates on prim_path, this MAY be useful for user scripts when some callback directly requires prim rather than prim_path. it can save time to directly parse from string.

source/isaaclab/isaaclab/utils/logger.py
pascal-roth, line 1: NIT: this should probably be a git mv instead of readding the same file

source/isaaclab/isaaclab/sim/utils/stage.py
```diff
@@ -214,223 +172,41 @@ def get_current_stage_id() -> int:
...
-    """
-    # Note: Need to import this here to prevent circular dependencies.
-    from .prims import get_prim_path
-
-    for prim in traverse_stage(fabric=fabric):
-        prim_path = get_prim_path(prim)
-        print(prim_path)
-
-
-def add_reference_to_stage(usd_path: str, prim_path: str, prim_type: str = "Xform") -> Usd.Prim:
```
pascal-roth, line 370: could image that function being used by some people that would add realisitc assets or which workflow would we recommend instead?

source/isaaclab/isaaclab/sim/utils/stage.py
```diff
@@ -214,223 +172,41 @@ def get_current_stage_id() -> int:
...
-    stage = get_current_stage()
-    if stage is None:
-        raise Exception("There is no stage currently opened")
-    rootLayer = stage.GetRootLayer()
-    rootLayer.SetPermissionToEdit(True)
-    with Usd.EditContext(stage, rootLayer):
-        UsdGeom.SetStageUpAxis(stage, AXES_TOKEN[axis])
-
-
-def get_stage_up_axis() -> str:
```
pascal-roth, line 268: I assume here we want users to directly use the UsdGeom call? Maybe we could tell users that this is ported for now and will be deprecrated in 3.1?

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -176,225 +145,148 @@ def create_prim(
...
-        >>> import isaaclab.utils.prims as prims_utils
+        >>> import isaaclab.sim as sim_utils
         >>>
-        >>> prims_utils.delete_prim("/World/Cube")
+        >>> sim_utils.delete_prim("/World/Cube")
     """
-    DeletePrimsCommand([prim_path]).do()
-
-
-def get_prim_at_path(prim_path: str, fabric: bool = False) -> Usd.Prim | usdrt.Usd._Usd.Prim:
```
pascal-roth, line 196: maybe we add first a deprecation warning and then remove it in 3.1?

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -176,225 +145,148 @@ def create_prim(
...
-        return None
+def move_prim(path_from: str, path_to: str, keep_world_transform: bool = True, stage: Usd.Stage | None = None) -> None:
+    """Moves a prim from one path to another within a USD stage.
 
+    This function moves the prim from the source path to the destination path. If the :attr:`keep_world_transform`
+    is set to True, the world transform of the prim is kept. This implies that the prim's local transform is reset
+    such that the prim's world transform is the same as the source path's world transform. If it is set to False,
+    the prim's local transform is preserved.
 
-def get_prim_path(prim: Usd.Prim) -> str:
```
pascal-roth, line 223: same for this function

source/isaaclab/isaaclab/sim/utils/prims.py
pascal-roth: also this makes it indepedent from fabric (in case someone ever uses it)

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -645,103 +539,6 @@ def find_matching_prim_paths(prim_path_regex: str, stage: Usd.Stage | None = Non
...
-    Returns:
-        bool: Return true if prim implements all apis. Return false otherwise.
-    """
-    if not isinstance(apis, list):
-        return prim.HasAPI(apis)
-    else:
-        return all(prim.HasAPI(api) for api in apis)
-
-
-def resolve_pose_relative_to_physx_parent(
```
pascal-roth, line 665: this one is used within the IMU 
https://github.com/isaac-sim/IsaacLab/blob/d13cb0b6043db6ae9b2efc3ab1ac64f7a77ed3ed/source/isaaclab/isaaclab/sensors/imu/imu.py#L144C62-L144C99

source/isaaclab/isaaclab/sim/utils/stage.py
pascal-roth: ah nevermind, saw the new function in prims.py, can we maybe make a reference to that function and say that this one here will be removed later? Otherwise the new function might be hard to find for users

source/isaaclab/isaaclab/sim/utils/prims.py
Mayankm96: This function code is outdated. In USD 22.0 onwards, it is possible to directly do `GetTypeName()` for both PXR USD APIs and Fabric APIs. Thus, I removed it.

We havevery sparse tests running for fabric, so until we have those and see a necessity, I think it is okay to remove this function.

source/isaaclab/isaaclab/sim/utils/prims.py
Mayankm96: I simplified the logic of this function. It was trying to do a lot of things and we have other utilities to handle this more generally. 

This is how it is now: https://github.com/Mayankm96/IsaacLab/blob/fix/cleanup/source/isaaclab/isaaclab/sensors/imu/imu.py#L142-L157

source/isaaclab/isaaclab/sim/utils/prims.py
Mayankm96: Maybe we could write a migration guide for these functions so users know what to do. Since most people would likely be using isaacsim prim utils, guiding them there should cover the majority of cases.

Open to feedback if you think these functions are heavily used elsewhere though.

source/isaaclab/isaaclab/sim/utils/stage.py
Mayankm96: I doubt people are using this. We don't in our codebase and most other users right now would be resorting to isaacsim functions.

source/isaaclab/isaaclab/sim/utils/prims.py
pascal-roth: ah perfect, did not see that

source/isaaclab/isaaclab/sim/simulation_context.py
Mayankm96: yeah grouping functions by the way you should use them :)

Mayankm96:
@greptile-apps want to take another look?

source/isaaclab/isaaclab/sim/utils/prims.py
Mayankm96: Added legacy module.

source/isaaclab/isaaclab/sim/utils/prims.py
Mayankm96: Added legacy module.

source/isaaclab/isaaclab/sim/utils/stage.py
Mayankm96: Added legacy module.

source/isaaclab/isaaclab/sim/utils/stage.py
Mayankm96: Added legacy module.

source/isaaclab/isaaclab/utils/logger.py
Mayankm96: Not sure why that happened but it might be because i added other changes?

Mayankm96:
There are only three tests failing:

* Skrl and SB3 wrapper tests -- Timeout
* Asset test for UR10e -- I tried to inspect this locally but seems like something is off in their USDs. I am unable to set the variants from the GUI as well

Since these two go beyond the scope of this MR, I think we can move forward with this merge, as there are other MRs waiting for it.

PR #4377, opened 2026-01-13
Switches to ruff in-built isort ordering

### LINKED ISSUE #4336: [Proposal] Reconsider `isort` Import Ordering to Align with Standard Python Conventions
### Proposal

The current `isort` configuration reorders imports in a way that merges standard library modules with pip-installable dependencies, and applies custom grouping logic across multiple package families.

While this works from a tooling perspective, it diverges from what most Python users expect when reading code.

One possible direction would be to realign the default import structure with the standard three-block convention, while limiting special-case handling to packages that are genuinely runtime-provided (e.g. `isaacsim`, `omni`, `carb`, etc.).

This would keep most imports predictable and familiar, while still allowing flexibility where runtime constraints require it.

### Motivation

Reordering imports can be helpful in a limited set of scenarios, most commonly to mitigate circular import issues. In practice, however, frequent reliance on import reordering often points to deeper package design concerns. While reordering can work around symptoms, it does not address underlying structural issues.

Even if circular imports are not a concern in this specific case, the current import grouping introduces cognitive overhead. From a Python user’s perspective, the widely understood and de-facto standard convention is:

- Standard library imports
- External dependencies (declared in pyproject.toml)
- Internal / local modules

This structure makes it easier to:

- Reason about dependencies
- Spot missing or undeclared dependencies
- Quickly understand where a symbol comes from
- Onboard new contributors

### Build Info

- Isaac Lab Version: main
- Isaac Sim Version: N/A

### Additional context

This issue is intended to start a discussion around balancing:

- Tooling consistency
- Runtime constraints
- Readability and developer expectations

Original discussion started here: https://github.com/isaac-sim/IsaacLab/pull/3619

### Checklist

### Acceptance Criteria

- [ ] Import grouping follows a clearly documented and predictable structure
- [ ] Standard library and external dependencies are not merged by default
- [ ] Any special-case handling is limited and well-justified

Mayankm96:
# Description

Fixes #4336

## Type of change

- Documentation update

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,10 +63,6 @@ max-complexity = 30
 convention = "google"
 
 [tool.ruff.lint.isort]
-# Match isort behavior
-combine-as-imports = false
-order-by-type = true
-from-first = false
 
 # Custom import sections with separate sections for each Isaac Lab extension
 section-order = [


pyproject.toml
```diff
@@ -166,10 +78,10 @@ section-order = [
     "omniverse-extensions",
     # Group Isaac Lab extensions together since they are all part of the Isaac Lab project
     "isaaclab",
-    "isaaclab-assets",
     "isaaclab-rl",
     "isaaclab-mimic",
     "isaaclab-tasks",
+    "isaaclab-assets",
```
jtigue-bdai, line 80: Is there a specific reason for putting assets ant the bottom?

pyproject.toml
```diff
@@ -152,7 +64,7 @@ convention = "google"
 
 [tool.ruff.lint.isort]
 # Match isort behavior
-combine-as-imports = true
+combine-as-imports = false
```
jtigue-bdai: is the primary motivation to reduce the number of lines in the imports for those using `as`?

pyproject.toml
Mayankm96: Assets only have pre-built configurations. It looked slightly cleaner to have those global objects in the end of the file (i.e. based on the order of their typical usage in the code).

```python
from isaaclab_tasks.locomotion.velocity import VelocityEnvCfg

from isaaclab_assets.robots import DIGIT_CFG

cfg = VelocityEnvCfg()
cfg.scene.robot = DIGIT_CFG
```

pyproject.toml
Mayankm96: It is false by default: https://docs.astral.sh/ruff/settings/#lint_isort_combine-as-imports

Mayankm96: I removed the default settings for the pyproject file so that only overwritten properties are there.

PR #4379, opened 2026-01-13
Update RSL-RL configs to work with newer versions 4.0 and 5.0

ClemensSchwarke:
This PR updates the rsl-rl config classes to be compatible with version 4.0 and 5.0. It also adds a configuration handling function to ensure that old configs work with version 4.0/5.0 and adapts `train.py` and `play.py` accordingly. Lastly, it updates the config classes for ANYmal D locomotion for reference and adds recurrent configs.

The main change for 4.0 is that the policy config is split into seperate actor and critic / student and teacher configs. For more details see the [Release Notes](https://github.com/leggedrobotics/rsl_rl/releases/tag/v4.0.0).

The main change for 5.0 is that the noise distribution is now configured via the `distribution_cfg` to allow for arbitrary distributions. For more details see the [Release Notes](https://github.com/leggedrobotics/rsl_rl/releases/tag/v5.0.0).

## Checklist


source/isaaclab_rl/isaaclab_rl/rsl_rl/rl_cfg.py
Wenkai-Dong, line 223: ### Missing share_cnn_encoders in config dataclass
The code uses `cfg["algorithm"].pop("share_cnn_encoders", None)` to configure shared encoders, but `share_cnn_encoders` is not defined in the `RslRlPpoAlgorithmCfg` dataclass.
This causes validation errors with strict config systems (e.g., Hydra/OmegaConf) when users attempt to set this parameter in YAML, as it's treated as an unexpected argument.
Consider adding `share_cnn_encoders: bool = False` to `RslRlPpoAlgorithmCfg` to support this feature properly and ensure type safety.

kellyguo11:
@ClemensSchwarke could you run the formatter?

ClemensSchwarke:
I am not sure if the `handle_deprecated_rsl_rl_cfg` function lives in the right place and the tests for it are currently somewhere else (the isaaclab_rl extension), which I think is not ideal. Do you have any preference where those things should live @kellyguo11 @ooctipus? Thanks :)

kellyguo11:
> I am not sure if the `handle_deprecated_rsl_rl_cfg` function lives in the right place and the tests for it are currently somewhere else (the isaaclab_rl extension), which I think is not ideal. Do you have any preference where those things should live @kellyguo11 @ooctipus? Thanks :)

hmm perhaps making it a utility under isaaclab_rl/rsl_rl could make sense?

ClemensSchwarke:
> > I am not sure if the `handle_deprecated_rsl_rl_cfg` function lives in the right place and the tests for it are currently somewhere else (the isaaclab_rl extension), which I think is not ideal. Do you have any preference where those things should live @kellyguo11 @ooctipus? Thanks :)
> 
> hmm perhaps making it a utility under isaaclab_rl/rsl_rl could make sense?

Sounds good, thanks!

PR #4427, opened 2026-01-20
Fix misleading docstring for resolve_matching_names_values

### LINKED ISSUE #3849: [Bug Report] Confusing or incorrect docstring for `resolve_matching_names_values`
### Describe the bug

The documentation of `resolve_matching_names_values` says:

https://github.com/isaac-sim/IsaacLab/blob/79fc49dceeaa961da552ec9da54412d6d13be9b7/source/isaaclab/isaaclab/utils/string.py#L274

> If the :attr:`preserve_order` is True, the ordering of the matched indices and names is the same as the order    of the provided list of strings. This means that the ordering is dictated by the order of the target strings and not the order of the query regular expressions.

> If the :attr:`preserve_order` is False, the ordering of the matched indices and names is the same as the order    of the provided list of query regular expressions.

but the example appears to show the exact opposite behavior:

> For example, consider the dictionary is {"a|d|e": 1, "b|c": 2}, the list of strings is ['a', 'b', 'c', 'd', 'e']. If :attr:`preserve_order` is False, then the function will return the indices of the matched strings, the matched strings, and the values as: ([0, 1, 2, 3, 4], ['a', 'b', 'c', 'd', 'e'], [1, 2, 2, 1, 1]).

That example shows `preserve_order=False`, but the list of returned indices is clearly the same as the order of the `list_of_strings`, which is the opposite of what the previous part of the docstring says. 

I think the docstring only makes sense if you interpret "provided list of strings" as "keys in the data dictionary", but that doesn't make sense given that there is a variable named `list_of_strings`. 

### System Info

- Commit: 79fc49dceeaa961da552ec9da54412d6d13be9b7

### Checklist

Wheeeeeeeeels:
## Summary

Fix the misleading docstring for `resolve_matching_names_values` function in `isaaclab/utils/string.py`.

The docstring incorrectly described the behavior of the `preserve_order` parameter - the descriptions for `True` and `False` were swapped.

Fixes #3849

## Changes

**Before (incorrect):**
- `preserve_order=True`: ordering follows the list of strings
- `preserve_order=False`: ordering follows the query regular expressions

**After (correct):**
- `preserve_order=False` (default): ordering follows `list_of_strings`
- `preserve_order=True`: ordering follows the regex keys in data dictionary

## Verification

The code logic confirms the corrected behavior:
- Default path (lines 314-339): iterates over `list_of_strings` in order
- When `preserve_order=True` (lines 341-360): reorders results to follow `data` keys order

The example in the docstring was already correct; only the parameter descriptions were swapped.

### DIFF
--- a/scripts/benchmarks/benchmark_view_comparison.py
+++ b/scripts/benchmarks/benchmark_view_comparison.py
@@ -8,10 +8,12 @@
 This script tests the performance of batched transform operations using:
 
 - Isaac Lab's XformPrimView (USD-based)
+- Isaac Lab's XformPrimView (Fabric-based)
 - PhysX RigidBodyView (PhysX tensors-based, as used in RigidObject)
 
 Note:
     XformPrimView operates on USD attributes directly (useful for non-physics prims),
+    or on Fabric attributes when Fabric is enabled.
     while RigidBodyView requires rigid body physics components and operates on PhysX tensors.
     This benchmark helps understand the performance trade-offs between the two approaches.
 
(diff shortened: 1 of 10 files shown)


StafaH:

PR #4470, opened 2026-01-27
Fixes terrain origin computing in the inverted pyramid slope terrain

luoye2333:
# Description
In `height_field_to_mesh` function, the terrain origin is computed as the max in the square area around the center of the terrain,
and the square has a fixed width of 2 meters.
```
        # compute origin
        x1 = int((cfg.size[0] * 0.5 - 1) / cfg.horizontal_scale)  # fixed width
        x2 = int((cfg.size[0] * 0.5 + 1) / cfg.horizontal_scale)
        y1 = int((cfg.size[1] * 0.5 - 1) / cfg.horizontal_scale)
        y2 = int((cfg.size[1] * 0.5 + 1) / cfg.horizontal_scale)
        origin_z = np.max(heights[x1:x2, y1:y2]) * cfg.vertical_scale
        origin = np.array([0.5 * cfg.size[0], 0.5 * cfg.size[1], origin_z])
```
In inverted pyramid slope terrain, the default `platform_width` is `1.0` meter. Therefore the area outside the platform is also calculated in the `np.max` function. This caused the terrain origin to be a little higher than the ground. See the `Screenshots` below. Non-inverted pyramid terrain do not have this problem because the function is `max`.

To fix, a param `terrain_origin_judge_width` is added to the `height_field_to_mesh` wrapper (default is 2.0 so the behaviour is same as before) and we can use decorator `@height_field_to_mesh_v2(terrain_origin_judge_width=1.0)` to change the `max` function area. For the default platform 1.0 meter in inverted pyramid slope terrain, a half width `terrain_origin_judge_width=1.0` is ok.
```
        # compute origin
        x1 = int((cfg.size[0] * 0.5 - terrain_origin_judge_width / 2.0) / cfg.horizontal_scale)
        x2 = int((cfg.size[0] * 0.5 + terrain_origin_judge_width / 2.0) / cfg.horizontal_scale)
        y1 = int((cfg.size[1] * 0.5 - terrain_origin_judge_width / 2.0) / cfg.horizontal_scale)
        y2 = int((cfg.size[1] * 0.5 + terrain_origin_judge_width / 2.0) / cfg.horizontal_scale)
```

## Type of change

## Screenshots

| Before | After |
| ------ | ----- |
| img_v3_02ub_d9221c73-c96d-4b7d-967a-f5aa279e240g | img_v3_02ub_d8c4226e-419a-4342-9e83-95357df2fefg|

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/terrains/height_field/utils.py
+++ b/source/isaaclab/isaaclab/terrains/height_field/utils.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2025, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
 # All rights reserved.
 #
 # SPDX-License-Identifier: BSD-3-Clause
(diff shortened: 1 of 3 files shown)


source/isaaclab/isaaclab/terrains/height_field/utils.py
```diff
@@ -77,6 +76,59 @@ def wrapper(difficulty: float, cfg: HfTerrainBaseCfg):
     return wrapper
 
 
+def height_field_to_mesh_v2(terrain_origin_judge_width: float = 2.0):
```
Mayankm96, line 79: Hi, thank you for sharing the bug and potential fix.

This parameter `terrain_origin_judge_width` seems to depend a lot on the choice of the platform width size. I don't think hard-coding it here like this is the right solution.

The proper solution should be that the height-field generator also returns the origin. Then all this hard-coding logic inside `height_field_to_mesh` can be avoided.

PR #4475, opened 2026-01-27
Replaces Isaac Sim SimulationManager with IsaacLab SimulationManager

ooctipus:
# Description

Adds simulation manager that replaces Isaac sim simulation manager, 
this is currently a draft, things are mostly working, but lacks clean up.

### DIFF
--- a/source/isaaclab/isaaclab/visualizers/newton_visualizer.py
+++ b/source/isaaclab/isaaclab/visualizers/newton_visualizer.py
@@ -367,10 +367,10 @@ class NewtonVisualizer(Visualizer):
         self._is_closed = True
 
     def is_running(self) -> bool:
-        """Check if visualizer window is still open."""
+        """Check if visualizer window is still open (independent of pause state)."""
         if not self._is_initialized or self._is_closed or self._viewer is None:
             return False
-        return self._viewer.is_running()
+        return True
 
     def is_stopped(self) -> bool:
         """Check if visualizer window is closed."""
--- a/source/isaaclab/isaaclab/visualizers/physx_ov_visualizer.py
+++ b/source/isaaclab/isaaclab/visualizers/physx_ov_visualizer.py
@@ -421,10 +421,8 @@ class PhysxOVVisualizer(Visualizer):
         return self._timeline.is_playing()
 
     def is_stopped(self) -> bool:
-        """Check whether the simulation is stopped."""
-        if self._timeline is None:
-            return True
-        return self._timeline.is_stopped()
+        """Check whether the visualizer is stopped (closed)."""
+        return self._is_closed
 
     def play(self) -> None:
         """Start playing the simulation."""
@@ -446,12 +444,8 @@ class PhysxOVVisualizer(Visualizer):
     # ------------------------------------------------------------------
 
     def is_running(self) -> bool:
-        """Check if visualizer is still running (app is alive, not whether playing)."""
+        """Check if visualizer is still running (independent of pause state)."""
         return self._is_initialized and not self._is_closed
-
-    def is_stopped(self) -> bool:
-        """Check if visualizer is stopped (closed)."""
-        return self._is_closed
 
     def supports_markers(self) -> bool:
         """Supports markers via USD prims."""
--- a/source/isaaclab/isaaclab/visualizers/rerun_visualizer.py
+++ b/source/isaaclab/isaaclab/visualizers/rerun_visualizer.py
@@ -259,20 +259,16 @@ class RerunVisualizer(Visualizer):
         self._is_initialized = False
 
     def is_running(self) -> bool:
-        """Check if visualizer is running.
+        """Check if visualizer is running (independent of pause state).
 
         Returns:
-            True if viewer is initialized and running, False otherwise.
+            True if viewer is initialized and not closed, False otherwise.
         """
-        if self._viewer is None:
-            return False
-        return self._viewer.is_running()
+        return self._is_initialized and not self._is_closed and self._viewer is not None
 
     def is_stopped(self) -> bool:
-        """Check if visualizer is stopped."""
-        if self._viewer is None:
-            return True
-        return False
+        """Check if visualizer is stopped (closed)."""
+        return self._is_closed
 
     def is_training_paused(self) -> bool:
         """Check if training is paused.
--- a/source/isaaclab/isaaclab/visualizers/ov_visualizer.py
+++ b/source/isaaclab/isaaclab/visualizers/ov_visualizer.py
@@ -213,35 +213,28 @@ class OVVisualizer(Visualizer):
             torch.cuda.set_device(self._simulation_context.device)
 
     def is_running(self) -> bool:
-        """Check if visualizer is still running."""
-        return self.is_playing()
-
-    def is_playing(self) -> bool:
-        """Check whether the simulation is playing."""
-        if self._timeline is None:
+        """Check if visualizer is still running (independent of pause state)."""
+        if not self._is_initialized or self._is_closed:
             return False
-        return self._timeline.is_playing()
+        if self._simulation_app is not None and hasattr(self._simulation_app, "is_running"):
+            return self._simulation_app.is_running()  # type: ignore[union-attr]
+        return self._simulation_app_running
 
     def is_stopped(self) -> bool:
-        """Check whether the simulation is stopped."""
-        if self._timeline is None:
-            return True
-        return self._timeline.is_stopped()
+        """Check if visualizer is stopped (closed)."""
+        return self._is_closed
 
     def play(self) -> None:
-        """Start playing the simulation."""
-        if self._timeline is not None:
-            self._timeline.play()
+        """Start playing the simulation (no-op for OV - always playing)."""
+        pass
 
     def pause(self) -> None:
-        """Pause the simulation."""
-        if self._timeline is not None:
-            self._timeline.pause()
+        """Pause the simulation (no-op for OV)."""
+        pass
 
     def stop(self) -> None:
-        """Stop the simulation."""
-        if self._timeline is not None:
-            self._timeline.stop()
+        """Stop the simulation (no-op for OV)."""
+        pass
 
     def is_training_paused(self) -> bool:
         """Check if training is paused (always False for OV)."""


source/isaaclab/isaaclab/assets/asset_base.py
```diff
@@ -8,16 +8,15 @@
...
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
-import torch
-
 import omni.kit.app
 import omni.timeline
-from isaacsim.core.simulation_manager import IsaacEvents, SimulationManager
+from isaaclab.sim import PhysxSimulationManager as SimulationManager, SimulationEvent as IsaacEvents
```
AntoineRichard: This should not depend on the PhysX simulation manager ideal, it should be agnostic

source/isaaclab/isaaclab/assets/asset_base.py
```diff
@@ -346,8 +349,12 @@ def _on_prim_deletion(self, prim_path: str) -> None:
...
-    def _clear_callbacks(self) -> None:
-        """Clears the callbacks."""
+    def _clear_callbacks(self, event: Any = None) -> None:
+        """Clears the callbacks.
+
+        Args:
+            event: Optional event that triggered the callback (unused but required for event handlers).
+        """
         if self._prim_deletion_callback_id:
             SimulationManager.deregister_callback(self._prim_deletion_callback_id)
```
AntoineRichard: This call should be doable on the base simulation manager.

source/isaaclab/isaaclab/assets/asset_base.py
```diff
@@ -309,15 +308,18 @@ def _initialize_callback(self, event):
...
+                    raise RuntimeError("Simulation Context is not initialized!")
+                # Obtain device and backend
+                self._device = sim.device
+                # initialize the asset
                 self._initialize_impl()
             except Exception as e:
-                if builtins.ISAACLAB_CALLBACK_EXCEPTION is None:
-                    builtins.ISAACLAB_CALLBACK_EXCEPTION = e
+                if builtins.ISAACLAB_CALLBACK_EXCEPTION is None:  # type: ignore
+                    builtins.ISAACLAB_CALLBACK_EXCEPTION = e  # type: ignore
```
AntoineRichard: Why do we use builtins over a class variable in the simulation context?

source/isaaclab/isaaclab/envs/direct_marl_env.py
```diff
@@ -151,18 +149,16 @@ def __init__(self, cfg: DirectMARLEnvCfg, render_mode: str | None = None, **kwar
...
-                # since the reset can trigger callbacks which use the stage,
-                # we need to set the stage context here
-                with use_stage(self.sim.get_initial_stage()):
-                    self.sim.reset()
-                # update scene to pre populate data buffers for assets and sensors.
-                # this is needed for the observation manager to get valid tensors for initialization.
-                # this shouldn't cause an issue since later on, users do a reset over all the environments
-                # so the lazy buffers would be reset.
-                self.scene.update(dt=self.physics_dt)
+        print("[INFO]: Starting the simulation. This may take a few seconds. Please wait...")
```
AntoineRichard, line 152: Ideally this should use the logger.info

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -158,18 +156,16 @@ def __init__(self, cfg: DirectRLEnvCfg, render_mode: str | None = None, **kwargs
...
-                # since the reset can trigger callbacks which use the stage,
-                # we need to set the stage context here
-                with use_stage(self.sim.get_initial_stage()):
-                    self.sim.reset()
-                # update scene to pre populate data buffers for assets and sensors.
-                # this is needed for the observation manager to get valid tensors for initialization.
-                # this shouldn't cause an issue since later on, users do a reset over all the environments
-                # so the lazy buffers would be reset.
-                self.scene.update(dt=self.physics_dt)
+        print("[INFO]: Starting the simulation. This may take a few seconds. Please wait...")
```
AntoineRichard, line 159: Ideally this should use logger.info

source/isaaclab/isaaclab/envs/manager_based_env.py
```diff
@@ -3,16 +3,14 @@
...
+import torch
 import warnings
 from collections.abc import Sequence
 from typing import Any
 
-import torch
-
 import omni.physx
-from isaacsim.core.simulation_manager import SimulationManager
+from isaaclab.sim.physx_simulation_manager import PhysxSimulationManager as SimulationManager
```
AntoineRichard: Here this should not depend on physx_simulation_manager but the base manager

source/isaaclab/isaaclab/envs/manager_based_env.py
```diff
@@ -164,20 +157,18 @@ def __init__(self, cfg: ManagerBasedEnvCfg):
...
-                with use_stage(self.sim.get_initial_stage()):
-                    self.sim.reset()
-                # update scene to pre populate data buffers for assets and sensors.
-                # this is needed for the observation manager to get valid tensors for initialization.
-                # this shouldn't cause an issue since later on, users do a reset over all the environments
-                # so the lazy buffers would be reset.
-                self.scene.update(dt=self.physics_dt)
-            # add timeline event to load managers
-            self.load_managers()
+        print("[INFO]: Starting the simulation. This may take a few seconds. Please wait...")
```
AntoineRichard, line 160: Use the logger

source/isaaclab/isaaclab/sensors/sensor_base.py
```diff
@@ -14,16 +14,16 @@
...
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
-import torch
-
 import omni.kit.app
 import omni.timeline
-from isaacsim.core.simulation_manager import IsaacEvents, SimulationManager
+from isaaclab.sim.physx_simulation_manager import PhysxSimulationManager as SimulationManager
```
AntoineRichard: This should not depend on the PhysXSimulationManager but the base one.

source/isaaclab/isaaclab/sensors/sensor_base.py
```diff
@@ -302,8 +301,8 @@ def _initialize_callback(self, event):
             try:
                 self._initialize_impl()
             except Exception as e:
-                if builtins.ISAACLAB_CALLBACK_EXCEPTION is None:
-                    builtins.ISAACLAB_CALLBACK_EXCEPTION = e
+                if builtins.ISAACLAB_CALLBACK_EXCEPTION is None:  # type: ignore
+                    builtins.ISAACLAB_CALLBACK_EXCEPTION = e  # type: ignore
```
AntoineRichard: Why do we use builtins here?

source/isaaclab/isaaclab/sensors/sensor_base.py
AntoineRichard: This call should be doable on the base manager

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    def enable_stage_open_callback(cls, enable: bool) -> None:
+        """Enable or disable stage open tracking.
+
+        This is optional and may be overridden by backend implementations
+        that need to track stage/scene changes.
+
+        Args:
+            enable: Whether to enable stage open tracking.
+        """
+        pass  # Default no-op, override in subclass if needed
```
AntoineRichard: Are we keeping the callbacks here? Shouldn't they move to the simulation context?

AntoineRichard: I don't think we should keep the callbacks here. enable_stage_open_callback sounds very IsaacSim centric, not even USD, or PhysX.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    @abstractmethod
+    def dispatch_prim_deletion(cls, prim_path: str) -> None:
+        """Dispatch prim deletion event.
+
+        Args:
+            prim_path: The USD path of the prim being deleted.
+                       Use "/" to signal clearing all prims.
+        """
+        raise NotImplementedError
+
```
AntoineRichard: Is that really a base method?

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+        Args:
+            backend: Either "torch" or "warp".
+
+        Raises:
+            ValueError: If backend is not "torch" or "warp".
+        """
+        if backend not in ("torch", "warp"):
+            raise ValueError(f"Backend must be 'torch' or 'warp', got '{backend}'")
+        cls._backend = backend
```
AntoineRichard: I don't think that's needed.

source/isaaclab/isaaclab/sim/simulation_manager.py
AntoineRichard, line 1: Is there a step function here? How would you go about configuring the physics backend?
I think it's missing configure option, that would eat a config (similar to the PhysX config in the simulation context cfg), as well as a method to set a bunch of physical settings through a dict.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+from __future__ import annotations
+
+import weakref
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Any, Callable
+
+if TYPE_CHECKING:
+    from .simulation_events import SimulationEvent
+
+class SimulationManagerBase(ABC):
```
AntoineRichard: Could it be named BaseSimulationManager? :)

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+        _callbacks: Registry of user-registered callbacks (id -> subscription).
+        _callback_id: Counter for generating unique callback IDs.
+        _handles: Named internal handles for subscriptions.
+    """
+
+    # ------------------------------------------------------------------
+    # Shared State (backend-agnostic)
+    # ------------------------------------------------------------------
+
+    _backend: str = "torch"
```
AntoineRichard: I wouldn't keep that since we will not allow a switch. Also that's TensorAPI specific!

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    _warmup_needed: bool = True
```
AntoineRichard: Isn't that PhysX specific?

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+
+import carb
+import omni.kit
+import omni.physics.tensors
+import omni.physx
+import omni.timeline
+import omni.usd
+from pxr import PhysxSchema
+
+from .simulation_manager import SimulationManagerBase
```
AntoineRichard: BaseSimulationManager!

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+    _physx_sim_interface = omni.physx.get_physx_simulation_interface()
+    _carb_settings = carb.settings.get_settings()
+
+    # ------------------------------------------------------------------
+    # Omniverse-specific State
+    # ------------------------------------------------------------------
+
+    # Simulation views (PhysX tensor API)
+    _view: omni.physics.tensors.SimulationView | None = None
+    _view_warp: omni.physics.tensors.SimulationView | None = None
```
AntoineRichard: I don't think that's needed?

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+        cls._handles.clear()
+        cls._callbacks.clear()
+
+        # Invalidate views before clearing
+        if cls._view:
+            cls._view.invalidate()
+            cls._view = None
+        if cls._view_warp:
+            cls._view_warp.invalidate()
+            cls._view_warp = None
```
AntoineRichard: This should not be needed

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+        return cls._view
+
+    @classmethod
+    def get_warp_sim_view(cls) -> omni.physics.tensors.SimulationView | None:
+        """Get the warp simulation view (Omniverse-specific).
+
+        Returns:
+            The warp simulation view, or None if not created.
+        """
+        return cls._view_warp
```
AntoineRichard: Should not be needed

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+
+    @classmethod
+    def _is_gpu_enabled(cls) -> bool:
+        """Check if GPU dynamics is enabled."""
+        if cls._physics_scene_apis:
+            api = list(cls._physics_scene_apis.values())[0]
+            bp = api.GetBroadphaseTypeAttr().Get()
+            gpu_dyn = api.GetEnableGPUDynamicsAttr().Get()
+            return bp == "GPU" and gpu_dyn
+        return False
```
AntoineRichard: Do we need this?

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+            if not api.GetPrim().IsValid():
+                continue
+            if api.GetBroadphaseTypeAttr().Get() is None:
+                api.CreateBroadphaseTypeAttr(bp_type)
+            else:
+                api.GetBroadphaseTypeAttr().Set(bp_type)
+            if api.GetEnableGPUDynamicsAttr().Get() is None:
+                api.CreateEnableGPUDynamicsAttr(enable)
+            else:
+                api.GetEnableGPUDynamicsAttr().Set(enable)
```
AntoineRichard: That should be done using a config method. And it should allow setting all the physX settings.

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
```diff
@@ -0,0 +1,435 @@
...
+        mgr = omni.kit.app.get_app().get_extension_manager()
+        was_enabled = mgr.is_extension_enabled("omni.physx.fabric")
+        if enable and not was_enabled:
+            mgr.set_extension_enabled_immediate("omni.physx.fabric", True)
+        elif not enable and was_enabled:
+            mgr.set_extension_enabled_immediate("omni.physx.fabric", False)
+        cls._carb_settings.set_bool("/physics/updateToUsd", not enable)
+        cls._carb_settings.set_bool("/physics/updateParticlesToUsd", not enable)
+        cls._carb_settings.set_bool("/physics/updateVelocitiesToUsd", not enable)
+        cls._carb_settings.set_bool("/physics/updateForceSensorsToUsd", not enable)
```
AntoineRichard: I think this should be done in the config.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    _view_created: bool = False
```
AntoineRichard: I would replace that with a more general name, something like `_initialized`. Newton for instance does not create a tensorAPI like structure for the whole simulation.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    _assets_loaded: bool = True
+
+    # Callback registry: id -> subscription handle
+    _callbacks: dict[int, Any] = {}
+    _callback_id: int = 0
+    _handles: dict[str, Any] = {}
```
AntoineRichard: Do we want callbacks here. IMO no.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @classmethod
+    def is_warmup_needed(cls) -> bool:
+        """Check if physics warmup is needed."""
+        return cls._warmup_needed
+
+    @classmethod
+    def is_view_created(cls) -> bool:
+        """Check if simulation views have been created."""
+        return cls._view_created
```
AntoineRichard: the same things about initialization

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+        """Warm-start physics and create simulation views.
+
+        This should:
+        - Force load physics from USD/scene
+        - Start the simulation engine
+        - Perform initial simulation step
+        - Dispatch PHYSICS_WARMUP event
+        - Create simulation views
+        - Dispatch SIMULATION_VIEW_CREATED and PHYSICS_READY events
+        """
```
AntoineRichard: The description is very PhysX centric.
Something along the lines of could do:
`Initializes the physics engine so that it's ready to run the simulation. After this call the simulation should be ready to go`

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+
+    @classmethod
+    @abstractmethod
+    def get_physics_sim_view(cls) -> Any | None:
+        """Get the physics simulation view.
+
+        Returns:
+            The simulation view for tensor operations, or None if not created.
+        """
+        raise NotImplementedError
```
AntoineRichard: That's the tensor API right? I don't think we should have that in the base API.

source/isaaclab/isaaclab/sim/simulation_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    def set_physics_sim_device(cls, device: str) -> None:
+        """Set the physics simulation device.
+
+        Args:
+            device: Device string like "cpu" or "cuda:0".
+
+        Raises:
+            ValueError: If device string is invalid.
+        """
+        raise NotImplementedError
```
AntoineRichard: Do we need two methods for that? Could we only have one? Also does it make sense to have this method now that we explicitly set it?

source/isaaclab/isaaclab/sim/physx_simulation_manager.py
AntoineRichard, line 1: Similarly to the other one it's missing the step function, the physx setup. I don't think the callbacks should be there.

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -3,46 +3,47 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+from __future__ import annotations
+
 import builtins
```
AntoineRichard: Do we need builtins?

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -245,73 +253,122 @@ def __init__(self, cfg: SimulationCfg | None = None):
...
-        if sim_params is not None:
-            if "physx" in sim_params:
-                physx_params = sim_params.pop("physx")
-                sim_params.update(physx_params)
+        # obtain interfaces for simulation
+        self._app_iface = omni.kit.app.get_app_interface()
+        self._framework = carb.get_framework()
+        self._physx_iface = omni.physx.get_physx_interface()
+        self._physx_sim_iface = omni.physx.get_physx_simulation_interface()
+        self._timeline_iface = omni.timeline.get_timeline_interface()
```
AntoineRichard: This should not be here since it's already in the manager?

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -497,38 +588,106 @@ def get_setting(self, name: str) -> Any:
...
+            # disable app control on stop handle
+            self._disable_app_control_on_stop_handle = True
+            if not self.is_stopped():
+                self.stop()
+            self._disable_app_control_on_stop_handle = False
+            # play the simulation
+            self.play()
+            # initialize the physics simulation
+            if SimulationManager.get_physics_sim_view() is None:
+                SimulationManager.initialize_physics()
```
AntoineRichard: SHould the SimulationManager have a reset?

AntoineRichard: (I think we may need that method)

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -711,6 +881,113 @@ def _apply_physics_settings(self):
...
+                " number of velocity iterations (setting it to 1 or 2 rather than 0), together with this flag, to"
+                " improve the accuracy of velocity updates."
+            )
+
+        if not self.cfg.physx.enable_stabilization and self.cfg.dt > 0.0333:
+            self.logger.warning(
+                "Large simulation step size (> 0.0333 seconds) is not recommended without enabling stabilization."
+                " Consider setting the `enable_stabilization` flag to True in the PhysxCfg, or reducing the"
+                " simulation step size if you run into physics issues."
+            )
```
AntoineRichard: All that can go in the SimulationManager of PhysX

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -623,68 +774,87 @@ def render(self, mode: RenderMode | None = None):
...
+        # how many physics steps to perform per rendering step.
+        # it is not used if step(render=False).
+        render_interval = max(self.cfg.render_interval, 1)
+
+        # set simulation step per second
+        steps_per_second = int(1.0 / self.cfg.dt)
+        self._physx_scene_api.CreateTimeStepsPerSecondAttr(steps_per_second)
+        # set minimum number of steps per frame
+        min_steps = int(steps_per_second / render_interval)
+        self.carb_settings.set_int("/persistent/simulation/minFrameRate", min_steps)
```
AntoineRichard: That could also go in the SImulationManager

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -573,10 +726,14 @@ def step(self, render: bool = True):
...
             # FIXME: This steps physics as well, which we is not good in general.
-            self.app.update()
+            self._app_iface.update()
 
         # step the simulation
-        super().step(render=render)
+        if render:
+            self._app_iface.update()
+        else:
+            self._physx_sim_iface.simulate(self.cfg.dt, 0.0)
```
AntoineRichard: PhysXSimulationManager step

source/isaaclab/isaaclab/sim/simulation_context.py
AntoineRichard, line 1: I need more time to review that one completely. But I would be in favor of taking it in, and then working on decoupling all the kit related calls.

AntoineRichard:
Also this should probably target develop?

apps/isaaclab.python.kit
AntoineRichard, line 1: Should we remove the isaacsim_4_5 apps or do we plan on supporting those still?

apps/isaacsim_4_5/isaaclab.python.kit
AntoineRichard, line 1: Should we remove the isaacsim_4_5 apps or do we plan on supporting those still?

apps/isaacsim_4_5/isaaclab.python.kit
```diff
@@ -17,7 +17,7 @@ keywords = ["experience", "app", "usd"]
 "isaacsim.core.api" = {}
 "isaacsim.core.cloner" = {}
 "isaacsim.core.nodes" = {}
-"isaacsim.core.simulation_manager" = {}
+# "isaacsim.core.simulation_manager" = {}  # Replaced by isaaclab.sim.simulation_manager
```
AntoineRichard: This can just be removed?

apps/isaaclab.python.kit
```diff
@@ -17,7 +17,7 @@ keywords = ["experience", "app", "usd"]
 "isaacsim.core.api" = {}
 "isaacsim.core.cloner" = {}
 "isaacsim.core.nodes" = {}
-"isaacsim.core.simulation_manager" = {}
+# "isaacsim.core.simulation_manager" = {}  # Replaced by isaaclab.sim.simulation_manager
```
AntoineRichard: This could be removed directly?

source/isaaclab/isaaclab/assets/asset_base.py
```diff
@@ -296,7 +295,7 @@ def safe_callback(callback_name, event, obj_ref):
             order=10,
         )
         # register prim deletion callback
-        self._prim_deletion_callback_id = SimulationManager.register_callback(
+        self._prim_deletion_callback_id = PhysxManager.register_callback(
```
AntoineRichard: I think the callbacks should live in the simulation context. It's something that we could take a look at in an other PR.

source/isaaclab/isaaclab/sim/__init__.py
```diff
@@ -28,8 +28,30 @@
...
+    # These callbacks were already set up when the extension loaded, so we need to disable them
+    _OriginalSimManager.enable_all_default_callbacks(False)
+
+    # Replace the class in both the module and impl module
+    _isaacsim_sim_manager.PhysxManager = PhysxManager
+    _isaacsim_sim_manager.IsaacEvents = IsaacEvents
+    _isaacsim_sim_manager_impl.SimulationManager = PhysxManager
+    _isaacsim_sim_manager_impl.IsaacEvents = IsaacEvents
+except ImportError:
+    pass  # Isaac Sim extension not loaded yet, will be patched when available
```
AntoineRichard, line 57: Do we really need to patch IsaacSim? That's not clear to me why.

source/isaaclab/isaaclab/visualizers/rerun_visualizer.py
AntoineRichard, line 1: This is newton only. Do we want to keep this in this PR?

source/isaaclab/isaaclab/visualizers/rerun_visualizer_cfg.py
AntoineRichard, line 1: This is newton only ATM, do we want to keep this in this PR?

source/isaaclab/isaaclab/visualizers/newton_visualizer.py
AntoineRichard, line 1: This is Newton only do we want to keep this in this PR.

source/isaaclab/isaaclab/visualizers/newton_visualizer_cfg.py
AntoineRichard, line 1: This is Newton only, do we want to keep this in this PR?

source/isaaclab/isaaclab/visualizers/physx_ov_visualizer.py
AntoineRichard, line 1: It's not clear to me why this is needed if you mock the IsaacSim classes in the first place?

source/isaaclab/isaaclab/visualizers/physx_ov_visualizer_cfg.py
AntoineRichard, line 1: Same

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
...
+        self._was_playing = False
+
+        # Initialize visualizers immediately
+        self.initialize_visualizers()
+
+    # -- Properties --
+
+    @property
+    def settings(self):
+        return self._sim.carb_settings
```
AntoineRichard: Could that be an abstract method? I feel like not everything will use carb_settings even when it comes to rendering.

source/isaaclab/isaaclab/sim/interface/interface.py
AntoineRichard, line 1: I'm not sure this abstraction is needed. I don't see a need to unify the physics things and the rendering things under the same umbrella.

source/isaaclab/isaaclab/sim/interface/interface.py
```diff
@@ -0,0 +1,65 @@
...
+        pass
+
+    @abstractmethod
+    def step(self, render: bool = True) -> None:
+        """Step the subsystem by one timestep.
+
+        Args:
+            render: Whether to render after stepping. Defaults to True.
+        """
+        pass
```
AntoineRichard, line 52: Like a render technically doesn't step.

source/isaaclab/isaaclab/sim/interface/interface.py
```diff
@@ -0,0 +1,65 @@
...
+        """Clean up resources."""
+        pass
+
+    def play(self) -> None:
+        """Handle simulation start."""
+        pass
+
+    def stop(self) -> None:
+        """Handle simulation stop."""
+        pass
```
AntoineRichard, line 69: I don't think those are used by the physics_manager for instance.

source/isaaclab/isaaclab/sim/interface/interface.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Args:
+            soft: If True, skip full reinitialization.
+        """
+        pass
+
+    @abstractmethod
+    def forward(self) -> None:
+        """Update state without stepping simulation."""
+        pass
```
AntoineRichard, line 43: A renderer should not have a forward.

source/isaaclab/isaaclab/sim/interface/physics_interface.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        super().__init__(sim_context)
+
+        # Get config reference
+        self._cfg = self._sim.cfg.physics_manager_cfg
+        self.physics_prim_path = self._cfg.physics_prim_path
+        self.backend = "torch"
+
+        # Get the physics manager class from config
+        self.physics_manager: type[PhysicsManager] = self._cfg.create_manager()
```
AntoineRichard: `self.physics_manager: type[PhysicsManager] = self._cfg.create_manager()` --> `self.physics_manager: 
PhysicsManager = self._cfg.create_manager()`

source/isaaclab/isaaclab/sim/interface/physics_interface.py
AntoineRichard, line 1: Is that class really needed if it justs pipes into the physics manager base class? I don't see the value of having this abstraction here.

The only thing that would justify this are here. But then I feel like this could live in the simulation context all the same?

```python
    def is_fabric_enabled(self) -> bool:
        """Returns whether the fabric interface is enabled."""
        return self.physics_manager.is_fabric_enabled()

    def _init_usd_physics_scene(self) -> None:
        """Create and configure the USD physics scene."""
        stage = self._sim.stage
        with sim_utils.use_stage(stage):
            # Set stage conventions for metric units
            UsdGeom.SetStageUpAxis(stage, "Z")
            UsdGeom.SetStageMetersPerUnit(stage, 1.0)
            UsdPhysics.SetStageKilogramsPerUnit(stage, 1.0)

            # Find and delete any existing physics scene
            for prim in stage.Traverse():
                if prim.GetTypeName() == "PhysicsScene":
                    sim_utils.delete_prim(prim.GetPath().pathString, stage=stage)

            # Create a new physics scene
            if stage.GetPrimAtPath(self._cfg.physics_prim_path).IsValid():
                raise RuntimeError(f"A prim already exists at path '{self._cfg.physics_prim_path}'.")

            physics_scene = UsdPhysics.Scene.Define(stage, self._cfg.physics_prim_path)

            # Pre-create gravity tensor to avoid torch heap corruption issues (torch 2.1+)
            gravity = torch.tensor(self._cfg.gravity, dtype=torch.float32, device=self._cfg.device)
            gravity_magnitude = torch.norm(gravity).item()

            # Avoid division by zero
            if gravity_magnitude == 0.0:
                gravity_direction = [0.0, 0.0, -1.0]
            else:
                gravity_direction = (gravity / gravity_magnitude).tolist()

            physics_scene.CreateGravityDirectionAttr(Gf.Vec3f(*gravity_direction))
            physics_scene.CreateGravityMagnitudeAttr(gravity_magnitude)
```

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
...
+        """Sync scene data and step all active visualizers.
+
+        Args:
+            dt: Time step in seconds (0.0 for kinematics-only).
+        """
+        if self._scene_data_provider:
+            self._scene_data_provider.update()
+
+        if not self._visualizers:
+            return
```
AntoineRichard: This is mostly meant for "forward kinematics". I would just call that what it is, update transforms or whatnot.

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
...
+        """Check whether the simulation is stopped."""
+        for viz in self.visualizers:
+            return not viz.is_running()
+        return False
+
+    def forward(self) -> None:
+        """Sync scene data and step all active visualizers.
+
+        Args:
+            dt: Time step in seconds (0.0 for kinematics-only).
```
AntoineRichard: No dt here

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
...
+        for viz_type in requested:
+            if viz_type in type_map:
+                try:
+                    configs.append(type_map[viz_type]())
+                except Exception as e:
+                    logger.error(f"Failed to create default config for '{viz_type}': {e}")
+            else:
+                logger.warning(f"Unknown visualizer type '{viz_type}'. Valid: {list(type_map.keys())}")
+
+        return configs
```
AntoineRichard: That coupling looks quite convoluted...

source/isaaclab/isaaclab/sim/physx_ov_visualizer_cfg.py
AntoineRichard, line 1: Why do we need a special coupling for this?

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
...
+            except Exception as e:
+                logger.error(f"Error closing visualizer: {e}")
+
+    def get_rendering_dt(self) -> float:
+        """Get rendering dt from visualizers, or fall back to physics dt."""
+        for viz in self._visualizers:
+            dt = viz.get_rendering_dt()
+            if dt is not None:
+                return dt
+        return self.dt
```
AntoineRichard: self.dt is not private, it's a bit of a NIT, but I feel like if there is a self.dt, why would I call get_rendering_dt. Also why not use a property here?

```python
    @property
    def rendering_dt(self) -> float:
        """Get rendering dt from visualizers, or fall back to physics dt * render_interval."""
        for viz in self._visualizers:
            dt = viz.get_rendering_dt()
            if dt is not None:
                return dt
        return self._dt
``

File: source/isaaclab/isaaclab/sim/interface/visualizer_interface.py [hunk 48]
```diff
@@ -0,0 +1,255 @@
...
+        for viz in to_remove:
+            try:
+                viz.close()
+                self._visualizers.remove(viz)
+                logger.info(f"Removed: {type(viz).__name__}")
+            except Exception as e:
+                logger.error(f"Error closing visualizer: {e}")
+
+    def get_rendering_dt(self) -> float:
+        """Get rendering dt from visualizers, or fall back to physics dt."""
```
AntoineRichard: It's not physics dt it's physics dt * render interval?

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
AntoineRichard, line 1: In general i'm not entirely sure this class is needed. Since there is base visualizer it could make more sense to have them straight into the simulation context? Also, would this support multi-visualizer things?

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -389,29 +196,13 @@ class SimulationCfg:
...
-        running under the hood.
+    This configuration determines which physics manager to use. Override with
+    a different config (e.g., NewtonManagerCfg) to use a different physics backend.
     """
 
-    physx: PhysxCfg = PhysxCfg()
-    """PhysX solver settings. Default is PhysxCfg()."""
-
     physics_material: RigidBodyMaterialCfg = RigidBodyMaterialCfg()
     """Default physics material settings for rigid bodies. Default is RigidBodyMaterialCfg().
```
AntoineRichard: Not sure if we should keep this here. It feels out of place.

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -385,8 +224,29 @@ def get_version(self) -> tuple[int, int, int]:
         """
         return get_isaac_sim_version().major, get_isaac_sim_version().minor, get_isaac_sim_version().micro
```
AntoineRichard: I feel like this shoudn't be in the simulation context. That's an IsaacSim's problem. Soon we'll only depend on PhysX / Kit / Newton versions.

source/isaaclab/isaaclab/physics/newton_manager.py
AntoineRichard, line 1: Let's not have that here. I understand that it was used for testing, but let's keep things pure PhysX for now.

source/isaaclab/isaaclab/physics/newton_manager_cfg.py
AntoineRichard, line 1: Let's not have that here

source/isaaclab/isaaclab/physics/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Implementation backends for simulation interfaces."""
+
+from .physics_manager import PhysicsManager
+from .physics_manager_cfg import PhysicsManagerCfg
+from .physx_manager import PhysxManager, IsaacEvents
+from .physx_manager_cfg import PhysxManagerCfg
+# from .newton_manager import NewtonManager
+# from .newton_manager_cfg import NewtonManagerCfg
```
AntoineRichard, line 13: Let's delete.

source/isaaclab/isaaclab/physics/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+__all__ = [
+    "PhysicsManager",
+    "PhysicsManagerCfg",
+    "PhysxManager",
+    "IsaacEvents",
+    "PhysxManagerCfg",
+    # "NewtonManager",
+    # "NewtonManagerCfg",
```
AntoineRichard, line 25: This too!

source/isaaclab/isaaclab/visualizers/__init__.py
AntoineRichard, line 1: In general let's try to have only things that we can use here.

scripts/environments/state_machine/open_cabinet_sm.py
```diff
@@ -289,7 +293,7 @@ def main():
     desired_orientation = torch.zeros((env.unwrapped.num_envs, 4), device=env.unwrapped.device)
     desired_orientation[:, 1] = 1.0
     # create state machine
-    open_sm = OpenDrawerSm(env_cfg.sim.dt * env_cfg.decimation, env.unwrapped.num_envs, env.unwrapped.device)
+    open_sm = OpenDrawerSm(env_cfg.sim.physics_manager_cfg.dt * env_cfg.decimation, env.unwrapped.num_envs, env.unwrapped.device)
```
kellyguo11, line 296: I think I like retrieving the attributes from env_cfg.sim.dt directly a bit more - would it make sense to still keep some common attributes in the SimulationCfg? probably also less breaking changes for user scripts

source/isaaclab/isaaclab/physics/newton_manager.py
```diff
@@ -0,0 +1,551 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Newton physics manager for Isaac Lab."""
```
kellyguo11, line 6: is this being added here to also refactor it to match the new PhysicsManager interface? I'm not sure if we anticipate more changes for this in the dev/newton branch before merging things in.

source/isaaclab/isaaclab/physics/physx_manager.py
```diff
@@ -0,0 +1,867 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""PhysX Manager for Isaac Lab.
```
kellyguo11, line 6: I think this would likely be the OV manager, since ovphysx will probably have a different set of APIs

source/isaaclab/isaaclab/sim/interface/interface.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .simulation_context import SimulationContext
+
+
+class Interface(ABC):
```
kellyguo11, line 17: this does feel maybe a bit too abstracted, perhaps we can just combine the interface idea into the SimulationContext?

we can probably go with the approach we talked about today where we design the classes with interface in mind, but flatten them out when we do the implementation.

source/isaaclab/isaaclab/sim/interface/visualizer_interface.py
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Visualizer interface for SimulationContext."""
```
kellyguo11: would be good for @matthewtrepte to provide feedback for the visualizers

source/isaaclab/isaaclab/sim/utils/__init__.py
```diff
@@ -5,9 +5,23 @@
...
+
 from .legacy import *  # noqa: F401, F403
 from .prims import *  # noqa: F401, F403
 from .queries import *  # noqa: F401, F403
 from .semantics import *  # noqa: F401, F403
 from .stage import *  # noqa: F401, F403
 from .transforms import *  # noqa: F401, F403
+
+
+def raise_callback_exception_if_any() -> None:
```
kellyguo11: I wonder if there's a better way to handle the exceptions if we follow our own callback system? I think this was kind of a hacky way we had to do to force exceptions to be propagated from the OV callbacks

source/isaaclab/isaaclab/sim/__init__.py
```diff
@@ -28,8 +28,30 @@
...
 from .simulation_context import SimulationContext, build_simulation_context  # noqa: F401, F403
+from isaaclab.physics.physx_manager import PhysxManager, IsaacEvents  # noqa: F401, F403
 from .spawners import *  # noqa: F401, F403
 from .utils import *  # noqa: F401, F403
 from .views import *  # noqa: F401, F403
+
+# Monkey-patch Isaac Sim's PhysxManager to use Isaac Lab's implementation
+# This ensures all code (including Isaac Sim internals) uses our manager
+try:
+    import isaacsim.core.simulation_manager as _isaacsim_sim_manager
```
kellyguo11, line 41: hmm...this feels a bit scary haha, is there any way we can avoid referencing Isaac Sim's SimulationManager all together?

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -1079,55 +313,36 @@ def build_simulation_context(
...
         yield sim
 
     except Exception:
-        sim.logger.error(traceback.format_exc())
+        logger.error(traceback.format_exc())
         raise
     finally:
-        if not sim.has_gui():
-            # Stop simulation only if we aren't rendering otherwise the app will hang indefinitely
+        if not sim.get_setting("/isaaclab/has_gui"):
```
kellyguo11, line 391: is the has_gui flag important for non-OV workflows as well?

source/isaaclab/isaaclab/visualizers/newton_visualizer.py
```diff
@@ -0,0 +1,399 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Newton OpenGL Visualizer implementation."""
```
kellyguo11, line 6: let's sync with @matthewtrepte on the visualizer classes since he's been looking at integrating these with the PhysX backend.

kellyguo11: we can probably also bring things in piece by piece so we can keep this PR somewhat smaller

source/isaaclab/isaaclab/sim/__init__.py
ooctipus: I think we want to avoid cases where there is isaacsim manager and isaaclab manager, since both are singeton, just like we point isaaclab stage as isaacsim stage x)

PR #4501, opened 2026-01-31
Adds support to benchmark scripts when Isaac Sim is not running

kellyguo11:
# Description

Allows running benchmark_non_rl, benchmark_rsl_rl, and benchmark_rlgames scripts without launching Isaac Sim. This will allow us to gather similar timing data for startup time and FPS metrics without using the Isaac Sim benchmarking extension.

Without Isaac Sim, we are not able to capture Render and Physics simulation data directly from the scripts, but there's other benchmarking work going on where we can inject timers into the code to capture these.

By default, it will not launch Isaac Sim unless `--kit` is added to the command.

## Type of change

## Checklist


AntoineRichard:
@kellyguo11 I think there is some overlapp with https://github.com/isaac-sim/IsaacLab/pull/4497 wrt to the benchmarking too. I don't think we shoud even rely on IS tools this way we don't have to follow their template / format.

kellyguo11:
> @kellyguo11 I think there is some overlapp with #4497 wrt to the benchmarking too. I don't think we shoud even rely on IS tools this way we don't have to follow their template / format.

I'm good with that as well if we can capture all the necessary data around simulation and rendering. Sounds like Isaac Sim is also making some breaking changes to their benchmarking tool, so we can also save some work to migrate.

PR #4523, opened 2026-02-03
most recent changes

sagarchotalia:
Work on the modification of the UR10e lifting task has now been pushed to the repo. With this we push to main.

Currently, the task is not working as expected -- we have the cube being pushed off the table as a result of exploitation of the lifting reward. This is to be changed and merged in the upcoming weeks :)
Changing the reward function was also tried here. I also added some new rewards (which ended up not working either) and so I discarded them for the sake of simplicity. Also, IsaacLab documentation says to define the task in as few observation and reward terms as possible(Occam's razor).


sagarchotalia:
Opened this PR by mistake from a forked branch; closing immediately. Apologies!

PR #4582, opened 2026-02-10
Adds support for RLinf library

KumoLiu:
# Description

This PR adds [RLinf](https://github.com/RLinf/RLinf.git) as a new reinforcement learning framework integration for IsaacLab, alongside the existing integrations (rsl_rl, skrl, sb3, rl_games).
RLinf is a flexible and scalable open-source RL infrastructure designed for Embodied and Agentic AI. This integration enables reinforcement learning fine-tuning of Vision-Language-Action (VLA) models (e.g., GR00T, OpenVLA) on IsaacLab simulation tasks, leveraging RLinf's FSDP-based distributed training across multiple GPUs/nodes.

## Key capabilities:
- Fine-tune pretrained VLA models on IsaacLab tasks using PPO / Actor-Critic / SAC
- Distributed training via RLinf's FSDP workers (Actor, Rollout, Env)
- Configurable observation/action mapping from IsaacLab to GR00T format via Hydra YAML configs
- Training (train.py / run.sh) and evaluation (play.py / play.sh) entry points following the same pattern as other RL integrations
- Automatic task registration into RLinf without modifying RLinf source code

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_physx/docs/CHANGELOG.rst
+++ b/source/isaaclab_physx/docs/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changelog
 ---------
 
+0.4.1 (2026-02-18)
+~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed a bug in :meth:~isaaclab_physx.assets.Articulation.process_actuators_cfg where explicit actuator joints could receive non-zero PhysX stiffness/damping, causing double PD control.
+
 
 0.4.0 (2026-02-13)
 ~~~~~~~~~~~~~~~~~~
--- a/source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
+++ b/source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
@@ -3531,30 +3531,6 @@ class Articulation(BaseArticulation):
             )
             # store actuator group
             self.actuators[actuator_name] = actuator
-            # set the passed gains and limits into the simulation
-            if isinstance(actuator, ImplicitActuator):
-                self._has_implicit_actuators = True
-                # the gains and limits are set into the simulation since actuator model is implicit
-                self.write_joint_stiffness_to_sim_index(actuator.stiffness, joint_ids=actuator.joint_indices)
-                self.write_joint_damping_to_sim_index(actuator.damping, joint_ids=actuator.joint_indices)
-            else:
-                # the gains and limits are processed by the actuator model
-                # we set gains to zero, and torque limit to a high value in simulation to avoid any interference
-                self.write_joint_stiffness_to_sim_index(0.0, joint_ids=actuator.joint_indices)
-                self.write_joint_damping_to_sim_index(0.0, joint_ids=actuator.joint_indices)
-
-            # Set common properties into the simulation
-            self.write_joint_effort_limit_to_sim_index(actuator.effort_limit_sim, joint_ids=actuator.joint_indices)
-            self.write_joint_velocity_limit_to_sim_index(actuator.velocity_limit_sim, joint_ids=actuator.joint_indices)
-            self.write_joint_armature_to_sim_index(actuator.armature, joint_ids=actuator.joint_indices)
-            self.write_joint_friction_coefficient_to_sim_index(actuator.friction, joint_ids=actuator.joint_indices)
-            self.write_joint_dynamic_friction_coefficient_to_sim_index(
-                actuator.dynamic_friction, joint_ids=actuator.joint_indices
-            )
-            self.write_joint_viscous_friction_coefficient_to_sim_index(
-                actuator.viscous_friction, joint_ids=actuator.joint_indices
-            )
-
             # Store the configured values from the actuator model
             # note: this is the value configured in the actuator model (for implicit and explicit actuators)
             joint_ids = actuator.joint_indices
@@ -3643,6 +3619,29 @@ class Articulation(BaseArticulation):
                     self.data._joint_viscous_friction_coeff,
                 ],
                 device=self.device,
+            )
+            # set the passed gains and limits into the simulation
+            if isinstance(actuator, ImplicitActuator):
+                self._has_implicit_actuators = True
+                # the gains and limits are set into the simulation since actuator model is implicit
+                self.write_joint_stiffness_to_sim_index(actuator.stiffness, joint_ids=actuator.joint_indices)
+                self.write_joint_damping_to_sim_index(actuator.damping, joint_ids=actuator.joint_indices)
+            else:
+                # the gains and limits are processed by the actuator model
+                # we set gains to zero, and torque limit to a high value in simulation to avoid any interference
+                self.write_joint_stiffness_to_sim_index(0.0, joint_ids=actuator.joint_indices)
+                self.write_joint_damping_to_sim_index(0.0, joint_ids=actuator.joint_indices)
+
+            # Set common properties into the simulation
+            self.write_joint_effort_limit_to_sim_index(actuator.effort_limit_sim, joint_ids=actuator.joint_indices)
+            self.write_joint_velocity_limit_to_sim_index(actuator.velocity_limit_sim, joint_ids=actuator.joint_indices)
+            self.write_joint_armature_to_sim_index(actuator.armature, joint_ids=actuator.joint_indices)
+            self.write_joint_friction_coefficient_to_sim_index(actuator.friction, joint_ids=actuator.joint_indices)
+            self.write_joint_dynamic_friction_coefficient_to_sim_index(
+                actuator.dynamic_friction, joint_ids=actuator.joint_indices
+            )
+            self.write_joint_viscous_friction_coefficient_to_sim_index(
+                actuator.viscous_friction, joint_ids=actuator.joint_indices
             )
 
         # perform some sanity checks to ensure actuators are prepared correctly


KumoLiu:
Hi @kellyguo11, as discussed offline, I've created this separate PR for the core logic changes needed to integrate RLinf into IsaacLab. This covers the training/evaluation entry points and the extension module for task registration and obs/action conversion. Could you please help review it? Thanks.

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,180 @@
...
+"""Rest of the script - launch RLinf evaluation."""
+
+import logging
+import torch.multiprocessing as mp
+
+from hydra import compose, initialize_config_dir
+from hydra.core.global_hydra import GlobalHydra
+from omegaconf import OmegaConf, open_dict
+
+from rlinf.config import validate_cfg
```
kellyguo11: what are the dependencies that will need to be added for rlinf? we'll also need to update setup.py to add the specific rlinf dependencies. if they are large, it's probably best that we don't pull them in by default unless the user explicitly installs isaaclab_rl[rlinf]

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def get_config_path_and_name(args_cli) -> tuple[Path, str]:
+    """Get config path and name.
+    
+    Priority:
+    1. RLINF_CONFIG_FILE environment variable (full path, set by play.sh)
+    2. CLI --config_name argument (looks in rlinf/config directory)
+    """
+    config_file = os.environ.get("RLINF_CONFIG_FILE", "")
```
kellyguo11: is this something that's set by rlinf? by does it need to be an environment variable?

scripts/reinforcement_learning/rlinf/play.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
```
kellyguo11: is there a reason why we need a separate .sh script instead of running play.py directly?

scripts/reinforcement_learning/rlinf/run.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
```
kellyguo11: similar comment as play.sh, we normally keep to only a single python script to launch training/inference. what are the additional configurations that need to be done through a .sh script?

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
```diff
@@ -0,0 +1,480 @@
...
+    return _isaaclab_cfg_cache
+
+
+def _patch_embodiment_tags(cfg: dict) -> None:
+    """Add custom embodiment tag to RLinf's EmbodimentTag enum and mapping if needed.
+    
+    Reads from YAML config (env.train.isaaclab.embodiment_tag and embodiment_tag_id).
+    Only adds tag if not already in RLinf's native registry.
+    """
+    embodiment_tag = cfg.get("embodiment_tag", "new_embodiment")
```
kellyguo11, line 143: If we expect the config file to follow certain structure, would it make sense to define a base configuration class similar to RSL RL's rl_cfg.py where we provide all the parameters users could specify? might make it more robust for users to provide structured inputs through the config class.

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
```diff
@@ -0,0 +1,480 @@
...
+    tag_id = cfg.get("embodiment_tag_id", 31)
```
kellyguo11, line 144: is there a special meaning to 31?

kellyguo11:
could you please also run `./isaaclab.sh -f` to format the code?

scripts/reinforcement_learning/rlinf/play.sh
KumoLiu: Good point — there's no strong reason to keep the separate .sh scripts. The shell wrappers were originally added to set up PYTHONPATH (since RLinf can't be pip-installed due to its repo layout) and a few environment variables (RLINF_EXT_MODULE, RLINF_CONFIG_FILE, etc.).
I've now moved all of that logic into train.py and play.py directly — the scripts will set up sys.path and os.environ at the top, before any RLinf imports. This makes the usage consistent with other RL frameworks in IsaacLab (rsl_rl, rl_games, skrl, sb3), which all run their .py scripts directly via isaaclab.sh -p.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: `RLINF_CONFIG_FILE` is not a requirement from RLinf — it's defined in extension.py to pass the config file path to RLinf worker processes. The reason it's an environment variable is that extension.py is loaded dynamically by RLinf workers (via `_load_user_extensions()`), which are separate Ray actor processes. There's no direct function-call parameter path from our train.py → RLinf worker → extension.py, so we use an env var as the simplest cross-process communication mechanism.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Hi @kellyguo11, RLinf currently cannot be installed directly via pip install due to its repository layout (multiple top-level packages cause setuptools discovery to fail). For now, users need to clone the RLinf repo and add it to their PYTHONPATH. Where would be the best place to document this setup requirement? currently I try to add a note in the setup.py, do you think it make sense?

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,235 @@
...
+from pathlib import Path
+
+# ---------------------------------------------------------------------------
+# Environment bootstrap — replaces the need for a separate play.sh wrapper.
+# Sets up PYTHONPATH, environment variables, and validates that RLinf is
+# reachable before any heavy imports.
+# ---------------------------------------------------------------------------
+SCRIPT_DIR = Path(__file__).parent.absolute()
+ISAACLAB_DIR = SCRIPT_DIR.parent.parent.parent  # scripts/reinforcement_learning/rlinf -> IsaacLab root
+RLINF_ROOT = "/ws/Code/RLinf-Orca/"  # parent of IsaacLab (where rlinf/ lives)
```
ooctipus: is this a local path or cluster path?

ooctipus:
@KumoLiu Thank you for the PR,  (b˙◁˙ )b
could you please share an example script that we can run and examine the running workflow?

scripts/reinforcement_learning/rlinf/play.py
kellyguo11: ah I see, maybe we can include the steps to install in the documentation. We have a section that walks through each of the RL libraries currently supported - https://isaac-sim.github.io/IsaacLab/main/source/overview/reinforcement-learning/rl_existing_scripts.html, we can also add RLinf as a section and then provide the installation steps for it

source/isaaclab_rl/setup.py
```diff
@@ -48,10 +48,22 @@
...
+    "rlinf": [
+        # RLinf is loaded via PYTHONPATH (see train.py / play.py bootstrap).
+        #   git clone https://github.com/RLinf/RLinf.git
+        #   git checkout 2036b8d2d98dff902f96dde5418ecc589dd1146d
+        # GR00T (Isaac-GR00T) must be installed separately:
+        #   git clone https://github.com/NVIDIA/Isaac-GR00T.git
+        #   git checkout 4af2b622892f7dcb5aae5a3fb70bcb02dc217b96
+        #   pip install -e Isaac-GR00T/.[base] --no-deps
+        #   pip install --no-build-isolation flash-attn==2.7.1.post4
+        "ray[default]==2.47.0",
```
kellyguo11: I guess if the rlinf installation has to happen separately, we can also just include the ray installation with the rest of the steps and avoid just this one dependency in lab.

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,235 @@
...
+cli_args.add_rlinf_args(parser)
+args_cli = parser.parse_args()
+
+# Validate that RLinf is importable
+try:
+    import rlinf  # noqa: F401
+except ImportError:
+    print(f"ERROR: Cannot import rlinf. Expected RLinf repo at: {RLINF_ROOT}/rlinf")
+    print("Please clone RLinf into the parent directory of IsaacLab:")
+    print(f"  git clone https://github.com/RLinf/RLinf.git {RLINF_ROOT}")
```
ClemensSchwarke: Why does this need to be hardcoded?

source/isaaclab_rl/setup.py
ClemensSchwarke: Generally, it would be much better if RLinf could be installed via pip I think.

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
KumoLiu: Yes, 31 has special meaning:
The GR00T Action Head uses category-specific parameters with 32 slots (indices 0-31) for different robot embodiments. Each embodiment ID selects a different set of weights in the encoder/decoder modules.
Index 31 is specifically reserved for NEW_EMBODIMENT - it's the last available slot and is designated for fine-tuning new robots. The other indices (17, 24, 26) are already used by pre-trained embodiments (OXE_DROID, GR1, AGIBOT_GENIE1).
This allows one model to support multiple robots by using different parameter sets for each embodiment.

refer to: https://github.com/NVIDIA/Isaac-GR00T/blob/n1.5-release/gr00t/data/embodiment_tags.py#L43

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: A local path, let me remove it and sorry for confuse here.

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
KumoLiu: I try to add a rl_cfg.py here like this: https://github.com/isaac-sim/IsaacLab/pull/4582/changes#diff-fcb10de01d5aecc572dfc3976fbf6d3087979d5211ca1546edfce5188e6d5e4c

scripts/reinforcement_learning/rlinf/run.sh
KumoLiu: Removed both bash file.

KumoLiu:
> @KumoLiu Thank you for the PR, (b˙◁˙ )b could you please share an example script that we can run and examine the running workflow?

Sure, I will add a task config and example cmd to show the quick run later.

source/isaaclab_rl/setup.py
KumoLiu: Hi @kellyguo11 and @ClemensSchwarke, thanks for the suggestion! I've discussed this with the RLinf team — they will create a pip-installable package, and we can update the dependency here once that's available.

One remaining concern is around Isaac-GR00T. There are two constraints that cannot be expressed in setup.py:
1. --no-deps — GR00T must be installed with --no-deps to avoid pulling in conflicting versions of packages already bundled with Isaac Sim (e.g. transformers, torch).
2. --no-build-isolation — flash-attn requires this flag so it can see the existing torch/CUDA installation at build time.
Since neither flag can be passed through a standard Python dependency specifier, GR00T currently has to be installed manually. Do you have any suggestions on how to handle this?

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
kellyguo11: I see, maybe we can add a comment for this :)

source/isaaclab_rl/setup.py
kellyguo11: That's great that they are looking into a pip-installable package!

For the GR00T dependencies, maybe we can leverage the installation scripts isaaclab.sh to handle the installation there? I think we'll probably want to first check if the dependency is needed, and then run the installation commands directly from the .sh file with the necessary flags. @myurasov-nv what are your thoughts on this?

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Hi @ClemensSchwarke, it's just a error message here, do you have any concern? I can remove it. Thanks.

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
KumoLiu: Added, thanks.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Hi @kellyguo11, thanks for the suggestion here, I try to update the rst file in this commit: https://github.com/isaac-sim/IsaacLab/pull/4582/commits/fc9c08614967ead3b75b4fd0577290d266650240.
And also move the files under isaaclab_rl to the isaaclab_contri, please help take a look at it again. Thanks.

source/isaaclab_rl/isaaclab_rl/rlinf/extension.py
KumoLiu: Thanks for the suggestion @kellyguo11! I try to add a rl_cfg.py with structured config classes (RLinfIsaacLabCfg, etc.) following RSL-RL's pattern — As I linked above.
However, there's a practical limitation with using it as the primary config source in the RLinf workflow: RLinf uses Ray for distributed training, where the config needs to be shared across multiple worker processes (EnvWorker, RolloutWorker, ActorWorker). These workers are spawned by Ray on potentially different nodes, so we can't pass Python objects directly between them. Additionally, IsaacLab's AppLauncher can only be initialized inside a child process (it takes over the process), so we can't load the config class in the main process either.
Because of these constraints, the runtime config is currently passed via a YAML file (through the RLINF_CONFIG_FILE environment variable), which all workers can read independently.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Removed it to make the code clean.

KumoLiu:
> could you please share an example script that we can run and examine the running workflow?

Hi @ooctipus, since i4h hasn't been released yet, we can't include the task config in this PR. As discussed, I've created a separate PR with the task config here: https://github.com/isaac-sim/IsaacLab-Internal/pull/826
You can merge that PR first, then test this one with:
```
# run training
python train.py --task Isaac-Assemble-Trocar-G129-Dex3-RLinf-v0 --model_path ckpt_path --config_name isaaclab_ppo_gr00t_assemble_trocar

# run playing
python play.py --task Isaac-Assemble-Trocar-G129-Dex3-RLinf-v0 --model_path ckpt_path --config_name isaaclab_ppo_gr00t_assemble_trocar --video --num_episodes 1
```

Thanks.

source/isaaclab_rl/setup.py
KumoLiu: Hi @kellyguo11, I linked the requirements in the setup.py under `isaaclab_contri` here: https://github.com/isaac-sim/IsaacLab/pull/4582/changes#diff-9cfcff087b936e45994034119ff820341ad35ad43cafca5db8cdf53020839169

I was testing in isaaclab 2.3.1 container and besides the ray, all other deps are required by GR00T.

KumoLiu: RLinf has upload a pip-installable version and I have updated it in this commit: [fc033ef](https://github.com/isaac-sim/IsaacLab/pull/4582/commits/fc033ef117a7171b08b2f53a509e3ad3f0aa233c)

scripts/reinforcement_learning/rlinf/cli_args.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    # -- training arguments
+    arg_group.add_argument(
+        "--only_eval", action="store_true", default=False, help="Only run evaluation without training."
+    )
+    # -- cluster arguments
+    arg_group.add_argument("--num_nodes", type=int, default=1, help="Number of nodes for distributed training.")
+
+
+def update_rlinf_cfg(agent_cfg, args_cli: argparse.Namespace):
```
kellyguo11: could you include the return type to the header type hinting as well? we have some style guideline documented here - https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html#coding-style

scripts/reinforcement_learning/rlinf/cli_args.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # Check if it's a dataclass (IsaacLab style) or DictConfig (Hydra style)
+    if is_dataclass(agent_cfg):
+        return _update_dataclass_cfg(agent_cfg, args_cli)
+    else:
+        return _update_omegaconf_cfg(agent_cfg, args_cli)
+
+
+def _update_dataclass_cfg(agent_cfg, args_cli: argparse.Namespace):
+    """Update dataclass configuration (RLinfPPORunnerCfg style)."""
```
kellyguo11: please include the arguments and return types in header and docstrings here as well

scripts/reinforcement_learning/rlinf/cli_args.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # Override cluster settings
+    if args_cli.num_nodes is not None:
+        agent_cfg.cluster.num_nodes = args_cli.num_nodes
+
+    return agent_cfg
+
+
+def _update_omegaconf_cfg(agent_cfg, args_cli: argparse.Namespace):
+    """Update OmegaConf DictConfig configuration (Hydra style)."""
```
kellyguo11: same comment about parameters and return types in header and docstring

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,186 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
```
kellyguo11: we can remove this extra header

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,186 @@
...
+parser.add_argument(
+    "--num_episodes", type=int, default=None, help="Number of evaluation episodes (overrides config if set)."
+)
+parser.add_argument("--video", action="store_true", default=False, help="Enable video recording.")
+cli_args.add_rlinf_args(parser)
+args_cli = parser.parse_args()
+
+# Default config directory (task-specific YAML lives here)
+_DEFAULT_CONFIG_DIR = str(
+    SCRIPT_DIR.parents[2] / "source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config"
```
kellyguo11: is it necessary to hardcode this path here? could we make this a general config that's not specific to the assemble_trocar environment?

scripts/reinforcement_learning/rlinf/play.py
```diff
@@ -0,0 +1,186 @@
...
+)
+_DEFAULT_CONFIG_NAME = "isaaclab_ppo_gr00t_assemble_trocar"
```
kellyguo11: similar to above, can this be an empty string by default?

scripts/reinforcement_learning/rlinf/README.md
```diff
@@ -0,0 +1,307 @@
...
+- A **pretrained VLA checkpoint** in HuggingFace format
+- Multi-GPU setup recommended (FSDP requires at least 1 GPU)
+
+## Quick Start
+
+### Training
+
+```bash
+# Basic training (uses default config)
+python train.py
```
kellyguo11, line 78: will this run headless by default? is it more performant to run with `--headless`?

scripts/reinforcement_learning/rlinf/train.py
```diff
@@ -0,0 +1,218 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
```
kellyguo11: we can remove this duplicated header

scripts/reinforcement_learning/rlinf/train.py
```diff
@@ -0,0 +1,218 @@
...
+Tasks can be either:
+1. Registered in IsaacLab with `rlinf_cfg_entry_point` - will be auto-registered into RLinf
+2. Already registered in RLinf's REGISTER_ISAACLAB_ENVS
+
+Usage:
+    # List available tasks
+    isaaclab.sh -p train.py --list_tasks
+
+    # Train an IsaacLab task
+    isaaclab.sh -p train.py --config_name isaaclab_ppo_gr00t_assemble_trocar
```
kellyguo11: same comment about --headless here, would it be helpful to add that?

scripts/reinforcement_learning/rlinf/train.py
```diff
@@ -0,0 +1,218 @@
...
+parser.add_argument("--list_tasks", action="store_true", default=False, help="List all available tasks and exit.")
+parser.add_argument("--model_path", type=str, default=None, help="Path to pretrained model checkpoint (required).")
+
+# append RLinf cli arguments
+cli_args.add_rlinf_args(parser)
+args_cli = parser.parse_args()
+
+# Default config directory (task-specific YAML lives here)
+_DEFAULT_CONFIG_DIR = str(
+    SCRIPT_DIR.parents[2] / "source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config"
```
kellyguo11: similar to the play.py script, could these be more general task-agnostic paths/names?

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/__init__.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
```
kellyguo11: duplicate header

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
```
kellyguo11: duplicate header

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+
+    _register_gr00t_converters(cfg)
+    _patch_gr00t_get_model(cfg)
+    _register_isaaclab_envs()
+
+    logger.info("isaaclab_contrib.rl.rlinf.extension: Registration complete.")
+
+
+def _load_full_cfg() -> dict:
+    """Load and cache the full YAML config from ``RLINF_CONFIG_FILE``."""
```
kellyguo11: could we include proper docstring with the return type?

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+    if not config_file:
+        raise ValueError("RLINF_CONFIG_FILE not set")
+    with open(config_file) as f:
+        _full_cfg_cache = yaml.safe_load(f)
+    logger.info(f"Loaded full config from {config_file}")
+    return _full_cfg_cache
+
+
+def _get_isaaclab_cfg() -> dict:
+    """Shortcut: return ``env.train.isaaclab`` from the cached full config."""
```
kellyguo11: pls add return type to docstring

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+    return _load_full_cfg().get("env", {}).get("train", {}).get("isaaclab", {})
+
+
+def _patch_embodiment_tags(cfg: dict) -> None:
+    """Add custom embodiment tag to RLinf's EmbodimentTag enum and mapping if needed.
```
kellyguo11, line 126: please add argument to docstring

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+
+    Reads from YAML config (env.train.isaaclab.embodiment_tag and embodiment_tag_id).
+    Only adds tag if not already in RLinf's native registry.
+    """
+    # GR00T uses embodiment tags to identify different robots.  Custom robots
+    # (like G129+Dex3) need a unique tag string and numeric ID so that the
+    # model's tokenizer can map them to the correct action/state dimensions.
+    embodiment_tag = cfg.get("embodiment_tag", "new_embodiment")
+    tag_id = cfg.get("embodiment_tag_id", 31)
```
kellyguo11, line 144: could we mention the exact mapping of what 31 represents in the comments

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+        embodiment_tags.EmbodimentTag = NewEmbodimentTag
+        logger.info(f"Added EmbodimentTag.{tag_upper} = '{embodiment_tag}'")
+
+    # Add to mapping
+    embodiment_tags.EMBODIMENT_TAG_MAPPING[embodiment_tag] = tag_id
+    logger.info(f"Added EMBODIMENT_TAG_MAPPING['{embodiment_tag}'] = {tag_id}")
+
+
+def _patch_gr00t_get_model(cfg: dict) -> None:
+    """Monkeypatch RLinf's GR00T get_model to support custom data_config.
```
kellyguo11: pls add argument to docstring

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+
+        logger.info(f"Loaded GR00T model with embodiment_tag='{model_cfg.embodiment_tag}'")
+        return model
+
+    rlinf_gr00t_mod.get_model = patched_get_model
+    logger.info(f"Patched get_model for data_config_class='{data_config_class}'")
+
+
+def _register_gr00t_converters(cfg: dict) -> None:
+    """Register GR00T obs/action converters for IsaacLab tasks.
```
kellyguo11, line 255: pls add argument to docstring

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+        simulation_io.OBS_CONVERSION[obs_converter_type] = _convert_isaaclab_obs_to_gr00t
+        logger.info(f"Registered obs converter: {obs_converter_type}")
+
+    if obs_converter_type not in simulation_io.ACTION_CONVERSION:
+        simulation_io.ACTION_CONVERSION[obs_converter_type] = _convert_gr00t_to_isaaclab_action
+        logger.info(f"Registered action converter: {obs_converter_type}")
+
+
+def _convert_isaaclab_obs_to_gr00t(env_obs: dict) -> dict:
+    """Convert IsaacLab env observations to GR00T format.
```
kellyguo11, line 278: pls add return type to docstring

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -0,0 +1,476 @@
...
+                    groot_obs[gr00t_key] = states_np[:, :, slice_range[0] : slice_range[1]]
+
+    # Pass through task descriptions
+    groot_obs["annotation.human.action.task_description"] = env_obs.get("task_descriptions", [])
+
+    return groot_obs
+
+
+def _convert_gr00t_to_isaaclab_action(action_chunk: dict, chunk_size: int = 1) -> Any:
+    """Convert GR00T action output to IsaacLab env action format.
```
kellyguo11, line 335: pls add arguments to docstring

kellyguo11:
please also run `./isaaclab.sh -f` to run the formatter

kellyguo11:
Is it possible to also add some unit tests for this? not sure how straight forward it is to do that though

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Update the comments, please help take a look again to see if it make sense, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Added, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
KumoLiu: Removed, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/__init__.py
KumoLiu: Removed, thanks.

scripts/reinforcement_learning/rlinf/train.py
KumoLiu: Removed, thanks.

scripts/reinforcement_learning/rlinf/README.md
KumoLiu: No need to add `--headless`, currently it's the default behaviour, thanks.

scripts/reinforcement_learning/rlinf/train.py
KumoLiu: Same as above, headless is the default behviour currently. thanks.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Updated the logic so the default config directory is the current working directory, and removed `_DEFAULT_CONFIG_NAME`. Users can now specify their config file directly via `--config_name`. Thanks.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Removed this as mentioned above, thanks.

scripts/reinforcement_learning/rlinf/train.py
KumoLiu: Updated as mentioned above, thanks.

scripts/reinforcement_learning/rlinf/play.py
KumoLiu: Removed, thanks.

scripts/reinforcement_learning/rlinf/cli_args.py
KumoLiu: Updated, thanks.

scripts/reinforcement_learning/rlinf/cli_args.py
KumoLiu: Updated, thanks.

scripts/reinforcement_learning/rlinf/cli_args.py
KumoLiu: Updated, thanks.

KumoLiu:
> Is it possible to also add some unit tests for this? not sure how straight forward it is to do that though

Yes, I've added unit tests in source/isaaclab_contrib/test/rl/test_rlinf_extension.py. The rlinf package is fully mocked so tests run without Isaac Sim, GPU, or RLinf installed.
Coverage includes:
- Config loading & caching (_load_full_cfg, _get_isaaclab_cfg)
- Obs conversion: image reshaping, state slicing, task description passthrough
- Action conversion: concatenation, prefix/suffix padding, chunk_size slicing
- Embodiment tag patching (add new / skip existing)
- Task & converter registration (including idempotency)
- Random policy simulation: multi-step roundtrip through the full obs→GR00T→action pipeline, validating shapes, no NaN/Inf, and value preservation across varying batch sizes
Run with: `python -m pytest source/isaaclab_contrib/test/rl/test_rlinf_extension.py -v`

KumoLiu:
> please also run `./isaaclab.sh -f` to run the formatter

Fixed format, thanks.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+# All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
```
kellyguo11:
```suggestion
```

kellyguo11:
Awesome, thanks so much for the addition and all the modifications and tests! will merge this in :)

PR #4608, opened 2026-02-13
Adds support for Newton Warp Rendering with PhysX

daniela-hase:
# Description

Adding support for Newton Warp Rendering with PhysX.

This is adding the Newton Warp Renderer as a new renderer in the `renderers` folder and updates the TiledCamera to optionally render with it.

## Checklist


source/isaaclab/isaaclab/sensors/camera/tiled_camera_cfg.py
```diff
@@ -14,3 +14,4 @@ class TiledCameraCfg(CameraCfg):
     """Configuration for a tiled rendering-based camera sensor."""
 
     class_type: type = TiledCamera
+    renderer: str = "rtx"
```
kellyguo11: would this enforce we only have one renderer running at a time? is it possible to extend the flexibility to allow rendering with both rtx and newton?

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+from pxr import Usd
+
+import isaaclab.sim as isaaclab_sim
+
+if TYPE_CHECKING:
+    from isaaclab.scene import InteractiveScene
+    from isaaclab.sensors import SensorBase
+
+
+class CameraManager:
```
kellyguo11: should we name this with something specific to Newton? like NewtonRenderer or something like that?

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+    @dataclass
+    class CameraData:
+        camera_rays: wp.array(dtype=wp.vec3f, ndim=4) = None
+        outputs: CameraManager.CameraOutputs = field(default_factory=lambda: CameraManager.CameraOutputs())
+        prims: list[Usd.Prim] = field(default_factory=lambda: [])
+        name: str | None = None
+        width: int = 100
+        height: int = 100
+
+    def __init__(self, scene: InteractiveScene):
```
kellyguo11: does this require any additional configurations to apply settings to the sensor? do we need to add a new RenderCfg class for this?

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
```diff
@@ -0,0 +1,221 @@
...
+                            )
+                        )
+
+    def get_camera_transforms(self, camera_data: CameraData) -> wp.array(dtype=wp.transformf):
+        camera_transforms = []
+        for prim in camera_data.prims:
+            camera_transforms.append(self.__resolve_camera_transform(prim))
+        return wp.array([camera_transforms], dtype=wp.transformf)
+
+    def save_images(self, filename: str):
```
kellyguo11: I feel this API is likely not needed here, we have separate utilities for saving images. this should probably be a more user-facing API then hiding it in the renderer class

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -20,10 +20,13 @@
 from isaaclab.sim.views import XformPrimView
 from isaaclab.utils.warp.kernels import reshape_tiled_image
 
+from ...renderers.newton_warp_renderer import NewtonWarpRenderer
```
kellyguo11: We may want to be careful with the import logic. Is it possible to follow the factory style similar to how the Articulation class is structured? We can have a base Render class and a factory Render class, which decides which class to import based on the config.

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -182,68 +187,74 @@ def _initialize_impl(self):
...
-        rp = rep.create.render_product_tiled(cameras=cam_prim_paths, tile_resolution=(self.cfg.width, self.cfg.height))
-        self._render_product_paths = [rp.path]
+        if self.cfg.renderer == "newton":
+            self.renderer = NewtonWarpRenderer(self.scene)
 
-        if any(data_type in self.SIMPLE_SHADING_MODES for data_type in self.cfg.data_types):
-            rep.AnnotatorRegistry.register_annotator_from_aov(
-                aov=self.SIMPLE_SHADING_AOV, output_data_type=np.uint8, output_channels=4
+        else:
+            # Create replicator tiled render product
```
kellyguo11, line 192: hmm I guess we should refactor this part into a KitRTXRender class maybe. and then the Camera class would not have renderer specific logic.

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
daniela-hase: This is just a helper class for the `NewtonWarpRenderer` (defined further down), I can rename it if you like - but ideally no one outside of this module should use it :)

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
daniela-hase: No, currently only the `InteractiveScene`. All required information is retrieved from there :)

source/isaaclab/isaaclab/sensors/camera/tiled_camera_cfg.py
daniela-hase: Currently, yes - but my implementation into the `TiledCamera` wasn't meant as the long term proper solution. More a temporary holdover until we have that proper solution :)

I believe someone (was it Nathan?) is working on an API layer to properly switch between the different supported renderers. So once we have that, my changes to the `TiledCamera` will be obsolete I believe.

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
daniela-hase: Ah yeah - I had that in there just for debugging purposes, so I could save images to confirm they are correct. I meant to remove it before doing the PR, but totally forgot - removed it now :)

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
daniela-hase: See my comment above about the proper long term solution for multiple renderers. I'd think long term this changes to something a bit more abstract.

I can make the change for it to be similar to the Articulation already though :)

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
daniela-hase: Yep, I assumed that was the long term proper plan. Have some abstract API layer that allows us to use any renderer we like, without having to hardcode them all :)

source/isaaclab/isaaclab/sensors/camera/tiled_camera_cfg.py
daniela-hase: PS: Does it make sense to have multiple renderers for the same `TiledCamera` though? In what case would one want to render the same Camera with two or more different renderers? 🤔

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
```diff
@@ -0,0 +1,229 @@
...
+    def __init__(self, render_context: newton.sensors.SensorTiledCamera.RenderContext, sensor: SensorBase):
+        self.render_context = render_context
+        self.sensor = sensor
+        self.num_cameras = 1
+
+        self.camera_rays: wp.array(dtype=wp.vec3f, ndim=4) = None
+        self.camera_transforms: wp.array(dtype=wp.transformf, ndim=2) = None
+        self.outputs = RenderData.CameraOutputs()
+        self.width = getattr(sensor.cfg, "width", 100)
+        self.height = getattr(sensor.cfg, "width", 100)
```
bareya: Typo!  Should be

```
self.height = getattr(sensor.cfg, "height", 100)
```

daniela-hase: Good catch! Fixed, thank you :)

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
```diff
@@ -0,0 +1,229 @@
...
+        if self.physx_to_newton_body_mapping:
+            return
+
+        self.physx_to_newton_body_mapping.clear()
+        for name, articulation in self.scene.articulations.items():
+            articulation_mapping = []
+            for prim in isaaclab_sim.find_matching_prims(articulation.cfg.prim_path):
+                body_indices = []
+                for body_name in articulation.body_names:
+                    prim_path = prim.GetPath().AppendChild(body_name).pathString
```
bareya: Unfortunately, this doesn't work with Dex environment.

This code:

```python
 prim_path = prim.GetPath().AppendChild(body_name).pathString
```

Assumes that each body is nested under articulation's root. Which is not the case in few instances i.e.

For Dex environment the articulation's root is:  `/World/envs/env_0/Robot/` 
The body name is: `allegro_mount`
The full prim path is: `/World/envs/env_0/Robot/allegro_mount`

Unfortunately, this path doesn't exist in usd stage. The real path of it is:

```
/World/envs/env_0/Robot/ee_link/allegro_mount
```

This means body parts are not always nested under the articulation's root.

ooctipus:
```
rigid_body_prims = sim_utils.get_all_matching_child_prims(
    template_prim_path,
    predicate=lambda prim: prim.HasAPI(UsdPhysics.RigidBodyAPI),
    traverse_instance_prims=True,
)
```

Something like this maybe address @bareya 's point

ooctipus:
The code looks good to me! I think it will also be great If you can add an option to run with warp camera to cartpole in this pr. x)

ooctipus (approved):
There are places abstraction could be made better but I like the simplicity and well organization of this PR. Some of the naming style like __<func> could be replace with single _ to be consistent with how lab hands private method in other parts.

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
daniela-hase: I updated the PR to use the SceneDataProvider, so this code no longer exists inside the renderer :)

daniela-hase:
> Some of the naming style like __ could be replace with single _ to be consistent with how lab hands private method in other parts.

I changed all `__` to just `_` :)
Usually I prefer double `__` to somewhat ensure the methods/variables are not accessed from a place where they are not supposed to be. But since it's just a single `_` everywhere in Labs, it makes sense to do the same here for sure.

source/isaaclab/isaaclab/renderers/newton_warp_renderer.py
ooctipus: Thanks Daniela

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -79,7 +81,7 @@ class TiledCamera(Camera):
     cfg: TiledCameraCfg
     """The configuration parameters."""
 
-    def __init__(self, cfg: TiledCameraCfg):
+    def __init__(self, cfg: TiledCameraCfg, renderer: Renderer | None = None):
```
bareya, line 84: We shouldn't pass instance of the `renderer` here. 

Setting `renderer` to default as `None` is not how `InteractiveScene` creation work. The configuration is built first. Then, the concrete instances are built from the configurations. 

One option to solve this problem is something I came up with in my PR #4660 

```python
    base_camera = TiledCameraCfg(
        ...
        renderer_cfg=OVRTXRendererCfg(),
    )
```

I do not explicitly pass strings, but configuration instance.

daniela-hase: Should I implement that in this PR? Or shall we update it to the method of your PR, once your PR lands?

bareya: I would just undo this change. After your change lands I can create MR that handles this through render configuration.

PR #4635, opened 2026-02-17
Add experiment name to be set correctly

ClemensSchwarke:
Add the `experiment_name` CLI argument to `update_rsl_rl_cfg()`, as passing it didn't have any effect before.

## Type of change


AntoineRichard (approved):
LGTM

PR #4660, opened 2026-02-20
Draft: Initial integration of OvRTX

bareya:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/test/renderer/__init__.py
+++ b/source/isaaclab/test/renderer/__init__.py
@@ -1,0 +1,6 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Tests for the renderer package."""
--- a/source/isaaclab/test/renderer/test_ovrtx_render_data.py
+++ b/source/isaaclab/test/renderer/test_ovrtx_render_data.py
@@ -1,0 +1,114 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Unit tests for OVRTX RenderData (buffer creation, shapes)."""
+
+import pytest
+
+# Skip entire module if ovrtx not available (ovrtx_renderer imports it)
+pytest.importorskip("ovrtx")
+
+# Warp/GPU tests may need app context
+from isaaclab.app import AppLauncher
+
+simulation_app = AppLauncher(headless=True).app
+
+"""Rest everything follows."""
+
+import torch
+import warp as wp
+
+from isaaclab.renderer.ovrtx_renderer import OVRTXRenderData
+
+
+@pytest.fixture
+def device():
+    """Use CUDA if available, else skip."""
+    if not torch.cuda.is_available():
+        pytest.skip("CUDA not available")
+    return "cuda:0"
+
+
+class TestOVRTXRenderData:
+    """Tests for OVRTXRenderData buffer creation."""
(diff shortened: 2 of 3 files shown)


source/isaaclab/isaaclab/renderer/ovrtx_usd.py
```diff
@@ -0,0 +1,216 @@
...
+        render_product_name,
+        render_var_path,
+        render_var_name,
+        source_name,
+        tiled_width,
+        tiled_height,
+    )
+    combined_usd = original_usd.rstrip() + "\n\n" + camera_content
+
+    Path("/tmp/ovrtx_test").mkdir(parents=True, exist_ok=True)
```
bareya: This probably needs to be configurable where renderer saves temporary usd file.

source/isaaclab/isaaclab/renderer/ovrtx_renderer_cfg.py
```diff
@@ -0,0 +1,31 @@
...
+class OVRTXRendererCfg(RendererCfg):
+    """Configuration for OVRTX Renderer.
+    
+    The OVRTX renderer uses the ovrtx library for high-fidelity RTX-based rendering.
+    """
+
+    renderer_type: str = "ov_rtx"
+    """Type identifier for OVRTX renderer."""
+    
+    simple_shading_mode: bool = True
```
bareya, line 23: This needs to be more than that, perhaps an enum to support other modes?

source/isaaclab/isaaclab/renderer/ovrtx_renderer.py
```diff
@@ -0,0 +1,622 @@
...
+        
+        Performance: O(1) or O(log N) vs O(N) for traditional loading.
+        """
+        print(f"[OVRTX OPTIMIZE] Cloning base environment to {self._num_envs - 1} targets...")
+        
+        # Source: base environment
+        source_path = "/World/envs/env_0"
+        
+        # Targets: all cloned environments
+        target_paths = [f"/World/envs/env_{i}" for i in range(1, self._num_envs)]
```
kellyguo11: we might need to incorporate this into the Lab cloner plans eventually to handle more complex scene setups like heterogeneous environments

source/isaaclab/isaaclab/renderer/ovrtx_renderer.py
```diff
@@ -0,0 +1,622 @@
...
+                    self._output_data_buffers[buffer_key][env_idx],
+                    tile_x,
+                    tile_y,
+                    self._width,
+                    self._height,
+                ],
+                device=DEVICE,
+            )
+
+    def _extract_depth_tiles(self, tiled_depth_data: wp.array) -> None:
```
kellyguo11: could this and the rgb kernel be merged into 1?

source/isaaclab/isaaclab/renderer/renderer_cfg.py
```diff
@@ -35,8 +35,8 @@ class RendererCfg:
     num_cameras: int = 1
     """Number of cameras to use for rendering. Defaults to 1."""
 
-    data_types: list[str] = MISSING
-    """List of data types to use for rendering."""
+    data_types: list[str] = field(default_factory=list)
+    """List of data types to use for rendering. Overridden by the camera at runtime when used with TiledCameraCfg."""
```
daniela-hase, line 39: Those are defined in the CameraCfg, and therefore different for each Camera. I don't think the Renderer should override the settings like resolution of the Camera.

bareya: Hey @daniela-hase , thanks for looking into this. I agree with you, I was wondering about `data_types` this seems like it should be per camera as well? We need some unified way to define outputs. It seems odd to put that in the renderer configuration because we could have multiple cameras with different outputs defined.

daniela-hase: Yep, `data_types` should be per camera as well I'd think. Not every camera may want to render every aov :)

PR #4677, opened 2026-02-21
Updates CI test setup and fixes camera tests

kellyguo11:
# Description

- Fixes some issues in unit tests
- Fixes rendering in recent visualizer change, which broke the render update
- Moves curobo into its own CI job so that we can switch back to the base container for default lab testing
- Adds assembly/disassembly automate environments to curobo test suite; fixes syntax issues in environments

## Type of change

## Checklist

### DIFF
--- a/tools/test_settings.py
+++ b/tools/test_settings.py
@@ -45,6 +45,7 @@ PER_TEST_TIMEOUTS = {
     "test_visuotactile_sensor.py": 1000,
     "test_visuotactile_render.py": 500,
     "test_rigid_object_collection.py": 1000,
+    "test_outdated_sensor.py": 500,
 }
 """A dictionary of tests and their timeouts in seconds.
 


source/isaaclab/isaaclab/visualizers/kit_visualizer.py
```diff
@@ -73,11 +73,15 @@ def step(self, dt: float) -> None:
         self._sim_time += dt
         self._step_counter += 1
         try:
+            import carb
```
ooctipus: we can use setting manager here

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -474,8 +474,31 @@ def step(self, render: bool = True) -> None:
...
+        and Replicator annotator buffers, unless a visualizer (e.g. KitVisualizer)
+        already does so in its own step().
+        """
         self.update_visualizers(self.get_rendering_dt())
+
+        # Check if a visualizer (e.g. KitVisualizer) already pumps the Kit app
+        # loop in its step(). If so, we skip to avoid a redundant double render.
+        visualizer_pumps_app = any(viz.pumps_app_update() for viz in self._visualizers)
+
+        if self.is_rendering and not visualizer_pumps_app:
```
ooctipus: is this kit visualizer specific? if is, would be better to implement in kit visualizer instead of simulation context

kellyguo11: but that would mean we need to enable the kit visualizer if we want to do rendering? I think eventually this should be moved to the renderer class, but we need that separation first from Piotr and team

ooctipus: i see this is for sensor rendering?
I thought sensor rendering responsibility if compeletely owned by sensor, if thats then, maybe this code needs come from piotr? Or could it be implemented in tiledCamera implementation instead?

kellyguo11: yes it's sensor rendering. I agree it should be refactored out with Piotr's changes, but I don't know what that looks like yet, so I just left it here for now. the issue with putting it in TiledCamera right now is I'm not sure how to handle synchronizing the step call across multiple cameras. This change was mainly to bring back rendering so that at least we can still train perception environments on the current develop branch.

PR #4678, opened 2026-02-21
fix: correct inverted preserve_order docstring in resolve_matching_names

### LINKED ISSUE #4493: Docstring error in resolve_matching_names and resolve_matching_names_values: preserve_order description is inverted
Noting a mistake in the docstring for `isaaclab.utils.string.resolve_matching_names`: the text description incorrectly describes how the function works under `preserve_order=True` and `preserve_order=False`.

https://github.com/isaac-sim/IsaacLab/blob/37ddf626871758333d6ed89cf64ad702aef127d0/source/isaaclab/isaaclab/utils/string.py#L178

```
"""
...
If the :attr:`preserve_order` is True, the ordering of the matched indices and names is the same as the order
of the provided list of strings. This means that the ordering is dictated by the order of the target strings
and not the order of the query regular expressions.

If the :attr:`preserve_order` is False, the ordering of the matched indices and names is the same as the order
of the provided list of query regular expressions.

For example, consider the list of strings is ['a', 'b', 'c', 'd', 'e'] and the regular expressions are ['a|c', 'b'].
If :attr:`preserve_order` is False, then the function will return the indices of the matched strings and the
strings as: ([0, 1, 2], ['a', 'b', 'c']). When :attr:`preserve_order` is True, it will return them as:
([0, 2, 1], ['a', 'c', 'b']).
...
"""
```

In terms of the actual function operation, matched strings are enumerated in the order of the list of strings, and then only reordered to the query keys if `preserve_order=True`. This is the opposite of what the text describes. Swapping `True` and `False` in the above text will fix the documentation issue. The provided examples in the docstring are already correct according to the function's operation.

The same issue is seen in `isaaclab.utils.string.resolve_matching_names_values` directly below the other function.

https://github.com/isaac-sim/IsaacLab/blob/37ddf626871758333d6ed89cf64ad702aef127d0/source/isaaclab/isaaclab/utils/string.py#L274

```
"""
...
If the :attr:`preserve_order` is True, the ordering of the matched indices and names is the same as the order
of the provided list of strings. This means that the ordering is dictated by the order of the target strings
and not the order of the query regular expressions.

If the :attr:`preserve_order` is False, the ordering of the matched indices and names is the same as the order
of the provided list of query regular expressions.

For example, consider the dictionary is {"a|d|e": 1, "b|c": 2}, the list of strings is ['a', 'b', 'c', 'd', 'e'].
If :attr:`preserve_order` is False, then the function will return the indices of the matched strings, the
matched strings, and the values as: ([0, 1, 2, 3, 4], ['a', 'b', 'c', 'd', 'e'], [1, 2, 2, 1, 1]). When
:attr:`preserve_order` is True, it will return them as: ([0, 3, 4, 1, 2], ['a', 'd', 'e', 'b', 'c'], [1, 1, 1, 2, 2]).
...
"""
```

There is a typo/comma error in the example sentence:  
`... return the indices of the matched strings, the matched strings, and the values as: ...`  
(remove the redundant "the matched strings,")

This small doc fix will help avoid confusion for the use of other downstream functionalities, like in `isaaclab.assets.Articulation`:

https://github.com/isaac-sim/IsaacLab/blob/37ddf626871758333d6ed89cf64ad702aef127d0/source/isaaclab/isaaclab/assets/articulation/articulation.py#L288
```
"""
...
Find joints in the articulation based on the name keys.

Please see the :func:`isaaclab.utils.string.resolve_matching_names` function for more information
on the name matching.
...
"""
```

Thanks!

alltheseas:
## Summary

Fix the inverted `preserve_order` parameter description in the `resolve_matching_names` docstring.

The docstring incorrectly stated that `preserve_order=True` returns results in list_of_strings order, and `preserve_order=False` returns in query-key order. The actual behavior is the opposite:

- `preserve_order=False` (default): results follow `list_of_strings` order
- `preserve_order=True`: results are reordered to follow query regex key order

This is the same bug that PR #4427 fixes for the sibling function `resolve_matching_names_values`. This PR addresses the remaining function noted in #4493.

Fixes #4493

## Changes

- Swapped the `True`/`False` descriptions in the `resolve_matching_names` docstring
- Added `(default)` annotation to the `False` case for clarity
- Added brief inline clarifications to the examples (matching PR #4427 style)

Signed-off-by: alltheseas <alltheseas@users.noreply.github.com>

### DIFF
--- a/source/isaaclab/isaaclab/utils/string.py
+++ b/source/isaaclab/isaaclab/utils/string.py
@@ -183,17 +183,18 @@ def resolve_matching_names(
     When a list of query regular expressions is provided, the function checks each target string against each
     query regular expression and returns the indices of the matched strings and the matched strings.
 
-    If the :attr:`preserve_order` is True, the ordering of the matched indices and names is the same as the order
-    of the provided list of strings. This means that the ordering is dictated by the order of the target strings
-    and not the order of the query regular expressions.
+    If the :attr:`preserve_order` is False (default), the ordering of the matched indices and names follows
+    the order of the provided list of strings. This means that the ordering is dictated by the order of the
+    target strings and not the order of the query regular expressions.
 
-    If the :attr:`preserve_order` is False, the ordering of the matched indices and names is the same as the order
-    of the provided list of query regular expressions.
+    If the :attr:`preserve_order` is True, the ordering of the matched indices and names follows the order
+    of the query regular expressions.
 
     For example, consider the list of strings is ['a', 'b', 'c', 'd', 'e'] and the regular expressions are ['a|c', 'b'].
-    If :attr:`preserve_order` is False, then the function will return the indices of the matched strings and the
-    strings as: ([0, 1, 2], ['a', 'b', 'c']). When :attr:`preserve_order` is True, it will return them as:
-    ([0, 2, 1], ['a', 'c', 'b']).
+    If :attr:`preserve_order` is False (default), then the function will return the indices of the matched strings and
+    the strings as: ([0, 1, 2], ['a', 'b', 'c']) - following the order of list_of_strings. When
+    :attr:`preserve_order` is True, it will return them as:
+    ([0, 2, 1], ['a', 'c', 'b']) - following the order of the regex keys.
 
     Note:
         The function does not sort the indices. It returns the indices in the order they are found.


StafaH:

PR #4746, opened 2026-02-26
Updates Isaac Teleop button-state query to use retargeting pipeline output

rwiltz:
# Description

Query controller button states from the retargeting pipeline output (TensorGroup via ControllerInputIndex) instead of directly from ControllerTracker / DeviceIOSession. This eliminates fragile recursive tracker discovery and the risk of XR_ERROR_NAME_DUPLICATED from duplicate OpenXR action sets.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #4752, opened 2026-02-26
Fixes crashes when Stop XR is pressed and when Kit is closed while Teleop is running

hougantc-nvda:
# Description

Before Kit tears down the OpenXR instance and session, we need to cleanup IsaacTeleop as the cached handles will become invalidated on Kit cleanup.

Stop XR: In the /xr/enabled callback, call _teardown_dead_session() instead of stop(). Only the session is torn down; pipeline and retargeting UI stay. step() then returns None (same as deferred start), the recording loop keeps rendering, and starting XR again works without restarting the script.

Kit close: Subscribe to GLOBAL_EVENT_PRE_SHUTDOWN with order=-100 so our cleanup runs before XRCore (order=0). We release our session and stop using our cached OpenXR handles before Kit XR tears down the OpenXR instance/session, avoiding use of invalid handles.

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
+++ b/source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
@@ -81,6 +81,34 @@ class TeleopSessionLifecycle:
         except (ImportError, ModuleNotFoundError):
             logger.info("isaacsim.kit.xr.teleop.bridge not available; IsaacTeleop will create its own OpenXR session")
 
+        try:
+            import carb.settings
+
+            # Subscribe to the setting (may not fire when Kit closes; see pre-shutdown below)
+            self._xr_enabled_subscription = carb.settings.get_settings().subscribe_to_node_change_events(
+                "/xr/enabled",
+                self._on_xr_enabled_changed,
+            )
+        except (ImportError, ModuleNotFoundError):
+            logger.info("carb.settings not available; IsaacTeleop will not be able to detect XR enabled state")
+
+        try:
+            import omni.kit.app
+            from carb.eventdispatcher import get_eventdispatcher
+
+            # Subscribe to Kit pre-shutdown so we tear down our session before XRCore
+            # tears down the OpenXR instance/session (XRCore uses order=0; lowest runs first).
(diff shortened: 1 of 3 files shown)


hougantc-nvda:
Rebased on top of https://github.com/isaac-sim/IsaacLab/pull/4746
That PR will need to be merged prior to this PR.

@rwiltz for vis

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
```diff
@@ -78,6 +81,34 @@ def __init__(self, cfg: IsaacTeleopCfg):
...
             logger.info("isaacsim.kit.xr.teleop.bridge not available; IsaacTeleop will create its own OpenXR session")
 
+        try:
+            import carb.settings
+
+            # Subscribe to the setting (may not fire when Kit closes; see pre-shutdown below)
+            self._xr_enabled_subscription = carb.settings.get_settings().subscribe_to_node_change_events(
+                "/xr/enabled",
+                self._on_xr_enabled_changed,
+            )
```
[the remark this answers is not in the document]
hougantc-nvda: The subscription is RAII handle. It will automatically shutdown on class cleanup.

hougantc-nvda:
Rebased on develop and ready to merge.

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
```diff
@@ -190,6 +217,22 @@ def _on_request_required_extensions(self) -> list[str]:
         logger.info(f"Required extensions: {required_extensions}")
         return required_extensions
 
+    def _on_xr_enabled_changed(self, item, event_type):
+        import carb.settings
```
rwiltz, line 221: You have try catch for carb import above but not here. Should we be consistant?

rwiltz: I guess if the above fails this will never be hit so probably ok

rwiltz (approved):

PR #4771, opened 2026-02-27
Forces setuptools to use a version <82.0.0 to avoid pkg_resources issues

kellyguo11:
# Description

Some downstream packages (e.g. tensorboard) still appear to be affected by the change in recent setuptools 82.0.0 where pkg_resources was removed. To avoid these issues, we'll force a downgrade of setuptools to always install a version less than 82.0.0.

## Type of change

## Checklist


Mayankm96:
Could this be fixed on main as well? Building the docker image and running things inside the container is throwing a similar issue.

Issue #4789, opened 2026-03-02
[Bug Report] Non-spawning Rigid Bodies Fail


alexmillane:
### Describe the bug

**This bug is for Lab 3.0. (i.e. develop branch)**

Creating a `RigidBodyCfg` without a spawner (i.e. referencing a rigid body that already exists in another asset) fails, claiming that it is unable to find the path to prim, even though the prim exists.

### Steps to reproduce

I have modified the franka reach task to repro this behaviour in an MR: #4788 

After checking out that branch run:

```
./isaaclab.sh -p scripts/environments/zero_agent.py --num_envs 1 --visualizer kit --task Isaac-Reach-Franka-v0
```

And you will get:

```
Traceback (most recent call last):
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/scripts/environments/zero_agent.py", line 70, in <module>
    main()
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/scripts/environments/zero_agent.py", line 48, in main
    env = gym.make(args_cli.task, cfg=env_cfg)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/_isaac_sim/kit/python/lib/python3.12/site-packages/gymnasium/envs/registration.py", line 734, in make
    env = env_creator(**env_spec_kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/source/isaaclab/isaaclab/envs/manager_based_rl_env.py", line 80, in __init__
    super().__init__(cfg=cfg)
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/source/isaaclab/isaaclab/envs/manager_based_env.py", line 140, in __init__
    self.scene = InteractiveScene(self.cfg.scene)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/source/isaaclab/isaaclab/scene/interactive_scene.py", line 178, in __init__
    self._add_entities_from_cfg()
  File "/workspaces/isaaclab_arena/submodules/IsaacLab/source/isaaclab/isaaclab/scene/interactive_scene.py", line 705, in _add_entities_from_cfg
[...]

PR #4788, opened 2026-03-02
Repro failing rigid body reference.

alexmillane:
# Description

This is a repro MR associated with the issue #4789 

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #4809, opened 2026-03-04
Fix prim path not found for RigidObjCfg

xyao-nv:
# Description

During scene building, it spawns assets, creates wrappers and then call `clone_environments()` to copy template. So when RigidObject was created, the demanded prim path did not exist yet. It only exists after `clone_environments() `copying ops.

Fixes [issue](https://github.com/isaac-sim/IsaacLab/issues/4789)(https://github.com/isaac-sim/IsaacLab/issues/4789) with [repro mr](https://github.com/isaac-sim/IsaacLab/pull/4788).

The fix is for any `RigidObjCfg` has no spawn and whose prim path containing the env_namespace, defer its creation by `clone_environment()` first and create `RigidObject` and add it to `_rigid_objects`.

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/apps/isaaclab.python.xr.openxr.kit
+++ b/apps/isaaclab.python.xr.openxr.kit
@@ -39,7 +39,8 @@ xr.skipInputDeviceUSDWrites = true
 [dependencies]
 "isaaclab.python" = {}
 
-# Kit extensions
+# Required for XR instruction widget (CopyFabricPrim)
+"omni.fabric.commands" = {}
 "omni.kit.xr.system.openxr" = {}
 "omni.kit.xr.bundle.generic" = {}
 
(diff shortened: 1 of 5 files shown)


AntoineRichard:
@zoctipus can you take a look at that one?

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -769,6 +779,14 @@ def _add_entities_from_cfg(self):  # noqa: C901
                 asset_paths = sim_utils.find_matching_prim_paths(asset_cfg.prim_path)
                 self._global_prim_paths += asset_paths
 
+        # Clone template to envs, then create deferred RigidObjects (reference-only under env path).
+        self.clone_environments(copy_from_source=(not self.cfg.replicate_physics))
```
alexmillane, line 783: This clone needs to come before adding the non-spawning assets? I assume yes given you've moved it in here. We're that not needed, I feel we could solve our issues by moving the non-spawning assets behind the spawning assets in `ordered_items`

What's the reasoning there?

alexmillane (commented):
To me, this looks reasonable, although I'm not the most experienced.

One issue is that this will only work for non-spawning rigid bodies. Other types of non-spawning `Cfg`s will have the same issue.

I think we need some advice from the Lab team as to their preferred approach here and why this behaviour changed from Lab 2.3.

PR #4817, opened 2026-03-04
Add Experimental OvRTX Integration

bareya:
# Description

image

```
diff --git a/source/isaaclab_tasks/isaaclab_tasks/utils/presets.py b/source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
index 3ca5ea256d2..69532aa21f6 100644
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
@@ -4,6 +4,7 @@
 
 from isaaclab_newton.renderers import NewtonWarpRendererCfg
+from isaaclab_ovrtx.renderers import OVRTXRendererCfg
 from isaaclab_physx.renderers import IsaacRtxRendererCfg
 
@@ -15,4 +16,5 @@ from isaaclab_tasks.utils import PresetCfg
     default: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
     newton_renderer: NewtonWarpRendererCfg = NewtonWarpRendererCfg()
-    ovrtx_renderer: IsaacRtxRendererCfg = default
+    ovrtx_renderer: OVRTXRendererCfg = OVRTXRendererCfg()
+    ovrtx: OVRTXRendererCfg = OVRTXRendererCfg()  # alias for presets=newton,ovrtx
```

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
@@ -5,6 +5,9 @@
 
 """Configuration for OVRTX Renderer."""
 
+import tempfile
+from pathlib import Path
+
 from isaaclab.renderers.renderer_cfg import RendererCfg
 from isaaclab.utils import configclass
 
@@ -29,7 +32,7 @@ class OVRTXRendererCfg(RendererCfg):
     Set to False to use full RTX path-traced rendering with LdrColor.
     """
 
-    temp_usd_dir: str = "/tmp/ovrtx_test"
+    temp_usd_dir: str = str(Path(tempfile.gettempdir()) / "ovrtx")
     """Directory for temporary combined USD files (scene + injected cameras).
     Used by the OVRTX renderer when building the render scope; must be writable.
     """


source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
```diff
@@ -15,4 +16,4 @@
 class MultiBackendRendererCfg(PresetCfg):
     default: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
     newton_renderer: NewtonWarpRendererCfg = NewtonWarpRendererCfg()
-    ovrtx_renderer: IsaacRtxRendererCfg = default
+    ovrtx_renderer: OVRTXRendererCfg = OVRTXRendererCfg()
```
bareya, line 19: Hey @ooctipus @kellyguo11!

Can you have a look here and check if we are in agreement about the naming? I fixed importing configurations, right now OVRTX is lazily loadable. We can import config, it should not fail. However when we instantiate renderer through factory, then imports should fail.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -235,7 +235,7 @@ def __str__(self) -> str:
 
     @property
     def num_instances(self) -> int:
-        return self._view.count
+        return self._num_envs
```
bareya: Motivation for this change is to fix a runtime error when callers of this property could make an attempt to access the property before `_view` is created. @ooctipus let me know if this is something we should be more careful about.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_kernels.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Warp kernels and device constant for OVRTX renderer."""
+
+import numpy as np
+import warp as wp
```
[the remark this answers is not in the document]
bareya: This is going to be addressed in subsequent MRs.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -0,0 +1,548 @@
...
+import math
+import os
+import weakref
+from typing import TYPE_CHECKING
+
+import numpy as np
+import torch
+import warp as wp
+
+os.environ["OVRTX_SKIP_USD_CHECK"] = "1"
```
ooctipus, line 38: usually we try to avoid writing os environ in isaaclab. If there is need to check global variable, we ususally do it through
source/isaaclab/isaaclab/app/settings_manager.py


        from isaaclab.app.settings_manager import SettingsManager
        settings = SettingsManager.instance()
        settings.get("OVRTX_SKIP_USD_CHECK", 1)


if this is required by ovrtx package and can only be done through os environment, then it should be fine to leave as it

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -0,0 +1,548 @@
...
+        num_envs = sensor.num_instances
+        data_types = sensor.cfg.data_types if sensor.cfg.data_types else ["rgb"]
+
+        camera_prim_path = sensor.cfg.prim_path
+        camera_prim_name = (camera_prim_path or "").strip().split("/")[-1] or "Camera"
+        stage = sensor.stage
+        usd_scene_path = None
+        use_cloning = self.cfg.use_cloning
+        if stage is not None:
+            print(f"[OVRTX] Preparing stage for export ({num_envs} envs, cloning={use_cloning})...")
```
ooctipus: do this cloning work for heterogeneous environment? I guess we can find out when dexsuite lands

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -0,0 +1,548 @@
...
+                print(f"  ⚠ Warning: Failed to write omni:resetXformStack: {e}")
+
+            if self._camera_binding is not None:
+                print(f"  ✓ Camera binding created successfully")
+            else:
+                print(f"  ✗ WARNING: Camera binding is None!")
+
+            self._setup_object_bindings()
+
+    def _clone_environments_in_ovrtx(self, num_envs: int):
```
ooctipus, line 223: ha! looks like it will only work for the homogeneous environment for the moment, maybe you cna checkout 

source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py
source/isaaclab_physx/isaaclab_physx/cloner/physx_replicate.py

and change this cloning logic to something like 

source/isaaclab_ov/isaaclab_ov/cloner/ov_rtx_replicate.py

and see if cursor can adapt the logic to support heterogeneous case as well : D

source/isaaclab/isaaclab/sensors/camera/camera.py
kellyguo11: hmm what if we end up having multiple cameras in the view per environment? I think self._num_envs won't necessarily always be equivalent to self._view.count

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
kellyguo11: if it's needed, maybe just add a comment on why it's necessary :)

bareya: I agree and  I am not happy about this approach either. Unfortunately, this environment variable needs to be set prior to importing `ovrtx`. Therefore, it needs to be through `os` for the time being. This is _experimental_ integration.

ooctipus (approved):
I think we could address cloning logic later, os.environ is a small concern but I don't think its a big deal neither, otherwise looks good to me! Thank you @bareya

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
bareya: FYI this isn't the permanent solution. Eventually, when problem is addressed on ovrtx side, we wouldn't have to set this environment variable.

source/isaaclab/isaaclab/sensors/camera/camera.py
bareya: Oh ok! this is interesting insight! Let me have a look into this!

bareya: Reverted back!

source/isaaclab_ov/config/extension.toml
```diff
@@ -0,0 +1,11 @@
+[package]
```
kellyguo11, line 1: could we also add the docs folder with changelog just for consistency with how other directories are set up?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
bareya: Discussed through DM, but I am leaving here a trace:

1. Current approach is very naive. We take all cloned environments and serialize them to USD file. This is super slow homogeneous cloning. The slowness comes from serialization.
2. Second iteration would include improved homogeneous cloning - we only serialize first(template) environment as USD file. Then we use ovrtx feature to clone environments at runtime.
3. Third iteration would include serialization of the N template heterogeneous environments. Then we use ovrtx feature to clone them at runtime.

We can solve this problem incrementally. For now the intention is to put this MR in and we can solve many problems in parallel.

@kellyguo11 for vis.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
bareya: Replied here: https://github.com/isaac-sim/IsaacLab/pull/4817#discussion_r2886547382

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
bareya: I am resolving this conversation.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -0,0 +1,47 @@
...
+
+    simple_shading_mode: bool = True
+    """Whether to use simple shading mode (default: True).
+
+    When enabled, uses SimpleShadingSD RenderVar instead of LdrColor for RGB rendering.
+    Provides faster, simpler rendering suitable for many vision-based tasks.
+    Set to False to use full RTX path-traced rendering with LdrColor.
+    """
+
+    temp_usd_dir: str = "/tmp/ovrtx_test"
```
ncournia-nv: I wonder if we should use some built-in Python library that will create a temporary directory for us.  The idea here being it would work across platforms.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -0,0 +1,212 @@
...
+    camera_rel_list = ", ".join([f"<{p}>" for p in camera_paths])
+    return f'''
+def Scope "Render"
+{{
+    def RenderProduct "{render_product_name}" (
+        prepend apiSchemas = ["OmniRtxSettingsCommonAdvancedAPI_1"]
+    ) {{
+        rel camera = [{camera_rel_list}]
+        token omni:rtx:background:source:type = "domeLight"
+        token omni:rtx:rendermode = "RealTimePathTracing"
```
ncournia-nv: This should be "Minimal" when simple_shading_mode is true.

huidongc: Good point. We will split the concern with a refactoring change that replaces simple shading mode with minimal rendering mode very soon.

source/isaaclab/isaaclab/renderers/renderer.py
```diff
@@ -13,13 +13,17 @@
 from .renderer_cfg import RendererCfg
 
 # This is mapping of where backends live in the isaaclab_<backend> package.
-_RENDERER_TYPE_TO_BACKEND = {"isaac_rtx": "physx", "newton_warp": "newton"}
+_RENDERER_TYPE_TO_BACKEND = {"isaac_rtx": "physx", "newton_warp": "newton", "ovrtx": "ov"}
```
huidongc, line 16: I know it is probably due to legacy, but still it looks a bit weird to map "physx" to IsaacRTXRenderer.

source/isaaclab_ov/config/extension.toml
```diff
@@ -0,0 +1,11 @@
...
+version = "0.1.0"
+title = "Omniverse renderers for IsaacLab"
+description = "Extension providing Omniverse renderers (OVRTX, ovphysx, etc.) for tiled camera rendering."
```
huidongc, line 4: is `ovphysx` a renderer??

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -0,0 +1,50 @@
...
+    Used by the OVRTX renderer when building the render scope; must be writable.
+    """
+
+    temp_usd_suffix: str = ".usda"
+    """File suffix for temporary combined USD files (e.g. '.usda' or '.usdc')."""
+
+    use_cloning: bool = False
+    """When True, export only env_0 and use OVRTX clone_usd. When False, export full stage."""
+
+    log_level: str = "verbose"
```
huidongc, line 46: is verbose logging level intentional?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -0,0 +1,50 @@
...
+
+    The OVRTX renderer uses the ovrtx library for high-fidelity RTX-based rendering.
+    width, height, num_envs, and data_types are obtained from the sensor when
+    create_render_data() is called (same pattern as Isaac RTX).
+    """
+
+    renderer_type: str = "ovrtx"
+    """Type identifier for OVRTX renderer."""
+
+    simple_shading_mode: bool = True
```
huidongc, line 27: To be clear, this MR only exposes this config but the support is not implemented yet, I will follow to add the implementation. Correct?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -0,0 +1,50 @@
...
+    temp_usd_suffix: str = ".usda"
+    """File suffix for temporary combined USD files (e.g. '.usda' or '.usdc')."""
+
+    use_cloning: bool = False
+    """When True, export only env_0 and use OVRTX clone_usd. When False, export full stage."""
+
+    log_level: str = "verbose"
+    """OVRTX carb log level: "verbose", "info", "warn", "error"."""
+
+    log_file_path: str = "/tmp/ovrtx_renderer.log"
```
huidongc, line 49: This is hardcoded to Linux system path. I think you want something like this:

```python
log_file_path: str = str(Path(tempfile.gettempdir()) / "ovrtx_renderer.log")
```

source/isaaclab/isaaclab/renderers/renderer.py
bareya: Yes it's for backwards compatibility.

source/isaaclab_ov/config/extension.toml
bareya: No it's not. It's a future simulator, distributed as a library, similar to ovrtx.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
bareya: For now it has no effect, it didn't do anything and i left it as verbose. This integration is experimental, we can change it in very near future.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
bareya: Good point, i will make a MR fixing it. Thanks!

source/isaaclab_ov/config/extension.toml
huidongc: I see, it was a rhetoric question. The description makes it sound like ovphysx is one Omniverse renderer.

PR #4822, opened 2026-03-04
[Newton] Implements Frame Transformer Sensor

camevor:
# Description

- Rewrites the Newton `FrameTransformer` sensor to use a site-injection pipeline backed by `SensorFrameTransform`. Source and target sites are registered before replication via
`NewtonManager.cl_register_site`, so they automatically replicate per-world.
- Replaces the old per-view body lookup and inline Warp kernels with zero-copy `transform_to_vec_quat` views and a flat strided sensor output.
- Adds wildcard body matching for target frames (e.g. `"Robot/finger.*"` expands to one target per matched body per env).
- Also includes a one-line fix to `RigidObjectData` initializing `_default_root_state = None` for lazy allocation, and a minor contact sensor fix.

## Checklist

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -10,13 +10,6 @@ Added
   zero-copy splitting of ``wp.transformf`` arrays into ``vec3f`` and ``quatf`` views.
 
 
-Changed
-^^^^^^^
-
-* Changed :class:`~isaaclab.scene.InteractiveScene` to format ``sensor_shape_prim_expr``
-  and ``filter_shape_prim_expr`` config attributes with the environment regex namespace.
-
-
 4.5.29 (2026-04-10)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab_newton/docs/CHANGELOG.rst
+++ b/source/isaaclab_newton/docs/CHANGELOG.rst
@@ -1,7 +1,7 @@
 Changelog
 ---------
 
-0.5.11 (2026-04-07)
+0.5.11 (2026-04-13)
 ~~~~~~~~~~~~~~~~~~~
 
 Added
@@ -10,13 +10,6 @@ Added
 * Added :class:`~isaaclab_newton.sensors.frame_transformer.FrameTransformer` sensor
   wrapping Newton's ``SensorFrameTransform``. Supports per-env source/target site
   registration, wildcard body matching, and zero-copy transform views.
-
-Changed
-^^^^^^^
-
-* Changed :meth:`~isaaclab_newton.physics.NewtonManager.start_simulation` to inject
-  pending frame-transformer sites via :meth:`_cl_inject_sites_fallback` in the
-  non-replication path.
 
 
 0.5.10 (2026-04-05)


source/isaaclab_newton/isaaclab_newton/sensors/frame_transformer/frame_transformer.py
```diff
@@ -211,231 +115,223 @@ def body_names(self) -> list[str]:
...
-
-        # Allocate intermediate buffers (not exposed through data class)
-        self._warp_source_transforms_w = wp.zeros((self._num_envs,), dtype=wp.transformf, device=self._device)
-        self._warp_target_transform_source = wp.zeros(
-            (self._num_envs, self._num_frames), dtype=wp.transformf, device=self._device
+        num_envs = self._num_envs
+        site_map = NewtonManager._cl_site_index_map
+
+        # Resolve and validate per-env site indices
+        assert self._world_origin_label in site_map
```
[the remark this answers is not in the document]
camevor: > Since these check external state that can legitimately be wrong

Asserts used only for internal invariants. E.g. here, `_world_origin_label` being in `site_map` when `_initialize_impl` is called is an internal invariant and does not rely on user input.

camevor:
@isaaclab-review-bot 

1. Accurate but not actionable for this PR. As stated, this is a pre-existing pattern in the code base. No change needed.
2. Re-stating the convention for every composition of two transforms adds noise. No change needed.
3. This convention change was made across the board and documented and is not related to this PR. No change needed.
4. This is existing behavior. This PR simply uses it. No change needed.
5. Technically accurate but not a real risk. Both `_cl_inject_sites` and `_cl_inject_sites_fallback` clear `_cl_pending_sites`. Adding a debug log for an impossible condition is unnecessary noise. No change needed.
6. Tightened type hint to list[str], matching Newton `ModelBuilder.shape_labels`. Sites are shapes and thus, `site_idx` is always a valid shape index by construction. The existing assert on site_map membership is the right validation layer and raising with an error message would be no more helpful than a bare `IndexError`.

Nits:
1. `_target_frame_names` is set to a real list in `FrameTransformer.__init__`, before any external code can access `data.target_frame_names`. The None window is internal to the constructor. No change needed.
2. Added to the PR description.
3. The comment was removed because it described a workaround for sensor template path resolution that is no longer the only concern in that block. No change needed.

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -736,9 +736,6 @@ def _add_entities_from_cfg(self):  # noqa: C901
                     template_base = asset_cfg.prim_path.replace(self.env_regex_ns, self.cloner_cfg.template_root)
                     proto_id = self.cloner_cfg.template_prototype_identifier
                     if isinstance(asset_cfg, SensorBaseCfg):
-                        # Sensor may be nested under a proto_asset_N prim (e.g. a camera on a robot
-                        # link). Search for the actual template location so spawning succeeds even
-                        # though the parent asset lives at template_root/<Asset>/proto_asset_0/...
```
AntoineRichard, line 741: Is this no longer true?

source/isaaclab_newton/isaaclab_newton/assets/rigid_object/rigid_object_data.py
```diff
@@ -756,6 +756,7 @@ def _create_buffers(self) -> None:
         # -- default root pose and velocity
         self._default_root_pose = wp.zeros((self._num_instances,), dtype=wp.transformf, device=self.device)
         self._default_root_vel = wp.zeros((self._num_instances,), dtype=wp.spatial_vectorf, device=self.device)
+        self._default_root_state = None  # lazily allocated by deprecated default_root_state property
```
AntoineRichard, line 759: Was that missing var not caught by the tests?

source/isaaclab_newton/isaaclab_newton/sensors/frame_transformer/frame_transformer_kernels.py
```diff
@@ -0,0 +1,60 @@
...
+
+    Launch with dim=(num_envs, num_targets).
+    """
+    env, tgt = wp.tid()
+
+    if env_mask:
+        if not env_mask[env]:
+            return
+
+    target_transforms_w[env, tgt] = wp.transform_multiply(source_transforms[env], target_transforms[env, tgt])
```
AntoineRichard: Is that better than a plain source_transforms[env] * target_transforms[env, tgt]

source/isaaclab/isaaclab/utils/warp/math_ops.py
```diff
@@ -0,0 +1,34 @@
...
+    if t.dtype != wp.transformf:
+        raise TypeError(f"Expected wp.transformf array, got dtype={t.dtype}")
+    floats = t.view(wp.float32)
+    if t.ndim == 1:
+        return floats[:, :3].view(wp.vec3f), floats[:, 3:].view(wp.quatf)
+    if t.ndim == 2:
+        return floats[:, :, :3].view(wp.vec3f), floats[:, :, 3:].view(wp.quatf)
+    if t.ndim == 3:
+        return floats[:, :, :, :3].view(wp.vec3f), floats[:, :, :, 3:].view(wp.quatf)
+    raise ValueError(f"Expected 1D, 2D, or 3D transform array, got ndim={t.ndim}")
```
AntoineRichard, line 34: We can do slicing now? Might be interesting to keep this in mind when we'll review the asset code. There we use some quite violent code to generate this sort of things.

source/isaaclab_newton/isaaclab_newton/sensors/frame_transformer/frame_transformer.py
AntoineRichard, line 1: Do we have a quick benchmark on the initialization time of this sensor? (To avoid a contact sensor init situation)

camevor:
@greptile-apps

> Score of 4 reflects a solid implementation with thorough testing (both unit and integration), correct handling of both replication and fallback paths, and clean separation of concerns. Minor concern: assert statements used for runtime validation could be silently disabled with Python -O. No critical bugs found.

Assertions are used in this PR to verify internal invariants, failures of which are unrecoverable. User input and configuration validation does not use assertions but raises errors instead.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
AntoineRichard, line 1: Thanks for commenting the mods to the cloner here. @kellyguo11 we need to start considering another refactor of this class. I think after milestone 2 we need to look into this. That class is getting out of hand.

source/isaaclab_newton/isaaclab_newton/assets/rigid_object/rigid_object_data.py
camevor: Good question, I'm not sure why it wasn't caught.

source/isaaclab_newton/isaaclab_newton/sensors/frame_transformer/frame_transformer_kernels.py
camevor: Good point, this is exactly the same.

source/isaaclab/isaaclab/scene/interactive_scene.py
camevor: Good catch, thanks. Restored it.

source/isaaclab/isaaclab/utils/warp/math_ops.py
camevor: Agreed, this could simplify code in a few places.

source/isaaclab_newton/isaaclab_newton/sensors/frame_transformer/frame_transformer.py
camevor: Good point, I'll check.

camevor: A brief checks on 128 and 2048 worlds looks good:

**Setup:** NVIDIA L40, Python 3.11, Newton 1.1.0.dev0, kitless. Scene: ANYmal C + cube + FrameTransformer (4 feet targets). 200 steps, 20 warmup, 5 runs each.

### Initialization (ms, first run discarded)

| Phase | develop 128 | this PR 128 | develop 2048 | this PR 2048 |
|-------|:----------:|:----------:|:----------:|:----------:|
| `scene_construction` | 5237 ±783 | 3878 ±415 | 27285 ±29592 | 23722 ±8969 |
| `sim_reset` | 3805 ±143 | 3650 ±65 | 8480 ±521 | 7666 ±176 |
| **total** | **9042 ±896** | **7528 ±394** | **35764 ±29403** | **31388 ±9018** |

`scene_construction` has very large run-to-run variance. It dominates the init measurement and makes the two branches indistinguishable on init time. `sim_reset` is more stable and shows this PR ~1s faster at 2048 (9036 vs 8005), likely from the slimmer `_initialize_impl`.

**No init regression from the new implementation.**

 ### Step Breakdown (ms, median, first run discarded)

 | Phase | develop 128 | this PR 128 | develop 2048 | this PR 2048 |
 |-------|:----------:|:----------:|:----------:|:----------:|
 | `write_data_to_sim` | 1.042 | 1.042 | 1.237 | 1.229 |
 | `sim.step` | 0.781 | 0.792 | 1.188 | 1.169 |
 | **`scene.update`** | **0.180** | **0.166** | **1.861** | **0.160** |
 | **total** | **2.003** | **1.999** | **4.289** | **2.560** |

 Step-time std across 5 runs is ≤0.03 ms for all phases.
Step time scales much better with this implementation.

PR #4831, opened 2026-03-05
Improves distinguish-ability for objects in semantic segmentation images

huidongc:
# Description

Semantic segmentation in :class:`~isaaclab_ov.renderers.OVRTXRenderer` maps semantic instance IDs to RGBA using the same pseudo-random per-ID HSV scheme as the Isaac Sim RTX render backend, so OVRTX and Isaac RTX produce matching colors for the same IDs. Numeric IDs ``0`` (BACKGROUND) and ``1`` (UNLABELLED) use fixed RGBA.

## Type of change

## Screenshots

| Before | After |
| ------ | ----- |
| before | after |

## Checklist

### DIFF
--- a/source/isaaclab_tasks/config/extension.toml
+++ b/source/isaaclab_tasks/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "1.5.14"
+version = "1.5.15"
 
 # Description
 title = "Isaac Lab Environments"


source/isaaclab/setup.py
```diff
@@ -87,6 +87,7 @@
     "contrib": ["isaaclab_contrib"],
     "mimic": ["isaaclab_mimic"],
     "newton": ["isaaclab_newton"],
+    "ov": ["isaaclab_ov"],
```
bareya: I didn't add it on purpose, this is experimental package.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -396,6 +397,17 @@ def write_output(
         if src.ptr != output_data.data_ptr():
             wp.copy(dest=wp.from_torch(output_data), src=src)
 
+    def _compute_crc32_hash(self, input_data: wp.array) -> wp.array:
+        """Compute CRC32 hash per pixel for a 2D uint32 buffer; returns (h, w) uint32 array."""
+        output_hash = wp.zeros(shape=input_data.shape, dtype=wp.uint32, device=DEVICE)
```
bareya: Can we put this into render data? Do we need to reallocate this buffer every time?

source/isaaclab/setup.py
huidongc: I think I had to add this because the unit tests need to import from isaaclab_ov.

source/isaaclab_ov/test/test_ovrtx_renderer_kernels.py
```diff
@@ -0,0 +1,104 @@
...
+    compute_crc32_hash_kernel,
+)
+
+
+def _crc32_reference(value: int) -> int:
+    """Reference CRC32 for a single uint32 (4 bytes, little-endian)."""
+    return binascii.crc32(value.to_bytes(4, "little"))
+
+
+class TestComputeCrc32HashKernel:
```
bareya: This is awesome but it will fail in our ci/cd. We don't install ovrtx yet.

source/isaaclab/setup.py
bareya: They will fail :( we do not have cicd setup for ovrtx

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Good point!

source/isaaclab_ov/test/test_ovrtx_renderer_kernels.py
huidongc: Good point. I will skip this test case.

PR #4832, opened 2026-03-05
Add OVRTX as multi-backend configuration

bareya:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_presets_env_cfg.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_presets_env_cfg.py
@@ -6,10 +6,7 @@
 from __future__ import annotations
 
 from isaaclab_newton.physics import NewtonCfg
-from isaaclab_newton.renderers import NewtonWarpRendererCfg
-from isaaclab_ov.renderers import OVRTXRendererCfg
 from isaaclab_physx.physics import PhysxCfg
-from isaaclab_physx.renderers import IsaacRtxRendererCfg
 
 import isaaclab.sim as sim_utils
 from isaaclab.assets import ArticulationCfg
@@ -20,6 +17,7 @@ from isaaclab.sim import SimulationCfg
 from isaaclab.utils import configclass
 
 from isaaclab_tasks.utils import PresetCfg
+from isaaclab_tasks.utils.presets import MultiBackendRendererCfg
 
 from isaaclab_assets.robots.cartpole import CARTPOLE_CFG
 
@@ -29,14 +27,6 @@ class PhysicsCfg(PresetCfg):
     default = PhysxCfg()
     physx = PhysxCfg()
     newton = NewtonCfg()
-
-
-@configclass
-class MultiBackendRendererCfg(PresetCfg):
-    default: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
-    newton_renderer: NewtonWarpRendererCfg = NewtonWarpRendererCfg()
-    ovrtx_renderer: OVRTXRendererCfg = OVRTXRendererCfg()
-    isaacsim_rtx_renderer: IsaacRtxRendererCfg = default
 
 
 @configclass


fatimaanes:
The only thing that I see is that there are now two MultiBackendRendererCfg classes — one in presets.py (shared) and one in cartpole_camera_presets_env_cfg.py (task-local). They need to be kept in sync manually. A cleaner long-term approach would be to have the cartpole task import from the shared one. But that's a refactoring choice, not a correctness issue, and there may be reasons for keeping task configs self-contained. @kellyguo11 @ooctipus @bareya

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
ooctipus: Do you mind change here to directly use `needs_kit_for_config` ?

fatimaanes:
Looks good, thank you, Piotr! Glad to see the benchmark scripts now use launch_simulation for Kit detection instead of the hardcoded renderer checks, and new renderer backends won't need manual updates to the Kit detection logic.

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
bareya: Good idea and it's done!

source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_presets_env_cfg.py
```diff
@@ -34,6 +35,7 @@ class PhysicsCfg(PresetCfg):
 class MultiBackendRendererCfg(PresetCfg):
```
huidongc: This is a duplicate class of the other cfg class defined in isaaclab_tasks.utils.presets. Why not import MultiBackendRendererCfg from there?

bareya: To my understanding we want to have this preset defined per environment. Modifying global preset would affect all environments. We don't want that because we want to be selective which rendering/simulation backend is supported in which environment.

kellyguo11: I think the global cfg was meant to be a utility that can be shared by all tasks that use the same configuration. If the cartpole one is identical to the global preset, then we should be able to just import that one.

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
```diff
@@ -96,6 +96,30 @@ def _is_kit_camera(node) -> bool:
...
+        env_cfg: Resolved environment config (e.g. from :func:`resolve_task_config`).
+        launcher_args: Optional CLI args; if ``visualizer=kit`` is set, needs_kit is True.
+
+    Returns:
+        (needs_kit, has_kit_cameras, visualizer_types)
+    """
+    is_newton, has_kit_cameras = _scan_config(env_cfg, [_is_newton_physics, _is_kit_camera])
+    needs_kit = has_kit_cameras or not is_newton
+    visualizer_types = _get_visualizer_types(launcher_args)
+    if "kit" in visualizer_types:
```
huidongc, line 118: According to [this doc](https://github.com/isaac-sim/IsaacLab/blob/develop/docs/source/experimental-features/newton-physics-integration/visualization.rst), "omniverse" is also a valid input for the visualizer arg. Should we add the str in this test as well?

huidongc (approved):

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
kellyguo11: we'll update the doc. we are switching from omniverse to kit, so only kit will be a valid option

source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_presets_env_cfg.py
bareya: Thank you for pointing it out and clarification! Changes pushed with a fix!

bareya: @fatimaanes your concern from here 

https://github.com/isaac-sim/IsaacLab/pull/4832#issuecomment-4006770778

is addressed in the commit below!

PR #4843, opened 2026-03-06
Remove hardcoded camera prim paths from OVRTX renderer

ncournia-nv:
Remove hardcoded camera prim paths from OVRTX renderer

Derive the camera prim path from the sensor's resolved view prims
instead of assuming it is always "Camera". This fixes scenes where the
camera lives at a non-default path (e.g. Robot/ee_link/palm_link/Camera).

- Extract camera relative path from sensor._view at initialization
- Use it for camera bindings, scene partitions, object binding filter,
  and USD camera injection
- Discover cameras by USD type (UsdGeom.Camera) in create_cloning_attributes
  instead of matching on the "Camera" string
- Add ovrtx_renderer preset to cartpole camera presets env config

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -120,6 +120,7 @@ class OVRTXRenderer(BaseRenderer):
         self._initialized_scene = False
         self._sensor_ref: weakref.ref[object] | None = None
         self._exported_usd_path: str | None = None
+        self._camera_rel_path: str | None = None
 
     def prepare_stage(self, stage: Any, num_envs: int) -> None:
         """Export the USD stage for OVRTX before create_render_data.
@@ -133,7 +134,7 @@ class OVRTXRenderer(BaseRenderer):
         use_cloning = self.cfg.use_cloning
 
         logger.info("Preparing stage for export (%d envs, cloning=%s)...", num_envs, use_cloning)
-        create_cloning_attributes(stage, "Camera", num_envs, use_cloning)
+        create_cloning_attributes(stage, num_envs, use_cloning)
 
         export_path = "/tmp/stage_before_ovrtx.usda"
         export_stage_for_ovrtx(stage, export_path, num_envs, use_cloning)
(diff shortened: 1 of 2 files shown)


source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_presets_env_cfg.py
```diff
@@ -6,7 +6,10 @@
 from __future__ import annotations
 
 from isaaclab_newton.physics import NewtonCfg
+from isaaclab_newton.renderers import NewtonWarpRendererCfg
+from isaaclab_ov.renderers import OVRTXRendererCfg
```
bareya: You can undo those changes, This MR addresses this problem:

https://github.com/isaac-sim/IsaacLab/pull/4832

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -154,6 +155,12 @@ def initialize(self, sensor: SensorBase):
         num_envs = sensor.num_instances
         data_types = sensor.cfg.data_types if sensor.cfg.data_types else ["rgb"]
 
+        env_0_prefix = "/World/envs/env_0/"
```
bareya, line 158: Replacing one hard coded value with another lesser evil hard coded value?

ooctipus: hahahah
I think for homogeneous case maybe we can assume that for now.  When we unified the cloner across all engine and scripts, the cloner should be reponsible for informing all backends where to look for spawning path and cloning path.

ooctipus (approved):
its already a lot less hardcoded than before, : D Maybe you can let user pass in `/World/envs/env_.*` as default argument if user doesn't have any other prefix preference. that will somewhat address the hardcoding concern.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -154,6 +155,12 @@ def initialize(self, sensor: SensorBase):
...
+        first_cam_path = sensor._view.prims[0].GetPath().pathString
```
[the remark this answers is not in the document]
ncournia-nv: Good idea.  Something to consider during the camera refactor.

PR #4860, opened 2026-03-06
Adding docker image caching

nv-apoddubny:
# Description

Adding Docker image caching to CI


myurasov-nv:
It seems to fail at getting image from nvcr.io even though I can pull when ssh-d into the machine. Weird...

myurasov-nv:
Merged with #4877.

PR #4939, opened 2026-03-11
Atempts to fix disappearing develop docs

myurasov-nv:
Small change to docs.yaml CI task, that:
- builds only develop branch docs on develop
- builds other versions on main/release
- passes keep_files: true flag to the action that deploys gh pages


myurasov-nv:
@greptile-apps

kellyguo11 (approved):

PR #4955, opened 2026-03-11
Fix headless Kit crash with non-kit visualizers

ooctipus:
## Summary

- **`compute_kit_requirements`** now auto-enables `enable_cameras` when the rendering pipeline is required by non-kit visualizers (`newton`, `rerun`, `viser`) or Kit camera sensors. This ensures the rendering-capable headless experience file (`isaaclab.python.headless.rendering.kit`) is selected instead of the plain headless one, which crashes on Vulkan initialization.
- The `enable_cameras` logic is consolidated into `compute_kit_requirements` instead of being duplicated in `launch_simulation`.

## Test plan

- [ ] Run `Isaac-Repose-Cube-Shadow-Vision-Benchmark-Direct-v0` with `presets=newton_renderer,rgb --num_envs=16 --visualizer=newton` in headless mode and verify no Kit startup crash.
- [ ] Run with `--visualizer=kit` and verify existing behavior is unchanged.
- [ ] Run without `--visualizer` and verify existing behavior is unchanged.

### DIFF
--- a/source/isaaclab_tasks/config/extension.toml
+++ b/source/isaaclab_tasks/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "1.5.8"
+version = "1.5.9"
 
 # Description
 title = "Isaac Lab Environments"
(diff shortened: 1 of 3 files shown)


ooctipus:
Closing: the headless Kit crash was caused by a stale/corrupted prebundle repoint in the local Isaac Sim installation, not by the experience file selection logic. After a clean reinstall, isaaclab.python.headless.kit works fine for all cases including presets=newton_renderer.

PR #4966, opened 2026-03-12
Add simple quick installation docs

rdsa-nv:
# Description

- Simplify documentation for `Quick Installation`
- Add `./isaaclab.sh -uv` to both create a uv env, and source paths for an IsaacSim build from source.
- Fix silent fails when targeting a visualizer that was not installed

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5003, opened 2026-03-13
Raise error when explicitly requested visualizer fails

rdsa-nv:
## Summary
- When a visualizer is explicitly requested via CLI (`--visualizer`), raise a `RuntimeError` instead of silently logging the failure

Split from #4966 per reviewer feedback (simulation context change only).

## Test plan
- [ ] Run with `--visualizer <invalid>` and verify RuntimeError is raised
- [ ] Run with `--visualizer newton` and verify normal initialization still works
- [ ] Run without `--visualizer` and verify silent fallback behavior is unchanged


AntoineRichard:
@greptile

ooctipus:
@greptile

ooctipus (approved):

PR #5017, opened 2026-03-14
Adds Newton + Isaac RTX Rendering Performance Optimizations

ncournia-nv:
# Newton + Isaac RTX Rendering Performance Optimizations

This document describes four performance optimizations applied to the Newton physics simulator when used with the Isaac Sim RTX renderer inside Isaac Lab. Together they reduce per-frame time from **~323 ms to ~60 ms** (a **5.4x speedup**), making Newton's rendering path slightly faster than PhysX's equivalent (~65 ms).

All live primarily in two files:

- `source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py`
- `source/isaaclab_newton/isaaclab_newton/physics/_cubric.py` (new)

with small additions to `PhysicsManager` and `SimulationContext` in the core `isaaclab` package.

---

## Baseline: ~323 ms per frame

The starting point is the unoptimized Newton + RTX rendering loop. A Nsight Systems trace reveals the structure:

- **Two physics steps** execute per frame (typical for 2× physics substeps per render frame).
- **After each physics step**, Newton writes updated body transforms to Fabric (Omniverse's GPU scene-graph cache) and then triggers a full CPU hierarchy update via `update_world_xforms()`. This hierarchy walk recomputes every world-space transform in the scene from parent-child relationships — even though Newton already computed the correct world transforms and wrote them directly.
- The Kit renderer also runs its own, lighter, internal hierarchy update.

The per-step Fabric sync and hierarchy update dominates the frame. Because it runs after *every* physics step (not just before rendering), the cost is multiplied by the number of substeps.

newton-rtx-baseline

---

### Problem

Every physics substep was calling `sync_transforms_to_usd()`, which writes Newton body poses to Fabric and then invokes `update_world_xforms()`. The hierarchy update is expensive and only needs to happen once before the renderer reads the scene — not after every substep.

### Solution

A **dirty-flag pattern** decouples physics stepping from Fabric synchronization:

1. **`_mark_transforms_dirty()`** — called at the end of each `_simulate()` call, sets `_transforms_dirty = True`. This is cheap (a boolean assignment).
2. **`sync_transforms_to_usd()`** — now checks `_transforms_dirty` at the top and returns immediately if transforms haven't changed. When dirty, it writes transforms and calls the hierarchy update, then clears the flag.
3. **`pre_render()`** — a new method added to `PhysicsManager` (base class) and overridden by `NewtonManager`. It calls `sync_transforms_to_usd()`. The `SimulationContext.render()` method calls `physics_manager.pre_render()` before updating visualizers and cameras, ensuring transforms are flushed exactly once per render frame.

The key insight is that the renderer only reads scene transforms during `render()`, not during `step()`. By deferring the Fabric write and hierarchy update to render time, we eliminate redundant work when multiple physics substeps run per render frame. For 2 substeps per frame, this cuts the hierarchy update count in half.

### Key code paths

- `_simulate()` → `_mark_transforms_dirty()` (just sets a flag)
- `SimulationContext.render()` → `PhysicsManager.pre_render()` → `NewtonManager.sync_transforms_to_usd()` (runs once, clears the flag)

newton-rtx-dirty

---

### Problem

Looking at the physics steps in the trace, the GPU is underutilized. Each Warp kernel launch (collision detection, constraint solve, integration, FK evaluation) incurs a round-trip to the CPU via Python — launch overhead, GIL acquisition, and driver calls. For a simulation with many small kernels per substep, this CPU-side overhead becomes the bottleneck while the GPU sits idle between dispatches.

Newton already supported CUDA graphs (pre-recording a sequence of kernel launches and replaying them with a single driver call), but CUDA graph capture was **disabled when RTX rendering was active**. The original code had:

```python
use_cuda_graph = cfg.use_cuda_graph and (cls._usdrt_stage is None)
```

This was necessary because RTX's background threads use CUDA's legacy stream (stream 0) for async operations like `cudaImportExternalMemory`. Warp's standard `ScopedCapture()` uses `cudaStreamCaptureModeThreadLocal` on a blocking stream, which implicitly synchronizes with legacy stream 0. If RTX ops happen during capture, the CUDA runtime raises error 906 (`cudaErrorStreamCaptureImplicit`).

### Solution

A **deferred, relaxed-mode CUDA graph capture** strategy that is compatible with RTX:

**Deferral:** Graph capture is postponed from `initialize_solver()` to the first `step()` call. By that time, RTX has finished its initialization (all `cudaImportExternalMemory` calls are done) and is idle between render frames, providing a clean capture window.

```python
# In initialize_solver():
cls._graph = None
cls._graph_capture_pending = True

# In step():
if cls._graph_capture_pending:
    cls._graph = cls._capture_relaxed_graph(device)
```

**Relaxed-mode capture** (`_capture_relaxed_graph`): This method works around two conflicting requirements:

1. **RTX compatibility**: RTX threads use legacy stream 0. A blocking stream (Warp's default) implicitly syncs with it, causing capture failures. Solution: create a **non-blocking stream** (`cudaStreamNonBlocking = 0x01`) that has no implicit synchronization with stream 0.

2. **Warp compatibility**: `mujoco_warp` internally calls `wp.capture_while`, which checks Warp's `device.captures` registry to decide whether to insert a conditional graph node or synchronize eagerly. Without a registered capture, it calls `wp.synchronize_stream` on the capturing stream — which is illegal inside graph capture. Solution: call `wp.capture_begin(external=True, stream=fresh_stream)` to register the capture in Warp's tracking without calling `cudaStreamBeginCapture` again (already done externally).

The capture sequence:

1. **Warmup run** — execute `_simulate_physics_only()` eagerly to pre-allocate all MuJoCo-Warp scratch buffers (allocations are forbidden inside graph capture).
2. **Create a non-blocking CUDA stream** via `cudaStreamCreateWithFlags(..., NonBlocking)`.
3. **Begin capture** with `cudaStreamBeginCapture(..., cudaStreamCaptureModeRelaxed)` — relaxed mode allows other streams to operate freely during capture.
4. **Register with Warp** via `wp.capture_begin(external=True, stream=...)`.
5. **Record physics kernels** — `_simulate_physics_only()` inside `wp.ScopedStream(fresh_stream)`.
6. **Finalize** — `wp.capture_end()` then `cudaStreamEndCapture()` to obtain the graph.

**Physics-only capture:** `_simulate_physics_only()` was factored out of `_simulate()` to exclude Fabric sync operations (`wp.synchronize_device`, `wp.fabricarray`) that are incompatible with graph capture. After graph replay, `step()` marks transforms dirty, and `pre_render()` handles the Fabric sync eagerly.

The ctypes binding to `libcudart.so` is used directly because Warp's `ScopedCapture` doesn't expose control over capture mode or stream type.

newton-rtx-cuda-graph

---

### Problem

Even with the dirty-flag pattern reducing hierarchy updates to once per render frame, the `update_world_xforms()` call is still a **CPU-side tree walk** over the entire Fabric scene graph. For scenes with thousands of prims (typical in multi-environment RL), this CPU hierarchy propagation is a significant bottleneck.

The PhysX backend avoids this problem by using **cubric** — a GPU-accelerated transform hierarchy library. cubric runs the parent-child transform propagation entirely on the GPU via `IAdapter::compute()`, which is dramatically faster than the CPU walk. However, cubric has no Python bindings.

### Solution

**Pure-Python ctypes bindings to cubric's Carbonite interface** (`_cubric.py`), allowing Newton to use the same GPU hierarchy propagation that PhysX uses.

cubric is implemented as a Carbonite plugin and exposes its API through the `omni::cubric::IAdapter` interface. The bindings work by:

1. **Acquiring the Carbonite Framework** — `libcarb.so`'s `acquireFramework()` returns the singleton `Framework*`.
2. **Acquiring the IAdapter interface** — calling `tryAcquireInterfaceWithClient()` with the interface descriptor `omni::cubric::IAdapter` version `0.1`.
3. **Wrapping function pointers** — the `IAdapter` struct is a C++ vtable-like struct with function pointers at known offsets.

Each function pointer is read from the struct at its byte offset and wrapped with `ctypes.CFUNCTYPE` to make it callable from Python.

**Integration in `sync_transforms_to_usd()`:**

The sync method now mirrors PhysX's `ScopedUSDRT` pattern:

1. **Pause Fabric change tracking** — `track_world_xform_changes(False)` and `track_local_xform_changes(False)`. This is critical: `SelectPrims` with `ReadWrite` access internally calls `getAttributeArrayGpu`, which marks Fabric buffers dirty. If tracking is still active, the hierarchy records the change and Kit's `updateWorldXforms` will do an expensive connectivity rebuild every frame.
2. **Write transforms** — the existing Warp kernel writes Newton body poses to Fabric's `omni:fabric:worldMatrix`.
3. **Resume tracking** — re-enable change tracking (in a `finally` block for safety).
4. **Run cubric compute** — `IAdapter::compute()` with `eRigidBody | eForceUpdate` options and `eAll` dirty mode. The `eRigidBody` flag tells cubric to use **inverse propagation** on prims tagged with `PhysicsRigidBodyAPI` (preserve the world matrix that Newton wrote, derive the local transform) and **forward propagation** on everything else (propagate parent transforms to children). `eForceUpdate` bypasses cubric's change-listener dirty check since we know transforms have changed.

The adapter is lazily created on the first `sync_transforms_to_usd()` call rather than during `initialize_solver()`, to avoid startup-ordering issues with the cubric plugin.

When cubric is unavailable (e.g., plugin not loaded, CPU-only), the code falls back gracefully to the CPU `update_world_xforms()` path.

```
sync_transforms_to_usd():
    ┌─────────────────────────────────┐
    │  Pause Fabric change tracking   │
    ├─────────────────────────────────┤
    │  SelectPrims (ReadWrite)        │
    │  wp.launch(_set_fabric_transforms) │  ← GPU: write Newton poses to Fabric
    │  wp.synchronize_device()        │
    ├─────────────────────────────────┤
    │  cubric IAdapter::compute()     │  ← GPU: propagate hierarchy
    ├─────────────────────────────────┤
    │  Resume Fabric change tracking  │
    └─────────────────────────────────┘
```

A future Kit release is expected to ship official Python bindings for cubric, at which point the ctypes approach can be replaced.

The result is a frame time of **~60 ms** — slightly faster than PhysX's **~65 ms** on the same scene.

newton-rtx-cubric

---

## Summary

| Optimization | Frame Time | Speedup vs. Baseline | Key Technique |
|---|---|---|---|
| Baseline | ~323 ms | — | Sync + hierarchy after every substep |
| Dirty-flag deferred sync | ~244 ms | 1.3× | Sync once per render frame, not per substep |
| CUDA graph (relaxed mode) | ~144 ms | 2.2× | Eliminate per-kernel CPU launch overhead |
| cubric GPU hierarchy | ~60 ms | 5.4× | GPU hierarchy propagation via ctypes bindings |

All four optimizations are complementary and stack on top of each other. The final result matches or slightly beats the PhysX rendering path (~65 ms) while using Newton as the physics backend.

*Co-developed with Toby Jones (NVIDIA).*


kellyguo11:
@ncournia-nv could you run the formatter `./isaaclab.sh -f`?

also seeing some failures in test_rendering_correctness.py - https://github.com/isaac-sim/IsaacLab/actions/runs/23513986493/job/68449749241?pr=5017

scripts/benchmarks/benchmark_startup.py
```diff
@@ -0,0 +1,356 @@
...
+    Args:
+        env_cfg: Resolved environment configuration for the task.
+        app_launch_profile: cProfile session from the app-launch phase.
+        app_launch_wall_ms: Wall-clock duration of the app-launch phase [ms].
+    """
+
+    # Override config with CLI args
+    env_cfg.scene.num_envs = args_cli.num_envs if args_cli.num_envs is not None else env_cfg.scene.num_envs
+    env_cfg.sim.device = args_cli.device if args_cli.device is not None else env_cfg.sim.device
+    env_cfg.seed = args_cli.seed
```
kellyguo11: **Bug: unconditional seed override clobbers config default.**

```python
env_cfg.seed = args_cli.seed
```

When `--seed` is not passed, `args_cli.seed` is `None`. This unconditionally overwrites whatever `env_cfg.seed` was — either the task config's default or the conditional assignment at module-level line 166.

Line 166 already does the right thing:
```python
env_cfg.seed = args_cli.seed if args_cli.seed is not None else env_cfg.seed
```

This line should match that pattern, or just be removed since the module-level assignment already ran:
```python
# Either:
env_cfg.seed = args_cli.seed if args_cli.seed is not None else env_cfg.seed
# Or just delete this line — line 166 already handled it.
```

scripts/benchmarks/benchmark_startup.py
```diff
@@ -0,0 +1,356 @@
...
+                sys.exit(1)
+        _WHITELIST = raw
+
+# Resolve top_n default: 5 when using whitelist (fallback phases stay compact), 30 otherwise
+if args_cli.top_n is None:
+    args_cli.top_n = 5 if _WHITELIST else 30
+
+# -- Create the benchmark instance ------------------------------------------
+
+env_cfg.seed = args_cli.seed if args_cli.seed is not None else env_cfg.seed
```
kellyguo11: **Nit: module-level seed assignment is redundant with `main()`.** This line correctly does the conditional assignment, but then `main()` line 206 unconditionally overwrites it (see other comment). If you fix line 206 to be conditional too, consider removing one of the two assignments to avoid confusion about which one "wins".

Since `env_cfg` is passed into `main()` and `main()` also sets `num_envs` and `device` there, it'd be cleaner to do **all** config overrides in one place — inside `main()` — and remove this line.

scripts/benchmarks/benchmark_startup.py
```diff
@@ -0,0 +1,356 @@
...
+                    phase_name, measurement=SingleMeasurement(name=label, value=round(tottime_ms, 2), unit="ms")
+                )
+                benchmark.add_measurement(
+                    phase_name,
+                    measurement=SingleMeasurement(name=f"{label} (cumtime)", value=round(cumtime_ms, 2), unit="ms"),
+                )
+
+        # Finalize benchmark output
+        benchmark.update_manual_recorders()
+        benchmark._finalize_impl()
```
kellyguo11: **Minor: calling private `_finalize_impl()` directly.**

I know other benchmark scripts (`benchmark_non_rl.py`) do this too, so this is a pre-existing pattern rather than a problem you introduced. Still worth noting: if `BaseIsaacLabBenchmark` ever gains a public `finalize()` wrapper (which it should), this should switch to it.

scripts/benchmarks/benchmark_startup.py
```diff
@@ -0,0 +1,356 @@
...
+import gymnasium as gym  # noqa: E402
+import torch  # noqa: E402
+
+from isaaclab.envs import DirectMARLEnvCfg, DirectRLEnvCfg, ManagerBasedRLEnvCfg  # noqa: E402
+
+from isaaclab_tasks.utils import launch_simulation, resolve_task_config  # noqa: E402
+
+imports_profile.disable()
+
+if torch.cuda.is_available() and torch.cuda.is_initialized():
```
kellyguo11: **Design note: `cuda.is_initialized()` check may miss lazy CUDA init.**

At this point in the script, CUDA may not yet be initialized (torch was just imported but no tensor ops have run). If CUDA isn't initialized, `is_initialized()` returns `False` and the synchronize is skipped — which is correct, since there's nothing to sync. But it means the wall-clock `imports_time_end` doesn't account for any deferred GPU init that gets triggered later.

This is fine for a profiling script (you're measuring import time, not GPU init time), but worth a comment explaining the intent — especially since the same pattern at lines 225 and 245 *does* catch active GPU work.

kellyguo11 (commented):
## Detailed Code Review — Newton + RTX Rendering Performance Optimizations

Excellent PR. The 5.4× speedup from three stacked optimizations (dirty-flag deferred sync, relaxed-mode CUDA graph capture, cubric GPU hierarchy) is well-motivated by Nsight traces and the implementation shows deep understanding of the CUDA capture + RTX interaction constraints. The PR description is production-quality documentation.

**NOTE:** This review's inline comments were posted against the wrong commit due to a tooling error. Please see the subsequent review for the actual inline code comments.

**CI Status:** pre-commit fails (ruff try-except-pass lint). Other failures (isaaclab core, newton, physx) appear to be pre-existing flaky tests unrelated to this PR.

**Overall verdict:** Approve with minor fixes. The architecture is sound.

scripts/benchmarks/utils.py
```diff
@@ -235,3 +236,115 @@ def log_convergence(
...
+    # The public get_stats_profile() (Python 3.9+) doesn't expose caller info, which
+    # we need for the first-level external call filter. If a future Python release
+    # breaks this, switch to get_stats_profile() and drop the caller-based filtering.
+    # stats.stats: dict[(filename, lineno, funcname)] -> (pcalls, ncalls, tottime, cumtime, callers)
+    # callers: dict[(filename, lineno, funcname)] -> (pcalls, ncalls, tottime, cumtime)
+    results = []
+    for func_key, (_, _, tottime, cumtime, callers) in stats.stats.items():
+        filename, _, funcname = func_key
+        if _is_isaaclab(filename):
+            label = _make_label(filename, funcname)
```
kellyguo11: **Observation (not a bug): external functions can appear with abbreviated labels.** When an external function (e.g., `torch.nn.modules.linear:forward`) is directly called by an IsaacLab function, `_make_label` truncates the path to `site-packages`-relative or last-3-components. This is fine for readability but means the label isn't a valid Python import path — it's a display-only string. Worth a brief docstring note on `_make_label` that labels are for display, not programmatic use.

scripts/benchmarks/benchmark_startup.py
```diff
@@ -0,0 +1,356 @@
...
+            env_creation_profile.disable()
+
+        if torch.cuda.is_available() and torch.cuda.is_initialized():
+            torch.cuda.synchronize()
+        env_creation_time_end = time.perf_counter_ns()
+        # -- First step profiled ------------------------------------------------
+
+        # Sample random actions
+        actions = (
+            torch.rand(env.unwrapped.num_envs, env.unwrapped.single_action_space.shape[0], device=env.unwrapped.device)
```
kellyguo11: **Nit: `single_action_space` may not exist for all env types.**

For `DirectMARLEnv`, the action space structure is different — `single_action_space` exists but might be a `Dict` space, not a `Box`, so `.shape[0]` would fail. The type signature of `main()` accepts `DirectMARLEnvCfg`, but this line assumes a flat action space.

Since this is a benchmarking script (not a general-purpose tool), this is acceptable if the docs/help text clarify it's for single-agent RL envs. But if MARL envs are intended to be supported, this needs a check.

source/isaaclab/isaaclab/test/benchmark/backends.py
```diff
@@ -276,6 +276,33 @@ def _print_summary(self) -> None:
...
+        for phase_name, phase in phases.items():
+            if phase_name in known_phases or not phase.measurements:
+                continue
+            self._print_box_line(f"Phase: {phase_name}")
+            for measurement in phase.measurements:
+                label = measurement.name
+                if isinstance(measurement, StatisticalMeasurement):
+                    unit_str = f" {measurement.unit.strip()}" if (measurement.unit and measurement.unit.strip()) else ""
+                    value = f"{self._format_scalar(measurement.mean)}{unit_str}"
+                elif isinstance(measurement, SingleMeasurement):
```
kellyguo11: **Good addition.** The catch-all rendering for unknown phases is clean and doesn't break existing behavior. One minor style nit: the `unit_str` construction is duplicated between the `frametime_phase` block above and this new block. Could extract a helper like `_format_measurement_value(measurement)` to DRY it up, but that's cosmetic.

kellyguo11:
## Detailed Code Review — Newton + RTX Rendering Performance Optimizations

Excellent PR. The 5.4× speedup from three stacked optimizations (dirty-flag deferred sync, relaxed-mode CUDA graph capture, cubric GPU hierarchy) is well-motivated by Nsight traces and the implementation shows deep understanding of the CUDA capture + RTX interaction constraints. The PR description is production-quality documentation.

Below are the issues I found, ordered roughly by severity. Inline comments are posted as review comments on the relevant lines.

**CI Status:** pre-commit fails (ruff `try-except-pass` lint on line 718-720 of newton_manager.py). Other failures (isaaclab core, newton, physx) appear to be pre-existing flaky tests unrelated to this PR.

**Overall verdict:** Approve with minor fixes. The architecture is sound; the issues are correctness edge cases and cleanup items, not design flaws.

---

### 1. CUDA Graph Handle Management (`_capture_relaxed_graph`, line ~737)

After `wp.capture_end()` returns a `wp.Graph`, you overwrite its `.graph` member with a `ctypes.c_void_p` and set `.graph_exec = None`. Potential issues:

- **Type mismatch:** `raw_graph` is a `ctypes.c_void_p` struct, not its integer `.value`. When `wp.capture_launch` passes this to `cudaGraphInstantiate`, the wrong type may cause issues. Use `raw_graph.value` if you need the integer handle.
- **Double capture end:** If `wp.capture_end()` in external mode still calls `cudaStreamEndCapture` internally, your subsequent manual `cudaStreamEndCapture` may return NULL or double-free the graph.
- **`graph_exec = None`:** Forces `wp.capture_launch` to re-instantiate on every launch, which may add overhead. Consider instantiating once.

**Suggestion:** Verify exactly what `wp.capture_end(stream=fresh_stream)` returns in external mode. If it already contains the correct graph, the manual `cudaStreamEndCapture` may be redundant.

### 2. `bind_to_stage` Called Every Frame (line 249)

The Fabric stage ID doesn't change between frames. Calling `bindToStage` every sync is at minimum unnecessary overhead and potentially triggers internal state reconstruction in cubric. PhysX's `DirectGpuHelper` binds once.

Consider caching:
```python
if cls._cubric_bound_fabric_id != fabric_id:
    cls._cubric.bind_to_stage(cls._cubric_adapter, fabric_id)
    cls._cubric_bound_fabric_id = fabric_id
```

### 3. Warmup Run During First `step()` Advances Physics State (line ~673)

The warmup call in `_capture_relaxed_graph()` runs `_simulate_physics_only()` eagerly, advancing `_state_0`/`_state_1` **at the first step() call** after reset. This means the first frame has an extra physics advancement compared to the non-RTX path (where warmup happens during `initialize_solver()`).

**Suggestion:** Save and restore state around the warmup, or document the first-frame side effect.

### 4. `_transforms_dirty` Not Cleared on Empty Selection (line 233)

When `GetCount() == 0`, the method returns without clearing `_transforms_dirty`. Every subsequent frame will re-enter the full sync path (pause/resume tracking, SelectPrims) for nothing. Add `cls._transforms_dirty = False` before the early return.

### 5. Hard-Coded Struct Offsets in `_cubric.py` (lines 37-56)

The vtable offsets are derived from specific Kit/cubric header versions. If a future Kit version reorders fields, these will silently read wrong function pointers, causing segfaults (not Python exceptions). Consider:
- Document the exact Kit/cubric version these offsets were verified against
- Add a smoke-test (create + release) during `initialize()` to validate function pointers
- Track the migration plan to official Python bindings

### 6. `tryAcquireInterfaceWithClient` ABI (`_cubric.py` line ~159)

The ctypes signature passes `_InterfaceDesc` by value. This is correct for the Carbonite C ABI on x86_64 (struct ≤16 bytes → registers), but verify the C++ side doesn't expect `const InterfaceDesc&` (by-reference), which would be an ABI mismatch.

### 7. Pre-commit Lint Fix Needed (line ~718)

`try-except-pass` → use `contextlib.suppress(Exception)` or add `# noqa: SIM105`.

### 8. `release_adapter` Shutdown Safety (line 340)

If cubric plugin is already torn down during Kit shutdown, the function pointer call in `release_adapter` could segfault. Wrap in try/except.

### 9. Double `sync_transforms_to_usd()` Call Path

`pre_render()` → `sync_transforms_to_usd()` runs first, then `update_visualizers()` → `NewtonSceneDataProvider.update()` → `sync_transforms_to_usd()` runs second. The dirty flag makes the second call a cheap no-op, but consider removing the scene data provider call since `pre_render()` now handles it.

kellyguo11 (approved):

source/isaaclab_newton/isaaclab_newton/physics/_cubric.py
```diff
@@ -0,0 +1,268 @@
...
+    def __init__(self) -> None:
+        self._ia_ptr: int = 0
+        self._create_fn = None
+        self._release_fn = None
+        self._bind_fn = None
+        self._compute_fn = None
+
+    # -- lifecycle -----------------------------------------------------------
+
+    def initialize(self) -> bool:
```
kellyguo11, line 116: could this eventually be an interface that can be exposed from kit? the code looks quite scary haha

ncournia-nv: Yes, we'll work on making equivalent bindings part of a future Kit release.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -139,49 +157,114 @@ def forward(cls) -> None:
...
+            try:
+                selection = cls._usdrt_stage.SelectPrims(
+                    require_attrs=[
+                        (usdrt.Sdf.ValueTypeNames.Matrix4d, "omni:fabric:worldMatrix", usdrt.Usd.Access.ReadWrite),
+                        (usdrt.Sdf.ValueTypeNames.UInt, cls._newton_index_attr, usdrt.Usd.Access.Read),
+                    ],
+                    device=str(PhysicsManager._device),
+                )
+                if selection.GetCount() == 0:
+                    return
```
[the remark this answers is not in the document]
ncournia-nv: Fixed.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -139,49 +157,114 @@ def forward(cls) -> None:
...
+                    device=PhysicsManager._device,
+                )
+                wp.synchronize_device(PhysicsManager._device)
+
+                cls._transforms_dirty = False
+
+                if use_cubric and fabric_hierarchy is not None:
+                    fabric_id = cls._usdrt_stage.GetFabricId().id
+                    cls._cubric.bind_to_stage(cls._cubric_adapter, fabric_id)
+                    cls._cubric.compute(cls._cubric_adapter)
```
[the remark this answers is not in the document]
ncournia-nv: Fixed.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -477,25 +585,169 @@ def initialize_solver(cls) -> None:
...
+            return None
+
+        if end_ret != 0 or not raw_graph.value:
+            logger.warning("cudaStreamEndCapture failed (code %d)", end_ret)
+            return None
 
-        # MJWarp can use its internal collision pipeline.
+        graph.graph = raw_graph
+        graph.graph_exec = None
+        return graph
```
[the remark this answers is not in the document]
ncournia-nv: Fixed.

ncournia-nv:
## CI Failure Analysis

All 6 failing jobs in the [latest CI run](https://github.com/isaac-sim/IsaacLab/actions/runs/23774029576) are **pre-existing failures unrelated to this PR**. The same tests fail across unrelated branches with identical errors.

### Failure Breakdown

| Job | Failing Test(s) | Pre-existing? | Evidence |
|-----|-----------------|---------------|----------|
| **isaaclab (core) [1/3]** | `test_first_frame_textured_rendering` (delta 18.7 > 15.0 threshold), `test_tiled_camera` semantic segmentation (negative mean assertion) | ✅ Yes | Fails on [`test_ci_issues`](https://github.com/isaac-sim/IsaacLab/actions/runs/23795170083), [`my-ci-fix-3`](https://github.com/isaac-sim/IsaacLab/actions/runs/23781161969), [`reb/multitask_demo_migrate`](https://github.com/isaac-sim/IsaacLab/actions/runs/23786015221) |
| **isaaclab (core) [2/3]** | `test_ideal_pd_actuator` (test report missing — timeout/crash) | ✅ Yes | Fails on `my-ci-fix-3`, `reb/multitask_demo_migrate` |
| **isaaclab (core) [3/3]** | `test_manager_based_rl_env_obs_spaces` (clip_constraint), `test_device_constructors`, `test_spawn_from_files`, `test_utils_transforms` (all timeout/crash — `-1/0` results) | ✅ Yes | Fails on `test_ci_issues`, `my-ci-fix-3`, `reb/multitask_demo_migrate` |
| **isaaclab_tasks [3/3]** | `test_rendering_correctness` (42 golden image mismatches at 5-32% diff — CI GPU vs golden image GPU mismatch) | ✅ Yes | Fails on all checked branches |
| **isaaclab_contrib** | `test_thruster`, `test_multirotor` (timeout — no test report generated) | ✅ Yes | `test_multirotor` crashes on [`my-ci-fix-3`](https://github.com/isaac-sim/IsaacLab/actions/runs/23781161969), fails on `reb/multitask_demo_migrate`. Flaky — passed on our previous run |
| **environments_training** | `test_train_environments[skrl-task_spec19]` (1 of 568 tests) | ✅ Yes | Fails on all checked branches |

### Cross-Branch Verification

Checked recent CI runs on **unrelated branches** to confirm these are not introduced by our changes:

- [`test_ci_issues`](https://github.com/isaac-sim/IsaacLab/actions/runs/23795170083) (Mar 31) — same core [1/3], core [3/3], tasks [3/3], environments_training failures
- [`my-ci-fix-3`](https://github.com/isaac-sim/IsaacLab/actions/runs/23781161969) (Mar 31) — same core [1/3], [2/3], [3/3], tasks [3/3], contrib, environments_training failures
- [`reb/multitask_demo_migrate`](https://github.com/isaac-sim/IsaacLab/actions/runs/23786015221) (Mar 31) — even more failures (10 jobs failing)

### What Passes

All jobs directly related to our changes pass cleanly: **pre-commit**, **isaaclab_newton**, **isaaclab_rl**, **isaaclab_physx**, **docs build**, **license-check**.

PR #5030, opened 2026-03-16
Escapes cmd.exe Special Characters

myurasov-nv:
When `subprocess.run()` receives a list whose first element is a `.bat` file, Python invokes `cmd.exe /c "..."` under the hood.  `subprocess.list2cmdline` executed down the line does not escape `cmd.exe` special characters (`<`, `>`, `|`, `&`, `^`), so they are interpreted as shell operators, which breaks commands that utilize IsacSim's `python.bat` as interpreter (`"C:\Users\Misha\IsaacLab\_isaac_sim\python.bat -m pip install setuptools<82.0.0"` etc).

`run_command()` now detects `.bat`/`.cmd` executables on Windows and converts the command list to a pre-quoted string, bypassing `list2cmdline`.

## Type of change

## Checklist

- [NA] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file


AntoineRichard (approved):
LGTM

PR #5051, opened 2026-03-18
Fix Newton model not built for sensors added in _setup_scene()

nblauch:
InteractiveScene.__init__() resolves renderer requirements from sensors declared in the scene config, but DirectRLEnv environments add sensors programmatically in _setup_scene() after the scene is constructed. This means PhysxSceneDataProvider is never told it needs a Newton model, so get_newton_model() returns None and Newton-based renderers/visualizers get a black screen.

Add _update_scene_data_requirements_from_sensors() to DirectRLEnv which re-scans all scene sensors after _setup_scene() and merges their renderer requirements into the simulation context before the provider is constructed.

# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/direct_rl_env.py
+++ b/source/isaaclab/isaaclab/envs/direct_rl_env.py
@@ -149,6 +149,12 @@ class DirectRLEnv(gym.Env):
             with use_stage(self.sim.stage):
                 self.scene = InteractiveScene(self.cfg.scene)
                 self._setup_scene()
+        # InteractiveScene.__init__() resolves renderer requirements from sensors
+        # declared in the scene config, but sensors added programmatically in
+        # _setup_scene() (the standard DirectRLEnv pattern) are missed. Re-scan
+        # all scene sensors and merge their requirements so that the scene data
+        # provider (constructed later) knows it needs a Newton model.
+        self._update_scene_data_requirements_from_sensors()
         print("[INFO]: Scene manager: ", self.scene)
 
         # set up camera viewport controller
@@ -656,6 +662,31 @@ class DirectRLEnv(gym.Env):
         any explicit scene setup, the function can be left empty.
         """
         pass
+
+    def _update_scene_data_requirements_from_sensors(self):
+        """Re-scan scene sensors and update simulation context requirements.
+
+        ``InteractiveScene.__init__()`` resolves renderer requirements from
+        sensors declared in the scene *config*, but sensors added
+        programmatically in :meth:`_setup_scene` (the standard ``DirectRLEnv``
+        pattern) are missed.  This method re-scans all scene sensors after
+        ``_setup_scene()`` returns and merges their renderer requirements into
+        the simulation context so that ``PhysxSceneDataProvider`` (constructed
+        later) knows whether it needs to build a Newton model.
+        """
+        from isaaclab.physics.scene_data_requirements import (
+            aggregate_requirements,
+            resolve_scene_data_requirements,
+        )
+
+        renderer_types = self.scene._sensor_renderer_types()
+        if not renderer_types:
+            return
+        sensor_req = resolve_scene_data_requirements(visualizer_types=[], renderer_types=renderer_types)
+        current_req = self.sim.get_scene_data_requirements()
+        merged = aggregate_requirements([current_req, sensor_req])
+        if merged != current_req:
+            self.sim.update_scene_data_requirements(merged)
 
     @abstractmethod
     def _pre_physics_step(self, actions: torch.Tensor):


source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -149,6 +149,12 @@ def _init_sim(self, render_mode: str | None = None, **kwargs):
             with use_stage(self.sim.stage):
                 self.scene = InteractiveScene(self.cfg.scene)
                 self._setup_scene()
+        # InteractiveScene.__init__() resolves renderer requirements from sensors
+        # declared in the scene config, but sensors added programmatically in
+        # _setup_scene() (the standard DirectRLEnv pattern) are missed. Re-scan
```
kellyguo11, line 154: 🟡 **`DirectMARLEnv` has the identical bug pattern but is not covered by this fix.**

`DirectMARLEnv` at `direct_marl_env.py:146` has the same `_setup_scene()` → `InteractiveScene` flow but does NOT inherit from `DirectRLEnv`. Sensors added programmatically in `DirectMARLEnv._setup_scene()` will still have missing renderer requirements.

Consider either:
1. Extracting `_update_scene_data_requirements_from_sensors()` into a shared utility function and calling it from both `DirectRLEnv._init_sim()` and `DirectMARLEnv._init_sim()`.
2. Adding the same method and call to `DirectMARLEnv`.

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -149,6 +149,12 @@ def _init_sim(self, render_mode: str | None = None, **kwargs):
...
+        # all scene sensors and merge their requirements so that the scene data
+        # provider (constructed later) knows it needs a Newton model.
+        self._update_scene_data_requirements_from_sensors()
         print("[INFO]: Scene manager: ", self.scene)
```
kellyguo11, line 159: 🔵 **Nit: The `_update_scene_data_requirements_from_sensors()` call is outside the `Timer` block.**

This is intentional and correct — the method doesn't need the `use_stage()` context and is a lightweight operation. Just confirming for future readers that this placement is deliberate (after scene creation timing, before provider construction during `sim.reset()`).

kellyguo11 (commented):
## 🤖 Isaac Lab Review Bot

### Summary

This PR fixes a real bug where sensors added programmatically in `DirectRLEnv._setup_scene()` don't have their renderer requirements (e.g., `requires_newton_model`) reflected in the simulation context's `_scene_data_requirements`. This is particularly important for the `OVRTXRenderer`, which does NOT do its own requirement merging (unlike `NewtonWarpRenderer` which handles it in its `__init__`). The fix correctly re-scans all scene sensors after `_setup_scene()` returns and merges their requirements before the scene data provider is constructed.

### Architecture Impact

- **`DirectRLEnv` only** — the identical `DirectMARLEnv` is not patched (see finding #1).
- No API surface changes. The new method is private (`_update_scene_data_requirements_from_sensors`) and called internally.
- Minimal overhead for environments that don't add sensors in `_setup_scene()` (early return on empty renderer_types).

### Implementation Verdict

**Minor fixes needed.** The core fix is correct and well-placed, but `DirectMARLEnv` has the exact same bug pattern and should be fixed in the same PR.

### Test Coverage

No tests added. The PR checklist acknowledges this. A test that creates a `DirectRLEnv` subclass adding a newton_warp sensor in `_setup_scene()` and verifies `sim.get_scene_data_requirements().requires_newton_model` would strengthen confidence.

### CI Status

Only the `labeler` check ran (✅ pass). No functional CI checks observed.

### Findings

1. 🟡 **`DirectMARLEnv` has the same bug but is not fixed.** `DirectMARLEnv(gym.Env)` does NOT inherit from `DirectRLEnv` — it's a separate class with its own `_setup_scene()` pattern at `direct_marl_env.py:146`. It needs the same `_update_scene_data_requirements_from_sensors()` call. Either extract the method to a shared mixin/utility, or duplicate it in `DirectMARLEnv`.

2. 🔵 **Missing CHANGELOG.rst entry.** This is a user-visible bug fix (black screen for Newton-based renderers/visualizers in DirectRLEnv). It should have a changelog entry under the next version's `Fixed` section.

3. 🔵 **Observation: `NewtonWarpRenderer` already self-heals, but `OVRTXRenderer` does not.** I traced both renderer `__init__` paths. `NewtonWarpRenderer.__init__()` independently merges its own requirement before creating the provider, so it would work even without this fix. However, `OVRTXRenderer.__init__()` does NOT merge requirements — it just calls `initialize_scene_data_provider()` later in `_setup_object_bindings()` and checks `get_newton_model()`. For `OVRTXRenderer`, this fix is essential. This context may be worth noting in the commit message or PR description.

ooctipus:
seems addressed, please reopen if the bug persists

PR #5059, opened 2026-03-18
Splits Tests More Evenly, Only Runs Relevant Ones, Adds Reporting

myurasov-nv:
# Description

Until now every PR ran **all** GPU test jobs regardless of what changed. That is no-good and needs to change. :)

This PR:
- Splits tests more evenly and according to functionality
- Only runs relevant ones; manually all can still be executed
- Adds reporting of the passed/failed tests on the Workflow page
- Improves job cancellation so resources are freed faster

## Detailed changes

- Added a `detect-changes` job that runs `git diff --name-only` to produce per-package `true`/`false` outputs; each test job skips itself when its package is untouched. Core/infra changes still trigger everything. Manual "run all" option available via GH UI.
- Split `test-general` into per-package jobs: `test-isaaclab` (2 shards), `test-isaaclab-rl`, `-mimic`, `-contrib`, `-teleop`, `-visualizers`, `-assets`. Each job is independently gated and sharded.
- Replaced hardcoded `include-files` whitelists with deterministic `shard-index % shard-count` sharding, fixing 7 silently skipped `isaaclab_tasks` tests.
- Broke out cuRobo Planner and SkillGen into separate test jobs with their own conditional execution gates.
- Fixed `conftest.py` exclude pattern that was matching a comma-separated string as one substring (never matched).
- Container is now killed on cancellation signals so jobs stop quickly via GH UI.
- JUnit results are converted to Markdown and displayed on the GH Actions job page, including skip reasons.
- Faster code checkout and updated `actions/checkout@v4`, `actions/upload-artifact@v4`, `actions/download-artifact@v4` to latest SHA-pinned versions.
- Extracted reusable composite action for repeating Docker test-run boilerplate.
- Added Docker image pull timing for diagnostics.
- Suppressed Node.js deprecation warnings in CI output.

## How test_train_environments.py is split

The test function `test_train_environments` is parametrized by *task_spec* (all registered Isaac environments) and *workflow* (`rl_games`, `rsl_rl`, `sb3`, `skrl`), so pytest expands it into `N_tasks x 4_workflows` individual test items. The `pytest_collection_modifyitems` hook in conftest.py runs after pytest builds this full list and keeps only every Nth item based on `TEST_SHARD_INDEX` / `TEST_SHARD_COUNT` env vars (e.g. shard 0/3 gets items 0, 3, 6, ...). Each CI runner sees ~1/3 of the parametrized combinations, cutting wall time from ~1.2h to ~24min per shard.

## Type of change

## Checklist


tools/conftest.py
```diff
@@ -301,14 +301,25 @@ def _collect_test_files(
...
+
+    # Apply sharding: sort deterministically, then select every Nth file
+    shard_index = os.environ.get("TEST_SHARD_INDEX", "")
+    shard_count = os.environ.get("TEST_SHARD_COUNT", "")
+    if shard_index and shard_count:
+        shard_index = int(shard_index)
+        shard_count = int(shard_count)
+        test_files.sort()
+        test_files = [f for i, f in enumerate(test_files) if i % shard_count == shard_index]
+        print(f"Shard {shard_index}/{shard_count}: selected {len(test_files)} test files")
```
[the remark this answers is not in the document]
myurasov-nv: Its not that hard to correctly provide in number between 0 and 2. :)

.github/workflows/build.yaml
```diff
@@ -83,551 +174,502 @@ jobs:
...
+  test-skillgen:
+    name: test-skillgen
+    runs-on: [self-hosted, gpu]
+    timeout-minutes: 120
+    continue-on-error: true
+    needs: [build-curobo, detect-changes]
+    if: >-
+      always() && needs.build-curobo.result == 'success' && (
+        needs.detect-changes.outputs.isaaclab-mimic == 'true' ||
+        needs.detect-changes.outputs.isaaclab-tasks == 'true' ||
```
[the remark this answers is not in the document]
myurasov-nv: Sure

myurasov-nv:
@greptile-apps

.github/workflows/build.yaml
```diff
@@ -42,12 +68,72 @@ env:
...
-    name: Build Base Docker image
+    name: Build Base Docker Image
     runs-on: [self-hosted, gpu]
     steps:
     - name: Checkout Code
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         fetch-depth: 1
         lfs: true
```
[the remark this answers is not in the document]
myurasov-nv: This is always needed

myurasov-nv:
@greptile-apps

myurasov-nv:
@greptile-apps

myurasov-nv:
@greptile-apps

AntoineRichard:
@greptile

myurasov-nv:
@greptile-apps

myurasov-nv:
@greptile-apps

.github/workflows/build.yaml
```diff
@@ -42,12 +67,132 @@ env:
...
+        echo "isaaclab-visualizers=$(has '^source/isaaclab_visualizers/')" >> "$GITHUB_OUTPUT"
+        echo "isaaclab-assets=$(has '^source/isaaclab_assets/')" >> "$GITHUB_OUTPUT"
+        echo "infra=$(has '^(\.github/|docker/|tools/|scripts/)')" >> "$GITHUB_OUTPUT"
+
+        echo ""
+        echo ""
+        echo "CHANGED FILES:"
+        echo "$changed" | while IFS= read -r f; do
+          case "$f" in
+            source/isaaclab/*)           echo "$f (isaaclab-core)" ;;
```
kellyguo11, line 162: we should test changes in apps/ folder as well, at least running the tests in core/physx/newton/tasks

.github/actions/run-package-tests/action.yml
```diff
@@ -0,0 +1,139 @@
...
+        filter-pattern: ${{ inputs.filter-pattern }}
+        shard-index: ${{ inputs.shard-index }}
+        shard-count: ${{ inputs.shard-count }}
+        curobo-only: ${{ inputs.curobo-only }}
+        quarantined-only: ${{ inputs.quarantined-only }}
+        include-files: ${{ inputs.include-files }}
+
+    - name: Report Test Results
+      if: always() && github.event.pull_request.head.repo.full_name == github.repository
+      uses: dorny/test-reporter@3d76b34a4535afbd0600d347b09a6ee5deb3ed7f # v2.6.0
```
kellyguo11: why do we need to switch to using SHA? feels harder to keep track across all the files than just the versions

.github/actions/combine-results/action.yml
```diff
@@ -1,103 +0,0 @@
-# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
kellyguo11, line 1: if we are removing this job, are the test jobs going to show fail? we previously set them up to continue on failure so that this combine-results job executes.

also, where are the final reports being presented with this removed?

.github/workflows/build.yaml
myurasov-nv: +1, done

.github/actions/combine-results/action.yml
ooctipus: It seems pretty clear to me to scroll and see the pass or failure on each tests, as I am reviewing, I am also trying to understand what is the benefit of /combine-results/action.yml beside it creates one more entry of pass and fail? It seems removable for me unless I am missing some important consideration.

.github/actions/run-package-tests/action.yml
myurasov-nv: According to G , potentially a malicious hijacking of third-party action's github can occur so that someone will run their code for free and steal all our secrets . Tbh, i think it is going a bit too far in being security-conscious. :joy_cat:

.github/actions/combine-results/action.yml
kellyguo11: It was supposed to generate a full report with the exact test cases that failed and their logs. it never really worked too well though, but if we were able to get that working, then it'll save us from going into each job, scrolling to the bottom to see which tests failed and why.

but I'm more concerned about capturing the fail/pass correctly, since before we depended on the combine-results job to check the overall pass/fail

myurasov-nv: Testing jobs are displaying results individually, so that's not really needed for capturing the results. If fact it succeeds always regardless of testing jobs. It was indeed combining JUnit XML files into one. Do we need that report? My suggestion is to bring it back once we do need that report. :)

myurasov-nv:
@greptile-apps

.github/actions/combine-results/action.yml
ooctipus: I think what kelly is suggesting is that we can go through each individual test to check the JUnit test result details when more inspection is needed, but if we have one summary report that collects all failure, it could be convenient to reference directly(think about case where we just need to copy one table and paste to github,chat,bugs and ask someone to fix rather than manually going through each tests and collect). But I do agree if you feel like this needs another PR to support, we don't have to have it in right now.

ooctipus (approved):

PR #5063, opened 2026-03-19
Fix randomize_rigid_body_material crash with Newton backend

ax-anoop:
## Summary

`randomize_rigid_body_material` crashes with `AttributeError: 'ArticulationView' object has no attribute 'link_paths'` when using the Newton physics backend (`presets=newton`).

The event term's `__init__` unconditionally uses PhysX-specific APIs to compute the number of collision shapes per body:

```python
for link_path in self.asset.root_view.link_paths[0]:
    link_physx_view = self.asset._physics_sim_view.create_rigid_body_view(link_path)
    self.num_shapes_per_body.append(link_physx_view.max_shapes)
```

Newton's `ArticulationView` does not expose `link_paths`, `max_shapes`, `get_material_properties()`, or `set_material_properties()`.

**Note:** The underlying MuJoCo-Warp engine *does* support runtime friction modification via `model.geom_friction` — this is a well-established MuJoCo capability used in production RL pipelines. The gap is in IsaacLab's Newton `ArticulationView` abstraction, which does not yet map these MuJoCo fields to the `get/set_material_properties` API that the event system expects.

### Changes

- Detect backend support by checking `hasattr(self.asset.root_view, "max_shapes")` before accessing material APIs
- Skip gracefully with a `logger.warning` when the backend lacks support
- When the backend provides `asset.num_shapes_per_body` directly (as Newton does), use it instead of the PhysX-specific `link_paths` workaround
- Guard `__call__` to return early if the backend doesn't support material randomization
- PhysX behavior is completely unchanged

### Reproduction

```bash
python scripts/train.py \
  --task Isaac-Velocity-Flat-Anymal-D-v0 \
  --num_envs 16 --max_iterations 1 \
  presets=newton
```

Crashes with:
```
AttributeError: 'ArticulationView' object has no attribute 'link_paths'. Did you mean: 'link_names'?
```

After this fix, Newton training runs successfully with the material randomization event skipped and a warning logged.

### Longer-term

Full material randomization on Newton requires exposing MuJoCo-Warp's `model.geom_friction` (and related geom properties) through the Newton `ArticulationView`'s `get_material_properties()` / `set_material_properties()` API. MuJoCo-Warp fully supports runtime per-environment friction modification — the gap is in the IsaacLab abstraction layer, not the physics engine.

## Test plan

- [x] Verified Newton training completes with this fix (`presets=newton`)
- [x] Verified PhysX training still works unchanged (default preset)
- [ ] Run existing unit tests for `randomize_rigid_body_material`


ooctipus:
https://github.com/isaac-sim/IsaacLab/pull/5098

can you check if this pr resolves your question?

StafaH:
Closing as #5098 resolved these issues

PR #5082, opened 2026-03-23
Add assemble_trocar task for G129-Dex3

mingxueg-nv:
# Description

Adds the **Assemble Trocar** manipulation task for the Unitree G1 (29-DoF + Dex3), with RLinf support.

Key additions:
- **Task MDP**: observations (body + Dex3 joint states), reward functions (4-stage sparse), termination conditions (timeout, success, object drop), and reset events (scene reset, task stage reset, random tray rotation).
- **Camera presets**: front camera and left/right wrist cameras (TiledCamera, 224×224) configured for GR00T visual input.
- **Robot presets**: G1 29-DoF + Dex3 articulation configuration.
- **GR00T data config**: `IsaacLabDataConfig` defining video/state/action modality keys, transforms (SinCos state encoding, min-max action normalization, color jitter), and model-specific settings.
- **RLinf extension update**: minor update to `isaaclab_contrib/rl/rlinf/extension.py` to support the new task registration.

## Type of change

## Checklist


Nic-Ma:
Hi @kellyguo11 , could you please help review this PR?

Thanks in advance.

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -468,6 +468,32 @@ def make_env_isaaclab() -> tuple:
...
 
                 env = gym.make(self.isaaclab_env_id, cfg=isaac_env_cfg, render_mode="rgb_array").unwrapped
+
+                _original_reset = env.reset
+                _n_warmup = max(getattr(env.cfg, "num_rerenders_on_reset", 0), 1)
+
+                def _patched_reset(*args, **kwargs):
+                    obs, extras = _original_reset(*args, **kwargs)
+                    try:
+                        import isaaclab_physx.renderers.isaac_rtx_renderer_utils as _rtx_utils
```
kellyguo11: why is this patch necessary? this is also restricting the renderer to RTX, is it not possible to use newton renderer as well?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/__init__.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Licensed under the Apache License, Version 2.0 (the "License");
```
kellyguo11: any reason why we also need the apache license?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/gr00t_config.py
```diff
@@ -0,0 +1,148 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers.
```
kellyguo11: duplicated license headers

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/gr00t_config.py
```diff
@@ -0,0 +1,148 @@
...
+This module defines customizable GR00T data configurations for different
+embodiments. Users can create their own data config classes by subclassing
+BaseDataConfig or copying/modifying the examples here.
+
+Example usage in run.sh:
+    export RLINF_DATA_CONFIG="policy.gr00t_config"
+    export RLINF_DATA_CONFIG_CLASS="policy.gr00t_config:IsaacLabDataConfig"
+"""
+
+from gr00t.data.dataset import ModalityConfig
```
kellyguo11, line 17: will these imports get picked up as well if we are not running this task? we want to make sure users are not forced to install gr00t if they don't need to run this particular task

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
mingxueg-nv: Hi, @kellyguo11 , during training and evaluation, the first frame of recorded videos shows the robot in its USD default pose instead of the environment's reset pose, the camera observation returned by `env.reset()` does not get the actual scene state.
I have already tried to set `num_rerenders_on_reset`, but it doesn't work: 

- In IsaacLab 2.3, [sim.render()](https://github.com/isaac-sim/IsaacLab/blob/5c2ec81cb17532d32f7922dd7fcaae40d123b71a/source/isaaclab/isaaclab/sim/simulation_context.py#L551) called `self.forward() + _app.update()` to flush data and trigger the full render pipeline. 

- In IsaacLab 3.0, this was removed, [sim.render()](https://github.com/isaac-sim/IsaacLab/blob/c26c70b4a9a33afbd2f4fff4fde2bf504cb8c059/source/isaaclab/isaaclab/sim/simulation_context.py#L665) now only updates visualizers. But [ManagerBasedEnv.reset()](https://github.com/isaac-sim/IsaacLab/blob/c26c70b4a9a33afbd2f4fff4fde2bf504cb8c059/source/isaaclab/isaaclab/envs/manager_based_env.py#L400) still relies on `sim.render()` for `num_rerenders_on_reset`, making it a no-op.

I have removed the limit of render to RTX in patch, thanks.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/__init__.py
mingxueg-nv: @kellyguo11 , thanks for reminder, I have removed all duplicated and apache headers from all files.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/gr00t_config.py
mingxueg-nv: Have removed, thanks

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/gr00t_config.py
mingxueg-nv: Hi, @kellyguo11 , the `config/__init__.py` only imports `camera_config` and `robot_config`, `gr00t_config.py` is not reachable through any import chain. This module is only loaded at runtime when RLinf resolves the string `gr00t_config:IsaacLabDataConfig` from the YAML config, so users who don't run this task will not trigger the gr00t import. For users running this task, we provide required installation step in the [RLinf setup docs](https://github.com/isaac-sim/IsaacLab/blob/develop/docs/source/overview/reinforcement-learning/rl_existing_scripts.rst#rlinf).

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
```diff
@@ -468,6 +468,26 @@ def make_env_isaaclab() -> tuple:
                 isaac_env_cfg.scene.num_envs = self.cfg.init_params.num_envs
 
                 env = gym.make(self.isaaclab_env_id, cfg=isaac_env_cfg, render_mode="rgb_array").unwrapped
+
+                _original_reset = env.reset
+
+                import omni.kit.app
+
+                _app = omni.kit.app.get_app()
```
kellyguo11: why is omni.kit.app needed to be explicitly defined here? could we call env.render() or access other APIs from the env?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/camera_config.py
```diff
@@ -0,0 +1,131 @@
...
+
+"""
+public camera configuration
+include the basic configuration for different types of cameras, support scene-specific parameter customization
+"""
+
+from collections.abc import Sequence
+
+import isaaclab.sim as sim_utils
+from isaaclab.sensors import CameraCfg, TiledCameraCfg
```
kellyguo11: we've deprecated TiledCamera recently, both can just use Camera directly now

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/robot_config.py
```diff
@@ -0,0 +1,310 @@
...
+    "right_hand_middle_0_joint": 60.0 * np.pi / 180.0,
+    "right_hand_thumb_0_joint": 0.0,
+    "right_hand_index_1_joint": 40.0 * np.pi / 180.0,
+    "right_hand_middle_1_joint": 40.0 * np.pi / 180.0,
+    "right_hand_thumb_1_joint": 0.0,
+    "right_hand_thumb_2_joint": 0.0,
+}
+
+
+G129_CFG_WITH_DEX3_BASE_FIX = ArticulationCfg(
```
kellyguo11: I think this could probably be moved to under isaaclab_assets so that other tasks could also use this robot if they'd like?

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/mdp/events.py
```diff
@@ -0,0 +1,259 @@
...
+        min_angle_deg, max_angle_deg = -rotation_range, rotation_range
+
+    # Get assets
+    tray = env.scene[tray_cfg.name]
+    trocar_1 = env.scene[trocar_1_cfg.name]
+    trocar_2 = env.scene[trocar_2_cfg.name]
+
+    # Get default states (initial positions from config)
+    # note: default_root_state is the local coordinate relative to the environment origin
+    tray_default_state = wp.to_torch(tray.data.default_root_state)[env_ids].clone()
```
kellyguo11: wp.to_torch is no longer necessary. we've added a new .torch mechanism to improve performance. the wp.to_torch wrapper will still be supported but is now deprecated

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/mdp/observations.py
```diff
@@ -0,0 +1,159 @@
...
+    joint_pos = wp.to_torch(env.scene["robot"].data.joint_pos)
+    joint_vel = wp.to_torch(env.scene["robot"].data.joint_vel)
+    joint_torque = wp.to_torch(env.scene["robot"].data.applied_torque)
+    device = joint_pos.device
+    batch = joint_pos.shape[0]
+
+    # Precompute and cache column indices
+    global _body_obs_cache
+    if _body_obs_cache["device"] != device or _body_obs_cache["idx_t"] is None:
+        body_joint_indices = [
```
kellyguo11: could this be defined together with the asset instead so that we can keep the mdp implementation clean?

kellyguo11 (commented):
can you also add some documentation for both RLinf and the environment? the environment can be added to https://isaac-sim.github.io/IsaacLab/develop/source/overview/environments.html and maybe RLinf can be added as a new section under https://isaac-sim.github.io/IsaacLab/develop/source/experimental-features/bleeding-edge.html

source/isaaclab_contrib/isaaclab_contrib/rl/rlinf/extension.py
mingxueg-nv: Hi, @kellyguo11 , I have removed this `_patch_reset` and set the `num_rerenders_on_reset`, it works well now, thanks.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/camera_config.py
mingxueg-nv: Have updated to CameraCfg directly, 🙏

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/config/robot_config.py
mingxueg-nv: I have moved to  `isaaclab_assets` , thanks.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/mdp/events.py
mingxueg-nv: Thank you for reminder, I have changed `wp.to_torch` to `.torch `.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/assemble_trocar/mdp/observations.py
mingxueg-nv: Have cleaned it, thank you.

mingxueg-nv:
Hi, @kellyguo11 , I have updated this PR and resolved comments. Also, we added the [environments.rst](https://github.com/isaac-sim/IsaacLab/pull/5082/changes#diff-dd6a414c0b9310329a78bb171df2a98aac6c4929fcc0c365ce1d74f88d894069) and [bleeding-edge.rst](https://github.com/isaac-sim/IsaacLab/pull/5082/changes#diff-47e2bbc262e7f48cf6727b5bc3fc071a8fcaae36fa9803b43e6cbed39a3bb833).

Could you please help to review again? Thank you!

kellyguo11:
Thanks a lot for addressing the feedback! I do feel like the overall implementation still deviates a bit from how other environments are implemented in Isaac Lab. Combined with the bot review, I think perhaps there are a few places we can try to clean up a bit more:

1. Task state monkey-patched onto the env object `env._task_stage`, `env._prev_stage_lift`, `env._lift_reward_locked`, etc. are created via setattr(env, ...) inside reward functions. The review bot's suggestion — a `@dataclass AssembleTrocarState` namespaced under `env.assemble_trocar_state` could work.

2. Stage machine runs inside a reward function `update_task_stage()` is called from `lift_trocars_reward()` and silently controls all four stage transitions using the thresholds passed to that one reward term. The other three reward functions call `get_task_stage()` and depend on `lift_trocars_reward` having already run. Stage updates should be an event term run once per step before rewards, not side-effectful reward logic.

3. Deprecated API usage `data.default_root_state` in `events.py` is deprecated. 

4. use logging instead of print()

5. `mdp/__init__.py` still uses manual imports + __all__ rather than lazy_export().
`events: EventCfg = EventCfg()` was missing a type annotation.

6. `get_trocar_tip_position()` hardcodes `/World/envs/env_0/...` - does this environment work with num_envs > 0? does it always have to be hard-coded to 0? we generally expect our environments to be parallelizable.

7. `tray` being defined as an `ArticulationCfg`, can it actually be a RigidObject instead?

kellyguo11:
@ooctipus what are your thoughts on this PR?

mingxueg-nv:
> Thanks a lot for addressing the feedback! I do feel like the overall implementation still deviates a bit from how other environments are implemented in Isaac Lab. Combined with the bot review, I think perhaps there are a few places we can try to clean up a bit more:
> 
> 1. Task state monkey-patched onto the env object `env._task_stage`, `env._prev_stage_lift`, `env._lift_reward_locked`, etc. are created via setattr(env, ...) inside reward functions. The review bot's suggestion — a `@dataclass AssembleTrocarState` namespaced under `env.assemble_trocar_state` could work.
> 2. Stage machine runs inside a reward function `update_task_stage()` is called from `lift_trocars_reward()` and silently controls all four stage transitions using the thresholds passed to that one reward term. The other three reward functions call `get_task_stage()` and depend on `lift_trocars_reward` having already run. Stage updates should be an event term run once per step before rewards, not side-effectful reward logic.
> 3. Deprecated API usage `data.default_root_state` in `events.py` is deprecated.
> 4. use logging instead of print()
> 5. `mdp/__init__.py` still uses manual imports + **all** rather than lazy_export().
>    `events: EventCfg = EventCfg()` was missing a type annotation.
> 6. `get_trocar_tip_position()` hardcodes `/World/envs/env_0/...` - does this environment work with num_envs > 0? does it always have to be hard-coded to 0? we generally expect our environments to be parallelizable.
> 7. `tray` being defined as an `ArticulationCfg`, can it actually be a RigidObject instead?

Thanks @kellyguo11 for the detailed review! 

- 1-5: Done — `@dataclass AssembleTrocarState` for task state, stage machine decoupled as weight=0 reward term, deprecated API migrated, print → logging, `lazy_export()` + type annotation added.
- 6: Intentional — `env_0` is only used once to read a static mesh offset; with `replicate_physics=True` all envs share identical geometry, so it works with any num_envs.
- 7: Keeping as-is — the USD asset has `PhysicsArticulationRootAPI` and `FixedJoints` baked in, so `RigidObjectCfg` would cause `PhysX` errors. `ArticulationCfg(actuators={}) `is for passive articulations.

I've also verified the updated code locally — the environment and pipeline runs well. Could you help verify if the implementation now better aligns with the patterns used in other Isaac Lab environments? Happy to make further adjustments if you spot anything else!

kellyguo11 (approved):

PR #5091, opened 2026-03-24
Caches string_to_callable module lookups with functools.cache

kellyguo11:
# Description

`string_to_callable()` is called during manager initialization and configclass deserialization to resolve `module:attribute` strings into callables. Each call runs `importlib.import_module()` even when the same string has been resolved before.

This PR extracts the non-lambda resolution path into `_resolve_module_callable()` and decorates it with `functools.cache`, so repeated lookups (e.g. across environment resets or multi-env setups) skip the import overhead.

Lambda expressions are **not** cached since they are `eval()`'d fresh each time by design.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.54.3"
+version = "0.54.4"
 
 # Description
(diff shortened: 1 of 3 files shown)


kellyguo11:
@greptileai review

ooctipus:
Each call runs importlib.import_module() even when the same string has been resolved before.

I don't think this is true?  It only resolve once, once resolved its a fully resolved class no string anymore

PR #5102, opened 2026-03-24
Makes torch/array imports lazy in dict.py for lightweight environments

kellyguo11:
## Summary

Defers the top-level `import torch` and `from .array import ...` in `isaaclab.utils.dict` so that the module can be imported in environments where `torch`/`warp` are not installed (e.g. lightweight CI runners, linters, pure-Python test jobs).

## Changes

- Replaces `isinstance(obj, torch.Tensor)` with a private `_is_torch_tensor()` helper that catches `ModuleNotFoundError`
- Moves `TENSOR_TYPE_CONVERSIONS` / `TENSOR_TYPES` access into a `_get_array_constants()` helper called lazily at first use inside `convert_dict_to_backend()`

## Why

The import chain `isaaclab.utils.__init__` → `configclass` → `dict` → `import torch` forces torch to be present even for code that never touches tensors. This blocks running pure-Python utility tests (string, dict, configclass, logger, version) on standard CI runners without GPU/torch.

## Testing

- Verified `from isaaclab.utils.dict import class_to_dict, update_class_from_dict` succeeds with torch blocked
- Verified `class_to_dict` correctly handles both plain objects and torch tensors when torch is available
- No behavioral change when torch is installed

### DIFF
--- a/source/isaaclab/isaaclab/utils/dict.py
+++ b/source/isaaclab/isaaclab/utils/dict.py
@@ -11,10 +11,24 @@ import json
 from collections.abc import Iterable, Mapping, Sized
 from typing import Any
 
-import torch
-
-from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
 from .string import ResolvableString, callable_to_string, string_to_slice
+
+
+def _is_torch_tensor(obj: object) -> bool:
+    """Check if *obj* is a ``torch.Tensor`` without importing torch at module level."""
+    try:
+        import torch  # noqa: F811
+    except ModuleNotFoundError:
+        return False
+    return isinstance(obj, torch.Tensor)
+
+
+def _get_array_constants():
+    """Lazily import TENSOR_TYPE_CONVERSIONS and TENSOR_TYPES from .array."""
+    from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
+
+    return TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
+
 
 """
 Dictionary <-> Class operations.
@@ -42,7 +56,7 @@ def class_to_dict(obj: object) -> dict[str, Any]:
     # convert object to dictionary
     if isinstance(obj, dict):
         obj_dict = obj
-    elif isinstance(obj, torch.Tensor):
+    elif _is_torch_tensor(obj):
         # We have to treat torch tensors specially because `torch.tensor.__dict__` returns an empty
         # dict, which would mean that a torch.tensor would be stored as an empty dict. Instead we
         # want to store it directly as the tensor.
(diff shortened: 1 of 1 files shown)


source/isaaclab/isaaclab/utils/dict.py
```diff
@@ -11,11 +11,25 @@
...
 from .string import ResolvableString, callable_to_string, string_to_slice
 
+
+def _is_torch_tensor(obj: object) -> bool:
+    """Check if *obj* is a ``torch.Tensor`` without importing torch at module level."""
+    try:
+        import torch  # noqa: F811
+    except ModuleNotFoundError:
+        return False
+    return isinstance(obj, torch.Tensor)
```
kellyguo11, line 23: 🟡 **Performance:** `_is_torch_tensor` is called recursively from `class_to_dict` — once per object in the tree. While `import torch` hits `sys.modules` when torch is installed, the repeated function-call + try/except + dict-lookup overhead adds up on large configs.

Consider caching the torch.Tensor type at module level on first use:

```suggestion
_torch_tensor_type = None


def _is_torch_tensor(obj: object) -> bool:
    """Check if *obj* is a ``torch.Tensor`` without importing torch at module level."""
    global _torch_tensor_type
    if _torch_tensor_type is None:
        try:
            import torch
        except ModuleNotFoundError:
            return False
        _torch_tensor_type = torch.Tensor
    return isinstance(obj, _torch_tensor_type)
```

This makes subsequent calls a simple `isinstance` check with zero import overhead.

source/isaaclab/isaaclab/utils/dict.py
```diff
@@ -11,11 +11,25 @@
...
+
+
+def _get_array_constants():
+    """Lazily import TENSOR_TYPE_CONVERSIONS and TENSOR_TYPES from .array."""
+    from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
+
+    return TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
```
kellyguo11, line 30: 🔵 **Missing return type hint.** Per style guide, type hints are required in function signatures.

```suggestion
def _get_array_constants() -> tuple[dict, dict]:
    """Lazily import TENSOR_TYPE_CONVERSIONS and TENSOR_TYPES from .array."""
    from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES

    return TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
```

Also consider caching the result in a module-level variable so repeated calls to `convert_dict_to_backend` don't re-run the import machinery each time:

```python
_array_constants: tuple[dict, dict] | None = None

def _get_array_constants() -> tuple[dict, dict]:
    global _array_constants
    if _array_constants is None:
        from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
        _array_constants = (TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES)
    return _array_constants
```

source/isaaclab/isaaclab/utils/dict.py
```diff
@@ -11,11 +11,25 @@
...
-import torch
-
-from .array import TENSOR_TYPE_CONVERSIONS, TENSOR_TYPES
 from .string import ResolvableString, callable_to_string, string_to_slice
 
+
+def _is_torch_tensor(obj: object) -> bool:
+    """Check if *obj* is a ``torch.Tensor`` without importing torch at module level."""
+    try:
+        import torch  # noqa: F811
```
kellyguo11, line 20: 🔵 **Stale noqa comment.** The `# noqa: F811` suppression was needed when there was a top-level `import torch` that this line would shadow. Since that top-level import is now removed, there's no redefinition to suppress.

```suggestion
        import torch
```

kellyguo11 (commented):
## PR Review: #5102

### Summary
This PR makes `torch` and `isaaclab.utils.array` imports lazy in `dict.py` so the module can be imported in lightweight environments without `torch`/`warp`. The approach is sound — it correctly breaks the `utils.__init__` → `configclass` → `dict` → `torch` import chain. The implementation is functional but has a performance concern worth addressing.

### Architecture Assessment
No breaking API changes — all public function signatures are unchanged. The lazy import pattern is consistent with the existing `lazy_loader` usage in `utils/__init__.py`. Cross-module impact is minimal since only the internal import mechanism changed, not behavior when torch is available.

### CI Status
- **pre-commit:** ✅ Pass
- **license-check:** ✅ Pass
- **isaaclab_newton:** ❌ Fail — this appears to be a pre-existing/infrastructure failure unrelated to this PR (the develop branch also has CI failures from different tests).
- Several core/tasks test jobs are still pending.

### Key Findings
1. 🟡 **Performance: `_is_torch_tensor()` and `_get_array_constants()` do an import lookup on every call** — see inline comments for suggested caching pattern.
2. 🔵 **Style: private helpers placed before section comment** — Per Isaac Lab file structure conventions, `_private` functions should appear after all public functions.
3. 🔵 **Stale `# noqa: F811`** — no longer needed since the top-level `import torch` was removed.
4. 🔵 **Missing return type hint on `_get_array_constants()`**
5. 🔵 **Missing CHANGELOG entry** — The isaaclab extension is modified but `source/isaaclab/docs/CHANGELOG.rst` is not updated.

*Note: the pre-existing bug where `convert_dict_to_backend`'s recursive call (line 289) doesn't forward `backend`/`array_types` args is not introduced by this PR, but worth a follow-up fix.*

PR #5124, opened 2026-03-29
[Newton] Fix kit-less mode and Newton v0.2.2 compatibility

Kripner:
The `dev/newton` branch supports running Isaac Lab without Omniverse (kit-less mode using the Newton physics backend), but several issues prevent it from working outside of Isaac Sim's bundled Python environment. Two fixes:

- Guard unconditional `import omni` statements that crash at import time when Omniverse is not installed. Add HTTPS fallback for downloading S3-hosted USD assets without `omni.client`. Remove `omniverseclient` from `setup.py` (never imported). Fix stale `warp-lang` dev pins.
- Pin Newton to v0.2.2 (the pinned commit `51ce35e8` lacks `body_label` used by `newton_replicate.py`). Fix API mismatches: `articulation_data.py` tensor indexing, `direct_rl_env_warp.py` kwarg name (`mask` -> `env_mask`), add `has_rtx_sensors()` stub.

All changes are backwards-compatible - when Omniverse is available, behavior is unchanged.

## Context

The Newton kit-less mode is a key feature of Isaac Lab 3.0 Beta, but the `dev/newton` branch currently crashes on import without Omniverse installed. These fixes enable headless RL training on HPC clusters using only CUDA (no Omniverse/Isaac Sim).

Tested environments (training + Newton OpenGL visualization):
- `Isaac-Cartpole-Direct-Warp-v0` (direct workflow)
- `Isaac-Reach-Franka-Warp-v0` (manager-based warp workflow)
- `Isaac-Open-Drawer-Franka-v0` (manager-based standard workflow)

### DIFF
--- a/source/isaaclab/isaaclab/envs/__init__.py
+++ b/source/isaaclab/isaaclab/envs/__init__.py
@@ -42,7 +42,12 @@ For more information about the workflow design patterns, see the `Task Design Wo
 .. _`Task Design Workflows`: https://isaac-sim.github.io/IsaacLab/source/features/task_workflows.html
 """
 
-from . import mdp, ui
+from . import mdp
+
+try:
+    from . import ui
+except ModuleNotFoundError:
+    pass
 from .common import VecEnvObs, VecEnvStepReturn, ViewerCfg
 from .direct_rl_env import DirectRLEnv
 from .direct_rl_env_cfg import DirectRLEnvCfg
--- a/source/isaaclab/isaaclab/envs/direct_rl_env.py
+++ b/source/isaaclab/isaaclab/envs/direct_rl_env.py
@@ -28,7 +28,10 @@ from isaaclab.utils.timer import Timer
 
 from .common import VecEnvObs, VecEnvStepReturn
 from .direct_rl_env_cfg import DirectRLEnvCfg
-from .ui import ViewportCameraController
+try:
+    from .ui import ViewportCameraController
+except ModuleNotFoundError:
+    ViewportCameraController = None
 from .utils.spaces import sample_space, spec_to_gym_space
 
 # import logger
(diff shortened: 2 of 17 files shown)


ooctipus:
dev/newton is not actively developed anymore.

develop branch should support kit-less path very well!

diegoferigo-rai:
> develop branch should support kit-less path very well!

Note: as I outlined in https://github.com/isaac-sim/IsaacLab/issues/5084#issuecomment-4109565852, `omniverseclient` is still needed in the kit-less setup also in the `develop` branch, and I'm not sure if this in intended.

ooctipus:
`omniverseclient` should be pip installable publicly , so It should be kit-less compatible. Ahh I see your concern in the comment, let me discuss this with team

Kripner:
> dev/newton is not actively developed anymore.
> 
> develop branch should support kit-less path very well!

OK, closing. I'd recommend deleting the stale branch - right now we have `develop`, `dev/newton`, and `feature/newton`, and it's unclear which one to use.

diegoferigo-rai:
> `omniverseclient` should be pip installable publicly , so It should be kit-less compatible. Ahh I see your concern in the comment, let me discuss this with team

That would be much appreciated. More broadly, it would be ideal if IsaacLab 3.0 could better isolate the usage of `omniverseclient`, ensuring it is only imported when strictly required.

From my current understanding, in a kit-less setup its role appears limited (primarily asset downloading). If that holds, it would be preferable to avoid making it a mandatory dependency for all users, particularly for applications that rely solely on Newton and do not need this functionality.

A cleaner approach might be to treat `omniverseclient` as an optional extra (e.g., via `extras_require` in `setup.py`), rather than a hard dependency. This would improve modularity and reduce unnecessary installation overhead.

cc @myurasov-nv

PR #5149, opened 2026-04-02
Fix pyglet version pin to exclude incompatible 3.0 pre-release

Kripner:
Pin `pyglet>=2.1.6,<3.0` in `source/isaaclab/setup.py` to prevent pip/uv from resolving to `pyglet==3.0.dev2`, which removes `pyglet.graphics.shader.Shader` and breaks the Newton OpenGL viewer.

Test of the fix: Newton visualizer launches without `ImportError`: `./isaaclab.sh -p scripts/environments/zero_agent.py --task Isaac-Reach-Franka-v0 --num_envs 4 env.sim.physics=newton --visualizer newton`


StafaH:
This was resolved by #5263

PR #5098, opened 2026-03-24
Unify physics randomization events across PhysX and Newton

ooctipus:
# Description

Asset APIs added:

- PhysX: set_material_properties_index / set_material_properties_mask on Articulation, RigidObject, and RigidObjectCollection for vectorized material property writes.
- Newton: set_friction_index / set_friction_mask, set_restitution_index / set_restitution_mask, and num_shapes property on all three asset types. Newton uses a single friction coefficient (mu), so friction and restitution are separate APIs.

Event terms updated:

- randomize_rigid_body_material — auto-detects backend at init; PhysX uses 3-tuple materials via set_material_properties_index, Newton uses separate set_friction_index / set_restitution_index.
- randomize_rigid_body_collider_offsets — converted from duplicated PhysX/Newton branches to a single __call__ using writer lambdas set at init. PhysX writes rest/contact offsets; Newton maps to shape_margin / shape_gap and notifies solver.
- randomize_rigid_body_com — passes body_ids through to set_coms_index and handles Newton's position-only (vec3) vs PhysX's full pose (pos + quat). NOTE: This is only implemented for completion but there seem to be an bug in mujoco warp in handing change of this value. 
- randomize_rigid_body_inertia — new event term for randomizing body inertia tensors.

## Type of change

- New feature (non-breaking change which adds functionality)

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5151, opened 2026-04-02
Skip material randomization on physics backends that don't support it

Kripner:
## Summary
- `randomize_rigid_body_material` uses PhysX-specific APIs (`link_paths`, `max_shapes`, `get/set_material_properties`) that don't exist on Newton's `ArticulationView`, causing `AttributeError` when running tasks like `Isaac-Open-Drawer-Franka-v0` with `env.sim=newton`.
- Detect backend support via `hasattr` check on `get_material_properties` and skip the term with a warning when unsupported.
- Also use the asset's `num_shapes_per_body` property when available (Newton provides this directly) instead of the PhysX-specific `link_paths` workaround.

### DIFF
--- a/source/isaaclab/isaaclab/envs/mdp/events.py
+++ b/source/isaaclab/isaaclab/envs/mdp/events.py
@@ -203,22 +203,37 @@ class randomize_rigid_body_material(ManagerTermBase):
                 f" with type: '{type(self.asset)}'."
             )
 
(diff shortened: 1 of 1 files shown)


ooctipus:
https://github.com/isaac-sim/IsaacLab/pull/5098

this PR should support it : D, I am closing this PR for now.

PR #5162, opened 2026-04-03
Move TiledCamera implementation into Camera and deprecate TiledCamera. 

nvsekkin:
# Description

Unify `Camera` and `TiledCamera` by moving TiledCamera's renderer-based implementation into `Camera` and making `TiledCamera` a deprecated subclass alias.

### What changed

**`base_renderer.py` (interface change):**
- `write_output(render_data, output_name, output_data)` replaced by `read_output(render_data, camera_data)`. The renderer now receives the full `CameraData` instance and owns the per-output iteration loop.

**`camera.py`:**
- `_initialize_impl` now uses `Renderer(cfg.renderer_cfg)` + `renderer.prepare_stage()` + `renderer.create_render_data()` instead of replicator imports and per-camera render product loops
- `_create_buffers` uses eager pre-allocation + `renderer.set_outputs()` instead of lazy allocation
- Removed: `_rep_registry`, `_render_product_paths`, `_create_annotator_data()`, `_process_annotator_output()`, `render_product_paths` property, `omni.replicator.core` import, `SyntheticData` import

**`camera_data.py`: `info` type change:**
- `CameraData.info` changed from `list[dict[str, Any]]` to `dict[str, Any]`, matching the Replicator annotator output structure. The old `list[dict]` was an artifact of the original `Camera`'s per-camera-annotator architecture. With tiled rendering, all cameras share a single annotator pipeline, so the per-camera list dimension carried no information. Access is now `camera.data.info["semantic_segmentation"]` directly.

**`tiled_camera.py`:**
- Replaced entire class with a thin `class TiledCamera(Camera)` subclass that emits a `DeprecationWarning` on instantiation

**`tiled_camera_cfg.py`:**
- Added `DeprecationWarning` in `__post_init__`

**Test suite:**
- Ported 6 unique tests from `test_tiled_camera.py` → `test_camera.py` (multi-regex init, all annotators, segmentation non-colorize, normals unit length, data types ordering, frame offset)
- Slimmed `test_tiled_camera.py` to 4 focused deprecation/compatibility tests
- Updated `info` assertions across all test files to use flat `dict` access pattern (`info["key"]` instead of `info[0]["key"]`)

### Breaking changes
- `render_product_paths` property removed (confirmed zero external usage across entire repo — only referenced internally in `camera.py`, `isaac_rtx_renderer.py`, and `ovrtx_renderer.py`)
- `CameraData.info` changed from `list[dict[str, Any]]` to `dict[str, Any]`. `Camera` users (old): replace `camera.data.info[cam_idx][data_type]` with `camera.data.info[data_type]`. `TiledCamera` users (old): access pattern `camera.data.info[data_type]` is unchanged.
- `BaseRenderer.write_output` replaced by `BaseRenderer.read_output(render_data, camera_data)`. Custom renderer implementations must update their signature.

## Type of change

## Checklist


source/isaaclab/config/extension.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "4.5.25"
+version = "4.5.26"
```
nvsekkin: maybe bump minor at least, since there is a breaking change and deprecation warning.

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,37 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+Changed
+^^^^^^^
+
+* Unified :class:`~isaaclab.sensors.camera.Camera` and :class:`~isaaclab.sensors.camera.TiledCamera`
+  into a single implementation. :class:`Camera` now delegates all rendering to the
+  :class:`~isaaclab.renderers.Renderer` abstraction (same approach :class:`TiledCamera` used).
+  The public API is unchanged for :class:`Camera` users.
+* **Breaking (TiledCamera users):** :attr:`~isaaclab.sensors.camera.CameraData.info` now correctly
```
nvsekkin: this was a bug from before, fixed while moving TiledCamera impl into Camera

source/isaaclab/test/sensors/test_multi_tiled_camera.py
```diff
@@ -29,6 +29,10 @@
 import isaaclab.sim as sim_utils
 from isaaclab.sensors.camera import TiledCamera, TiledCameraCfg
 
+# Deprecation warnings from TiledCamera/TiledCameraCfg are expected in this file;
+# the deprecation mechanism itself is validated in test_tiled_camera.py.
+pytestmark = pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
nvsekkin, line 34: this is needed otherwise, the TiledCamera gets noisy in tests

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -68,6 +73,26 @@ def create_render_data(self, sensor: SensorBase) -> IsaacRtxRenderData:
         import omni.replicator.core as rep
         from pxr import UsdGeom
 
+        settings = get_settings_manager()
+        isaac_sim_version = get_isaac_sim_version()
+
+        if isaac_sim_version.major >= 6:
```
nvsekkin, line 91: This was inside Camera impl, but should be in the renderer

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -476,135 +442,38 @@ def _initialize_impl(self):
...
-                self._rep_registry[name].append(rep_annotator)
-
-        # Create internal buffers
+            self._sensor_prims.append(UsdGeom.Camera(cam_prim))
+
+        # View needs to exist before creating render data
+        self.render_data = self.renderer.create_render_data(self)
+
+        # Create internal buffers (includes intrinsic matrix and pose init)
         self._create_buffers()
```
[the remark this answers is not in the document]
nvsekkin: good catch, restored it for RTX path

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -476,135 +442,38 @@ def _initialize_impl(self):
...
+            if output_name == "rgb":
+                continue
+            self.renderer.write_output(self.render_data, output_name, output_data)
+
+        # Broadcast renderer info (e.g. segmentation label mappings) to all per-camera entries
+        renderer_info = getattr(self.render_data, "renderer_info", None)
+        if renderer_info:
+            for data_type, metadata in renderer_info.items():
+                for cam_idx in range(self._view.count):
+                    self._data.info[cam_idx][data_type] = metadata
```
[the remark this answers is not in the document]
nvsekkin: You’re right. that’s intentional here because the segmentation metadata applies to all envs in a single pass and scoping it to env_ids would actually desync info.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -632,19 +501,82 @@ def _check_supported_data_types(self, cfg: CameraCfg):
...
+                data_dict["instance_id_segmentation_fast"] = torch.zeros(
+                    (self._view.count, self.cfg.height, self.cfg.width, 1), device=self.device, dtype=torch.int32
+                ).contiguous()
+
+        self._data.output = data_dict
         self._data.info = [{name: None for name in self.cfg.data_types} for _ in range(self._view.count)]
+        self.renderer.set_outputs(self.render_data, self._data.output)
 
     def _update_intrinsic_matrices(self, env_ids: Sequence[int]):
         """Compute camera's matrix of intrinsic parameters.
```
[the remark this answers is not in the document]
nvsekkin: Added a short note on CameraData.output that dict key order is not guaranteed to match data_types and callers should index by name or iterate data_types explicitly

source/isaaclab/isaaclab/sensors/camera/tiled_camera_cfg.py
```diff
@@ -15,6 +16,18 @@
...
+    """
 
     class_type: type["TiledCamera"] | str = "{DIR}.tiled_camera:TiledCamera"
+
+    def __post_init__(self):
+        warnings.warn(
+            "TiledCameraCfg is deprecated. Use CameraCfg directly.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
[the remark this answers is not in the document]
nvsekkin: no need to make changes for hypothetical future parent hooks on a deprecated shim, also super().__post_init__() is not safe with _combined_function

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -3,359 +3,34 @@
...
-            f"\tcolorize instance id segm.: {self.cfg.colorize_instance_id_segmentation}\n"
-            f"\tupdate period (s): {self.cfg.update_period}\n"
-            f"\tshape        : {self.image_shape}\n"
-            f"\tnumber of sensors : {self._view.count}"
+    def __init__(self, cfg):
+        warnings.warn(
+            "TiledCamera is deprecated and will be removed in a future release. "
+            "Use Camera directly — it now uses the same Renderer abstraction.",
+            DeprecationWarning,
+            stacklevel=2,
```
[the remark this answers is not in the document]
nvsekkin: done

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -632,19 +501,82 @@ def _check_supported_data_types(self, cfg: CameraCfg):
...
-        # the memory will be allocated when the buffer() function is called for the first time.
-        self._data.output = {}
+        # -- output data (eagerly pre-allocated so renderer.set_outputs() can hold tensor references)
+        data_dict = dict()
+        if "rgba" in self.cfg.data_types or "rgb" in self.cfg.data_types:
+            data_dict["rgba"] = torch.zeros(
+                (self._view.count, self.cfg.height, self.cfg.width, 4), device=self.device, dtype=torch.uint8
+            ).contiguous()
+        if "rgb" in self.cfg.data_types:
+            data_dict["rgb"] = data_dict["rgba"][..., :3]
```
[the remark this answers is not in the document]
nvsekkin: @ooctipus do you have an input for this?

source/isaaclab/config/extension.toml
nvsekkin: done

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -205,14 +252,16 @@ def tiling_grid_shape():
 
         num_tiles_x = tiling_grid_shape()[0]
 
+        if render_data.renderer_info is None:
```
bareya: What is semantical meaning of `renderer_info` being set to `None`?  Why it can not be initialized to empty dictionary? You never check if `renderer_info` is `None` but here?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -96,13 +95,11 @@ class Camera(SensorBase):
     }
     """The set of sensor types that are not supported by the camera class."""
 
-    SIMPLE_SHADING_MODES: dict[str, int] = {
-        "simple_shading_constant_diffuse": 0,
-        "simple_shading_diffuse_mdl": 1,
-        "simple_shading_full_mdl": 2,
+    SIMPLE_SHADING_TYPES: set[str] = {
```
bareya, line 98: This belongs to RTX renderer.  Is this intentional that you left it here, and you are going to move it into subsequent MR?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -187,6 +156,9 @@ def __init__(self, cfg: CameraCfg):
         self._sensor_prims: list[UsdGeom.Camera] = list()
         # Create empty variables for storing output data
         self._data = CameraData()
+        # Renderer and render data — initialized in _initialize_impl
+        self.renderer: Renderer | None = None
```
bareya: self.renderer type hint is incorrect - it should be BaseRenderer.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -443,12 +406,19 @@ def _initialize_impl(self):
...
+        self.renderer = Renderer(self.cfg.renderer_cfg)
+        logger.info("Using renderer: %s", type(self.renderer).__name__)
+
+        # Stage preprocessing must happen before creating the view because the view keeps
+        # references to prims located in the stage.
+        self.renderer.prepare_stage(self.stage, self._num_envs)
+
+        # Create a view for the sensor with Fabric enabled for fast pose queries.
+        # TODO: remove sync_usd_on_fabric_write=True once the GPU (cuda:0) Fabric sync bug in
+        # renderer.update_transforms() is fixed. Without it, poses are stale on the GPU path.
```
bareya: Why do we keep update_transforms commented out?  Perhaps this should be removed?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -476,135 +442,38 @@ def _initialize_impl(self):
...
+        self.renderer.update_transforms()
+        self.renderer.render(self.render_data)
+
+        for output_name, output_data in self._data.output.items():
+            if output_name == "rgb":
+                continue
+            self.renderer.write_output(self.render_data, output_name, output_data)
+
+        # Broadcast renderer info (e.g. segmentation label mappings) to all per-camera entries
+        renderer_info = getattr(self.render_data, "renderer_info", None)
```
bareya: nope this is wrong, you are leaking implementation details of the renderer, we are missing some API call here?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
nvsekkin: true, this was just a workaround for the 'dataclass', will swap it with something like field(default_factory=dict)

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: agreed, copied directly from TiledCamera impl, will update on this PR

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: this was additional context for when we fix the sync issue on the cuda path, I will remove second line

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: good point, will push this up to BaseRenderer

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: agreed, I didn't want to crowd this PR more. I think we can address in a follow-up PR that moves logic from _create_buffers into renderer.

source/isaaclab/isaaclab/sensors/camera/camera.py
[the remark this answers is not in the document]
kellyguo11: I think this should be ok, rgb is just an alias for rgba to return only a 3 channel buffer. as long as we don't duplicate the buffers for them, it's not a big deal if both data types are exposed

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -3,359 +3,35 @@
...
-            f"\tsemantic filter : {self.cfg.semantic_filter}\n"
-            f"\tcolorize semantic segm.   : {self.cfg.colorize_semantic_segmentation}\n"
-            f"\tcolorize instance segm.   : {self.cfg.colorize_instance_segmentation}\n"
-            f"\tcolorize instance id segm.: {self.cfg.colorize_instance_id_segmentation}\n"
-            f"\tupdate period (s): {self.cfg.update_period}\n"
-            f"\tshape        : {self.image_shape}\n"
-            f"\tnumber of sensors : {self._view.count}"
+        warnings.warn(
+            "TiledCamera is deprecated and will be removed in a future release. "
+            "Use Camera directly — it now uses the same Renderer abstraction.",
```
kellyguo11: could we also explicitly mention that Camera will now support the vectorization optimizations of TiledCamera

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,37 @@
...
+  returns ``list[dict[str, Any]]`` (per-camera, then per-data-type) as documented in
+  :class:`~isaaclab.sensors.camera.CameraData`. :class:`TiledCamera` previously returned a flat
+  ``dict``, which violated the documented contract. Migration: replace
+  ``camera.data.info[data_type]`` with ``camera.data.info[cam_idx][data_type]``.
```
kellyguo11: this may have been done intentionally. If we don't flatten the buffer for the tiled version, would it still allow slicing if we need to index a subset of environments like `camera.data.info[0:10][data_type]`?

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
nvsekkin: added this in docstring and warnings

source/isaaclab/docs/CHANGELOG.rst
nvsekkin: Good point, the docs in camera.rst do show the flat dict pattern (tiled_camera.data.info['semantic_segmentation']), so it may have been intentional for TiledCamera.

That said, CameraData.info is typed as list[dict[str, Any]], and Camera's original implementation also used list[dict]. The renderer info is shared metadata (e.g. label mappings), identical for all cameras, so info[0]["semantic_segmentation"] works for the common case.

I went with list[dict] to match the documented type. The practical impact is minimal: [0] indexing is the only change for TiledCamera users?

source/isaaclab/isaaclab/sensors/camera/camera_data.py
```diff
@@ -53,6 +53,9 @@ class CameraData:
     The format of the data is available in the `Replicator Documentation`_. For semantic-based data,
     this corresponds to the ``"data"`` key in the output of the sensor.
 
+    Key iteration order is not guaranteed to match the order of entries in the camera configuration's
+    ``data_types`` list; index by name or iterate ``data_types`` and look up each key explicitly.
+
```
huidongc: this behavior isn't introduced by your refactoring, right?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -187,6 +156,9 @@ def __init__(self, cfg: CameraCfg):
         self._sensor_prims: list[UsdGeom.Camera] = list()
         # Create empty variables for storing output data
         self._data = CameraData()
+        # Renderer and render data — initialized in _initialize_impl
+        self.renderer: BaseRenderer | None = None
+        self.render_data = None
```
huidongc: what is IsaacLab naming convention for member variables? Do we need to prefix with underscore?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -205,14 +177,12 @@ def __init__(self, cfg: CameraCfg):
...
+        """Unsubscribes from callbacks and cleans up renderer resources."""
         # unsubscribe callbacks
         super().__del__()
-        # delete from replicator registry
-        for _, annotators in self._rep_registry.items():
-            for annotator, render_product_path in zip(annotators, self._render_product_paths):
-                annotator.detach([render_product_path])
-                annotator = None
+        # cleanup render resources (renderer may be None if never initialized)
+        if hasattr(self, "renderer") and self.renderer is not None:
```
huidongc: why do you have to run this defensive hasattr() check?

source/isaaclab/docs/CHANGELOG.rst
kellyguo11: ah I see, if they are always the same across all cameras, would it make sense to just return a dict instead of list[dict]? so automatically we just return info[0]

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: moved the population logic into the renderer implementation

source/isaaclab/isaaclab/sensors/camera/camera.py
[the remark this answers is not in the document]
nvsekkin: thanks Kelly, resolving.

source/isaaclab/isaaclab/sensors/camera/camera_data.py
nvsekkin: @huidongc Technically the old Camera preserved cfg.data_types order (it built _rep_registry from data_types and iterated that to populate output). The new implementation uses a fixed allocation order in _create_buffers, so key order may differ from data_types. The docstring documents this new behavior. In practice nobody should rely on dict iteration order for semantic key lookup, but I wanted to be explicit about the change. (TiledCamera already worked like this)

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: this was a direct copy from TiledCamera original implementation. I think it's safe to remove this check in this PR

source/isaaclab/docs/CHANGELOG.rst
nvsekkin: I am in favor of moving to dict, this was a side effect of the inital Camera implementation and is not needed. I think we can document this as a breaking change.

nvsekkin: done

kellyguo11:
One of the doc pages will also need an update

/home/runner/work/IsaacLab/IsaacLab/docs/source/how-to/save_camera_output.rst:37: WARNING: end-at pattern not found: single_cam_info = camera.data.info[camera_index] [docutils]

huidongc (approved):

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -93,13 +93,18 @@ def render(self, render_data: Any) -> None:
         pass
 
     @abstractmethod
-    def write_output(self, render_data: Any, output_name: str, output_data: torch.Tensor) -> None:
-        """Write a specific output type to the given buffer.
+    def read_output(self, render_data: Any, output_name: str, camera_data: Any) -> None:
```
bareya: `camera_data` has a very specific type `CameraData` and it's defined in `camera_data.py`, so I think `Any` is incorrect here.

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -93,13 +93,18 @@ def render(self, render_data: Any) -> None:
...
+        """Read a named output from the renderer into the camera data container.
+
+        Implementations should populate ``camera_data.output[output_name]`` with
+        rendered data and, when available, ``camera_data.info[output_name]`` with
```
bareya: Those are implementation details I would leave it as vague as it is in the first line.

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -93,13 +93,18 @@ def render(self, render_data: Any) -> None:
...
+        annotator metadata (e.g. ``idToLabels`` for segmentation types).
 
         Args:
             render_data: The render data object from :meth:`create_render_data`.
             output_name: Name of the output (e.g. ``"rgba"``, ``"depth"``).
-            output_data: Pre-allocated tensor to write the output into.
+            camera_data: The :class:`~isaaclab.sensors.camera.camera_data.CameraData`
```
bareya, line 102: Interesting you quoted here actual type but not in `read_output`

source/isaaclab/isaaclab/sensors/camera/camera_data.py
bareya: question so this is unrelated to changing this from list to dictionary? Because I am reading here:

https://docs.python.org/3.11/library/stdtypes.html#dict.values

> Dictionaries preserve insertion order.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -476,135 +441,31 @@ def _initialize_impl(self):
...
-                        0.0 if self.cfg.depth_clipping_behavior == "zero" else self.cfg.spawn.clipping_range[1]
-                    )
+
+        self._renderer.update_transforms()
+        self._renderer.render(self._render_data)
+
+        for output_name in self._data.output:
+            if output_name == "rgb":
+                continue
+            self._renderer.read_output(self._render_data, output_name, self._data)
```
bareya: Why we can not simplify this, and move entire loop and just call:

```python
 self._renderer.read_output(self._render_data, self._data)
```

and renderer will do the job copying/moving everything what's necessary back to camera data?

bareya (changes requested):

source/isaaclab/isaaclab/sensors/camera/camera.py
nvsekkin: yeah we can fold in the output name

source/isaaclab/isaaclab/renderers/base_renderer.py
nvsekkin: fixed

docs/source/how-to/save_camera_output.rst
```diff
@@ -32,18 +32,11 @@ images in a numpy format. For more information on the basic writer, please check
 
 While stepping the simulator, the images can be saved to the defined folder. Since the BasicWriter only supports
 saving data using NumPy format, we first need to convert the PyTorch sensors to NumPy arrays before packing
-them in a dictionary.
+them in a dictionary and writing with the BasicWriter.
```
bareya, line 39: Is that true or is that very specific to implementation - Isaac Sim RTX Renderer?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -387,18 +388,21 @@ def update_camera(
...
-            return
-        src = render_data.warp_buffers[output_name]
-        if src.ptr != output_data.data_ptr():
-            wp.copy(dest=wp.from_torch(output_data), src=src)
+        """Copy from render_data warp buffers to camera data output tensors."""
+        for output_name in camera_data.output:
+            if output_name == "rgb":
+                continue
+            src = render_data.warp_buffers.get(output_name)
+            if src is None:
```
bareya, line 401: We shouldn't be silently discarding outputs that were requested but renderer can not produce.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -280,10 +327,12 @@ def tiling_grid_shape():
...
-    def write_output(self, render_data: IsaacRtxRenderData, output_name: str, output_data: torch.Tensor):
-        """No-op for Isaac RTX - all outputs written in render().
-        See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.write_output`."""
-        pass
+    def read_output(self, render_data: IsaacRtxRenderData, camera_data: CameraData) -> None:
+        """Populate per-output metadata collected during render(). Pixel data already written in render().
+        See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.read_output`."""
+        for output_name, info in render_data.renderer_info.items():
+            if info is not None:
+                camera_data.info[output_name] = info
```
bareya, line 335: That looks much better!

bareya (approved):

source/isaaclab/isaaclab/sensors/camera/camera_data.py
nvsekkin: Removed the note. The old Camera preserved data_types order (via _rep_registry which was keyed by data_types), but the old TiledCamera used the same hardcoded if-block pattern and never guaranteed order. 

This is a behavioral change for old Camera users: they could previously iterate output.keys() and get data_types order, and now they can't. I wanted to document this for legacy Camera users, but this note is confusing. all access should be by key name and not ordered IMO.

docs/source/how-to/save_camera_output.rst
nvsekkin: Good catch - Camera is now renderer-agnostic so this section is RTX-specific, not universal. Added a note clarifying. Tutorial was specific to previous Camera.

PR #5176, opened 2026-04-05
Fixes NaN after env reset by syncing body_q before collision detection

ooctipus:
# Description

This is an important fix for locomotion rough terrain in newton to work

- Fixed NaN explosions during Newton training caused by stale body_q after env resets. Articulation write methods (write_root_pose_to_sim_*, write_joint_position_to_sim_*) updated joint_q but never recomputed body_q via forward kinematics. The collision pipeline read the old body_q, generating contacts for the wrong position. The MuJoCo Warp solver then computed constraint Jacobians from new qpos but old contacts — a geometry mismatch that overflowed the solver gradient, producing NaN.
- Added NewtonManager.invalidate_fk() flag set by articulation write methods, checked in step() before collision detection to conditionally run eval_fk.
- Added regression test that patches _simulate_physics_only to capture body_q at collide() time and asserts consistency with joint_q. Fails with 10m mismatch without the fix.

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_newton/config/extension.toml
+++ b/source/isaaclab_newton/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.5.9"
+version = "0.5.10"
 
 # Description
 title = "Newton simulation interfaces for IsaacLab core package"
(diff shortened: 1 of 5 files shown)


kellyguo11 (approved):

AntoineRichard (approved):
I think this is the right way of doing this. We could make the eval fk specific to a given robot but it adds overhead, and I'm not sure we'd be gaining much perf in doing so.
LGTM! Thanks Octi!

PR #5177, opened 2026-04-06
Simplify preset resolution with single-pass resolve_presets

ooctipus:
# Description

- Add `resolve_presets(cfg, selected)` — a single-pass function that walks the config tree and replaces each `PresetCfg` with the CLI-selected or default alternative. This fixes bugs where presets inside dict-valued alternatives and `PresetCfg(default=None)` were not being discovered or resolved, causing deeply nested configs to silently use wrong defaults. 

- Unknown or unresolvable preset names now raise `ValueError` with a grouped listing of available presets and the paths they affect.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
@@ -38,8 +38,6 @@ except ImportError:
 from isaaclab.envs.utils.spaces import replace_env_cfg_spaces_with_strings, replace_strings_with_env_cfg_spaces
 from isaaclab.utils import configclass, replace_slices_with_strings, replace_strings_with_slices
 
-from isaaclab_tasks.utils.parse_cfg import load_cfg_from_registry
-
 
 @configclass
 class PresetCfg:
@@ -152,6 +150,10 @@ def collect_presets(cfg, path: str = "") -> dict:
         for alt in fields.values():
             if hasattr(alt, "__dataclass_fields__"):
                 result.update(collect_presets(alt, preset_path))
+            elif isinstance(alt, dict):
+                for v in alt.values():
+                    if hasattr(v, "__dataclass_fields__"):
+                        result.update(collect_presets(v, preset_path))
 
     if isinstance(cfg, PresetCfg):
         _record(cfg, path)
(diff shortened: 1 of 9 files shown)


source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -32,14 +32,21 @@
     import hydra
     from hydra.core.config_store import ConfigStore
     from omegaconf import OmegaConf
+
+    _HYDRA_AVAILABLE = True
 except ImportError:
-    raise ImportError("Hydra not installed. Run: pip install hydra-core")
```
AntoineRichard, line 36: Do we want to support environments without HYDRA? At this point HYDRA is so important to lab that I don't see why we would bother supporting non-HYDRA environments.

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -161,6 +163,74 @@ def _record(preset_obj, preset_path):
...
+        replacement = _pick_alternative(cfg, selected)
+        if replacement is cfg:
+            return cfg
+        return resolve_presets(replacement, selected)
+
+    def _resolve(parent, key, preset_obj, _path):
+        val = _pick_alternative(preset_obj, selected)
+        if val is preset_obj:
+            return
+        while isinstance(val, PresetCfg):
```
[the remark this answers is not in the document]
AntoineRichard: @ooctipus the bot looks correct? Maybe this would never happen though

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
AntoineRichard, line 1: I feel like the hydra resolution is getting out of hand a bit

AntoineRichard (changes requested):
I would look into the bot's feedback, otherwise it looks like it's doing what it should. Though I feel like the hydra thing is starting to be heavy.

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
ooctipus: I think there were a bunch of patch fixes that complicated things a bit, this PR supposes to clean them up, In my opinion this PR made hydra logic a lot nicer than it was used to!

ooctipus: I address the finite cycle issue and added a test case for it :D

AntoineRichard (approved):

PR #5179, opened 2026-04-06
Refactors Newton XformPrimView: proper local poses, warp-native API, and shared contract tests

ooctipus:
# Description

### Summary

Rewrites the Newton `XformPrimView` from scratch with correct local-pose semantics, a clean site-based architecture, and a shared test contract that enforces the same invariants across all backends (USD, Fabric, Newton).

**Key changes:**

- **Fix local poses**: `get_local_poses` / `set_local_poses` now correctly compute parent-relative transforms on GPU (`inv(parent_world) * prim_world`) instead of incorrectly returning world poses
- **Fix set_world_poses**: Updates `site_local` offset instead of writing `body_q` directly (which would move the parent body)
- **Guard against misuse**: Raises `ValueError` if prim path resolves to a physics body or collision shape — XformPrimView is for non-physics child prims only (cameras, sensors, markers)
- **Warp-native API**: All inputs/outputs are `wp.array` — no torch/list conversion overhead
- **Factory dispatch**: `from isaaclab.sim.views import XformPrimView` now auto-selects the correct backend (USD, Fabric, Newton) via `XformPrimViewFactory`
- **Composition over inheritance**: PhysX `FabricXformPrimView` uses composition (`self._usd_view`) instead of inheriting from `UsdXformPrimView`
- **Explicit class names**: `UsdXformPrimView`, `FabricXformPrimView`, `NewtonSiteXformPrimView` — no more ambiguous `XformPrimView` in every package
- **Shared contract tests**: 16 test functions in `xform_contract_tests.py` that any backend imports and runs by providing a `view_factory` fixture
- **Benchmark updates**: Both benchmark scripts support Newton, use warp-native arrays, and include per-backend round-trip verification

### Type of change

- [x] Bug fix (Newton local poses were fundamentally broken — `local == world`)
- [x] New feature (shared contract test infrastructure, factory dispatch)
- [x] Breaking change (Newton `XformPrimView` renamed to `NewtonSiteXformPrimView`, PhysX to `FabricXformPrimView`; indices parameter changed from `Sequence[int]` to `wp.array`)
- [x] Documentation update

### Expected failures

- `test_set_world_updates_local[cuda:0]` in Fabric — pre-existing limitation: `set_world_poses` writes to `omni:fabric:worldMatrix` but `get_local_poses` reads from USD, so local poses are stale after a Fabric world write. This will be fixed by the Fabric backend PR (#4923) which adds `omni:fabric:localMatrix` support.

### Test results

| Backend | Passed | Failed | Skipped |
|---|---|---|---|
| Newton | 40 | 0 | 0 |
| USD | 45 | 0 | 0 |
| Fabric | 15 | 1 (xfail) | 16 (CPU) |
| Camera | 20 | 0 | 0 |
| TiledCamera | 61 | 0 | 0 |
| RayCaster | 5 | 0 | 0 |

### Benchmark (1024 prims, 50 iterations, RTX 5090)

```
========================================================================================================================
BENCHMARK RESULTS: 1024 prims, 50 iterations
========================================================================================================================
Operation                         Isaaclab Usd (ms)   Isaaclab Fabric (ms) Isaaclab Newton Site (ms)
------------------------------------------------------------------------------------------------------------------------
Initialization                               3.7168                 3.6596                39.0608
Get World Poses                              6.6730                 0.0296                 0.0180
Set World Poses                             15.5574                 0.0640                 0.0186
Get Local Poses                              4.6086                 4.5637                 0.0216
Set Local Poses                              6.4680                 6.6221                 0.0218
Get Both (World+Local)                      12.1240                 4.7361                 0.0374
Interleaved World Set->Get                  23.4141                 0.1050                 0.0344
========================================================================================================================

Total                                       72.5619                19.7800                39.2126

========================================================================================================================
```

### Checklist

- [ ] I have run the pre-commit checks with `./isaaclab.sh --format`


source/isaaclab/isaaclab/sim/views/xform_prim_view.py
```diff
@@ -206,394 +120,93 @@ def prims(self) -> list[Usd.Prim]:
...
         Args:
-            indices: Indices of prims to get scales for. Defaults to None, in which case scales are retrieved
-                for all prims in the view.
-
-        Returns:
-            A tensor of shape (M, 3) containing the scales of each prim, where M is the number of prims queried.
+            positions: World-space positions of shape ``(M, 3)``.
+            orientations: World-space quaternions ``(w, x, y, z)`` of shape ``(M, 4)``.
+            indices: Indices of prims to set poses for. Defaults to None (all prims).
         """
```
kellyguo11: 🔴 **Breaking change: callers not updated**

This method now returns `wp.array` instead of `torch.Tensor`. However, multiple core callers were **not** updated:

1. **`camera.py:695-696`**: `poses, quat = self._view.get_world_poses(env_ids)` followed by `self._data.pos_w[env_ids] = poses` — assigning `wp.array` into a `torch.Tensor` slice will crash.

2. **`ray_cast_utils.py:38`**: `pos_w, quat_w = physx_view.get_world_poses(env_ids)` — returned to callers that do `pos_w += self.drift[env_ids]` (torch arithmetic on `wp.array` will fail).

3. **`benchmark_view_comparison.py`**: Uses return values as torch tensors for timing comparisons.

These files must either:
- Be updated to wrap returns with `wp.to_torch()`, or
- The base class should return `torch.Tensor` and let Newton/PhysX return `wp.array` only when the caller opts in.

This is a runtime crash, not a silent error — it will be caught immediately, but it blocks any existing workflow that uses cameras or ray casters.

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
```diff
@@ -0,0 +1,495 @@
...
+
+    inv_local = wp.transform_inverse(site_local[si])
+    w_pos = world_pos[i]
+    w_q = world_quat[i]
+    world_tf = wp.transform(w_pos, wp.quatf(w_q[0], w_q[1], w_q[2], w_q[3]))
+    body_q[bid] = wp.transform_multiply(world_tf, inv_local)
+
+
+@wp.kernel
+def _gather_scales(
```
kellyguo11, line 102: 🟡 **Performance: O(sites × total_shapes) kernel**

This kernel performs a linear scan over **all** shapes in the model for **every** site:

```python
for s in range(num_shapes):
    if shape_body[s] == bid and found == 0:
```

For a typical scene with 4096 envs × ~10 shapes/body = ~40K shapes, this is:
- 4096 sites × 40K shapes = **163 million** comparisons per `get_scales()` call

This is a GPU kernel so each thread does O(num_shapes) work, but it's still a significant serial loop inside each warp thread. For larger scenes, this will dominate.

**Suggestion:** Pre-compute a `site_to_shape` index mapping at init time (a simple `wp.array` mapping each site index to its first shape index). This converts the kernel from O(num_shapes) per thread to O(1).

```python
# At init time:
site_shape = wp.array([...], dtype=wp.int32, device=device)  # -1 if no shape

# In kernel:
si = site_shape[i]  # O(1) lookup
if si >= 0:
    out_scales[i] = shape_scale[si]
```

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_xform_prim_view.py
```diff
@@ -0,0 +1,495 @@
...
+                dim=self.count,
+                inputs=[state.body_q, self._site_body, self._site_local, pos_wp, quat_wp],
+                device=self._device,
+            )
+
+    # ------------------------------------------------------------------
+    # Local poses -- delegate to world (Newton bodies live in world space)
+    # ------------------------------------------------------------------
+
+    def get_local_poses(self, indices: Sequence[int] | None = None) -> tuple[wp.array, wp.array]:
```
kellyguo11: 🟡 **Semantic concern: `get_local_poses` returns world poses**

For Newton, `get_local_poses()` delegates directly to `get_world_poses()`. While documented ("Newton bodies live in world space"), this is semantically wrong for prims with non-root parents.

Consider a camera mount at `(0.1, 0, 0.05)` relative to a rigid body. The USD `XformPrimView` would return the local offset `(0.1, 0, 0.05)` from `get_local_poses()`, but the Newton view returns the world position. Code that relies on local-pose semantics (e.g., computing relative transforms) will silently produce wrong results.

If true local poses aren't available from Newton's state, consider either:
1. Computing local poses from `world_pose * inverse(parent_world_pose)` using the site data you already have
2. Raising `NotImplementedError` so callers know the limitation
3. At minimum, adding a prominent warning in the docstring that this method does NOT return local poses for Newton

source/isaaclab_newton/isaaclab_newton/sim/views/xform_prim_view.py
```diff
@@ -0,0 +1,495 @@
...
+        site_bodies: list[int] = []
+        site_locals: list[list[float]] = []
+
+        identity_xform = [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0]
+
+        for prim in self._prims:
+            pp = prim.GetPath().pathString
+
+            if pp in shape_labels:
+                si = shape_labels.index(pp)
```
kellyguo11: 🔵 **Performance: O(N) `list.index()` in init loop**

For each matched prim, `shape_labels.index(pp)` does a linear scan of the entire `shape_labels` list. With 4096 envs × ~10 shapes = 40K labels, and 4096 prims, this is ~164M string comparisons.

Same issue on lines 295 and 334 with `body_labels.index()`.

```suggestion
                si = shape_label_to_idx[pp]
```

Where `shape_label_to_idx = {label: i for i, label in enumerate(shape_labels)}` is precomputed once. Similarly for body labels. You already have `body_label_set` as a set for O(1) membership test — just make it a dict mapping to index.

source/isaaclab/isaaclab/sim/views/xform_prim_view_factory.py
```diff
@@ -0,0 +1,44 @@
...
+"""Factory for creating backend-specific XformPrimView instances."""
+
+from __future__ import annotations
+
+from isaaclab.utils.backend_utils import FactoryBase
+
+from .base_xform_prim_view import BaseXformPrimView
+
+
+class XformPrimViewFactory(FactoryBase, BaseXformPrimView):
```
kellyguo11: 🟡 **No tests for the factory**

The `XformPrimViewFactory` is the core architectural addition that ties the backends together, but there are zero tests for it. Missing test cases:

1. Factory returns `XformPrimView` (USD) when no SimulationContext exists
2. Factory returns PhysX `XformPrimView` when PhysX backend is active
3. Factory returns Newton `XformPrimView` when Newton backend is active
4. Factory `isinstance` check: `isinstance(factory_result, BaseXformPrimView)` returns True

Without these, any regression in the factory dispatch logic (e.g., wrong module path, missing registration) will go undetected.

source/isaaclab/isaaclab/sim/views/base_xform_prim_view.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+
+    @property
+    @abc.abstractmethod
+    def count(self) -> int:
+        """Number of prims in this view."""
+        ...
+
+    @abc.abstractmethod
+    def get_world_poses(self, indices: Sequence[int] | None = None) -> tuple[wp.array, wp.array]:
```
kellyguo11: 🟡 **Inconsistent `wp.array` dtypes between backends**

The ABC declares returns as `tuple[wp.array, wp.array]` but doesn't specify the expected dtype or shape convention. The implementations diverge:

- **USD/PhysX**: Returns `wp.array(dtype=wp.float32, shape=(M, 3))` and `wp.array(dtype=wp.float32, shape=(M, 4))`
- **Newton**: Returns `wp.array(dtype=wp.vec3f, shape=(M,))` and `wp.array(dtype=wp.vec4f, shape=(M,))`

Both convert to the same `torch.Tensor` shape via `wp.to_torch()`, but any code inspecting the raw `wp.array.shape` or `wp.array.dtype` will get different results depending on the backend. This violates the Liskov Substitution Principle for the ABC.

Consider documenting the expected dtype/shape in the docstring (or better, standardizing on one convention). The `wp.vec3f`/`wp.vec4f` approach is more semantically correct and avoids ambiguity.

source/isaaclab/isaaclab/sim/views/xform_prim_view.py
```diff
@@ -206,394 +120,93 @@ def prims(self) -> list[Usd.Prim]:
...
 
     def set_world_poses(
         self,
-        positions: torch.Tensor | None = None,
-        orientations: torch.Tensor | None = None,
+        positions: wp.array | None = None,
+        orientations: wp.array | None = None,
         indices: Sequence[int] | None = None,
     ):
         """Set world-space poses for prims in the view.
```
kellyguo11: 🔵 **Type hints say `wp.array` but callers pass `torch.Tensor`**

The setters are annotated as `positions: wp.array | None` but the `_to_numpy` helper (line 301) explicitly handles both `wp.array` and `torch.Tensor`. All existing tests also pass `torch.Tensor` to the setters.

The type hints should reflect what the code actually accepts:

```suggestion
        positions: wp.array | torch.Tensor | None = None,
        orientations: wp.array | torch.Tensor | None = None,
```

Or at minimum, document that `torch.Tensor` is accepted as an input type. Without this, type checkers will flag every existing caller as an error.

kellyguo11 (commented):
## 🤖 Isaac Lab Review Bot

### Summary
This PR refactors `XformPrimView` from a monolithic class into a backend-polymorphic hierarchy: `BaseXformPrimView` (ABC) → `XformPrimView` (USD-only, in `isaaclab`), `XformPrimView` (PhysX/Fabric, in `isaaclab_physx`), `XformPrimView` (Newton/GPU, in `isaaclab_newton`), connected by `XformPrimViewFactory`. The API changes from `torch.Tensor` to `wp.array` for all getters/setters.

### Architecture Impact
**High.** The return type change from `torch.Tensor` → `wp.array` on all getters is a **breaking change** that affects every downstream consumer of `XformPrimView`. Multiple files in the core `isaaclab` package (camera, ray_caster, benchmarks) directly consume the return values of `get_world_poses()` as `torch.Tensor` and were **NOT updated** in this PR. This will cause runtime crashes.

### Implementation Verdict
The refactoring direction (backend-polymorphic views, warp-native data) is sound architecture. However, the PR is **incomplete as submitted**: it changes the return types of a widely-consumed API without updating all callers. The Newton and PhysX implementations are well-structured, but there are several issues ranging from critical (broken callers) to moderate (performance, inconsistent dtypes).

### Test Coverage
- Core USD tests: ✅ Updated to convert via `wp.to_torch()`
- PhysX Fabric tests: ✅ New dedicated test file
- Newton tests: ✅ New comprehensive test file
- **Missing**: No tests for `XformPrimViewFactory` (the whole point of the abstraction)
- **Missing**: Camera / ray_caster integration tests would have caught the broken callers

### CI Status
All checks **pending** at time of review.

### Findings

| # | Severity | File | Finding |
|---|----------|------|---------|
| 1 | 🔴 | `ray_cast_utils.py` (not in diff) | `obtain_world_pose_from_view()` returns `wp.array` to callers expecting `torch.Tensor` — crashes ray_caster |
| 2 | 🔴 | `camera.py` (not in diff) | `_update_poses()` assigns `wp.array` into `torch.Tensor` slice — crashes camera |
| 3 | 🔴 | `benchmark_view_comparison.py` (not in diff) | Uses `get_world_poses()` return as torch tensors for arithmetic |
| 4 | 🟡 | `newton/xform_prim_view.py` | `_gather_scales` kernel is O(sites × total_shapes) — quadratic blowup |
| 5 | 🟡 | `newton/xform_prim_view.py` | `get_local_poses` = `get_world_poses` — semantically wrong for hierarchical prims |
| 6 | 🟡 | `newton/xform_prim_view.py` | Inconsistent `wp.array` dtype: Newton uses `wp.vec3f`/`wp.vec4f`, USD/PhysX use `wp.float32` |
| 7 | 🟡 | `xform_prim_view_factory.py` | No `XformPrimViewFactory` tests |
| 8 | 🔵 | `newton/xform_prim_view.py` | Init uses O(N) `list.index()` instead of O(1) dict lookup |
| 9 | 🔵 | `xform_prim_view.py` | Type hints say `wp.array` but tests and callers pass `torch.Tensor` to setters |

See inline comments for details on findings 4–9.

source/isaaclab_physx/isaaclab_physx/sim/views/xform_prim_view.py
```diff
@@ -0,0 +1,380 @@
...
+            super().set_world_poses(positions, orientations, indices)
+            return
+
+        if not self._fabric_initialized:
+            self._initialize_fabric()
+
+        indices_wp = self._resolve_indices_wp(indices)
+        count = indices_wp.shape[0]
+
+        positions_wp = positions if positions is not None else wp.zeros((0, 3), dtype=wp.float32, device=self._device)
```
kellyguo11: 🔵 **Fabric path doesn't explicitly convert `torch.Tensor` → `wp.array`**

The old code explicitly called `wp.from_torch(positions)` before passing to `wp.launch()`. The new code passes the input directly:

```python
positions_wp = positions if positions is not None else wp.zeros(...)
```

This works because warp's `pack_arg` auto-converts via `__cuda_array_interface__` (CUDA) or `__array__()` (CPU). However, it's fragile and undocumented — relying on warp internal behavior rather than explicit conversion.

The old pattern was more robust:
```python
positions_wp = wp.from_torch(positions) if isinstance(positions, torch.Tensor) else positions
```

Not a bug today, but a reliability concern if warp changes its auto-conversion behavior.

source/isaaclab_physx/isaaclab_physx/sim/views/xform_prim_view.py
```diff
@@ -0,0 +1,380 @@
...
+
+import isaaclab.sim as sim_utils
+from isaaclab.app.settings_manager import SettingsManager
+from isaaclab.sim.views.xform_prim_view import XformPrimView as UsdXformPrimView
+from isaaclab.utils.warp import fabric as fabric_utils
+
+logger = logging.getLogger(__name__)
+
+
+class XformPrimView(UsdXformPrimView):
```
bareya: I think this is kind of inconsistent with what you have designed and I would consider inheritance as anti-pattern in this situation. 

Whatever someone does to `UsdXformPrimView` in the future, it's going to be exposed as public API. I would suggest to derive from interface you designed - `BaseXFormPrimView`, and do composition.

If you look at #4923 MR I designed xform data backend through composition - this gives us additional benefits - we can decouple, xform interface, backend and caching, and we can keep adding those layers. We can also test them separately.

I thought we can have Usd, Newton, Fabric, UsdRt backends, that are wrapped in caching layer that can be inserted into `XFormPrimView`.

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_xform_prim_view.py
bareya: May I ask why are we doing this in the first place? This seems like something that can lead to many many errors.

I agree with @kellyguo11

source/isaaclab_physx/isaaclab_physx/sim/views/xform_prim_view.py
bareya: Also I would rename `UsdXformPrimView` alias into something more descriptive, `IssacSimXformPrivmView`?

ooctipus: addressed

source/isaaclab/isaaclab/sensors/ray_caster/ray_cast_utils.py
```diff
@@ -34,8 +34,10 @@ def obtain_world_pose_from_view(
     Raises:
         NotImplementedError: If the prim view is not of the supported type.
     """
-    if isinstance(physx_view, XformPrimView):
-        pos_w, quat_w = physx_view.get_world_poses(env_ids)
+    if isinstance(physx_view, BaseXformPrimView):
```
bareya: Not that this needs to be addressed in this MR but there is elegant way of wrapping those multiple definitions into single dispatch, example:

```python
from functools import singledispatch

@singledispatch
def _world_poses(view, env_ids):
    raise NotImplementedError(type(view))

@_world_poses.register
def _(view: BaseXformPrimView, env_ids):
    ...


def obtain_world_pose_from_view(physx_view, env_ids, clone=False):
    return _world_poses(physx_view, env_ids)
```

source/isaaclab/isaaclab/sensors/ray_caster/ray_cast_utils.py
```diff
@@ -12,11 +12,11 @@
 
 import omni.physics.tensors.impl.api as physx
 
-from isaaclab.sim.views import XformPrimView
+from isaaclab.sim.views import BaseXformPrimView
 
 
 def obtain_world_pose_from_view(
-    physx_view: XformPrimView | physx.ArticulationView | physx.RigidBodyView,
+    physx_view: BaseXformPrimView | physx.ArticulationView | physx.RigidBodyView,
```
bareya: Question, it seems like all of those are _transformables_. Why do we do this kind of _"ugly"_ dispatch and not wrap this into something that is called transformable, xform view or whatever? They all seem very similar.  However listing those types here brings some complexity?

source/isaaclab/isaaclab/sensors/ray_caster/ray_caster.py
bareya: Return type is incorrect? I think this pattern of prim view articulation and rbd transforms repeats.

source/isaaclab/isaaclab/sim/spawners/sensors/sensors_cfg.py
```diff
@@ -222,3 +222,16 @@ class FisheyeCameraCfg(PinholeCameraCfg):
 
     fisheye_polynomial_f: float = 0.0
     """Sixth component of fisheye polynomial. Defaults to 0.0."""
+
+
+@configclass
+class RayCasterXformCfg(SpawnerCfg):
```
Mayankm96: I’m wondering about the choice of the name `RaycasterXformCfg`: is there a specific reason for that? Since this prim doesn’t seem to carry any operations or schemas itself, would it be clearer to represent it as a `DummyPrim` (or something similar) to better reflect its role?

Also, it feels a bit unintuitive that this prim is created, but the sensor offset is then defined separately in the raycaster configuration. Would it make sense to consolidate that, or is there a design reason for keeping those concerns separate?

source/isaaclab/isaaclab/sensors/ray_caster/multi_mesh_ray_caster.py
```diff
@@ -14,10 +14,8 @@
 import trimesh
 import warp as wp
 
-import omni.physics.tensors.impl.api as physx
-
 import isaaclab.sim as sim_utils
-from isaaclab.sim.views import XformPrimView
+from isaaclab.sim.views import BaseXformPrimView
```
Mayankm96: We should probably stick to a consistent naming convention like `<Type>Base`, as we already do with `AssetBase` and `SensorBase`. Would it make sense to align this with that pattern?

scripts/demos/sensors/raycaster_sensor.py
```diff
@@ -62,7 +62,8 @@ class RaycasterSensorSceneCfg(InteractiveSceneCfg):
     robot = ANYMAL_C_CFG.replace(prim_path="{ENV_REGEX_NS}/Robot")
 
     ray_caster = RayCasterCfg(
-        prim_path="{ENV_REGEX_NS}/Robot/base/lidar_cage",
+        prim_path="{ENV_REGEX_NS}/Robot/base/raycaster",
+        spawn=sim_utils.RayCasterXformCfg(),
```
kellyguo11: oof, this would also be a pretty big breaking change for existing user code that's using the ray caster sensor right?

source/isaaclab/isaaclab/sensors/ray_caster/ray_cast_utils.py
```diff
@@ -30,18 +28,11 @@ def obtain_world_pose_from_view(
...
     """
-    if isinstance(physx_view, XformPrimView):
-        pos_w, quat_w = physx_view.get_world_poses(env_ids)
-    elif isinstance(physx_view, physx.ArticulationView):
-        pos_w, quat_w = wp.to_torch(physx_view.get_root_transforms())[env_ids].split([3, 4], dim=-1)
-    elif isinstance(physx_view, physx.RigidBodyView):
-        pos_w, quat_w = wp.to_torch(physx_view.get_transforms())[env_ids].split([3, 4], dim=-1)
-    else:
-        raise NotImplementedError(f"Cannot get world poses for prim view of type '{type(physx_view)}'.")
+    indices = wp.from_torch(env_ids.to(dtype=torch.int32), dtype=wp.int32) if env_ids is not None else None
```
kellyguo11: does this mean we don't allow attaching ray caster sensor to a physics body and assume it's always attached to a non-physics body?

source/isaaclab/isaaclab/sensors/ray_caster/ray_caster.py
```diff
@@ -332,77 +326,67 @@ def _debug_vis_callback(self, event):
...
+        Before this refactor, ``prim_path`` was expected to point at an existing link prim (e.g.
+        ``{ENV_REGEX_NS}/Robot/base``) that typically carried an ``ArticulationRootAPI`` or
+        ``RigidBodyAPI``.  The new convention requires ``prim_path`` to be a **non-physics** child
+        (e.g. ``{ENV_REGEX_NS}/Robot/base/raycaster``) where a plain Xform is spawned.
+
+        This helper preserves backward compatibility: when ``spawn`` is set and the resolved
+        path already exists as a prim with a physics API, the path is automatically extended
+        with ``/raycaster`` and a deprecation warning is emitted.
+        """
+        if cfg.spawn is None:
```
kellyguo11: was spawn needed in previous ray caster configurations as well? or users will need to add this to their previous setup?

source/isaaclab/isaaclab/sim/views/usd_frame_view.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+        self._prim_path = prim_path
+        self._device = device
+
+        stage = sim_utils.get_current_stage() if stage is None else stage
+        self._prims: list[Usd.Prim] = sim_utils.find_matching_prims(prim_path, stage=stage)
+
+        if validate_xform_ops:
+            for prim in self._prims:
+                sim_utils.standardize_xform_ops(prim)
```
kellyguo11, line 89: how long does this call take for a typical RL environment workflow? is there benefit in defaulting validate_xform_ops to False instead?

kellyguo11: actually on second thought, accuracy is probably more important than performance, but maybe worth documenting somewhere if the overhead is large.

source/isaaclab/test/sim/frame_view_contract_utils.py
```diff
@@ -0,0 +1,359 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
kellyguo11, line 1: is this meant to be unit tests? why does the file not start with test_*?

source/isaaclab_mimic/isaaclab_mimic/locomanipulation_sdg/scene_utils.py
```diff
@@ -122,7 +124,7 @@ def set_pose(self, pose: torch.Tensor):
         xform_prim = self._get_xform_view()
         position = pose[..., :3]
         orientation = pose[..., 3:]
-        xform_prim.set_world_poses(position, orientation, None)
+        xform_prim.set_world_poses(wp.from_torch(position.contiguous()), wp.from_torch(orientation.contiguous()), None)
```
kellyguo11, line 127: @AntoineRichard - do you think the `.contiguous()` call could also be something we can handle for users under the hood somehow? I'm not sure how easy it'd be for users who are unfamiliar with Warp to know this

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
```diff
@@ -0,0 +1,747 @@
...
+    indices: wp.array(dtype=wp.int32),
+    num_shapes: wp.int32,
+    out_scales: wp.array(dtype=wp.vec3f),
+):
+    """Indexed variant of _gather_scales."""
+    i = wp.tid()
+    si = indices[i]
+    bid = site_body[si]
+    found = int(0)
+    for s in range(num_shapes):
```
kellyguo11, line 156: is it worth having this as a kernel dimension as well?

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -0,0 +1,403 @@
...
+
+    def get_local_poses(self, indices=None):
+        return self._usd_view.get_local_poses(indices)
+
+    # ------------------------------------------------------------------
+    # Scales — Fabric-accelerated or USD fallback
+    # ------------------------------------------------------------------
+
+    def set_scales(self, scales, indices=None):
+        if not self._use_fabric:
```
kellyguo11, line 224: what does it mean to be in the FabricXformPrimView class but not having use_fabric enabled? should this path be just using the USD view class instead?

source/isaaclab/isaaclab/sensors/ray_caster/ray_cast_utils.py
ooctipus: yes

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
ooctipus: there are buggy cases where fabric doesn't work, and @bareya has fixes, self._use_fabric could be good WAR for buggy case handling for the time being before @bareya fixes land.

source/isaaclab/isaaclab/sim/spawners/sensors/sensors_cfg.py
ooctipus: yes the offset should be consolidate to frameview instead of separately maintained, much like the current camera class

source/isaaclab/isaaclab/sensors/ray_caster/ray_caster.py
ooctipus: its default added, user don't need to add this to their previous setup

source/isaaclab_physx/isaaclab_physx/sim/__init__.pyi
kellyguo11, line 27:
```suggestion
__all__ = [
    "define_deformable_body_properties",
    "modify_deformable_body_properties",
    "DeformableBodyPropertiesCfg",
    "DeformableObjectSpawnerCfg",
    "spawn_deformable_body_material",
    "DeformableBodyMaterialCfg",
    "SurfaceDeformableBodyMaterialCfg",
    "views",
]

from .schemas import (
    define_deformable_body_properties,
    modify_deformable_body_properties,
    DeformableBodyPropertiesCfg
)
from .spawners import (
    DeformableObjectSpawnerCfg,
    spawn_deformable_body_material,
    DeformableBodyMaterialCfg,
    SurfaceDeformableBodyMaterialCfg,
)
from . import views
```

source/isaaclab_physx/isaaclab_physx/sim/__init__.pyi
```diff
@@ -3,6 +3,30 @@
...
+from .schemas import (
+    define_deformable_body_properties,
+    modify_deformable_body_properties,
+    DeformableBodyPropertiesCfg
+)
+from .spawners import (
+    DeformableObjectSpawnerCfg,
+    spawn_deformable_body_material,
+    DeformableBodyMaterialCfg,
+    SurfaceDeformableBodyMaterialCfg,
```
kellyguo11:
```suggestion

```

docs/source/api/lab/isaaclab.sim.views.rst
```diff
@@ -7,11 +7,11 @@
 
   .. autosummary::
 
-    XformPrimView
+    FrameView
```
kellyguo11, line 10: let's make sure to document this in the migration guide

kellyguo11 (approved):

kellyguo11:
doc build failure:
`/home/runner/work/IsaacLab/IsaacLab/source/isaaclab/isaaclab/sensors/__init__.py:docstring of isaaclab.sensors:17: ERROR: Malformed table.`

PR #5188, opened 2026-04-06
Adds release version to benchmark output

kellyguo11:
## Description

Adds the Isaac Lab release version (from the root `VERSION` file) to the benchmark output files.

## Motivation

The `VersionInfoRecorder` currently collects package versions via `importlib.metadata` (pip) and module `__version__` attributes. These reflect what is *installed* in the current environment, but do not capture the **release version** of Isaac Lab itself (e.g. `3.0.0` from the root `VERSION` file).

This makes it hard to correlate benchmark results with specific releases, especially when the installed pip version does not match the source tree (e.g. editable installs during development).

## Changes

- `source/isaaclab/isaaclab/test/benchmark/recorders/record_version_info.py`:
  - Added reading of the top-level `VERSION` file → recorded as `isaaclab_release`

## Example Output

Before (existing fields preserved):
```
isaaclab: 4.5.25
isaaclab_tasks: 1.5.15
```

After (new field added):
```
isaaclab_release: 3.0.0        # ← from VERSION file
```

## Testing

Verified the recorder produces correct output with the expected field populated. No existing fields are modified — this is purely additive.

### DIFF
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+4.5.27 (2026-04-08)
+~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+
+* Added release version and extension.toml versions to
+  :class:`~isaaclab.test.benchmark.recorders.VersionInfoRecorder` output.
+
+
 4.5.26 (2026-04-08)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -7,7 +7,7 @@ Changelog
 Added
 ^^^^^
 
-* Added release version and extension.toml versions to
+* Added release version to
   :class:`~isaaclab.test.benchmark.recorders.VersionInfoRecorder` output.
 
 
--- a/source/isaaclab/isaaclab/test/benchmark/recorders/record_version_info.py
+++ b/source/isaaclab/isaaclab/test/benchmark/recorders/record_version_info.py
@@ -7,15 +7,12 @@ import importlib.metadata
 import os
 import subprocess
 
-try:
-    import tomllib
-except ModuleNotFoundError:
-    import tomli as tomllib  # Python < 3.11
+import tomllib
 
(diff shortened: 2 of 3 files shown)


source/isaaclab/isaaclab/test/benchmark/recorders/record_version_info.py
```diff
@@ -7,14 +7,23 @@
 import os
 import subprocess
 
+try:
+    import tomllib
+except ModuleNotFoundError:
+    import tomli as tomllib  # Python < 3.11
```
AntoineRichard: We should be using python3.11 at minimum. So none of this is needed.

AntoineRichard: Remove and import tomllib only.

source/isaaclab/isaaclab/test/benchmark/recorders/record_version_info.py
```diff
@@ -49,6 +58,39 @@ def _record(self, key: str, version: str | None) -> None:
...
+                toml_path = os.path.join(source_dir, entry, "config", "extension.toml")
+                if os.path.isfile(toml_path):
+                    try:
+                        with open(toml_path, "rb") as f:
+                            data = tomllib.load(f)
+                        version = data.get("package", {}).get("version")
+                        if version:
+                            self._record(f"{entry}_ext", version)
+                    except Exception:
+                        pass
```
AntoineRichard: Why not use get_version directly?

I.e.
```python
if os.path.isdir(source_dir):
            for entry in sorted(os.listdir(source_dir)):
                version = get_version(entry)
                if version:
                    self._record(f"{entry}_ext", version
```

AntoineRichard: Also could this not be done in _get_version_info directly?

AntoineRichard (dismissed):
I feel like this could be better structured.

AntoineRichard:
@greptile

source/isaaclab/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
 Changelog
 ---------
 
+4.5.27 (2026-04-08)
+~~~~~~~~~~~~~~~~~~~
+
+Added
+^^^^^
+
+* Added release version and extension.toml versions to
```
```suggestion
* Added release version to
```

PR #5202, opened 2026-04-07
Caches resolve_matching_names on AssetBase for all finder methods

hougantc-nvda:
# Description

`Articulation.find_joints` and `find_bodies` delegate to `resolve_matching_names`, which runs a Python regex double-loop on every call. Nsight profiling of the pick-place task showed this costs ~881μs per call (~1.83 ms / 1.5% of step time) for work that always returns the same result — joint and body names are fixed after construction.

This PR adds a module-level `@functools.cache` on a new private `_resolve_matching_names_impl` helper. Hashable tuples are used for the cache key; cached results are immutable tuples that the public wrapper copies into fresh lists per caller, so callers cannot mutate shared state. `resolve_matching_names_values` is left uncached because its current callers only use it during init, never in the step loop.

A `clear_resolve_matching_names_cache()` helper is called from `SimulationContext.clear_instance()` so cached entries from destroyed assets do not accumulate across scene rebuilds in long-lived processes.

Fixes # (issue)

## Type of change

## Screenshots

## Checklist


hougantc-nvda:
@greptileai

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -11,6 +11,7 @@
 import logging
 import warnings
 from collections.abc import Sequence
+from functools import cache
```
AntoineRichard: I'm wondering how much of this could be handled by a `@lru_cache` there would be less custom code.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -113,8 +114,25 @@ def __init__(self, cfg: ArticulationCfg):
...
+            return resolve_matching_names(keys, self.body_names, preserve_order)
+
+        @cache
+        def _find_joints_cached(
+            keys: tuple[str, ...], joint_names: tuple[str, ...], preserve_order: bool
+        ) -> tuple[list[int], list[str]]:
+            return resolve_matching_names(keys, list(joint_names), preserve_order)
+
+        self._find_bodies_cached = _find_bodies_cached
+        self._find_joints_cached = _find_joints_cached
```
AntoineRichard: Could this be extended to cover the tendons? Right now this looks like it's missing half the code. Also this pattern would have to be applied to the other assets (rigid object collection + rigid object(?)).

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard:
```python
  def find_bodies(self, name_keys, preserve_order=False):
      keys = (name_keys,) if isinstance(name_keys, str) else tuple(name_keys)
      indices, names = self._find_bodies_cached(keys, preserve_order)
      return list(indices), list(names)

  @lru_cache(maxsize=None)
  def _find_bodies_cached(self, keys, preserve_order):
      return resolve_matching_names(keys, self.body_names, preserve_order)
```

AntoineRichard: That would remove the need to initialize the cache.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -113,8 +114,25 @@ def __init__(self, cfg: ArticulationCfg):
...
+
```
AntoineRichard: In general I don't feel like this is implementation specific. The code looks to be the same in both backends, maybe this could be move to the abstract class? (BaseArticulation)

AntoineRichard (changes requested):
Thanks for the optimization, I think it's important some of these results get cached to avoid overhead. However, right now, this PR misses the rest of the assets (RigidObjectCollection, RigidObject) as well as some other properties in the Articulation (Tendons, SpatialTendons). Could you extend the pattern to these. Also I would consider relying on an LRU cache rather than our own custom caching. I'm not completely against the custom caching, but I'd be curious to know why a custom cache was favored over the LRU cache.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hougantc-nvda: Thanks for the suggestion. I actually ended up switching over to @cache yesterday per your recommendation as it seems it is equivalent to @lru_cache(maxsize=None) in more recent python.

```
        @cache
        def _find_joints_cached(
```

hougantc-nvda: The reason there is
```
        self._find_bodies_cached = _find_bodies_cached
        self._find_joints_cached = _find_joints_cached
```

Is so the caches get cleaned when the articulation gets destroyed. I had toyed with the idea of making the cache function "static", but it would grow unbounded. I'm worried there are use cases where developers are rotating articulations in a single run.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hougantc-nvda: Happy to make the addition to tendons. As I was only going from the nsight profiles, I tried to target the areas where hotspots were showing up. I was hesitant to make large changes myself.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hougantc-nvda: Updated code. Pushed the caching up to AssetBase. Now there is a drop in replacement for
resolve_matching_names > self._resolve_matching_names_cached.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hougantc-nvda: Updated code. Pushed the caching up to AssetBase. Now there is a drop in replacement for
resolve_matching_names > self._resolve_matching_names_cached.

RigidObject also updated to use drop in replacement.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hougantc-nvda: Updated code. Pushed the caching up to AssetBase. Now there is a drop in replacement for
resolve_matching_names > self._resolve_matching_names_cached.

Tendons updated to include the caching.

source/isaaclab/isaaclab/assets/asset_base.py
```diff
@@ -19,6 +20,7 @@
 from isaaclab.physics import PhysicsEvent, PhysicsManager
 from isaaclab.sim.simulation_context import SimulationContext
 from isaaclab.sim.utils.stage import get_current_stage
+from isaaclab.utils.string import resolve_matching_names, resolve_matching_names_values
```
camevor: It seems like this would be fairly easy to implement centrally by adding caching to `resolve_matching_names` (a decorator would suffice) and `resolve_matching_names_values`. The latter still only _matches_ on the names, which are hashable, so at least the matching part can also be cached.

AntoineRichard (approved):
Made the code simpler following @camevor comment.

source/isaaclab/isaaclab/utils/string.py
```diff
@@ -247,50 +248,19 @@ def __deepcopy__(self, memo):
...
-    preserve_order: bool = False,
-    *,
-    raise_when_no_match: bool = True,
-) -> tuple[list[int], list[str]]:
-    """Match a list of query regular expressions against a list of strings and return the matched indices and names.
-
-    When a list of query regular expressions is provided, the function checks each target string against each
-    query regular expression and returns the indices of the matched strings and the matched strings.
+@functools.cache
+def _resolve_matching_names_impl(
```
hougantc-nvda, line 252: @AntoineRichard This is much simpler. I was not sure of taking that approach at first as it seems the internal caching dict would never clean up. In normal use, this wouldn't be an issue, but I was not sure if we need to consider the case of long standing runs where unique assets were being rotated through causing a "leak". If we don't need to consider this case, then this is the far simpler solution.

AntoineRichard: We could also use a LRU cache with a large cache size like 512.

AntoineRichard: @alexmillane you guys are using permanent IsaacLab instances. Do you think you might be impacted by this caching system?

hougantc-nvda: It may be a theoretical issue and perhaps a large LRU cache would suffice. I don't have context on how customers are using IL for long standing runs, just putting it out there in case.

AntoineRichard: @hougantc-nvda I found a reasonable place to clear the cache I think.

hougantc-nvda:
I ran this branch through profiler and compared it against develop

NVTX range	
Develop avg	
Branch avg	
Speedup

Articulation.find_joints
682 µs
52 µs
13.2×
string.resolve_matching_names
659 µs
31 µs
21.0×
string._resolve_matching_names_impl call count
405
9
~97.8% cache hit

The optimization is working as intended. If no other comments, we can merge.
@AntoineRichard @kellyguo11 ?

kellyguo11 (approved):

PR #5208, opened 2026-04-08
Compute mass from collision shapes for zero-mass moving bodies in Newton

Kripner:
Some USD assets - including stock Isaac Lab assets like Franka `panda_leftfinger` / `panda_rightfinger` and the OpenArm hand link - declare zero mass on intermediate moving bodies. This is not an authoring mistake: those assets rely on Isaac Sim's PhysX "auto-compute mass from collider" feature, which Newton's USD importer does not replicate.

The bug: Newton's `bound_mass` validator only floors positive masses below the bound; zero / negative is treated as a "static body". For any body that is the child of a non-fixed joint, the MuJoCo solver then rejects the model with an `mjMINVAL` check at finalize, because it requires positive mass for every non-fixed joint. Manipulation tasks that use these assets (e.g. Isaac-Lift-Cube-Franka-v0) cannot start at all.

The fix: Compute mass and inertia from the body's own collision shapes, before `finalize`.

The obvious alternative is to floor zero-mass moving bodies at some small epsilon (e.g. 1e-6 kg). This makes the solver accept the model, but the result is physically meaningless: a microgram-mass gripper finger cannot transmit contact force to a 200 g cube - the manipulator simply passes through it. The task starts, but no policy can ever solve it. The fix has to give the body a physically reasonable mass, not just a non-zero one.

## How

A new private method `NewtonManager._fix_zero_mass_moving_bodies()` runs once in `_finalize_model`, immediately before `builder.finalize`:

1. Identify all bodies that are the child of a non-fixed joint.
2. For each such body whose authored mass is `<= 0`, walk its collision shapes and call [`newton._src.geometry.inertia.compute_inertia_shape`](https://github.com/newton-physics/newton) on each one with a default density of **1000 kg/m³**. Sum the per-shape masses and inertia tensors.
3. Bodies without any usable collision shape fall back to a small default (`fallback_mass = 0.01 kg`, `fallback_inertia_scalar = 1e-5 kg·m²`) so the solver still has positive values to work with.
4. Write the computed mass / inertia (and their inverses) back into the builder, and emit a single warning summarizing how many bodies were patched.

The fix is non-invasive - bodies with valid authored mass are completely untouched.

## Validation

For a Franka `panda_leftfinger`, the fix computes ~22 g, which is close to the manufacturer spec (~15 g) - close enough that the gripper can actually grip a cube. With this fix in place, the lift / stack manipulation tasks train cleanly under PPO with `presets=newton`. Without it, the same tasks either crash at finalize or never converge because contact forces through the fingers are effectively zero.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [x] My changes generate no new warnings (the fix itself emits an informational warning summarizing what was patched)
- [ ] I have added tests that prove my fix is effective or that my feature works *(see note below)*
- [x] I have added my name to the `CONTRIBUTORS.md`

**Note on tests:** the failure mode is asset-specific (it requires a USD asset that uses PhysX auto-mass-from-collider, which means a real Franka / OpenArm USD or a fixture that mimics one). I'd welcome guidance on where the right home for such a regression test would be. Manually validated against `Isaac-Lift-Cube-Franka-v0`.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -116,6 +116,7 @@ Guidelines for modifications:
 * Manuel Schweiger
 * Masoud Moghani
 * Mateo Guaman Castro
+* Matěj Kripner
 * Maurice Rahme
 * Michael Gussert
 * Michael Lin
--- a/source/isaaclab_newton/config/extension.toml
+++ b/source/isaaclab_newton/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.5.10"
+version = "0.5.11"
 
 # Description
 title = "Newton simulation interfaces for IsaacLab core package"
--- a/source/isaaclab_newton/docs/CHANGELOG.rst
+++ b/source/isaaclab_newton/docs/CHANGELOG.rst
@@ -1,6 +1,26 @@
 Changelog
 ---------
 
+0.5.11 (2026-04-08)
+~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed zero-mass moving bodies in :class:`~isaaclab_newton.physics.NewtonManager`.
+  Some USD assets (e.g. Franka ``panda_leftfinger`` / ``panda_rightfinger``,
+  OpenArm hand link) have zero authored mass because they rely on Isaac Sim's
+  PhysX "auto-compute mass from collider" feature, which Newton's USD importer
+  does not replicate. Newton's ``bound_mass`` validator only floors *positive*
+  masses, so these bodies would either fail the MuJoCo solver's positivity
+  check at finalize, or get a near-zero epsilon mass that made them physically
+  meaningless (e.g. a microgram-mass gripper finger could not transmit force
+  to a 200 g cube). Mass and inertia are now computed from the body's
+  collision shapes via :func:`newton._src.geometry.inertia.compute_inertia_shape`
+  with a default density of 1000 kg/m^3, falling back to a small default for
+  bodies without any usable shape.
+
+
 0.5.10 (2026-04-05)
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -417,6 +417,13 @@ class NewtonManager(PhysicsManager):
         device = PhysicsManager._device
         logger.info(f"Finalizing model on device: {device}")
         cls._builder.up_axis = Axis.from_string(cls._up_axis)
+        # The MuJoCo solver requires moving bodies (those with non-fixed joints) to have
+        # positive mass. Some USD assets (e.g. Franka panda finger, OpenArm hand link)
+        # have zero mass authored because they rely on Isaac Sim's "auto-compute mass
+        # from collider" PhysX feature, which Newton's USD importer does not replicate.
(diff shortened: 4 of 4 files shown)


AntoineRichard:
Hi @Kripner, thanks a lot for debugging this! I'm checking with the newton team if this should not land on newton rather lab. I'll keep you updated. Thanks again :)

eric-heiden:
@Kripner thanks, but this mass correction has been done in the recent version of Newton already (probably in https://github.com/newton-physics/newton/pull/1912), see the following script to verify:

```py
from pxr import Usd, UsdPhysics
import newton

ASSET = ".../IsaacLab/Robots/FrankaEmika/panda_instanceable.usd"

print(f"Inspecting: {ASSET}")
stage = Usd.Stage.Open(ASSET)
for path in [
    "/panda/panda_hand",
    "/panda/panda_leftfinger",
    "/panda/panda_rightfinger",
]:
    prim = stage.GetPrimAtPath(path)
    mass_api = UsdPhysics.MassAPI(prim)
    print(f"\n{path}")
    print("  schemas:", prim.GetAppliedSchemas())
    print("  mass:", mass_api.GetMassAttr().Get())
    print("  diagonal inertia:", mass_api.GetDiagonalInertiaAttr().Get())
    print("  center of mass:", mass_api.GetCenterOfMassAttr().Get())
    print("  density:", mass_api.GetDensityAttr().Get())

builder = newton.ModelBuilder()
result = builder.add_usd(ASSET)
print("\nImporter result:")
for path, idx in sorted(result["path_body_map"].items(), key=lambda kv: kv[1]):
    if any(tok in path for tok in ["panda_hand", "panda_leftfinger", "panda_rightfinger"]):
        print(path)
        print("  body index:", idx)
        print("  imported mass:", builder.body_mass[idx])
        print("  imported com:", builder.body_com[idx])
        print("  body shapes:", builder.body_shapes.get(idx, []))
```
Prints:
```
/panda/panda_hand
  schemas: ['PhysicsRigidBodyAPI', 'PhysicsMassAPI']
  mass: 0.0
  diagonal inertia: (0, 0, 0)
  center of mass: (-inf, -inf, -inf)
  density: 1000.0

/panda/panda_leftfinger
  schemas: ['PhysicsRigidBodyAPI', 'PhysicsMassAPI']
  mass: 0.0
  diagonal inertia: (0, 0, 0)
  center of mass: (-inf, -inf, -inf)
  density: 1000.0

/panda/panda_rightfinger
  schemas: ['PhysicsRigidBodyAPI', 'PhysicsMassAPI']
  mass: 0.0
  diagonal inertia: (0, 0, 0)
  center of mass: (-inf, -inf, -inf)
  density: 1000.0
Module newton._src.geometry.inertia 5855353 load on device 'cuda:0' took 1.42 ms  (cached)

Importer result:
/panda/panda_hand
  body index: 8
  imported mass: 0.708855018173126
  imported com: [-2.804908945108764e-06, -0.0006155292503535748, 0.02392951399087906]
  body shapes: [8, 19]
/panda/panda_leftfinger
  body index: 9
  imported mass: 0.021976267085157316
  imported com: [-2.40856570599135e-05, 0.011958077549934387, 0.023781796917319298]
  body shapes: [9, 20]
/panda/panda_rightfinger
  body index: 10
  imported mass: 0.021976267085157316
  imported com: [2.4086704797809944e-05, -0.011958077549934387, 0.023781796917319298]
  body shapes: [10, 21]
```
These kinds of fixes to the USD importer would also belong to Newton, not here in Isaac Lab.

AntoineRichard:
Closing this as it should already be covered in latest newton. Feel free to re-open if it's not working for you.

PR #5214, opened 2026-04-09
Adds initial support for the Kamino solver

AntoineRichard:
# Description

Adds initial support for the Kamino solver.
Cartpole is training.

## Type of change

## Checklist


source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -637,6 +736,14 @@ def initialize_solver(cls) -> None:
...
                     cls._graph = capture.graph
                     logger.info("Newton CUDA graph captured (standard Warp mode)")
+
+                    # Kamino allocates internal state arrays (joint_q_prev, body_f_total,
+                    # joint_lambdas, FK solver arrays) during graph capture via wp.clone/wp.zeros.
+                    # These memory-pool allocations must be "pinned" by replaying the graph once
+                    # before any non-graphed Kamino operations (e.g. solver.reset()) can safely
+                    # access them.  Without this warm-up, pool-allocated addresses may be stale.
+                    if isinstance(cls._solver, SolverKamino):
+                        wp.capture_launch(cls._graph)
```
AntoineRichard, line 1009: @camevor can you check if this is correct? It looks like we call a solver reset before the first sim step which requires us to do a warm-up step. If we don't do the warm-up step it looks like not everything is initialized.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -166,6 +175,19 @@ def pre_render(cls) -> None:
...
         cls.sync_transforms_to_usd()
 
+    @classmethod
+    def _forward_kamino(cls) -> None:
+        """Kamino-specific forward kinematics via ``solver.reset()``.
+
+        Kamino's ``joint_q`` / ``joint_u`` include coordinates for **all** joints
+        (including free joints), so we pass Newton's full state arrays directly.
+        """
+        cls._solver.reset(
```
rubengrandia, line 243: The full reset without world_mask will have a performance penalty. Kamino does a solver cold-start after a reset. We can think about that later though. Perhaps ` _kamino_needs_fk` needs to be a per world thing?

rubengrandia: Thoughts @vastsoun?

AntoineRichard: Thanks @rubengrandia! Agreed. I already had a chat about this with @camevor we'd have to rework the API a bit to make that happen though.

AntoineRichard: @rubengrandia added the mask logic.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -280,6 +334,14 @@ def step(cls) -> None:
         if sim is None or not sim.is_playing():
             return
 
+        # Kamino: run solver.reset() with the accumulated world mask to reinitialise
+        # internal state (warm-start containers, constraint multipliers) for reset worlds.
```
AntoineRichard, line 365: @rubengrandia do you know if the call here is a no-op if _world_reset_mask is a mask full of zero?

rubengrandia: I think all of the kernels have an early return based on the mask, but there are still a fair bit of kernels launched. Either way I think this would be an optimization to apply inside the Kamino reset.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
vastsoun: There is a pending task/issue to ensure resets will invoke a world-masked cold/warm start so only reset worlds get cold-started. Also,  I think we could add a graph-capturable mechanism to ensure that an all-zeros world mask could be used to skip unnecessary resets. I've created an issue for the second aspect (https://github.com/vastsoun/newton/issues/343)

camevor (approved):

PR #5215, opened 2026-04-09
Fixes multi-GPU device resolution in AppLauncher

kellyguo11:
## Description

Centralizes multi-GPU device resolution in AppLauncher and launch_simulation() so that individual training scripts do not need per-script distributed device logic.

### Problem

In distributed (multi-GPU) training, every RL script duplicates device resolution. This breaks when CUDA_VISIBLE_DEVICES restricts each process to a single GPU (e.g., SLURM), because local_rank=1 exceeds the visible device count. Additionally, physics backends like Newton/Warp that allocate on the current CUDA device during init may all default to cuda:0.

### Fix

**app_launcher.py:** CUDA_VISIBLE_DEVICES-aware device ID resolution, early torch.cuda.set_device(), new self.device attribute.

**sim_launcher.py:** New _resolve_distributed_device() helper centralizes device logic in launch_simulation(). After AppLauncher creation, propagates AppLauncher.device to env_cfg.sim.device. Handles kitless Newton path. Also treats PresetCfg renderers as non-Kit in _is_kit_camera().

**presets.py:** Adds newton and physx aliases to MultiBackendRendererCfg.

## Type of change

## Checklist


ClawLabby (commented):
Reviewed this — the architecture is solid. Centralizing device resolution in AppLauncher + `_resolve_distributed_device()` is exactly right. A few observations from testing the same fixes in #5236:

**1. `isaacsim_rtx_renderer` mutable alias in presets.py (not covered here)**

In `MultiBackendRendererCfg`, the original code has:
```python
isaacsim_rtx_renderer = default
```
This creates an alias — both names point to the same `IsaacRtxRendererCfg` instance. If preset resolution mutates `isaacsim_rtx_renderer`, it also mutates `default`. The fix is:
```python
isaacsim_rtx_renderer: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
```
Small issue but can cause subtle bugs during preset resolution. Happy to open a separate PR for this.

**2. `_is_kit_camera` returning `False` for PresetCfg — timing consideration**

The new check assumes PresetCfg renderers will match the physics backend and therefore aren't necessarily Kit. This is correct for the Newton path, but worth noting: if someone uses a PresetCfg that resolves to `IsaacRtxRendererCfg` (which IS a Kit camera), returning `False` here means `launch_simulation` won't start Kit. In practice this works because the preset resolution happens later and the Kit path is also triggered by other checks, but it's worth a comment.

**3. Tested configurations**

We've been running this exact device resolution logic (same CUDA_VISIBLE_DEVICES fallback + early `set_device`) in 4-GPU distributed training with both Newton and PhysX+RTX for the past few weeks. Configurations tested:
- 4×L40 Newton + Warp renderer + cameras (2048 envs/GPU)
- 4×L40 PhysX + Isaac RTX + cameras (2048 envs/GPU)
- Resume with different env counts across runs

All stable. The early `set_device` is critical for the Newton/Warp path — without it, all ranks allocate on cuda:0 during physics init.

**4. Test coverage looks thorough** — the mocked `device_count`/`set_device` approach is what we converged on too after trying to test with actual GPU detection. 👍

PR #5223, opened 2026-04-10
Pin Sphinx <9 on main for sphinx-multiversion compatibility

AntoineRichard:
## Summary

- Pins `sphinx>=7.0,<9` in `docs/requirements.txt` on `main` to prevent Sphinx 9.x from being installed transitively via `sphinx-book-theme`.
- `sphinx-multiversion==0.2.4` passes positional arguments to `Config.read()`, which became keyword-only in Sphinx 9.0, causing multi-version doc builds to fail with `TypeError: Config.read() takes 2 positional arguments but 3 were given`.
- This is the same fix as #5211 (applied to `develop`), now applied to `main` where the multi-version build for `main`, `release/*`, and version tags actually runs.

## Context

PR #5211 fixed the `develop` branch, but `develop` only builds its own docs (`SMV_BRANCH_WHITELIST=^develop$`). The `main` branch workflow builds all release branches and tags, and its `docs/requirements.txt` had no explicit Sphinx pin — allowing Sphinx 9.x to be resolved transitively.

## Test plan

- [ ] Docs CI passes on this PR (single-version build)
- [ ] After merge, the multi-version Docs build on `main` succeeds and deploys correctly

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -218,6 +218,10 @@ suppress_warnings = [
     # analysis. Since those hints are actual hints that *CANNOT* by definition
     # by canonicalized, our only recourse is to squelch warnings altogether.
     "ref.python",
+    # Sphinx 7+ emits warnings when autosummary entries use fully-qualified
+    # names that include the current module. These are cosmetic and do not
+    # affect the generated output.
+    "autosummary.import_cycle",
 ]
 
 # -- Internationalization ----------------------------------------------------
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,6 +1,6 @@
 # for building the docs
 sphinx>=7.0,<9  # sphinx-multiversion 0.2.4 incompatible with Sphinx 9.x (Config.read() signature change)
-sphinx-book-theme==1.0.1
+sphinx-book-theme>=1.1
 myst-parser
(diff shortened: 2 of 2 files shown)


ooctipus (approved):

kellyguo11 (approved):
Clean fix that mirrors what's already on `develop`. The three changes are well-motivated:

1. **`sphinx>=7.0,<9`** — prevents Sphinx 9.x from breaking `sphinx-multiversion==0.2.4`'s positional `Config.read()` call. Good inline comment explaining the constraint.
2. **`sphinx-book-theme>=1.1`** — necessary because 1.0.1 pins `sphinx<7`, which would conflict with the new lower bound. Aligns with `develop`.
3. **`autosummary.import_cycle` suppression** — Sphinx 7+ emits this for fully-qualified autosummary entries. Cosmetic warning, but would fail the `-W` build. Worth noting this isn't on `develop` yet — might want to port it there too.

One minor note: `sphinx-book-theme>=1.1` (no upper bound) is slightly more permissive than the pinned version it replaces. Not a concern today since 1.1.4 is the latest and the theme API has been stable, but if reproducibility ever becomes an issue, an upper bound could be added. Not blocking.

PR #5227, opened 2026-04-10
Replace sphinx-multiversion with per-branch docs builds

AntoineRichard:
## Summary

Same changes as the develop PR (#5226), applied to main's diverged files.

- Removes `sphinx-multiversion` from docs build
- Each deploy branch builds only its own docs
- Adds version switcher (theme built-in + injected JS for old versions)
- Adds one-time `docs-migration.yaml` workflow
- Pins `sphinx>=7.0,<9` and `sphinx-book-theme>=1.1` (supersedes #5223)

See spec: `docs/superpowers/specs/2026-04-10-multi-version-docs-design.md`

## Test plan

- [ ] Docs CI builds successfully on this PR
- [ ] After merge, verify docs deploy correctly for main


docs/conf.py
```diff
@@ -278,21 +281,14 @@
         },
     ],
     "icon_links_label": "Quick Links",
+    "switcher": {
```
kellyguo11, line 302: This PR supersedes #5223, which added `suppress_warnings = ["autosummary.import_cycle"]` to fix the 18 import_cycle warnings that break `-W` builds on main. That suppression isn't carried forward here, so the docs build will continue to fail.

Also, the pydata-sphinx-theme version switcher emits a warning when `versions.json` returns 404 (which it will until first deploy), adding another `-W` failure.

Suggestion: add the suppression from #5223 here too:
```python
suppress_warnings = ["autosummary.import_cycle"]
```
And consider suppressing the version-switcher fetch warning as well, or providing a placeholder `versions.json`.

.github/workflows/docs-release.yml
```diff
@@ -20,49 +22,64 @@ concurrency:
...
     outputs:
       trigger-deploy: ${{ steps.trigger-deploy.outputs.defined }}
     steps:
     - id: trigger-deploy
       env:
         REPO_NAME: ${{ secrets.REPO_NAME }}
-      if: "${{ github.repository == env.REPO_NAME && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/devel' || startsWith(github.ref, 'refs/heads/release/')) }}"
+      if: >-
+        github.repository == env.REPO_NAME &&
+        (
```
kellyguo11, line 41: Both `actions/checkout@v6` and `actions/upload-artifact@v7` / `actions/download-artifact@v8` referenced in this workflow don't exist yet. Current latest versions are `@v4` for all of these. The workflow will fail to run as-is.

Same issue in `docs-migration.yaml`.

.github/workflows/docs-release.yml
```diff
@@ -74,14 +91,116 @@ jobs:
...
+          SLUG="${SLUG//\//-}"
+        fi
+        echo "slug=$SLUG" >> "$GITHUB_OUTPUT"
+        echo "Version slug: $SLUG"
+
+    - name: Assemble deploy directory
+      env:
+        SLUG: ${{ steps.slug.outputs.slug }}
+      run: |
+        mkdir -p site
```
kellyguo11, line 142: The `SLUG` and `NAME` variables are interpolated into the Python heredoc via `'${SLUG}'` / `'${NAME}'`. If a branch/tag name ever contains a single quote, this breaks the Python string literal. Safer approach:

```python
import os
slug = os.environ['SLUG']
name = os.environ['NAME']
```

(with `SLUG` and `NAME` passed as env vars to the step, which you already have for `SLUG`).

kellyguo11 (commented):
## Review: Replace sphinx-multiversion with per-branch docs builds

Good direction — `sphinx-multiversion` has been unmaintained since 2020 and has been a recurring source of build pain. Per-branch builds are simpler, more reliable, and easier to reason about. This mirrors the same approach as #5226 on develop, applied to main's diverged files.

### CI Status

Both CI failures are **pre-existing / not caused by this PR**:

1. **Docs build failure** — 18 `autosummary.import_cycle` warnings (pre-existing on main) treated as errors by `-W`. Additionally, the `pydata-sphinx-theme` version switcher tries to fetch `versions.json` at build time and logs a warning when it 404s (expected — the file won't exist until the first deploy). The previous PR #5223 added `suppress_warnings = ["autosummary.import_cycle"]` to `conf.py` but this PR doesn't include that suppression. Since this PR says it supersedes #5223, it should carry forward the warning suppression.

2. **Link checker failure** — `https://graphics.pixar.com/usd/docs/index.html` returns 404. Pre-existing, unrelated to this PR.

### Issues

**Should-fix:**

1. **Missing `suppress_warnings` for `autosummary.import_cycle`** — PR #5223 (which this supersedes) added suppression for these warnings. Without it, the `-W` flag in `make current-docs` will keep failing on main. Need to add `suppress_warnings = ["autosummary.import_cycle"]` to `conf.py` or the build will never pass.

2. **`versions.json` 404 warning breaks `-W` build** — When building locally or in CI before the first deploy, the pydata-sphinx-theme version switcher fetches `versions.json` and emits a warning on 404. With `-W` (warnings as errors), this is fatal. Consider either:
   - Suppressing this specific warning category in conf.py
   - Adding a placeholder `versions.json` to the repo
   - Or documenting that the first build will need `-W` relaxed

3. **Shell injection risk in `docs.yaml` deploy step** — The `SLUG` variable is interpolated directly into shell via `${SLUG}` in the "Assemble deploy directory" and "Update versions.json" steps. While the branch/tag names are constrained by the workflow trigger patterns, the Python script uses `'${SLUG}'` and `'${NAME}'` (single-quoted shell substitution inside a Python heredoc), which could break if a tag/branch name contains single quotes. Consider passing these as environment variables that Python reads via `os.environ` instead.

4. **`docs-migration.yaml` uses `actions/download-artifact@v8`** — This action version doesn't exist yet (current latest is v4). The main `docs.yaml` also uses `@v7` for upload and `@v8` for download. These should be pinned to actually-released versions (v4 for both).

5. **`actions/checkout@v6` doesn't exist** — Current latest is v4. Both workflow files reference `@v6`. This will cause the workflows to fail.

**Suggestions:**

6. **Duplicate slug determination logic** — The "Determine version slug" step appears in both `build-docs` and `deploy-docs` jobs in `docs.yaml` with identical logic. Consider extracting it once and passing via job outputs, or at minimum adding a comment noting the intentional duplication.

7. **Migration workflow `actions/upload-artifact@v7`** — Same version concern as above; should be `@v4`.

8. **XHR in version-switcher.js** — `XMLHttpRequest` is legacy. Since this is new code, consider using `fetch()` with a fallback, or just `fetch()` since all modern browsers support it. Not blocking, but the code would be cleaner.

9. **HEAD request for page existence check** — The version switcher does a HEAD request to check if a page exists in the target version before navigating. This is a nice UX touch, but be aware some CDN/hosting setups don't handle HEAD well for static files. A comment explaining the fallback behavior would help future maintainers.

10. **`devel` → `develop` rename in trigger** — The workflow trigger changes `devel` to `develop`. Make sure all references to the old branch name are cleaned up across the repo (not just this file).

### Overall

Solid architectural improvement. The standalone `version-switcher.js` approach is clever — it allows injecting version switching into old builds that never had it. The migration workflow for one-time historical rebuild is well thought out.

Main blockers are the action version references (v6/v7/v8 don't exist) and the missing warning suppression from #5223. Fix those and this is good to go.

PR #5219, opened 2026-04-10
Exposes Newton collision pipeline configuration through NewtonCfg

ooctipus:
# Description

Expose Newton's `CollisionPipeline` parameters through a new `NewtonCollisionPipelineCfg` configclass, allowing users to customize broad phase, contact limits, and hydroelastic mode via `NewtonCfg.collision_cfg`.

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5228, opened 2026-04-10
Add SDF collision and hydroelastic shape configuration

AntoineRichard:
## Summary

- Add `SDFCfg` configclass for SDF mesh collision configuration (body/shape regex patterns, resolution overrides, hydroelastic flag enablement)
- Add missing `HydroelasticSDFCfg` pipeline params (`moment_matching`, `buffer_mult_*`, `grid_size`)
- Add manager methods to build SDF on matching shapes before model finalization
- Integrate with Newton cloner to apply SDF on prototypes before replication and skip convex hull for SDF shapes
- Force Newton collision pipeline when SDF is configured
- Add warning when hydroelastic config has no effect (no qualifying shape pairs)

Built on top of #5219. Ports SDF/hydroelastic shape preparation from #5160 into #5219's config design.

## Test plan

- [x] 14 unit tests for `_build_sdf_on_mesh` and `_apply_sdf_config` (pattern matching, resolution overrides, hydroelastic flags)
- [ ] Integration test with a Newton scene using `SDFCfg` + `NewtonCollisionPipelineCfg`


source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -387,6 +392,190 @@ def add_model_change(cls, change: SolverNotifyFlags) -> None:
...
+            for pat, res in res_overrides:
+                if pat.search(label):
+                    resolution = res
+                    break
+        sdf_kwargs: dict = dict(narrow_band_range=sdf_cfg.narrow_band_range)
+        if resolution is not None:
+            sdf_kwargs["max_resolution"] = resolution
+        if sdf_cfg.target_voxel_size is not None:
+            sdf_kwargs["target_voxel_size"] = sdf_cfg.target_voxel_size
+        mesh.build_sdf(**sdf_kwargs)
```
kellyguo11, line 424: **`max_resolution` and `target_voxel_size` forwarded simultaneously despite docs saying one overrides the other**

+1 to Greptile's finding. The `SDFCfg.max_resolution` docstring says "Ignored when `target_voxel_size` is set", but both are unconditionally passed to `mesh.build_sdf()`. Either:

1. Make the code match the docs (use `elif` so `target_voxel_size` takes precedence), or
2. Update the docs to say they're both forwarded and let Newton decide.

Option 1 is safer — it's what users will expect from reading the config:

```python
        sdf_kwargs: dict = dict(narrow_band_range=sdf_cfg.narrow_band_range)
        if sdf_cfg.target_voxel_size is not None:
            sdf_kwargs["target_voxel_size"] = sdf_cfg.target_voxel_size
        elif resolution is not None:
            sdf_kwargs["max_resolution"] = resolution
        mesh.build_sdf(**sdf_kwargs)
```

And update `test_target_voxel_size_passed_alongside_resolution` accordingly.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -592,22 +797,38 @@ def initialize_solver(cls) -> None:
...
+                        "NewtonManager: collision_cfg cannot be set when use_mujoco_contacts=True."
+                        " Either set use_mujoco_contacts=False or remove collision_cfg."
+                    )
             else:
                 cls._needs_collision_pipeline = True
 
+            # Force Newton pipeline when collision_cfg or SDF is configured
+            if cfg.collision_cfg is not None and not cls._needs_collision_pipeline:
+                logger.warning("collision_cfg set — enabling Newton collision pipeline.")
+                cls._needs_collision_pipeline = True
```
kellyguo11: **Dead code — this warning block is unreachable**

+1 to Greptile. By this point:
- MuJoCo + `use_mujoco_contacts=True` + `collision_cfg` → already raised `ValueError` above
- MuJoCo + `use_mujoco_contacts=False` → `_needs_collision_pipeline` is already `True`
- Non-MuJoCo solvers → `_needs_collision_pipeline` is already `True`

Suggest removing this block. If it's intended as a safety net for future solver types, add a comment explaining that.

kellyguo11 (commented):
## Review: Add SDF collision and hydroelastic shape configuration

Solid PR that cleanly ports SDF/hydroelastic shape preparation from #5160 into #5219's config design. The config hierarchy is well-structured, the pattern-matching system is flexible, and the cloner integration correctly preserves triangle meshes for SDF shapes. The 14 unit tests cover the key logic paths well.

I agree with Greptile's P2 findings — the `max_resolution`/`target_voxel_size` doc-vs-code mismatch and the dead-code warning block should be cleaned up. A few additional items below.

source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py
```diff
@@ -74,7 +87,33 @@ def _build_newton_builder_from_mapping(
...
+                    # Skip shapes that will use SDF (matched by body or shape pattern)
+                    if p.shape_body[i] in sdf_bodies:
+                        continue
+                    if shape_pats is not None:
+                        lbl = p.shape_label[i] if i < len(p.shape_label) else ""
+                        if any(pat.search(lbl) for pat in shape_pats):
+                            continue
+                    approx_indices.append(i)
+                if approx_indices:
+                    p.approximate_meshes("convex_hull", shape_indices=approx_indices, keep_visual_shapes=True)
```
kellyguo11, line 121: **`_apply_sdf_config` called on prototype builders, but it reads `PhysicsManager._cfg`**

This works at runtime because `PhysicsManager._cfg` is set before cloning begins. But it creates an implicit coupling: `_apply_sdf_config` is a classmethod on `NewtonManager` that reads global state from `PhysicsManager._cfg`, and now it's being called with prototype `ModelBuilder` objects (not the final builder).

This is fine for now, but consider:
1. Adding a brief comment on the prototype builder noting that `_apply_sdf_config` reads from `PhysicsManager._cfg`, not from the builder itself.
2. Long-term, passing `sdf_cfg` explicitly as a parameter rather than reading global state would make this more testable and less fragile.

Non-blocking — the current approach works correctly.

source/isaaclab_newton/isaaclab_newton/physics/newton_collision_cfg.py
```diff
@@ -0,0 +1,331 @@
...
+    max_resolution: int | None = None
+    """Maximum voxel dimension for the SDF grid.
+
+    Must be divisible by 8. Typical values: 128, 256, 512.
+    Ignored when :attr:`target_voxel_size` is set.
+
+    Defaults to ``None``.
+    """
+
+    target_voxel_size: float | None = None
```
kellyguo11, line 252: **`SDFCfg.max_resolution` docstring says "Ignored when `target_voxel_size` is set" but that's not what the code does**

As noted on the manager side, the code currently passes both. Either update this docstring to match reality, or fix the code to match this docstring. Keeping them in sync avoids user confusion.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -387,6 +392,190 @@ def add_model_change(cls, change: SolverNotifyFlags) -> None:
...
+                body=body_idx,
+                xform=builder.shape_transform[visual_si],
+                mesh=mesh,
+                scale=builder.shape_scale[visual_si],
+                cfg=sdf_shape_cfg,
+                label=f"{body_lbl}/sdf_collision",
+            )
+            num_added += 1
+            if sdf_cfg.k_hydro is not None:
+                num_hydro += 1
```
kellyguo11, line 497: **Silent truncation to first visual mesh per body**

+1 to Greptile. When a body has multiple visual meshes, only the first is used with no warning. At minimum, log a warning when `len([si for si in sdf_shape_indices if builder.shape_body[si] == body_idx and builder.shape_source[si] is not None]) > 1`.

Better yet, iterate over all visual meshes for the body, since SDF collision should cover the full geometry.

source/isaaclab_newton/isaaclab_newton/physics/newton_collision_cfg.py
```diff
@@ -0,0 +1,331 @@
...
+    """Collision margin [m] for SDF shapes.
+
+    When ``None``, the Newton builder default is used.
+
+    Defaults to ``None``.
+    """
+
+    body_patterns: list[str] | None = None
+    """Regex patterns for body labels.
+
```
kellyguo11, line 280: nit: The docstring says `(-0.1, 0.1)` for the default, which matches the field default. But consider documenting the physical meaning more explicitly — e.g. "10cm inside / 10cm outside the mesh surface" — since users unfamiliar with SDF narrow bands may not understand the units immediately.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager_cfg.py
```diff
@@ -218,3 +240,32 @@ class NewtonCfg(PhysicsCfg):
...
+    """
+
+    sdf_cfg: SDFCfg | None = None
+    """SDF collision configuration.
+
+    When set, mesh collision shapes matching the configured patterns will have
+    SDF built via Newton's ``mesh.build_sdf()`` at simulation start.  This
+    also forces Newton's collision pipeline to be active (overriding
+    ``use_mujoco_contacts=True`` if necessary).
+
```
kellyguo11, line 268: **`sdf_cfg` docstring says it forces Newton collision pipeline, but the actual override only happens for `use_mujoco_contacts=True`**

The docstring says: "This also forces Newton's collision pipeline to be active (overriding `use_mujoco_contacts=True` if necessary)."

But in `initialize_solver`, the SDF override only fires when `not cls._needs_collision_pipeline`. For non-MuJoCo solvers, the pipeline is already forced on, so there's no override. For MuJoCo with `use_mujoco_contacts=True`, the `collision_cfg` check above already raises a `ValueError` — so SDF with MuJoCo contacts would need to bypass that check first.

Is the intent that SDF should work with `use_mujoco_contacts=True` by silently overriding it? If so, the `ValueError` raised for `collision_cfg + use_mujoco_contacts=True` would block that path. Consider clarifying the intended behavior.

docs/superpowers/plans/2026-04-10-newton-sdf-hydroelastic.md
```diff
@@ -0,0 +1,1106 @@
+# Newton SDF & Hydroelastic Config Implementation Plan
```
kellyguo11: nit: Consider whether this 1100-line plan doc should be included in the final PR. It's great for development context but adds significant weight to the repo's `docs/` tree. If this is an internal-only workflow artifact, it might be better in a separate location or excluded from the PR.

source/isaaclab_newton/docs/CHANGELOG.rst
```diff
@@ -1,6 +1,46 @@
 Changelog
 ---------
 
+0.5.12 (2026-04-10)
```
kellyguo11, line 4: nit: RST heading underline should match the title length. `0.5.12 (2026-04-10)` is 19 chars but the underline `~~~~~~~~~~~~~~~~~~~` is 19 tildes — looks correct actually. 👍

source/isaaclab_newton/config/extension.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.5.10"
+version = "0.5.12"
```
kellyguo11, line 4: **Version jumps from 0.5.10 to 0.5.12, skipping 0.5.11**

The changelog adds both 0.5.11 and 0.5.12 entries, but this PR's base presumably already has 0.5.10. Since this PR is built on top of #5219, 0.5.11 likely comes from that PR. Make sure the version bump chain is correct when both PRs merge — if #5219 sets 0.5.11, then this PR setting 0.5.12 is correct, but if #5219 doesn't bump the version, you'll have a gap in the version history.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -387,6 +392,217 @@ def add_model_change(cls, change: SolverNotifyFlags) -> None:
...
+        sdf_kwargs: dict = dict(narrow_band_range=sdf_cfg.narrow_band_range)
+        if resolution is not None:
+            sdf_kwargs["max_resolution"] = resolution
+        if sdf_cfg.target_voxel_size is not None:
+            sdf_kwargs["target_voxel_size"] = sdf_cfg.target_voxel_size
+        mesh.build_sdf(**sdf_kwargs)
+        return True
+
+    @classmethod
+    def _create_sdf_collision_from_visual(
```
ooctipus, line 428: Hmmmm physx sdf is defined in colliders, and modified through collision_prop_cfg, e.g. you can have convex hull nut collider, and sdf nut collider while convex hull nut collide convex hull robot, while sdf nut collides sdf bolt, but this implementation seems like that newton is global overriding all visual meshes to sdf?

PR #5011, opened 2026-03-13
Adds perspective video recording via Newton GL viewer for Kitless backends; Fix for Kitfull backends camera position

bdilinila:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5231, opened 2026-04-10
Add documentation for setup of perspective video recording

bdilinila:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


bareya (approved):

docs/source/features/record_video_configuration.rst
```diff
@@ -0,0 +1,137 @@
+.. _features_record_video:
```
kellyguo11: I think this may be better placed as a how-to guide than a highlighted feature

docs/source/features/record_video_configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+
+.. literalinclude:: ../../../source/isaaclab/isaaclab/envs/common.py
+   :language: python
+   :lines: 20-28
+
+
+Backend selection: Kit vs Newton GL
+------------------------------------
+
+``VideoRecorder`` resolves the implementation from the live :class:`~isaaclab.scene.InteractiveScene`: PhysX physics or Isaac RTX in the sensor stack selects the Kit path (``omni.replicator`` on ``/OmniverseKit_Persp``). The Newton GL path applies to kit-less stacks where Newton physics is paired with either the Newton Warp renderer or the OVRTX renderer (from ``isaaclab_ov``): a headless ``ViewerGL`` draws the Newton model supplied by the scene data provider. The resolver also treats ``newton_warp`` appearing among sensor renderer types as selecting this path (see the snippet below). If both Kit and Newton GL signals are present, the Kit path is chosen.
```
kellyguo11: could newton visualizer not also work with isaac sim physx? I don't think it's purely for kit-less

docs/source/features/record_video_configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+.. list-table::
+   :widths: 28 36 36
+   :header-rows: 1
+
+   * - Stack example
+     - Video backend
+     - Capture mechanism
+   * - PhysX simulation + Isaac RTX renderer (Kit / full Sim)
+     - Kit (``"kit"``)
+     - ``/OmniverseKit_Persp`` + Replicator RGB
```
kellyguo11: @matthewtrepte can you make sure this documentation still aligns with your latest changes?

kellyguo11:
Spotted a failure in the doc build job `/home/runner/work/IsaacLab/IsaacLab/docs/source/features/record_video_configuration.rst:4: WARNING: Title underline too short.`

docs/source/features/record_video_configuration.rst
bdilinila: There is an existing how-to guide /home/horde/git/IsaacLab/docs/source/how-to/record_video.rst. 

Would it be best to just append my sections to that document? My changes focus more on how the recording is set up and how the viewer is resolved.

docs/source/features/record_video_configuration.rst
bdilinila: This will result in an error:
`RuntimeError: Newton GL perspective video requires a Newton model on the scene data provider. Do not use --video for this setup.`

If we want to use the Newton GL Viewer, then we need Newton to be loaded. I can rephrase it along the lines of 
`The Newton GL path applies to stacks where Newton is loaded by the simulation or the warp renderer backend: ...`

kellyguo11: oh, even if we specify --visualizer newton, it won't build the Newton model for the visualization? @matthewtrepte do you know if that path would work?

docs/source/features/record_video_configuration.rst
kellyguo11: Yes, I think we can combine them into the existing record_video page. if should all be one big feature that does the same functionality, but just for different combinations of simulation/visualization

docs/source/features/record_video_configuration.rst
bdilinila: I did not try that. When I worked on PR5011 I never tried/considered running --visualizer along with --video.

kellyguo11: I see, might be worthwhile to clarify what paths this current support. I was imagining we can capture recordings for any combinations of physics engine and visualizer

docs/source/features/record_video_configuration.rst
bdilinila: Done

docs/source/features/record_video_configuration.rst
bdilinila: Done. Thank you

docs/source/features/record_video_configuration.rst
bdilinila: Just a heads up, the documentation was moved to the existing docs/source/how-to/record_video.rst.

docs/source/features/record_video_configuration.rst
matthewtrepte: @bdilinila hm in headless mode, how do we specify newton vs kit based recording? does it just use the renderer configured in the env?

bdilinila: It resolves by checking whether a Kit backend is being used in order to use the OmniversePersp_Kit camera for render output; otherwise, it would use Newton GL Viewer to create a camera:
- [checking backends](https://github.com/isaac-sim/IsaacLab/pull/5011/changes#diff-1051e0a7801abd2a2ce22a559b3a5085424673841de77b18b976fa7e362d28d5R48)

If you have `presets=physx,...`, `presets=isaacsim_rtx_renderer,...`, then it will resolve to using the Kit camera. 

If we have a case like `presets=ovrtx_renderer,...`, then it depends on the physics backend.

docs/source/how-to/record_video.rst
kellyguo11, line 17: I think this might no longer be necessary?

docs/source/how-to/record_video.rst
```diff
@@ -23,3 +26,153 @@ Example usage:
...
+.. literalinclude:: ../../../source/isaaclab/isaaclab/envs/utils/video_recorder.py
+   :language: python
+   :lines: 70-114
+
+
+Step 1 - Manager-based RL (``ManagerBasedRLEnv``)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The recorder is created inside :class:`~isaaclab.envs.manager_based_env.ManagerBasedEnv`, so
+``ManagerBasedRLEnv`` must set ``cfg.video_recorder`` before ``super().__init__``.
```
kellyguo11, line 103: is this needed even with the `--video` flag? the previous workflow we had was that adding `--video` would enable recording automatically without requiring additional configs

bdilinila: These sections probably do not need to be added anymore now that they were added to the how-to guide.

I had initially written this more for developers to explain the flow of how the video recording was activated and set up behind the scenes (e.g., getting the camera position and target).  Octi had previously asked about how the eye and lookat / position and target were calculated ([thread](https://nvidia.slack.com/archives/C0ADR7ELXAB/p1775520324047669?thread_ts=1773196123.491359&cid=C0ADR7ELXAB)), so I could just focus on where those values are gathered from and how they are used to set up the perspective view. Would that make sense to do?

docs/source/how-to/record_video.rst
bdilinila: Before I remove the line, you are referring to the --enable_cameras option, correct? Yes, we can run --video without it.

docs/source/features/record_video_configuration.rst
kellyguo11: so I think the current implementation requires a renderer to be defined. the next step would be to extend this to allow for capturing of visualizers as well without requiring an explicit renderer

docs/source/how-to/record_video.rst
kellyguo11: yup exactly, the --enable_cameras option

docs/source/how-to/record_video.rst
kellyguo11: sounds good, we can focus on just how users would specify these and define the configurations for recording. would be good to also mention the default configurations if users just pass `--video` without overriding any other cfg objects.

kellyguo11:
hmm looks like we are still getting some errors in the doc build job

bdilinila:
Fixed doc build locally; however, we have a new doc [error](https://github.com/isaac-sim/IsaacLab/actions/runs/24739356241/job/72374598915?pr=5231#step:5:209):

`### Errors in source/isaaclab_tasks/isaaclab_tasks/manager_based/locomotion/velocity/config/spot/README.md
`

This file is not part of my changes. Would this be ignored, @kellyguo11 ?

I also requested isaaclab-review-bot to re-review, but I do not know how to check whether it has done so. Do I reject the previously requested changes since I resolved them?

kellyguo11 (approved):

PR #5232, opened 2026-04-10
OMPE-88188: Enable frame stacking for newton+newton_renderer by default

jmart-nv:
# Description

This enables frame stacking for newton+warp by default.

*Previously this PR also included changes to RTX's default AA, which have been reverted. This was intended to remove the implicit temporal information introduced by DLSS and TAA. Instead, we now focus on explicitly adding temporal data via frame stacking where needed.*

**Newton Frame Stacking:** _Provides explicit temporal data to newton._

RTX uses DLSS anti-aliasing by default, which provides implicit temporal data. The newton_renderer does not provide temporal data. However, newton's energy-conserving physics solver requires temporal velocity data in order to compensate for the lack of damping, which causes convergence problems when paired with newton_renderer.

This commit provides explicit temporal information via 2-frame stacking by default for camera tasks that use newton+newton_renderer. This allows newton to provide the damping it needs to converge at the same rate as physx. This adds 36% GPU memory overhead, but the wall clock overhead is negligible.

The default for all other physics/renderer backends is still stack size = 1 (disabled) since physx has implicit damping built-in via its TGS solver, and RTX provides temporal data implicitly.

Implementation:

CameraCfg now has a `frame_stack` field that controls a ring buffer of previous frames. These are concatenated to the channel dimension automatically in `DirectRLEnv`. The new `MultiBackendCameraCfg` wraps the `MultiBackendRendererCfg` to provide defaults for the different presets, driven by nested config classes and capability flags. Warnings will be logged if it detects that a physics backend that needs explicit temporal data is paired with a renderer that does not provide it.

Updated existing tasks to use the new MultiBackendCameraCfg and added 65 new unit tests.

## Type of change

## Screenshots

profiling_charts
convergence_final
frame_comparison_grid

## Checklist

- [ ] My changes generate no new warnings *(N/A - new warnings are intentional)*


source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -220,10 +226,36 @@ def _update_buffers_impl(self, env_mask: wp.array):
...
-            self.renderer.write_output(self.render_data, output_name, output_data)
+        if self._frame_stack_size > 1:
+            # Renderer wrote into _single_frame_output. Update ring buffer and stacked output.
+            needs_init = self._frame_stack_needs_init
+            for name, single in self._single_frame_output.items():
+                if name != "rgb":
+                    self.renderer.write_output(self.render_data, name, single)
+                history = self._frame_history[name]
+                # For envs that just reset, fill all history slots with current frame
+                if needs_init.any():
```
kellyguo11, line 237: **should-fix:** `needs_init.any()` triggers a GPU→CPU synchronization on every update step, for every output channel. In the common case (no environments were just reset), this is a wasted sync that stalls the pipeline.

Consider tracking reset state with a CPU-side flag instead:

```python
# In reset():
self._frame_stack_needs_init_cpu = True  # simple bool

# Here:
if self._frame_stack_needs_init_cpu:
    init_ids = self._frame_stack_needs_init.nonzero(as_tuple=False).squeeze(-1)
    if len(init_ids) > 0:
        for i in range(self._frame_stack_size):
            history[i, init_ids] = single[init_ids]
    self._frame_stack_needs_init.zero_()
    self._frame_stack_needs_init_cpu = False
```

This avoids the GPU→CPU sync entirely in the steady-state path.

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
```diff
@@ -220,10 +226,36 @@ def _update_buffers_impl(self, env_mask: wp.array):
...
+                    init_ids = needs_init.nonzero(as_tuple=False).squeeze(-1)
+                    for i in range(self._frame_stack_size):
+                        history[i, init_ids] = single[init_ids]
+                # Write current frame into ring buffer
+                history[self._frame_stack_idx].copy_(single)
+                # Build stacked output: concatenate along channel dim, oldest first
+                ordered = torch.cat(
+                    [
+                        history[(self._frame_stack_idx + 1 + i) % self._frame_stack_size]
+                        for i in range(self._frame_stack_size)
```
kellyguo11, line 247: **nit (perf):** The `torch.cat` with a list comprehension per update is fine for `frame_stack=2` but scales as O(frame_stack) allocations per step. Since the API allows arbitrary `frame_stack` values, worth either:
- Documenting that values > 4 are not recommended for performance reasons, or
- Using pre-allocated views with `torch.narrow` + in-place copies instead of `cat`

Not blocking for the current use case (frame_stack=2).

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -591,6 +591,26 @@ def _configure_gym_env_spaces(self):
...
+                frame_stack = getattr(attr, "frame_stack", None)
+                if isinstance(frame_stack, int) and frame_stack > 1:
+                    h, w, c = self.cfg.observation_space
+                    self.cfg.observation_space = [h, w, c * frame_stack]
+                    logger.info(
+                        "Auto-adjusted observation_space channels %d -> %d for frame_stack=%d on '%s'.",
+                        c,
+                        c * frame_stack,
+                        frame_stack,
+                        attr_name,
```
kellyguo11, line 610: **suggestion:** This `break` means only the first camera with `frame_stack > 1` adjusts the observation space. For single-camera tasks this is correct, but it silently ignores additional cameras.

Worth adding a comment like `# NOTE: Only the first camera with frame_stack > 1 is used to adjust observation_space. Multi-camera tasks should set observation_space explicitly.`

Also: the observation_space auto-adjust assumes the space is `[H, W, C]` from a single camera. If a task uses a different observation layout, the multiplication could corrupt the space dimensions. A guard like checking that `c` matches the camera's expected channel count would make this more robust.

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
```diff
@@ -173,6 +180,17 @@ def launch_simulation(
     visualizer_intent = _compute_visualizer_intent(env_cfg)
     _set_visualizer_intent_on_launcher_args(launcher_args, visualizer_intent)
 
+    # Warn when Newton physics is used with camera observations but no frame stacking.
```
kellyguo11, line 183: **should-fix:** `_has_camera_without_frame_stack` checks `isinstance(node, CameraCfg)` but `frame_stack` is only defined on `TiledCameraCfg`. A regular `CameraCfg` (non-tiled) will hit `getattr(node, "frame_stack", 1) <= 1` → `True`, causing a false-positive warning even though frame stacking isn't applicable to non-tiled cameras.

Should be:
```python
from isaaclab.sensors import TiledCameraCfg

def _has_camera_without_frame_stack(node) -> bool:
    if not isinstance(node, TiledCameraCfg):
        return False
    return getattr(node, "frame_stack", 1) <= 1
```

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -56,17 +56,26 @@ class RenderCfg:
...
-
-    - **DLSS**: Boosts performance by using AI to output higher resolution frames from a lower resolution input.
-      DLSS samples multiple lower resolution images and uses motion data and feedback from prior frames to reconstruct
-      native quality images.
-    - **DLAA**: Provides higher image quality with an AI-based anti-aliasing technique. DLAA uses the same
-      Super Resolution technology developed for DLSS, reconstructing a native resolution image to maximize
-      image quality.
-
-    This is set by the variable: ``/rtx/post/dlss/execMode``.
+    antialiasing_mode: Literal["Off", "FXAA", "DLSS", "TAA", "DLAA"] | None = "FXAA"
```
kellyguo11, line 59: **suggestion (breaking change):** Changing the default from `None` (preserve renderer default, which was DLSS) to `"FXAA"` is a user-facing behavioral change — existing code that relied on the previous default will now get different rendering output and potentially different performance characteristics.

This is the right default for RL training, but it should be called out in the changelog since it affects all users, not just Newton users. The PR checklist has the changelog item unchecked — please add an entry noting the default AA mode change.

source/isaaclab/test/sensors/test_tiled_camera_frame_stack.py
```diff
@@ -0,0 +1,329 @@
...
+def test_frame_stack_partial_reset_preserves_others(setup_scene):
+    """Resetting env 0 should not affect env 1's history."""
+    sim = setup_scene
+    camera = TiledCamera(_make_camera_cfg(frame_stack=2))
+    sim.reset()
+
+    # Step to build history
+    for _ in range(3):
+        sim.step()
+        camera.update(dt=0.01)
```
kellyguo11, line 186: **nit:** `import logging` has no dependency on the simulator and can be moved to the top-level imports (before `AppLauncher`). Only `isaaclab_tasks` and `pxr` need to be deferred.

source/isaaclab/test/sensors/test_tiled_camera_frame_stack.py
```diff
@@ -0,0 +1,329 @@
...
+    assert cfg.frame_stack == 1
+
+
+# -- Newton warning and preset resolution tests --
+# These require isaaclab_tasks for launch_simulation and preset configs.
+
+import logging
+
+import isaaclab_tasks  # noqa: F401
+from isaaclab_tasks.utils import resolve_task_config
```
kellyguo11, line 214: **suggestion:** Manipulating `sys.argv` directly is fragile — if tests run in parallel, another test could see the modified argv. Consider using `unittest.mock.patch('sys.argv', [...])` as a context manager, which is both cleaner and thread-safe:

```python
from unittest.mock import patch

with patch('sys.argv', [sys.argv[0], 'presets=newton']):
    env_cfg, _ = resolve_task_config(...)
```

source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
```diff
@@ -18,3 +20,19 @@ class MultiBackendRendererCfg(PresetCfg):
...
+class MultiBackendTiledCameraCfg(TiledCameraCfg):
+    """TiledCameraCfg with multi-backend renderer and automatic Newton frame stacking.
+
+    When ``presets=newton`` is used, ``frame_stack`` is automatically set to 2.
+    Newton's energy-conserving (symplectic) integrator produces dynamics that
+    require velocity information for effective camera-based control. Frame stacking
+    provides this temporal information by concatenating consecutive frames along
+    the channel dimension, enabling the policy to infer velocity from pixel
+    differences between frames.
+    """
```
kellyguo11, line 35: **suggestion:** Nice API — `MultiBackendTiledCameraCfg` as a `TiledCameraCfg` subclass with embedded `MultiBackendRendererCfg` is cleaner than the previous pattern of separate `renderer_cfg` fields.

One thing to document: downstream code using `type(cam) == TiledCameraCfg` (exact type check) would now fail. `isinstance` checks still work. Might be worth a migration note in the PR description or changelog.

kellyguo11 (commented):
## Review Summary

Good two-part PR that addresses a real sim2real gap: (1) disabling temporal AA (DLSS) by default so RL policies don't learn from renderer-specific temporal artifacts, and (2) adding frame stacking for Newton to explicitly provide temporal information that Newton's energy-conserving integrator needs.

The implementation is solid — ring buffer approach for frame stacking is clean, the `MultiBackendTiledCameraCfg` preset pattern is a nice API improvement over the previous `renderer_cfg=MultiBackendRendererCfg()` approach, and the test coverage is thorough (12+ new tests covering frame stacking, ring buffer correctness, reset behavior, and AA mode warnings).

A few performance and correctness issues worth addressing below.

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
jmart-nv: Added `_frame_stack_needs_init_cpu` (now in camera.py)

source/isaaclab/isaaclab/sensors/camera/tiled_camera.py
jmart-nv: Replaced `torch.cat` with a per-slot narrow + copy (now in capera.py)

source/isaaclab/isaaclab/envs/direct_rl_env.py
jmart-nv: This now collects every camera field with frame_stack > 1 and auto-adjusts the observation space (raises an error if the values mismatch)

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
jmart-nv: No longer applicable (now implemented in CameraCfg since TiledCameraCfg is deprecated)

source/isaaclab/isaaclab/sim/simulation_cfg.py
jmart-nv: No longer applicable (RTX changes reverted)

source/isaaclab/test/sensors/test_tiled_camera_frame_stack.py
jmart-nv: Moved to top

source/isaaclab/test/sensors/test_tiled_camera_frame_stack.py
jmart-nv: Refactored to use `unittest.mock.patch` instead of `sys.argv`

source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
jmart-nv: No longer applicable (this now uses CameraCfg directly since TiledCameraCfg is deprecated)

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -610,6 +610,34 @@ def _configure_gym_env_spaces(self):
             if isinstance(self.cfg.state_space, type(MISSING)):
                 self.cfg.state_space = self.cfg.num_states
 
+        # Auto-adjust observation_space for camera frame stacking. A 3-tuple
+        # observation_space implies a single camera drives the policy obs; we collect
+        # every camera-like field with ``frame_stack > 1`` and verify they all share
+        # the same value before adjusting, rather than silently picking the first.
+        if isinstance(self.cfg.observation_space, (list, tuple)) and len(self.cfg.observation_space) == 3:
```
kellyguo11, line 617: I don't think this should belong in the base class. the stacking mechanism should be on a per-task implementation. I would add this specifically for the cartpole environment

source/isaaclab/isaaclab/physics/physics_manager_cfg.py
```diff
@@ -30,3 +30,13 @@ class PhysicsCfg:
 
     class_type: type[PhysicsManager] | Any = MISSING
     """The physics manager class to use. Must be set by subclasses."""
+
+    requires_temporal_camera_data: bool = False
```
kellyguo11, line 34: I think this probably also doesn't belong in the physics manager cfg. there's no reason the physics manager should care about camera data

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -347,6 +349,13 @@ def reset(self, env_ids: Sequence[int] | None = None, env_mask: wp.array | None
         self._update_poses(env_ids)
         # Reset the frame count
         self._frame[env_ids] = 0
+        # Reset frame-stack history for these envs; next _update_buffers_impl fills all
+        # history slots with the first post-reset frame so the policy never sees stale frames.
+        if self._frame_stack_size > 1:
```
kellyguo11, line 354: I would recommend handling the frame stacking logic in the cartpole environment directly

source/isaaclab_tasks/isaaclab_tasks/utils/sim_launcher.py
```diff
@@ -71,6 +71,63 @@ def _is_kitless_physics(node) -> bool:
     return isinstance(node, PhysicsCfg) and type(node).__name__ in ("NewtonCfg", "OvPhysxCfg")
 
 
+def _physics_needs_temporal_camera_data(node) -> bool:
```
kellyguo11, line 74: these are probably also not necessary if we move the logic into the task class directly?

ooctipus:
Good Job figuring this out @jmart-nv  I 100% agree with the motivation and surprised by your finding. Implementation wise, similar to Kelly's advise, I’d prefer this stay task-local for now. The current PR makes frame stacking a framework-level concern across the camera sensor, base env, physics config, renderer capabilities, presets, and launcher. That feels too broad for behavior currently motivated by specific Newton camera tasks.

In particular, I don’t think DirectRLEnv should auto-adjust observation_space from camera config fields. The task owns the observation layout, so it should explicitly stack frames and declare the resulting observation shape. If this becomes common across tasks, we can promote a shared helper later with clearer boundaries.

For example, for CartpoleCameraEnv, I’d rather see something like a task config field:
```
camera_frame_stack: int = 1
observation_space = [tiled_camera.height, tiled_camera.width, 3 * camera_frame_stack]
```
and then handle the history directly in CartpoleCameraEnv._get_observations():
```
def _stack_camera_obs(self, camera_data: torch.Tensor) -> torch.Tensor:
    frame_stack = self.cfg.camera_frame_stack
    if frame_stack <= 1:
        return camera_data
    if self._camera_frame_history is None:
        self._camera_frame_history = camera_data.unsqueeze(0).repeat(frame_stack, 1, 1, 1, 1)
        self._camera_frame_stack_idx = 0
    self._camera_frame_history[self._camera_frame_stack_idx].copy_(camera_data)
    stacked = torch.cat(
        [
            self._camera_frame_history[(self._camera_frame_stack_idx + 1 + i) % frame_stack]
            for i in range(frame_stack)
        ],
        dim=-1,
    )
    self._camera_frame_stack_idx = (self._camera_frame_stack_idx + 1) % frame_stack
    return stacked
```
Then _get_observations() can remain explicit:
```
camera_data = self._tiled_camera.data.output[data_type] / 255.0
camera_data -= torch.mean(camera_data, dim=(1, 2), keepdim=True)
camera_data = self._stack_camera_obs(camera_data)
return {"policy": camera_data.clone()}
```
Reset handling can also stay task-local by clearing or reinitializing the history for env_ids in _reset_idx(). That keeps the observation contract, memory cost, reset behavior, and Newton-specific policy choice visible in the environment that actually needs it.

I’d also document the Newton/Warp renderer class subtlety directly: unlike RTX modes with temporal AA, the Warp renderer does not inject prior-frame information into the current image. Camera-control tasks that need velocity-like visual cues should therefore add explicit temporal observations, such as task-local frame stacking, rather than relying on renderer-specific artifacts.

Feel free to give my comment to agent and and have it work on next iteration x)

jmart-nv:
Closing this PR in favor of per-task frame stacking

PR #5235, opened 2026-04-10
Adds async retarget to IsaacTeleopDevice (on by default)

hougantc-nvda:
# Description

Retargeting now runs asynchronously by default so work overlaps with env.step() instead of blocking the render path. This PR also hardens async scheduling by combining a consumption gate (strict 1:1 retarget-to-advance() cadence) with paced timing, and then refactors pacing into a pluggable estimator interface (EMA remains the default).

To disable async and use synchronous retargeting:

device.set_async(False)

Key changes:
Added AsyncRetargetLoop and wired IsaacTeleopDevice to use async retargeting by default.
Added consumption-gated pacing to prevent double-retargets and keep inputs fresh.
Refactored pacing into pluggable TimingEstimator / TimingEstimatorCfg with EmaTimingEstimator as default.

Fixes # (issue)

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
+++ b/source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-"""EMA-paced, consumption-gated background retarget loop for :class:`IsaacTeleopDevice`.
+"""Consumption-gated background retarget loop with pluggable timing estimation.
 
 The background thread produces exactly one retarget result per
 :meth:`consume` call.  After posting a result it waits for the main
@@ -11,15 +11,18 @@ thread to consume it before retargeting again, ensuring the retarget
 pipeline is invoked at the same cadence as ``advance()`` (just like the
 synchronous path).
 
-Within each cycle the thread uses an Exponential Moving Average (EMA)
-to predict *when* the next :meth:`consume` will happen, then delays
-reading inputs and retargeting until just before that deadline.  This
-maximises input freshness while still overlapping the retarget
-computation with ``env.step()``.
+Within each cycle a :class:`TimingEstimator` predicts *when* the next
+:meth:`consume` will happen and delays reading inputs until just before
+that deadline.  The default strategy uses Exponential Moving Average
+(EMA) via :class:`EmaTimingEstimator`.  To swap in a different
+prediction algorithm, subclass :class:`TimingEstimator` and
+:class:`TimingEstimatorCfg`, then pass the custom cfg to
+:class:`AsyncRetargetLoop`.
 """
 
 from __future__ import annotations
 
+import abc
 import logging
 import threading
 import time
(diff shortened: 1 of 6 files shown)


kellyguo11 (commented):
Nice work — well-structured async retarget with solid EMA pacing and thorough tests. The thread-safety contract (USD on main thread, retarget in background) is cleanly separated. A few issues below, two of which are real thread-safety concerns.

**Should-fix (3)**
1. `target_T_world` torch.Tensor not cloned in `update_inputs()` — breaks the isolation contract
2. `consume(block=True)` can deadlock if the background thread dies outside the max-failure path
3. `__exit__` permanently disables async mode even if user explicitly enabled it

**Suggestions (3)**
4. Document thread-safety requirement on `step_fn` parameter in `__init__` docstring
5. Consider resetting `_gen`/`_consumed_gen` in `start()` for clarity
6. Minor: double blank line in `_pace` docstring

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,269 @@
...
+            self._anchor_matrix = np.array(anchor_matrix, dtype=np.float32, copy=True)
+            if isinstance(target_T_world, np.ndarray):
+                target_T_world = np.array(target_T_world, dtype=np.float32, copy=True)
+            self._target_T_world = target_T_world
+
+    def consume(self, block: bool = True) -> tuple[torch.Tensor | None, bool]:
+        """Read the most recent retarget result.
+
+        Updates the EMA estimate of the consumer's call period so the
+        pacer can predict the next deadline.
```
kellyguo11: **should-fix: torch.Tensor not cloned — thread-safety hole**

The `np.ndarray` path correctly copies with `np.array(..., copy=True)`, but when `target_T_world` is a `torch.Tensor` it's stored by reference. The main thread can mutate the tensor between `update_inputs()` and the background thread's `_run()` reading it under `_input_lock`, violating the isolation contract documented in the class docstring.

```suggestion
            if isinstance(target_T_world, np.ndarray):
                target_T_world = np.array(target_T_world, dtype=np.float32, copy=True)
            elif isinstance(target_T_world, torch.Tensor):
                target_T_world = target_T_world.clone()
```

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,269 @@
...
+        Returns:
+            ``(action, is_fresh)`` where *is_fresh* is ``True`` when the
+            result has been updated since the last :meth:`consume`.
+
+        Raises:
+            Exception: Re-raises any exception that killed the background
+                thread, chained under a :class:`RuntimeError`.
+        """
+        if block:
+            with self._cond:
```
kellyguo11: **should-fix: `consume(block=True)` can deadlock on a dead thread**

The wait loop exits on `self._stop.is_set()` or a new generation, but if the background thread dies from an unexpected `BaseException` (e.g. `KeyboardInterrupt`, or a bug that bypasses the retry path), `_stop` is never set and `_gen` never increments. The 0.1 s timeout keeps the main thread responsive to `_stop`, but a dead thread won't set `_stop` either — so this spins forever in 0.1 s increments.

Cheap fix — add a thread-alive check in the wait loop:
```python
while not self._stop.is_set() and self._gen == self._consumed_gen:
    if self._thread is not None and not self._thread.is_alive():
        break
    self._cond.wait(timeout=0.1)
```

After breaking, raise a `RuntimeError("background thread died unexpectedly")` if `_gen == _consumed_gen` and no `_exc` is set.

source/isaaclab_teleop/isaaclab_teleop/isaac_teleop_device.py
```diff
@@ -165,6 +180,8 @@ def __enter__(self) -> IsaacTeleopDevice:
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Exit the context manager and clean up the IsaacTeleop session."""
```
kellyguo11, line 198: **should-fix: `__exit__` permanently disables async**

`self._async_enabled = False` here means if the user re-enters the context manager (e.g. retry loop), async mode is silently lost even though they explicitly enabled it (or it defaulted on). If re-entry isn't supported, a guard in `__enter__` would make that clear. If it is supported, this should restore the user's original preference rather than forcing sync mode.

Possible fix: only stop the loop here, don't mutate the preference flag — or save/restore it.

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,269 @@
...
+        # alpha=0.3 gives an EMA half-life of ~2 samples (ln2/ln(1/0.7)≈1.9).
+        # At a 45 Hz target frame rate that adapts within ~44 ms — fast
+        # enough to track frame-rate changes (e.g. VSync toggle, workload
+        # shifts) while still filtering single-frame timing jitter.
+        self._step_fn = step_fn
+
+        # Result state protected by a Condition rather than a bare Event to
+        # avoid a TOCTOU (Time-of-Check-to-Time-of-Use) race: with Event the
+        # consumer would check is_set() then separately read _latest, but
+        # the producer could overwrite _latest between those two steps.
```
kellyguo11: suggestion: The class docstring's "Thread-safety contract" section describes the *intended usage pattern*, but the `step_fn` parameter itself doesn't document that it will be called from a non-main thread. Adding a note to `__init__`'s Args section (e.g. "Called from a background thread; must not touch USD/Kit/Omniverse APIs directly") would make the requirement explicit at the API boundary.

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,269 @@
...
+        self._ema_alpha = ema_alpha
+        self._margin_s = margin_s
+        self._step_period = 0.022
+        self._retarget_dur = 0.020
+        self._last_consume = 0.0
+
+    def start(self) -> None:
+        """Start the retarget loop in a daemon thread."""
+        self._stop.clear()
+        self._thread = threading.Thread(target=self._run, daemon=True)
```
kellyguo11, line 303: suggestion: `start()` doesn't reset `_gen`/`_consumed_gen`, so after `stop()` + `start()` the counters carry over from the previous run. This works correctly because `consume(block=True)` waits for `_gen > _consumed_gen`, and the new background thread will increment `_gen`. But it's subtle — consider resetting both to 0 in `start()` (and clearing `_latest`/`_exc`) to make the fresh-start semantics explicit and avoid confusion if someone reads the code later.

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,269 @@
...
+                    torch.cuda.current_stream(action.device).synchronize()
+
+            with self._cond:
+                self._latest = action
+                self._gen += 1
+                self._cond.notify_all()
+
+    def _pace(self) -> None:
+        """Sleep until the optimal moment to begin the next retarget.
+
```
kellyguo11, line 448: Nit: double blank line in the `_pace` docstring body (PEP 257 / project style typically uses single blank lines within docstrings).

source/isaaclab_teleop/isaaclab_teleop/async_retarget_loop.py
```diff
@@ -0,0 +1,298 @@
...
+                action tensor or ``None``.  **Called from a background
+                thread**; must not touch USD/Kit/Omniverse APIs directly.
+            ema_alpha: EMA smoothing factor in ``(0, 1]``.
+            margin_s: Safety margin [s] subtracted from the predicted
+                deadline.
+        """
+        if not (0.0 < ema_alpha <= 1.0):
+            raise ValueError(f"ema_alpha must be in (0, 1], got {ema_alpha}")
+
+        # alpha=0.3 gives an EMA half-life of ~2 samples (ln2/ln(1/0.7)≈1.9).
```
nvddr: Can you split the timing estimation into a separate small class, so that ema can be swapped to something better later if needed?

source/isaaclab_teleop/isaaclab_teleop/isaac_teleop_device.py
```diff
@@ -226,17 +318,88 @@ def advance(self, target_T_world: np.ndarray | torch.Tensor | SupportsDLPack | N
...
+        First call runs synchronously to seed the cache and starts the
+        background loop only once the session is confirmed alive.
+        Subsequent calls push fresh inputs from the main thread and
+        consume the latest result, blocking for freshness when
+        ``self._async_blocking`` is set.
+
+        If the background thread died with an exception, it is re-raised
+        here on the main thread after cleaning up the loop.
+        """
+        if self._retarget_loop is None:
```
nvddr, line 360: Move into a separate function so this function is more readable

PR #5253, opened 2026-04-13
Fixes for the wrench composer

ClemensSchwarke:
# Description

The wrench composer currently has two bugs if I am not mistaken.

1. `cast_to_link_frame()` computes the position relative to a link frame's origin, but expressed in world frame with:
```
if is_global:
        return position - link_position
```
This position is then used in the cross product with the force from `cast_force_to_link_frame()` which is expressed in link frame:
```
if is_global:
        return wp.quat_rotate_inv(link_quat, position - link_position)
```
which is incorrect.

2. In `set_forces_and_torques_at_position()`, user-provided torques are overwritten when positions are provided.

Both issues are fixed in a separate commit.

## Type of change

### DIFF
--- a/source/isaaclab/isaaclab/utils/warp/kernels.py
+++ b/source/isaaclab/isaaclab/utils/warp/kernels.py
@@ -310,6 +310,10 @@ wp.overload(
 def cast_to_link_frame(position: wp.vec3f, link_position: wp.vec3f, is_global: bool) -> wp.vec3f:
     """Casts a position to the link frame of the body.
 
+    .. deprecated::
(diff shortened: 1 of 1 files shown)


AntoineRichard:
The bug is bigger than that. I have a fix here: https://github.com/gbionics/IsaacLab/pull/1 I'm double checkign some stuff for it

ClemensSchwarke:
@AntoineRichard I left a review on that PR :)

ClemensSchwarke:
Closing in favor of https://github.com/isaac-sim/IsaacLab/pull/5265

PR #5265, opened 2026-04-14
Port WrenchComposer dual-buffer fix to develop branch

AntoineRichard:
# Description

Replaces the single-buffer WrenchComposer with a dual-buffer architecture that stores global (world-frame) and local (body-frame) forces separately.

Follows develop's warp-first paradigm: all internal buffers are warp arrays, inputs accept both torch.Tensor and wp.array (warp ingests both natively), outputs are wp.array.

Updates PhysX and Newton asset write_data_to_sim to use add_raw_buffers_from
+ compose_to_body_frame instead of the old composed_force/composed_torque merge pattern.

## Type of change

- 
## Checklist


source/isaaclab/isaaclab/utils/warp/kernels.py
```diff
@@ -302,10 +302,240 @@ def reshape_tiled_image(
...
+                            positions[tid_env, tid_body], forces[tid_env, tid_body]
+                        )
+                    else:
+                        local_torque_b[tid_env, tid_body] = wp.cross(
+                            positions[tid_env, tid_body], forces[tid_env, tid_body]
+                        )
+
+
+@wp.kernel
+def add_forces_to_dual_buffers_mask(
```
mmichelis, line 448: Could set and add be combined, so set is just calling a zeroing func first then the add warp kernel? It feels like a lot of code duplication.

source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -44,61 +60,121 @@ def __init__(self, asset: BaseArticulation | BaseRigidObject | BaseRigidObjectCo
...
     @property
     def active(self) -> bool:
-        """Whether the wrench composer is active."""
+        """Whether the wrench composer is active (has pending forces/torques)."""
         return self._active
 
     @property
-    def composed_force(self) -> wp.array:
-        """Composed force at the body's link frame.
+    def global_only(self) -> bool:
```
mmichelis: The _has_global and _has_local parameters feel unnecessary, is global_only called somewhere for potential future optimizations? If so, should be clarified here why these flags are needed.

source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -184,65 +243,58 @@ def set_forces_and_torques_index(
...
+
         self._active = True
+        self._dirty = True
+        # set replaces everything, so flags match is_global exactly
+        self._has_global = is_global
+        self._has_local = not is_global
 
         wp.launch(
-            set_forces_and_torques_at_position_index,
+            set_forces_to_dual_buffers_index,
```
mmichelis, line 279: The zeroing is done anyways before, is there a reason not to call the add kernel and remove the set kernel entirely?

source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -357,91 +389,246 @@ def set_forces_and_torques_mask(
...
+            self._local_torque_b.zero_()
+            self._out_force_b.zero_()
+            self._out_torque_b.zero_()
             self._active = False
+            self._dirty = False
+            self._has_global = False
+            self._has_local = False
         elif env_mask is not None:
+            # Partial reset via mask: 7 buffers = 3.5 pairs, call 4 times
+            # Pair 1: global_force_w, global_torque_w
```
mmichelis: These 3.5 pairs of resetting should be done with a more general reset kernel. Either a generic one buffer at a time reset, or do all 7 together.

source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -27,13 +29,27 @@
...
-        This class is used to compose forces and torques at the body's link frame.
-        It can compose global wrenches and local wrenches. The result is always in the link frame of the body.
+        This class composes forces and torques applied to rigid bodies. Forces and torques can be
+        specified in either the global (world) frame or the local (body) frame. Internally, they are
+        stored in separate global and local input buffers. When the final composed wrench is needed,
+        the global contributions are rotated into the body frame and combined with the local
+        contributions to produce the output force and torque in the body's link frame.
+
+        The dual-buffer architecture uses five input buffers:
+            - global_force_w: Global forces applied at a position offset (world frame).
```
mmichelis: naming feels superfluous, force_w implies world frame, global could be left out?

mmichelis (commented):
Nice fix of permanent wrenches not using up-to-date link poses.

source/isaaclab/isaaclab/utils/wrench_composer.py
ClemensSchwarke: I agree, this can all be removed as it is not used anywhere

source/isaaclab/isaaclab/utils/wrench_composer.py
ClemensSchwarke: I think a set kernel makes sense, especially for user code

source/isaaclab/isaaclab/utils/wrench_composer.py
ClemensSchwarke: Not sure, just because it is expressed in world frame doesn't mean it is computed with respect to the world frames origin. I would personally leave it as is.

ClemensSchwarke:
1. I updated `global_force_at_com_w` to `global_force_at_link_w` and adapted the docstrings to mirror what the code is doing. 

2. I agree with @mmichelis that `_has_global` etc. can be pruned and that this zeroing kernel being called 4 times is not ideal :D Other than that it looks good to me.

3. One more comment I had was that including '_ensure_composed()` gives the impression that the output forces and torques are always "up to date", but it could still be that the frame transforms are super outdated which is currently not checked, which would lead to wrong outputs nevertheless. But I think this can be addressed later.

ClemensSchwarke:
The wrench composer now has the following behavior compared to PhysX:
- is_global=True and no positions -> Forces act at COM (in both)
- is_global=True and positions -> Positions are offsets from the world origin (in both)
- is_global=False and no positions -> Forces act at COM (in both)
- is_global=False and positions -> Positions are offsets from the link origin (PhysX), Positions are offsets from the CoM (Wrench Composer)

ClemensSchwarke:
Hey @AntoineRichard, I just checked the latest changes and it seems like the docstring updates changed things to the wrong version again. I put some time into formulating those to be as precise and well-understandable as possible, so I would consider keeping them. I am mainly referring to:
```
positions: The positions at which forces act. If `is_global` is True, these are global positions expressed
                in the world frame. If `is_global` is False, these are offsets from the body's CoM expressed in the
                body frame. If None, forces are assumed to act at the body's CoM, independent of the `is_global` flag.
                Shape: (len(env_ids), len(body_ids), 3). Defaults to None.
```
becoming
```
positions: Positions [m] at which forces act. If ``is_global`` is True, these are absolute
                world-frame coordinates of the force application point. If ``is_global`` is False, these
                are body-frame coordinates of the force application point (offset from the body frame
                origin). If None, forces are assumed to act at the body's CoM, independent of the
                ``is_global`` flag. Shape: (len(env_ids), len(body_ids), 3). Defaults to None.
```
in https://github.com/isaac-sim/IsaacLab/pull/5265/commits/29dc9f8c5c390d975611f5093c585def8a2d785b.

AntoineRichard:
> Hey @AntoineRichard, I just checked the latest changes and it seems like the docstring updates changed things to the wrong version again. I put some time into formulating those to be as precise and well-understandable as possible, so I would consider keeping them. I am mainly referring to:
> 
> ```
> positions: The positions at which forces act. If `is_global` is True, these are global positions expressed
>                 in the world frame. If `is_global` is False, these are offsets from the body's CoM expressed in the
>                 body frame. If None, forces are assumed to act at the body's CoM, independent of the `is_global` flag.
>                 Shape: (len(env_ids), len(body_ids), 3). Defaults to None.
> ```
> 
> becoming
> 
> ```
> positions: Positions [m] at which forces act. If ``is_global`` is True, these are absolute
>                 world-frame coordinates of the force application point. If ``is_global`` is False, these
>                 are body-frame coordinates of the force application point (offset from the body frame
>                 origin). If None, forces are assumed to act at the body's CoM, independent of the
>                 ``is_global`` flag. Shape: (len(env_ids), len(body_ids), 3). Defaults to None.
> ```
> 
> in [29dc9f8](https://github.com/isaac-sim/IsaacLab/commit/29dc9f8c5c390d975611f5093c585def8a2d785b).

Thanks Clemens, I'll revert it. Mike is going to take another look, and then we'll merge.

ClemensSchwarke:
Cool thanks :)

source/isaaclab/test/utils/test_wrench_composer_vs_physx.py
mmichelis, line 1: Great testing against physx!

mmichelis (approved):
Looks great to me, thanks for checking and updating the documentation for the is_global/position arguments!

ClemensSchwarke (approved):
Looks good :)

PR #5270, opened 2026-04-14
Fix RTX color rendering disabled for albedo and simple-shading cameras

AntoineRichard:
# Description

The TiledCamera-to-Camera merge (b1aba48) moved a disableColorRender optimization into IsaacRtxRenderer.create_render_data. The old TiledCamera never had this optimization, so albedo and simple-shading camera types now incorrectly have color rendering disabled, causing CUDA illegal memory access when the annotator returns invalid buffers.

Expand the needs_color_render check to include "albedo" and simple-shading data types that also require RTX color rendering.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
+++ b/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
@@ -89,7 +89,12 @@ class IsaacRtxRenderer(BaseRenderer):
         isaac_sim_version = get_isaac_sim_version()
 
         if isaac_sim_version.major >= 6:
-            needs_color_render = "rgb" in sensor.cfg.data_types or "rgba" in sensor.cfg.data_types
+            needs_color_render = (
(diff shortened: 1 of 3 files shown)


source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -89,7 +89,12 @@ def create_render_data(self, sensor: SensorBase) -> IsaacRtxRenderData:
         isaac_sim_version = get_isaac_sim_version()
 
         if isaac_sim_version.major >= 6:
-            needs_color_render = "rgb" in sensor.cfg.data_types or "rgba" in sensor.cfg.data_types
+            needs_color_render = (
```
kellyguo11, line 92: I think this flag should only apply to rgb and rgba? IIRC, this was the magic flag that enables the fast paths in albedo/simple shading modes so for those, we wanted to set this to False

PR #5203, opened 2026-04-08
Moves gravity compensation from schemas to MDP event

vidurv-nvidia:
# Description

Adds gravity compensation support for the Newton/MuJoCo backend as an MDP event, addressing review feedback from @ooctipus on #4847 to keep Newton-specific attributes out of the PhysX schema layer.

**New MDP event:** `set_gravity_compensation()` writes `mjc:gravcomp` (body-level) and `mjc:actuatorgravcomp` (joint-level) USD attributes at `prestartup` time. This follows the same pattern as existing USD-modifying events like `randomize_rigid_body_scale`.

**Bug fix:** `SolverMuJoCo.register_custom_attributes(builder)` was missing from `NewtonManager.instantiate_builder_from_stage()`, causing all `mjc:` prefixed custom attributes (gravcomp, actuatorgravcomp, etc.) to be silently ignored when parsing USD. The cloning path (`newton_replicate.py`) already had this call — it was only missing from the normal path.

**Usage example:**
```python
events = {
    "gravity_compensation": EventTermCfg(
        func=mdp.set_gravity_compensation,
        mode="prestartup",
        params={
            "asset_cfg": SceneEntityCfg("robot"),
            "body_gravity_compensation_scale": 1.0,          # uniform for all bodies
            # or per-body: {".*link[0-3]": 1.0, ".*link4": 0.5}
            "joint_gravity_compensation": [".*"],             # regex patterns for joints
        },
    ),
}
```

The pipeline: IsaacLab config → USD attributes → Newton ModelBuilder → MuJoCo solver / MJCF XML.

Supersedes #4847

## Type of change

- New feature (non-breaking change which adds functionality)
- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #4847, opened 2026-03-06
Enable gravity compensation in Isaaclab with Newton + Mujoco backend

vidurv-nvidia:
# Description

<!--
Thank you for your interest in sending a pull request. Please make sure to check the contribution guidelines.

Link: https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html

💡 Please try to keep PRs small and focused. Large PRs are harder to review and merge.
-->
Adds gravity compensation support for both rigid bodies (body-level) and joints (joint-level) when using the Newton/MuJoCo
   backend.                                                                                                                 
                                                                                                                            
  Body-level: A new gravity_compensation_scale field on RigidBodyPropertiesCfg writes the mjc:gravcomp attribute to rigid
  body USD prims, allowing Newton's MuJoCo solver to apply partial or full gravity compensation per body (0.0 = no          
  compensation, 1.0 = full).
                                                                                                                            
  Joint-level: A new gravity_compensation field on JointDrivePropertiesCfg writes mjc:actuatorgravcomp to joint USD prims   
  via modify_joint_drive_properties(). This is applied at spawn time through the schemas layer and flows automatically
  through FileCfg.joint_drive_props.                                                                                        
                                                            
  The pipeline is: IsaacLab config → USD attributes → Newton ModelBuilder → MuJoCo solver / MJCF XML.

## Type of change

<!-- As you go through the list, delete the ones that are not applicable. -->

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

<!--
As you go through the checklist above, you can mark something as done by putting an x character in it

For example,
- [x] I have done this task
- [ ] I have not done this task
-->

PR #5275, opened 2026-04-14
Refactor schema cfgs to separate solver-common from PhysX-specific fields

vidurv-nvidia:
# Description

Splits IsaacLab's USD-physics cfg classes into solver-common base classes and backend-specific subclasses, and refactors the writers (`modify_*_properties`, `spawn_rigid_body_material`) so that schema application is data-driven rather than hard-coded per-class. Prepares the schema layer for multi-backend support (PhysX today, Newton/Mjc next) without polluting base classes with silently-ignored fields or stamping backend-specific schemas onto prims that didn't opt in.

## Architecture

Two layered concepts:

1. **Per-declaring-class routing.** Each cfg field's USD namespace is determined by the class that declares it (walking the MRO). Base-class fields write under `physics:*`; subclass fields write under their own namespace (`physxRigidBody:*`, etc.). When a `PhysxRigidBodyPropertiesCfg` instance is written, base fields still go under `physics:*` because `_usd_namespace` is read from the declaring class via `__dict__`, not via `getattr` (which would hit the subclass override).
2. **Per-field exceptions.** Some "universal physics" fields have no USD path except through a backend-namespaced attribute today (e.g., `disable_gravity` only exists at `physxRigidBody:disableGravity`). These are declared as `_usd_field_exceptions = {applied_schema: (namespace, [fields...])}` on the base class; the writer applies the exception schema only when one of the listed fields is non-None.

The single helper `_apply_namespaced_schemas(prim, cfg, cfg_dict)` in `schemas.py` does both passes for every writer (rigid body, collision, articulation root, joint drive, mesh collision, rigid-body material).

## Design constraints

**One cfg class per spawner slot.** Spawners (`UsdFileCfg`, `MeshCuboidCfg`, etc.) carry a single field for each property group: `rigid_props: RigidBodyBaseCfg | None`, `collision_props: CollisionBaseCfg | None`, `joint_drive_props: JointDriveBaseCfg | None`, etc. The user cannot pass two cfgs into the same slot, so the cfg class hierarchy must be **single-rooted per spawner field** — one base class per group, with backend-specific subclasses below.

This rules out a "PhysX cfg sits next to a Newton cfg as siblings" design and drives several placement decisions:

| Constraint | Consequence |
|---|---|
| Universal-physics fields must be reachable from any backend's cfg | Goes on the **base** class, not a sibling backend cfg. Users on Newton-only deployments can use `RigidBodyBaseCfg(disable_gravity=True)` without importing `isaaclab_physx`. |
| A PhysX-namespaced field whose semantics are universal (e.g., `disable_gravity`) | Lives on the base but routes to the PhysX namespace via `_usd_field_exceptions`. The base stays backend-clean; the writer dispatches the PhysX write only when the field is non-None. |
| Writer logic must not branch on cfg subclass | Every writer is the same code path regardless of subclass. The cfg metadata (`_usd_namespace`, `_usd_applied_schema`, `_usd_field_exceptions`) drives behavior; the writer is a pure data interpreter. |
| Adding a new backend (Newton, Mjc) | Requires a new subclass with its own `_usd_namespace` / `_usd_applied_schema`. No spawner-side changes, no writer-side changes, no base-cfg-side changes. |
| A field has multiple USD paths today (one PhysX-namespaced, one Newton-namespaced) | Belongs on the **PhysX subclass**, not the base. A future `NewtonArticulationRootPropertiesCfg` will own the same conceptual field on the Newton side. ("Rule 2" — e.g., `enabled_self_collisions`.) |
| A field has only one USD path today, namespaced under PhysX, but the conceptual quantity is universal | Belongs on the **base** with an `_usd_field_exceptions` entry. ("Rule 1" — e.g., `disable_gravity`, `articulation_enabled`, `contact_offset`, `rest_offset`, `max_joint_velocity`.) When Newton ships its own native attribute, the exception namespace switches transparently with no API change. |

### Base (solver-common) classes — `physics:*` namespace via `UsdPhysics.*API`

| Cfg class | Field | USD attribute |
|---|---|---|
| `RigidBodyBaseCfg` | `rigid_body_enabled` | `physics:rigidBodyEnabled` |
| `RigidBodyBaseCfg` | `kinematic_enabled` | `physics:kinematicEnabled` |
| `CollisionBaseCfg` | `collision_enabled` | `physics:collisionEnabled` |
| `MassPropertiesCfg` | `mass` | `physics:mass` |
| `MassPropertiesCfg` | `density` | `physics:density` |
| `RigidBodyMaterialBaseCfg` | `static_friction` | `physics:staticFriction` |
| `RigidBodyMaterialBaseCfg` | `dynamic_friction` | `physics:dynamicFriction` |
| `RigidBodyMaterialBaseCfg` | `restitution` | `physics:restitution` |
| `JointDriveBaseCfg` | `drive_type` | `drive:<axis>:physics:type` |
| `JointDriveBaseCfg` | `max_force` | `drive:<axis>:physics:maxForce` |
| `JointDriveBaseCfg` | `stiffness` | `drive:<axis>:physics:stiffness` |
| `JointDriveBaseCfg` | `damping` | `drive:<axis>:physics:damping` |
| `MeshCollisionBaseCfg` | `mesh_approximation_name` | `physics:approximation` (token) |
| `ArticulationRootBaseCfg` | `fix_root_link` | (synthesizes `UsdPhysics.FixedJoint`) |

`JointDriveBaseCfg` and `MeshCollisionBaseCfg` use the typed `UsdPhysics.DriveAPI` / `UsdPhysics.MeshCollisionAPI` accessors at the writer level (multi-instance namespace and `TfToken` with `allowedTokens`, respectively); all other base fields flow through the helper's per-class routing.

### PhysX subclasses — `physx*:*` namespaces, `Physx*API` schemas

| Cfg class | `_usd_namespace` | `_usd_applied_schema` | Adds fields |
|---|---|---|---|
| `PhysxRigidBodyPropertiesCfg` | `physxRigidBody` | `PhysxRigidBodyAPI` | `linear_damping`, `angular_damping`, `max_linear_velocity`, `max_angular_velocity`, `max_depenetration_velocity`, `max_contact_impulse`, `enable_gyroscopic_forces`, `retain_accelerations`, solver iter counts, sleep / stabilization thresholds |
| `PhysxCollisionPropertiesCfg` | `physxCollision` | `PhysxCollisionAPI` | `torsional_patch_radius`, `min_torsional_patch_radius` |
| `PhysxArticulationRootPropertiesCfg` | `physxArticulation` | `PhysxArticulationAPI` | `enabled_self_collisions`, solver iter counts, sleep / stabilization thresholds |
| `PhysxJointDrivePropertiesCfg` | `physxJoint` | `PhysxJointAPI` | (currently empty; reserved for future PhysX-only knobs) |
| `PhysxRigidBodyMaterialCfg` | `physxMaterial` | `PhysxMaterialAPI` | `compliant_contact_stiffness`, `compliant_contact_damping`, `friction_combine_mode`, `restitution_combine_mode` |
| `PhysxConvexHullPropertiesCfg` | `physxConvexHullCollision` | `PhysxConvexHullCollisionAPI` | `hull_vertex_limit`, `min_thickness` |
| `PhysxConvexDecompositionPropertiesCfg` | `physxConvexDecompositionCollision` | `PhysxConvexDecompositionCollisionAPI` | hull / voxel / shrink-wrap tunables |
| `PhysxTriangleMeshPropertiesCfg` | `physxTriangleMeshCollision` | `PhysxTriangleMeshCollisionAPI` | `weld_tolerance` |
| `PhysxTriangleMeshSimplificationPropertiesCfg` | `physxTriangleMeshSimplificationCollision` | `PhysxTriangleMeshSimplificationCollisionAPI` | `simplification_metric`, `weld_tolerance` |
| `PhysxSDFMeshPropertiesCfg` | `physxSDFMeshCollision` | `PhysxSDFMeshCollisionAPI` | `sdf_margin`, `sdf_narrow_band_thickness`, `sdf_resolution`, etc. |

### `_usd_field_exceptions` table

These fields are declared on a *base* class but the only USD path today goes through a non-base namespace. Each entry says: "if any listed field on this cfg is non-None, apply the exception schema and write that one attribute under the exception namespace." All other fields on the cfg follow the per-declaring-class routing rule.

| Base cfg class | Exception schema | Namespace | Field(s) | Why on the base |
|---|---|---|---|---|
| `RigidBodyBaseCfg` | `PhysxRigidBodyAPI` | `physxRigidBody` | `disable_gravity` | Per-body gravity exclusion is universal physics; PhysX honors per-body, Newton consumes the same attribute via the bridge resolver (scene-level today; per-body fix is a Newton-side kernel change, not a cfg-API change) |
| `CollisionBaseCfg` | `PhysxCollisionAPI` | `physxCollision` | `contact_offset`, `rest_offset` | Collision-pair generation distance and rest gap are universal physics; Newton importer consumes both via PhysX bridge to populate `Model.shape_collision_radius` / `_thickness` (`import_usd.py:2104, 2111`) |
| `ArticulationRootBaseCfg` | `PhysxArticulationAPI` | `physxArticulation` | `articulation_enabled` | PhysX honors at sim time; IsaacLab Newton wrapper reads it as a spawn-time guard at `rigid_object.py:1035`. Universal user-facing intent |
| `JointDriveBaseCfg` | `PhysxJointAPI` | `physxJoint` | `max_joint_velocity` | Sole USD path to `Model.joint_velocity_limit` in Newton (no `newton:*` equivalent today). The exception namespace switches transparently when Newton ships `newton:maxJointVelocity` as a registered applied API |

When any exception field is non-None, the corresponding `Physx*API` schema is applied to the prim. When all exception fields are None, no PhysX schema is stamped — Newton-targeted prims authored from `*BaseCfg` stay free of PhysX schemas they didn't opt in to.

## Field renames (with deprecation aliases)

To enforce the convention that python `snake_case` cfg field names map identity-style to USD `camelCase` attribute names, two legacy fields were renamed. Both keep the old name as a deprecation alias forwarded via `__post_init__` (emits `DeprecationWarning`, scheduled for removal in 5.0).

| Old name | New name | USD attribute |
|---|---|---|
| `JointDriveBaseCfg.max_velocity` | `max_joint_velocity` | `physxJoint:maxJointVelocity` |
| `JointDriveBaseCfg.max_effort` | `max_force` | `drive:<axis>:physics:maxForce` |

## Type of change

The split is non-breaking at the spawner-cfg level — every base-class type accepts any subclass via polymorphism, and every legacy `RigidBodyPropertiesCfg` / `JointDrivePropertiesCfg` / `CollisionPropertiesCfg` / `ArticulationRootPropertiesCfg` / `MeshCollisionPropertiesCfg` / `RigidBodyMaterialCfg` / `FixedTendonPropertiesCfg` / `SpatialTendonPropertiesCfg` import path continues to work via deprecation-alias subclasses and `__getattr__` shims on `isaaclab.sim`, `isaaclab.sim.schemas`, and `isaaclab.sim.schemas.schemas_cfg`. Direct attribute access to the renamed fields still works through deprecation aliases. Removal scheduled for 5.0.

The breaking aspect: cfg classes in `isaaclab_physx.sim.schemas` and `isaaclab_physx.sim.spawners.materials` are physically relocated. Anyone importing from internal paths (rather than `isaaclab.sim`) needs to update.

## Migration

```python
# Before
rigid_props = sim_utils.RigidBodyPropertiesCfg(disable_gravity=True, linear_damping=0.1)
joint_props = sim_utils.JointDrivePropertiesCfg(max_effort=80.0, max_velocity=5.0)
collision_props = sim_utils.CollisionPropertiesCfg(contact_offset=0.02, torsional_patch_radius=1.0)
material = sim_utils.RigidBodyMaterialCfg(static_friction=0.7, compliant_contact_stiffness=1000.0)

# After (PhysX-targeted)
from isaaclab_physx.sim.schemas import (
    PhysxRigidBodyPropertiesCfg,
    PhysxJointDrivePropertiesCfg,
    PhysxCollisionPropertiesCfg,
)
from isaaclab_physx.sim.spawners.materials import PhysxRigidBodyMaterialCfg

rigid_props = PhysxRigidBodyPropertiesCfg(disable_gravity=True, linear_damping=0.1)
joint_props = PhysxJointDrivePropertiesCfg(max_force=80.0, max_joint_velocity=5.0)
collision_props = PhysxCollisionPropertiesCfg(contact_offset=0.02, torsional_patch_radius=1.0)
material = PhysxRigidBodyMaterialCfg(static_friction=0.7, compliant_contact_stiffness=1000.0)

# After (Newton-targeted — base classes only, no PhysX schemas applied)
from isaaclab.sim.schemas import RigidBodyBaseCfg, JointDriveBaseCfg, CollisionBaseCfg
from isaaclab.sim.spawners.materials import RigidBodyMaterialBaseCfg

rigid_props = RigidBodyBaseCfg(disable_gravity=True)  # only base + exception fields available
joint_props = JointDriveBaseCfg(max_force=80.0, max_joint_velocity=5.0)
material = RigidBodyMaterialBaseCfg(static_friction=0.7)
```

Spawner type annotations remain unchanged — they accept any subclass via polymorphism.

## Internal helper

```python
def _apply_namespaced_schemas(prim, cfg, cfg_dict):
    # 1. Per-field exceptions: pop listed fields, apply exception schema if any non-None,
    #    write under exception namespace.
    # 2. Per-declaring-class routing: walk MRO to find each remaining field's owner class;
    #    write under that class's _usd_namespace; apply that class's _usd_applied_schema.
```

Used by all five `modify_*_properties` writers and `spawn_rigid_body_material`. Replaced ~125 lines of duplicated gating logic with a single ~30-line helper.

## Side change: configclass

`source/isaaclab/isaaclab/utils/configclass.py:_process_mutable_types` now detects string-form `ClassVar` annotations under PEP 563 (`from __future__ import annotations`) so it doesn't wrap `ClassVar[dict]` defaults in `field(default_factory=...)`. Matches Python stdlib `dataclasses` semantics. No pre-existing IsaacLab class used `ClassVar` inside a `@configclass` block, so the change has no effect on existing code; it enables the `ClassVar` metadata pattern this PR introduces.

## Test plan

- [x] `test_schemas.py` (38 → 40 tests): all schema-cfg classes write correct attributes under the right namespace; PhysX schemas are NOT applied when only base/UsdPhysics fields are set; deprecation aliases (`max_velocity` → `max_joint_velocity`, `max_effort` → `max_force`) forward correctly and emit `DeprecationWarning`. **40 passed.**
- [x] `test_schemas_shim.py`: legacy import paths (`isaaclab.sim.schemas.RigidBodyPropertiesCfg` etc.) resolve via `__getattr__` shims. **All passing.**
- [x] `test_articulation.py`, `test_rigid_object_iface.py`, `test_valid_configs.py`, `test_spawn_*` — no regressions.
- [x] Full suite (`./isaaclab.sh -t`): 8768/9205 pass, 437 unrelated baseline failures (rendering, `omni.physics.tensors.api` missing, OSC controller, `install_ci`, `pyglet`, Newton env-path, Anymal-C determinism). Zero new regressions; +123 passing tests vs. earlier state.
- [x] `./isaaclab.sh -f` (pre-commit) clean.

## Supersedes

Together with #5276, supersedes #4847 and #5203 with a cleaner schema-layer design.

## Checklist

- [x] I have made corresponding changes to the documentation (changelog fragments under `source/isaaclab/changelog.d/`)
- [x] I have updated the changelog (fragment-based system) and the corresponding version in the extension's `config/extension.toml` file


source/isaaclab/isaaclab/sim/schemas/schemas_cfg.py
```diff
@@ -116,6 +144,26 @@ class RigidBodyPropertiesCfg:
     """The mass-normalized kinetic energy threshold below which an actor may participate in stabilization."""
 
 
+@configclass
+class RigidBodyPropertiesCfg(PhysxRigidBodyPropertiesCfg):
+    """Deprecated: use :class:`PhysxRigidBodyPropertiesCfg` or :class:`RigidBodyBaseCfg`.
+
+    .. deprecated:: 4.6.2
```
kellyguo11, line 151: let's use 3.0.0 for the version

kellyguo11 (commented):
how would this change look like for the asset cfgs? would we need different physx and newton variations of the assets? or somehow pick between the cfgs based on the engine used?

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -318,22 +316,34 @@ def modify_rigid_body_properties(
         return False
     # retrieve the USD rigid-body api
     usd_rigid_body_api = UsdPhysics.RigidBodyAPI(rigid_body_prim)
-    # ensure PhysX rigid body API is applied
-    applied_schemas = rigid_body_prim.GetAppliedSchemas()
-    if "PhysxRigidBodyAPI" not in applied_schemas:
-        rigid_body_prim.AddAppliedSchema("PhysxRigidBodyAPI")
```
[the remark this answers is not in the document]
AntoineRichard: This should only be ran once.

AntoineRichard (approved):
I think the PR overall is nice, It's great that it work as is without touching any other files. One concern is how user friendly this will be in the long run but at least it's making things explicit.

ooctipus:
maybe baseCfg lives in core, Physx related schema lives in isaaclab_physx?

vidurv-nvidia:
bot do a full review

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -46,21 +45,140 @@
...
+    ]
+)
+
+
+"""
+Schema-application helper.
+"""
+
+
+def _apply_namespaced_schemas(prim, cfg, cfg_dict: dict) -> None:
```
vidurv-nvidia, line 132: Crux of the reworked spawner system. The schema and namespace is now dictated by the class. This function automatically detects the schema and the namespace the parameter belong to and applies it to the asset. This also has a small section on exceptions. The base class carries a dictionary of exceptions - parameters that do not belong to the base class but are present because of an architectural reasons detailed in the exceptional parameters' docstring. Would be able to use team's expertise to vet those exceptions in each base cfg class.

ooctipus (commented):
Did a careful walk through the placement layer and the writer/helper with a agent. Most of it lands cleanly against the consumption-gated rule — the per-field audit comes out matching for ~16 of the 18 cases I checked, and the "don't stamp PhysX schemas onto Newton-targeted prims" invariant has solid test coverage (`test_*_no_physx_schema_when_*`). Two placement calls worth confirming, and a handful of framework-shape ideas that I think would make the cfg layer even nicer to live with.

## Two placement calls

**`disable_gravity` on `RigidBodyBaseCfg`.** The strict reading of the consumption-gated rule would push this to the PhysX subclass. Newton's importer reads `physxRigidBody:disableGravity` from the scene prim (`newton/_src/utils/import_usd.py:1655`, `prim_type=PrimType.SCENE`) and uses it to drive scene-wide gravity — so a Newton user who writes `RigidBodyBaseCfg(disable_gravity=True)` per-body either no-ops or globally toggles, depending on which prim the importer happens to consult. The field's docstring is honest about this, which I genuinely appreciate, but the user-facing API will mislead Newton users today. Two options: keep on the base with the documented caveat, or move to `PhysxRigidBodyPropertiesCfg` until Newton ships per-body kernel support ("not yet supported on Newton" is a cleaner failure mode than "supported but lies"). I lean toward keeping it where it is, but would like a sanity check from someone with Newton-kernel context.

**`enabled_self_collisions` on `PhysxArticulationRootPropertiesCfg`.** This is a divergence from the rule as written. Newton has a native `newton:selfCollisionEnabled` (resolver checks it first) and consumes it per-articulation at `import_usd.py:2294-2299`. The PR introduces a separately stated "Rule 2" — when both backends have native namespaces, place on the backend-specific subclass and let a future `NewtonArticulationRootPropertiesCfg` carry the Newton side. That framing is internally consistent and forward-looking, but it's a real reframing, and it sets a precedent for future dual-namespace fields. Worth a moment to confirm this is the intended evolution.

## Framework-shape ideas

None of these are blockers — the PR works. But the cfg layer carries some machinery I think we can lean out, and the deprecation boilerplate is repeating itself enough that I think it deserves a small framework lift.

### 1. Move routing onto the field via `Annotated`

*Intent:* the cfg body should read like a schema definition, not a mix of class-level metadata and fields. Per-field exceptions live on the field, the class default lives in a decorator above the class, and the body becomes 100% field declarations. Field renames stop desyncing from a separate string-keyed registry, and the MRO walk in `_apply_namespaced_schemas` (a workaround for "subclass `_usd_namespace` shouldn't apply to base fields") goes away.

**Before** (`source/isaaclab/isaaclab/sim/schemas/schemas_cfg.py:1325`):
```python
class RigidBodyBaseCfg:
    _usd_namespace: ClassVar[str | None] = "physics"
    _usd_applied_schema: ClassVar[str | None] = None
    _usd_field_exceptions: ClassVar[dict] = {
        "PhysxRigidBodyAPI": ("physxRigidBody", ["disable_gravity"]),
    }
    rigid_body_enabled: bool | None = None
    kinematic_enabled: bool | None = None
    disable_gravity: bool | None = None
```

**After:**
```python
@usd_route(namespace="physics")
@configclass
class RigidBodyBaseCfg:
    rigid_body_enabled: bool | None = None
    kinematic_enabled: bool | None = None
    disable_gravity: Annotated[
        bool | None,
        Route("physxRigidBody", "PhysxRigidBodyAPI"),
    ] = None
```

PhysX subclasses compose the same way — decorator override for the subclass's own fields, inherited base routing untouched:

```python
@usd_route(namespace="physxRigidBody", applied_schema="PhysxRigidBodyAPI")
@configclass
class PhysxRigidBodyPropertiesCfg(RigidBodyBaseCfg):
    linear_damping: float | None = None
    angular_damping: float | None = None
    # ...
```

The decorator is two lines:
```python
def usd_route(*, namespace: str | None = None, applied_schema: str | None = None):
    def _apply(cls):
        cls._usd_class_route = Route(namespace, applied_schema)
        return cls
    return _apply
```

Resolved via `get_type_hints(cls, include_extras=True)`, cached on the class. The triplet (`_usd_namespace`, `_usd_applied_schema`, `_usd_field_exceptions`) collapses to one decorator + per-field `Annotated` markers; `_usd_field_exceptions` disappears entirely. Classes that route nothing at the class level (e.g. `JointDriveBaseCfg` — every field is either typed-`UsdPhysics.DriveAPI` or `Annotated`-routed) just omit the decorator. Same `Annotated` machinery powers idea #3, and `@usd_route` composes cleanly with the `@deprecated_subclass` factory in idea #2.

### 2. Hoist class-deprecation boilerplate out of cfg files

*Intent:* cfg classes go back to being pure data declarations, no `def` in user-facing class bodies. Single source of truth for the deprecation list lives next to the writer that consumes it. Side benefit: messages don't drift (the current PR has both "scheduled for removal in 5.0" and "and will be removed in 5.0" wordings).

**Before** (`source/isaaclab_physx/isaaclab_physx/sim/schemas/schemas_cfg.py:3585`, ×14):
```python
@configclass
class RigidBodyPropertiesCfg(PhysxRigidBodyPropertiesCfg):
    """Deprecated: ..."""
    def __post_init__(self):
        warnings.warn(
            "'RigidBodyPropertiesCfg' is deprecated and will be removed in 5.0. Use ...",
            DeprecationWarning, stacklevel=2,
        )
        super().__post_init__()
```

**After** — empty cfg + registry next to the writer:
```python
class RigidBodyPropertiesCfg(PhysxRigidBodyPropertiesCfg):
    pass
```
```python
# in schemas.py
DEPRECATED_CFGS: dict[type, tuple[str, str]] = {
    RigidBodyPropertiesCfg: ("PhysxRigidBodyPropertiesCfg | RigidBodyBaseCfg", "5.0"),
    JointDrivePropertiesCfg: ("PhysxJointDrivePropertiesCfg | JointDriveBaseCfg", "5.0"),
    # ... 12 more
}

def _warn_if_deprecated(cfg) -> None:
    entry = DEPRECATED_CFGS.get(type(cfg))
    if entry is None:
        return
    repl, removed = entry
    warnings.warn(
        f"{type(cfg).__name__!r} is deprecated and will be removed in {removed}. Use {repl} instead.",
        DeprecationWarning, stacklevel=3,
    )
```

Each writer calls `_warn_if_deprecated(cfg)` at the top. Warnings fire on use rather than instantiation, which means config-loading-at-import stays quiet, and `simplefilter("once")` dedupes naturally. ~110 lines of repeated boilerplate become ~14 dict entries.

### 3. Field aliases on `JointDriveBaseCfg` — same trick

*Intent:* the only `__post_init__` on a non-deprecated cfg, expressed as field-level data so the cfg has zero `def`. The alias rule self-documents next to the field it aliases.

**Before** (`source/isaaclab/isaaclab/sim/schemas/schemas_cfg.py:1519`):
```python
@configclass
class JointDriveBaseCfg:
    def __post_init__(self):
        _deprecate_field_alias(self, "max_velocity", "max_joint_velocity")
        _deprecate_field_alias(self, "max_effort", "max_force")
    max_force: float | None = None
    max_effort: float | None = None
    max_joint_velocity: float | None = None
    max_velocity: float | None = None
```

**After:**
```python
@configclass
class JointDriveBaseCfg:
    max_force: float | None = None
    max_effort: Annotated[float | None, DeprecatedField(canonical="max_force")] = None
    max_joint_velocity: float | None = None
    max_velocity: Annotated[float | None, DeprecatedField(canonical="max_joint_velocity")] = None
```

A `_resolve_field_aliases(cfg)` helper in `schemas.py` walks the markers, warns, and forwards old → new before the routing pass. Same `Annotated` infrastructure as #1.

### 4. Ask the schema registry for the Sdf type instead of guessing

*Intent:* the schema already knows the right wire type — just ask it. Removes the Python-value isinstance ladder and incidentally fixes the latent TfToken bug for combine-mode fields. Pre-existing on main, so the PR doesn't introduce it, but this feels like the right moment.

**Before** (`source/isaaclab/isaaclab/sim/utils/prims.py:355`):
```python
if isinstance(value, bool):    sdf_type = Sdf.ValueTypeNames.Bool
elif isinstance(value, int):   sdf_type = Sdf.ValueTypeNames.Int
elif isinstance(value, float): sdf_type = Sdf.ValueTypeNames.Float
elif isinstance(value, str):   sdf_type = Sdf.ValueTypeNames.String  # ← combine modes land here
```

`physxMaterial:frictionCombineMode` and `restitutionCombineMode` are TfTokens with `allowedTokens`; the `Literal[str]` falls into the String branch.

**After:**
```python
@functools.cache
def _attr_type(applied_schema: str, attr: str) -> Sdf.ValueTypeName:
    spec = Usd.SchemaRegistry().FindAppliedAPIPrimDefinition(applied_schema)
    return spec.GetSchemaAttributeSpec(attr).typeName
```

Cached per (schema, attr). Future schemas with float3, asset paths, etc. work without expanding the ladder.

## Small things alongside

- `MeshCollisionBaseCfg.__getattr__` for `usd_api` / `physx_api` (`schemas_cfg.py:1711`): the dispatch that consumed these is gone. Consider deleting the shim — `AttributeError` with a docstring pointer to `_usd_applied_schema` reads cleaner than a working-but-deprecated accessor that reconstructs a string. Also removes the `self.__dict__.get(...)` ClassVar lookup that only works because `_custom_post_init` deepcopies ClassVars into instance dicts — fragile to read without configclass internals in mind.
- `MeshCollisionBaseCfg._usd_attr_name_map: ClassVar[dict] = {}` (`schemas_cfg.py:1677`) is declared but never read.
- Worth one explicit `Changed` changelog line: the new `PhysxRigidBodyMaterialCfg` defaults `compliant_contact_*` / `*_combine_mode` to `None`, where the legacy class had concrete defaults. Probably no observable change because PhysX engine defaults match, but the shift from "always authored" to "authored only when user sets it" is meaningful enough to call out.
- PR description's `PhysxRigidBodyMaterialCfg` field table mentions `improve_patch_friction`, but I don't see the field in the code.
- `cfg_dict = {k: v for k, v in cfg.to_dict().items() if not k.startswith("_")}` appears in 6 writers — replacing with a `dataclasses.fields(cfg)` walk drops the underscore filter and the recursive `to_dict()` deepcopy. Pairs with #1.

If only one of the framework lifts lands, I'd push for #1 — it's the foundation that makes #3 cheap and removes the rename hazard structurally. #2 is the largest readability win at ~110 lines of cfg-file boilerplate.

Overall this is solid work — the consumption-gating tests are exactly the invariant I'd want to see covered, and the architecture is honest about its trade-offs in docstrings. Happy to see it merge once the two placement calls land.

vidurv-nvidia:
`disable_gravity` and `enabled_self_collisions` placements are intentional. Per-body gravity exclusion is universal physics — Newton consumes the same attribute today (scene-level partial honor, documented on the field) and the per-body fix is a Newton kernel change, not a cfg-API one. Moving it now means a rename later. `enabled_self_collisions` is the dual-namespace case: Newton has its own native attribute, future `NewtonArticulationRootPropertiesCfg` owns the `newton:*` side. Either on the base would lock the writer into one namespace forever.

On #1 — `Annotated` isn't used anywhere in the repo today (no `typing.Annotated` imports, no `get_type_hints(include_extras=True)` callers). Adopting here introduces a routing dialect that diverges from every other configclass in the codebase. Worth doing as a project-wide convention with a separate RFC, not a local cleanup in this PR.

#3 — same concern as #1: same `Annotated` mechanism for two fields scheduled for removal in 5.0. Not enough lifetime to justify introducing the pattern just for them.

#4 — agree, right call. Pre-existing on main and fixes the latent TfToken bug for `*_combine_mode` fields independently of this PR. Will file as a separate issue.

The cfg_dict → `dataclasses.fields(cfg)` swap landed in e0312ccf252 along with the material-defaults changelog line.

ooctipus (approved):

PR #5248, opened 2026-04-13
[Newton] Add Rough terrain locomotion Part 1: Foundation + quadrupeds on Newton

hujc7:
# Description

Enables Newton rough-terrain locomotion training on all locomotion velocity envs (Go1, Go2, A1, Anymal-B/C/D, H1, Cassie, Digit, G1) on top of [@ooctipus](https://github.com/ooctipus)'s Anymal-D foundation work, cherry-picked from PR #5225.

## Why this PR exists

PR #5225 (Octi's draft) added Newton support for Anymal-D rough terrain. The other 9 locomotion envs were left out of scope. Octi is away and his PR has CI failures, so per maintainer guidance ([Kelly Guo's comment](https://github.com/isaac-sim/IsaacLab/pull/5225#issuecomment-2120849317)) the required changes from #5225 are cherry-picked here so the rough-terrain stack can move forward without depending on his still-open WIP PR.

## Dependencies

- PR #5365 — adds `isaaclab.utils.checked_apply`, used by `NewtonManager.create_builder` to forward `NewtonShapeCfg` onto Newton's upstream `ShapeConfig`. Required for stable rough-terrain contact.

## 1. Cherry-pick from #5225 (`614ea2dbb74`)

Commit authored by Octi Zhang with `Co-authored-by` trailer. Subset of #5225 — what's kept and dropped:

| # | Item | Status | Reason |
|---:|---|---|---|
| 1 | `anymal_d/rough_env_cfg.py` Anymal-D Newton config | **KEPT** (then hoisted into shared parent in commit 2 below) | Defines the Newton physics shape used for rough terrain |
| 2 | `velocity_env_cfg.py` — hoist `physics_material`, `add_base_mass`, `base_com` startup events into shared `EventsCfg` | **KEPT** | All envs need them |
| 3 | `base_com` guard `preset(default=..., newton=None)` | **KEPT** | Ablation A4 on #5225 (posted 2026-04-17): without the gate, 99.99% of episodes terminate from `body_lin_vel` runaway on Newton. Upstream Newton fix newton-physics/newton#2332 will let us drop the guard once it ships |
[...]

PR #4883, opened 2026-03-09
Updates contact sensor API to use Newton new API that allows fast initialization.

ooctipus:
# Description

- This PR add contact sensor test and test utilities to newton contacts sensors 

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #4761, opened 2026-02-27
Adds newton engine

ooctipus:
# Description

This PR adds newton engine to isaaclab together with backend agnostic syntax
this pr happens after lazy import
Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5289, opened 2026-04-16
Fixes default shape margin field name in Newton manager

hujc7:
## Summary

Fix incorrect attribute name `contact_margin` → `gap` on Newton `ShapeConfig` in `NewtonManager.create_builder()`.

Newton PR #1732 renamed `contact_margin` to `gap` (broad-phase AABB expansion). The IsaacLab code was never updated, creating a dead attribute that Python silently accepted. The intended 1 cm default shape gap was never applied.

## Newton PR #1732 rename mapping

| Old field | New field | Semantics |
|-----------|-----------|-----------|
| `thickness` | `margin` | Shape surface extension (affects contact forces) |
| `contact_margin` | **`gap`** | Broad-phase AABB expansion (detection range only) |
| `rigid_contact_margin` | `rigid_gap` | Builder-level default gap (already 0.1) |

## Timeline

| Date | Newton | IsaacLab | `contact_margin` valid? |
|------|--------|----------|------------------------|
| Feb 19 | pin: `51ce35e` (has `contact_margin`) | #4646 adds `contact_margin = 0.01` | Yes |
| Feb 24 | **PR #1732 renames `contact_margin` → `gap`** | — | — |
| Mar 2 | pin updated to `v0.2.3` (after rename) | #4761 keeps `contact_margin` | **No — broken** |
| Mar 9 | — | #4883 removes the line | Removed |
| Apr 13 | — | #4822 re-adds `contact_margin` | **No — still broken** |

## Ablation: gap vs margin

We conducted an ablation study to understand the impact. Note: `margin` (shape surface
extension) is a different field from `gap` (broad-phase range). The original code intended
to set `gap`, but setting `margin` also has a significant positive effect on training for
rough terrain locomotion.

| Setting | `gap` | `margin` | Go1 Reward (300 iter) | Effect |
|---------|-------|----------|----------------------|--------|
| Dead field (baseline) | 0.1 (default) | 0.0 | ~1.0 | — |
| `gap=0.01` only | 0.01 | 0.0 | 0.66 | No training improvement |
| `margin=0.01` only | 0.1 (default) | 0.01 | 18.77 | Major improvement |
| `gap=0.01` + `margin=0.01` | 0.01 | 0.01 | 16.54 | Similar to margin-only |

**This PR fixes the correct field migration** (`contact_margin` → `gap`). The `margin` setting
for rough terrain contact quality will be addressed separately in the rough terrain env PR
(#5248) via a new `default_shape_margin` config field on `NewtonCfg`.

## Test plan

- [x] Verified `contact_margin` is not a field on `ShapeConfig` (Newton 1.1.0.dev0)
- [x] Verified `gap` is the correct replacement field per Newton PR #1732
- [x] Confirmed by camevor (Newton developer)
- [x] Ablation study: gap alone doesn't change training, so existing tests should pass

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -426,10 +426,10 @@ class NewtonManager(PhysicsManager):
             **kwargs: Forwarded to :class:`ModelBuilder`.
 
         Returns:
-            New builder with up-axis and contact margin defaults applied.
+            New builder with up-axis and gap defaults applied.
         """
         builder = ModelBuilder(up_axis=up_axis or cls._up_axis, **kwargs)
-        builder.default_shape_cfg.contact_margin = 0.01
+        builder.default_shape_cfg.gap = 0.01
         return builder
 
     @classmethod
--- a/source/isaaclab_newton/config/extension.toml
+++ b/source/isaaclab_newton/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.5.15"
+version = "0.5.16"
 
 # Description
 title = "Newton simulation interfaces for IsaacLab core package"
--- a/source/isaaclab_newton/docs/CHANGELOG.rst
+++ b/source/isaaclab_newton/docs/CHANGELOG.rst
@@ -1,6 +1,19 @@
 Changelog
 ---------
 
+0.5.16 (2026-04-17)
+~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
+* Fixed incorrect attribute name ``contact_margin`` on Newton
+  ``ShapeConfig`` in
+  :meth:`~isaaclab_newton.physics.NewtonManager.create_builder`. The
+  field was renamed to ``gap`` in Newton PR #1732. The typo created a
+  dead attribute so the intended 1 cm default shape gap was never applied.
+
+
 0.5.15 (2026-04-16)
 ~~~~~~~~~~~~~~~~~~~
 


AntoineRichard (approved):
LGTM, thanks @hujc7! As mentioned, make a separate PR for testing these fields!

AntoineRichard:
@camevor do you know why the newton tests fails? It looks like the force reported are larger than expected when the contact margin are large?

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -422,7 +422,7 @@ def create_builder(cls, up_axis: str | None = None, **kwargs) -> ModelBuilder:
             New builder with up-axis and contact margin defaults applied.
         """
         builder = ModelBuilder(up_axis=up_axis or cls._up_axis, **kwargs)
-        builder.default_shape_cfg.contact_margin = 0.01
+        builder.default_shape_cfg.margin = 0.01
```
camevor: `ShapeConfig.margin` is not the same thing as `ShapeConfig.contact_margin` used to be. You want to set `ShapeConfig.gap` or `ModelBuilder.rigid_gap` (fallback)  instead.

camevor:
Please note that this change does not migrate the former `contact_margin` correctly to `gap`, because the semantics of these terms has been changed in Newton. Please take a look at the [Shape Configuration Documentation](https://newton-physics.github.io/newton/stable/concepts/collisions.html#shape-configuration).

camevor:
> @camevor do you know why the newton tests fails? It looks like the force reported are larger than expected when the contact margin are large?

This is likely a test stability issue: The contact sensor is inherently quite noisy, so we should add some averaging for these quantitative tests. I'll prepare a PR with averaging to make the tests less flaky. It's also worth considering adding this to the sensor itself (e.g. averaging over solver substeps).

hujc7:
Thanks, @camevor ! Running a test with updated field name. It did previous improve training so I will also investigate the impact of setting margin. Will check back

hujc7:
gap alone does not fix rough env but should still remain the right fix for the intended field.

PR #5296, opened 2026-04-16
Antoiner/feat/torch array

AntoineRichard:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


StafaH:
Might be interesting to also add the automatic unwarp for warp kernels as well? Something like below. Might needs some solid tests (cuda vs cpu kernels, make sure they get passed in, no functional or performance difference between passing array vs array.warp.

```
   # ------------------------------------------------------------------
    # Warp kernel interop
    # ------------------------------------------------------------------
    #
    # Expose the array-interface protocols from the underlying warp array so that
    # a TorchArray can be passed directly as an input to :func:`warp.launch`.
    # Warp's ``pack_arg`` routine looks for these attributes when the runtime
    # type does not match the expected :class:`warp.array` and builds an
    # ``array_t`` descriptor from the interface (keeping ``self`` alive via its
    # ``owner`` reference). Delegating to ``self._warp`` reuses the cached
    # interface dict that warp already maintains per array, so there is no extra
    # allocation on each launch.
    #
    # AttributeError is propagated (not caught) when the device doesn't match
    # the protocol — warp's ``pack_arg`` relies on the ``AttributeError`` to
    # fall through to the next candidate protocol.

    @property
    def __cuda_array_interface__(self):
        """Delegate the CUDA array interface to the underlying warp array.

        Raises:
            AttributeError: If the underlying warp array is not on a CUDA device.
        """
        return self._warp.__cuda_array_interface__

    @property
    def __array_interface__(self):
        """Delegate the NumPy array interface to the underlying warp array.

        Raises:
            AttributeError: If the underlying warp array is not on a CPU device.
        """
        return self._warp.__array_interface__

```

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -787,10 +788,10 @@ def joint_pos(self) -> wp.array:
             # read data from simulation and set the buffer data and timestamp
             self._joint_pos.data = self._root_view.get_dof_positions()
             self._joint_pos.timestamp = self._sim_timestamp
-        return self._joint_pos.data
+        return TorchArray(self._joint_pos.data)
```
hougantc-nvda, line 791: Profiling the physx side, it doesn't appear the caching in working as intended.
All these TorchArray constructors are causing wp.to_torch to be called on every call to these functions since __init__ sets self._torch_cache: to None
Including the overhead of TorchArray it's actually a slight bit slower.

PR #5304, opened 2026-04-17
ProxyArray and Asset/Sensor level property caching

AntoineRichard:
# Description

Follows the TorchArray concept from MJLab and augment it with some lab requirements.

## Type of change

## Checklist


source/isaaclab/isaaclab/utils/warp/__init__.pyi
```diff
@@ -4,10 +4,12 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 __all__ = [
+    "TorchArray",
```
camevor: Does `TorchArray` need to be exported publicly?

docs/source/how-to/torch_array.rst
```diff
@@ -0,0 +1,127 @@
...
+  and vice versa.
+
+- ``.warp`` returns the original :class:`warp.array`. Use this when you need to pass data to
+  warp kernels explicitly or access warp-specific attributes (``ptr``, ``strides``, etc.).
+
+
+Deprecation Bridge
+~~~~~~~~~~~~~~~~~~
+
+To ease migration, ``TorchArray`` includes a deprecation bridge that allows existing code to
```
camevor: This is a nice feature; however, I wonder if it is necessary or if we can avoid the associated complexity (and avoid any new code being written against it despite the deprecation warning).

source/isaaclab/isaaclab/utils/warp/torch_array.py
```diff
@@ -0,0 +1,292 @@
...
+                " If you have a TorchArray, use it directly instead of wrapping it again."
+            )
+        self._warp = wp_array
+        self._torch_cache: torch.Tensor | None = None
+
+    # ------------------------------------------------------------------
+    # Core accessors
+    # ------------------------------------------------------------------
+
+    def rebind(self, wp_array: wp.array) -> None:
```
camevor: I can see a few issues coming out of this caching layer introduced by `TorchArray`:
1. Failing to `rebind()` on each instance after memory has been rotated will lead to invalid memory access.
2. It is fair to assume that users will expect type and shape continuity across rebind operations (that they are unaware of), so this contract should be specified and validated.

camevor (commented):
Nice convenience and performance improvements! There are some remaining questions regarding the caching `TorchArray` introduces. It might be simpler to expose torch-specific attributes on data containers, or make `TorchArray` a pure bundle of a warp array and torch array.

source/isaaclab/isaaclab/utils/warp/torch_array.py
```diff
@@ -0,0 +1,292 @@
...
+        """Rebind this wrapper to a new warp array, invalidating the torch cache.
+
+        This is needed when the underlying simulation memory is re-created (e.g. after
+        a full simulation reset) and the old warp array pointer becomes stale.
+
+        Args:
+            wp_array: The new warp array to wrap.
+        """
+        self._warp = wp_array
+        self._torch_cache = None
```
camevor: Abandoning the cached torch tensor means that of the following patterns, one is deprecated and one will break upon rebind:

```python
# init
a = sensor.data.val
b = sensor.data.val.torch
# rebind
a.mean()   # deprecated
b.mean()  # invalid memory access
a.torch.mean()  # preferred
```

Updating the torch array could improve the issue with `b`. But the more straightforward and cleaner pattern would probably just be having separate torch attributes (which could be auto-generated, and cache based on array identity):

```python
sensor.data.val_torch.mean()
```

AntoineRichard: Remove the rebind, and create a new torch array instead of calling rebind

source/isaaclab/isaaclab/utils/warp/torch_array.py
```diff
@@ -0,0 +1,292 @@
...
+import torch
+import warp as wp
+
+
+class TorchArray:
+    """Warp-first array wrapper providing cached zero-copy ``.torch`` and ``.warp`` accessors.
+
+    This class wraps a :class:`warp.array` and provides:
+
+    * A ``.warp`` property that returns the original warp array (for kernel interop).
```
AntoineRichard: raise on set

source/isaaclab/isaaclab/utils/warp/torch_array.py
```diff
@@ -0,0 +1,292 @@
...
+from __future__ import annotations
+
+import warnings
+from typing import ClassVar
+
+import torch
+import warp as wp
+
+
+class TorchArray:
```
AntoineRichard: rename

source/isaaclab/isaaclab/utils/warp/torch_array.py
AntoineRichard: Removed

source/isaaclab/isaaclab/utils/warp/torch_array.py
AntoineRichard: removed

source/isaaclab/isaaclab/utils/warp/__init__.pyi
AntoineRichard: I think so.

docs/source/how-to/torch_array.rst
AntoineRichard: Agreed, there are massibe user friendlyness benefits here. I don't like it either.

kellyguo11 (approved):

PR #5349, opened 2026-04-21
Add render flag to environment step() API

kellyguo11:
Expose a render: bool = True parameter on the step() method of all environment base classes (ManagerBasedEnv, ManagerBasedRLEnv, DirectRLEnv, DirectMARLEnv) and the MARL utility wrappers.

When render=False is passed, all rendering calls are skipped:
- GUI / RTX sensor renders inside the decimation loop
- Post-reset re-renders for RTX sensors

Physics simulation continues normally regardless of the flag. This allows user workflows that do not need rendering every step (e.g. headless RL training, fast rollouts) to opt out per-step.

Also adds unit tests for render=False and mixed render flag stepping.

# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/direct_rl_env.py
+++ b/source/isaaclab/isaaclab/envs/direct_rl_env.py
@@ -223,7 +223,12 @@ class DirectRLEnv(gym.Env):
         # initialize data and constants
         # -- counter for simulation steps
         self._sim_step_counter = 0
-        # -- whether step() performs rendering (set to False to skip all render calls)
+        # -- controls camera/Kit rendering in step().
+        # When False, the Kit app loop (app.update()) and camera/RTX sensor updates are
+        # skipped, but standalone visualizers (Newton, Rerun, Viser) continue to update.
+        # This is because Kit bundles camera rendering with its app loop and the two
+        # cannot be separated.  Non-Kit visualizers have independent step() methods
+        # that do not trigger camera or GUI updates, so they remain active.
         self.render_enabled: bool = True
         # -- counter for curriculum
         self.common_step_counter = 0
@@ -389,6 +394,16 @@ class DirectRLEnv(gym.Env):
 
         Rendering can be controlled per-step via :attr:`render_enabled`.
 
+        When ``render_enabled`` is False:
+
+        - The Kit app loop (``app.update()``) is **skipped**, which also disables
+          camera/RTX sensor rendering and GUI viewport updates.  Kit bundles these
+          operations together, so they cannot be separated.
+        - Standalone visualizers (Newton, Rerun, Viser) **continue to update**
+          normally because their ``step()`` methods are independent of the Kit
+          app loop.
+        - Post-reset re-renders for RTX sensors are also skipped.
+
         Args:
             action: The actions to apply on the environment. Shape is (num_envs, action_dim).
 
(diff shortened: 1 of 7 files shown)


Toni-SM:
@kellyguo11 IMO, the implementation breaks the agreed standard API for representing general RL problems ([gymnasium.Env.step](https://gymnasium.farama.org/main/api/env/#gymnasium.Env.step)) forcing any reinforcement library (already integrated, or with the intention of supporting Isaac Lab) to change their implementation and include a custom case to support this feature for Isaac Lab only. 

If such behavior (skipping rendering calls) is desired, it should be implemented using other mechanisms (e.g., config passed to the class constructor) that don't modify the agreed standard API ([gymnasium.Env](https://gymnasium.farama.org/main/api/env/#gymnasium.Env)).

kellyguo11:
> @kellyguo11 IMO, the implementation breaks the agreed standard API for representing general RL problems ([gymnasium.Env.step](https://gymnasium.farama.org/main/api/env/#gymnasium.Env.step)) forcing any reinforcement library (already integrated, or with the intention of supporting Isaac Lab) to change their implementation and include a custom case to support this feature for Isaac Lab only.
> 
> If such behavior (skipping rendering calls) is desired, it should be implemented using other mechanisms (e.g., config passed to the class constructor) that don't modify the agreed standard API ([gymnasium.Env](https://gymnasium.farama.org/main/api/env/#gymnasium.Env)).

Yes, I left it with a default so that it's not a requirement for wrappers following gymnasium APIs to set this, but it does diverge from the gymnasium Env API. alternatively, we can make this an internal variable, similar to your config idea, that can be set by external calls. We want this flag to be dynamic for each reset call.

xyao-nv (approved):

xyao-nv:
Thanks for making the change!

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -434,7 +438,7 @@ def step(self, action: torch.Tensor) -> VecEnvStepReturn:
         if len(reset_env_ids) > 0:
             self._reset_idx(reset_env_ids)
             # if sensors are added to the scene, make sure we render to reflect changes in reset
-            if self.has_rtx_sensors and self.cfg.num_rerenders_on_reset > 0:
+            if is_rendering and self.has_rtx_sensors and self.cfg.num_rerenders_on_reset > 0:
```
matthewtrepte: hm from a design point of view, if render_enabled=False, do we always want to skip updating visualizers? since visualizers are stepped in self.sim.render()

source/isaaclab/test/envs/test_env_rendering_logic.py
```diff
@@ -239,3 +239,167 @@ def wrapped_step(dt):
...
+            original_step(dt)
+            render_cb(render_dt)
+
+        viz.step = wrapped_step
+
+        # create a zero action tensor for stepping the environment
+        actions = torch.zeros((env.num_envs, 0), device=env.device)
+
+        # Step with render_enabled=False for several steps
+        env.render_enabled = False
```
matthewtrepte, line 309: hm if theres a very short episode env we can test with (<10 episode length), then the step path can cover the post reset case

matthewtrepte (commented):
few comments

source/isaaclab/isaaclab/envs/direct_rl_env.py
kellyguo11: good point, enabled visualizer stepping

source/isaaclab/test/envs/test_env_rendering_logic.py
kellyguo11: addressed

PR #5358, opened 2026-04-22
Fixes compatibility issues with latest Isaac Sim updates

kellyguo11:
# Description

Fixes two regressions introduced by recent Isaac Sim `develop` builds (between 2026-04-13 and 2026-04-22) that broke the standard training entry point:

```bash
./isaaclab.sh -p scripts/reinforcement_learning/rsl_rl/train.py --task=Isaac-Cartpole-v0
```

### 1. Isaac Sim `SimulationManager` invalidating the shared physics view

A recent Isaac Sim commit (`8df6beeb0`, 6.0.0-alpha.180) decorated `SimulationManager._on_stop` / `_on_play` / `_on_stage_*` with `@staticmethod` so they finally fire from their Carb subscriptions. The newly-working `_on_stop` calls `SimulationManager.invalidate_physics()`, which invalidates the shared `omni.physics.tensors` simulation view that `PhysxManager` and every articulation `_root_view` rely on, raising:

```
[Error] [omni.physx.tensors.plugin] Simulation view object is invalidated and cannot be used again to call getDofVelocities
Exception: Failed to get DOF velocities from backend
```

on the first `scene.update()` after `sim.reset()`.

- `source/isaaclab_physx/isaaclab_physx/__init__.py`: `_patch_isaacsim_simulation_manager()` now (a) force-imports `isaacsim.core.simulation_manager` so the patch always runs deterministically, (b) calls `original_class.enable_all_default_callbacks(False)` to tear down the original class's Carb subscriptions, and (c) only then swaps the module attribute. `PhysxManager` becomes the single owner of the simulation view's lifecycle.

### 2. PhysX tensors API module relocation

Recent Isaac Sim builds removed `omni.physics.tensors.impl` and now expose the PhysX Tensor API types directly under `omni.physics.tensors.api`. The old import raises `ModuleNotFoundError` at Isaac Lab import time, taking down every entry point that touches articulations, rigid objects, deformables, contact sensors, or ray caster sensors.

- Updated all imports of the form `import omni.physics.tensors.impl.api as physx` → `import omni.physics.tensors.api as physx` across `isaaclab` (ray-caster sensors) and `isaaclab_physx` (assets and contact sensor), including `TYPE_CHECKING`-gated imports.
- Updated the `PhysX Tensor API` Sphinx cross-reference URLs in the articulation data classes and the example console output in the contact sensor / IMU docs.
- Added a migration note to `docs/source/migration/migrating_to_isaaclab_3-0.rst`.

Fixes # (issue)

## Type of change

## Screenshots

N/A — runtime fixes with no UI surface.

## Checklist


source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -462,6 +462,13 @@ def _update_buffers_impl(self, env_mask: wp.array):
         self._renderer.update_transforms()
         self._renderer.render(self._render_data)
 
+        # Synchronize warp's CUDA stream before handing GPU buffers to torch.
+        # render() launches warp kernels that write into the output tensors;
+        # without an explicit sync, torch may read the buffers on a different
+        # CUDA stream before the warp kernels finish, causing illegal-memory-
+        # access errors (observed with newer Isaac Sim nightly images).
+        wp.synchronize()
```
pbarejko: My suggestion would be to push to the back end that needs it.

kellyguo11: reverted synchronize calls from this PR

preeves-nv:
Hi @kellyguo11 @pbarejko 

While looking at the CI failures on this PR I - with Claude, traced the root cause of some of the failures  to _patch_isaacsim_simulation_manager() in isaaclab_physx/__init__.py. The call to enable_all_default_callbacks(False) disables all four Carb subscriptions including warm_start (PLAY). The warm_start callback is required to initialise the rendering pipeline on sim.reset() — without it, tiled-camera RGB output stays black and visualization-marker tests fail.

Does this make sense to you - I am not familiar enough with the whole code base to be sure - but I thought it worth trying to investigate? 

Only _default_callback_on_stop needs to be unsubscribed — it's the one that calls invalidate_physics() and wrecks the shared omni.physics.tensors view. The other three callbacks, including warm_start, should be left intact.

Looking at the isaaclab (core) [1/3] CI log from this PR, the specific failures consistent with this are:

test_first_frame_textured_rendering (0/1)
test_tiled_camera (6/8)
test_multi_tiled_camera (6/11)
test_visualization_markers (5/6)
The fix is a single-file change to isaaclab_physx/__init__.py:


original_class = getattr(original_module, "SimulationManager", None)
if original_class is not None and original_class is not PhysxManager:
    if hasattr(original_class, "_default_callback_on_stop"):
        # Carb subscription objects unsubscribe on destruction — setting to
        # None drops the reference and silently cancels the subscription.
        original_class._default_callback_on_stop = None
    else:
        # Fallback for older Isaac Sim builds where the attribute doesn't exist.
        # Note: this may cause tiled-camera black frames; the targeted fix above is preferred.
        try:
            original_class.enable_all_default_callbacks(False)
        except Exception:
            pass
The commit is cherry-pickable from preeves-nv:preeves/fix-on-stop-targeted (379915312) — also referenced in #5371. Hap

PR #5376, opened 2026-04-23
Modifies ROOT user related setup to Dockerfile.curobo

njawale42:
# Description

Refactors `docker/Dockerfile.curobo` so the editable Isaac Lab installs and the final container runtime drop privileges to a non-root **`isaaclab`** user (uid/gid 1000) instead of staying as root the whole way. The cuRobo CI image (`build-curobo` → `test-curobo` / `test-skillgen`) currently runs as root, with `OMNI_KIT_ALLOW_ROOT=1` papering over the upstream Isaac Sim base image's intent of running non-root.

## What changed

1. **Explicit `USER root` for the system-level setup block.** apt, cuda-keyring repo, `install_deps.py apt`, `${DOCKER_USER_HOME}` cache dirs, `/bin/nvidia-*` placeholders, the pre-bundled torch removal, and the `get-pip.py` bootstrap all require root and stay as root.

2. **New `isaaclab` user at uid/gid 1000.** Created via explicit `groupadd --non-unique --gid 1000` followed by `useradd --non-unique --uid 1000 --gid 1000 isaaclab`. The `--non-unique` flag on both is required because some base image revisions already provision a user/group at numeric ID 1000 (e.g. an `ubuntu`/`kit`-style account); the file system only cares about numeric IDs, so two names mapping to the same ID is fine. Pinning the GID explicitly via `groupadd` (instead of letting `useradd` auto-pick one when the default GID is taken) keeps the BuildKit pip cache mount's `gid=1000` matched to the user's primary group. UID 1000 is chosen so that the runtime user matches the typical CI runner host user that owns the bind-mounted `/workspace/isaaclab` — required for the test runner's `mkdir tests` etc. to not fail with `EACCES` on the bind-mount overlay.

3. **`chown` + `chmod` + `USER isaaclab` before the pip installs.**
   - The paths the new user needs to write to are chowned: `${ISAACLAB_PATH}`, `${ISAACSIM_ROOT_PATH}/kit/python/.../site-packages`, `${DOCKER_USER_HOME}`, `/home/isaaclab`. `/isaac-sim` itself stays mode 755 and isaac-sim-owned by the base image — re-chowning it would force an OverlayFS copy-up of the multi-GB Isaac Sim install.
   - `chmod 755 ${ISAACSIM_ROOT_PATH}` opens up traversal of `/isaac-sim` for non-owner users. The base image creates `/isaac-sim` as `isaac-sim`'s home directory; on recent Ubuntu (`HOME_MODE 0700` in `/etc/login.defs`), that means mode 0700 isaac-sim:isaac-sim, which our isaaclab user cannot enter to find `python.sh` or the rest of the Kit runtime. Inner files retain their original perms (NVIDIA ships them world-readable per standard distribution conventions), so only the top-level directory's mode needs to change — single-inode metadata change, no recursive layer cost.

4. **BuildKit cache mount pinned to `uid=1000,gid=1000`** for `${DOCKER_USER_HOME}/.cache/pip`, otherwise the default root-owned cache mount would be unwritable from `isaaclab` (BuildKit doesn't consult `/etc/passwd` at mount time, so the numeric form is required).

5. **`${HOME}/.bashrc` used consistently** in the alias `RUN` block. Two lines hardcoded `/root/.bashrc` while the surrounding eight used `${HOME}/.bashrc`; after the user switch this inconsistency would silently land the trailing config in the wrong file.

6. **Final `COPY` annotated with `--chown=isaaclab:isaaclab`** for ownership consistency.

CI build args (`DOCKER_USER_HOME_ARG=/root` in `.github/actions/ecr-build-push-pull/action.yml`, `.github/actions/docker-build/action.yml`, and `.github/workflows/postmerge-ci.yml`) are unchanged — `/root` is chowned to `isaaclab` inside the image, so they continue to work.

## Type of change
- CI/CD fix (non-breaking change which fixes an issue)

## Checklist


njawale42:
@greptileai

PR #5395, opened 2026-04-24
perf(buffers): remove GPU syncs from CircularBuffer/DelayBuffer hot path

haixuanTao:
# Human written summary

Did some performance check on WBC-Agile, and notice 10% fps increase when doing the below changes in simulation.

This should not change any of the simulator logic.

# AI Generated Summary

``CircularBuffer.append`` and ``CircularBuffer.__getitem__`` each used ``torch.any(...)`` to detect post-reset state, forcing a GPU→CPU synchronization on every call. For consumers that invoke the delay buffer once per physics step (``DelayedPDActuator``, ``RemotizedPDActuator``, ``DelayedDCMotor``, observation-history buffers in ``observation_manager``), this produces two syncs per call and measurably starves the GPU at high ``num_envs``.

Replace the ``torch.any()`` probes with a CPU-side ``_any_first_push_pending`` flag maintained by ``reset()`` (or buffer lazy-init) and cleared by ``append()`` once the first-push replication runs. The public API is unchanged:

- ``_num_pushes`` stays a per-env ``long`` GPU tensor; the ``current_length`` property still returns ``min(num_pushes, max_len)`` per batch index.
- First-push replication still fills the entire history for any env whose ``num_pushes`` is zero at append time. The replication is now written via ``torch.where`` with a broadcast mask so the scatter is dynamic-shape-free and cannot reintroduce a sync through advanced-indexing shape inference.
- ``__getitem__`` still raises ``RuntimeError`` on a read from an unwritten env — just gated on the CPU flag instead of a GPU probe.
- All other error/shape invariants preserved.

``DelayBuffer.compute`` also drops a redundant ``.clone()`` on the returned gather. ``CircularBuffer.__getitem__`` uses advanced indexing (``self._buffer[index_in_buffer, self._ALL_INDICES]``) which already allocates fresh storage, so callers that mutate the return value in place do not affect the internal ring. Saves one ``(batch, *feat)`` D2D copy per call.

Microbenchmark on CUDA (RTX 5090, ``num_envs=6144, features=12, history_length=10``, resetting ~10% of envs every 50 steps):

    upstream DelayBuffer.compute : 619 µs/call
    patched  DelayBuffer.compute : 354 µs/call  (1.75x)

Per actuator step with three delay buffers: ~265 µs/call × 3 ≈ 800 µs saved per physics tick.

Validated against the existing ``test_circular_buffer.py`` (10 cases) and ``test_delay_buffer.py`` (3 cases) suites. End-to-end verified by re-running a 6144-env locomotion training job on the patched tree and confirming the wandb FPS plateau matches a prior run that used a hand-rolled sync-free sibling of ``DelayBuffer``.

# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -74,6 +74,7 @@ Guidelines for modifications:
 * Gary Lvov
 * Giulio Romualdi
 * Grzegorz Malczyk
+* Haixuan Xavier Tao
 * Haoran Zhou
 * Harsh Patel
 * HoJin Jeon
--- a/source/isaaclab/config/extension.toml
+++ b/source/isaaclab/config/extension.toml
@@ -1,7 +1,7 @@
 [package]
 
 # Note: Semantic Versioning is used: https://semver.org/
-version = "0.54.3"
+version = "0.54.4"
 
 # Description
 title = "Isaac Lab framework for Robot Learning"
--- a/source/isaaclab/docs/CHANGELOG.rst
+++ b/source/isaaclab/docs/CHANGELOG.rst
@@ -1,6 +1,26 @@
 Changelog
 ---------
 
+0.54.4 (2026-04-24)
+~~~~~~~~~~~~~~~~~~~
+
+Changed
+^^^^^^^
+
+* Removed per-call GPU→CPU synchronizations from
+  :class:`~isaaclab.utils.buffers.CircularBuffer` by replacing the
+  ``torch.any(...)`` probes in ``append`` and ``__getitem__`` with a
+  CPU-side flag maintained by :meth:`reset`. Also removed a redundant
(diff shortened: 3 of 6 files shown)


Mayankm96:
Similar to the PR: https://github.com/isaac-sim/IsaacLab/pull/4275

ooctipus:
Thanks @haixuanTao — this PR called the problem correctly, and nearly all of it has since landed on `develop` through the frame-stacking refactor (#5574, #5849): the CPU-side flag exists as `_need_reset`, `__getitem__` no longer probes the device, and `DelayBuffer.compute` no longer clones. The one remaining sync this PR targeted — the masked first-push backfill in `append` — is removed in #6471 (branchless `torch.where`, with benchmarks), which also carries over this PR's empty-`batch_ids` reset guard. Closing as superseded; the direction here was right and is now fully covered.

PR #5297, opened 2026-04-16
Decouple Renderer from Camera

r-schmitt:
# Description

The renderer was previously held by the camera.  this PR refactors this relationship, creating a RenderContext as a property of the SimulationContext to manage the renderer's lifetime independent of the camera(s)

## Type of change

- refactor

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5398, opened 2026-04-24
Patch newton model reqs

matthewtrepte:
# Description

Patch newton model reqs issue where in some caes env configs were constructed in a way where newton model reqs were not correctly determined, leading to downstream issues.

Also, revert the revert to physx scene data provider, since the fix above no longer requires the revert, which is non ideal since building newton model from usd fallback is slow.

## Type of change

## Screenshots

## Checklist


source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -133,6 +134,30 @@ def __init__(self, cfg: CameraCfg):
         self._renderer: BaseRenderer | None = None
         self._render_data = None
 
+    def _register_renderer_scene_data_requirements(self) -> None:
```
kellyguo11, line 137: @r-schmitt @nvsekkin @pbarejko would this conflict with the refactoring in https://github.com/isaac-sim/IsaacLab/pull/5297?

r-schmitt: im not sure that this explicitly conflicts, i dont see anything that would break here.  if these are scene data requirements for a specific sensor/camera then it would make sense to call it from camera.py, but maybe the function would make more sense living in the new render_context module in my PR

nvsekkin: I don’t see a blocker with #5297 this probably does not belong in renderer instance creation anyways. so may be okay to leave as is in `camera.py`

nvsekkin: to @r-schmitt's point may belong to `render_context`. The only risk is timing, but we can get around this by calling a `RenderContext` method from `Camera` `__init__`

kellyguo11 (approved):

source/isaaclab/isaaclab/sensors/camera/camera.py
kellyguo11: thanks! @matthewtrepte pls keep an eye on #5297 and see if we should move this to `render_context` once those changes land

PR #5399, opened 2026-04-25
Fixes ActionTerm.set_debug_vis crash in kitless mode

kellyguo11:
## Description

`ActionTerm.set_debug_vis` crashes with `NameError: name 'omni' is not defined` when running in Kit-based headless mode. This affects any environment with `debug_vis=True` on an action term (e.g. `Isaac-Navigation-Flat-Anymal-C` via `PreTrainedPolicyAction`).

**Root cause:** Two issues compound:

1. The top-level `import omni.kit.app` is guarded by `if has_kit():`, but `has_kit()` returns `False` at module import time (before Kit is initialized in `sys.modules`), so the import is skipped.
2. Later at runtime, `set_debug_vis` calls `omni.kit.app.get_app_interface()` unconditionally — `omni` is not in scope because the import was skipped.

```
NameError: name 'omni' is not defined
  File "action_manager.py", line 138, in set_debug_vis
    app_interface = omni.kit.app.get_app_interface()
```

## Fix

Two changes to `ActionTerm.set_debug_vis`:

1. Add `has_kit()` guard to the `debug_vis` check, matching the existing pattern in `CommandTerm.set_debug_vis` (`command_manager.py:112`).
2. Add a **local** `import omni.kit.app` inside the guarded block, so the import resolves correctly even when the top-level conditional import was skipped at module load time.

```python
# Before
if debug_vis:
    if self._debug_vis_handle is None:
        app_interface = omni.kit.app.get_app_interface()

# After
if debug_vis and has_kit():
    if self._debug_vis_handle is None:
        import omni.kit.app

        app_interface = omni.kit.app.get_app_interface()
```

**Tested:** `Isaac-Navigation-Flat-Anymal-C-v0` now trains successfully in headless mode (1500 iterations, previously crashed immediately).

## Type of change

## Checklist


kellyguo11:
@ooctipus could you give this a review?

Issue #2529, opened 2025-05-20
[Question] run_diff_ik.py can't work without reset the arm to the default pose


nolann030628:
### Question

I want to write an IK function basd on the tutorial, and I don't want the pos to go back to the default pose so I comment the  robot.write_joint_state_to_sim(joint_pos, joint_vel), but the arm seems to can not compute the right joint_pos_des.
```
 if count % 150 == 0:
            # reset time
            count = 0
        # reset joint state
            joint_pos = robot.data.default_joint_pos.clone()
            joint_vel = robot.data.default_joint_vel.clone()
            # robot.write_joint_state_to_sim(joint_pos, joint_vel)
            robot.reset()
```
the code I commented is show above, I want to know how to solve this and use this ik controller in my custom env. I would be more than appreciated for any useful suggestions!
![Image](https://github.com/user-attachments/assets/9b06a7b5-4cad-4f04-853f-88d92163ce2e)

RandomOakForest:
Thanks for posting this. What would be then the pose you expect the robot arm to be at? It seems `write_joint_state_to_sim` should be used. See the [example in this tutorial](https://isaac-sim.github.io/IsaacLab/main/source/tutorials/01_assets/add_new_robot.html). I will move this post to a discussion for the team to follow up.

PR #5400, opened 2026-04-25
[Newton] Backend-agnostic task-space accessors for IK/OSC

hujc7:
## 1. Summary

Make IK / OSC / RMPFlow task-space controllers backend-agnostic so Franka manipulation envs run under Newton.

The action terms previously called PhysX-only methods on `asset.root_view` directly, which crashed under Newton with `AttributeError`. The Franka reach envs worked around this by hardcoding `self.sim.physics = PhysxCfg(...)` with the comment *"{IK,OSC} control is not supported with Newton physics; use PhysX only"*. This PR removes that workaround.

End-to-end result: `Isaac-Reach-Franka-{IK-Abs,IK-Rel,OSC}-v0` run on either backend, picked via `presets=newton`.

## 2. Design

Four new properties on `BaseArticulationData`:

- `body_link_jacobian_w` — geometric Jacobian, linear rows at the link origin (USD prim frame). What IK / OSC consumers want.
- `body_com_jacobian_w` — geometric Jacobian, linear rows at the center of mass. Engine-natural form; useful for dynamics reasoning.
- `mass_matrix` — joint-space generalized mass matrix `M(q)`.
- `gravity_compensation_forces` — joint-space gravity-loading torques `g(q)`.

Plus one metadata property on `BaseArticulation`:

- `num_base_dofs` — number of free DoFs of the floating base. `0` for fixed-base, `6` for floating-base. Maps an actuated-joint id `j` to its column in J / M / g via `j + num_base_dofs`.

All four data properties are concrete-with-`NotImplementedError` defaults; backends override what they support.

### 2.1 DoF axis: `num_joints + num_base_dofs` (industry-standard)

The DoF axis prepends `num_base_dofs` columns at the front of the joint axis: `0` for fixed-base, `6` for floating-base. The 6 columns are the floating-base spatial velocity in world frame, ordered `[lin_x, lin_y, lin_z, ang_x, ang_y, ang_z]`.

This matches every major rigid-body library:

| Library | Floating-base layout |
|---------|----------------------|
| Pinocchio | Free-flyer joint contributes 6 to `nv`; reduced-coordinate `nv = 6 + n_actuated`. `pinocchio::computeJointJacobians` writes `(6, nv)`. |
| Drake | `MultibodyPlant` ephemeral floating joint; `MultibodyPlant::CalcJacobianSpatialVelocity` returns `(6, nv)` with leading 6 free-floating cols. |
| MuJoCo | `<freejoint/>` introduces 6 DoFs at the front of `qvel`; `mj_jac*` returns `(_, nv)`. |
| RBDL | `JointTypeFloatingBase` adds 6 to `dof_count`; `CalcPointJacobian` returns `(_, dof_count)`. |
| OCS2 | `generalizedCoordinatesNum = 6 + actuatedJointsNum` for floating-base robots. |
| iDynTree | `getFreeFloatingMassMatrix` returns `(6 + dofs, 6 + dofs)`. |

Consumers operating in actuated-joint space (action terms keyed by `joint_ids`) compute `[j + asset.num_base_dofs for j in joint_ids]` once at init — same application-level reduction as every surveyed library.

### 2.2 Output shapes by backend

Concrete shapes for each engine output and the wrapper transform applied. `N` = num_instances, `B` = num_bodies (full count incl. root), `J` = num_joints (actuated only). "Default" is the engine-native per-view shape (post-gather). "passthrough" = wrapper applies no shape change. The COM→origin shift is values-only (does not change shape).

| Property | Base | Newton (default ⟶ transform) | PhysX (default ⟶ transform) | Aligned |
|---|---|---|---|---|
| `body_link_jacobian_w` | fixed | `(N, B, 6, J)` ⟶ drop fixed-root row + COM→origin shift | `(N, B−1, 6, J)` ⟶ COM→origin shift | `(N, B−1, 6, J)` |
| `body_link_jacobian_w` | floating | `(N, B, 6, J+6)` ⟶ COM→origin shift | `(N, B, 6, J+6)` ⟶ COM→origin shift | `(N, B, 6, J+6)` |
| `body_com_jacobian_w` | fixed | `(N, B, 6, J)` ⟶ drop fixed-root row | `(N, B−1, 6, J)` ⟶ passthrough | `(N, B−1, 6, J)` |
| `body_com_jacobian_w` | floating | `(N, B, 6, J+6)` ⟶ passthrough | `(N, B, 6, J+6)` ⟶ passthrough | `(N, B, 6, J+6)` |
| `mass_matrix` | fixed | `(N, J, J)` ⟶ passthrough | `(N, J, J)` ⟶ passthrough | `(N, J, J)` |
| `mass_matrix` | floating | `(N, J+6, J+6)` ⟶ passthrough | `(N, J+6, J+6)` ⟶ passthrough | `(N, J+6, J+6)` |
| `gravity_compensation_forces` | fixed | (no upstream primitive) ⟶ `NotImplementedError` | `(N, J)` ⟶ passthrough | `(N, J)` |
| `gravity_compensation_forces` | floating | (no upstream primitive) ⟶ `NotImplementedError` | `(N, J+6)` ⟶ passthrough | `(N, J+6)` |

The aligned column collapses across base type using two derived symbols:

- `num_base_dofs` = `0` fixed | `6` floating — exposed as `BaseArticulation.num_base_dofs`
- `num_jacobi_bodies` = `B − 1` fixed | `B` floating — fixed-root row excluded for fixed-base

| Property | Aligned (generalized) |
|---|---|
| `body_link_jacobian_w`, `body_com_jacobian_w` | `(N, num_jacobi_bodies, 6, J + num_base_dofs)` |
| `mass_matrix` | `(N, J + num_base_dofs, J + num_base_dofs)` |
| `gravity_compensation_forces` | `(N, J + num_base_dofs)` |

Two observations:
1. The only body-axis asymmetry is **Newton's fixed-base Jacobian**: Newton's `eval_jacobian` includes a zero row for the fixed-root joint that the wrapper drops. PhysX's engine already drops it.
2. The DoF axis is in the industry-standard `J + num_base_dofs` form on both engines natively. The only DoF-axis asymmetry is `gravity_compensation_forces` (Newton NIE pending upstream primitive).

### 2.3 Why on `BaseArticulationData`, not `BaseArticulation`

`BaseArticulationData` already exposes per-body / per-joint state as cached lazy `@property` accessors with `TimestampedBuffer` invalidation (`body_link_pose_w`, `joint_pos`, etc.). The four new accessors fit the same shape — read-only state indexed by body / joint, refreshed per sim step — so they reuse that infrastructure and let consumer code stay symmetric:

```python
# IK action term
ee_pose = articulation.data.body_link_pose_w.torch[:, ee_idx]
ee_jac  = articulation.data.body_link_jacobian_w.torch[:, ee_jac_idx]   # same prefix
```

## 3. Fixed: latent PhysX IK / OSC frame mismatch

PhysX's `_root_view.get_jacobians()` returns linear rows referenced at each body's **center of mass**, not the link origin. Undocumented behavior — verified empirically by bypassing the IsaacLab wrapper and confirming `J · q̇` matches `body_com_lin_vel_w` to 1e-8 and differs from `body_link_lin_vel_w` by exactly `ω × r_com_world` (the rigid-body shift). The PhysX data layer already encoded this convention for velocities: `body_com_vel_w` is the raw passthrough of `_root_view.get_link_velocities()`; `body_link_vel_w` is **derived** via a shift kernel.

Before this PR, IK / OSC / RMPFlow action terms on PhysX consumed `_root_view.get_jacobians()` directly while using `data.body_link_pose_w` as the EE pose setpoint. The frame mismatch contributes `ω × r_com_world` per body to the linear-row contract — undetected in CI because no existing test compared `J · q̇` against `body_link_lin_vel_w` directly. The new contract test `test_get_jacobians_link_origin_contract` parametrized to `anymal` catches it explicitly: 0.32 m/s residual on PhysX without the fix.

After this PR, PhysX's `body_link_jacobian_w` applies the same COM→origin shift kernel that Newton uses, mirroring the existing `body_link_vel_w` derivation. Both backends now satisfy `body_link_jacobian_w · q̇ == body_link_vel_w` to numerical precision (test tolerance 5e-3 absolute, 1e-2 relative).

The COM-referenced sibling `body_com_jacobian_w` is exposed for callers that intentionally want the engine-native form (e.g. dynamics-side reasoning at the COM).

Also: the three PhysX passthrough properties (`body_com_jacobian_w`, `mass_matrix`, `gravity_compensation_forces`) now pin a single `ProxyArray` in `_create_buffers` instead of allocating one per property read. PhysX tensor-view getters return pointer-stable buffers for the articulation's lifetime — verified manually across `sim.step`, a manual joint write, and `sim.reset`.

### 4.1 Pre-allocation for capture safety

Dynamics-scratch allocation is grouped in a private `ArticulationData._create_jacobian_buffers(model)` helper called from `_create_buffers`. The helper is sectioned by which property each buffer feeds, and names reflect each buffer's physical role:

- **Shared scratch** (eval_jacobian output, reused as eval_mass_matrix's `J` input to skip a re-compute): `_jacobian_buf_flat`, `_joint_S_s_buf` (Featherstone motion subspace — shared between the two evals, hence the property-prefix-free name).
- **Per-view gather config**: `_jacobian_link_offset` (fixed-base row-0 skip), `_jacobian_view_art_ids` (flattened view-to-model index map).
- **`body_com_jacobian_w`**: `_jacobian_buf` (zero-copy 4-D view of `_jacobian_buf_flat`, kernel input) and `_body_com_jacobian_w_buf` (gather output, per-view).
- **`body_link_jacobian_w`**: `_body_link_jacobian_w_buf` (Center-Of-Mass-to-link-origin shift kernel output, per-view).
- **`mass_matrix`**: `_mass_matrix_full_buf` (model-wide `H` scratch — Composite Rigid Body Algorithm (CRBA) output), `_mass_matrix_body_I_s_buf` (Featherstone per-body spatial inertia aux, CRBA-only), `_mass_matrix_buf` (gather output, per-view).

Properties are allocation-free at step time. The kernel-launch sequence runs against fixed buffer pointers, which is what makes the per-step path safe under CUDA-graph capture.

### 4.2 View-level row gather

Newton's `eval_jacobian` / `eval_mass_matrix` write every articulation in the model into a single buffer (shape includes `model.articulation_count`), regardless of which `ArticulationView` invoked them. PhysX returns view-scoped data already. Two Warp kernels (`gather_jacobian_rows` 4-D, `gather_mass_matrix_rows` 3-D) gather just this view's rows into a contiguous view-sized destination so the caller-facing shape contract matches PhysX. The view-to-model index map is reused across both gathers.

### 4.3 `eval_mass_matrix` "J as input" gotcha

Newton's `eval_mass_matrix(state, H, J=None, body_I_s=None, joint_S_s=None)` treats `J` as **input** when provided — it skips the internal `eval_jacobian` and uses the buffer as-is. Passing an empty pre-allocated `J` produces `H = J^T·M·J = 0` → singular → `LinAlgError` in OSC's `torch.inverse(mass_matrix)`. The wrapper explicitly populates `J` by calling `eval_jacobian` first; we reuse `_jacobian_buf_flat` (same shape) so no separate scratch is needed.

### 4.4 COM→origin shift on `body_link_jacobian_w`

Newton's `eval_jacobian` writes linear-velocity rows at each link's **center of mass**. After `gather_jacobian_rows`, the `shift_jacobian_com_to_origin` Warp kernel applies `v_origin = v_com - ω × (R · body_com_pos_b)` per `(env, body, dof)` thread, writing to `_body_link_jacobian_w_buf`. The COM-referenced source buffer is reused as-is for `body_com_jacobian_w` and `mass_matrix`.

### 4.5 FK staleness

`eval_jacobian` and `eval_mass_matrix` read `state.body_q` (per-body world transforms). After a manual `write_joint_position_to_sim_*` (no sim step), `state.joint_q` is updated but `state.body_q` is stale until `eval_fk` runs. The new properties match the existing `body_link_pose_w` convention — refresh FK lazily via `_ensure_fk_fresh()` (Python-guarded `SimulationManager.forward()`) before invoking the eval kernels. PhysX has its own internal refresh on the equivalent getters (verified empirically by the new manual-write tests passing on PhysX without an explicit trigger).

## 5. Action-term gating

`OperationalSpaceControllerAction._compute_dynamic_quantities` previously fetched mass matrix and gravity-compensation forces unconditionally on every step. Under the new abstraction this would still call Newton's gravity-comp stub even when the user disabled gravity compensation in the controller config.

The fetches are now gated to match what the controller actually consumes:

- Mass matrix fetched when `inertial_dynamics_decoupling=True` **or** `nullspace_control != "none"` (the null-space torque term in `OperationalSpaceController.compute()` consumes mass matrix independently of inertial decoupling).
- Gravity comp fetched only when `gravity_compensation=True`.

Side benefit: skips a per-step engine call on PhysX when neither flag is set.

## 6. Known limitation: gravity compensation on Newton

Newton's `ArticulationView` has no gravity-compensation primitive (only `eval_fk` / `eval_jacobian` / `eval_mass_matrix`). `gravity_compensation_forces` raises `NotImplementedError` on Newton; OSC users on Newton must set `gravity_compensation=False` until upstream lands the primitive (newton-physics/newton#2497, #2529, #2625). The strict-xfail test `test_get_gravity_compensation_forces_not_implemented_on_newton` flips to XPASS when that happens, signaling the maintainer to remove the wrapper stub and OSC guidance.

## 7. Reach-env cfg cleanup

Removed the `self.sim.physics = PhysxCfg(bounce_threshold_velocity=0.2)` override from `Isaac-Reach-Franka-{IK-Abs,IK-Rel,OSC}-v0`. Tasks now inherit the parent `ReachPhysicsCfg` preset, so `presets=newton` selects `NewtonCfg` and `presets=physx` (the default) keeps the previous behavior — same `bounce_threshold_velocity=0.2` lives on `ReachPhysicsCfg.default = PhysxCfg(bounce_threshold_velocity=0.2)`. No information lost.

### 8.1 Existing controller tests migrated

- [x] `test_differential_ik.py` migrated to `robot.data.body_link_jacobian_w.torch`. Passes on PhysX (2/2).
- [x] `test_operational_space.py` migrated to `robot.data.body_link_jacobian_w.torch`, `robot.data.mass_matrix.torch`, `robot.data.gravity_compensation_forces.torch`. PhysX (12/18 — the 6 failures are pre-existing `ContactSensor` env issues unrelated to this PR; affected tests fail on `omni.physics.tensors.api` import which is independent of the bridge).
- [x] `test_floating_base_osc_action_term_indexing` cfg unchanged.

### 8.2 New tests in this PR

**Shape contracts** — lock the public DoF / body axes:

| Test | Backend | Purpose | Result |
|------|---------|---------|--------|
| `test_get_jacobians_shape_fixed_base` | PhysX + Newton | `body_link_jacobian_w` drops the fixed-root row → `(N, B−1, 6, J)`. | PASSES |
| `test_get_jacobians_shape_floating_base` | PhysX + Newton | Floating base keeps root row + prepends 6 base cols → `(N, B, 6, J+6)`. | PASSES |
| `test_get_mass_matrix_shape_and_nonsingular_fixed_base` | PhysX + Newton | `(N, J, J)` + strictly positive diagonal (catches the model-wide-padding bug that masks heterogeneous-scene mismatch as a singular matrix). | PASSES |
| `test_get_mass_matrix_shape_floating_base` | Newton | `(N, J+6, J+6)` — floating-base includes the 6 free-root DoFs. | PASSES |
| `test_heterogeneous_scene_per_view_shapes` | Newton | Mixed Franka+Anymal scene: each view returns its OWN asset shape, not `model.max_*`. Direct regression test for the heterogeneous-padding bug. | PASSES |

**Math / physics contracts** — values, not just shapes:

| Test | Backend | Purpose | Result |
|------|---------|---------|--------|
| `test_get_jacobians_link_origin_contract[panda \| anymal]` | PhysX | `J · q̇ == body_link_lin_vel_w` identity (sharp J reference-point check). Anymal exercise catches the latent COM/link mismatch this PR fixes. | PASSES |
| `test_get_jacobians_link_origin_contract[panda \| anymal]` | Newton | Same identity, ground truth from `state.body_qd` minus the COM offset shift. | PASSES |
| `test_get_mass_matrix_symmetry_pd[panda \| anymal]` | PhysX + Newton | `M(q)` is square, symmetric, positive-definite. | PASSES (4/4) |

**Freshness contracts** — Forward Kinematics (FK) refresh after manual writes:

| Test | Backend | Purpose | Result |
|------|---------|---------|--------|
| `test_jacobian_refreshes_after_manual_joint_write[panda \| anymal]` | PhysX + Newton | After `write_joint_position_to_sim_index` (no sim step), reading the Jacobian reflects the new joint state. Locks in the FK-staleness contract on the manual-write code path. | PASSES (4/4) |
| `test_mass_matrix_refreshes_after_manual_joint_write[panda \| anymal]` | PhysX + Newton | Same contract for mass matrix. | PASSES (4/4) |

**Gravity compensation** — accessor + Operational Space Control (OSC) integration + negative control:

| Test | Backend | Purpose | Result |
|------|---------|---------|--------|
| `test_get_gravity_compensation_forces_static_equilibrium` | PhysX | Apply only `τ_gc` to a non-trivial Franka pose; assert it stays static. Pins the accessor in isolation, no controller masking. | PASSES |
| `test_franka_osc_gravity_compensation_holds_under_gravity` | PhysX | OSC + `gravity=g(q)` under scene gravity holds the EE pose. Pins (a) `_jacobi_joint_idx + num_base_dofs` indexing, (b) `OSC.compute(gravity=...)` torque math, (c) reachability through the action-term pipeline. | PASSES |
| `test_franka_osc_no_gravity_compensation_sags_under_gravity` | PhysX | Negative control — OSC **without** gravity comp DOES drift under gravity. Proves the with-comp test isn't passing because `g(q)=0`. | PASSES |
| `test_get_gravity_compensation_forces_not_implemented_on_newton` | Newton | Strict-xfail pin for the upstream Newton gap. Flips to XPASS when upstream lands the primitive. | XFAIL (correct) |

**End-to-end Inverse Kinematics (IK) and OSC accuracy** — production sentinels:

| Test | Backend | Purpose | Result |
|------|---------|---------|--------|
| `test_franka_ik_tracking_accuracy` | PhysX + Newton | Damped-Least-Squares IK convergence sentinel via the new accessors. | PASSES (PhysX 0.01 mm, Newton 0.00 mm) |
| `test_franka_osc_tracking_accuracy` | PhysX + Newton | OSC convergence sentinel via Jacobian + mass matrix. | PASSES (both at machine precision) |

Latest CI: `isaaclab_physx` and `isaaclab_newton` both green on the most recent push.

### 8.3 Determinism and stability

The accuracy sentinels are bit-for-bit deterministic on both backends at the 5 cm short-target setup: 20× consecutive runs each give the same `pos_mean` to 5 decimal places. Both tests assert on tail mean rather than tail min — the latter is the bottom of any oscillation envelope and can pass spuriously while the actual tracking error is much larger. Tight regression sentinels rather than flaky bounds. No `pytest-rerunfailures` retry decoration — a CI failure should be a real regression, not noise to retry away.

### 8.4 Smoke runs

- [x] `random_agent` on `Isaac-Reach-Franka-IK-Abs-v0` under Newton: 5,205 physics substeps zero-error.
- [x] `random_agent` on `Isaac-Reach-Franka-IK-Rel-v0` under Newton: 306 substeps zero-error.
- [x] `random_agent` on `Isaac-Reach-Franka-OSC-v0` under Newton: 290 substeps zero-error.

### 8.5 Test-setup notes

The accuracy tests need two fixes that the standalone path doesn't get for free (production envs do):

- Teleport to `init_state.joint_pos` post-`sim.reset()`. Without it, the robot sits at the URDF-neutral pose where Franka's wrist axes nearly align — rank-deficient Jacobian, multi-cm DLS plateau.
- Override `sim_cfg.gravity = (0, 0, 0)` in the Newton `sim` fixture (`build_simulation_context(gravity_enabled=False)` is silently ignored when an explicit `sim_cfg` is passed).

The OSC test additionally zeros actuator PD gains so OSC's joint-effort output isn't opposed by `kp·(target − q)` (same way OSC is wired in production action terms). With these in place, Newton hits machine precision and PhysX hits ~10 µm. Newton's PD does have a real `g_torque/kp` gravity-sag that PhysX's TGS masks via constraint projection — surfaces only when gravity is on without gravity compensation, which is the upstream gap in § 6.

## 9. Files

Touched five extensions; per-package changelog fragments under `source/<pkg>/changelog.d/jichuanh-ik-newton-compat-mvp*.rst`:

- `isaaclab` (minor): four new properties on `BaseArticulationData`; `num_base_dofs` on `BaseArticulation`; controller-action gating + migration to data-layer accessors; doc updates.
- `isaaclab_physx`: data-layer impls + `shift_jacobian_com_to_origin` kernel; docs the latent frame-mismatch fix.
- `isaaclab_newton` (minor): data-layer impls (eval + gather + shift); model-sized scratch and view-sized output buffers migrated from articulation to data layer; gravity-comp NIE stub.
- `isaaclab_ovphysx`: bridge methods removed (inherits NIE).
- `isaaclab_tasks`: hardcoded `PhysxCfg` removals + direct-workflow caller migrations to data-layer accessors.


hujc7:
@greptileai review

hujc7:
@greptileai review

hujc7:
@greptileai review

hujc7:
@greptileai review

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
```diff
@@ -329,13 +329,11 @@ def _apply_gravity_compensation(self) -> None:
             # Get gravity compensation forces using cached tensor
             if self._asset.is_fixed_base:
                 gravity = torch.zeros_like(
-                    wp.to_torch(self._asset.root_view.get_gravity_compensation_forces())[
-                        :, self._controlled_joint_ids_tensor
-                    ]
+                    wp.to_torch(self._asset.get_gravity_compensation_forces())[:, self._controlled_joint_ids_tensor]
                 )
```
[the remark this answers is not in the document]
hujc7: I do think crash is the right behavior, otheriwise it's hiding problem if user turns on gravity comp with newton.

aravindskumar-nvidia:
@hujc7 @kellyguo11 

Thanks for the Newton IK compatibility scaffold. I tested it locally on Franka reach and wanted to sanity-check two things with you:

- Newton’s Jacobian output appears to need a spatial-to-geometric shift before feeding the existing IK controller, which seems to expect the PhysX-style body-origin Jacobian. 
- I also found that we need to preserve disableGravity semantics in the Newton/MuJoCo loader for the Franka asset; otherwise the arm still sags even after the Jacobian fix.

Does that sound right to you?

hujc7:
@aravindskumar-nvidia Thanks! I do miss some testing in this PR. Let me take a look once my machine is back.

source/isaaclab/isaaclab/envs/mdp/actions/rmpflow_task_space_actions.py
```diff
@@ -61,10 +60,13 @@ def __init__(self, cfg: rmpflow_actions_cfg.RMPFlowActionCfg, env: ManagerBasedE
...
             self._jacobi_body_idx = self._body_idx - 1
-            self._jacobi_joint_ids = self._joint_ids
         else:
             self._jacobi_body_idx = self._body_idx
-            self._jacobi_joint_ids = [i + 6 for i in self._joint_ids]
+        # Jacobian / mass-matrix DoF axis prepends ``num_base_dofs`` floating-base columns
+        # (0 for fixed-base, 6 for floating-base), so actuated-joint id ``j`` maps to
+        # Jacobian column ``j + num_base_dofs``.
+        offset = self._asset.num_base_dofs
+        self._jacobi_joint_ids = [j + offset for j in self._joint_ids]
```
hujc7: combine

source/isaaclab/isaaclab/envs/mdp/actions/task_space_actions.py
```diff
@@ -77,10 +76,13 @@ def __init__(self, cfg: actions_cfg.DifferentialInverseKinematicsActionCfg, env:
...
             self._jacobi_body_idx = self._body_idx - 1
-            self._jacobi_joint_ids = self._joint_ids
         else:
             self._jacobi_body_idx = self._body_idx
-            self._jacobi_joint_ids = [i + 6 for i in self._joint_ids]
+        # Jacobian / mass-matrix DoF axis prepends ``num_base_dofs`` floating-base columns
+        # (0 for fixed-base, 6 for floating-base), so actuated-joint id ``j`` maps to
+        # Jacobian column ``j + num_base_dofs``.
+        offset = self._asset.num_base_dofs
+        self._jacobi_joint_ids = [j + offset for j in self._joint_ids]
```
hujc7: combine

hujc7:
**Note for reviewers — current CI red is upstream, not in this PR's code.**

The failing tests (`test_get_jacobians_shape_fixed_base[panda]`, `test_get_mass_matrix_shape_and_nonsingular_fixed_base[panda]`, `test_get_gravity_compensation_forces_static_equilibrium[panda]`, and the Newton-side equivalents) all share one symptom: PhysX's `_root_view.get_generalized_mass_matrices()` returns a `(N, 15, 15)` mass matrix for Franka where it should be `(N, 9, 9)` — i.e., Franka is being treated as floating-base. Newton's `articulation.is_fixed_base` correspondingly returns `False`.

Root cause: a transitional regression in Isaac Sim's `latest-develop` rolling Docker tag (`nvcr.io/nvidian/isaac-sim:latest-develop`, per `.github/workflows/config.yaml`) where articulation property queries skip the USD-side metadata population. Articulation root type and DoF count fall back to a wrong default.

Repro scope:
- CI (`latest-develop`): fails
- Local against stable Isaac Sim 6.0.0.0: **all tests pass** (Newton 1.2.0rc2 — same as CI's pin)
- Local against env_isaaclab_test (Newton 1.2.0rc2, stable Isaac Sim 6.0.0.0): **all tests pass**
- Other recent PRs (#5535, #5523, #5504): pass `isaaclab_physx` because they don't add tests that assert articulation shape contracts. This PR is the first to exercise the property-query path against shape expectations, so it's the first to surface the upstream regression.

The upstream fix is already implemented in the property-query path; awaiting it landing in the rolling tag. No code change to this PR is needed. Will re-trigger CI once the upstream fix lands.

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
AntoineRichard, line 1: Why is the gravity comp mixed into this PR?

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -857,6 +857,56 @@ def body_com_pose_b(self) -> ProxyArray:
...
+    """
+
+    @property
+    def body_com_jacobian_w(self) -> ProxyArray:
+        """See :attr:`isaaclab.assets.BaseArticulationData.body_com_jacobian_w`.
+
+        PhysX implementation: passthrough of ``_root_view.get_jacobians()``, which is
+        natively COM-referenced.
+        """
+        return ProxyArray(self._root_view.get_jacobians())
```
AntoineRichard: Did you check if this pointer was stable/fixed? This way we could avoid creating a proxy array everytime?

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -857,6 +857,56 @@ def body_com_pose_b(self) -> ProxyArray:
...
+        )
+        return self._body_link_jacobian_w_ta
+
+    @property
+    def mass_matrix(self) -> ProxyArray:
+        """See :attr:`isaaclab.assets.BaseArticulationData.mass_matrix`.
+
+        PhysX implementation: passthrough of ``_root_view.get_generalized_mass_matrices()``.
+        """
+        return ProxyArray(self._root_view.get_generalized_mass_matrices())
```
AntoineRichard: Did you check if this pointer was stable/fixed? This way we could avoid creating a proxy array everytime?

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -857,6 +857,56 @@ def body_com_pose_b(self) -> ProxyArray:
...
+
+    @property
+    def gravity_compensation_forces(self) -> ProxyArray:
+        """See :attr:`isaaclab.assets.BaseArticulationData.gravity_compensation_forces`.
+
+        PhysX implementation: passthrough of ``_root_view.get_gravity_compensation_forces()``.
+        """
+        return ProxyArray(self._root_view.get_gravity_compensation_forces())
```
AntoineRichard: Did you check if this pointer was stable/fixed? This way we could avoid creating a proxy array everytime?

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -127,6 +127,19 @@ def update(self, dt: float) -> None:
...
+        Newton's ``state.body_q`` (per-body world transforms) is updated by ``eval_fk``,
+        invoked here through ``SimulationManager.forward()``. After a manual joint or root
+        write that bypassed the sim step (``write_*_to_sim_*``), ``_fk_timestamp`` is set
+        to ``-1.0`` to force a refresh on the next read of any property that depends on
+        body poses (``body_link_pose_w``, the Jacobian properties, ``mass_matrix``).
+        """
+        if self._fk_timestamp < self._sim_timestamp:
+            SimulationManager.forward()
+            self._fk_timestamp = self._sim_timestamp
+
```
AntoineRichard, line 142: Don't we already have something like that? If not can we harmonize to other articulation props like the bodies links and the writers?

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -812,6 +823,124 @@ def body_com_pose_b(self) -> ProxyArray:
...
+        ``ArticulationView`` exposes only ``eval_fk`` / ``eval_jacobian`` /
+        ``eval_mass_matrix``. Use PhysX, or set the controller's
+        ``gravity_compensation=False`` until upstream Newton adds the primitive.
+        """
+        raise NotImplementedError(
+            "Newton has no gravity-compensation primitive. Use PhysX, or set the controller's"
+            " ``gravity_compensation=False`` until upstream Newton adds an"
+            " ``eval_gravity_compensation`` API."
+        )
+
```
AntoineRichard, line 949: Did we open a ticket for this? @preist-nvidia for viz.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -1516,6 +1648,60 @@ def _create_buffers(self) -> None:
...
+        # body spatial inertias, written by ``eval_mass_matrix``.
+        self._mass_matrix_buf = wp.zeros(
+            (model.articulation_count, max_dofs, max_dofs), dtype=wp.float32, device=self.device
+        )
+        self._mass_matrix_body_I_s_buf = wp.zeros(model.body_count, dtype=wp.spatial_matrix, device=self.device)
+        self._mass_matrix_buf_view = wp.zeros(
+            (self._num_instances, self._num_joints + num_base_dofs, self._num_joints + num_base_dofs),
+            dtype=wp.float32,
+            device=self.device,
+        )
```
AntoineRichard: Can we spin this out as a different method: `_create_ jacobian_buffers` or something along these lines?

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -1516,6 +1648,60 @@ def _create_buffers(self) -> None:
...
         )
+        # -- dynamics quantities for task-space controllers
+        # Newton's eval_jacobian / eval_mass_matrix write into model-sized scratch buffers
+        # spanning every articulation in the model; the gather kernels below extract the rows
+        # belonging to this view. The output buffers are sized using THIS articulation's body /
+        # DoF counts (not the model-wide ``max_*``) so heterogeneous scenes do not leak
+        # zero-padded rows / cols into the returned tensor. The DoF axis includes
+        # ``num_base_dofs`` floating-base columns up front (0 for fixed-base, 6 for floating-
+        # base), matching the cross-library industry convention used by PhysX, Pinocchio,
+        # Drake, MuJoCo, RBDL, OCS2, and iDynTree.
```
AntoineRichard: In general it looks like this is something that may not fall on lab considering how important this looks? @preist-nvidia @kellyguo11  for viz.

AntoineRichard: @camevor should the selection API provide tools for this similarly to how it provides per articulation root and FK?

AntoineRichard (approved):
If we need it now, this is fine, I didn't dig too deep, didn't have time. But it looks like some features here could be interesting to newton in general.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
AntoineRichard: @jcarius-nv instead of @preist-nvidia :D

source/isaaclab/isaaclab/envs/mdp/actions/pink_task_space_actions.py
hujc7: Just to unify the API to provide it in a backend agnostic way if newton side has plan to provide it in the physx way. Can be removed if there's no such plan

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
hujc7: Updated.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
hujc7: Fixed all to follow existing pattern.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
jcarius-nv: @AntoineRichard could I ask you (or your AI agent) to file a ticket for us at https://github.com/newton-physics/newton/issues , specifying precisely how you'd like us to provide Jacobians and mass matrices (and anything else) and in which format?

A paragraph or two are fine, but it should include acceptance criteria and ideally a way to test it, which shouldn't be hard given this MR here. We'll then triage and assign it accordingly.
Thank you!

PR #5415, opened 2026-04-28
Add Metrics/success_rate logging across RL tasks

ooctipus:
# Description

Adds a unified ``Metrics/success_rate`` log key across all rsl_rl-trainable RL
environments so cross-task success curves land on a single TensorBoard / wandb
card instead of fragmenting into per-term keys.

The semantics are deliberately uniform:

- For **velocity locomotion** (manager-based and direct AnymalC), success is
  the per-env binary \"episode-mean XY-vel error AND episode-mean yaw-vel error
  both stay below thresholds\", mean-reduced across the resetting envs.
- For **manipulation** tasks with a clear functional reward term that already
  reads the goal state (cabinet ``open_drawer_bonus``, lift ``object_goal_distance``),
  the success bit is hosted on that term via an opt-in ``success_threshold`` param.
  No zero-weight bookkeeping reward terms.
- For **pose-driven** tasks (manipulation/reach, navigation), ``UniformPoseCommandCfg``
  / ``UniformPose2dCommandCfg`` gain ``position_success_threshold``; setting it
  enables the metric, leaving it ``None`` stays a no-op.
- For **survival** tasks (classic cartpole, classic humanoid + ant, direct
  cart_double_pendulum, locomotion-derived ant/humanoid), survival = timed-out
  without termination is logged from the appropriate ``reset()`` hook.
- For **direct envs** (anymal_c, quadcopter, franka_cabinet, shadow_hand_over,
  inhand_manipulation/{allegro,shadow}, automate, factory, cartpole), the metric
  is flushed from ``_reset_idx`` using per-episode accumulators that are
  guaranteed pre-reset state.

Threshold defaults are the empirically-validated values from training plots:

- ``vel_xy_success_threshold = 0.5`` m/s (uniform across all robots).
- ``vel_yaw_success_threshold = 0.4`` rad/s (quadruped baseline). Bipeds (h1,
  cassie, g1) override to ``0.8`` rad/s in their ``rough_env_cfg.__post_init__``;
  flat variants inherit the looser biped value through the standard
  ``flat_env_cfg → rough_env_cfg`` chain.

Also fixes a long-standing wart in ``UniformVelocityCommand._update_metrics``: the
old \"/ max_command_step\" divisor was actually a per-episode integral scaled by
``episode_length / max_command_time``, which made ``error_vel_xy/yaw`` silently
depend on ``episode_length_s`` and ``resampling_time_range``. Replaced with a
true per-step running mean finalized in a ``reset()`` override. In the typical
config (``episode_length_s == max_command_time``) the value is unchanged.

Fixes # (issue)

## Type of change

## Screenshots

N/A — log keys only.

## Checklist

- [x] I have updated the changelog and the corresponding version in the extension's \`config/extension.toml\` file


ooctipus:
Good catch — the review is correct. ``UniformVelocityCommand.reset()`` was leaving
``success_rate`` in ``self.metrics``, so ``CommandManager.reset()`` auto-prefixed it
to ``Metrics/base_velocity/success_rate`` and the 40 manager-based velocity tasks
never reached the unified ``Metrics/success_rate`` key.

Pushed ``f74cb6e7`` which mirrors the pattern from ``InHandReOrientationCommand``:
keep the per-env tensor flowing through the base-class ``reset()`` for free, then
pop the float from the returned extras and write it directly to
``self._env.extras["log"]["Metrics/success_rate"]``. Now matches the routing used
by ``UniformPoseCommand`` and ``UniformPose2dCommand``.

ooctipus:
Pushed ``736d60c5`` addressing the three P2 comments:

1. **Threshold three-way inconsistency** (`commands_cfg.py:93`): cfg defaults now
   match the PR description and our empirical findings — ``vel_xy = 0.5`` (uniform)
   and ``vel_yaw = 0.4`` (quadruped baseline). Docstrings updated. Dropped the
   redundant ``0.6`` override in ``LocomotionVelocityRoughEnvCfg.__post_init__``,
   updated ``AnymalCFlatEnvCfg`` to ``vel_yaw = 0.4`` so flat / rough / direct all
   agree. Bipeds (h1, cassie, g1) override ``vel_yaw = 0.8`` in their per-robot
   ``rough_env_cfg`` since biped yaw control is empirically harder; flat variants
   inherit it through the existing ``flat → rough`` chain.

2. **Breaking metric key rename** (`pose_2d_command.py:59`): reverted the rename
   ``error_pos_2d`` → ``error_pos``, restoring the original key that downstream
   dashboards expect. The original init/writer mismatch is fixed by aligning
   both on ``error_pos``.

3. **Unused persistent buffer** (`quadcopter_env.py:166`): collapsed
   ``self._last_episode_success`` into an inline expression in ``_reset_idx``.

fatimaanes (approved):

kellyguo11 (approved):

PR #5419, opened 2026-04-28
Fix lazy physics preset imports

AntoineRichard:
# Description

This PR fixes task config loading when optional physics backend packages are not installed.

Direct Cartpole, Ant, and Humanoid configs currently import `isaaclab_ovphysx` while defining
their physics preset alternatives. That makes default PhysX runs fail before the `ovphysx`
preset is selected if the optional package is not installed.

The change adds a generic `lazy_preset("module:Cfg")` helper for `PresetCfg` alternatives.
Lazy presets are collected for discoverability, but their config class is imported and
instantiated only when the preset is selected. If the selected preset cannot be imported,
the resolver raises an actionable error with the preset name, import path, and optional
install hint.

Fixes #N/A.

## Type of change

## Screenshots

N/A.

## Checklist

## Local Test Plan

- `PYTHONPYCACHEPREFIX=/tmp/isaaclab_pycache ./isaaclab.sh -p -m py_compile source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/direct/ant/ant_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/direct/humanoid/humanoid_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/direct/locomotion/locomotion_env.py source/isaaclab_tasks/test/test_hydra.py`
- `./isaaclab.sh -f`
- `git diff --check`

## Not Run Locally

- `PYTHONPYCACHEPREFIX=/tmp/isaaclab_pycache ./isaaclab.sh -p -m pytest source/isaaclab_tasks/test/test_hydra.py -q`: blocked because this shell's `/usr/bin/python3.12` environment is missing `torch`.
- `./isaaclab.sh -d`: blocked by PEP 668 externally-managed system Python while installing docs requirements.

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/direct/ant/ant_env_cfg.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/direct/ant/ant_env_cfg.py
@@ -6,7 +6,6 @@
 from __future__ import annotations
 
 from isaaclab_newton.physics import MJWarpSolverCfg, NewtonCfg
-from isaaclab_ovphysx.physics import OvPhysxCfg
 from isaaclab_physx.physics import PhysxCfg
 
 import isaaclab.sim as sim_utils
@@ -17,7 +16,7 @@ from isaaclab.sim import SimulationCfg
 from isaaclab.terrains import TerrainImporterCfg
 from isaaclab.utils import configclass
 
-from isaaclab_tasks.utils import PresetCfg
+from isaaclab_tasks.utils import PresetCfg, lazy_preset
 
 from isaaclab_assets.robots.ant import ANT_CFG
 
@@ -37,7 +36,10 @@ class AntPhysicsCfg(PresetCfg):
         num_substeps=1,
         debug_mode=False,
     )
-    ovphysx: OvPhysxCfg = OvPhysxCfg()
+    ovphysx = lazy_preset(
+        "isaaclab_ovphysx.physics:OvPhysxCfg",
+        error_hint="Install the OV-PhysX backend before selecting the 'ovphysx' preset.",
+    )
 
 
 @configclass
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,6 @@ _build
 
 # Superpowers (Claude Code plugin artifacts)
 docs/superpowers/
+
+# Codex local state
+/.codex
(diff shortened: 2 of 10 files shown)


source/isaaclab_tasks/isaaclab_tasks/direct/ant/ant_env_cfg.py
```diff
@@ -37,7 +36,10 @@ class AntPhysicsCfg(PresetCfg):
         num_substeps=1,
         debug_mode=False,
     )
-    ovphysx: OvPhysxCfg = OvPhysxCfg()
+    ovphysx = lazy_preset(
```
kellyguo11, line 39: hmm I'm not sure I'm a fan of this syntax, feels like more overload onto the users.

is it possible to import OvPhysxCfg without importing the package itself? maybe better isolation between the cfg and the APIs?

AntoineRichard: I agree, it could also be wrapped in a try catch? In general it's an issue. Alternatively, we could install all our extensions by default but not their dependencies. This would result in all imports working (thanks to the lazy imports) but the calls failing if not explicitly installed. @kellyguo11 @ooctipus what do you think. I feel like this would be the most elegant, but maybe the least transparent?

ooctipus: > we could install all our extensions by default but not their dependencies.

 I agree with this path, which I opened a pr here
 https://github.com/isaac-sim/IsaacLab/pull/5428
 
 >is it possible to import OvPhysxCfg without importing the package itself? maybe better isolation between the cfg and the APIs?
 
 this is exactly how isaaclab_teleop is handled

kellyguo11: I agree, I think that makes more sense to me. if the packages are not installed, then we can fail with a meaningful error that you'll need to install the dependencies.

PR #5070, opened 2026-03-20
Disables USD notice handlers during usd_replicate for 25-35% faster startup

aloysbaillet:
# Description

Summary

Disable IFabricUsd and SimulationManager USD notice handlers during bulk environment cloning in usd_replicate(), reducing startup time by 25-35% for manager-based tasks.

Problem

When usd_replicate() calls Sdf.CopySpec to clone thousands of environment prims, every single copy triggers IFabricUsd UsdNoticeListener::Handle(UsdNotice::ObjectsChanged) — an expensive Fabric↔USD sync operation. For Reach-Franka (4096 envs), this added 17 seconds (438 notice handler calls) during scene construction.

IsaacSim's own Cloner.disable_change_listener() already handles this by disabling the notice handler before cloning, but IsaacLab v3's replacement usd_replicate() function didn't replicate this behavior.

Additionally, the import path from isaacsim.core.simulation_manager import SimulationManager resolves to PhysxManager under IsaacLab's lazy_export module system, which lacks the enable_fabric_usd_notice_handler method. The correct import is from isaacsim.core.simulation_manager.impl.simulation_manager import SimulationManager.

Changes

source/isaaclab/isaaclab/cloner/cloner_utils.py:

• Add _disable_usd_notice_handlers() / _enable_usd_notice_handlers() helpers that mirror isaacsim.core.cloner.Cloner.disable_change_listener()
• Wrap usd_replicate() body in try/finally to disable handlers during bulk USD authoring
• Use correct full import path for SimulationManager

source/isaaclab/isaaclab/utils/string.py:

• Cache string_to_callable() with @functools.cache to avoid repeated importlib.import_module calls
• Fix triple ast.parse() call in is_lambda_expression()

Benchmark Results

Tested on DGXC (2x NVIDIA L40, 4096 envs, 1 GPU, warm run):
| Task                            | Before | After | Improvement           |
| ------------------------------- | ------ | ----- | --------------------- |
| Isaac-Reach-Franka-v0           | 29.0s  | 20.0s | -31%                  |
| Isaac-Velocity-Rough-H1-v0      | 37.9s  | 24.7s | -35%                  |
| Isaac-Velocity-Flat-Anymal-D-v0 | 29.8s  | 22.1s | -26%                  |
| Isaac-Cartpole-Direct-v0        | 8.9s   | 9.1s  | ~0% (minimal cloning) |
| Isaac-Ant-Direct-v0             | 11.9s  | 12.0s | ~0% (minimal cloning) |
[...]

PR #5432, opened 2026-04-28
Suspend Fabric USD notice listener during cloning for faster startup

ooctipus:
# Description

Adds a re-entrant context manager `disabled_fabric_change_notifies` that suspends the `omni::fabric::IFabricUsd` USD notice listener for the duration of bulk cloning. During `Sdf.CopySpec` loops in `usd_replicate`, every per-prim mutation otherwise fires `IFabricUsd::UsdNoticeListener::Handle` to do an immediate Fabric↔USD sync — for moderately heavy scenes that single hot path can dominate scene-load time. Toggling the listener's soft flag (`IFabricUsd.cpp:739`) gates the work without unregistering the listener, then the natural `SimulationContext.reset` path performs the catch-up resync in one pass.

The same handler is what `isaacsim.core.cloner.Cloner.disable_change_listener` toggles, but this implementation reaches it through the underlying `omni::fabric::IFabricUsd` Carbonite interface directly, so the cloner has **no `isaacsim.core.simulation_manager` dependency**. Since `omni.fabric` has no public Python binding for `setEnableChangeNotifies`, acquisition is via a small ctypes module modelled on the in-tree `isaaclab_newton.physics._cubric` pattern.

The context manager is wired in at the cloner-session boundary — `clone_from_template` and the two cloning regions in `InteractiveScene` — not inside `usd_replicate` itself, keeping the leaf USD-authoring primitive pure.

## Motivation

This PR supersedes the approach taken in #5070 (still open). #5070 reaches the same toggle via `isaacsim.core.simulation_manager.SimulationManager.enable_fabric_usd_notice_handler`, which:

1. Adds an `isaacsim.core.simulation_manager` dependency to the cloner (project policy: avoid `isaacsim.*` implementation deps).
2. Implements via a state-dict pattern + a private `_manage_notice_handlers` kwarg leaked through public `usd_replicate`.
3. Disables the handler from inside `usd_replicate` and never re-enables it on exit (global-state leak that survives exceptions).

This PR keeps #5070's measured perf win while replacing the above with a single context manager wired in at the orchestrator boundary, exception-safe, re-entrant, and IsaacSim-implementation-free.

### Architecture

| Concern | Where it lives |
|---|---|
| ABI-coupled ctypes binding | `source/isaaclab/isaaclab/cloner/_fabric_notices.py` (private, ~135 lines) |
| Public context manager | `disabled_fabric_change_notifies(stage, *, restore=True)` in `cloner_utils.py` |
| Application | `clone_from_template` body and `InteractiveScene.clone_environments` |

The `restore=False` kwarg, used at the two scene-init sites, opts out of the on-exit re-enable. This avoids the redundant `forceMinimalPopulate` batch that fires when the flag flips back on; downstream `SimulationContext.reset` performs the same Fabric resync as part of normal startup. `restore=True` (the default) is exception-safe and is what tests and any future direct callers get.

When the Carbonite interface can't be acquired (e.g. running outside a live Kit application), the context manager falls through to a no-op so callers never break — they just don't get the perf win.

## Benchmarks

`scripts/benchmarks/benchmark_startup.py`, RTX 5090, headless, warm run (2nd of 2 invocations after kernel/extension caches populate).

| Task | num_envs | Backend | Before | After | Saved | % |
|---|---:|---|---:|---:|---:|---:|
| `Isaac-Cartpole-Direct-v0` | 4096 | PhysX | 7.41 s | 6.47 s | 0.94 s | −12.7% |
| `Isaac-Cartpole-Direct-v0` | 4096 | Newton | 22.06 s | 22.01 s | 0.05 s | ~0% |
| `Isaac-Velocity-Flat-Anymal-C-v0` | 4096 | PhysX | 28.09 s | 14.49 s | 13.60 s | **−48.4%** |
| `Isaac-Velocity-Flat-Anymal-C-v0` | 4096 | Newton | 36.72 s | 32.99 s | 3.73 s | −10.1% |
| `Isaac-Dexsuite-Kuka-Allegro-Reorient-v0` | 8192 | PhysX | 104.84 s | 41.94 s | 62.90 s | **−60.0%** |
| `Isaac-Dexsuite-Kuka-Allegro-Reorient-v0` | 8192 | Newton | 81.00 s | 74.17 s | 6.83 s | −8.4% |

Savings come entirely from `env_creation` (Scene Creation + Simulation Start) and `first_step`. Non-cloning phases (`app_launch`, `python_imports`, `task_config`) are unchanged within run-to-run noise. Newton kitless runs see smaller wins because Fabric is only touched when `omni.fabric` is loaded for rendering; PhysX with Kit is the primary beneficiary.

The `first_step` collapse on heavy scenes (e.g. Dexsuite −93%, 10.8 s → 0.74 s) is the deferred Fabric resync work folding into the existing `SimulationContext.reset` pass with much less overhead than an eager `forceMinimalPopulate`.

## Type of change

- Performance improvement

## Test plan

- `pytest source/isaaclab/test/sim/test_cloner.py` → 18 passed
- `pytest source/isaaclab_physx/test/sim/test_cloner.py` → 22 passed (2 xfailed/2 xpassed are pre-existing)
- `./isaaclab.sh -f` (pre-commit) clean
- Smoke test confirms `restore=True` round-trips the flag, `restore=False` leaves it off, and nested context managers are re-entrant.

## Follow-ups

- File a Kit-team request for an `omni.fabric` Python binding to `setEnableChangeNotifies`, after which `_fabric_notices.py` can be deleted.
- Consider an ABI-drift smoke test under `test_cloner.py` to catch Kit-side vtable changes in CI.

## Checklist


ooctipus:
Re: the review bot's "Critical: Missing `this` pointer in vtable function calls" — this finding is incorrect, and I want to flag it so it doesn't block human review.

The premise of the bot's objection is that the function pointers read out of the `IFabricUsd` interface struct are C++ virtual member functions, and therefore each call must pass the interface pointer as an implicit `this`. They are not. Carbonite interfaces (`omni::fabric::IFabricUsd`, `omni::cubric::IAdapter`, etc.) are CARB_ABI plain-old-data structs of free function pointers — `setEnableChangeNotifies(FabricId, bool)` is a free function with exactly the signature its declaration shows. There is no `this`.

The reference for this is the in-tree `isaaclab_newton/physics/_cubric.py`, which the PR description names as the pattern this module mirrors. It uses the exact same `CFUNCTYPE`-without-`this` shape across all four bound entry points, e.g.:

```python
self._release_fn = ctypes.CFUNCTYPE(
    ctypes.c_bool,
    ctypes.c_uint64,           # AdapterId, no `this`
)(release_addr)

self._compute_fn = ctypes.CFUNCTYPE(
    ctypes.c_bool,
    ctypes.c_uint64,           # adapterId
    ctypes.c_uint32,           # options
    ctypes.c_int32,            # dirtyMode
    ctypes.c_void_p,           # outAccountFlags*
)(compute_addr)
```

That module ships in production and has been working since merge. Adding a `c_void_p` `this` parameter as the bot suggests would actually break the call by shifting all subsequent arguments off by 8 bytes — i.e. the bot's recommended "fix" is what would corrupt state. I'm leaving the binding as-is.

The other findings — `StageCache.Insert` leak, undocumented `clone_from_template` side-effect, scope-of-`restore=False` clarity in `InteractiveScene.__init__` — are valid and I'm pushing fixes for those.

kellyguo11 (approved):

PR #5433, opened 2026-04-28
[Newton] Rename per-env labels in physics replication (depends on #5523)

hujc7:
## Summary

Extends `_rename_builder_labels` in `isaaclab_newton.cloner.newton_replicate` so that every label-bearing column on the merged Newton `ModelBuilder` is rewritten to per-env USD paths after replication. Previously, only the built-in body/joint/shape/articulation columns were rewritten; tendon labels (and any other string-typed custom-attribute column) kept the source proto path on every replicated environment.

## Stack / dependencies

- **Depends on #5523** (\"[Newton] Bump Newton pin to v1.2.0rc2\"). After #5523 lands, this PR rebases cleanly on develop.

The Newton 1.2 release ([newton-physics/newton#2659](https://github.com/newton-physics/newton/pull/2659)) also includes the upstream tendon-scoping fix that obsoletes the IsaacLab-side \`_scope_custom_frequencies\` workaround a previous version of this PR carried — that workaround has been removed in favor of relying on the Newton bump in #5523.

## Why the rename is needed

Newton's \`add_builder\` copies each proto's bodies, joints, shapes, articulations, etc. into the merged builder verbatim, and tags each row with a \`*_world\` integer column to track env identity. Labels (path strings) are copied as-is. So after cloning N environments from one proto, the merged builder has N copies of every row, all with the **same proto-path string label**, distinguished only by the integer \`*_world\` column.

IsaacLab keys most of its data flow off **USD prim paths** (sensor binding, event-term scope, visualization, logging). It needs labels to be unique per-env paths so a body called \`/World/envs/env_3/Robot/Forearm\` is reachable by path lookup. The rename function is the bridge: it walks every label-bearing column post-replication and rewrites the source-root prefix to the per-env destination root using each row's \`*_world\` value.

Until this PR, the rename only walked **5 built-in label arrays**. Tendon labels and any string-typed custom-attribute column were missed, so e.g. \`mujoco:tendon_label\` showed \`/World/envs/env_0/...\` for every env — surfaced on Shadow Hand fixed tendons.

### \`_rename_builder_labels\` extension

* **Pass 1 (built-in label arrays)** — extended from 5 to 6 entity types: \`body\`, \`joint\`, \`shape\`, \`articulation\`, \`constraint_mimic\`, **\`equality_constraint\`** (the latter was missing — would have surfaced for any env using \`MjcEquality\` constraints, currently none).
* **Pass 2 (string custom-attribute columns)** — new. Walks every registered custom attribute, finds string-typed columns whose frequency has a \`references=\"world\"\` companion column, and applies the same prefix rewrite. Any future solver-registered string column at such a frequency is handled automatically without changes here.
* **Path-separator boundary** on the prefix match: \`startswith(src_path.rstrip(\"/\") + \"/\")\`. Prevents source paths that are string prefixes of one another (\`/Sources/protoA\` vs \`/Sources/protoAB\`) from cross-contaminating when both feed the same envs.
* **Hard error on length mismatch**: raises \`ValueError\` if the parallel \`(labels, worlds)\` arrays differ in length, instead of silently truncating. By contract Newton's \`add_builder\` keeps them in lockstep.

### Tests

New \`source/isaaclab_newton/test/cloner/test_rename_builder_labels.py\` with 10 cases covering:

- Both passes with built-ins and \`mujoco:tendon_label\` rewrite correctly per world.
- Cross-pass consistency: every renamed label lives under the per-env root.
- Guards: non-path strings pass through untouched; rows whose world id is not in \`env_ids\` keep their original label.
- \`test_sparse_env_ids\` — non-contiguous env ids \`[10, 20, 30]\`.
- \`TestRenamePass2Generality\` — multiple coexisting custom frequencies, multiple string columns at one frequency, registered-but-empty string column.
- \`TestRenameMultiSource::test_prefix_overlap_does_not_cross_contaminate\` — explicit regression for the \`/Sources/protoA\` vs \`/Sources/protoAB\` boundary fix; both sources feed the same envs so the world-id guard cannot mask the boundary bug. Fails without the fix; passes with it.

## Test plan

- [x] All 10 unit tests pass.
- [x] \`./isaaclab.sh -f\` clean (pre-commit hooks).
- [x] Verified the boundary-prefix regression test fails when the boundary terminator is removed and passes when it's restored.
- [x] Smoke (Shadow-Hand-Over MAPPO 4 envs / iter 1) shows tendon labels go from \`/World/envs/env_0/.../T_FFJ\` (every env) to \`/World/envs/env_<wid>/.../T_FFJ\` (per-env paths) after the rename.


hujc7:
@greptileai Review

hujc7:
@greptileai Review

source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py
```diff
@@ -19,6 +19,35 @@
...
+    Args:
+        builder: A proto ``ModelBuilder`` that has already had custom attributes
+            registered (e.g. via ``SolverMuJoCo.register_custom_attributes``).
+        root_path: USD path prefix; only prims whose path starts with this string
+            will pass the patched filter.
+    """
+    for freq in builder.custom_frequencies.values():
+        if freq.usd_prim_filter is None:
+            continue
+        orig = freq.usd_prim_filter
```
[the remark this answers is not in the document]
hujc7: Resolved by removing the shim entirely. The current PR bumps the Newton pin to `v1.2.0rc1`, which includes upstream [newton-physics/newton#2659](https://github.com/newton-physics/newton/pull/2659) — `parse_usd` now scopes the custom-frequency walk to `root_path` natively, so the IL-side `_scope_custom_frequencies` workaround is no longer needed.

The same boundary-prefix concern was applied to the surviving `_rename_builder_labels` function: the prefix match now uses `src_path.rstrip("/") + "/"` so a source path that is a string prefix of another (`/Sources/protoA` vs `/Sources/protoAB`) does not cross-contaminate. See `source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py:143`.

source/isaaclab_newton/test/cloner/test_mjc_tendon_cloner.py
```diff
@@ -0,0 +1,464 @@
...
+        _scope_custom_frequencies(b, self.ROOT_A)
+
+    def test_heterogeneous_plan_no_cross_contamination(self):
+        """Each proto only accumulates tendon entries from its own source subtree.
+
+        Simulates the heterogeneous scenario: two proto builders, each scoped to
+        a different source path.  After both receive tendon entries representing
+        their own prims (as the scoped traversal would inject), neither contains
+        entries from the other source.
+        """
```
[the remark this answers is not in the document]
hujc7: The `test_mjc_tendon_cloner.py` file is removed in the current PR (the shim it tested was dropped after the Newton bump). 

The boundary-prefix correctness concern is now exercised in the new `test_rename_builder_labels.py::TestRenameMultiSource::test_prefix_overlap_does_not_cross_contaminate`, which constructs a multi-source plan with `/Sources/protoA` and `/Sources/protoAB` both feeding the same envs and verifies neither over-matches the other. The test fails without the boundary-prefix fix and passes with it.

kellyguo11:
@AntoineRichard @StafaH could you help review this one?

hujc7:
@greptileai Review

source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py
```diff
@@ -136,21 +145,55 @@ def _rename_builder_labels(
...
+            for k in range(len(values)):
+                world_id = int(worlds[k])
+                if world_id in world_roots and isinstance(values[k], str) and values[k].startswith(src_prefix):
+                    values[k] = swap(values[k], world_roots[world_id])
+
+        # Pass 1: built-in label arrays. Each has a paired ``*_world`` int column.
+        # Use ``is None`` (not ``or``) so an empty-but-defined ``*_label`` column
+        # is recognized — falling through to ``*_key`` would over-match a
+        # builder that legitimately exposes both attributes.
+        for t in ("body", "joint", "shape", "articulation", "constraint_mimic", "equality_constraint"):
```
StafaH: Let's pull these out as a constant _BUILTIN_LABEL_TYPES? That way they can be re-used especially for tests

source/isaaclab_newton/test/cloner/test_rename_builder_labels.py
```diff
@@ -0,0 +1,281 @@
...
+import torch
+from isaaclab_newton.cloner.newton_replicate import _rename_builder_labels
+from newton.solvers import SolverMuJoCo
+
+_TENDON_FREQ = "mujoco:tendon"
+_SRC = "/Sources/protoA"
+_DST = "/World/envs/env_{}"
+
+
+# ─── helpers ─────────────────────────────────────────────────────────────────
```
StafaH: Let's remove the excessive comment decoration

source/isaaclab_newton/test/cloner/test_rename_builder_labels.py
```diff
@@ -0,0 +1,281 @@
...
+    b = newton.ModelBuilder()
+    SolverMuJoCo.register_custom_attributes(b)
+    _inject_builtins(
+        b, ("body", "joint", "shape", "articulation", "constraint_mimic", "equality_constraint"), _SRC, worlds
+    )
+    _inject_tendon_strings(b, _SRC, worlds)
+    return b
+
+
+# ─── tests ───────────────────────────────────────────────────────────────────
```
StafaH: Same as above and elsewhere

source/isaaclab_newton/isaaclab_newton/cloner/newton_replicate.py
```diff
@@ -136,21 +145,55 @@ def _rename_builder_labels(
     """
     # per-source, per-world renaming (strict prefix swap), compact style preserved
     for i, src_path in enumerate(sources):
-        src_prefix_len = len(src_path.rstrip("/"))
+        # Boundary-terminated prefix prevents over-matching when one source path is a
+        # prefix of another (e.g. ``/Sources/protoA`` vs ``/Sources/protoAB``).
```
StafaH: What about the behavior for when the label matches exactly the source root? The old code used to cover that case but the new code does not

StafaH (approved):
LGTM, just left minor comments

PR #5444, opened 2026-04-29
Fix silent ABI mismatch in cubric Python shim

jmart-nv:
# Description

Background: The _cubric.py ctypes shim was pinned to IAdapter v0.1 vtable offsets, but newer Kit builds ship v0.2 — compute calls were silently landing on unbind, disabling cubric's GPU transform hierarchy propagation. carb accepts the version mismatch with only a stderr warning.

Originally, this change updated offsets to the v0.2 layout, requested v0.2 from the framework, and added a runtime InterfaceDesc check that refused to acquire on any unexpected version.

The kit team is fixing the ABI-breaking semver contract violation upstream, so it won't actually make it into a release. So the pinned version in Isaac Lab remains on v0.1 but keeps the validation code as a safety net.

This problem will go away once we have official python bindings for cubric in a future kit release.

If usdrt eventually exposes the required `eRigidBody` options via the `IFabricHierarchy` API then that would massively simplify the implementation of newton manager. Will pursue a feature request.

## Type of change

## Checklist

- [ ] I have made corresponding changes to the documentation *(N/A)*
- [ ] My changes generate no new warnings *(New warnings on ABI mismatch are intentional)*
- [ ] I have added tests that prove my fix is effective or that my feature works *(N/A - spoofing kit versions for unit test would be non-trivial; verified manually)*


pbarejko (approved):

source/isaaclab_newton/isaaclab_newton/physics/_cubric.py
```diff
@@ -222,6 +241,55 @@ def initialize(self) -> bool:
...
+            return False
+
+        interfaces_ptr = _read_u64(plugin_desc_ptr + _PD_OFF_INTERFACES)
+        interface_count = _read_u64(plugin_desc_ptr + _PD_OFF_INTERFACE_COUNT)
+        if interfaces_ptr == 0 or interface_count == 0:
+            logger.warning("PluginDesc reports zero interfaces for cubric plugin")
+            return False
+
+        for i in range(interface_count):
+            entry_addr = interfaces_ptr + i * _INTERFACE_DESC_STRIDE
```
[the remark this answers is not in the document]
jmart-nv: Added bound of 64

source/isaaclab_newton/isaaclab_newton/physics/_cubric.py
```diff
@@ -31,27 +31,34 @@
...
 # ---------------------------------------------------------------------------
 # Counting only CARB_ABI fields from the top of ``struct Framework``:
-#   0: loadPluginsEx
-#   8: unloadAllPlugins
-#  16: acquireInterfaceWithClient
-#  24: tryAcquireInterfaceWithClient  ← we use this one
+#   24 = tryAcquireInterfaceWithClient
+#   96 = getInterfacePluginDesc
 _FW_OFF_TRY_ACQUIRE = 24
+_FW_OFF_GET_INTERFACE_PLUGIN_DESC = 96
```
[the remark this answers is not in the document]
jmart-nv: Clarified the full vtable in comment

source/isaaclab_newton/isaaclab_newton/physics/_cubric.py
```diff
@@ -222,6 +241,55 @@ def initialize(self) -> bool:
         logger.info("cubric IAdapter bindings ready")
         return True
 
+    @staticmethod
+    def _verify_iadapter_version(fw_ptr: int, ia_ptr: int) -> bool:
```
gevans-nvda, line 250: I don't think we should go so far as to distrust minor interface increments because they broke this one time. If and when the minor version of IAdapter is bumped again, that should not part-and-parcel come with having to babysit this pathway of code, to essentially allowlist versions of cubric that will preserve expected behavior.

jmart-nv: Higher minor version mismatches will now log a loud warning and proceed, respecting the minor semver contract. Lower minor versions and mismatched major versions will still fall back on the CPU (non-cubric) path, but carb should prevent that from happening in the first place.

pbarejko: After carefully reading this, I think I am with @gevans-nvda, he raises a good point here. The vtable + minor update was one time situation and it shouldn't happen in the future. If we switch, in near future, to new kit version, with newer minor version of `IAdapter` interface, we will observe big perf regression, with fallback to CPU. I think it's okay to relax this to trust the minor version bump.

> Higher minor version mismatches will now log a loud warning and proceed.

Isaac Lab in some instances can produce quite a lot of warnings, and the laud warning could not be as laud as you would want.

jmart-nv: Made this a louder (stderr) warning

jmart-nv:
kit will be un-breaking the semver minor version compatibility contract violation upstream, so the breaking ABI change will never actually make it into a kit release. Reverted our shim back to requesting v0.1, relaxed the warning to use the logger instead, but kept the validation code as a safeguard.

jmart-nv:
@isaaclab-review-bot updated the PR description

PR #5445, opened 2026-04-29
Add per-env omni:scenePartition authoring in InteractiveScene

ooctipus:
# Description

Adds `InteractiveScene.author_scene_partitions()` that, for each env root in `env_prim_paths`, writes the inheriting primvar `primvars:omni:scenePartition` on the root and the matching non-primvar `omni:scenePartition` token on every per-env `UsdGeom.Camera`. RTX (and OvRTX) honor primvar inheritance, so the env-root primvar propagates to all descendant geometry; renderers that don't recognize the primvar simply ignore it, so authoring is unconditional.

The method is called at the end of `clone_environments()`, so both `ManagerBasedEnv` (which calls `InteractiveScene.__init__` to auto-clone) and Direct envs that drive cloning manually from `_setup_scene` (e.g. `anymal_c_env`) get partitioning for free. Authoring runs after USD/physics cloning but before the physics backend takes its tensor view in `sim.reset`, so PhysX and Newton tensor views are not invalidated. Cost is `O(N + cameras)`; per-descendant authoring is intentionally not done because RTX is spec'd to honor primvar inheritance here.

## Type of change

## Tests

Two regression tests under `source/isaaclab/test/scene/test_scene_partitioning.py`:

| Test | Status today | Behavior on RTX fix |
| ---- | ------------ | ------------------- |
| `test_partitioning_isolates_rigid_object` | **PASSES** — dexsuite Lift with the robot hidden, RigidObject tile diff ≈ 13 (well above the 5.0 threshold). | Continues to pass. |
| `test_partitioning_isolates_articulation` | **XFAILS** — dexsuite Lift with object + table hidden, robot tile diff ≈ 2.2 (same as partitioning OFF; RTX articulation path doesn't honor the primvar today). | XPASSes once the RTX articulation path is fixed; signal to drop the xfail mark. |

Confirmed via ablation against `Isaac-Dexsuite-Kuka-Allegro-Lift-v0` at `num_envs=4`, `env_spacing=0`, `replicate_physics=False`, with selected per-env prims hidden via `UsdGeom.Imageable.MakeInvisible` (visibility-only — preserves PhysX tensor view validity).

The articulation-side bug is confirmed by Nathan Cournia; Steven Bloemer / FSD has the RTX-side fix in flight. The xfail test is the regression hook for that fix landing.

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/utils/camera_view.py
+++ b/source/isaaclab/isaaclab/envs/utils/camera_view.py
@@ -14,6 +14,8 @@ from typing import Any
 import torch
 import warp as wp
 
+from pxr import Sdf
+
 import isaaclab.sim as sim_utils
 from isaaclab.sensors.camera import Camera, CameraCfg
 
@@ -117,6 +119,14 @@ def create_visualizer_camera(
     for path in generated_paths:
         if len(sim_utils.find_matching_prims(path)) == 0:
             spawn.func(path, spawn, translation=(0.0, 0.0, 0.0), orientation=(0.0, 0.0, 0.0, 1.0))
+
+    stage = sim_utils.get_current_stage()
+    for path in generated_paths:
+        cam_prim = stage.GetPrimAtPath(path)
+        attr = cam_prim.GetAttribute("omni:scenePartition")
+        if not attr.IsValid():
+            attr = cam_prim.CreateAttribute("omni:scenePartition", Sdf.ValueTypeNames.Token)
+        attr.Set(path.split("/")[-2])
     cfg = CameraCfg(
         prim_path=f"/World/envs/env_.*/{camera_name}",
         update_period=0.0,
--- a/source/isaaclab_ov/changelog.d/zhengyuz-scene-partitioning.skip
+++ b/source/isaaclab_ov/changelog.d/zhengyuz-scene-partitioning.skip
@@ -1,0 +1,1 @@
+No user-facing isaaclab_ov changelog entry needed for scene partitioning test changes.
--- a/apps/isaaclab.python.headless.kit
+++ b/apps/isaaclab.python.headless.kit
@@ -82,6 +82,7 @@ app.useFabricSceneDelegate = false
 
 # Enable Iray and pxr by setting this to "rtx,iray,pxr"
 renderer.enabled = "rtx"
+renderer.scenePartitioning.enabled = true # Enable before RTX initialization so env-root partitions are honored
 
 # Avoid warning on shutdown from audio context
 app.audio.enabled = false
--- a/apps/isaaclab.python.headless.rendering.kit
+++ b/apps/isaaclab.python.headless.rendering.kit
@@ -41,6 +41,7 @@ app.version = "3.0.0"
 app.useFabricSceneDelegate = true
 # Temporary, should be enabled by default in Kit soon
 rtx.hydra.readTransformsFromFabricInRenderDelegate = true
+renderer.scenePartitioning.enabled = true # Enable before RTX initialization so env-root partitions are honored
 
 # Disable print outs on extension startup information
 # this only disables the app print_and_log function
--- a/apps/isaaclab.python.kit
+++ b/apps/isaaclab.python.kit
@@ -122,6 +122,7 @@ exts."omni.kit.viewport.window".windowMenu.label = "" # Put Viewport menuitem un
 exts."omni.rtx.window.settings".window_menu = "Window" # Where to put the render settings menuitem
 exts."omni.usd".locking.onClose = false # reduce time it takes to close/create stage
 renderer.asyncInit = true # Don't block while renderer inits
+renderer.scenePartitioning.enabled = true # Enable before RTX initialization so env-root partitions are honored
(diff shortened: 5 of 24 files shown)


kellyguo11 (approved):

huidongc (approved):

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -353,6 +357,42 @@ def _aggregate_scene_data_requirements(self, visualizer_types=()) -> None:
...
+                    continue
+                token = env_path.rsplit("/", 1)[-1]
+                for prim in Usd.PrimRange(env_prim):
+                    if prim == env_prim:
+                        attr_path = prim.GetPath().AppendProperty("primvars:omni:scenePartition")
+                    elif prim.IsA(UsdGeom.Camera):
+                        attr_path = prim.GetPath().AppendProperty("omni:scenePartition")
+                    else:
+                        continue
+                    Sdf.JustCreatePrimAttributeInLayer(root_layer, attr_path, token_type, variability, False)
```
huidongc: Could you try this change on your local machine?

```py
    Sdf.JustCreatePrimAttributeInLayer(root_layer, attr_path, token_type, Sdf.VariabilityUniform, True)
```

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -353,6 +357,41 @@ def _aggregate_scene_data_requirements(self, visualizer_types=()) -> None:
         if requirements != current_req:
             self.sim.update_scene_data_requirements(requirements)
 
+    def author_scene_partitions(self) -> None:
```
huidongc: RE your question in our meeting about the best place to land the `author_scene_partition`()` function considering it is RTX-specific requirement and shouldn't be in a common place here:

How do you feel if we just put it in the IsaacRTXRenderer.prepare_stage() for now? This renderer is likely going to be replaced by OVRTX renderer as time goes on. So it is probably fine not having a shared base class for the two renderer classes that can host the shared code - like this function.

@pbarejko for vis.

huidongc: Or maybe we define this method as a shared util/helper that OVRTX and RTX can call into from their own `prepare_stage()` function to dedup.

pbarejko: I agree with @huidongc - those are renderer's implementation details what attributes it consumes. However we should consider pre/post clone hooks. It would be great to look into this from design perspective:
1. Does any renderer need to mutate pre-cloned stage? - based on this PR the answer is yes.
2. Does any renderer need to mutate post-cloned stage? - I don't recall if we do. I would leave it @huidongc to decide

ooctipus: I think prepare_stage() makes sense!

source/isaaclab/isaaclab/envs/utils/camera_view.py
```diff
@@ -117,6 +119,14 @@ def create_visualizer_camera(
     for path in generated_paths:
         if len(sim_utils.find_matching_prims(path)) == 0:
             spawn.func(path, spawn, translation=(0.0, 0.0, 0.0), orientation=(0.0, 0.0, 0.0, 1.0))
+
+    stage = sim_utils.get_current_stage()
```
huidongc, line 123: Could you add some comment in the code to explain what the code block here is intended for? It is not as obvious to me. :)

huidongc: One thing I don't follow: the visualization camera is supposed to show all the envs, how do we need to modify the env cameras for visualization purpose?

huidongc (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -223,7 +222,7 @@ def prepare_stage(self, stage: Any, num_envs: int) -> None:
             return
 
         logger.info("Preparing stage for export (%d envs, cloning=%s)...", num_envs, self._use_ovrtx_cloning)
-        create_scene_partition_attributes(stage, num_envs, self._use_ovrtx_cloning)
+        # create_scene_partition_attributes(stage, num_envs, self._use_ovrtx_cloning)
```
huidongc: Please restore this call.

source/isaaclab_ov/changelog.d/zhengyuz-scene-partitioning.skip
```diff
@@ -0,0 +1 @@
+No user-facing isaaclab_ov changelog entry needed for scene partitioning test changes.
```
huidongc, line 1: This file can be reverted.

source/isaaclab/isaaclab/scene/interactive_scene.py
huidongc: Octi, I see you are moving back to earlier approach. Why?

ooctipus: oh it must be a bad rebase...

PR #5470, opened 2026-05-01
Skip cloner perf test and fix ray caster intrinsics camera pose

ooctipus:
# Description

Two unrelated CI flakes in the same test infrastructure area, bundled because both are minimal targeted fixes.

### 1. `test_disabled_fabric_change_notifies_speedup_regression` (`source/isaaclab_physx/test/sim/test_cloner.py`)

This is a wall-clock perf regression test that's intended to run **locally only** — it asserts a >= 1.2× speedup of clone+reset with listener suspension. The result is platform-sensitive (deferred Fabric resync in `sim.reset` can offset the scene-time savings on some hardware), so it was meant to be skipped in CI.

The original guard was `if os.getenv(\"CI\", \"\").lower() in (\"true\", \"1\"): pytest.skip(...)` inside the test body. However, this project's CI doesn't set the `CI` env var — it selects tests via the `isaacsim_ci` pytest marker registered in `pyproject.toml` and applied module-wide via `pytestmark` at the top of `test_cloner.py`. Result: the env-var skip never fires, and the test runs (and occasionally flakes) on CI.

Fix: replace the dead env-var branch with a top-level `@pytest.mark.skip(...)` decorator so the test is collected and skipped unconditionally regardless of how CI selects tests. The correctness of the suspension mechanism is still covered by `test_disabled_fabric_change_notifies_toggles_ifabricusd_flag`, which is unaffected. Re-enable the perf test manually when touching listener suspension.

### 2. `test_output_equal_to_usd_camera_when_intrinsics_set` (`source/isaaclab/test/sensors/test_ray_caster_camera.py`)

Intermittent CI failure where the ray caster camera output mismatched the USD camera reference with inf-valued differences across all 518400 elements:

```
E   Mismatched elements: 518400 / 518400 (100.0%)
E   Greatest absolute difference: inf at index (0, 0, 0, 0) (up to 0.0001 allowed)
E   Greatest relative difference: inf at index (0, 0, 0, 0) (up to 0.005 allowed)
```

Root cause: the test placed the camera at `eye=(0, 0, 5)` looking at `target=(0, 0, 0)`, which is colinear with the default up vector and produces a degenerate view transform. Fix: nudge `eye` to `(0.001, 0, 5)` for both the ray caster and USD camera, keeping them at identical poses while breaking the singularity. The underlying degeneracy is tracked in a separate internal ticket; this is the test-side mitigation.

Fixes # (n/a)

## Type of change

## Screenshots

N/A — test infrastructure changes.

## Checklist

### DIFF
--- a/source/isaaclab/test/sensors/test_ray_caster_camera.py
+++ b/source/isaaclab/test/sensors/test_ray_caster_camera.py
@@ -898,11 +898,11 @@ def test_output_equal_to_usd_camera_when_intrinsics_set(setup_sim, focal_length_
 
     # set camera position
     camera_warp.set_world_poses_from_view(
-        eyes=torch.tensor([[0.0, 0.0, 5.0]], device=camera_warp.device),
+        eyes=torch.tensor([[0.001, 0.0, 5.0]], device=camera_warp.device),
         targets=torch.tensor([[0.0, 0.0, 0.0]], device=camera_warp.device),
     )
     camera_usd.set_world_poses_from_view(
-        eyes=torch.tensor([[0.0, 0.0, 5.0]], device=camera_usd.device),
+        eyes=torch.tensor([[0.001, 0.0, 5.0]], device=camera_usd.device),
         targets=torch.tensor([[0.0, 0.0, 0.0]], device=camera_usd.device),
     )
 
--- a/source/isaaclab_physx/test/sim/test_cloner.py
+++ b/source/isaaclab_physx/test/sim/test_cloner.py
@@ -552,10 +552,17 @@ def test_disabled_fabric_change_notifies_toggles_ifabricusd_flag(sim):
     assert bindings.is_enabled(fabric_id), "outer exit should restore the flag"
 
 
+@pytest.mark.skip(
+    reason=(
+        "Local-only perf regression; correctness is covered by"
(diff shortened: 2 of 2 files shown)


kellyguo11 (approved):

source/isaaclab/test/sensors/test_ray_caster_camera.py
```diff
@@ -843,6 +843,7 @@ def test_output_equal_to_usd_camera_intrinsics(setup_sim, focal_length):
 
 @pytest.mark.parametrize("focal_length_aperture", [(0.193, 0.20955), (1.93, 2.0955), (19.3, 20.955), (0.193, 20.955)])
 @pytest.mark.isaacsim_ci
+@pytest.mark.flaky(max_runs=3, min_passes=1)
```
pbarejko: We know what is the problem with this one. I created an internal ticket for it. I would just change:

```
 eyes=torch.tensor([[0.0, 0.0, 5.0]], device=camera_warp.device),
```
to 
```
 eyes=torch.tensor([[0.001, 0.0, 5.0]], device=camera_warp.device),
```
and remove the flakiness flag.

pbarejko (approved):

PR #5477, opened 2026-05-03
Fix FrameView staleness for sensors parented under physics bodies

ooctipus:
## Summary

- Fixes the bug demonstrated by #5476: FrameView-based sensors (RayCaster, Camera, IMU under articulation / rigid bodies) silently returned their spawn-time pose forever in headless training. Height-scan obs in rough-terrain locomotion was a frozen patch from spawn.
- Root cause: #5179 switched RayCaster from PhysX tensor views to FrameView (Fabric). The PhysX → Fabric write path is gated by three independent flags, all off in headless: `/app/useFabricSceneDelegate`, `/physics/fabricUpdateTransformations`, and a per-step `_update_fabric(dt, time)` call.
- Companion regression test in #5476 goes green with this PR.

## Changes

| file | what |
|---|---|
| `source/isaaclab_physx/.../physx_manager.py` | `PhysicsManager.step` calls `_update_fabric(dt, sim_time)` after each `simulate/fetch_results` with monotonic `_fabric_sim_time`. Body-write block is gated by the carb setting — call is free when nothing's registered. |
| `source/isaaclab/isaaclab/app/app_launcher.py` | Stop unconditionally setting `/physics/fabricUpdateTransformations` to `_rendering_enabled()`. Respect the value if a kit file / kit_args / `sim_launcher` already set it True. |
| `source/isaaclab_tasks/.../sim_launcher.py` | New `_needs_fabric_reads` predicate scanning `env_cfg` for `RayCasterCfg` / `CameraCfg`. When detected, `launch_simulation` injects `--/app/useFabricSceneDelegate=1 --/physics/fabricUpdateTransformations=1` into `kit_args` before `AppLauncher` (FSD must be set at kit boot). Mirrors the existing `_is_kit_camera` auto-detect. |
| `compute_kit_requirements` signature | Now returns 4-tuple including `needs_fabric_reads`. Three internal test files updated. |
| `test_views_xform_prim_fabric.py` | Module-level `AppLauncher` passes the kit_args `sim_launcher` would inject. Regression test `test_world_pose_tracks_physics_body_parent` (also added in #5476) goes green. |

## Performance

`anymal_c` rough terrain, 1024 envs, 500 steps after 100 warmup, RTX 5090:

| variant | per-step | env-steps/s | vs baseline |
|---|---:|---:|---:|
| fix off (today's broken state) | 4.481 ms | 228k | baseline |
| **fix on** | **5.073 ms** | **201k** | **+13.2%** |
| no FrameView env | 4.456 ms | 230k | noise |

~13% physics-step cost when a heightmap / camera sensor is in the env (cost of pushing all anymal articulation transforms into Fabric every step). **Zero overhead for envs without FrameView sensors.** This is the cost of correct observations vs the silent training corruption we had.

## Why not just enable `--enable_cameras`?

That works but loads RTX shaders, hydra texture, replicator, and viewport extensions — adds seconds to startup and significant GPU memory for users who don't need rendering. The `_needs_fabric_reads` auto-detect uses only the FSD scene-delegate flag (free routing setting; no shaders, no extensions) plus the PhysX → Fabric writer toggle.

## What about Newton?

Newton has its own implementation in `newton_manager.sync_transforms_to_usd` that doesn't depend on `omni.physx.fabric`, so this PR does not affect the Newton backend. (Aside: Newton's sync is gated on `pre_render`, which suggests a similar bug for pure-headless Newton training — worth a separate investigation.)

## Test plan

- [ ] CI for `test_world_pose_tracks_physics_body_parent[cuda:0]` goes green.
- [ ] Existing `test_views_xform_prim_fabric.py` contract tests still pass (18 passed locally).
- [ ] `test_ray_caster_integration.py` still passes (6 passed locally).
- [ ] `Isaac-Velocity-Rough-Anymal-C-v0` height_scan obs now varies as the robot walks (verified by re-recording the heightmap video from before/after).

Linked: #5476

### DIFF
--- a/source/isaaclab/isaaclab/app/app_launcher.py
+++ b/source/isaaclab/isaaclab/app/app_launcher.py
@@ -1163,8 +1163,14 @@ class AppLauncher:
         # set setting to indicate no RTX sensors are used (set to True when RTX sensor is created)
         settings.set_bool("/isaaclab/render/rtx_sensors", False)
 
-        # set fabric update flag to disable updating transforms when rendering is disabled
-        settings.set_bool("/physics/fabricUpdateTransformations", self._rendering_enabled())
+        # Enable PhysX -> Fabric rigid-body / articulation-link transform writes when
+        # rendering is enabled. Don't stomp the setting back to False if it was already
+        # turned on by a kit experience or by ``sim_launcher`` after detecting a
+        # FrameView-based sensor (RayCaster / Camera) in the env config — those readers
+        # depend on Fabric being fresh in headless training too. ``settings.get`` may
+        # return ``1``/``0`` from kit_args overrides, so cast through ``bool``.
+        existing = bool(settings.get("/physics/fabricUpdateTransformations"))
+        settings.set_bool("/physics/fabricUpdateTransformations", self._rendering_enabled() or existing)
 
         # use fixed time stepping disabled; custom loop runner from Isaac Sim is used instead
         settings.set_bool("/app/player/useFixedTimeStepping", False)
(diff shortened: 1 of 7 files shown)


kellyguo11:
Properly addressed raycaster issue in https://github.com/isaac-sim/IsaacLab/pull/5510, but FrameView fabric path will not work with non-rendering headless workflows.

PR #5499, opened 2026-05-05
Apply PPISP in Camera Wrapper Across Renderers

moennen:
## Description
  
  Introduces a wrapper-side Physically Plausible Image Signal Processing (PPISP) pipeline on the Camera sensor that converts  the renderer's HDR AOV to LDR RGBA on every render tick. The pipeline applies responsivity → exposure → vignetting →  color homography → camera response function → uint8 clamp in a single Warp kernel, called once per tick from the camera.
The kernel runs after the renderer fills the HDR buffer and writes back into the user-requested rgb/rgba output, so the rest of the pipeline stays untouched.

The integration is renderer-agnostic. New public surface in isaaclab.sensors.camera:

  - CameraPPISPCfg — explicit ISP coefficients (or a USD shader prim path).
  - CameraISPMode — AUTO_CAMERA (find an ISP shader bound to this camera via a RenderProduct), AUTO_ANY (fall back to the  first PPISP shader anywhere on stage).
  - CameraISP — owns HDR-buffer allocation and per-tick kernel dispatch.
  - CameraCfg.isp_cfg — accepts CameraPPISPCfg | CameraISPMode | None; defaults to CameraISPMode.AUTO_ANY.

Renderer contract extensions (in isaaclab.renderers):

  - RenderBufferKind.RGB_HDR (3-channel float, scene-linear).
  - RendererCfg.apply_pre_reset_settings(*, require_hdr_output) — backend hook to author global carb settings before sim.reset().
  - BaseRenderer.prepare_cameras(stage, spec) — backend hook for per-camera USD setup.
  - CameraRenderSpec — immutable snapshot of a tiled camera bundle so backends don't hold a reference to the sensor.

  Backend changes:

  - IsaacRtxRenderer (isaaclab_physx): advertises RGB_HDR via the Replicator HdrColor annotator; prepare_cameras authors a  neutral OmniRtxCameraExposureAPI_1 schema on each camera prim so RTX-side tonemapping does not double-process the ISP output; apply_pre_reset_settings sets /isaaclab/render/rtx_sensors=True and flips /rtx/rtpt/gaussian/skipTonemapping/enabled=False when HDR is requested (required to keep 3DGS splats in HDR).
  - OVRTXRenderer (isaaclab_ov): mirrors the IsaacRtx changes (HDR render var, neutral exposure schema, skipTonemapping).
  The exposure-override helper is intentionally duplicated rather than imported across sibling packages during the release window.
  - NewtonWarpRenderer (isaaclab_newton): advertises RGB_HDR (its scene-linear color is already what we need).

## Dependencies

No new runtime dependencies. Uses existing IsaacLab dependencies: Warp, Torch, USD, and renderer-specific backends.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Screenshots
Not applicable.

## Checklist
- [x] I have run the pre-commit checks with `./isaaclab.sh --format`

### DIFF
--- a/source/isaaclab/isaaclab/renderers/base_renderer.py
+++ b/source/isaaclab/isaaclab/renderers/base_renderer.py
@@ -25,6 +25,21 @@ class BaseRenderer(ABC):
         """Post-physics one-time initialization hook. Called only once."""
         return
 
+    def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
+        """Pre-render per-camera setup the backend needs.
+
+        The default implementation is a no-op. Renderer subclasses override
+        to perform whatever per-camera initialization their backend requires
+        — e.g. authoring stage attributes on the resolved camera prims,
+        configuring per-tile GPU buffers, or any other state setup.
+
+        Args:
+            stage: Scene stage the camera prims live on, or ``None``
+                when no stage context applies. Stage-less backends ignore it.
+            spec: Immutable description of the tiled camera bundle.
+        """
+        return
+
     @abstractmethod
     def supported_output_types(self) -> dict[RenderBufferKind, RenderBufferSpec]:
         """Per-output layout (channels + dtype) this renderer can produce.
--- a/source/isaaclab/isaaclab/sensors/camera/camera.py
+++ b/source/isaaclab/isaaclab/sensors/camera/camera.py
@@ -17,7 +17,6 @@ from pxr import UsdGeom
 
 import isaaclab.sim as sim_utils
 import isaaclab.utils.sensors as sensor_utils
-from isaaclab.app.settings_manager import get_settings_manager
 from isaaclab.renderers import BaseRenderer, CameraRenderSpec
 from isaaclab.sim.views import FrameView
 from isaaclab.utils import to_camel_case
@@ -148,16 +147,7 @@ class Camera(SensorBase):
         # initialize base class
         super().__init__(cfg)
 
-        # TODO(follow-up PR): move this flag flip out of Camera. The cleanest path is
-        # an apply_pre_reset_settings() hook on RendererCfg (default no-op) that
-        # IsaacRtxRendererCfg overrides to flip /isaaclab/render/rtx_sensors. The
-        # flag must be set pre-sim.reset() because SimulationContext.is_rendering
-        # and several env classes read it before the renderer's __init__ runs.
-        renderer_type = getattr(self.cfg.renderer_cfg, "renderer_type", None)
-        if renderer_type == "isaac_rtx":
-            get_settings_manager().set_bool("/isaaclab/render/rtx_sensors", True)
-
-        # Compute camera orientation (convention conversion) and spawn
+        # Compute camera orientation (convention conversion) and spawn.
         rot = torch.tensor(self.cfg.offset.rot, dtype=torch.float32, device="cpu").unsqueeze(0)
         rot_offset = convert_camera_frame_orientation_convention(
             rot, origin=self.cfg.offset.convention, target="opengl"
@@ -166,6 +156,34 @@ class Camera(SensorBase):
         if self.cfg.spawn is not None and self.cfg.spawn.vertical_aperture is None:
             self.cfg.spawn.vertical_aperture = self.cfg.spawn.horizontal_aperture * self.cfg.height / self.cfg.width
         self._resolve_and_spawn("camera", translation=self.cfg.offset.pos, orientation=rot_offset)
+
+        # An ISP (any ``isp_cfg`` other than ``None``) requires the HDR AOV;
+        # an explicit ``"rgb_hdr"`` in ``data_types`` also requires the
+        # HDR-routing flag flipped on the RTX-bearing backends.
+        require_hdr_output = "rgb_hdr" in self.cfg.data_types or self.cfg.isp_cfg is not None
+
+        # TODO(follow-up PR): move this flag flip out of Camera. The cleanest path is
+        # an apply_pre_reset_settings() hook on RendererCfg (default no-op) that
+        # IsaacRtxRendererCfg overrides to flip /isaaclab/render/rtx_sensors. The
+        # flag must be set pre-sim.reset() because SimulationContext.is_rendering
+        # and several env classes read it before the renderer's __init__ runs.
+        renderer_type = getattr(self.cfg.renderer_cfg, "renderer_type", None)
+        if renderer_type == "isaac_rtx":
+            from isaaclab.app.settings_manager import get_settings_manager
+
+            settings = get_settings_manager()
+            settings.set_bool("/isaaclab/render/rtx_sensors", True)
+            if require_hdr_output:
+                settings.set_bool("/rtx/rtpt/gaussian/skipTonemapping/enabled", False)
+        elif renderer_type == "ovrtx" and require_hdr_output:
+            from isaaclab.app.settings_manager import get_settings_manager
+
+            get_settings_manager().set_bool("/rtx/rtpt/gaussian/skipTonemapping/enabled", False)
+            # FIXME: settings set_bool is a no-op for ovrtx
+            # warning only since it affects only ParticleField3DGaussianSplat scene
+            logger.warning(
+                "OVRTX backend with PPISP/HDR requires /rtx/rtpt/gaussian/skipTonemapping/enabled to be false."
+            )
 
         # UsdGeom Camera prim for the sensor
         self._sensor_prims: list[UsdGeom.Camera] = list()
@@ -453,6 +471,30 @@ class Camera(SensorBase):
         self._renderer = sim_ctx.render_context.get_renderer(self.cfg.renderer_cfg)
         logger.info("Using renderer: %s", type(self._renderer).__name__)
 
+        # Build the render spec early — both the wrapper ISP (which delegates
+        # any renderer-side per-camera setup) and ``create_render_data`` consume
+        # it, and the prims are already authored at this point.
+        cam_paths = tuple(str(p.GetPath()) for p in sim_utils.find_matching_prims(self.cfg.prim_path, self.stage))
+        env_0_prefix = "/World/envs/env_0/"
+        rel_under_env0 = (
+            cam_paths[0].removeprefix(env_0_prefix) if cam_paths and cam_paths[0].startswith(env_0_prefix) else ""
+        )
+        device_str = self._device if isinstance(self._device, str) else str(self._device)
+        render_spec = CameraRenderSpec(
+            cfg=self.cfg,
+            device=device_str,
+            num_instances=len(cam_paths),
+            camera_prim_paths=cam_paths,
+            view_count=len(cam_paths),
+            camera_path_relative_to_env_0=rel_under_env0,
+        )
+
+        # Delegate per-camera USD setup to the renderer — must run **before**
+        # ``ensure_prepare_stage`` so renderers that snapshot the stage
+        # (ovrtx's ``stage.Export``) capture the resulting overrides in their
+        # exported USD.
+        self._renderer.prepare_cameras(self.stage, render_spec)
+
         # Stage preprocessing must happen before creating the view because the view keeps
         # references to prims located in the stage.
         sim_ctx.render_context.ensure_prepare_stage(self.stage, self._num_envs)
@@ -480,21 +522,6 @@ class Camera(SensorBase):
             # Add to list
             self._sensor_prims.append(UsdGeom.Camera(cam_prim))
 
-        # View needs to exist before creating render data
-        cam_paths = tuple(cam_prim.GetPath().pathString for cam_prim in self._view.prims)
-        env_0_prefix = "/World/envs/env_0/"
-        rel_under_env0 = (
-            cam_paths[0].removeprefix(env_0_prefix) if cam_paths and cam_paths[0].startswith(env_0_prefix) else ""
-        )
-        device_str = self._device if isinstance(self._device, str) else str(self._device)
-        render_spec = CameraRenderSpec(
-            cfg=self.cfg,
-            device=device_str,
-            num_instances=self.num_instances,
-            camera_prim_paths=cam_paths,
-            view_count=self._view.count,
-            camera_path_relative_to_env_0=rel_under_env0,
-        )
         self._render_data = self._renderer.create_render_data(render_spec)
 
         # Create internal buffers (includes intrinsic matrix and pose init)
--- a/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
+++ b/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
@@ -24,6 +24,8 @@ from ..physics.newton_manager import NewtonManager
 from .newton_warp_renderer_cfg import NewtonWarpRendererCfg
 
 if TYPE_CHECKING:
+    from isaaclab_ppisp import PpispPipeline
+
     from isaaclab.sensors.camera.camera_data import CameraData
     from isaaclab.utils.warp import ProxyArray
 
@@ -39,6 +41,7 @@ class RenderData:
     # but the Newton sensor API consumes it as (world_count,1,H,W) uint32 (same bytes, packed view).
     _OUTPUT_MAP: dict[str, tuple[str, type]] = {
         str(RenderBufferKind.RGBA): ("color_image", wp.uint32),
+        str(RenderBufferKind.RGB_HDR): ("hdr_color_image", wp.vec3f),
         str(RenderBufferKind.ALBEDO): ("albedo_image", wp.uint32),
         str(RenderBufferKind.DEPTH): ("depth_image", wp.float32),
         str(RenderBufferKind.NORMALS): ("normals_image", wp.vec3f),
@@ -48,6 +51,7 @@ class RenderData:
     @dataclass
     class CameraOutputs:
         color_image: wp.array(dtype=wp.uint32, ndim=4) = None
+        hdr_color_image: wp.array(dtype=wp.vec3f, ndim=4) = None
         albedo_image: wp.array(dtype=wp.uint32, ndim=4) = None
         depth_image: wp.array(dtype=wp.float32, ndim=4) = None
         normals_image: wp.array(dtype=wp.vec3f, ndim=4) = None
@@ -63,6 +67,24 @@ class RenderData:
         self.outputs = RenderData.CameraOutputs()
         self.width = getattr(spec.cfg, "width", 100)
         self.height = getattr(spec.cfg, "height", 100)
+        # Post-render PPISP pipeline composed when ``spec.cfg.isp_cfg`` is set.
+        # ``isp_cfg`` is already fully normalised by Camera by the time it reaches here.
+        self.ppisp_pipeline: PpispPipeline | None = None
+        if spec.cfg.isp_cfg is not None:
+            from isaaclab_ppisp import PpispPipeline
+
+            self.ppisp_pipeline = PpispPipeline(spec.cfg.isp_cfg)
+        self._hdr_scratch_wp: wp.array | None = None
+        """Internal HDR scratch buffer allocated when PPISP is composed but the
+        user did not request ``"rgb_hdr"`` in ``data_types``. Also exposed to
+        the Newton sensor through :attr:`CameraOutputs.hdr_color_image` as a
+        vec3f reinterpretation of this same backing storage."""
+        self._ppisp_hdr_source: wp.array | None = None
+        """PPISP HDR source bound once in :meth:`set_outputs` from the caller's
+        ``rgb_hdr`` output or :attr:`_hdr_scratch_wp`."""
+        self._ppisp_rgba_dest: wp.array | None = None
+        """PPISP LDR destination bound once in :meth:`set_outputs` from the
+        caller's ``rgba`` output."""
 
     def set_outputs(self, output_data: dict[str, ProxyArray]):
         shape = (self.newton_sensor.model.world_count, self.num_cameras, self.height, self.width)
@@ -79,10 +101,41 @@ class RenderData:
                 field_name,
                 wp.array(ptr=wp_arr.ptr, dtype=dtype, shape=shape, device=wp_arr.device, copy=False),
             )
+        # When PPISP is composed but the user did not request the raw HDR AOV,
+        # allocate an internal HDR scratch buffer and route a vec3f-shaped view
+        # of it as the Newton sensor's ``hdr_color_image`` so the renderer
+        # fills it directly.
+        if self.ppisp_pipeline is not None and self.outputs.hdr_color_image is None:
+            ref_proxy = next(iter(output_data.values()))
+            self._hdr_scratch_wp = wp.zeros(
+                (self.newton_sensor.model.world_count, self.height, self.width, 3),
+                dtype=wp.float32,
+                device=ref_proxy.device,
+            )
+            self.outputs.hdr_color_image = wp.array(
+                ptr=self._hdr_scratch_wp.ptr,
+                dtype=wp.vec3f,
+                shape=shape,
+                device=self._hdr_scratch_wp.device,
+                copy=False,
+            )
+        # Bind the two warp arrays the per-frame PPISP dispatch needs.
+        if self.ppisp_pipeline is not None:
+            if str(RenderBufferKind.RGBA) not in output_data:
+                raise ValueError(
+                    "Newton renderer ISP requires 'rgba' (or 'rgb', which aliases into rgba) as the"
+                    " LDR output destination, but neither was provided. Add 'rgb' or 'rgba' to"
+                    " Camera.cfg.data_types when isp_cfg is set."
+                )
+            hdr_proxy = output_data.get(str(RenderBufferKind.RGB_HDR))
+            self._ppisp_hdr_source = hdr_proxy.warp if hdr_proxy is not None else self._hdr_scratch_wp
+            self._ppisp_rgba_dest = output_data[str(RenderBufferKind.RGBA)].warp
 
     def get_output(self, output_name: str) -> wp.array:
         if output_name == RenderBufferKind.RGBA:
             return self.outputs.color_image
+        elif output_name == RenderBufferKind.RGB_HDR:
+            return self.outputs.hdr_color_image
         elif output_name == RenderBufferKind.ALBEDO:
             return self.outputs.albedo_image
         elif output_name == RenderBufferKind.DEPTH:
@@ -195,11 +248,25 @@ class NewtonWarpRenderer(BaseRenderer):
         return {
             RenderBufferKind.RGBA: RenderBufferSpec(4, wp.uint8),
             RenderBufferKind.RGB: RenderBufferSpec(3, wp.uint8),
+            RenderBufferKind.RGB_HDR: RenderBufferSpec(3, wp.float32),
             RenderBufferKind.ALBEDO: RenderBufferSpec(4, wp.uint8),
             RenderBufferKind.DEPTH: RenderBufferSpec(1, wp.float32),
             RenderBufferKind.NORMALS: RenderBufferSpec(3, wp.float32),
             RenderBufferKind.INSTANCE_SEGMENTATION_FAST: seg_spec,
         }
+
+    def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
+        """Resolve the camera's PPISP cfg before rendering.
+
+        :mod:`isaaclab.sensors.camera` does not depend on PPISP; the renderer
+        owns the sentinel-resolution + cfg-normalization step. Newton has no
+        USD-side overrides to author beyond this.
+        """
+        if not spec.camera_prim_paths:
+            return
+        from isaaclab_ppisp import resolve_and_normalize
+
+        spec.cfg.isp_cfg = resolve_and_normalize(spec.cfg.isp_cfg, stage, spec.camera_prim_paths[0])
 
     def prepare_stage(self, stage: Any, num_envs: int) -> None:
         """No-op for Newton Warp - uses Newton scene directly without stage export.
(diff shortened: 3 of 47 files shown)


source/isaaclab/isaaclab/renderers/camera_ppisp.py
```diff
@@ -0,0 +1,272 @@
...
+    shader_prim_path: str | None = None
+    """Optional source USD shader prim path used to populate :attr:`inputs`."""
+
+    inputs: dict[str, float | tuple[float, float]] = field(default_factory=default_camera_ppisp_inputs)
+    """Flat PPISP shader input values keyed by USD input name."""
+
+
+@dataclass
+class RenderProductInfo:
+    """Parsed USD RenderProduct information used for PPISP validation."""
```
pbarejko: Our intention for this module(`isaaclab/isaaclab/renderers/`) was to keep it back-end agnostic. You added a lot of features there that are back-end specific. Is USD Product something that is used for Newton Warp?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -117,6 +131,15 @@ def __init__(self, cfg: CameraCfg):
...
             get_settings_manager().set_bool("/isaaclab/render/rtx_sensors", True)
 
+        # When an HDR consumer is present (wrapper PPISP or an explicit ``rgb_hdr``
+        # data type), route ParticleField / 3DGS content through the HDR → tonemap
+        # pipeline so the ``HdrColor`` AOV is populated. By default RTX skips
+        # tonemap for splats. Must be set pre-``sim.reset()`` so the RTX pipeline
+        # picks it up before the first  render product is created.
+        # Applies to any RTX-backed renderer.
+        if renderer_type in ("isaac_rtx", "ovrtx") and (self.cfg.ppisp is not None or "rgb_hdr" in self.cfg.data_types):
+            get_settings_manager().set_bool("/rtx/rtpt/gaussian/skipTonemapping/enabled", False)
```
pbarejko: We put a lot of effort to separate rendering backends from camera, so the camera knows very little about the renderer that runs in the background. Now you are breaking our efforts and you started to bring implementation details to places where they don't belong in my opinion. Can we keep Camera renderer agnostic?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -376,6 +400,14 @@ def _initialize_impl(self):
         self._renderer = sim_ctx.render_context.get_renderer(self.cfg.renderer_cfg)
         logger.info("Using renderer: %s", type(self._renderer).__name__)
 
+        # Apply the neutral exposure + ``OmniRtxCameraExposureAPI_1`` schema on
```
pbarejko: Why this post processing can not be put into backends? It seems like you have two code paths. One is for warp and the other one is for RTX/OVRTX.

pbarejko (changes requested):
This work scatters implementation details almost everywhere.

source/isaaclab/isaaclab/sensors/camera/camera.py
moennen: Having this opimtized ISP implemented in all render backends is definitely the end goal. 
I do have a branch for that but : its impact is definitely too big for this release and there are several missing features RTX side to support it fully (tiled vignetting for example)

moennen:
> This work scatters implementation details almost everywhere.

@pbarejko 
Agree and sorry about that. I did refactor to comply to what I understood from your comments about the rendering design principles. 
Please let me know if I'm still missing something :pray:

source/isaaclab/isaaclab/renderers/camera_ppisp.py
moennen: Understood : moved and refactor camera changes into sensors module.

source/isaaclab/isaaclab/sensors/camera/camera.py
moennen: Understood : replaced all backend specific hook as 2 backend agnostic hooks.

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -26,6 +26,21 @@ def initialize(self) -> None:
         """Post-physics one-time initialization hook. Called only once."""
         return
 
+    def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
```
pbarejko, line 28: I think this method is redundant. This is what happens now:

```
isaac_rtx_renderer -> apply_isp_camera_overrides
ovrtx_renderer -> apply_isp_camera_overrides
newton_warp_renderer -> doesn't do anything.
```
You could use: `prepare_stage` and just pass camera config or camera spec?

moennen: Yes this is really tempting but the prepare_stage is called only once per-renderer.  If we have several cameras we want to call  prepare-camera once for every one. Thus it would require to call prepare_stage several times, IDK the codebase enough yet to decide if it is safe or not.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -42,3 +42,17 @@ class OVRTXRendererCfg(RendererCfg):
 
     log_file_path: str = "/tmp/ovrtx_renderer.log"
     """Path for OVRTX log file."""
+
+    def apply_pre_reset_settings(self, *, require_hdr_output: bool = False) -> None:
```
pbarejko: I don't think the configurations are meant to be used this way. They are ment to be serialized and de-serialized. At least this is my understanding.

If the above is true, `require_hdr_output` would be lost during serialization. Why we can not make it a field and apply appropriate setting in the rendering backend?

pbarejko: To add more context to this. Yesterday I merged #5573. It would allow you to set this setting before reset is called?

source/isaaclab/isaaclab/sensors/camera/camera_ppisp_warp.py
```diff
@@ -0,0 +1,300 @@
...
+    out_rgb[1] = _apply_crf(out_rgb[1], crf_toe_g, crf_shoulder_g, crf_gamma_g, crf_center_g)
+    out_rgb[2] = _apply_crf(out_rgb[2], crf_toe_b, crf_shoulder_b, crf_gamma_b, crf_center_b)
+
+    out_rgba[camera_id, height_id, width_id, 0] = wp.uint8(wp.clamp(out_rgb[0], 0.0, 1.0) * 255.0)
+    out_rgba[camera_id, height_id, width_id, 1] = wp.uint8(wp.clamp(out_rgb[1], 0.0, 1.0) * 255.0)
+    out_rgba[camera_id, height_id, width_id, 2] = wp.uint8(wp.clamp(out_rgb[2], 0.0, 1.0) * 255.0)
+    out_rgba[camera_id, height_id, width_id, 3] = wp.uint8(255)
+
+
+def apply_ppisp_to_rgba(hdr_color: torch.Tensor, out_rgba: torch.Tensor, cfg: CameraPPISPCfg) -> None:
```
pbarejko: This seems like an implementation detail of newton warp based rendering? I dealt with ISP in the past during AVSim project. Correct me if I am wrong, but conceptually this an additional pass we would want to run on the output from rendering. Why we can not push this to the renderer backend? I am curious about your motivation why to expose this to _public_? (I consider isaaclab/renderer, sensors/camera a place where public interfaces are)

moennen: As discussed : 

"Having this opimtized ISP implemented in all render backends is definitely the end goal.
I do have a branch for that but : its impact is definitely too big for this release and there are several missing features RTX side to support it fully (tiled vignetting for example)"

In other words this was the shortest and most secure path to have the feature in the release. When this will be correctly plugged in, I'm expecting this code to be swallowed up by Newton (and in my other branch it is already in Newton) ... RTX rendering will be done through SPG.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
moennen: require_hdr_output is not a renderer_cfg that should be serialized, it is driven by the scene and camera.

source/isaaclab/isaaclab/sensors/camera/camera_ppisp_warp.py
moennen: @pbarejko : I put in place the refactoring we discussed about.

source/isaaclab/isaaclab/renderers/renderer_cfg.py
```diff
@@ -13,3 +13,15 @@ class RendererCfg:
     """Configuration for a renderer."""
 
     renderer_type: str = "default"
+
+    def apply_pre_reset_settings(self, *, require_hdr_output: bool = False) -> None:
```
nvsekkin: suggestion:  I am guilty of the TODO comments suggesting this addition. one of the decisions we made is not to add methods into this class. there is work planned to move away from the lazy initialization of the renderers which should allow us to use their init methods for applying render config before sim reset.

this is the opposite direction but for now with a clear TODO should we put this in the Camera __init__? @pbarejko any thoughts on this?

nvsekkin: https://github.com/isaac-sim/IsaacLab/pull/5499/changes#diff-0a707c8334bb9561207df80d3f75b7bf611b45677a83da800b48bc866bf74020L111

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -16,6 +16,7 @@
 import numpy as np
 import torch
 import warp as wp
+from isaaclab_ppisp import PpispPipeline
```
nvsekkin: question: should this be a lazy-import? or do we actually want this as a hard dependency?

nvsekkin: there is already lazy-import here for example: https://github.com/isaac-sim/IsaacLab/pull/5499/changes#diff-905e05ce312e243b3b895b0f580ebea056758a2ad540c8f75aca5af01b09fe9aR104

source/isaaclab_ppisp/isaaclab_ppisp/rtx_camera_overrides.py
nvsekkin, line 1: suggestion: `source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_isp.py` and this file are effectively identical, is there a way to use a single file?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
nvsekkin: if we want to make this a truly optional backend we may need to make this lazy-import. 

then there will be another potential issue:

https://github.com/isaac-sim/IsaacLab/pull/5499/changes#diff-686bbec945243f47723ce4db0292ab1216378f3c146f748a4dd6f4fec2bc60adR21

the new module here `isaaclab_ppisp` is not part of `install.py` and if we keep it optional therefore none of the backends pull it as a hard dependency we need to find a way to install it. 

(note: if we assume everyone will just use `-i` with no flags this is not a real concern, if they use -i physx / -i ov / -i newton this package will not be pulled)

moennen: Made them lazy.

source/isaaclab_ppisp/isaaclab_ppisp/rtx_camera_overrides.py
moennen: :+1:

source/isaaclab/isaaclab/renderers/renderer_cfg.py
moennen: @ooctipus for viz

moennen: I will revert to the TODO adding my logic here.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -506,7 +516,8 @@ def _create_buffers(self):
         self._data.pos_w = torch.zeros((self._view.count, 3), device=self._device)
         self._data.quat_w_world = torch.zeros((self._view.count, 4), device=self._device)
         self._update_poses(self._ALL_INDICES)
-        self._renderer.set_outputs(self._render_data, self._data.output)
+        self._renderer_output_data = dict(self._data.output)
```
pbarejko: Why do we have to do this? The `self._data.output` is already a dictionary: https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab/isaaclab/sensors/camera/camera_data.py#L56

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -51,11 +54,26 @@ def __init__(self, newton_sensor: newton.sensors.SensorTiledCamera, spec: Camera
...
+        if spec.cfg.isp_cfg is not None:
+            from isaaclab_ppisp import PpispPipeline
+
+            self.ppisp_pipeline = PpispPipeline(spec.cfg.isp_cfg)
+        self.hdr_scratch_torch: torch.Tensor | None = None
+        """Internal HDR scratch torch tensor allocated when PPISP is composed but
+        the user did not request ``"rgb_hdr"`` in ``data_types``."""
 
     def set_outputs(self, output_data: dict[str, torch.Tensor]):
+        self.output_data = output_data
```
pbarejko: Why are we storing this?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -51,11 +54,26 @@ def __init__(self, newton_sensor: newton.sensors.SensorTiledCamera, spec: Camera
         self.outputs = RenderData.CameraOutputs()
         self.width = getattr(spec.cfg, "width", 100)
         self.height = getattr(spec.cfg, "height", 100)
+        self.output_data: dict[str, torch.Tensor] | None = None
```
pbarejko: This seems redundant.

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -127,13 +117,21 @@ def __init__(self, cfg: CameraCfg):
...
+        self.cfg.renderer_cfg.apply_pre_reset_settings(require_hdr_output=require_hdr_output)
+
         # UsdGeom Camera prim for the sensor
         self._sensor_prims: list[UsdGeom.Camera] = list()
         # Allocated in :meth:`_create_buffers` once the renderer's output contract is known.
         self._data: CameraData | None = None
         # Renderer and render data — assigned in _initialize_impl.
         self._renderer: BaseRenderer | None = None
         self._render_data = None
+        self._renderer_output_data: dict[str, torch.Tensor] | None = None
```
moennen: Correct, it is a residue of my previous imp, it was containing the intermediate hdr. Will remove.

source/isaaclab/isaaclab/sensors/camera/camera.py
moennen: Yes, this was previously done for copying the dictionnary (instead of reference). Not needed anymore.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -51,11 +54,26 @@ def __init__(self, newton_sensor: newton.sensors.SensorTiledCamera, spec: Camera
...
+        """Caller-owned torch output tensors; kept for the PPISP dispatch (which needs torch refs)."""
+        # Post-render PPISP pipeline composed when ``spec.cfg.isp_cfg`` is set.
+        # ``isp_cfg`` is already fully normalised by Camera by the time it reaches here.
+        self.ppisp_pipeline: PpispPipeline | None = None
+        if spec.cfg.isp_cfg is not None:
+            from isaaclab_ppisp import PpispPipeline
```
pbarejko, line 74: Is this installed by default? 

https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab/isaaclab/cli/commands/install.py#L436

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -506,7 +536,8 @@ def _create_buffers(self):
         self._data.pos_w = torch.zeros((self._view.count, 3), device=self._device)
         self._data.quat_w_world = torch.zeros((self._view.count, 4), device=self._device)
         self._update_poses(self._ALL_INDICES)
-        self._renderer.set_outputs(self._render_data, self._data.output)
+        self._renderer_output_data = dict(self._data.output)
```
pbarejko: This is redundant.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -120,6 +123,18 @@ def __init__(self, spec: CameraRenderSpec, device):
         self.num_cols = math.ceil(math.sqrt(self.num_envs))
         self.num_rows = math.ceil(self.num_envs / self.num_cols)
         self.warp_buffers: dict[str, wp.array] = {}
+        self.output_data: dict[str, torch.Tensor] | None = None
```
pbarejko: I don't think we need this.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
moennen: Reference to be updated in render by the isp computation.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
moennen: Redundant with self.outputs ? This just to avoid all reshaping gymnastic... no copy here.

source/isaaclab/isaaclab/sensors/camera/camera.py
moennen: See my previous answer.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
moennen: Good catch ! 
added to the three backends install_requires
added it to  VALID_ISAACLAB_SUBMODULES 
updated the install ordering + auto-include

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
pbarejko: Well, the renderer shouldn't own/share the `CameraData.outputs`, and now it does. I think this is a design flaw here. I would solve this differently - bind once, run per frame:

You are essentially rendering hdr, and then you save it in the rgb destination? How about you manage those buffers as follows:

```python
# RenderData.__init__
self._ppisp_hdr_source: torch.Tensor | None = None
self._ppisp_rgba_dest: torch.Tensor | None = None # not sure if it should be rgba, or rgb
```

Then, you initialize those buffers through `set_outputs`:

```python
# RenderData.set_outputs (after the existing hdr_scratch_torch allocation)
if self.ppisp_pipeline is not None:
    # PPISP source is the user-requested rgb_hdr if present, else our internal pre-allocated scratch
    self._ppisp_hdr_source = output_data.get("rgb_hdr", self.hdr_scratch_torch)
    # PPISP destination is the rgba
    self._ppisp_rgba_dest = output_data["rgba"]  # Throw an error if rgb/rgba doesn't exist? 
```

Then render becomes as simple as using the existing buffers?

```python
# NewtonWarpRenderer.render()
if render_data.ppisp_pipeline is not None:
    render_data.ppisp_pipeline.apply(
        render_data._ppisp_hdr_source,
        render_data._ppisp_rgba_dest,
    )
```

I think this way you can clean buffer management and you don't need to store camera data.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
pbarejko: https://github.com/isaac-sim/IsaacLab/pull/5499/changes#r3230699051

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pbarejko: https://github.com/isaac-sim/IsaacLab/pull/5499/changes#r3230699051

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -389,6 +424,7 @@ def set_outputs(self, render_data: OVRTXRenderData, output_data: dict[str, torch
 
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.set_outputs`.
         """
+        render_data.output_data = output_data
```
pbarejko: https://github.com/isaac-sim/IsaacLab/pull/5499/changes#r3230699051

source/isaaclab/isaaclab/renderers/base_renderer.py
pbarejko: Ok sounds good. To unblock you we can merge it with this and later we can re-review this interface @nvsekkin

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -470,6 +497,9 @@ def _check_supported_data_types(self, cfg: CameraCfg):
                 "\n\tHint: If you need to work with these sensor types, we recommend using their fast counterparts."
                 f"\n\t\tFast counterparts: {fast_common_elements}"
             )
+        # Renderer-side ISP requires an LDR colour output to write into.
+        if cfg.isp_cfg is not None and not any(data_type in ("rgb", "rgba") for data_type in cfg.data_types):
```
pbarejko: I think this should be an error on the renderer side. 

> Hey I can not do ISP because there is no destination buffer for it! 

Camera shouldn't really care about this problem. This could happen as a runtime error in set outputs.

pbarejko (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_kernels.py
```diff
@@ -109,6 +109,25 @@ def extract_all_rgba_tiles_kernel(
         output_buffer[env_idx, y, x, 3] = tiled_buffer[src_y, src_x, 3]
 
 
+@wp.kernel
+def extract_all_rgb_float_tiles_kernel(
```
huidongc, line 111: Implementation of this kernel is the same as `extract_all_rgba_tiles_kernel`. I think you can dedup a bit by using the same pattern as `reshape_tiled_image` in `isaaclab.utils.warp.kernels`:

1. Loosen the Python annotations (e.g. to Any) so static checkers do not insist on one dtype.
2. Register concrete signatures with wp.overload, one per dtype (or per pair of dtypes) you actually support.

```python
@wp.kernel(enable_backward=False)
def reshape_tiled_image(
    tiled_image_buffer: Any,
    batched_image: Any,
    ...
):
    ...
        batched_image[camera_id, height_id, width_id, i] = batched_image.dtype(tiled_image_buffer[pixel_start + i])

# uint32 -> int32 conversion is required for non-colored segmentation annotators
wp.overload(
    reshape_tiled_image,
    {"tiled_image_buffer": wp.array(dtype=wp.uint32), "batched_image": wp.array(dtype=wp.uint32, ndim=4)},
)
# uint8 is used for 4 channel annotators
wp.overload(
    reshape_tiled_image,
    {"tiled_image_buffer": wp.array(dtype=wp.uint8), "batched_image": wp.array(dtype=wp.uint8, ndim=4)},
)
# float32 is used for single channel annotators
wp.overload(
    reshape_tiled_image,
    {"tiled_image_buffer": wp.array(dtype=wp.float32), "batched_image": wp.array(dtype=wp.float32, ndim=4)},
)
```

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -63,6 +64,8 @@
 )
 
 if TYPE_CHECKING:
+    from isaaclab_ppisp import PpispPipeline
```
huidongc, line 70: This introduceds the ppisp dependency. You need to add `isaaclab_ppisp` dep in `source/isaaclab_ov/config/extension.toml`

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -203,6 +242,8 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
         height = spec.cfg.height
         num_envs = spec.num_instances
         data_types = spec.cfg.data_types if spec.cfg.data_types else ["rgb"]
+        if spec.cfg.isp_cfg is not None and "rgb_hdr" not in data_types:
+            data_types = [*data_types, "rgb_hdr"]
```
huidongc, line 251: nit pick: data_types.append("rgb_hdr")

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -412,6 +453,36 @@ def set_outputs(self, render_data: OVRTXRenderData, output_data: dict[str, torch
                 device=torch_array.device,
                 copy=False,
             )
+        # When PPISP is composed but the user did not request the raw HDR AOV,
+        # allocate an internal HDR scratch tensor for the HdrColor extractor to
+        # write into. We also register it as a warp buffer under "rgb_hdr" so
+        # the existing ``_extract_hdr_color_tiles`` path finds a destination.
+        if render_data.ppisp_pipeline is not None and "rgb_hdr" not in render_data.warp_buffers:
+            hdr_spec = self.supported_output_types()[RenderBufferKind.RGB_HDR]
+            scratch_torch = torch.zeros(
```
huidongc: this will allocate the tensor for every render? What is the cost like with and without caching it?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -19,20 +19,39 @@
...
 
-    if use_depth and not (use_rgb or use_albedo or use_semantic):
-        return "/Render/Vars/depth", "depth", "DistanceToImagePlaneSD"
-    if use_albedo and not (use_rgb or use_semantic):
-        return "/Render/Vars/albedo", "albedo", "DiffuseAlbedoSD"
-    if use_semantic and not (use_rgb or use_albedo):
-        return "/Render/Vars/semantic", "semantic", "SemanticSegmentation"
-    return "/Render/Vars/LdrColor", "LdrColor", "LdrColor"
+    if not render_vars:
+        render_vars.append(("/Render/Vars/LdrColor", "LdrColor", "LdrColor"))
```
huidongc: nit pick: if data types at least contain "rgb", I would assert render_vars being never empty.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -19,20 +19,39 @@
...
+    if use_ldr:
+        render_vars.append(("/Render/Vars/LdrColor", "LdrColor", "LdrColor"))
+    if use_hdr:
+        render_vars.append(("/Render/Vars/HdrColor", "HdrColor", "HdrColor"))
+    if use_depth:
+        render_vars.append(("/Render/Vars/depth", "depth", "DistanceToImagePlaneSD"))
+    if use_albedo:
+        render_vars.append(("/Render/Vars/albedo", "albedo", "DiffuseAlbedoSD"))
+    if use_semantic:
+        render_vars.append(("/Render/Vars/semantic", "semantic", "SemanticSegmentation"))
```
huidongc: Why do you make this change? Is it because we can actually request multiple render vars for the same training session? Have you tested how they work together?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
moennen: I do not really want to mutate the spec.cfg.data_types.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
moennen: Did I miss something ? set_outputs is called from the camera _create_buffers only once, not per render :thinking:

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
moennen: Since we need to pull both Ldr and Hdr, I tried to generalize the API. But you are right since I'm testing only the case Ldr + Hdr I will revert this generalization.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_kernels.py
moennen: Yes make complete sense. However since this PR is due for the release. could we take this as a follow-up ? I'm afraid this would be too much :pray:

huidongc (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: My impression is that it is called every frame. But I might be wrong because I've been with Lab not long.

PR #5496, opened 2026-05-05
Stabilize test_non_headless_launch and test_surface_gripper in CI

hujc7:
## Summary

Two CI flake fixes:

1. **Bump `STARTUP_DEADLINE` in `tools/conftest.py` from 45 s to 120 s.** Kit startup with the non-headless experience can exceed 60 s on cold CI workers; `test_non_headless_launch.py` was observed at 51–79 s wall on green runs, sitting right against the previous deadline. The lower threshold flagged legitimate slow launches as `STARTUP_HANG`.

2. **Avoid an Isaac Sim `SurfaceGripperView` deadlock during init in CI:**
   - `SurfaceGripper._initialize_impl` now performs the CPU-backend check **before** loading the upstream `isaacsim.robot.surface_gripper` extension. CUDA configs fail fast without triggering the extension load that can hang.
   - `test_surface_gripper.py::test_initialization` is skipped in CI because the upstream `GripperView` constructor can deadlock on the CI Docker image. `test_raise_error_if_not_cpu` continues to run and exercises the device guard above. This keeps the CI signal stable until the upstream fix lands.

Closes #5402.
Closes isaac-sim/IsaacLab-Internal#869.

### Pre-fix `STARTUP_HANG` rate

Measured across 60 recent Docker + Tests runs (excluding fix branches): **1 STARTUP_HANG out of 29 runs that actually executed the test, ≈ 3.4% per attempt**.

Wall-time distribution on the 28 passing runs:

| min | p50 | p90 | max |
|---|---|---|---|
| 48.2 s | 55.3 s | 78.5 s | 106.3 s |

100% of passing runs had wall > 45 s; 21% had wall > 60 s. The new 120 s deadline sits well outside the observed distribution.

### Post-fix stress test (10 attempts on sha 5e0281d6)

5 full-workflow reruns + 5 single-job reruns (`gh run rerun --job`) on PR #5496:

| attempt | rerun mode | `test_non_headless_launch` wall | `test_surface_gripper` | overall |
|---|---|---|---|---|
| 1 | full | 78.82 s | 1/2 (skip applied) | green |
| 2 | full | 51.71 s | 1/2 (skip applied) | green |
| 3 | full | 53.02 s | 1/2 (skip applied) | green |
| 4 | full | 52.49 s | 1/2 (skip applied) | green |
| 5 | full | 78.27 s | 1/2 (skip applied) | green |
| 6 | core only | 58.74 s | (carry) | green |
[...]

PR #5509, opened 2026-05-06
[STRESS / DO NOT MERGE] Stress harness for #5496

hujc7:
## Purpose

Stress-test harness for the fix in https://github.com/isaac-sim/IsaacLab/pull/5496. Disables every test job in \`build.yaml\` except the two that exercise the fix (\`isaaclab_physx\` and \`isaaclab (core) [3/3]\`) and pins each surviving job's \`include-files\` to a single test:

- \`test_non_headless_launch.py\`
- \`test_surface_gripper.py\`

This drops cycle time from ~75 min (gated by \`isaaclab_tasks\` / \`test-curobo\` / \`test-skillgen\`) to ~10–15 min, so we can collect many sequential reruns and tighten the post-fix STARTUP_HANG rate measurement.

## Status

Throwaway. Not for merge. Will be deleted after stress test concludes.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -219,7 +219,7 @@ jobs:
     name: Build cuRobo Docker Image
     runs-on: [self-hosted, gpu]
     needs: [changes, config]
-    if: needs.changes.outputs.run_docker_tests == 'true'
+    if: false  # [STRESS] disabled for stress-test
     steps:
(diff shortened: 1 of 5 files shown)


hujc7:
Closing — fork-PR workflow-file changes are silently ignored by upstream Actions, so this approach can't run. Pivoting to job-level reruns on PR #5496.

PR #5510, opened 2026-05-06
Split RayCaster into backend-specific implementations

ooctipus:
# Description

Refactors `RayCaster` and `RayCasterCamera` into a backend-agnostic base + per-backend implementations under `isaaclab_physx.sensors.ray_caster` and `isaaclab_newton.sensors.ray_caster`, mirroring the iconic split pattern already used by `Pva`, `FrameTransformer`, and `ContactSensor`.

The PhysX backend tracks the parent rigid body directly via `RigidObjectView` instead of going through `FabricFrameView`, which fixes the staleness regression from #5179: sensors parented under an articulation / rigid body were returning their spawn-time pose forever during headless training, silently freezing height-scan observations in rough-terrain locomotion (and any similar IMU / camera path that read through `FrameView`).

The Newton backend uses the site-based pattern from `Pva` / `FrameTransformer`: walk USD to the rigid-body ancestor, register a body-attached site via `NewtonManager.cl_register_site`, and read per-step transforms off a `SensorFrameTransform` against a shared world-origin reference. Static parents bypass the site machinery (a single `body=-1` global site can't represent per-env world origins) and serve a cached per-env `wp.transformf` array.

`MultiMeshRayCaster` / `MultiMeshRayCasterCamera` re-parent onto the new base but keep their `FrameView`-backed body tracker, so the staleness behavior persists there. Tracked as `xfail` in `test_ray_caster_sensor.py` — extending the backend split to MultiMesh is a follow-up.

The cfg surface, `class_type` strings, and runtime semantics are unchanged for callers; existing user code does not need to migrate.

Fixes #5476 (the `FabricFrameView` contract regression-test PR — the bug it documents is fixed for the single-mesh path here).

## Type of change

## Screenshots

N/A — backend refactor, no UI changes.

## Checklist

- [x] I have added a changelog fragment under \`source/<pkg>/changelog.d/\` for every touched package (do **not** edit \`CHANGELOG.rst\` or bump \`extension.toml\` — CI handles that)


ooctipus:
Thanks for the review! Triage on each item:

**🔴 Newton `_invalidate_body_tracker` — push back.** The cached flag (`_site_args_cached=True`) only short-circuits the *USD walk* — the cached `(body_pattern, site_xform)` pair. The two `cl_register_site` calls always run on the rigid-body branch (lines 104–105 in `ray_caster.py`), so both the world-origin label and the source label get re-issued every invalidate cycle. Mirrors the same pattern in `isaaclab_newton.sensors.frame_transformer.FrameTransformer._invalidate_initialize_callback`.

**🟡 `_zero_data_buffers` `.zero_()` vs `wp.launch` / mid-step race — push back.** The `.zero_()` pattern is the established one for this lifecycle event: see `isaaclab_newton.sensors.pva.Pva._invalidate_initialize_callback` (lines 240–252) which uses the identical `for buf in [...]: buf.zero_()` loop. Race-with-kernel-writes isn't a concern: `PhysicsEvent.STOP` is dispatched synchronously from `SimulationContext.stop()` after the sim loop has exited — there's no concurrent kernel running at that point.

**🟡 Unnecessary `.clone()` in `BaseRayCasterCamera.reset` — punt.** Correct observation but pre-existing — those clones predate this PR (verbatim from the original `RayCasterCamera`). Removing them is a separate cleanup; perf-irrelevant since `reset` / `set_world_poses` are cold paths.

**🟡 PhysX offset composition / camera ordering — already validated** in the same comment (\"does, per line 135-137\"). No action.

**🔵 Multi-mesh `_ray_distance_w` documentation — addressed in ebcc2b9.** Comment now contrasts the full-size atomic_min target with the base's `(1, 1)` placeholder dummy.

**🔵 Deprecated method TODO version comments — push back.** `_compute_view_world_poses` / `_compute_camera_world_poses` are pre-existing methods I just renamed-through (the rename was `_get_view_world_poses` → `_get_sensor_world_poses` upstream of these). Their existing `.. deprecated:: v2.3.1` Sphinx markers carry the version metadata. Adding `# TODO: Remove in v3.0` would be a guess at a target the project hasn't publicly committed to.

**🔵 Newton ground-truth in regression test — push back.** A real Newton ground-truth would need to read `state.body_q[body_idx]` after resolving the parent body's index from `NewtonManager._model.body_label` — that's nontrivial test plumbing. Comparing against `sensor._newton_transforms` would be tautological (it *is* the sensor pose post-composition, not an independent body pose). The change-detection assertion (`drop > 1 cm`) already pins the bug class on Newton; the tight ground-truth on PhysX is the bonus that comes for free because `_body_view` is the right shape there.

So: 1 fix landed, 6 push-backs above. Happy to reconsider any of them — let me know if I'm misreading something.

source/isaaclab_physx/isaaclab_physx/sensors/ray_caster/ray_caster.py
```diff
@@ -0,0 +1,162 @@
...
+        # (no fixed offset to compose). ``wp.to_torch`` returns zero-copy alias
+        # views; slice-writes update warp memory the kernel reads.
+        if self._fixed_pos_b is not None:
+            offset_pos = wp.to_torch(self._offset_pos_wp)
+            offset_quat = wp.to_torch(self._offset_quat_wp)
+            cur_pos, cur_quat = offset_pos.clone(), offset_quat.clone()
+            offset_pos[:] = self._fixed_pos_b + math_utils.quat_apply(self._fixed_quat_b, cur_pos)
+            offset_quat[:] = math_utils.quat_mul(self._fixed_quat_b, cur_quat)
+
+    def _get_sensor_transforms_wp(self) -> wp.array:
```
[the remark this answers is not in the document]
ooctipus: Good catch — this is a real correctness bug (and my regression tests did not exercise it because they use `cfg.offset = identity` *and* a sensor Xform with identity USD-local translation, so `T_body_camera == cfg.offset`).

Fixed in 32d91099ab7 with **option (a)**: drop the bake entirely and compose `body_pose * T_body_to_xform` per step via a tiny `wp.kernel` into a pre-allocated `_sensor_transforms_wp` buffer. `_offset_pos_wp` / `_offset_quat_wp` (and their torch aliases on the camera path) stay holding `cfg.offset`, so `reset()` and `set_world_poses()` see the right value.

Added two regression tests in `test_ray_caster_sensor.py`:

- `test_camera_offset_buffer_survives_body_tracker_init` asserts `_offset_pos` still equals `cfg.offset` after `_initialize_body_tracker` under a rigid parent. (Direct invariant — would have caught the original bake.)
- `test_camera_set_world_poses_under_rigid_parent` exercises the user-visible failure: kinematic rigid parent at `z=5`, sensor Xform at `(0, 0.1, 0)` USD-local, `set_world_poses(positions=[(1, 2, 6)])`, then assert `data.pos_w[0] ≈ (1, 2, 6)` within 1 cm. Pre-fix path would land at `(1, 1.9, 6)` (off by the `y=0.1` body-to-Xform offset).

Per-step cost is one extra `wp.launch` of a transform-multiply kernel (`(num_envs,)` threads). Negligible vs the existing ray-cast launch.

ooctipus:
@greptile-apps review again

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+
+        Args:
+            cfg: The configuration parameters.
+        """
+        super().__init__(cfg)
+
+        self._num_meshes_per_env: dict[str, int] = {}
+
+        self._raycast_targets_cfg: list[MultiMeshRayCasterCfg.RaycastTargetCfg] = []
+        for target in self.cfg.mesh_prim_paths:
```
StafaH, line 113: Why not merge the two target_cfg loops?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+                f"No meshes found for ray-casting! Please check the mesh prim paths: {self.cfg.mesh_prim_paths}"
+            )
+
+        dummy_record = (dummy_mesh_id, (1.0e9, 1.0e9, 1.0e9), (0.0, 0.0, 0.0, 1.0))
+        multi_mesh_ids_flattened: list[list[int]] = []
+        mesh_positions: list[list[tuple[float, float, float]]] = []
+        mesh_orientations: list[list[tuple[float, float, float, float]]] = []
+
+        # Pack all targets into the rectangular table used by the Warp kernel.
+        for env_id in range(self._num_envs):
```
StafaH: This can be vectorized with numpy. Also seems the dummy record is only used to extend the list. Can we not allocate the array in advance?

Even better it seems you could populate it with a warp kernel, since the end result is a warp kernel below?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+
+        total_n_meshes_per_env = len(multi_mesh_ids_flattened[0])
+        self._mesh_ids_wp = wp.array2d(multi_mesh_ids_flattened, dtype=wp.uint64, device=self.device)
+        self._mesh_positions_w = wp.zeros((self._num_envs, total_n_meshes_per_env), dtype=wp.vec3, device=self.device)
+        self._mesh_orientations_w = wp.zeros(
+            (self._num_envs, total_n_meshes_per_env), dtype=wp.quat, device=self.device
+        )
+        self._mesh_positions_w_torch = wp.to_torch(self._mesh_positions_w)
+        self._mesh_orientations_w_torch = wp.to_torch(self._mesh_orientations_w)
+        self._mesh_positions_w_torch[:] = torch.tensor(mesh_positions, dtype=torch.float32, device=self.device)
```
StafaH: Can we use proxy array here?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+        self._mesh_orientations_w_torch[:] = torch.tensor(mesh_orientations, dtype=torch.float32, device=self.device)
+
+    def _build_mesh_records(self, target_cfg: _RaycastTargetCfg, plan: ClonePlan | None, dummy_mesh_id: int | None):
```
StafaH: Why use an alias for _RaycastTargetCfg?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+        """Build mesh records for the target configuration."""
+        records_per_env: list[list[_MeshRecord]] = [[] for _ in range(self._num_envs)]
+        target_in_plan = False
+
+        # Prefer ClonePlan data for env-scoped targets; destination USD prims may not exist.
+        if plan is not None:
+            target_path = re.sub(r"env_\.\*", "env_0", target_cfg.prim_expr)
```
StafaH, line 212: Can we not use target_cfg.prim_expr.replace here?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
StafaH: This could be split into two smaller functions which would make it much easier to read/maintain and use, `_build_mesh_records_from_clone_plan` and `_build_mesh_records_from_usd`

```
records_per_env, dummy_mesh_id = self._build_mesh_records_from_clone_plan(plan, target_cfg)
if records_per_env is not None:
    return records_per_env, dummy_mesh_id

return self._build_mesh_records_from_usd(target_cfg)
```

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -31,3 +33,9 @@ class ClonePlan:
     """Boolean tensor of shape ``[len(sources), num_envs]``;
     ``clone_mask[i, j]`` is ``True`` iff env ``j`` was populated from
     :attr:`sources` ``[i]``."""
+
+    env_root_poses: torch.Tensor
```
StafaH: Unrelated to this PR, but it seems env_root_poses uses pos + quat, but only pos is ever used. Is there a reason for the env root to not be identity?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+                " enable `merge_prim_meshes` in the configuration or specify each mesh separately."
+            )
+
+        wp_mesh = convert_to_warp_mesh(trimesh_mesh.vertices, trimesh_mesh.faces, device=self._device)
+        BaseMultiMeshRayCaster.meshes[prim_key] = wp_mesh
+        logger.info(
+            f"Read '{len(mesh_prims)}' mesh prims under path '{target_prim.GetPath()}' with"
+            f" {len(trimesh_mesh.vertices)} vertices and {len(trimesh_mesh.faces)} faces."
+        )
+        return wp_mesh.id
```
StafaH, line 370: Is the warp mesh stored somewhere? As far as I'm aware, in this scenario, when wp_mesh is garbage collected, you would lose the underlying mesh data and your results would be invalid. Maybe somehow this is avoided, but this has definately been the case before

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+    def _update_buffers_impl(self, env_mask: wp.array):
+        """Fills the buffers of the sensor data."""
+        self._update_ray_infos(env_mask)
+        self._update_mesh_transforms()
+
+        # Fill output and distance buffers with inf for masked environments
+        wp.launch(
+            fill_vec3_inf_kernel,
+            dim=(self._num_envs, self.num_rays),
+            inputs=[env_mask, float("inf"), self._data._ray_hits_w],
```
StafaH: wp.inf? Here and elsewhere

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+            self._mesh_orientations_w_torch[:, mesh_idx : mesh_idx + meshes_per_env] = ori_w
+            mesh_idx += self._num_meshes_per_env[target_cfg.prim_expr]
+
+    def _update_buffers_impl(self, env_mask: wp.array):
+        """Fills the buffers of the sensor data."""
+        self._update_ray_infos(env_mask)
+        self._update_mesh_transforms()
+
+        # Fill output and distance buffers with inf for masked environments
+        wp.launch(
```
StafaH, line 432: There two kernels could be combined to one kernel.

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+        )
+        wp.launch(
+            fill_float2d_masked_kernel,
+            dim=(self._num_envs, self.num_rays),
+            inputs=[env_mask, float("inf"), self._ray_distance_w],
+            device=self._device,
+        )
+
+        n_meshes = self._mesh_ids_wp.shape[1]
+        return_normal = False
```
StafaH, line 441: You can inline these directly into the kernel launch

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,291 @@
...
+        if self.cfg.update_mesh_ids:
+            self._ray_mesh_id_w = wp.zeros((self._view.count, self.num_rays), dtype=wp.int16, device=self._device)
+            self._data.ray_mesh_ids = wp.to_torch(self._ray_mesh_id_w).unsqueeze(-1)
+        else:
+            self._ray_mesh_id_w = wp.empty((1, 1), dtype=wp.int16, device=self._device)
+
+        # Dummy face-id buffer (not used by camera but required by kernel signature)
+        self._ray_face_id_w = wp.empty((1, 1), dtype=wp.int32, device=self._device)
+
+    def _update_ray_infos(self, env_mask: wp.array):
```
StafaH, line 134: Why not make the whole function use warp instead of jumping to torch and then back to warp

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,291 @@
...
+
+        # Increment frame count for updated environments
+        self._frame[env_ids] += 1
+
+        self._update_mesh_transforms()
+
+        return_normal = "normals" in self.cfg.data_types
+
+        # Fill ray hit and distance buffers with inf for masked environments
+        wp.launch(
```
StafaH, line 176: Same comment as before, can be combined

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster.py
```diff
@@ -0,0 +1,378 @@
...
+        is a work in progress.
+    """
+
+    cfg: RayCasterCfg
+    """The configuration parameters."""
+
+    meshes: ClassVar[dict[tuple[str, str], wp.Mesh]] = {}
+    """A dictionary to store warp meshes for raycasting, shared across all instances.
+
+    The keys are ``(prim_path, device)`` tuples and values are the corresponding warp Mesh objects.
```
StafaH: Why are warp meshes being created on both devices? Should it not just be on the device currently set globally?

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster.py
```diff
@@ -0,0 +1,378 @@
...
+    def _initialize_warp_meshes(self):
+        # check number of mesh prims provided
+        if len(self.cfg.mesh_prim_paths) != 1:
+            raise NotImplementedError(
+                f"RayCaster currently only supports one mesh prim. Received: {len(self.cfg.mesh_prim_paths)}"
+            )
+
+        # read prims to ray-cast
+        for mesh_prim_path in self.cfg.mesh_prim_paths:
+            mesh_key = (mesh_prim_path, self._device)
```
StafaH, line 171: Same as previous comment, seems _device will not change? Then we don't need to key with tuple

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster_camera.py
```diff
@@ -0,0 +1,609 @@
...
+        up_axis = UsdGeom.GetStageUpAxis(self.stage)
+        # camera position and rotation in opengl convention
+        orientations = math_utils.quat_from_matrix(
+            math_utils.create_rotation_matrix_from_view(eyes, targets, up_axis=up_axis, device=self._device)
+        )
+        self.set_world_poses(eyes, orientations, env_ids, convention="opengl")
+
+    def _initialize_rays_impl(self):
+        # Create all indices buffer
+        self._ALL_INDICES = torch.arange(self._view.count, device=self._device, dtype=torch.long)
```
StafaH: nit: this does not need to be caps

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,442 @@
...
+if TYPE_CHECKING:
+    from isaaclab.cloner import ClonePlan
+
+    from .multi_mesh_ray_caster_cfg import MultiMeshRayCasterCfg
+
+    _RaycastTargetCfg = MultiMeshRayCasterCfg.RaycastTargetCfg
+
+logger = logging.getLogger(__name__)
+
+_MeshRecord = tuple[int, tuple[float, float, float], tuple[float, float, float, float]]
```
StafaH: In alot of places this tuple exists for vec3 and vec4/quat, maybe we can alias the type somewhere so it's a bit easier to read. If we convert more of the code to warp we can use wp.vec3 and wp.quat which would be alot cleaner

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster.py
```diff
@@ -0,0 +1,378 @@
...
+        super().__init__(cfg)
+        # Resolve physics-body paths and spawn the sensor Xform child if needed.
+        self._requested_prim_path = self.cfg.prim_path
+        self._resolve_and_spawn("raycaster")
+        self._data = RayCasterData()
+
+    def __str__(self) -> str:
+        """Returns: A string containing information about the instance."""
+        return (
+            f"Ray-caster @ '{self.cfg.prim_path}': \n"
```
StafaH, line 80: BaseRayCaster has the same __str__ overload, can we just call that one?

StafaH (commented):
Nice PR! Left some comments.

Overall there's alot of switching between torch and warp ops, which could be cleaned up to use warp only. Also we could use ProxyArrays in all the base classes to clean things up even more. Nothing critical for this PR though, can merge it in and clean up later

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: I can't think of a reason now , but I could think of usecases for camera to be place at idfferent angle and look at different background , then it would be nice to read from quaternion

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster.py
ooctipus: I don't fully understand how original code was written this way, seems like its trying to make sure gpu to cross each other to raycast onto other devices mesh?

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
ooctipus: Good thoughts , I removed these typing

source/isaaclab/isaaclab/sensors/ray_caster/base_ray_caster_camera.py
ooctipus: I think _ALL_INDICES is a common thing many files, I agree, but I will rename every one of them in one go for consistency

ooctipus:
@greptile-apps review

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,459 @@
...
+        super().__init__(cfg)
+
+        self._num_meshes_per_env: dict[str, int] = {}
+
+        self._raycast_targets_cfg: list[MultiMeshRayCasterCfg.RaycastTargetCfg] = []
+        for target in self.cfg.mesh_prim_paths:
+            if isinstance(target, str):
+                target_cfg = cfg.RaycastTargetCfg(prim_expr=target, track_mesh_transforms=False)
+            else:
+                target_cfg = target
```
StafaH, line 117: Should we allow users to pass in the config? Why not allow only paths? The variable is called mesh_prim_paths

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,459 @@
...
+        for mesh_prim in mesh_prims:
+            if mesh_prim is None or not mesh_prim.IsValid():
+                raise RuntimeError(f"Invalid mesh prim path: {target_prim}")
+            if mesh_prim.GetTypeName() == "Mesh":
+                mesh = create_trimesh_from_geom_mesh(mesh_prim)
+            else:
+                mesh = create_trimesh_from_geom_shape(mesh_prim)
+            mesh.apply_scale(sim_utils.resolve_prim_scale(mesh_prim))
+            relative_pos, relative_quat = sim_utils.resolve_prim_pose(mesh_prim, reference_prim)
+            relative_pos = torch.tensor(relative_pos, dtype=torch.float32)
```
StafaH: We could probably just use numpy, since we end up with warp arrays in the end, and all of this is being done on CPU

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,459 @@
...
+    def _create_tracked_target_view(self, target_prim_paths: str | list[str]):
+        raise NotImplementedError("Tracked multi-mesh targets must be implemented by the active physics backend.")
+
+    def _initialize_rays_impl(self):
+        super()._initialize_rays_impl()
+        # Persistent buffer for tracking closest-hit distance across meshes (for atomic_min)
+        self._ray_distance_w = wp.zeros((self._view_count, self.num_rays), dtype=wp.float32, device=self._device)
+        if self.cfg.update_mesh_ids:
+            self._ray_mesh_id_w = wp.zeros((self._view_count, self.num_rays), dtype=wp.int16, device=self._device)
+            self._data.ray_mesh_ids = ProxyArray(self._ray_mesh_id_w)
```
StafaH: Here and elsewhere, we can just directly do ProxyArray(wp.empty(xxxx)), no need to store a second reference to the warp array, since the ProxyArray reference can be used in the same way.

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py
```diff
@@ -0,0 +1,459 @@
...
+        else:
+            # Dummy 1×1 buffer so the kernel launch always has a valid array to bind
+            self._ray_mesh_id_w = wp.empty((1, 1), dtype=wp.int16, device=self._device)
+        # Persistent dummy buffers for unused kernel outputs; allocated once to avoid per-step allocations.
+        self._dummy_normal_w = wp.empty((1, 1), dtype=wp.vec3, device=self._device)
+        self._dummy_face_id_w = wp.empty((1, 1), dtype=wp.int32, device=self._device)
```
StafaH: For warp arrays let's suffix with _wp, since _w means world frame in many other parts of the repo

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,295 @@
...
+        # chain. The intermediate classes (RayCasterCamera, MultiMeshRayCaster) use different internal
+        # buffer names and orderings that are incompatible with the camera's full init path:
+        #   - RayCasterCamera creates single-mesh ray buffers (_ray_distance, _ray_normal_w, etc.)
+        #   - MultiMeshRayCaster creates _ray_distance_w / _ray_mesh_id_w for multi-mesh use
+        # The camera replaces all of these with its own camera-named equivalents below.
+        # If either parent class gains new shared buffers, they must be added here explicitly.
+
+        # Camera-specific bookkeeping buffers
+        self._frame_wp = wp.zeros(self._view.count, dtype=wp.int64, device=self._device)
+        self._frame = wp.to_torch(self._frame_wp)
```
StafaH, line 93: In another PR I changed this to be ProxyArray so depending on which ones gets merged first this might need to be updated

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,295 @@
...
+        self._create_offset_buffers(self._view.count)
+
+        # Warp buffer for distance_to_image_plane output (if requested)
+        if "distance_to_image_plane" in self.cfg.data_types:
+            self._distance_to_image_plane_wp = wp.zeros(
+                (self._view.count, self.num_rays), dtype=wp.float32, device=self._device
+            )
+
+        # World-frame ray buffers.
+        self._ray_starts_w = wp.zeros((self._view.count, self.num_rays), dtype=wp.vec3f, device=self._device)
```
StafaH: Here and elsewhere, use wp.empty instead of wp.zeros, it's much faster (zero-ing out an array takes some time). Only need wp.zeros when you specifically need a zero valued array

source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster_camera.py
```diff
@@ -0,0 +1,295 @@
...
+        return_normal = "normals" in self.cfg.data_types
+
+        # Fill ray hit and distance buffers with inf for masked environments
+        wp.launch(
+            ray_caster_kernels.fill_ray_hits_distance_inf_kernel,
+            dim=(self._num_envs, self.num_rays),
+            inputs=[env_mask, self.ray_hits_w.warp, self._ray_distance_cam_w],
+            device=self._device,
+        )
+        if return_normal:
```
StafaH: Combine with above and pass return_normal as argument to the kernel?

source/isaaclab_newton/isaaclab_newton/sensors/ray_caster/ray_caster.py
```diff
@@ -0,0 +1,309 @@
...
+        site_indices = self._resolve_site_indices(self._sensor_site_labels, self.cfg.prim_path, self._num_envs)
+        # The base classes still use ``self._view.count`` in a few generic
+        # places. Point it at the sensor instead of constructing an adapter.
+        self._view = self
+        self._view_count = len(site_indices)
+        self._sensor_site_indices = wp.array(site_indices, dtype=wp.int32, device=self._device)
+        self._newton_pose_w = wp.zeros(self._view_count, dtype=wp.transformf, device=self._device)
+        self._newton_pos_w = wp.zeros(self._view_count, dtype=wp.vec3f, device=self._device)
+        self._newton_quat_w = wp.zeros(self._view_count, dtype=wp.quatf, device=self._device)
+        self._newton_pos_w_proxy = ProxyArray(self._newton_pos_w)
```
StafaH: Combine with above initialization? ProxyArray(wp.empty(...))

StafaH (approved):
LGTM

PR #5434, opened 2026-04-29
Add fragment-based changelog system to eliminate per-PR merge conflicts

hujc7:
## Problem

Every PR that touches `source/<pkg>/` updates `source/<pkg>/docs/CHANGELOG.rst` and `source/<pkg>/config/extension.toml`. Both files are append-only at the top, so concurrent PRs collide on the same lines. Contributors burn time resolving merge conflicts that say nothing about the actual change. Maintainers occasionally land an entry under the wrong version heading because the resolution is mechanical and easy to mis-do.

## Solution

Adopt the towncrier-style fragment-per-PR pattern (used by pip, urllib3, Twisted, Trio, attrs, Sphinx, …):

- Each PR adds **one fragment file** per touched package under `source/<pkg>/changelog.d/<slug>.<tier>.rst`. Different PRs touch different files — no conflict.
- A **nightly CI workflow** rolls accumulated fragments into per-package `CHANGELOG.rst` entries, bumps each `extension.toml`, deletes consumed fragments, and pushes the result back to `develop`.
- A **PR gate** rejects any PR that touches a managed package's source without adding a fragment, and rejects modifications to existing fragments (immutability).

CHANGELOG.rst and extension.toml become append-only-by-CI files; humans never edit them.

## Contributor workflow (one fragment per touched package)

```bash
# Branch: jdoe/fix-spawn-pose
$EDITOR source/isaaclab/changelog.d/jdoe-fix-spawn-pose.rst
```

```rst
Fixed
^^^^^

* Fixed :meth:`~isaaclab.assets.Articulation.write_root_pose` ignoring the env-id mask.
```

Filename convention:

| Filename | Effect |
|---|---|
| `<slug>.rst` | patch bump (`X.Y.Z → X.Y.Z+1`) |
| `<slug>.minor.rst` | minor bump (`X.Y.Z → X.Y+1.0`) |
| `<slug>.major.rst` | major bump (`X+1.0.0`) — breaking change |
| `<slug>.skip` | no entry, no bump (CI / docs / test-only PRs) |

The slug is any short, unique name. **Branch name with `/` replaced by `-` is the recommended default** — already in scope at commit time, naturally unique because of the `<user>/<feature>` branch convention. Within a batch, the highest tier wins for the package (`major > minor > patch`).

## What the nightly does

[...]

PR #5516, opened 2026-05-06
perf: Implement Fabric-aware get/set_local_poses on FabricFrameView

pv-nvidia:
# Description

`FabricFrameView.set_world_poses` updated Fabric's `omni:fabric:worldMatrix`, but `get_local_poses` still delegated to USD — so after a world write the next local read returned stale values until USD was re-synced. Symmetrically, `set_local_poses` went through USD and Fabric never saw the write until the next sync. The shared contract test `test_set_world_updates_local` was marked `xfail(strict=True)` to acknowledge this.

This PR replaces the USD-fallback local-pose path with a fully Fabric-native implementation. World and local matrices are read and written directly on `omni:fabric:worldMatrix` / `omni:fabric:localMatrix` via `wp.indexedfabricarray`; consistency in both directions is maintained by Warp kernels.

The design rests on three pieces: three persistent `PrimSelection` instances differing only in per-attribute access mode (`trans_ro`, `world_rw`, `local_rw`); path-based view→fabric index mapping (no custom prim attributes written); and `wp.indexedfabricarray` for every (selection × attribute) pair, which bakes the view→fabric mapping into the array itself so kernels read element `i` of the view directly instead of taking a separate index-mapping argument. World↔local consistency is maintained entirely on the GPU through two Warp kernels — no Python-side parent walks, no per-call USD reads. The result is faster than the USD baseline by two orders of magnitude on every measured operation (see *Benchmark results* below).

# Implementation summary

* **Three persistent `PrimSelection` instances** — `trans_ro` (read both matrices), `world_rw` (write world, read local), `local_rw` (read world, write local). Each selection caches its own view→fabric index mapping built from `selection.GetPaths()`; selections do not guarantee a shared path ordering, so the mappings are kept independent.
* **Path-based view→fabric index mapping** — no `isaaclab:view_index:HASH` attributes written anywhere on the prims. Fabric groups prims into buckets by their attribute schema; writing a per-view index attribute would split each touched bucket into a "with-attr" and "without-attr" half and proliferate further with each additional view on the stage. Keeping the mapping out of Fabric leaves the bucket layout untouched, which keeps `SelectPrims` enumeration and `wp.fabricarray` reads more contiguous.
* **`wp.indexedfabricarray` everywhere** — bakes the mapping into the array itself, removing a per-launch indirection.
* **Per-view dirty flag for world↔local consistency:**
  * After `set_local_poses` the view is marked dirty; the next world read fires a Warp kernel that recomputes `child_world = parent_world * child_local` for this view's children. Tracking is per-view (not per-stage) so multiple `FabricFrameView` instances on the same stage cannot clear each other's dirty flag.
  * After `set_world_poses` / `set_scales`, a mirror Warp kernel recomputes `child_local = inv(parent_world) * child_world` so subsequent `get_local_poses` returns consistent values.
* **Stage-level `IFabricHierarchy` cache** keyed by `(stage_id, fabric_id)` so a recycled `stage_id` paired with a new Fabric attachment never returns a stale handle. `FabricFrameView.clear_static_caches()` is the explicit teardown hook.
* **Four new public Warp kernels** in `isaaclab.utils.warp.fabric`: `decompose_indexed_fabric_transforms`, `compose_indexed_fabric_transforms`, `update_indexed_local_matrix_from_world`, `update_indexed_world_matrix_from_local`.

## Transpose-storage convention

Fabric stores 4×4 transforms in column-transposed form. The two sync kernels apply the identity `(A·B)ᵀ = Bᵀ·Aᵀ` (and `inv(A)ᵀ = inv(Aᵀ)`) to compute the equivalent multiply on the stored matrices without explicit transposes:

* math `local = inv(parent) · world`  →  stored `local = world · inv(parent)`
* math `world = parent · local`        →  stored `world = local · parent`

Two new regression tests pin this convention; see *Tests* below.

# Benchmark results

1024 prims, 50 iterations, NVIDIA RTX A6000.

| Operation | USD (ms) | Fabric (ms) | Speedup |
|---|---|---|---|
| Get World Poses | 13.4 | 0.057 | **238×** |
| Set World Poses | 31.3 | 0.139 | **226×** |
| Interleaved Set→Get | 44.6 | 0.172 | **259×** |
| Get Local Poses | 6.1 | 0.039 | **158×** |
| Set Local Poses | 8.5 | 0.054 | **159×** |

Reproduce with:

```bash
./isaaclab.sh -p scripts/benchmarks/benchmark_view_comparison.py \
  --num_envs 1024 --num_iterations 50 --backends usd fabric \
  --device cuda:0 --headless
```

# Tests

Local run of `source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py` on `cpu` and `cuda:0`: **45 passed, 0 failed**.

* Removed the `xfail(strict=True)` on the shared `test_set_world_updates_local` — it now passes.
* `test_set_local_via_fabric_path` — `set_local_poses` writes `omni:fabric:localMatrix`; the deferred `world = parent * local` recompute lands the right world pose.
* `test_get_scales_fabric_path` — exercises the Fabric-native `get_scales`.
* `test_prepare_for_reuse_detects_topology_change` — confirms `PrimSelection.PrepareForReuse` is wired up and returns the expected topology-change signal.
* `test_fabric_rebuild_after_topology_change` — simulates a topology change and confirms the indexed arrays rebuild and continue producing correct results.
* **Transpose-convention verification** (`test_set_local_then_get_world_with_rotated_parent`, `test_set_world_then_get_local_with_rotated_parent`). The standard fixture's parents are translation-only, so the rotation block is identity and equal to its own transpose — a wrong transpose convention in either sync kernel would silently pass every other test. These two tests place the parent at a non-trivial rotation (+90° around Z) and verify the round-trip in both directions. Locally confirmed that flipping the multiply order in either of the two sync kernels makes the matching test fail.

# Iterations during review

* **Stale `localMatrix` after `set_scales`** — added the missing `_sync_local_from_world` call so a write through scale also propagates the new diagonal back into the local matrix.
* **Shared `_fabric_indices` across selections** — each selection now caches its own indices array; `_build_indexed_array` takes them as an explicit parameter, removing the silent-corruption hazard the shared field invited.
* **Stage-keyed dirty flag** — replaced with a per-view bool. With multiple views on a stage, view B reading worlds no longer clears the flag that view A's `set_local_poses` had set (which would have left A's worlds silently stale).
* **`_static_hierarchy_cache` lifecycle** — cache key now `(stage_id, fabric_id)`; added `FabricFrameView.clear_static_caches()` and call it from the test teardown fixture so handles do not accumulate across the suite.
* **Redundant `PrepareForReuse` in syncs** — both `_sync_local_from_world` and `_sync_world_from_local_if_dirty` now refresh `trans_sel_ro` exactly once per call, then read `_world_ifa_ro`, `_local_ifa_ro` and `_parent_world_ifa_ro` directly from the fields, avoiding the duplicated USDRT round-trip.
* **Docstring vs reality on `PrepareForReuse`** — the class docstring now describes the actual behaviour: every accessor calls it; it is cheap and idempotent in the steady state.
* **Clearer error for root-level prims** — `_compute_parent_fabric_indices` now raises a dedicated error when a prim has no parent path, rather than bubbling up the generic "not found in selection" message.

# Type of change

# Checklist

> Per the fragment-based changelog system (#5434), this PR ships fragments at `source/isaaclab/changelog.d/fix-fabric-local-matrix.rst` and `source/isaaclab_physx/changelog.d/fix-fabric-local-matrix.rst` instead of editing `CHANGELOG.rst` and `config/extension.toml` directly — the nightly CI workflow compiles those from the fragments.

### DIFF
--- a/scripts/benchmarks/benchmark_view_comparison.py
+++ b/scripts/benchmarks/benchmark_view_comparison.py
@@ -271,26 +271,71 @@ def _run_pose_benchmarks(
     positions: wp.array,
     orientations: wp.array,
 ):
-    """Shared benchmark loop for get/set world poses on any FrameView."""
+    """Shared benchmark loop for get/set {world,local} poses on any FrameView."""
+
+    # FrameView getters now return ProxyArray; older callers worked with wp.array
+    # directly. Support both transparently.
+    def _as_wp(a):
+        return a.warp if hasattr(a, "warp") else a
+
+    positions_wp = _as_wp(positions)
+    orientations_wp = _as_wp(orientations)
+
     start_time = time.perf_counter()
     for _ in range(num_iterations):
         view.get_world_poses()
     timing_results["get_world_poses"] = (time.perf_counter() - start_time) / num_iterations
 
-    new_positions = wp.clone(positions)
+    new_positions = wp.clone(positions_wp)
     new_positions_t = wp.to_torch(new_positions)
     new_positions_t[:, 2] += 0.5
     expected_positions = new_positions_t.clone()
 
     start_time = time.perf_counter()
     for _ in range(num_iterations):
-        view.set_world_poses(new_positions, orientations)
+        view.set_world_poses(new_positions, orientations_wp)
     timing_results["set_world_poses"] = (time.perf_counter() - start_time) / num_iterations
 
+    # Interleaved set→get on world poses — the realistic write/read pattern for
+    # downstream consumers (e.g. cameras updating their pose then immediately
+    # querying it).
+    start_time = time.perf_counter()
+    for _ in range(num_iterations):
+        view.set_world_poses(new_positions, orientations_wp)
+        view.get_world_poses()
+    timing_results["interleaved_world"] = (time.perf_counter() - start_time) / num_iterations
+
+    # Local poses — Fabric-aware path on FabricFrameView, USD path otherwise.
+    if hasattr(view, "get_local_poses"):
+        start_time = time.perf_counter()
+        for _ in range(num_iterations):
+            view.get_local_poses()
+        timing_results["get_local_poses"] = (time.perf_counter() - start_time) / num_iterations
+
+    if hasattr(view, "set_local_poses"):
+        local_pos, local_ori = view.get_local_poses()
+        local_pos_t = (
+            local_pos.torch
+            if hasattr(local_pos, "torch")
+            else (wp.to_torch(local_pos) if isinstance(local_pos, wp.array) else local_pos)
+        )
+        local_ori_t = (
+            local_ori.torch
+            if hasattr(local_ori, "torch")
+            else (wp.to_torch(local_ori) if isinstance(local_ori, wp.array) else local_ori)
+        )
+        new_local_pos = wp.from_torch(local_pos_t.clone().contiguous())
+        new_local_ori = wp.from_torch(local_ori_t.clone().contiguous())
+
+        start_time = time.perf_counter()
+        for _ in range(num_iterations):
+            view.set_local_poses(translations=new_local_pos, orientations=new_local_ori)
+        timing_results["set_local_poses"] = (time.perf_counter() - start_time) / num_iterations
+
     ret_pos, ret_quat = view.get_world_poses()
-    ret_pos_t = wp.to_torch(ret_pos)
-    ret_quat_t = wp.to_torch(ret_quat)
-    ori_t = wp.to_torch(orientations)
+    ret_pos_t = ret_pos.torch if hasattr(ret_pos, "torch") else wp.to_torch(ret_pos)
+    ret_quat_t = ret_quat.torch if hasattr(ret_quat, "torch") else wp.to_torch(ret_quat)
+    ori_t = wp.to_torch(orientations_wp)
 
     pos_ok = torch.allclose(ret_pos_t, expected_positions, atol=1e-4, rtol=0)
     quat_ok = torch.allclose(ret_quat_t, ori_t, atol=1e-4, rtol=0)
@@ -327,6 +372,9 @@ def print_results(results_dict: dict[str, dict[str, float]], num_prims: int, num
         ("Initialization", "init"),
         ("Get World Poses", "get_world_poses"),
         ("Set World Poses", "set_world_poses"),
+        ("Interleaved Set->Get", "interleaved_world"),
+        ("Get Local Poses", "get_local_poses"),
+        ("Set Local Poses", "set_local_poses"),
     ]
 
     for op_name, op_key in operations:
--- a/source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
+++ b/source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
@@ -642,7 +642,7 @@ class FabricFrameView(BaseFrameView):
         cache_key = (self._stage_id, fabric_id.id)
         if cache_key not in FabricFrameView._static_hierarchy_cache:
             hierarchy = usdrt.hierarchy.IFabricHierarchy().get_fabric_hierarchy(
-                fabric_id, self._stage.GetStageIdAsUInt()
+                fabric_id, self._stage.GetStageIdAsStageId()
             )
             hierarchy.track_local_xform_changes(True)
             hierarchy.track_world_xform_changes(True)
--- a/source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
+++ b/source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
@@ -5,8 +5,7 @@
 
 """PhysX FrameView with Fabric GPU acceleration.
 
-Design follows the prototype in
-``https://github.com/bareya/IsaacLab/tree/pbarejko/camera-update``:
+Design:
 
 * Three persistent ``PrimSelection`` instances differing only in per-attribute access
   mode (one for each of {trans_ro, world_rw, local_rw}).
@@ -28,6 +27,7 @@ Design follows the prototype in
 from __future__ import annotations
 
 import logging
+from typing import ClassVar
 
 import torch
 import warp as wp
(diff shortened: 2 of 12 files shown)


source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -297,171 +417,399 @@ def get_scales(self, indices=None):
...
-                (usdrt.Sdf.ValueTypeNames.Matrix4d, "omni:fabric:worldMatrix", usdrt.Usd.Access.ReadWrite),
-            ],
-            device=self._device,
-        )
+    def _compute_fabric_indices(self, selection) -> wp.array:
+        fabric_paths = selection.GetPaths()
+        path_to_fabric_idx: dict[str, int] = {str(p): i for i, p in enumerate(fabric_paths)}
+        indices: list[int] = []
+        for prim_path in self.prim_paths:
+            fabric_idx = path_to_fabric_idx.get(prim_path)
```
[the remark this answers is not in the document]
pv-nvidia: Not a bug — these are two different "indices" arrays:                                                               
                                                                  
  1. **`wp.indexedfabricarray(indices=...)`** (built by `_compute_fabric_indices`): Warp's `check_index_array`        
  validator explicitly requires `int32` and raises `ValueError` otherwise. `uint32` would break IFA construction
  immediately.                                                                                                        
                                                                  
  2. **Kernel `indices: ArrayUInt32` parameter** (the per-launch `view_indices`): a separate array passed at launch   
  time; correctly `uint32`, matching the `ArrayUInt32` annotation.
                                                                                                                      
  Source: [`warp/_src/types.py#L4474`](https://github.com/NVIDIA/warp/blob/release-1.12/warp/_src/types.py#L4474)     
  ```python
  def check_index_array(indices, expected_device):                                                                    
      ...                                                         
      if indices.dtype != int32:
          raise ValueError(f"Index array must use int32, got dtype {indices.dtype}")
      ```                                                                                                                                    
  The 41 unit tests and the comparison benchmark exercise both paths end-to-end on cpu and cuda:0 without issue — if  
  the IFA dtype were wrong, construction would fail on the first call.

scripts/benchmarks/benchmark_view_comparison.py
```diff
@@ -271,26 +271,71 @@ def _run_pose_benchmarks(
...
+
     start_time = time.perf_counter()
     for _ in range(num_iterations):
         view.get_world_poses()
     timing_results["get_world_poses"] = (time.perf_counter() - start_time) / num_iterations
 
-    new_positions = wp.clone(positions)
+    new_positions = wp.clone(positions_wp)
     new_positions_t = wp.to_torch(new_positions)
     new_positions_t[:, 2] += 0.5
```
pv-nvidia, line 291: Does this also change the positions buffer itself? Or is a copy made?

And why are we changing the positions but not the orientations?

And we completely ignore scaling here?

What is the point of this all?

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -49,23 +50,71 @@ def _to_float32_2d(a: wp.array | torch.Tensor) -> wp.array | torch.Tensor:
...
+    # share one ``IFabricHierarchy`` handle, keyed by ``(stage_id, fabric_id_int)``
+    # so that a recycled ``stage_id`` paired with a fresh Fabric attachment never
+    # reuses a stale handle.  The Fabric id is stored as a plain ``int`` (extracted
+    # from ``FabricId.id``) because the ``FabricId`` wrapper itself is freshly
+    # allocated on every ``GetFabricId()`` call and has no value equality, which
+    # would defeat the cache.  ``ClassVar`` plus the ``_static_`` prefix make it
+    # explicit (and enforceable by type checkers) that this is class-level — never
+    # shadow it with ``self.<name> = ...``.  The cache is not protected by a lock;
+    # Isaac Lab's simulation loop is single-threaded, and adding locking would
+    # negate the per-stage hit-cache cost it exists to avoid.  Call
```
pv-nvidia: TODO: Improve this comment, it is not clear IMO

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -49,23 +50,71 @@ def _to_float32_2d(a: wp.array | torch.Tensor) -> wp.array | torch.Tensor:
...
+    # :meth:`clear_static_caches` on stage teardown if you tear down many stages
+    # in one process.
+    _static_hierarchy_cache: ClassVar[dict[tuple[int, int], object]] = {}
+
+    @classmethod
+    def clear_static_caches(cls) -> None:
+        """Drop all cached ``IFabricHierarchy`` handles.
+
+        Call this after tearing down a USD stage if the process will continue
```
pv-nvidia: TODO: Improve this comment, it is not clear IMO

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -101,14 +150,45 @@ def __init__(
...
-        self._fabric_selection = None
-        self._fabric_to_view: wp.array | None = None
-        self._view_to_fabric: wp.array | None = None
-        self._default_view_indices: wp.array | None = None
+        self._stage_id: int | None = None
+        self._stage = None
         self._fabric_hierarchy = None
-        self._view_index_attr = f"isaaclab:view_index:{abs(hash(self))}"
+        # Set by ``set_local_poses``; cleared by ``_sync_world_from_local_if_dirty``.
+        # Per-view (not per-stage) so concurrent views on the same stage don't clear
```
pv-nvidia, line 129: "concurrent" contradicts the mention of single threaded earlier

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -153,48 +233,43 @@ def set_world_poses(self, positions=None, orientations=None, indices=None):
         if not self._fabric_initialized:
             self._initialize_fabric()
 
-        self._prepare_for_reuse()
+        # If a prior set_local_poses left worldMatrix stale, propagate local → world first.
```
pv-nvidia, line 206: TODO: Explain this is needed because not all world poses are overwritten here

pv-nvidia:

PR #5523, opened 2026-05-06
[Newton] Bump Newton pin to v1.2.0rc2

hujc7:
## Summary

Bumps the Newton pin to [`v1.2.0rc2`](https://pypi.org/project/newton/1.2.0rc2/), which pulls in IsaacLab-relevant fixes plus the upstream tendon-scoping fix.

## What's new in Newton v1.2.0rc2 vs IsaacLab's current pin (`a27277e`)

The current IsaacLab Newton pin is from late April; v1.2.0rc2 is the latest release-candidate cut. Notable fixes pulled in:

- **[newton-physics/newton#2659](https://github.com/newton-physics/newton/pull/2659)** \"Scope USD custom-frequency parsing\" — `parse_usd` now scopes the custom-frequency walk to `root_path` natively.
- **[newton-physics/newton#2678](https://github.com/newton-physics/newton/pull/2678)** Regression fix.
- **[newton-physics/newton#2720](https://github.com/newton-physics/newton/pull/2720)** `SolverKamino` reset under `world_mask`.
- **[newton-physics/newton#2710](https://github.com/newton-physics/newton/pull/2710)** VRAM leak fix on example reset.
- Plus 16 other smaller fixes between rc1 and rc2.

## Required dep bumps

Newton 1.2.0rc2's \`pyproject.toml\` requires:

- \`warp-lang==1.13.0\`
- \`mujoco==3.8.0\` (was 3.6.0)
- \`mujoco-warp==3.8.0.1\` (was 3.6.0)

Pins updated in:

| File | Change |
|---|---|
| \`source/isaaclab/setup.py\` | \`warp-lang==1.12.0\` → \`==1.13.0\`; \`mujoco==3.6.0\` → \`==3.8.0\`; \`mujoco-warp==3.6.0\` → \`==3.8.0.1\` |
| \`source/isaaclab_newton/setup.py\` | mujoco / mujoco-warp bumps; Newton pin → \`v1.2.0rc2\` |
| \`source/isaaclab_visualizers/setup.py\` | 3× Newton pin → \`v1.2.0rc2\` |
| \`tools/wheel_builder/res/python_packages.toml\` | All four pins mirrored |

## Code adapts

\`warp-lang\` 1.13 removed the \`wp.math\` namespace. Two IsaacLab call sites use it:

- \`source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py:72\`
[...]

PR #5525, opened 2026-05-06
[changelog] Allow dots in fragment slugs

hujc7:
## Summary

The changelog fragment gate (\`tools/changelog/cli.py\`) currently rejects any slug containing a dot. That contradicts AGENTS.md's recommendation to use the branch name as the slug, since branches routinely carry version numbers — \`bump-newton-1.2.0rc2\`, \`fix-foo-v2.5\`, etc. Following the docs literally produces a slug that the CLI then rejects, requiring a manual rename step on every version-bearing PR.

This PR loosens the parser to accept any character valid in a git refname segment, except \`/\` (which would create a subdirectory inside \`changelog.d/\`).

### Parser

- Replaced the \`FRAGMENT_RE\` / \`SKIP_RE\` regex pair with a \`FragmentFilename\` value object.
- Reverse-parses against a closed suffix tuple — \`(.minor.rst, .major.rst, .skip, .rst)\` — with the longest match winning. So \`foo.minor.rst\` is unambiguously \`slug='foo' tier='minor'\` no matter how many dots the slug otherwise carries.
- Slug validation mirrors \`git check-ref-format\` per-segment rules: non-empty, no leading \`.\` or \`-\`, no trailing \`.\` or \`.lock\`, no \`..\` or \`@{\`, no whitespace or \`~ ^ : ? * [ \\\`. Plus \`/\` (path separator).

### OO consolidation

Before this PR, three classes parsed filenames three different ways:

| Class | Mechanism |
|---|---|
| \`Fragment\` | \`@cached_property _match\` calling \`FRAGMENT_RE.match\` |
| \`Fragment.parse_slug\` | \`@staticmethod\` calling both regex constants |
| \`FragmentBatch.from_dir\` / \`PRDiff.check_fragments\` | direct \`SKIP_RE.match\` / \`FRAGMENT_RE.match\` |

After: all three go through one \`FragmentFilename\` value object that exposes \`.is_fragment\`, \`.is_skip\`, \`.slug\`, \`.tier\`. One source of truth, no scattered regex matches.

### Error message

The contributor-facing error now spells out the rules explicitly so a future contributor with a dotted branch name doesn't have to dig through the source:

\`\`\`
invalid filename — must be <slug>.rst, <slug>.minor.rst,
<slug>.major.rst, or <slug>.skip. Slug rules mirror git refnames
(excluding \`/\`): non-empty, no whitespace or any of \`~ ^ : ? * [ \\\`,
no leading \`.\` or \`-\`, no trailing \`.\` or \`.lock\`, no \`..\` or \`@{\`.
Dots inside the slug are fine (e.g. \`bump-newton-1.2.0rc2.minor.rst\`).
\`\`\`

## Tests

- New parametrized cases in \`test_bump_suffix.py\` covering dotted version-bearing slugs and the eight git-refname rejection cases (leading \`.\`, leading \`-\`, trailing \`.\`, \`..\`, \`.lock\`, whitespace, \`~\`, \`^\`, \`/\`).
- Replaced the two stale \`invalid_filenames/\` fixtures (\`multi.dot.slug.rst\`, \`1234.notabump.rst\` — both now legitimately valid) with two genuinely-invalid ones (\`-leading-dash.rst\`, \`has..consecutive-dots.rst\`).
- Updated \`test_parse.py\` and \`test_validate.py\` to reference the new fixtures.
- Full suite: 92 passed.

## Test plan

- [x] \`pytest tools/changelog/test/\` — all 92 pass.
- [x] \`./isaaclab.sh -f\` clean.
- [x] Smoke-tested the parser against the original dotted slug from #5523 (\`jichuanh-newton-1.2.0rc2-bump.minor.rst\`) plus version-bearing slugs and the rejection cases — behaves correctly in all 8 spot-checks.

## Why no changelog fragment

This PR only touches \`tools/changelog/\` (CI tooling), not any package under \`source/\`. The gate only requires fragments for touched \`source/\` packages.


hujc7:
Folded into https://github.com/isaac-sim/IsaacLab/pull/5867 and closing this in its favour.

Both PRs rewrite the same code — `FRAGMENT_RE` / `SKIP_RE` and every consumer of them — and #5867 additionally moves that code into a new `packages.py`. Landing them separately meant whichever went second would need a hard rebase over a file that had moved, so carrying the change there costs one resolution instead of two.

Nothing was dropped. `FragmentFilename`, the derived `pattern_summary()` / `help_lines_for_package()`, and the fixture swap all came across intact, and the tests moved with them.

Two things were added on top:

- The docstring claimed the slug rules mirror `git check-ref-format`. That's the raw refname check, which is *more* permissive than what's implemented — it accepts a leading `-`. The rule actually wanted, and implemented, is `git check-ref-format --branch`: the names git will let you give a branch. That matters because the convention is "slug = your branch name", and `git branch -- -foo` fails for the same reason the gate rejects `-foo.rst`. Wording corrected.
- A differential test now asks git directly across a corpus of slugs and asserts our validator agrees, rather than restating the rules a third time. `/` is the single deliberate divergence — git allows it in a branch, a filename cannot contain it — and has its own test saying so. Verified the guard bites: loosening the rule by one character fails the test and names the input that diverged.

Thanks — the motivating case (`bump-newton-1.2.0rc2.minor.rst`) works, and the argument that three classes were each re-deriving what counts as a fragment is the same duplication argument #5867 is built on, one layer down.

PR #5539, opened 2026-05-08
[DO NOT MERGE] Test revert of PR #5521 to isolate viewergl failure

hujc7:
## Purpose

Hypothesis test only — do not merge. Validates whether https://github.com/isaac-sim/IsaacLab/pull/5521 introduced the \`test_cartpole_newton_visualizer_viewergl_rgb_motion\` deterministic failure observed across https://github.com/isaac-sim/IsaacLab/pull/5523, https://github.com/isaac-sim/IsaacLab/pull/5534, https://github.com/isaac-sim/IsaacLab/pull/5024, https://github.com/isaac-sim/IsaacLab/pull/5495, https://github.com/isaac-sim/IsaacLab/pull/5492 (skip in https://github.com/isaac-sim/IsaacLab/pull/5538).

If the visualizer test **passes** here → https://github.com/isaac-sim/IsaacLab/pull/5521's camera-cfg lazy-import change is the trigger and the fix lives in adjusting that init order.

If it **fails** here → https://github.com/isaac-sim/IsaacLab/pull/5521 isn't the cause and the regression is elsewhere (likely the daily Isaac Sim image).

## Diff

Single revert of \`7b44452e9fa\` (https://github.com/isaac-sim/IsaacLab/pull/5521 merge commit) on top of develop tip.

PR #5521, opened 2026-05-06
use RendererCfg as default renderer_cfg in CameraCfg

r-schmitt:
# Description

the camera config was importing `isaaclab_physx.renderers` because the default render_cfg was set to that config.  this PR sets that to RendererConfig to remove the import, but provides a get_default_render_config method to the backend_utils to lazily import the config if needed.  this is called __post_init__ on the camera config to replace the generic config as soon as possible to avoid downstream issues referencing the renderer config.  this action can be moved to the factory if downstream references are cleaned up.

## Type of change

- Refactor to remove imports in cfg class

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5540, opened 2026-05-08
[debug] Instrument ViewerGL.get_frame for fully-black diagnosis

hujc7:
Diagnostic-only PR. Adds conftest.py to log FBO/PBO/CUDA-GL state during the failing viewergl_rgb_motion test.

The viewergl test has been deterministically failing on develop with 'Viewer frame appears fully black' since 2026-05-07 17:45 UTC. PR https://github.com/isaac-sim/IsaacLab/pull/5521 hypothesis was disproven by https://github.com/isaac-sim/IsaacLab/pull/5539. Same Kit session: tiled_camera (Kit RTX) passes, viewergl (Newton pyglet/EGL) fails. Bug is specific to Newton ViewerGL after Kit init.

This patch adds direct CPU FBO readback (bypassing PBO/CUDA) so we can tell whether the FBO is empty (rendering didn't happen) vs the CUDA-GL interop silently zeros the readback.

DO NOT MERGE - diagnostic only.

### DIFF
--- a/source/isaaclab_visualizers/test/conftest.py
+++ b/source/isaaclab_visualizers/test/conftest.py
@@ -1,0 +1,129 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Diagnostic instrumentation for ViewerGL.get_frame.
+
+Logs context state and the PBO/CUDA-GL interop result so we can tell whether
+the FBO itself is empty (rendering didn't happen) or whether the PBO/CUDA
+readback path returned zeros while the FBO had content.
+
+Active whenever this conftest is loaded.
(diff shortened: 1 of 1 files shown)


hujc7:
Data captured. Diagnostic confirmed the FBO itself is empty — direct CPU `glReadPixels` of the FBO returns all zeros with `GL_NO_ERROR` before and after, identical to the PBO/CUDA path's result. Rules out CUDA-GL interop / sync / context-currency as the root cause:

```
[VIZDIAG] fbo=c_uint(8)  pbo=None  size=600x600
[VIZDIAG] glGetError before: GL_NO_ERROR
[VIZDIAG] CPU-readback: nonzero=0/1080000  max=0  err=GL_NO_ERROR
[VIZDIAG] PBO-result: nonzero=0/1080000  max=0
```

The Newton pyglet/EGL surface is attached but inert under Kit 110.1.1 — `render()` deposits no pixels into the FBO. Investigation continues without this PR. Skip is shipping in https://github.com/isaac-sim/IsaacLab/pull/5538.

PR #5541, opened 2026-05-08
[H6 fix attempt] Make Newton ViewerGL context current before get_frame

hujc7:
Fix-attempt PR for the viewergl-fully-black regression on develop.

Hypothesis: Newton's `ViewerGL.get_frame` does not call `self._make_current()`. If Kit's GLFW context is current when `get_frame()` runs (after `env.step()` ticks Kit), Newton's GL calls (`glBindFramebuffer`, `glReadPixels`, `RegisteredGLBuffer.map`) operate on Kit's context — Newton's FBO/PBO IDs are invalid there, so the readback returns zeros. The bug presents as 'Viewer frame appears fully black'.

This patch wraps `get_frame` via a test conftest to call `_make_current()` first.

If `test_cartpole_newton_visualizer_viewergl_rgb_motion` passes with this in place, the proper fix is upstream in Newton (`newton/_src/viewer/viewer_gl.py:get_frame` should call `self._make_current()` at the start).

Related diagnostic in https://github.com/isaac-sim/IsaacLab/pull/5540. PR https://github.com/isaac-sim/IsaacLab/pull/5521 hypothesis already disproven by https://github.com/isaac-sim/IsaacLab/pull/5539.

DO NOT MERGE — testing only.

### DIFF
--- a/source/isaaclab_visualizers/test/conftest.py
+++ b/source/isaaclab_visualizers/test/conftest.py
@@ -5,16 +5,16 @@
 
 """H6 fix attempt: make Newton ViewerGL's GL context current before get_frame().
 
-Newton's ViewerGL.get_frame issues GL calls (glBindFramebuffer, glReadPixels,
-RegisteredGLBuffer.map) without first calling self._make_current(). If Kit's
-GLFW window context is current at the moment get_frame() runs, those GL calls
-operate on Kit's context where Newton's FBO/PBO ids are invalid. The result
-silently reads from Kit's framebuffer (empty for the visualizer test) or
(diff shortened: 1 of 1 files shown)


hujc7:
Hypothesis disproven. With `self._make_current()` wrapped around `get_frame`, the test still fails with the same 'Viewer frame appears fully black' assertion. Combined with the diagnostic in https://github.com/isaac-sim/IsaacLab/pull/5540 (which shows `GL_NO_ERROR` throughout), context-currency is not the cause — Newton's pyglet/EGL context is current, GL is happy, the FBO is just empty. Closing.

PR #5542, opened 2026-05-08
[H4 fix attempt] Add explicit glFinish + wp.synchronize_device before PBO map

hujc7:
Fix-attempt PR for the viewergl-fully-black regression on develop.

Hypothesis (H4): warp 1.13 (or a Kit/Replicator version flip in the daily `nvcr.io/nvidian/isaac-sim:latest-develop` image) changed the implicit CUDA-OpenGL sync semantics inside `wp.RegisteredGLBuffer.map`, so the CUDA read of the GL PBO can race the asynchronous `glReadPixels` write into that same PBO.

This patch adds an explicit `glFinish()` + `wp.synchronize_device()` between the PBO write and the CUDA-side map. If the test passes with this, the proper fix is upstream in Newton's `get_frame`.

Companion of https://github.com/isaac-sim/IsaacLab/pull/5540 (diagnostic) and https://github.com/isaac-sim/IsaacLab/pull/5541 (H6 _make_current attempt). PR https://github.com/isaac-sim/IsaacLab/pull/5521 hypothesis disproven by https://github.com/isaac-sim/IsaacLab/pull/5539.

DO NOT MERGE — testing only.

### DIFF
--- a/source/isaaclab_visualizers/test/conftest.py
+++ b/source/isaaclab_visualizers/test/conftest.py
@@ -1,0 +1,105 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""H4 hypothesis: warp 1.13 changed implicit CUDA-OpenGL sync semantics.
+
+Monkey-patches ``newton._src.viewer.viewer_gl.ViewerGL.get_frame`` to insert an
+explicit ``glFinish()`` and ``warp.synchronize_device()`` between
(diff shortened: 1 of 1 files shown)


hujc7:
Hypothesis disproven. With explicit `glFinish()` + `wp.synchronize_device()` inserted between `glReadPixels` and `wp.RegisteredGLBuffer.map`, every frame still reads back `nonzero=0`:

```
[H4] returned shape=(600, 600, 3)  nonzero=0  first24=[0, 0, 0, ...]
```

Combined with the diagnostic in https://github.com/isaac-sim/IsaacLab/pull/5540 (CPU FBO readback also zero), GL/CUDA sync is not the cause — the FBO itself is empty, so no amount of post-readback sync can produce non-zero pixels. Closing.

PR #5536, opened 2026-05-07
Fix OvPhysX integration patches for kitless mode

fatimaanes:
# Description

Three small fixes for OvPhysX (kitless) physics backend integration discovered during benchmarking:

1. **SensorBase prim deletion callback guard** — `SensorBase` checked `"physx" in __name__.lower()` to register a PhysX-specific prim deletion callback, but this also matched `OvPhysxManager` (which contains "physx" in its name). Added an explicit exclusion for OvPhysX so the PhysX-only callback is not incorrectly registered.

2. **Remove unsupported `gpu_index` parameter** — `OvPhysxManager._warmup_and_load()` passed `gpu_index` to the `ovphysx.PhysX()` constructor, but the current ovphysx wheel does not accept that parameter, causing a `TypeError` on GPU device initialization.

3. **Add OvPhysX physics backend preset for Cartpole Camera** — The `PhysicsCfg` preset class in the Cartpole Camera Direct environment was missing an `ovphysx = OvPhysxCfg()` entry, so `presets=ovphysx` could not be used for this task.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5543, opened 2026-05-08
Fix docs build: pin warp intersphinx to /latest/

hujc7:
## Summary

`https://nvidia.github.io/warp/objects.inv` now returns 404 — the Warp docs published `objects.inv` only at versioned subpaths (`/latest/`, `/stable/`, etc.). With Sphinx's `warnings treated as errors`, the broken intersphinx fetch fails the docs build on **every** PR (#5536, #5538, #5540-#5542 all failing on `Build Latest Docs`).

Pinning to `/latest/` matches what other intersphinx targets in this file already do (see the existing PyTorch `/docs/2.11/` pin, also added to work around a /stable/ 404).

## Test plan

- [x] Verified `https://nvidia.github.io/warp/latest/objects.inv` returns 200
- [ ] CI `Build Latest Docs` job passes on this PR (was failing on develop)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -140,7 +140,8 @@ intersphinx_mapping = {
     "torch": ("https://docs.pytorch.org/docs/2.11/", None),
     "isaacsim": ("https://docs.isaacsim.omniverse.nvidia.com/6.0.0/py/", None),
     "gymnasium": ("https://gymnasium.farama.org/", None),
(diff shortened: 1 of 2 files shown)


hujc7:
Folded into https://github.com/isaac-sim/IsaacLab/pull/5538 — both unblock develop CI in one PR.

PR #5551, opened 2026-05-08
Update configs to new conventions from rsl_rl >= 5.0

StafaH:
# Description

Formatted articulation root predicate expressions for readability. No behavior change and no new dependencies.

## Type of change
 
Code cleanup and migration to rsl_rl >= 5.0

## Checklist


ClemensSchwarke:
Hi @StafaH, thanks a lot for the PR! The only problem I am seeing (and why I did not update the configs) is that there is a function that updates old configs to the new structure but not the other way around because it is already quite lengthy. Updating the configs would thus break the tasks if the user uses an older rsl-rl version. The options I am seeing are 1. also including logic for converting new configs to the old structure, 2. not updating the configs, or 3. accepting that the default tasks break for older rsl-rl versions.

Cheers,
Clemens

StafaH:
Hi @ClemensSchwarke, I synced with @kellyguo11 and we prefer option 3. The develop branch pins rsl_rl >= 5.0 so we can begin migrating users to this with a warning or an error if we detect < 5.0 in the train script.

I can also flip the old script to create the old config from the new config if an older version of rsl_rl is detected, inverting the whole setup (isaaclab 3.0 will use new configs and rsl_rl >= 5.0. Not sure if this needed but if you recommend then we can do that.

ClemensSchwarke:
I agree for lab 3.0, I missed that this PR isn't going to main. Thanks @StafaH !

ClemensSchwarke (approved):

PR #5578, opened 2026-05-11
Migrate camera/renderer/camera data to warp

StafaH:
# Description

Migrates `Camera`, `CameraData`, and all renderer backends to warp-backed `ProxyArray`,
consistent with the rest of IsaacLab's sensors (ContactSensor, RayCaster, IMU, etc.).

`CameraData` fields (`pos_w`, `quat_w_world`, `intrinsic_matrices`, `output`) now return
`ProxyArray` instead of `torch.Tensor`. Use `.torch` for a zero-copy tensor view or pass
directly to warp kernels. `RenderBufferSpec.dtype` is now a warp dtype (e.g. `wp.float32`).

The `ProxyArray` deprecation bridge means existing torch usage continues to work with a
one-time `DeprecationWarning`.

## Type of change

- Possible breaking change (existing functionality will not work without user modification)

## Checklist

- [ x I have added my name to the `CONTRIBUTORS.md` or my name already exists there


source/isaaclab_visualizers/test/test_visualizer_cartpole_integration.py
```diff
@@ -341,7 +341,8 @@ def _step_until_non_black_camera(env, actions: torch.Tensor, *, max_steps: int =
         rgb = env._tiled_camera.data.output.get("rgb")
         if rgb is None:
             rgb = env._tiled_camera.data.output[env.cfg.tiled_camera.data_types[0]]
-        last_rgb = rgb
+        # TODO: We should make the assert not use torch tensors as we migrate to warp.
```
pbarejko: Is this left out on purpose?

source/isaaclab/isaaclab/utils/warp/warp_math.py
```diff
@@ -0,0 +1,207 @@
...
+    indices: wp.array(dtype=wp.int32),
+    q_const: wp.quatf,
+):
+    """Apply constant-quaternion convention conversion to indexed elements.
+
+    Reads ``src[i]`` and writes to ``dst[indices[i]]``.  Use this for partial
+    camera updates (e.g. environment resets targeting a subset of cameras).
+    """
+    i = wp.tid()
+    dst[indices[i]] = quat_mul_xyzw(src[i], q_const)
```
fatimaanes: If the above is removed, this and the other kernel simplify to:
`dst[i] = src[i] * q_const`

source/isaaclab/isaaclab/utils/warp/warp_math.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Returns:
+        Product quaternion ``(x, y, z, w)``.
+    """
+    x = q1[3] * q2[0] + q1[0] * q2[3] + q1[1] * q2[2] - q1[2] * q2[1]
+    y = q1[3] * q2[1] - q1[0] * q2[2] + q1[1] * q2[3] + q1[2] * q2[0]
+    z = q1[3] * q2[2] + q1[0] * q2[1] - q1[1] * q2[0] + q1[2] * q2[3]
+    w = q1[3] * q2[3] - q1[0] * q2[0] - q1[1] * q2[1] - q1[2] * q2[2]
+    return wp.quatf(x, y, z, w)
+
```
fatimaanes: This custom quat_mul_xyzw kernel might be redundant. warp's wp.quatf uses xyzw order (not wxyz as the comment states), so the native q1 * q2 computes the same Hamilton product. The Isaac Lab 3.0 migration switched to XYZW specifically to align with Warp, PhysX, and Newton. Can this function be removed?

StafaH: Thanks, I assumed wrongly it would be wxyz since that's what I'm used to in MuJoCo. Removed and fixed the related comment as well

source/isaaclab_visualizers/test/test_visualizer_cartpole_integration.py
StafaH: I have removed the comment, but in general as we migrate to warp, we need to begin removing all these torch testing utilities. We should opt for np.testing for array comparisons using the warp array with .numpy(). Removed the comment for now as this is actually a wider issue that we can address altogether later.

pbarejko (approved):

fatimaanes (approved):

ooctipus (approved):

PR #5587, opened 2026-05-12
[PresetCLI] Add typed preset selection with task-aware help

hujc7:
## Summary

Adds typed preset selection via Hydra-style tokens — `physics=NAME` / `renderer=NAME` / `presets=NAME[,...]` — that fold into the existing `presets=<csv>` Hydra-decorator flow. Makes `--task=X --help` list the actual `PresetCfg` variants present in that task's env_cfg, bucketed by typed target. Adopted in all 16 Hydra-using scripts (`rl_games/sb3/skrl/rsl_rl` train+play, `environments/*`, `benchmarks/*`, `sim2sim_transfer`, `leapp/rsl_rl/export`).

## API shape

```python
parser = argparse.ArgumentParser(...)
# ... script-specific args ...
add_launcher_args(parser)
args_cli, hydra_args = setup_preset_cli(parser)
sys.argv = [sys.argv[0]] + hydra_args
```

`setup_preset_cli` returns `(args, hydra_argv)` without mutating `sys.argv`. It registers no argparse flags for preset selection — the typed selectors are recognized as Hydra-style tokens in the `parse_known_args` remainder and folded into `hydra_argv[0]` as a single `presets=<csv>` token.

## Grammar

```
python train.py --task=X  physics=newton_mjwarp  renderer=newton_renderer  presets=albedo,depth
```

- `physics=NAME` — typed selector for `PhysicsCfg` variants
- `renderer=NAME` — typed selector for `RendererCfg` variants
- `presets=NAME[,NAME,...]` — broadcast: applied to every matching `PresetCfg`

All three fold into one `presets=<csv>` token: `presets=newton_mjwarp,newton_renderer,albedo,depth`.

The grammar matches Hydra's, so the same line can carry path-targeted overrides (`env.sim.dt=0.001`) that flow through untouched.

## Namespace contract

No preset selector is registered with argparse, so the parsed `args` namespace gains no `physics` / `renderer` / `presets` attribute. AppLauncher's name-based forwarding (`set(_SIM_APP_CFG_TYPES) & set(vars(args))`, `app_launcher.py:681`) therefore cannot pick up a preset value and push it into `SimulationApp.config` — the historical `--renderer` → `config["renderer"]` → `None.lower()` crash class is structurally impossible. Two regression tests lock the contract.

## Help text layout

`--task=Isaac-Cartpole-v0 --help` renders each selector with its available variants inline directly below it (bullets aligned with the description column):

```
preset selection:
  Select named PresetCfg alternatives via Hydra-style overrides (key=value, no leading dashes):
      physics=NAME              (typed) selects a PhysicsCfg variant. Available:
                                - newton_kamino
                                - newton_mjwarp
                                - physx
      renderer=NAME             (typed) selects a RendererCfg variant. Available:
                                (none)
      presets=NAME[,NAME,...]   broadcast: applied to every matching PresetCfg. Available:
                                (none)

  Hydra also accepts path-targeted overrides like env.sim.physics=NAME.
```

Typed variants appear only under their own typed selector. The `presets=` listing shows only DOMAIN-bucket variants (cfgs whose type doesn't subclass any typed target's base class).

Without `--task`, each row shows just the selector + description and the section adds a `Pass --task=X` hint on its own paragraph.

## Test plan

- [x] `pytest source/isaaclab_tasks/test/test_preset_cli.py` — 24 tests pass. Coverage: enum wiring, token folding/dedupe/passthrough, `_ArgvHelper` semantics, type-based bucketing, all four help-text branches (parametrized), no-`sys.argv`-mutation contract, namespace-clean contract, AppLauncher intersection contract, `hydra_args[0]` preserves the `presets=` token for benchmark telemetry.
- [x] `pytest source/isaaclab_tasks/test/test_hydra.py` — 76 tests pass; legacy-alias `FutureWarning` behavior unchanged.
- [x] `pre-commit` clean.
- [x] Manual: `--task=Isaac-Cartpole-v0 --help` and `--task=Isaac-Cartpole-RGB-Camera-Direct-v0 --help` render correctly.
- [x] Manual: `physics=newton_mjwarp renderer=newton_renderer presets=albedo` folds into one `presets=<csv>` token at `hydra_argv[0]`.
- [x] Manual: unknown name → grouped error from resolver; legacy alias `newton` → `FutureWarning` and resolves to `newton_mjwarp`.


source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager_cfg.py
```diff
@@ -9,8 +9,10 @@
 
 from isaaclab.physics import PhysicsCfg
 from isaaclab.utils import configclass
+from isaaclab.utils.preset_registry import PresetTarget, register
 
 
+@register(PresetTarget.PHYSICS, "ovphysx")
```
ooctipus: I wonder if hydra/preset should ever be a dependency in non-isaaclab-task, when I was designing the hydra.py. I wanted to be only task-level dependency and not a cross package dependency.

scripts/benchmarks/benchmark_non_rl.py
```diff
@@ -46,15 +48,11 @@
 
 # append AppLauncher cli args
 AppLauncher.add_app_launcher_args(parser)
-# parse the arguments
-args_cli, hydra_args = parser.parse_known_args()
-# always enable cameras to record video
+args_cli, hydra_args = setup_preset_cli(parser)
+sys.argv = [sys.argv[0]] + hydra_args
 if args_cli.video:
     args_cli.enable_cameras = True
```
StafaH, line 55: Out of scope for this PR but is enable_cameras still used?

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager_cfg.py
hujc7: The main problem is that without this thin register, we would need to have a hard coded list in hydra for those predefined names or types, which could potentially drift. IMO, It's more natural to put things where it's first appearing, so it's easier to maintain or scale.

ooctipus: my main concern is that preset is really just build around idea of user configuration customization, and the issue we are having is that its a bit too flexible that is confusing. Preset has 0 value beyond task-time variation, and becomes very concerning when it leaks to core package. If you want to add a registry like this, I would think it around Idea beyond just isaaclab-task useage. Maybe a general tagging system is what you want, and preset is just leveraging the tagging system?

hujc7: Let me check if there's a way to work without the register. To some extent yes I think, adding meta info to types so it's easier to identify them.

ooctipus:
Thanks for taking this work from me@hujc7 !
Few questions: 

If I am understanding correctly, preset_registry is just assign naming convention to each subclass? if that's the case can we build something that more along idea of generic tagging system, rather than blending with idea of preset? preset is basically just leveraging existing name and avoiding flexibile naming

we prototyped some naming idea in https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab_tasks/isaaclab_tasks/utils/presets.py

```
from isaaclab_tasks.utils import PresetCfg


@configclass
    default: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
    newton_renderer: NewtonWarpRendererCfg = NewtonWarpRendererCfg()
    ovrtx_renderer: OVRTXRendererCfg = OVRTXRendererCfg()
    isaacsim_rtx_renderer = default
```

and if user import as their renderer, the environment is ensured to have consistent naming.


My second questions is:

how would adding this preset_registry help the isaaclab_tasks? as far as I am understanding we still need to list each option with its implementation like code snippet above, I was hoping for if we have this tagging system some of these burden can be lifted, but it doesn't seem to help in that direction? or correct me if I am wrong

scripts/reinforcement_learning/rsl_rl/play.py
```diff
@@ -64,24 +64,25 @@
...
+# Call an external callback if requested. setup_preset_cli does not mutate
+# sys.argv, so the callback sees the user's original command line with the
+# typed --physics / --renderer / --presets flags intact. The callback returns
+# the args it did NOT consume; we intersect with hydra_args' non-preset tail
+# so the resolver only sees args neither side consumed, while the folded
+# "presets=..." token is held aside and always reaches Hydra.
 if args_cli.external_callback:
     external_callback_function = string_to_callable(args_cli.external_callback, separator=".")
     remaining_args_env_registration = external_callback_function()
+    presets_tokens = [t for t in hydra_args if t.startswith("presets=")]
```
ooctipus: is this addition needed?

scripts/reinforcement_learning/rsl_rl/train.py
```diff
@@ -69,24 +69,25 @@
...
-remaining_args_env_registration = None
+# Call an external callback if requested. setup_preset_cli does not mutate
+# sys.argv, so the callback sees the user's original command line with the
+# typed --physics / --renderer / --presets flags intact. The callback returns
+# the args it did NOT consume; we intersect with hydra_args' non-preset tail
+# so the resolver only sees args neither side consumed, while the folded
+# "presets=..." token is held aside and always reaches Hydra.
 if args_cli.external_callback:
     external_callback_function = string_to_callable(args_cli.external_callback, separator=".")
     remaining_args_env_registration = external_callback_function()
```
ooctipus, line 89: is this addition needed?

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
```diff
@@ -0,0 +1,333 @@
...
+        free-form) was given or a pre-existing ``presets=...`` token was
+        present in the remainder; otherwise the list contains only the
+        non-preset remainder.
+    """
+    # Peek for --task before argparse parses. argparse short-circuits on --help,
+    # so help text that depends on the task has to find it ahead of parser run.
+    # Skip the env_cfg load when --help isn't requested -- normal training runs
+    # don't need the variant enumeration and hydra walks the cfg later anyway.
+    task_name = _peek_task()
+    actual_variants = _enumerate_variants(task_name) if (task_name and _help_requested()) else None
```
ooctipus: _peek_task seems to do 1 scan 
```
  def _peek_task() -> str | None:
      last_task: str | None = None
      for i in range(1, len(sys.argv)):
```

_help_requested is doing the 2nd scan:
```
  def _help_requested() -> bool:
      return any(token in ("--help", "-h") for token in sys.argv[1:])   # ← scan #2
```

maybe it can be reduced to only one scan of argv?

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -0,0 +1,107 @@
...
+        Resolver-layer code (in :mod:`isaaclab_tasks.utils.hydra`) needs a
+        target-agnostic lookup -- the ``presets=...`` token is target-agnostic
+        on the wire. Builds fresh from per-target tables so this enum stays
+        the single source of truth.
+
+        Returns:
+            Mapping of every legacy alias to its canonical replacement,
+            aggregated across all members.
+        """
+        return {name: rep for target in cls for name, rep in target.legacy_aliases.items()}
```
ooctipus, line 128: Be careful this is rebuilding the dictionary for every lookup, before it was a dictionary constant and has O(1) cost. now it is rebuilding dictionary for every look up

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
```diff
@@ -0,0 +1,333 @@
...
+
+    Returns:
+        Mapping with one entry per :class:`PresetTarget` member.
+    """
+    typed_targets = [t for t in PresetTarget if t.base_classes]
+    result: dict[PresetTarget, set[str]] = {target: set() for target in PresetTarget}
+    for path_dict in walked.values():
+        for name, cfg in path_dict.items():
+            if name == "default":
+                continue
```
ooctipus, line 346: doing this this way will rewalk the preset dictionary output that collect_preset has already walked, I am wondering maybe this implementation can just go hydra.py's collect_presets or a sibling, so not introducing duplicate walking.

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -407,9 +402,9 @@ def _format_unknown_presets_error(unknown: set[str], name_to_paths: dict[str, li
         fingerprint_to_names.setdefault(key, []).append(name)
 
     lines = [f"Unknown preset(s): {', '.join(sorted(unknown))}"]
-    deprecated_hits = sorted(name for name in unknown if name in _LEGACY_PRESET_ALIASES)
+    deprecated_hits = sorted(name for name in unknown if name in PresetTarget.all_legacy_aliases())
```
StafaH, line 405: Is the sort here important?

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
```diff
@@ -0,0 +1,333 @@
...
+        if token == "--task" and i + 1 < len(sys.argv):
+            last_task = sys.argv[i + 1]
+        elif token.startswith("--task="):
+            last_task = token[len("--task=") :]
+    return last_task
+
+
+def _help_requested() -> bool:
+    """Return True if ``--help`` or ``-h`` appears in ``sys.argv`` (excluding ``sys.argv[0]``)."""
+    return any(token in ("--help", "-h") for token in sys.argv[1:])
```
ooctipus: short helper function like this could hide the performance cost when repeated used(in this case its only used once), but my my recommendation is just inline it.

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -83,7 +78,7 @@ def _normalize_preset_name(name: str, known_names: set[str]) -> str:
         * ``name`` is itself a real field in ``known_names`` (a user-defined preset
           legitimately reusing the deprecated spelling shadows the alias).
     """
-    replacement = _LEGACY_PRESET_ALIASES.get(name)
+    replacement = PresetTarget.all_legacy_aliases().get(name)
```
StafaH, line 81: Can we call this once and build the dict and keep it static?

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
ooctipus: I'd consider one more time _enumerate_variants, _help_requested, _peek_task needs to be helper function, or just inline them to make operations transparent

StafaH (approved):
Left some minor comments

kellyguo11:
@fatimaanes for viz

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -0,0 +1,107 @@
...
+class PresetTarget(enum.Enum):
+    """Typed-flag preset categories.
+
+    Adding a new target = appending one enum member.
+    """
+
+    # Members. Tuple values are (label, base_classes, legacy_aliases); the
+    # enum metaclass collects the whole namespace before constructing members,
+    # so ``__new__`` below unpacks each tuple regardless of declaration order.
+    PHYSICS = ("physics", (PhysicsCfg,), {"newton": "newton_mjwarp", "kamino": "newton_kamino"})
```
ooctipus, line 58: importing 
```
from isaaclab.renderers.renderer_cfg import RendererCfg
```

what is the reason this import needs to happen? If a user write their own physics engine thats not subclass of PhysicsCfg, say class RandomPhysics? will it be will --physics=random_physics be disallowed? If it is allowed and any Physics engine or non physics can be allowed, then we don't need to import.

scripts/benchmarks/benchmark_non_rl.py
ooctipus: nope I think we can drop enable_cameras

ooctipus:
@hujc7 
Maybe since you are writing the preset now, and there is possibility that you will take this effort from me. There design thoughts about preset that would be good to know:

1. Cfgclass walk should ideally only happen once, all the decision on resolution, compatibility alias should Ideally not introducing additional walk.

2. This is a feature I haven't implemented, current way is to construct the full configclass tree first, and discarding nodes in the walk. But a more ideal way is to read the presets first, and avoid building full configclass in the first place. This could be done in multiple way, for example, when building configtree and hit a config class, stop walking further until resolution. and alternative is that you resolve while building.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
hujc7: It's hard to check from the name so we need to check from the type, which is also more for at discovery time that --help could print available values for the predefined preset targets (e.g. physics). If not inheriting PhysicsCfg, I believe it should still be allowed.

scripts/reinforcement_learning/rsl_rl/train.py
hujc7: reverted

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
hujc7: addressed

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
hujc7: addressed

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
ooctipus: i see make sense then!

ooctipus (approved):

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
hujc7: I think kind of yes to show persistent printing

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -0,0 +1,128 @@
...
+    exist because Newton-backend solver presets were renamed to use the
+    ``newton_`` prefix so they group together in autocomplete and read
+    distinctly from backend / package / visualizer names that also contain the
+    word ``newton``. Hydra's resolver (see
+    :func:`~isaaclab_tasks.utils.hydra._normalize_preset_name`) consults these
+    and emits a :class:`FutureWarning`; the aliases will be removed in a
+    future release.
+    """
+
+    RENDERER = ("renderer", (RendererCfg,))
```
YizeWang, line 71: Any options for "renderer"?

hujc7: Could you elaborate the question? Not too sure what it means

YizeWang: Never mind. I was mistaken. Sorry for the inconvenience.

PR #5591, opened 2026-05-12
Enable OVRTX cloning by default

huidongc:
# Description

* Added :func:`~isaaclab.cloner.cloner_utils.is_homogeneous` to detect whether a :class:`~isaaclab.cloner.ClonePlan`
  assigns every environment from every source (a homogeneous clone mask).

* Fixed cloned environments disappearing from tiled camera output if
  :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_ovrtx_cloning` is set to ``True``,
  by correcting scene-partition attribute creation on env roots and cameras.

* Renamed the ``use_cloning`` field on :class:`~isaaclab_ov.renderers.OVRTXRendererCfg` to ``use_ovrtx_cloning``.
  Changed its default value to ``True``. This will bring notable speedup for the total startup time (Launch to Train),
  esp. for large-scale env setups. On Isaac-Dexsuite-Kuka-Allegro-Lift-v0 with 1024 env clones, the total startup time
  dropped from ~78s to ~43s. Note that if ``use_ovrtx_cloning`` is enabled but the env setup is heterogeneous, the
  OVRTX renderer will disable the internal cloning path and logs a warning, exporting the full multi-environment stage
  instead (same effect as setting ``use_ovrtx_cloning`` to ``False`` for that run).

## Type of change

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
+++ b/source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
@@ -2,13 +2,15 @@ Fixed
 ^^^^^
 
 * Fixed cloned environments disappearing from tiled camera output if
-  :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_cloning` is set to ``True``,
+  :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_ovrtx_cloning` is set to ``True``,
   by correcting scene-partition attribute creation on env roots and cameras.
 
 Changed
 ^^^^^^^
 
-* Changed the default of :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_cloning` to ``True``. This will bring
-  notable speedup for the total startup time (Launch to Train), esp. for large-scale env setups. On
-  Isaac-Dexsuite-Kuka-Allegro-Lift-v0 with 1024 env clones, the total startup time (Launch to Train) dropped from
-  ~78s to ~43s.
+* Renamed the ``use_cloning`` field on :class:`~isaaclab_ov.renderers.OVRTXRendererCfg` to ``use_ovrtx_cloning``.
+  Changed its default value to ``True``. This will bring notable speedup for the total startup time (Launch to Train),
+  esp. for large-scale env setups. On Isaac-Dexsuite-Kuka-Allegro-Lift-v0 with 1024 env clones, the total startup time
+  dropped from ~78s to ~43s. Note that if ``use_ovrtx_cloning`` is enabled but the env setup is heterogeneous, the
+  OVRTX renderer will disable the internal cloning path and logs a warning, exporting the full multi-environment stage
+  instead (same effect as setting ``use_ovrtx_cloning`` to ``False`` for that run).
--- a/source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
+++ b/source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
@@ -8,4 +8,7 @@ Fixed
 Changed
 ^^^^^^^
 
-* Changed the default of :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_cloning` to ``True``.
+* Changed the default of :attr:`~isaaclab_ov.renderers.OVRTXRendererCfg.use_cloning` to ``True``. This will bring
+  notable speedup for the total startup time (Launch to Train), esp. for large-scale env setups. On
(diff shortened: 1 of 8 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -162,43 +162,37 @@ def inject_cameras_into_usd(
...
+        env_prim.CreateAttribute("primvars:omni:scenePartition", Sdf.ValueTypeNames.Token).Set(scene_partition)
+        logger.debug("Set scene partition '%s' on env root '%s'", scene_partition, env_prim.GetPath())
+
         for prim in Usd.PrimRange(env_prim):
-            if prim.GetPath() == env_prim.GetPath():
-                continue
             if prim.IsA(UsdGeom.Camera):
-                prim.CreateAttribute("omni:scenePartition", Sdf.ValueTypeNames.Token).Set(partition_name)
-            else:
-                prim.CreateAttribute("primvars:omni:scenePartition", Sdf.ValueTypeNames.Token).Set(partition_name)
```
huidongc, line 199: We shouldn't set the scene partition on the child prims. This causes bug because we don't write scene partition  attributes on child prims after cloning - in `_update_scene_partitions_after_clone`. OVRTX can handle hierarchical inheritance once the env roots are set with proper scene partition value.

source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
```diff
@@ -0,0 +1,14 @@
+Fixed
```
ncournia-nv, line 1: Did you intend to commit this file?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
ncournia-nv: I recall we had to set `primvars:omni:scenePartition` on child prims because the top-level env `primvar` wasn't properly propagated.  Was this fixed in FSD?

source/isaaclab_ov/changelog.d/huidongc-ovrtx-cloning.rst
pbarejko: Yes this is a new way of committing to the change log without conflicts.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
pbarejko: I don't know if this was fixed in 0.3, but not in 0.2. The pipeline will probably fail because it's running on 0.2.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -34,7 +34,7 @@ class OVRTXRendererCfg(RendererCfg):
     temp_usd_suffix: str = ".usda"
     """File suffix for temporary combined USD files (e.g. '.usda' or '.usdc')."""
 
-    use_cloning: bool = False
+    use_cloning: bool = True
```
pbarejko: Would this work for heterogeneous cloning?  Or the code would take `env_0` and clone it n-times? If this wouldn't work for heterogeneous cloning, we need to put this as known limitation.

zoctipus (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
huidongc: Well, CI captures a bug when using `ovrtx 0.2.0.280040`, it can be reproduced it locally. However, it works very well with `ovrtx 0.3.0`.

huidongc: now this creates a problem for us:

To make ovrtx cloning work for ovrtx 0.3.0, we need to remove the scene partition attribute writes, however this change will break ovrtx 0.2.0. I will have to make some branching based on the version here...

I think we will soon need to discuss whether we want to remove the code paths for OVRTX 0.2.0.

huidongc: We can keep the OVRTX 0.2.0 code path for debugging and regression tracking. Once OVRTX 0.3.0 is validated by QA, we should discuss whether we should remove 0.2 code path.

pbarejko (approved):

PR #5443, opened 2026-04-29
Add experimental Newton deformable support with rigid-deformable coupling

mmichelis:
# Description

Introduces experimental Newton-backend deformable simulation alongside the existing PhysX deformables, refactors deformable APIs to be backend-neutral, and adds the manager-based Franka soft-body lifting environment used to validate two-way coupling.

### Backend-neutral deformable APIs (`isaaclab`)
- Moved deformable schemas, materials, spawners, and `DeformableObjectCfg` out of `isaaclab_physx` into `isaaclab.sim` / `isaaclab.assets` so user code imports a single backend-agnostic API.
- **Breaking (PhysX):** `isaaclab_physx.sim.DeformableBodyPropertiesCfg`, `DeformableObjectSpawnerCfg`, `DeformableBodyMaterialCfg`, `SurfaceDeformableBodyMaterialCfg`, `spawn_deformable_body_material`, `define/modify_deformable_body_properties` and `isaaclab_physx.assets.DeformableObjectCfg` are now imported from `isaaclab.sim` / `isaaclab.assets`. See `docs/source/migration/migrating_deformables.rst`.
- USD spawners now support assets that embed tetrahedral mesh data, with automatic surface-mesh extraction.
- Added `pytetwild` dependency for tet-mesh generation.

### Newton manager abstraction (`isaaclab_newton`)
- Added deformable registration hooks to Newton cloning + Fabric sync fixes for particle meshes and particle-only scenes.

### Experimental Newton deformables (`isaaclab_contrib.deformable`)
- `DeformableObject` (Newton backend), `VBDSolverCfg`, and coupled solver configs `CoupledMJWarpVBDSolverCfg` / `CoupledFeatherstoneVBDSolverCfg` providing one- and two-way rigid–deformable coupling.
- New docs: `docs/source/features/newton-physics-integration/using-vbd-solver.rst` (VBD solver tuning + Franka soft-body lift guidance).

### Tasks, demos, tutorials
- Added `Isaac-Lift-Soft-Franka-v0` (manager-based), the documented rigid–deformable coupling task.
- Updated `scripts/demos/deformables.py`, the `01_assets/run_deformable_object` tutorial, and `00_sim/spawn_prims` to use the backend-neutral API with selectable PhysX or Newton backends.
- Added IK + video-record state-machine demo `scripts/environments/state_machine/lift_franka_soft.py`.

Continues on #5439 adding VBD solver and coupled MJWarp + VBD solver as newton managers in isaaclab_contrib.

[...]

PR #5641, opened 2026-05-15
Adds cable (1D rod) assets via Newton

mmichelis:
# Description

  Adds end-to-end support for cable / rod assets driven by Newton, layered on top  of the existing deformable contrib pattern.

  - **`isaaclab.sim.spawners.shapes`**: new `CableCfg` / `spawn_cable` that  author cables as `UsdGeomBasisCurves` prims. Physics is materialized by the Newton replicate hook, not at spawn time.
  - **`isaaclab_newton`**: new `NewtonCableMaterialCfg` (stretch / bend  stiffness, damping, density) and a per-cable ignore-paths block in   `newton_physics_replicate` so `add_usd` skips cable `BasisCurves`.
  - **`isaaclab_contrib.cable`**: 
    - new `CableObject` (subclass of the Newton  `Articulation`) + `CableObjectCfg`, plus the replicate-hook plumbing  (`CableRegistryEntry`, `add_cable_entry_to_builder`, `add_registered_cables_to_builder`, `install_cable_builder_hooks`)  mirroring   the deformable contrib layout.
    - **`CableObject.reset(env_ids=…)`** — new public method that snaps the  cable's per-env body slice back to its spawn pose. Restores `state.body_q` and `solver.body_q_prev` from `model.body_q`, and zeros  `state.body_qd` and `solver.body_inertia_q`. Resetting `body_q` alone  leaves stale `solver.body_q_prev`, which feeds AVBD's implicit velocity  `(body_q − body_q_prev)/dt` and produced ~700 m/s spurious velocities on the next step.
    - **USD-loadable cables** — `spawn_cable` now authors an `int2[] connections` attribute on the `UsdGeomBasisCurves` prim, and `CableObject._register_cable` reads it to recover the edge topology. Cables can now be loaded from arbitrary curve USDs (e.g. authored in Blender) via `UsdFileCfg`, as long as the curve carries this attribute.
  - **VBD manager**:
    - Overrides `forward()` to mask cable articulations out of `eval_fk`,   which has no `JointType.CABLE` case and was collapsing rod segments  onto   their parent anchors on every Kit pre-render FK pass (cable explosion fix). [Open Bug in Newton 2887](https://github.com/newton-physics/newton/issues/2887).
  - **Kit / Fabric sync**: cable `UsdGeomBasisCurves` points are refreshed   from Newton cable body transforms at render cadence (CPU only for now).
  - **Demo + tests**: `scripts/demos/cables.py`, `test_spawn_cable.py`,  `test_cable_object.py`.

  Builds on top of `mym/deformable_experimental` (#5443 ) because cables build on the  in-progress Newton deformable contrib stack.

  ## Test plan

  - [ ] `./isaaclab.sh -p -m pytest  source/isaaclab/test/sim/test_spawn_cable.py`
  - [ ] `./isaaclab.sh -p -m pytest source/isaaclab_contrib/test/cable/test_cable_object.py`
  - [ ] `./isaaclab.sh -p scripts/demos/cables.py` — visual sanity check in Kit
  - [ ] `./isaaclab.sh -p scripts/demos/deformables.py` — confirm no regression in mixed scenes
  - [ ] `./isaaclab.sh -f`

## Type of change

## Checklist


docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
+.. _newton-using-cables:
+
+Using Cables
+============
+
+Isaac Lab exposes 1D rod / cable assets through Newton's
+:meth:`newton.ModelBuilder.add_rod_graph`. A cable is spawned as a
+``UsdGeomBasisCurves`` prim, and the cable's physics (per-segment capsules,
+inter-segment cable joints, stretch / bend stiffness, damping, density) is
+materialized at Newton model-build time by a contrib replicate hook.
```
AntoineRichard: This is a very technical phrasing that doesn't add much. It might be better to rephrase that into regular human lingo.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+active development.
+
+.. note::
+    Cables are currently **only supported on the Newton physics backend**.
+    The spawner authors valid USD on any backend (so the scene loads in PhysX
+    or PhysX-Fabric viewports), but the resulting cable is not registered with
+    a PhysX articulation. :class:`~isaaclab.sim.spawners.shapes.CableCfg`
+    requires ``physics_material`` to be a
+    :class:`~isaaclab_newton.sim.spawners.materials.NewtonCableMaterialCfg`
+    and rejects ``rigid_props`` / ``mass_props`` up front.
```
AntoineRichard, line 22: Should that be the case, I feel like we should raise an error immediately if a cable is defined in PhysX to not deceive the users.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+spawner, the cable replicate hook, the VBD solver, and the Kit / Fabric
+viewport sync are all working in your environment:
+
+.. code-block:: bash
+
+    ./isaaclab.sh -p scripts/demos/cables.py
+    ./isaaclab.sh -p scripts/demos/cables.py --num_cables 40
+
+The demo spawns a pile of randomly oriented cables onto a ground plane under
+the Newton VBD solver. Source: ``scripts/demos/cables.py``.
```
AntoineRichard, line 38: Maybe add something around the lines of: "Since this feature is currently experimental, a good sanity check is to ...." Otherwise it sounds like we don't know what we are doing.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+
+
+Authoring a Cable
+-----------------
+
+A cable is configured as a :class:`~isaaclab.sim.spawners.shapes.CableCfg`
+plus a Newton-specific physics material. The cfg's ``positions`` field is a
+list of at least two control points in the cable's local frame; adjacent pairs
+become individual rod segments, each materialized as a capsule body of
+diameter ``width`` and joined to its neighbour by a Newton cable joint.
```
AntoineRichard: I feel like this could be phrased better for people to understand.

AntoineRichard: Why not add a table like done elsewhere? That might be more informative to give precise information to the users while allowing for a more technically relaxed description.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+        visual_material=sim_utils.PreviewSurfaceCfg(diffuse_color=(0.7, 0.2, 0.2)),
+        physics_material=NewtonCableMaterialCfg(
+            stretch_stiffness=1.0e6,
+            bend_stiffness=1.0e-4,
+            stretch_damping=1.0e-4,
+            bend_damping=1.0e-4,
+            density=1000.0,
+        ),
+        collision_props=sim_utils.CollisionPropertiesCfg(),
+    )
```
AntoineRichard, line 106: Could this code block be documented? Create a cable with 20 elements all spaced by 10cm....

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+.. code-block:: python
+
+    import isaaclab.sim as sim_utils
+    from isaaclab_newton.sim.spawners.materials import NewtonCableMaterialCfg
+
+    cable_spawn = sim_utils.CableCfg(
+        positions=[(i * 0.1, 0.0, 0.0) for i in range(10)],
+        width=0.03,
+        visual_material=sim_utils.PreviewSurfaceCfg(diffuse_color=(0.7, 0.2, 0.2)),
+        physics_material=NewtonCableMaterialCfg(
```
AntoineRichard, line 98: Are we keeping VBD specific materials for later? That would help with picking a solver?

AntoineRichard: I recall discussing this with you @mmichelis?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+        cfg=CableObjectCfg(
+            prim_path="/World/Origin/Cable",
+            spawn=cable_spawn,
+            init_state=CableObjectCfg.InitialStateCfg(pos=(0.0, 0.0, 0.5)),
+        )
+    )
+
+The :class:`~isaaclab_contrib.cable.CableObject` constructor appends a
+:class:`~isaaclab_contrib.cable.CableRegistryEntry` to the contrib cable
+registry. The Newton VBD manager installs a per-world builder hook that walks
```
AntoineRichard, line 135: Are we limited to VBD?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+    physics_cfg.model_cfg = NewtonModelCfg(
+        shape_material_ke=1.0e3,
+        shape_material_kd=1.0e1,
+        shape_material_mu=1.0,
+    )
+
+A cable-only scene can use a bare
+:class:`~isaaclab_contrib.deformable.VBDSolverCfg`. Mixed rigid + cable scenes
+(robot manipulating a cable) should use a coupled solver — see
+:doc:`using-vbd-solver`.
```
AntoineRichard, line 173: This should state if VBD is the only supported solver. It should also state if we explicitely error out if set with a non-compatible solver?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+        :meth:`newton.ModelBuilder.add_rod_graph`.
+
+
+Spawner Parameters
+------------------
+
+:class:`~isaaclab.sim.spawners.shapes.CableCfg` fields specific to cables
+(beyond the inherited :class:`~isaaclab.sim.spawners.shapes.ShapeCfg` slots):
+
+.. list-table::
```
AntoineRichard: Ah ok list is here. Can we make sure there is a reference to this so users don't have to scroll all the way down to figure it out?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+each environment's cable bodies back to the spawn pose stored in
+``newton.Model.body_q`` and zeroes both ``state.body_qd`` and the AVBD
+``solver.body_inertia_q`` buffer. The implicit-velocity buffer
+``solver.body_q_prev`` is also restored to the rest pose — without this,
+AVBD's ``(body_q - body_q_prev) / dt`` velocity estimate would emit ~700 m/s
+spurious velocities the step after a snap-back. Joint state and AVBD
+penalty / Dahl buffers are intentionally left alone: they are either global
+to the world or would require joint offsets in the registry to slice
+per-env, and the body-side reset is sufficient to keep post-reset dynamics
+bounded in practice.
```
AntoineRichard: I feel like this is not done in a user friendly way. It starts with the solution and then goes back to what it does. Can we start by stating that it's supported with X and Y and then go into the actual solution. In general this doc is very solution focused when I think it should be more user centric, introduce the concept first before firing a bunch of technical words at the users.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+    cable = CableObject(
+        cfg=CableObjectCfg(
+            prim_path="/World/Origin/Cable",
+            spawn=sim_utils.UsdFileCfg(
+                usd_path="path/to/cable.usda",
+                physics_material=NewtonCableMaterialCfg(density=100.0),
+            ),
+            init_state=CableObjectCfg.InitialStateCfg(pos=(0.0, 0.0, 0.5)),
+        )
+    )
```
AntoineRichard, line 245: Again you give a snippet, but it's never said what this snippet does. It's never referenced.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+The Newton cable material is taken from the spawner's ``physics_material``
+binding on the curve prim. If no Newton cable material is bound, the
+:class:`~isaaclab_contrib.cable.CableRegistryEntry` defaults are used.
+
+
+Limitations
+-----------
+
+* Newton-only. PhysX has no cable joint, so the cable will load as inert
+  geometry under a PhysX backend.
```
AntoineRichard: It should error out. It should also error out on incorrect solvers. This (the solver is also a limit that should be mentioned)

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+* No actuators. :class:`~isaaclab_contrib.cable.CableObjectCfg` overrides
+  ``actuators`` to ``{}``; per-cable stiffness is treated as material, not as
+  a controllable joint. The inherited
+  ``logger.warning("Not all actuators are configured!")`` is expected and
+  harmless.
```
AntoineRichard, line 292: Well no kiding, why would we need actuators on a cable? How is that relevant? Is it because under the hood it's a hidden articulation?

AntoineRichard: Or is it a selection API limitation?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+* :meth:`newton.eval_fk` has no
+  :attr:`newton.JointType.CABLE` case at present. The VBD manager
+  works around this by building a non-cable articulation mask in
+  :meth:`~isaaclab_contrib.deformable.vbd_manager.NewtonVBDManager._build_non_cable_articulation_mask`
+  and overriding
+  :meth:`~isaaclab_contrib.deformable.vbd_manager.NewtonVBDManager.forward`
+  so Kit-triggered pre-render FK passes don't collapse rod segments onto their
+  parent anchors. Once Newton patches cable joints in ``eval_fk``, that mask
+  and override can be removed.
```
AntoineRichard, line 301: That sounds nasty. Is this know to Newton? We would need a fix ASAP. @mmichelis / claude /codex can you compile a list of tasks we'd need from newton (Claude / codex if you do it, first make the list and then check these are not already open issues). They are trying to wrap-up the 1.3 roadmap this week. If we miss the window we may need to wait for 1.4.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,306 @@
...
+* Self-contact between cable segments uses the rigid contact pipeline
+  (``shape_material_ke`` / ``kd`` / ``mu`` on
+  :class:`~isaaclab_contrib.deformable.NewtonModelCfg`), not VBD particle
+  self-contact. For dense cable piles, lower ``shape_material_ke``, raise
+  ``shape_material_kd``, and increase
```
AntoineRichard: Same here we'd need support from Newton to handle collision parsing properly. MJWarp has automatic parent / child filtering by default. Could we request that of VBD as well.

scripts/demos/cables.py
```diff
@@ -0,0 +1,159 @@
...
+
+"""Launch Isaac Sim Simulator first."""
+
+
+import argparse
+
+from isaaclab.app import AppLauncher
+
+parser = argparse.ArgumentParser(description="Spawn a pile of cables at varied z-axis rotations.")
+parser.add_argument("--num_cables", type=int, default=25, help="Number of cables to spawn.")
```
AntoineRichard, line 24: Can we add the number of segment in the cables as a parameter? :D

scripts/demos/cables.py
```diff
@@ -0,0 +1,159 @@
...
+from isaaclab_visualizers.newton.newton_visualizer_cfg import NewtonVisualizerCfg
+
+import isaaclab.sim as sim_utils
+
+from isaaclab_contrib.cable import CableObject, CableObjectCfg
+
+
+def z_axis_quat(angle_rad: float) -> tuple[float, float, float, float]:
+    """Quaternion (x, y, z, w) for a rotation of ``angle_rad`` about +Z."""
+    return (0.0, 0.0, math.sin(0.5 * angle_rad), math.cos(0.5 * angle_rad))
```
AntoineRichard: Don't we have a math function for that?

AntoineRichard: Please use our base of functions when possible.

scripts/demos/cables.py
```diff
@@ -0,0 +1,159 @@
...
+        collision_cfg=NewtonCollisionPipelineCfg(rigid_contact_max=65536),
+    )
+    # Soften body-body contact: lower ke + nonzero kd damps out the
+    # spikes when many cable segments pile onto one segment. mu=1.0 keeps
+    # cables from sliding off the pile.
+    physics_cfg.model_cfg = NewtonModelCfg(
+        shape_material_ke=1.0e3,
+        shape_material_kd=1.0e0,
+        shape_material_mu=1.0,
+    )
```
AntoineRichard, line 141: Do you think this could be randomized per cable post init? This way the demo would biundle as a randomization demo too?

scripts/demos/deformables.py
AntoineRichard, line 1: This is adding a noodle to the demo with all the other deformable bodies? The if statement if nice, but it makes it non-homogeneous between PhysX and Newton.  Can the demo description explicitely state this?

source/isaaclab/isaaclab/sim/spawners/shapes/shapes.py
```diff
@@ -233,6 +233,79 @@ def spawn_cone(
...
+    helper. Cable physics is materialized later by the Newton replicate hook
+    calling :meth:`newton.ModelBuilder.add_rod_graph`; ``rigid_props`` and
+    ``mass_props`` are rejected up front because they don't apply to cables.
+
+    .. warning::
+        Cables are currently **only supported on the Newton physics backend**. The
+        spawner itself only authors USD (which works on any backend), but the
+        resulting cable will not be simulated under PhysX —
+        :class:`~isaaclab_contrib.cable.CableObject` will refuse to register if
+        the active backend is not Newton.
```
AntoineRichard: This is fine, but I think we should also raise when using the wrong backend. Also make sure to specify the solver?

AntoineRichard: (VBD, it's the only that supports it too?)

source/isaaclab/isaaclab/sim/spawners/shapes/shapes.py
```diff
@@ -233,6 +233,79 @@ def spawn_cone(
...
+    orientation: tuple[float, float, float, float] | None = None,
+    **kwargs,
+) -> Usd.Prim:
+    """Create a ``UsdGeomBasisCurves`` cable prim.
+
+    Authors a USD curve under ``{prim_path}/geometry/mesh`` (same layout as the
+    other shape spawners) and binds the visual / physics material via the shared
+    helper. Cable physics is materialized later by the Newton replicate hook
+    calling :meth:`newton.ModelBuilder.add_rod_graph`; ``rigid_props`` and
+    ``mass_props`` are rejected up front because they don't apply to cables.
```
AntoineRichard, line 251: That's interesting. Why can't we set the mass?

source/isaaclab/isaaclab/sim/spawners/shapes/shapes.py
AntoineRichard: Not sure if we should raise here, but we should definitely raise somewhere.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
```diff
@@ -120,3 +120,47 @@ class ConeCfg(ShapeCfg):
...
+    :class:`UsdPhysics.CollisionAPI` to author a usable physics-material binding;
+    the cable geometry is ignored by the Newton USD importer so this has no
+    PhysX runtime effect.
+    """
+
+    func: Callable | str = "{DIR}.shapes:spawn_cable"
+
+    positions: list[tuple[float, float, float]] = MISSING
+    """Control points in cable-local frame [m]. Must contain at least 2 points.
+    Adjacent pairs define one cable segment each."""
```
AntoineRichard, line 168: Is it a "control point" (xyz, or x, + control of the derivative like a bezier curve would?)

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
```diff
@@ -120,3 +120,47 @@ class ConeCfg(ShapeCfg):
     """Height of the v (in m)."""
     axis: Literal["X", "Y", "Z"] = "Z"
     """Axis of the cone. Defaults to "Z"."""
+
+
+@configclass
+class CableCfg(ShapeCfg):
+    """Configuration parameters for a 1D cable / rod prim.
```
AntoineRichard, line 136: Do 2D cables exist?

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
```diff
@@ -120,3 +120,47 @@ class ConeCfg(ShapeCfg):
...
+        Cables are currently **only supported on the Newton physics backend**.
+        ``physics_material`` must be a
+        :class:`~isaaclab_newton.sim.spawners.materials.NewtonCableMaterialCfg`
+        (``ValueError`` otherwise), and any :class:`~isaaclab_contrib.cable.CableObject`
+        built from this cfg only registers in Newton's solver pipeline.
+
+    The cable's stretch/bend stiffness, damping, and density live on
+    ``physics_material`` (inherited slot from :class:`ShapeCfg`). ``rigid_props``
+    and ``mass_props`` are inherited from :class:`ShapeCfg` but are not used by
+    cables — :func:`spawn_cable` raises ``ValueError`` if either is non-None.
```
AntoineRichard, line 157: Ha so the mass is defined here. It's a bit messy. Why can't the mass be defined normally? Would we warn users that if they define the mass the normal way it will be ignored? I think it's really important to get these configs right otherwise it's going to be misleading for users, and without clear signal that something is wrong they may miss it completely. We can't expect them to read the fineprints in the doc.

source/isaaclab/test/sim/test_spawn_cable.py
AntoineRichard, line 1: Can we add some tests here to validate that we raise/warn when using the wrong backend, the wrong solver, the mass is defined incorrectly.

source/isaaclab_contrib/changelog.d/mym-cable.major.rst
AntoineRichard, line 1: I would argue this is a major.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Configuration for the cable asset class."""
+
+from __future__ import annotations
+
+from isaaclab.actuators import ActuatorBaseCfg
```
AntoineRichard, line 10: Why is that imported here?

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
```diff
@@ -0,0 +1,37 @@
...
+    - ``actuators = {}`` — cables have no user-defined actuators (cable joint
+      stiffness is material-like, applied internally by the solver). The
+      inherited ``_process_actuators_cfg`` iterates an empty dict safely and
+      emits a harmless ``logger.warning("Not all actuators are configured!")``
+      — expected and not suppressed in Phase 1.
+    """
+
+    class_type: type = "{DIR}.cable_object:CableObject"
+    articulation_root_prim_path: str | None = "/cable_articulation"
+    actuators: dict[str, ActuatorBaseCfg] = {}
```
AntoineRichard, line 36: Do we need this here? Super confusing.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
```diff
@@ -0,0 +1,37 @@
...
+      that :meth:`newton.ModelBuilder.add_rod_graph` produces under the cable's
+      source prim path (``f"{label}_articulation"`` where ``label`` is
+      ``"{prim_path}/cable"``). The base method composes this with
+      ``cfg.prim_path`` and uses the result as the label pattern for
+      :class:`newton.selection.ArticulationView`.
+    - ``actuators = {}`` — cables have no user-defined actuators (cable joint
+      stiffness is material-like, applied internally by the solver). The
+      inherited ``_process_actuators_cfg`` iterates an empty dict safely and
+      emits a harmless ``logger.warning("Not all actuators are configured!")``
+      — expected and not suppressed in Phase 1.
```
AntoineRichard: What's Phase 1? Claude/Codex can you not leave implementation traces in docs.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+
+The structure mirrors :mod:`isaaclab_contrib.deformable.deformable_object`. Cables
+differ from deformables in two respects only:
+
+1. They subclass :class:`Articulation` (not :class:`BaseDeformableObject`) because
+   ``newton.ModelBuilder.add_rod_graph`` produces a Newton articulation, and
+   ``ArticulationView`` already covers state read/write.
+2. Their material is consumed in-memory by the cable replicate hook (no USD
+   read-back), since :class:`CableObject` always holds the source cfg.
+"""
```
AntoineRichard, line 16: Yeah sure, but it also comes with a massive amount of useless methods that will just confuse people, subclassing a huge file just to benefit from 2 methods is insane.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+    last_edge_length: float = 0.0
+
+
+from isaaclab_newton.assets.articulation.articulation import Articulation  # noqa: E402
+from isaaclab_newton.physics import NewtonManager as SimulationManager  # noqa: E402
+
+import isaaclab.sim as sim_utils  # noqa: E402
+
+if TYPE_CHECKING:
+    from .cable_object_cfg import CableObjectCfg
```
AntoineRichard: Why delay these imports after the registry definition?

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def add_cable_entry_to_builder(
+    builder,
+    entry: CableRegistryEntry,
+    env_idx: int,
+    env_position: list[float],
+    env_rotation: list[float] | tuple[float, float, float, float],
+    cable_idx: int = 0,
+) -> None:
```
AntoineRichard, line 72: Why is this here? This has nothing to do with GPU accelerated asset settings, can't that be in the schema? This file is mirroring what an assets/cable would be, it is meant for setters not spawning / replicating logic.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+        2. Walk the template prim's descendants and find the single
+           ``UsdGeomBasisCurves`` prim, then read its ``points`` and ``widths``
+           attributes. This works for both :func:`spawn_cable` (which authors
+           the curve at ``{prim_path}/geometry/mesh``) and arbitrary curve
+           USDs loaded via :class:`~isaaclab.sim.spawners.UsdFileCfg`.
+        3. Bake the template prim's xform into the per-node positions so the
+           replicate hook only needs to apply the env transform.
+        4. Look up the bound Newton cable physics material on the curve prim
+           and read each ``newton:*`` attribute into the entry. If no Newton
+           material is bound, fall back to :class:`CableRegistryEntry`
```
AntoineRichard, line 237: Outch, Can't we do this cleanly? Is that required because Newton can't parse stuff? It's really ugly. Can this not be in the cable accessor code? This is again just setting stuff at the USD level. Wouldn't that be the schema's responsability?

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+            edges=edges,
+            radius=radius,
+            stretch_stiffness=float(stretch_stiffness),
+            bend_stiffness=float(bend_stiffness),
+            stretch_damping=float(stretch_damping),
+            bend_damping=float(bend_damping),
+            density=float(density),
+        )
+        SimulationManager._cable_registry.append(entry)
+        return entry
```
AntoineRichard, line 405: I feel like the schema could do this as well.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
```diff
@@ -0,0 +1,439 @@
...
+        # Per-call zero buffer for velocity slices (one segment chain wide).
+        zero_qd = wp.zeros(n, dtype=state.body_qd.dtype, device=state.body_qd.device)
+        zero_q = wp.zeros(n, dtype=solver.body_inertia_q.dtype, device=solver.body_inertia_q.device)
+        env_iter = range(len(body_offsets)) if env_ids is None or env_ids == slice(None) else list(env_ids)
+        for env_idx in env_iter:
+            offset = int(body_offsets[env_idx])
+            wp.copy(dest=state.body_q, src=model.body_q, dest_offset=offset, src_offset=offset, count=n)
+            wp.copy(dest=solver.body_q_prev, src=model.body_q, dest_offset=offset, src_offset=offset, count=n)
+            wp.copy(dest=state.body_qd, src=zero_qd, dest_offset=offset, count=n)
+            wp.copy(dest=solver.body_inertia_q, src=zero_q, dest_offset=offset, count=n)
```
AntoineRichard, line 448: Is that VBD only? I didn't know we had body_q_prev by default? Again we really need to raise on solver :)

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
AntoineRichard, line 1: Do you have a data equivalent?

source/isaaclab_contrib/isaaclab_contrib/deformable/deformable_object.py
AntoineRichard, line 1: This looks aggressive and unrelated (color change) is there any reason why we are doing this?

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
```diff
@@ -31,6 +31,8 @@ class VBDSolverCfg(NewtonSolverCfg):
 
     solver_type: str = "vbd"
 
+    requires_graph_coloring: bool = True
+
```
AntoineRichard: THis is just dropped here? Can we have some descriptions? In the main newton extension, don't we split these configs per solver?

AntoineRichard: Does VBD lives here and in the newton extension? Do we have two? Or it's just here. Might be confusing...

source/isaaclab_contrib/test/cable/test_cable_object.py
```diff
@@ -0,0 +1,547 @@
...
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Launch Isaac Sim Simulator first."""
+
+from isaaclab.app import AppLauncher
+
+# launch omniverse app
+simulation_app = AppLauncher(headless=True).app
```
AntoineRichard: This is newton only, do we need to start kit?

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
AntoineRichard, line 1: Is this really parsed or is it for show only? If it's not parsed yet, we should advertize this clearly?

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -980,7 +980,7 @@ def start_simulation(cls) -> None:
         NewtonManager._state_0 = cls._model.state()
         NewtonManager._state_1 = cls._model.state()
         NewtonManager._control = cls._model.control()
-        eval_fk(cls._model, cls._state_0.joint_q, cls._state_0.joint_qd, cls._state_0, None)
+        cls.forward()
```
AntoineRichard: Who is defining that now? I don't see an explicit forward method, was it already defined? If so was it the only thing it did? If not should we be concerned about perf implications?

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1013,6 +1013,9 @@ def start_simulation(cls) -> None:
                 NewtonManager._usdrt_stage = get_current_stage(fabric=True)
                 for i, prim_path in enumerate(body_paths):
                     prim = cls._usdrt_stage.GetPrimAtPath(prim_path)
+                    # Cable segment bodies have no per-body USD prim.
+                    if not prim.IsValid():
+                        continue
                     prim.CreateAttribute(cls._newton_index_attr, usdrt.Sdf.ValueTypeNames.UInt, True)
```
AntoineRichard: Dear lord, we're parsing the stage here now :'(. @ooctipus we really need to do the cloner thing. We have artifacts everywhere :/.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
```diff
@@ -156,6 +207,193 @@ def start_simulation(cls) -> None:
...
+                NewtonManager._usdrt_stage = get_current_stage(fabric=True)
+
+            stage = get_current_stage()
+            curves_registered = False
+            for entry in cls._cable_registry:
+                curve_template_path = entry.curve_prim_path or f"{entry.prim_path}/geometry/mesh"
+                for inst_idx, body_offset in enumerate(entry.body_offsets):
+                    resolved = re.sub(r"(?<=[Ee]nv_)\.\*", str(inst_idx), curve_template_path)
+                    resolved = re.sub(r"\.\*", str(inst_idx), resolved)
+                    curve_prim = stage.GetPrimAtPath(resolved)
```
AntoineRichard, line 183: I'm wondering how much this kind of methods could be externalized to better isolate the code. The risk here is that this code might need to be duplicated to XPBD, or any other solver, and then that as we add more things (cloth etc...) it just ends up bloating the code base making the sinulation start completely unreadable.

AntoineRichard (changes requested):
Thanks Mike for the cool work! Couple changes are required, but I think it's a really good step in the right direction. @kellyguo11 if we need this now, this can be merged as is and all the comments could be addressed in a follow-up however some are quite important to get right before that.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: Yes I would keep this separate, since it would require overhauling all newton materials.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: yes. I've clarified it's vbd only

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: during the cable object initialization, we check if the cable registry exists, which only works for the vbd manager. The error thrown details why VBD solver must be used

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: I shortened it since I don't think the tehcnicalities are needed here

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: Indeed it is because it's an articulation, but isaaclab articulation throws a warning that the articulation does not have its actuators configured. Currently articulation is expected to always be active, and not passive. We could either have another flag for active/passive articulation in isaaclab, or we don't inherit cables from articulation at all. We could construct our own version of passive articulation if desired. What do you think?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: I've edited the comment, we currently disable cable self contact by adding the segments to the same collision group. We chatted about making it so direct neighboring segments don't collide, I can look into that for this PR as well if you want.

scripts/demos/cables.py
mmichelis: yessir

scripts/demos/cables.py
mmichelis: No currently the contact parameters are not set per shape yet, it is already on the list for todos in the future that we can do per shape contact parameter definitions. We'd introduce a new config field for deformables here. I would do this in a follow-up PR.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes.py
mmichelis: I've added a check here that the physics manager is newton backend. Later we check specifically for VBD as well and raise error otherwise.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes.py
mmichelis: The shape config for newton cables only sets density: https://github.com/newton-physics/newton/blob/main/newton/_src/sim/builder.py#L231.

The higher level reason is that mass properties are set for single rigid bodies, not chains of rigid bodies.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
mmichelis: I mean, I call them 1D since I want the 1D/2D/3D cable/cloth/deformable dimensionality idea, do you think it's confusing to call them 1D without the context of 2d/3d?

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
mmichelis: Currently it's a vertex that is interpolated linearly, since we choose the type as linear. UsdGeom BasisCurve supports cubic as type as well, then you can have bezier/bspline as basis as well, in which case the points are control points/tangent handles. This is not yet there in how we read basiscurves, I added this as a comment.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
mmichelis: Answered above why mass cannot be defined normally (only density supported on rigid body chain). We raise an error if mass props is defined, and the message should be clear to the end user. If we don't want mass_props at all, we would have to not inherit from shapecfg and as a result rigidobjectspawnercfg. We cannot inherit from deformableobjectspawnercfg either since it also defines mass props, so would a cable specific one make sense? I thought this would be too messy, and breaks too much consistency with existing shapes.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
mmichelis: Actuators have to be set for articulation. This could be a bigger discussion whether or not cables should be articulation, since it seems articulation is currently thought for only active articulation, not passive.

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
mmichelis: So the articulation root prim path is so we can use the selection api for articulation from newton, and we need to disable actuators (which by default is MISSING for articulation).

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object_cfg.py
mmichelis: oopsie, my bad

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
mmichelis: yes absolutely, it's the same problem as with deformables, import_usd from newton. Can be removed same as deformable registry once parsed. I just need to put together the required parsing fields from newton, I want to make sure we have everything in place for this

source/isaaclab_contrib/isaaclab_contrib/cable/cable_object.py
mmichelis: it's indeed VBD specific, and yes there is a raise now for solver

source/isaaclab_contrib/isaaclab_contrib/deformable/deformable_object.py
mmichelis: we unified this to vbd solver, so it always colors, since this is applicable to all VBD solver objects, not just deformables.

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
mmichelis: VBD solver is only in the contrib, we don't have a manager in isaaclab_newton for vbd yet cuz of the experimental nature. But yes I will add some description for this parameter. This newton manager cfg is indeed just adding together all the manager configs, I thought we could keep the contrib from being too crowded, once in isaaclab_newton we split this into separate managers, or should we split here already?

source/isaaclab_contrib/test/cable/test_cable_object.py
mmichelis: yes we can remove this here! Thanks for noting this

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
mmichelis: The fields are parsed — `CableObject._register_cable` reads them back from the bound material prim  via `_get_material_attr("newton:stretchStiffness", ...)` etc. (`cable_object.py:383-387`), with the cfg values as fallbacks. The spawn-side authoring is covered by  `test_spawn_cable_authors_newton_material_attrs` in `test_spawn_cable.py`.

  That said, you're right that the storage is provisional — there's no official UsdNewton applied  schema yet (`_usd_applied_schema = None`), so the values live as plain `newton:*` attributes on a  `UsdShade.Material` prim. The previous docstring was also misleading (it implied `CableObject`  reads from the cfg object directly). I've rewritten the class docstring to describe the actual cfg  → USD → read-back flow and added a `Note::` calling out that the namespace convention is  experimental and may change once a typed schema lands.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
mmichelis: this should be handled by the import_usd and selection api from newton side, same with deformables, a lot of things can be removed once this is done. The resolving of paths is still hardcoded in lab, not sure how we want to approach that, but the body_offset things can be fixed with selection api.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
mmichelis: newton manager already had a forward method, but for some reason a develop commit removed the call to the class forward and just calls the immediate eval_fk. We need VBD to overwrite the forward, so this is quite important to use the class forward instead of eval_fk. No perf implications, it's one extra Python frame around a Warp kernel launch, and it only fires once during `start_simulation`. The motivation was just to stop duplicating the unmasked-FK call between this  site and `forward()` itself.

PR #5643, opened 2026-05-15
[Docs] Expand Isaac Teleop XR performance optimization guide

rwiltz:
# Description

Expands the **Optimize XR Performance** section of the Isaac Teleop feature guide
with the most common levers users reach for when XR teleop cannot sustain the
headset's display rate -- particularly on lower-spec GPUs or in heavy scenes.
What changed:
- **RTX - Minimal renderer**: new dropdown explaining when to use it, how to
  enable it from the viewport renderer dropdown, the recommended **Render
  Settings** (**Minimal Shading Mode = Diffuse/Glossy/Emission**), and the
  current `DistantLight`-only lighting limitation, with a snippet showing how
  to swap a `DomeLight` for a `DistantLight`.
- **XR Resolution Multiplier slider**: new dropdown describing the
  **XR -> Advanced Settings -> Render Resolution** slider for trading image
  sharpness for GPU headroom (`0.8` as a sensible starting point).
- **Physics / render time step**: refreshed to focus on Quest 3 / Pico 4 Ultra
  (90 Hz), explain what `sim.render_interval` actually controls, and call out
  the `sim.dt` stability/performance trade-off.
- Removed the **Try running physics on CPU** dropdown -- this is already the
  default for these workflows.

Fixes # (issue)

## Type of change

- Documentation update

## Checklist


docs/source/features/isaac_teleop.rst
```diff
@@ -956,27 +956,95 @@ Optimize XR Performance
...
+   .. figure:: ../_static/teleop/xr-resolution-slider.png
+      :width: 80%
+      :alt: XR Render Resolution slider in the Advanced Settings panel
+
+      The **Resolution Multiplier** under **XR -> Advanced Settings -> Render Resolution**.
+      Values below ``1.0`` reduce the render-buffer size before upscaling to the headset.
+
+   A value around ``0.8`` is usually a good starting point: noticeable GPU savings with minimal
+   perceptible quality loss. Reduce further only if you still cannot hit the headset's display
+   rate.
```
hougantc-nvda, line 1047: Can we add section:

.. dropdown:: Check CloudXR frame pacing

   The CloudXR runtime frame pacer attempts to keep the client experience smooth. If the
   application has repeated frame-time spikes, the pacer may settle at a lower stable frame
   rate instead of oscillating between rates. This can make a connected client appear slower
   even when Isaac Lab profiling does not show a proportional simulation-side regression.

   When diagnosing this case, check the CloudXR runtime configuration and restart the runtime
   after changing settings such as ``NV_ENABLE_POSE_WAIT=false``. See
   :ref:`isaac-teleop-cloudxr-profiles` for details.

hougantc-nvda: One more section:

.. dropdown:: Configure retargeting execution

   Isaac Teleop can run retargeting either synchronously on the application thread or
   asynchronously through a pipelined worker. This is controlled by
   ``RetargetingExecutionConfig``.

   In synchronous mode, retargeting runs inline with the simulation step. This can be the
   best choice for lightweight retargeting or retargeting implemented mostly in Python,
   since a background Python worker can still contend with the application thread through
   the GIL.

   In pipelined mode, Isaac Teleop submits retargeting work to a background worker and the
   application uses the most recent completed result. This is useful when retargeting has
   enough native work to overlap with simulation or rendering, or when the retargeting cost
   is large enough that running it inline would directly extend the frame.

   .. code-block:: python

      retargeting_execution=RetargetingExecutionConfig(
          mode="pipelined",
          pacing=DeadlinePacingConfig(safety_margin_s=0.025),
      )

   ``DeadlinePacingConfig`` intentionally delays the background retargeting work until
   closer to when the next result is needed, instead of starting it immediately when the
   request is submitted. This helps avoid competing with the Python work Isaac Lab performs
   at the beginning of the frame, and tends to line the retargeting work up with rendering
   or other native work where overlap is more useful.

   The ``safety_margin_s`` value controls how early retargeting starts before the predicted
   deadline. A larger margin starts retargeting earlier, which gives heavier or more variable
   retargeting work more time to finish before the next frame consumes the result. The tradeoff
   is that the input sample may be slightly older, and Python-heavy retargeting may introduce
   more GIL contention.

   If retargeting is mostly Python and lightweight, consider ``mode="sync"``. If retargeting
   performs substantial native work or has occasional long spikes, use ``mode="pipelined"``
   and increase ``safety_margin_s`` so the work starts earlier.

hougantc-nvda (approved):
LGTM

rwiltz:
@greptile

PR #5651, opened 2026-05-16
Migration of camera.py to completely remove torch

StafaH:
# Description

Cleaned up camera internals and converted more camera state handling to Warp-based paths. This keeps camera pose, intrinsic, and frame updates closer to the Warp-backed data buffers used by the renderer and sensor views.

## Checklist


source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -185,7 +185,7 @@ def image_shape(self) -> tuple[int, int]:
     """
 
     def set_intrinsic_matrices(
-        self, matrices: torch.Tensor, focal_length: float | None = None, env_ids: Sequence[int] | None = None
```
ooctipus, line 188: is this type change ok? maybe just require warp? x)

StafaH: I wasn't sure if I should be using numpy array or warp array as expected input, what is the expected usage of this? Will the user have warp or numpy array when they try to set this?

kellyguo11: I think all of our APIs should just accept warp arrays for consistency. we don't use numpy for most APIs

source/isaaclab/isaaclab/utils/math.py
```diff
@@ -1611,6 +1658,54 @@ def convert_camera_frame_orientation_convention(
         return quat_gl.clone()
 
 
+def convert_camera_frame_orientation_convention_np(
```
kellyguo11: why do we need a numpy version of this?

kellyguo11 (commented):
could you also fill in some details in the PR description?

source/isaaclab/isaaclab/sensors/camera/camera.py
StafaH: Converted to accept torch (and also warp). Torch gets wrapped to warp. Below it gets converted to numpy either way though but API is consistent now

source/isaaclab/isaaclab/utils/math.py
StafaH: The codepath was using numpy, but for now I reverted it to just use the old conversion.

kellyguo11 (approved):

PR #5660, opened 2026-05-17
[CLI] Enable kitless URDF and MJCF conversion

kellyguo11:
## 1. Summary

- Enables URDF and MJCF command-line conversion without Isaac Sim when the standalone `isaacsim-asset-isolated` wheel is installed.
- Preserves the existing Isaac Sim extension-backed path when the full runtime is available.
- Limits the converter CLI to converter arguments plus `--viz {kit,none}`; the launcher is configured internally, so `AppLauncher` flags (which never affected conversion output) are no longer accepted on these scripts.
- Other tool scripts (`convert_mesh.py`, `convert_instanceable.py`, `check_instanceable.py`) keep develop's exact `add_app_launcher_args` behavior, including its `--help`-with-required-positionals limitation.

## 2. Implementation

- Loads importer APIs through normal Python imports without rewriting `sys.path` or clearing `sys.modules`.
- Keeps standalone importer dependencies optional so the wheel owns its compatible OpenUSD and converter dependency set.
- Adds installation guidance, a changelog fragment, and focused regression coverage.

## 3. Validation

- [x] 19 focused tests on Python 3.12.
- [x] Regression test fails with the previous module-search-path mutation and passes with this change.
- [x] URDF and MJCF converter help works without Isaac Sim.
- [x] Direct smoke conversion with `urdf-usd-converter==0.2.0` and `mujoco-usd-converter==0.3.0`.
- [x] Full `./isaaclab.sh -f` pre-commit suite.
- [ ] End-to-end conversion through the exact `isaacsim-asset-isolated==1.0.0` wheel.

## 4. Current blocker

The exact standalone wheel is not available from the tested public PyPI or NVIDIA package indexes. This PR remains a draft until that artifact or its package index is available for end-to-end validation.


kellyguo11:
closing in favor of https://github.com/isaac-sim/IsaacLab/pull/6460

PR #5673, opened 2026-05-18
refactor: move Fabric/USD dispatch from FabricFrameView to FrameView factory

pv-nvidia:
## Problem

`FabricFrameView` has an internal `_use_fabric` flag that falls back to `UsdFrameView` when Fabric is disabled or the device is unsupported. This is a misleading abstraction: the class claims to be a Fabric-accelerated view but sometimes silently delegates everything to a different (slower) implementation. Callers have no way to know which code path actually ran.

## Solution

Move the Fabric-enabled + device-supported check from `FabricFrameView.__init__` up to the `FrameView` factory's `_get_backend()` method. The factory now dispatches to three implementations:

| Condition | Backend key | Class |
|-----------|------------|-------|
| PhysX + Fabric enabled + supported device | `physx` | `FabricFrameView` |
| PhysX without Fabric or unsupported device | `usd` | `UsdFrameView` |
| Newton | `newton` | `NewtonSiteFrameView` |

`UsdFrameView` is eagerly registered on the factory since it lives in `isaaclab` (not a backend package like `isaaclab_physx`), so `FactoryBase`'s dynamic import (`isaaclab_{backend}.sim.views`) can't discover it.

## Changes

- **`FrameView._get_backend()`**: checks `/physics/fabricEnabled` setting + device support via `_is_fabric_supported_device()`, returns `"physx"`, `"usd"`, or `"newton"`
- **`_is_fabric_supported_device()`**: accepts any valid `cuda:<N>` index (not just `cuda:0`) for multi-GPU support
- **`FabricFrameView`**: stripped of all `_use_fabric` conditionals, `SettingsManager` import, `_fabric_supported_devices` constant — it just does Fabric, nothing else
- **`UsdFrameView`**: eagerly registered as `"usd"` backend on the factory

## Result

Each FrameView implementation now has a single responsibility. `FabricFrameView` assumes Fabric is available (the factory guarantees this). No runtime behavior change for existing callers — the same code paths execute, just dispatched from a cleaner location.

Multi-GPU setups (`cuda:1`, `cuda:2`, etc.) are properly supported by the device check function.

## Dependencies

None — this PR applies cleanly to current develop.

### DIFF
--- a/source/isaaclab/isaaclab/sim/views/frame_view.py
+++ b/source/isaaclab/isaaclab/sim/views/frame_view.py
@@ -6,15 +6,37 @@
 """Backend-dispatching FrameView.
 
 ``FrameView(path, device=...)`` automatically selects the right backend:
-- PhysX: :class:`~isaaclab_physx.sim.views.FabricFrameView`
+- PhysX + Fabric enabled + supported device: :class:`~isaaclab_physx.sim.views.FabricFrameView`
+- PhysX without Fabric (or unsupported device): :class:`~isaaclab.sim.views.UsdFrameView`
+- OVPhysX: :class:`~isaaclab_ovphysx.sim.views.OvPhysxFrameView`
 - Newton: :class:`~isaaclab_newton.sim.views.NewtonSiteFrameView`
 """
 
 from __future__ import annotations
 
+import logging
+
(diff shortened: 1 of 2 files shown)


pv-nvidia:

PR #5674, opened 2026-05-18
refactor: fuse set_world_poses/set_scales into single _compose_fabric_transform

pv-nvidia:
## Problem

`FabricFrameView` had duplicated kernel-launch logic in `set_world_poses` and `set_scales`, and the initial USD→Fabric sync called both methods sequentially. This meant:

1. **Duplicated code** — the same 20-line kernel launch pattern appeared twice with minor variations (positions/orientations vs scales).
2. **Double `PrepareForReuse`** — the initial USD→Fabric sync in `_sync_fabric_from_usd_once` called `set_world_poses` then `set_scales`, each invoking `PrepareForReuse`. A second non-idempotent `PrepareForReuse` call could mask a topology-change signal that should have triggered a fabricarray rebuild.

## Solution

Extract `_compose_fabric_transform(positions=None, orientations=None, scales=None, indices=None)` — a single method that composes any subset of transform components into one kernel launch. Components left as `None` are skipped via empty arrays.

- `set_world_poses` → delegates to `_compose_fabric_transform(positions=..., orientations=...)`
- `set_scales` → delegates to `_compose_fabric_transform(scales=...)`
- `_sync_fabric_from_usd_once` → single fused call with all three components

## Additional fix

The topology-change invariant guard in `_rebuild_fabric_arrays` used `assert`, which is stripped under `python -O`. Replaced with `raise RuntimeError` so it's always active.

## Tests

All 36 existing Fabric tests pass (+ 2 xfail).

---

## Merge Order

This PR is part of a stacked series for Fabric-accelerated local poses:

1. **#5672** — service locator ← `this depends on`
2. **#5676** — FabricStageCache
3. **#5675** — indexed fabric kernels
4. **#5673** — factory dispatch
5. **#5674** — fused compose ← **this PR** (depends on #5672)
6. PR C+D — local poses (depends on all above)

### DIFF
--- a/source/isaaclab/isaaclab/sim/views/frame_view.py
+++ b/source/isaaclab/isaaclab/sim/views/frame_view.py
@@ -6,15 +6,25 @@
 """Backend-dispatching FrameView.
 
 ``FrameView(path, device=...)`` automatically selects the right backend:
-- PhysX: :class:`~isaaclab_physx.sim.views.FabricFrameView`
+- PhysX + Fabric enabled + supported device: :class:`~isaaclab_physx.sim.views.FabricFrameView`
+- PhysX without Fabric (or unsupported device): :class:`~isaaclab.sim.views.UsdFrameView`
+- OVPhysX: :class:`~isaaclab_ovphysx.sim.views.OvPhysxFrameView`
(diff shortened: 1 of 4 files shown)


pv-nvidia:

PR #5677, opened 2026-05-18
feat: GPU-accelerated FrameView local poses + a new xform_space_writer API

pv-nvidia:
# Fast GPU pose/scale ops on Fabric + a new writer-scope API

## What this does

Two things:

1. **Local-space pose and scale ops are now GPU-fast on the Fabric backend.**
   They go through the same Warp-kernel path that world-space ops already used.
   The old USD fallback for local-space is ~100× slower (see the table below).

2. **All transform writes go through a small context manager.** You open a
   scope, do your writes inside it, and the scope handles the cleanup:

   ```python
   with view.xform_world_space_writer() as writer:
       writer.set_poses(positions=p, orientations=o)
       writer.set_scales(scales=s)
   ```

   Use `view.xform_local_space_writer()` for local-space writes.

Builds on Piotr's prototype at
[`bareya/pbarejko/camera-update`](https://github.com/bareya/IsaacLab/tree/pbarejko/camera-update).

## Why a scope?

A FrameView keeps two copies of each prim's transform in Fabric storage:
the `omni:fabric:worldMatrix` and `omni:fabric:localMatrix` attributes.
When you write one, the other has to be recomputed so they stay
consistent.

Without a scope, that recompute has to run after every single write. With
a scope, it runs **once**, when the scope closes. You can call `set_poses`
and `set_scales` as many times as you want inside the scope; on exit, one
Warp kernel derives the other space and one `wp.synchronize()` runs.
Empty scopes cost nothing.

The scope also pauses **Fabric Hierarchy**'s transform-change tracking
while you're writing, then restores it on exit. (Fabric itself is just
the data store; Fabric Hierarchy, exposed through USDRT's
`IFabricHierarchy`, is the plugin that watches writes to
`omni:fabric:localMatrix` / `omni:fabric:worldMatrix` and keeps them
mutually consistent across the hierarchy.) Its tracking is pull-based:
a per-attribute listener records writes into a changelog, and the
plugin drains and processes that changelog on the next call to
`IFabricHierarchy::update_world_xforms()`. Pausing the listener does
not block the writes themselves — they still land in Fabric storage —
it just keeps them out of the changelog. With the changelog empty, the
next tick has nothing to "catch up on" for these prims, and can't
decide one of the two spaces is canonical and derive the other from
it. See the next section for why that matters.

Rules:
- Only one writer can be open on a view at a time. Opening a second one
  raises `RuntimeError`.
- While a writer is open, the view's own getters (`get_world_poses`, etc.)
  raise. Read through the writer (`writer.get_poses()`) inside the scope,
  or close the scope first.
- **Do not advance the simulation or render from inside the scope.**
  No `sim.step()`, no `world.render()`, no `SimulationApp.update()`.
  The matrices are mid-write until exit, and rendering against that state
  would read torn data. Keep scopes short and close them before stepping.
- If something raises inside the scope (a user-code bug, a notebook
  cell interrupt, etc.), the scope still runs the opposite-space recompute
  on exit so the world and local matrices stay consistent prim-by-prim.
  The partial write itself is not rolled back — if you need
  all-or-nothing, snapshot the matrices yourself before entering.

## Why we have to coordinate with Fabric Hierarchy (renderer correctness)

A writer scope is synchronous Python code, so the renderer can't see
anything mid-scope — no tick runs until we exit. The risk is on the
**next** render/sim tick after the scope closes.

On that next tick, Fabric Hierarchy runs its
`IFabricHierarchy::update_world_xforms()` step, which can recompute
`omni:fabric:worldMatrix` from `omni:fabric:localMatrix` (or vice
versa). If it thinks one of the two was the user's most recent edit,
it derives the other from it; the derived value — not what we wrote —
is then what the **Fabric Scene Delegate (FSD)** hands to RTX on the
render path.

Three things in this PR keep that next-tick recompute from clobbering our
work:

1. **Opposite-space derive at scope exit.** By the time the scope closes,
   `omni:fabric:worldMatrix` and `omni:fabric:localMatrix` are mutually
   consistent prim-by-prim, so any recompute Fabric Hierarchy does is a
   no-op.

2. **Fabric-Hierarchy change tracking is paused for the whole scope.**
   The writer calls `track_local_xform_changes(False)` and
   `track_world_xform_changes(False)` on enter (saving the prior state)
   and restores them on exit. While tracking is paused, the listeners
   don't record our writes in their changelog, so when the next tick
   runs there's nothing for `update_world_xforms()` to "catch up on".

3. **Two persistent selections with explicit read-only / read-write
   flags.** Outside the scope, the view's selection is fully read-only —
   both `omni:fabric:worldMatrix` and `omni:fabric:localMatrix` are
   flagged RO. Inside the scope, the writer flips to a fully read-write
   selection.

   ```
   _sel_ro :  worldMatrix=RO, localMatrix=RO   (steady state, between writes)
   _sel_rw :  worldMatrix=RW, localMatrix=RW   (active during a writer scope)
   ```

   Both selections are built once when the view is initialized and kept
   alive for its lifetime. The writer scope flips a single flag (`_is_rw`)
   on enter/exit; nothing is rebuilt. The RO steady state tells Fabric
   Hierarchy's next-tick `update_world_xforms()` that no attribute is
   user-authored, so it leaves both alone.

## Backend support

- **Fabric (PhysX):** the fast GPU path described above.
- **USD, Newton, OVPhysX:** the writer is a thin wrapper. Writes go straight
  to the backend's existing storage. No batching, no extra recompute (there
  is no second matrix to keep in sync). Same API across all four backends.

Newton has two different "scale" ideas, and the API keeps them apart:

- `NewtonSiteFrameView.set_scales(...)` (deprecated) writes the **Newton
  collision-shape geometry** size.
- The writer's `set_scales(...)` writes per-site **transform scale**,
  matching what USD FrameView does.

They operate on different state and are not merged.

## API changes

**New (recommended):**
- `view.xform_world_space_writer()` / `view.xform_local_space_writer()` —
  a context manager with `set_poses`, `set_scales`, `get_poses`,
  `get_scales`.

**Deprecated (still works, warns once per class on first use):**
- `set_world_poses` / `set_local_poses` — use the writer scope.
- `get_scales` / `set_scales` — use `get_local_scales` /
  `get_world_scales`, or the writer's `set_scales`.

## Benchmark (1024 prims, 50 iterations, Blackwell RTX PRO 5000)

Per-iteration timings (lower is better):

| Operation                  | USD (ms) | Fabric (ms) | Newton Site (ms) |
|----------------------------|---------:|------------:|-----------------:|
| Get World Poses            |   8.6148 |      0.0375 |           0.0307 |
| Set World Poses            |  19.1357 |      0.1090 |           0.0368 |
| Get Local Poses            |   5.9556 |      0.0374 |           0.0497 |
| Set Local Poses            |   7.9794 |      0.0979 |           0.0790 |
| Get World Scales           |  13.1193 |      0.0371 |           0.0015 |
| Set World Scales           |  21.4571 |      0.1014 |           0.0612 |
| Get Local Scales           |   3.2544 |      0.0375 |           0.0022 |
| Set Local Scales           |   3.8350 |      0.0937 |           0.0587 |
| Get Both (World+Local)     |  14.7506 |      0.0738 |           0.1302 |
| Interleaved World Set→Get  |  28.1046 |      0.1432 |           0.1588 |
| **Per-iter total**         | **126.2** |  **0.77**  |        **0.61**  |

Speedup vs USD (per-iter ops; one-time view construction excluded):

| Operation                  | Fabric × | Newton Site × |
|----------------------------|---------:|--------------:|
| Get World Poses            |   229.6× |        280.2× |
| Set World Poses            |   175.6× |        519.8× |
| Get Local Poses            |   159.0× |        119.9× |
| Set Local Poses            |    81.5× |        101.0× |
| Get World Scales           |   353.9× |       8911.5× |
| Set World Scales           |   211.6× |        350.4× |
| Get Local Scales           |    86.8× |       1489.9× |
| Set Local Scales           |    40.9× |         65.4× |
| Get Both (World+Local)     |   200.0× |        113.3× |
| Interleaved World Set→Get  |   196.2× |        177.0× |
| **Overall (per-iter)**     | **164×** |       **207×** |

One-time view construction (reported separately, not part of the per-iter total): USD 4.6 ms, Fabric 4.4 ms, Newton Site 1013 ms — the Newton number is dominated by stage population on the first call. Steady-state per-iteration cost is what the speedup row reflects.

## Notes for reviewers

- Getters call `wp.synchronize()` before returning, so the returned
  `ProxyArray` is always immediately readable from both GPU and host
  code — no caller-side sync needed. (Both the cached and the
  per-indices paths sync; this used to be asymmetric.)


pv-nvidia:

scripts/benchmarks/benchmark_xform_prim_view.py
```diff
@@ -217,6 +229,72 @@ def to_torch(a):
         computed_results["local_translations_after_set"] = to_torch(ta).clone()
         computed_results["local_orientations_after_set"] = to_torch(ola).clone()
 
+        # -- get_world_scales ----------------------------------------------
+        if is_newton:
+            torch.cuda.synchronize()
```
kellyguo11: we could use wp.synchronize directly here instead of torch

source/isaaclab/isaaclab/sim/views/base_frame_view.py
```diff
@@ -60,28 +245,26 @@ def set_world_poses(
...
-        ...
+        if not BaseFrameView._set_world_poses_deprecated_warned:
+            BaseFrameView._set_world_poses_deprecated_warned = True
+            warnings.warn(
+                "set_world_poses() is deprecated. Use 'with view.xform_world_space_writer() as w:"
+                " w.set_poses(...)' instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+        with self.xform_world_space_writer() as writer:
```
kellyguo11, line 266: I think forcing users to apply the scope could be extra complexity burden on the users. what if we don't deprecate these APIs so that users can still use them if they want to, but we still update our examples to use the scopes so that we get the performance benefit? so we leave both options available, depending on whether users want extra performance / more code complexity or code simplicity / possibly lower perf

kellyguo11 (approved):

PR #5681, opened 2026-05-18
chore: update ruff target-version to py312

pv-nvidia:
The project requires `python >= 3.12` but ruff was configured with `target-version = "py310"`. This causes false positives (e.g. `ExceptionGroup` flagged as undefined) and prevents correct stdlib import sorting for modules like `tomllib`.

This PR updates `target-version` to `"py312"` and applies the resulting auto-fixes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,7 +60,7 @@ all = [
 
 [tool.ruff]
 line-length = 120
-target-version = "py310"
+target-version = "py312"
 
 # Exclude directories
 extend-exclude = [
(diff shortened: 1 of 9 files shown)


pbarejko (approved):

pbarejko:
The intention for develop is to be used with Isaac Sim 6.0 that has embedded Python 3.12. I think we should accept this change @kellyguo11 ?

kellyguo11 (approved):

kellyguo11:
yup LGTM

PR #5687, opened 2026-05-19
[Test] Stabilize RTX MDL shader-warmup flake in rendering correctness tests

hujc7:
## Description

The rendering-correctness tests intermittently fail on cold-cache CI runners with "Camera output is all zeros or all inf" for `simple_shading_*_mdl` and `simple_shading_constant_diffuse` variants — the GPU returns a still-zero framebuffer because the RTX MDL material has not finished compiling by the time the test reads the camera tensor.

Add a poll-until-non-zero warmup that exits as soon as every camera-output tensor has a non-zero max:

1. **`test_shadow_hand_vision_presets.py::test_camera_renders_not_empty`** — polls via `env.step()` with a 60-step cap. (This test has no `flaky` retry, so a single-step fixture is the actual failure surface.)
2. **`rendering_test_utils.warmup_render_until_nonzero`** — iterates every sensor in `env.scene.sensors` and polls via `sim.render()` + `scene.update()` with a 30-pass cap. Called from the four rendering helpers in `rendering_test_utils.py` (`rendering_test_shadow_hand` / `_cartpole` / `_dexsuite_kuka`) and from `test_rendering_registered_tasks.py`. Mirrors the pattern already in `DirectRLEnvCfg.num_rerenders_on_reset` / `wait_for_textures`; physics state is not advanced, so the existing goldens at the post-init "first non-zero frame" stay valid.

An earlier attempt at a fixed-pass warmup over-rendered in the goldenfile helpers and broke `dexsuite_kuka` goldens at 91% pixel diff — early-exit on first-non-zero avoids that by landing at the same frame the goldens were captured at.

## Type of change

## Checklist

- [x] I have updated the changelog and the corresponding version in the extension's \`config/extension.toml\` file

### DIFF
--- a/source/isaaclab_tasks/changelog.d/jichuanh-stabilize-rendering-mdl-warmup.rst
+++ b/source/isaaclab_tasks/changelog.d/jichuanh-stabilize-rendering-mdl-warmup.rst
@@ -1,0 +1,13 @@
+Fixed
+^^^^^
+
+* Stabilized RTX MDL shader-warmup flakes in the rendering-correctness tests
+  by stepping the env 10 frames before reading the camera tensor, instead of
+  a single step. Affected ``test_shadow_hand_vision_presets.py``'s
+  ``test_camera_renders_not_empty`` and the three helper-driven tests in
+  ``rendering_test_utils.py``
+  (``rendering_test_shadow_hand``/``_cartpole``/``_dexsuite_kuka``) — all of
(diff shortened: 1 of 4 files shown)


hujc7:
does not seem working. need to change golden file.

hujc7:
Reopening. Going back at this with the warmup + golden regeneration approach, validated locally before pushing.

PR #5691, opened 2026-05-19
[OvPhysX] Add RayCaster sensor for the OVPhysX backend

AntoineRichard:
# Description

Adds `isaaclab_ovphysx.sensors.ray_caster` so `RayCaster(cfg)`,
`RayCasterCamera`, `MultiMeshRayCaster`, and `MultiMeshRayCasterCamera`
dispatch correctly under an OVPhysX `SimulationContext`. Unblocks
`Isaac-Velocity-Rough-Anymal-D-v0` under `presets=ovphysx` (the
height_scanner is a `RayCaster`).

Mirrors `isaaclab_physx.sensors.ray_caster` structure: a single
`_OvPhysxRayCasterMixin` (~200 lines in `ray_caster.py`) carries the
backend-specific pose-tracking surface, reading live body poses via the
ovphysx wheel's
`create_tensor_binding(pattern=..., tensor_type=RIGID_BODY_POSE)` API.
Three 14-line sibling modules compose the mixin with the matching
`Base*` class. Static (non-physics) sensor frames fall back to a
one-time USD pose snapshot, matching PhysX's
`_initialize_static_pose_tracking`.

Multi-mesh dynamic targets create a second binding via the same
`create_tensor_binding` call. v1 limitation: target paths must dedup
to a single env-wildcard pattern; multi-pattern targets raise
`NotImplementedError` (mirrors the OVPhysX `ContactSensor`
`track_pose` single-body constraint).

Also wires `isaaclab_ovphysx.sensors.ContactSensorCfg` and
`isaaclab_ovphysx.physics.OvPhysxCfg` into the shared
`LocomotionVelocityRoughEnvCfg.RoughPhysicsCfg` so
`Isaac-Velocity-Rough-Anymal-D-v0` can pick the right preset under
`presets=ovphysx`.

Built on top of #5678 (FrameView, **now merged**). Anymal-D Rough at
4096 envs needs FrameView's `clone_usd=True` + env_0 USD stripping to
avoid a wheel-side hang during init; with #5678 in develop, this PR is
now a standalone three-commit diff:

- `Wire OvPhysX preset into shared RoughPhysicsCfg`
- `Route RayCaster dispatch to ovphysx backend`
- `Add isaaclab_ovphysx.sensors.ray_caster package`

Verified locally:

- `./isaaclab.sh -p scripts/environments/random_agent.py --task Isaac-Velocity-Rough-Anymal-D-v0 --num_envs 64 --headless presets=ovphysx` reaches the agent loop cleanly.
- `./isaaclab.sh -p scripts/reinforcement_learning/rsl_rl/train.py --task Isaac-Velocity-Rough-Anymal-D-v0 --num_envs 4096 --max_iterations 100 --headless presets=ovphysx` completes 100 iterations with non-degenerate mean reward.

Fixes # (issue)

## Type of change

## Screenshots

N/A — backend dispatch + sensor implementation, no UI surface.

## Checklist


marcodiiga (dismissed):
Superseded by the consolidated follow-up comment: https://github.com/isaac-sim/IsaacLab/pull/5691#issuecomment-4633689504

marcodiiga:
Consolidating my earlier notes on the OVPhysX multi-mesh ray-caster path.

I do not think the documented multi-pattern limitation, by itself, should block the whole PR. The Anymal-D single-mesh height-scanner path is separate. The two concrete issues I would still address for the multi-mesh dynamic-target path are:

1. A single env-wildcard dynamic target can be expanded into multiple concrete env patterns and rejected.

`BaseMultiMeshRayCaster` preserves ClonePlan-backed owner expressions such as the destination template plus owner suffix (`source/isaaclab/isaaclab/sensors/ray_caster/base_multi_mesh_ray_caster.py:258`) and returns those tracked expressions (`base_multi_mesh_ray_caster.py:268-273`). The OVPhysX override then immediately resolves each expression against live USD prims (`source/isaaclab_ovphysx/isaaclab_ovphysx/sensors/ray_caster/ray_caster.py:222-242`). In clone-USD scenes, that means an env-wildcard target like `/World/envs/env_.*/Object` can become concrete paths like `/World/envs/env_0/Object`, `/World/envs/env_1/Object`, etc. `_ovphysx_body_glob()` only converts `{}` and `.*`, not concrete `env_N` segments (`ray_caster.py:60-67`), so the dedup at `ray_caster.py:247` can see multiple patterns and raise the v1 multi-pattern `NotImplementedError` at `ray_caster.py:248-253` even though this is one body type across envs.

Please preserve the ClonePlan wildcard expression, or collapse cloned env paths back to one ovphysx wildcard before the pattern-dedup check. A targeted test with one env-wildcard, single-body dynamic target would cover this.

2. Dynamic-target bindings are not destroyed on invalidation.

`_invalidate_initialize_callback()` calls `super()._invalidate_initialize_callback(event)` first (`ray_caster.py:311-313`). For `MultiMeshRayCaster`, the base callback clears `self._mesh_views` (`base_multi_mesh_ray_caster.py:462-466`) before the OVPhysX loop tries to destroy the tracked-target bindings (`ray_caster.py:320-322`). That makes the destroy loop iterate an empty list, so the `TensorBinding` handles created by `_create_tracked_target_view()` (`ray_caster.py:255-261`) can leak across sim stop/reinit.

Please snapshot/destroy `_mesh_views` before calling `super()`, or otherwise ensure backend-specific cleanup runs before the base class clears the list.

marcodiiga:
One more OVPhysX RayCaster pose-tracking issue: this path appears to assume that the `RIGID_BODY_POSE` binding row order matches the USD prim/env order returned by `find_matching_prims()`.

`_initialize_pose_tracking()` creates the binding from `body_glob`, then builds `offset_pos` / `offset_quat` by iterating the USD `prims` list. Later `_get_view_transforms_wp()` reads the binding into `transforms`, and `update_ray_caster_kernel` indexes `transforms[env_id]` together with `offset_pos[env_id]`. If the OVPhysX binding row order differs from USD discovery order, the sensor applies the wrong per-env offset to the body transform. There is also no validation that the binding row count/order matches the resolved sensor prims.

Other OVPhysX code that consumes `RIGID_BODY_POSE`, such as `OvPhysxFrameView`, uses `binding.prim_paths` to map body paths to binding rows before building indices. Can this raycaster path either build offsets in binding-row order using `prim_paths`, or add a hard validation that the binding rows match the expected env/body order? A multi-env test with distinct per-env sensor offsets would catch this.

AntoineRichard:
Follow-up on the OVPhysX RayCaster pose-tracking offset ordering concern: PhysX RayCaster does not currently support distinct per-env sensor-to-body offsets either. It resolves the source-side sensor offset and broadcasts that across the rigid-body view rows.

Keeping OVPhysX aligned with that backend contract is the intended scope for this PR, so adding binding-row-specific/per-env-authored offsets is out of scope here. If we want that behavior later, it should be added consistently across RayCaster backends with shared contract coverage.

marcodiiga (approved):
Approving after re-review. Existing review comments still stand where applicable.

PR #5695, opened 2026-05-20
[CI] Cross-platform — Part 1: pytest markers and shared scaffolding (base)

hujc7:
## Summary

Base PR for cross-platform CI on Windows and NVIDIA DGX Spark (ARM/aarch64 Linux). No user-facing API change.

- Registers `windows_ci` and `arm_ci` pytest markers in pyproject.toml.
- AppLauncher argv strip recognizes the new markers so `pytest -m windows_ci` (etc.) does not leak into Isaac Sim's argparse.
- AssetConverterBase USD scratch dir: hardcoded `/tmp/IsaacLab` → `tempfile.gettempdir()` for cross-platform compatibility.

This base PR ships only the marker registration and the shared scaffolding. Tagging specific suites with these markers, and the CI workflows that consume them, ship in the follow-up parts below — that keeps the marker/scaffolding change reviewable on its own and lets the consumer PRs rebase cleanly once it lands.

## Prior art and credits

This work is a derivative of and builds directly on two prior PRs by other contributors:

- #3900 by Kelly Guo (kellyguo11), against `main`, 2025-10-31 — "test windows CI". Designed and introduced the cross-platform marker scheme (`windows`, `windows_ci`, `arm`, `arm_ci`), tagged ~50 existing tests across the codebase, added the AppLauncher argv-strip pattern so the new pytest `-m` filters do not leak into Isaac Sim's argument parser, and pioneered the Windows CI job wiring. The marker design and AppLauncher behavior in this PR are taken directly from #3900.
- #3845 by Alexander Poddubny (nv-apoddubny), against `main`, 2025-10-27 — "Adds basic CI verification on Windows". Built the original Windows self-hosted runner infrastructure: `isaaclab.bat` entry script, the initial `.github/workflows/build.yml` Windows job, and the self-hosted-runner labeling conventions this PR's downstream Part 2/Part 3 workflows still rely on.

Both PRs target `main` and have not been merged. This PR rebases that work onto `develop` (where the active development happens — `main` is behind by hundreds of commits), narrows the base PR to just the marker scheme plus shared scaffolding (so the marker/runner plumbing can be validated end-to-end before tests are tagged and re-expanded), and adds an `arm_ci` marker variant + ARM/Spark runner support that #3900/#3845 did not include.

Credit for the architecture goes to the authors above; this PR is the rebase + minimization step.

## Series

PRs prefixed with `[CI] Cross-platform —` all contribute to bringing Isaac Lab CI up on Windows and NVIDIA DGX Spark runners. This PR is the base everything else builds on. Current siblings:

- Part 1 (this PR, base) — Markers and shared scaffolding
- Part 2 — #5698 — ARM/Spark workflow (`.github/workflows/arm-ci.yaml`)
- Part 3 — #5700 — Windows workflow (`.github/workflows/windows-ci.yaml`)

Siblings depend on this PR's marker scheme; the same scaffolding lines are carried in their branches and dedupe on rebase after this lands. Additional follow-up parts can be added later without renumbering existing ones.

## Test plan

- [x] `pytest --markers` lists `windows_ci` and `arm_ci` (registered, no `PytestUnknownMarkWarning`).
- [x] AppLauncher argv strip drops `-m windows_ci` / `-m arm_ci` before Isaac Sim's argparse runs.
- [x] `./isaaclab.sh -f` (pre-commit) passes.
- [x] Merged latest `develop`: the markers list is the union with the newly added `device_split` marker, and `--capture=no` argv stripping from develop is preserved.


source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1127,12 +1127,18 @@ def _create_app(self):
...
+        # and -c **/pyproject.toml
         indexes_to_remove = []
         for idx, arg in enumerate(sys.argv[:-1]):
             if arg == "-m":
                 value_for_dash_m = sys.argv[idx + 1]
-                if "pytest" in value_for_dash_m or "isaacsim_ci" in value_for_dash_m:
+                if (
+                    "pytest" in value_for_dash_m
+                    or "isaacsim_ci" in value_for_dash_m
+                    or "windows_ci" in value_for_dash_m
```
kellyguo11, line 1205: are these new annotators necessary for this check? I think this was initially added for the isaac sim pipeline to work properly by making sure the correct arguments gets passed over. will that also be impacted by the windows and arm annotations?

pyproject.toml
```diff
@@ -194,6 +194,10 @@ ignore-words-list = "haa,slq,collapsable,buss,reacher,thirdparty"
 
 markers = [
     "isaacsim_ci: mark test to run in isaacsim ci",
+    "windows: mark test as runnable on Windows platforms",
```
kellyguo11: would this be a lot of maintenance to tag every test that would work on windows? I think most of our workflows should support windows already with a few exceptions. perhaps it's easier to mark the ones that are not supported instead?

source/isaaclab/isaaclab/app/app_launcher.py
hujc7: It's for pytest selecting tests, which uses the same mechanism as isaacsim_ci tag. While this is not quite scalable at the moment, some investigation into argparse refactor should help. e.g. to only pass appLauncher recognized arguments to it.

pyproject.toml
hujc7: As a start, I think whitelist should be better. Windows pool size is also just 1/4 compare to linux ones. I can test on the overall duration + test support and see if a negative tag works better, but I think that will be in followup PRs given that windows still have driver + isaacsim develop access issue not resolved.
The PR serves as the common setup for followups so they can all go independently.

kellyguo11: I see, then is there a need to keep both windows and windows_ci tags? I was thinking if we have both, windows could be a blacklist that determines which tests are not supported on windows. and windows_ci could be a whitelist that enables a small set of tests to run

hujc7: "windows" is dropped, and only "windows_ci" is added in the latest patch. Personally, I would whitelist should be sufficient and it's more verbose at per test/file level that where the test runs.

source/isaaclab/isaaclab/app/app_launcher.py
AntoineRichard: Are we not running the the whole batch of tests on windows? Would it make sense to make a specific test shard just for windows? (that would remove the annotation)

pyproject.toml
```diff
@@ -194,6 +194,8 @@ ignore-words-list = "haa,slq,collapsable,buss,reacher,thirdparty"
 
 markers = [
     "isaacsim_ci: mark test to run in isaacsim ci",
+    "windows_ci: mark test to run on Windows platforms in CI",
+    "arm_ci: mark test to run on ARM platforms in CI (e.g. NVIDIA DGX Spark)",
```
AntoineRichard, line 201: Is this better than a specific shard?

AntoineRichard (commented):
I'm not sure what's best between annotated tests and a specific test shard for windows / arm.

pyproject.toml
hujc7: Not exactly sure what a shard means here. Does it mean selecting the tests at workflow level? If so, I am trying to make it obvious at test level all the configuration.

AntoineRichard: Here we'd mark specific tests in our CI for windows / arm right? What I'm wondering is if we setup a specific job for windows that would list the test files we want to run. It seems weird to me to only mark a couple of tests here and here across the code base for these platform tests. Or it's not how this works and these markers are used to mark files instead?

hujc7: Yes both. It can opt in at file level or at per test level. mainly two problems 
1. avoid modification at workflow level to include tests, so it's verbose at per test/file level that it's included for cross platform runs, which might have more traction.
2. the CI resources are skewed. The current setup has up to 10 windows machines and 2 spark. It might be better to auto-include new tests but being conservative for the initial setup.

AntoineRichard (approved):

PR #5698, opened 2026-05-20
[CI] Cross-platform — Part 2: ARM/Spark workflow

hujc7:
## Summary

- Adds an `arm-ci` job to the existing `Docker + Tests` workflow for NVIDIA DGX Spark self-hosted runners. It builds and tests on the same ARM runner because ECR is not wired for arm64.
- Adds marker-driven ARM test selection through `CI_MARKER`, plus a shared kitless Cartpole state/camera smoke used by ARM and installation CI.
- Extracts Docker dependency-hash and registry-login setup into shared composite actions used by the local Docker and ECR build paths.
- Makes Docker dependency-cache keys platform-specific and bounds stale ARM cache tags on the long-lived Spark runner.

## 1. ARM/Spark CI

1.1 The `arm-ci` job shares the existing `changes` and `config` gating. It builds `Dockerfile.base` for `linux/arm64`, runs the `arm_ci`-selected tests, and then runs the shared Cartpole smoke.

1.2 OVRTX and OVPhysX are installed explicitly for the rendering probes. Missing ARM runtime dependencies fail instead of being hidden by architecture-specific skips.

1.3 The job uses the same workflow and shared actions as the amd64 jobs. ARM enables 14-day stale dependency-cache eviction because its runner has persistent local Docker storage and no ECR handoff.

1.4 The `arm_ci` marker run passes `test-k-expr: not ovphysx`, the same per-file `-k` expression the x86 `rendering-correctness-kitless` job applies on public pip-index runs. The ovphysx-backed rendering params require OVPhysX >= 0.5.1, which has no aarch64 wheel.

## 2. Shared CI utilities

2.1 `_lib/compute-deps-hash` computes one dependency-cache identity from install inputs, the resolved base-image digest, and target platform. Both `docker-build` and `ecr-build-push-pull` consume it.

2.2 `_lib/setup-docker-config` centralizes the credential-helper workaround and optional NGC login. It is idempotent when the ECR action delegates to `docker-build`.

2.3 `run-tests` forwards both upstream's per-file `test-k-expr` and this PR's `ci-marker` inputs into the test container.

## 3. Shared Cartpole smoke

3.1 `source/isaaclab/test/install_ci/misc/cartpole_training_smoke.py` is the common prepared-environment probe for state Cartpole training, camera observation/reward validity, and camera Cartpole training.

3.2 CLI and wheel installation tests execute the same probe inside their newly installed environments; ARM executes it inside its prepared image.

3.3 The base image installs `git-lfs` so RSL-RL can snapshot repository state from a mounted LFS checkout.

## 4. Cross-platform support

4.1 `AppLauncher` derives `EXP_PATH` from the installed `isaacsim` package when the variable is missing, while leaving existing configured environments unchanged.

4.2 `tools/conftest.py` supports `CI_MARKER` as a file-selection and pytest marker filter parallel to `ISAACSIM_CI_SHORT`, exact node-ID selection, and per-file `-k` filtering.

## Known problems

- The three rendering modules apply `arm_ci` at module scope, so ARM collects all 32 cases per module. The ovphysx-backed cases are now deselected via `test-k-expr: not ovphysx` (section 1.4), leaving 15 Newton cases running per module. Without the filter, the extra ovphysx-ovrtx cases in `test_rendering_cartpole_kitless.py` pushed a single process past the OVRTX per-process SyncScopeId limit (~15 renderer creations), crashing HydraEngine initialization on the 16th env; this also reproduces on x86, so the failure is not ARM-specific. Narrowing ARM to the two intended representative tests is still pending.
- Current `develop` pins OVPhysX `0.5.2+head.f62c22207c`, while this job installs the publicly available ARM `0.4.13` wheel. With the `not ovphysx` filter the ARM rendering selection is Newton-only, matching the x86 public pip-index fallback; revisit if an aarch64 OVPhysX >= 0.5.1 wheel or wheelhouse becomes available.

## 5. Series

Part 2 of `[CI] Cross-platform —`. Depends on #5695 (Part 1). The Windows workflow is #5700 (Part 3).

## Test plan

- [x] Merge current `develop` (`5f2db9902baa724427b7aa654835aa559fce64e8`).
- [x] Run `./isaaclab.sh -f` before commit and before push.
- [x] Collect the refreshed ARM rendering selection locally: 96 items (32 per rendering module), confirming the current over-selection.
- [x] Deselect ovphysx rendering params in the ARM job (`test-k-expr: not ovphysx`), mirroring the x86 kitless public pip-index fallback.
- [ ] Narrow ARM rendering coverage to the two intended representative tests.
- [ ] Run the two selected tests and shared Cartpole smoke on Spark/aarch64.
- [ ] Confirm the refreshed `Docker + Tests` workflow is green.
- [ ] Trigger `daily-compatibility` before merge.


myurasov-nv:
Why are we creating yet another separate CI for ARM instead of just letting some of the tests from run on ARM? I vote against that.

hujc7:
> Why are we creating yet another separate CI for ARM instead of just letting some of the tests from run on ARM? I vote against that.

Should be fixed.

source/isaaclab_tasks/test/core/test_cartpole_training_smoke.py
AntoineRichard, line 1: Why not run this similarly to the other training tests and just add a windows/arm test list like we have the short / long list?

AntoineRichard (commented):
I'm not sure about the git action things

source/isaaclab_tasks/test/core/test_cartpole_training_smoke.py
hujc7: The goal is to delegate config to per test/file level, while keep workflow more stable if I understand the question correctly here

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1089,8 +1089,22 @@ def _resolve_experience_file(self, launcher_args: dict):
...
+        # If nothing is provided resolve the experience file based on the headless flag.
+        # EXP_PATH is normally set by ``isaacsim.bootstrap_kernel()`` on first import.
+        # If it is not set (e.g. on aarch64 where the bootstrap early-return triggered
+        # under certain install layouts), derive it from the installed isaacsim package.
+        kit_app_exp_path = os.environ.get("EXP_PATH")
+        if not kit_app_exp_path:
+            try:
+                import isaacsim as _isaacsim_for_paths
+            except ImportError as e:
+                raise RuntimeError(
```
kellyguo11, line 1150: is this path only triggered when we are launching with isaac sim? just want to make sure we don't throw this error if we try to run a kitless path

source/isaaclab_tasks/test/core/test_cartpole_training_smoke.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
kellyguo11: is this one similar to some of the cartpole tests we have under `source/isaaclab/test/install_ci/cli/test_install_rl_uv_env.py`? if we could consolidate some of the tests, that'd help keep files more organized.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -420,6 +421,8 @@ def _require_ovlibs_install(request):
 
                 print(f"ovrtx version: {ovrtx.__version__}")
             except ImportError as exc:
+                if platform.machine() == "aarch64":
```
kellyguo11: ovrtx and ovphysx seem to provide aarch64 wheels - https://pypi.nvidia.com/ovrtx/ https://pypi.nvidia.com/ovphysx/

source/isaaclab/isaaclab/app/app_launcher.py
hujc7: I believe that's determined in simulation context related utilities. In kitless path, applauncher should be skipped.

source/isaaclab_tasks/test/core/test_cartpole_training_smoke.py
hujc7: There's a limitation that install tests needs to be in the same folder. the current approach is to keep the test in the install_ci folder, but keep it in separate file such that other CI can use it.

.github/actions/docker-build/action.yml
```diff
@@ -24,61 +24,222 @@ inputs:
...
-          echo "Logging into NGC registry..."
-          docker login -u \$oauthtoken -p ${{ env.NGC_API_KEY }} nvcr.io
-          echo "✅ Successfully logged into NGC registry"
+        DEPS_HASH="${{ inputs.deps-hash || steps.deps-hash.outputs.hash }}"
+        LOCAL_DEPS_TAG="$(echo "${{ inputs.image-tag }}" | cut -d: -f1):deps-${DEPS_HASH}"
+
+        echo "🔵 Local deps tag: ${LOCAL_DEPS_TAG}"
+        echo "LOCAL_DEPS_TAG=${LOCAL_DEPS_TAG}" >> "$GITHUB_ENV"
+
+        if docker image inspect "${LOCAL_DEPS_TAG}" >/dev/null 2>&1; then
```
kellyguo11, line 134: does this guarantee that it will always use the latest isaac lab code? the agent picked up a concerning behavior that I'm not too sure if it's true or not

```
Commit A: build dependencies + source A → cached image
Commit B: source changes, dependencies do not

Current behavior:
cached image containing source A → renamed as Commit B image
```

.github/workflows/build.yaml
```diff
@@ -273,6 +273,67 @@ jobs:
...
+  # aarch64 build + marker-gated tests on NVIDIA DGX Spark self-hosted runners.
+  # Build and test must share one runner because ECR is not wired for arm64 —
+  # the locally-built image cannot be handed off across machines.
+  arm-ci:
+    name: arm-ci
+    runs-on: [self-hosted, arm64]
+    needs: [changes, config]
+    if: needs.changes.outputs.run_docker_tests == 'true'
+    timeout-minutes: 60
+    continue-on-error: true
```
kellyguo11, line 285: would this prevent us from getting an accurate signal on whether the tests passed or failed?

hujc7: It just means that if the job failed, merge is not blocked. If we are confident, this can be made required test.

.github/actions/docker-build/action.yml
hujc7: image is deduped w.r.t dependency hash. source code is always mounted.

kellyguo11 (approved):

PR #5710, opened 2026-05-20
Fix NewtonManager scene data provider backend detection when _sim is unset

daniela-hase:
## Summary

- `NewtonManager._backend_is_newton` returned `False` when
  `PhysicsManager._sim` was `None`, even if a `SimulationContext`
  instance existed — leaving backend detection inconsistent with
  `update_visualization_state`, which already fell back to
  `SimulationContext.instance()`.
- Introduce a shared `NewtonManager.get_scene_data_provider` classmethod
  that prefers `PhysicsManager._sim` and falls back to
  `SimulationContext.instance()`.
- Route both `_backend_is_newton` and `update_visualization_state`
  through the new helper so they agree on the active scene-data
  provider.

## Test plan

- [ ] Run the relevant `isaaclab_newton` physics tests:
      `./isaaclab.sh -p -m pytest source/isaaclab_newton/test`
- [ ] Exercise a PhysX-backed scene where `PhysicsManager._sim` is unset
      at the moment `get_state` is called and confirm visualization
      transforms are populated.
- [ ] Confirm Newton-backed runs still short-circuit
      `update_visualization_state` (no shadow model built).
- [ ] `./isaaclab.sh -f`

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -1795,15 +1795,15 @@ class NewtonManager(PhysicsManager):
         coordinate the sync explicitly.
         """
 
+        if scene_data_provider is None:
+            scene_data_provider = cls.get_scene_data_provider()
+        if scene_data_provider is None:
+            return
+
         if cls._backend_is_newton(scene_data_provider):
             return
         cls._ensure_visualization_model()
         if cls._state_0 is None or cls._model is None or cls._state_0.body_q is None:
-            return
-
-        if scene_data_provider is None:
-            scene_data_provider = cls.get_scene_data_provider()
-        if scene_data_provider is None:
             return
 
         if cls._scene_data is None:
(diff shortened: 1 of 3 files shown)


fatimaanes (commented):
Nice clean up!

I'm slightly questioning Greptile's review - it flags a "double resolution" of the scene data provider in `update_visualization_state`, but that's actually not the case in your new code. You resolve the provider upfront and pass it into `_backend_is_newton(scene_data_provider)`, which takes the `is not None` branch and never calls `get_scene_data_provider()` again. So it's resolved exactly once in that flow. There is a second resolution through `_ensure_visualization_model `→ `_backend_is_newton()` but that's pre-existing and only fires on first init, I don't think it's something this PR needs to address. 

But I could be wrong, I'm going against a bot. lol

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1761,6 +1761,23 @@ def _build_visualization_model_from_stage(cls, stage) -> ModelBuilder | None:
...
+        """
+        sim = PhysicsManager._sim
+        if sim is None:
+            from isaaclab.sim import SimulationContext
+
+            sim = SimulationContext.instance()
+
+        if sim is not None:
+            return sim.get_scene_data_provider()
+        return None
```
pbarejko: More pythonic way would be to throw an exception here.

pbarejko (approved):

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
daniela-hase: Done! :)

PR #5715, opened 2026-05-21
[PresetCLI] Make physics=/renderer=/presets= work on the unified train/play scripts

hujc7:
## 1. Summary

- Typed preset selectors (`physics=NAME`, `renderer=NAME`, `presets=NAME`) from #5587 silently fail on the unified `scripts/reinforcement_learning/train.py` dispatcher with `Key 'physics' is not in struct`.
- Root cause is merge-ordering between #5587 (preset CLI) and #5623 (unified train/play refactor): the new `train_<library>.py` / `play_<library>.py` files were authored without the `setup_preset_cli` + `fold_preset_tokens` integration that #5587 had applied to the older scripts.
- Wire the integration into the 4 unified train scripts and the 4 unified play scripts. 8 files, +71 / −20 LOC.

## 2. Repro

Before the fix (on develop):

```
$ python scripts/reinforcement_learning/train.py --rl_library rsl_rl \
      --task=Isaac-Ant-v0 physics=newton_mjwarp --headless

[INFO]: Parsing configuration from: isaaclab_tasks.manager_based.classic.ant.ant_env_cfg:AntEnvCfg
[INFO]: Parsing configuration from: isaaclab_tasks.manager_based.classic.ant.agents.rsl_rl_ppo_cfg:AntPPORunnerCfg
Could not override 'physics'.
To append to your config use +physics=newton_mjwarp
Key 'physics' is not in struct
    full_key: physics
    object_type=dict
```

After the fix: training launches normally with the Newton MuJoCo-Warp backend.

For an invalid value (`physics=does_not_exist`) the resolver emits a helpful error listing available variants and their cfg paths:

```
ValueError: Unknown preset(s): does_not_exist

Available presets (grouped by affected paths):

  newton_kamino
    -> env.sim.physics

  newton_mjwarp, physx
    -> env.observations
    -> env.sim.physics
```

## 3. Fix

Apply the canonical pattern from `isaaclab_tasks.utils.preset_cli`:

For train scripts:
```python
args_cli, hydra_args = setup_preset_cli(parser, argv)  # was: parser.parse_known_args(argv)
# ...
set_hydra_args(fold_preset_tokens(hydra_args))         # was: set_hydra_args(hydra_args)
```

For play scripts (which set `sys.argv` directly):
```python
args_cli, hydra_args = setup_preset_cli(parser)        # was: parser.parse_known_args()
sys.argv = [sys.argv[0]] + fold_preset_tokens(hydra_args)
```

`rsl_rl/train_rsl_rl.py` has an extra `--external_callback` intersection step; the fold runs after the intersection, matching the documented pattern.

## 4. Out of scope

- `train_rlinf.py` / `play_rlinf.py` — these use their own `GlobalHydra` instance and were never integrated with the typed-preset CLI. Adding support is a separate feature, not a regression.

## 5. Test plan

- [x] Repro confirmed on develop before the fix.
- [x] All 4 train scripts smoke-tested with `physics=newton_mjwarp` on Isaac-Ant-v0 — each reaches Newton sim init (no Hydra struct error). Verified via the unified dispatcher AND via direct library-script invocation.
- [x] All 4 play scripts smoke-tested with `physics=newton_mjwarp` + a fake `--checkpoint` — each resolves env_cfg + agent_cfg successfully and fails downstream at checkpoint loading (expected; proves the typed selector was accepted). Both invocation paths verified.
- [x] `--help` output includes `physics=NAME` / `renderer=NAME` / `presets=NAME[,NAME,...]` syntax with available variants per task. Verified via both invocation paths.
- [x] Invalid value (`physics=does_not_exist`) produces a clear error listing available presets.
- [x] Added `scripts/reinforcement_learning/test/test_typed_preset_cli_train_play.py` — parametrized subprocess test that drives `physics=does_not_exist` through the unified dispatcher for all 8 train/play entrypoints. Asserts the Hydra struct error is absent and the resolver's "Unknown preset(s)" error is present. ~20 s, no GPU. Verified to fail when any of the 8 scripts is reverted to develop's broken state.
- [x] Pre-commit clean.
- [ ] CI green on `Docker + Tests`.

### DIFF
--- a/scripts/reinforcement_learning/rl_games/play_rl_games.py
+++ b/scripts/reinforcement_learning/rl_games/play_rl_games.py
@@ -27,7 +27,14 @@ from isaaclab_rl.rl_games import RlGamesGpuEnv, RlGamesVecEnvWrapper
 from isaaclab_rl.utils.pretrained_checkpoint import get_published_pretrained_checkpoint
 
 import isaaclab_tasks  # noqa: F401
-from isaaclab_tasks.utils import add_launcher_args, get_checkpoint_path, launch_simulation, resolve_task_config
+from isaaclab_tasks.utils import (
+    add_launcher_args,
+    fold_preset_tokens,
+    get_checkpoint_path,
+    launch_simulation,
+    resolve_task_config,
+    setup_preset_cli,
+)
 
 # PLACEHOLDER: Extension template (do not remove this comment)
 with contextlib.suppress(ImportError):
@@ -59,12 +66,12 @@ parser.add_argument(
 )
 parser.add_argument("--real-time", action="store_true", default=False, help="Run in real-time, if possible.")
 add_launcher_args(parser)
-args_cli, hydra_args = parser.parse_known_args()
+args_cli, hydra_args = setup_preset_cli(parser)
 
 if args_cli.video:
     args_cli.enable_cameras = True
 
-sys.argv = [sys.argv[0]] + hydra_args
+sys.argv = [sys.argv[0]] + fold_preset_tokens(hydra_args)
 
 
 def main():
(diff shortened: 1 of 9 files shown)


kellyguo11 (approved):

kellyguo11:
thanks! would be good to also have some unit tests for these scripts

PR #5738, opened 2026-05-21
[CI] Enable multi-GPU pytest workflow

hujc7:
## 1. Summary

- Re-enables `on: pull_request` trigger in `.github/workflows/test-fabric-multi-gpu.yaml`. Single file changed, 5 insertions / 12 deletions.
- Unblocks the cuda:1-parameterized unit tests added in #5514 (`source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py`) — they are currently only reachable via `workflow_dispatch`.
- First end-to-end exercise of the new multi-GPU runner pool (`g6e.12xlarge`, label `[self-hosted, linux, x64, gpu, multi-gpu]`) provisioned 2026-05-21.

## 2. Why now

The trigger block was commented out at #5514 merge time because no self-hosted runner advertised the `multi-gpu` label; any `pull_request` event would have queued indefinitely. The infra precondition is now satisfied, so the original trigger can be restored verbatim.

## 3. Scope

- No test code changes. The cuda:1 tests, the `ISAACLAB_TEST_MULTI_GPU=1` env gate, the 2-GPU pre-flight check, and the runner label all stay as-is.
- This PR is the minimal "is the runner alive" check. If the workflow goes green, follow-ups can broaden coverage (path filter, additional `cuda:1` parameterized tests).

## 4. CI footprint

The diff only touches `.github/workflows/test-fabric-multi-gpu.yaml`, which is not in `build.yaml`'s `Detect Changes` allowlist. The standard L40 docker-test matrix self-skips with `run_docker_tests=false`, so iteration on this branch does not consume the production single-GPU pool. Only the new multi-GPU workflow runs.

## 5. Test plan

- [ ] Workflow appears in the checks list on this draft PR.
- [ ] 2-GPU pre-flight step passes (`torch.cuda.device_count() >= 2`).
- [ ] All three `cuda:1`-parameterized tests in `test_views_xform_prim_fabric.py` pass.
- [ ] If multi-GPU pool capacity is constrained (us-west-2d g6e.12xlarge shortage per Alexander 2026-05-21), confirm the job queues rather than fails and picks up once capacity returns.

### DIFF
--- a/.github/workflows/test-fabric-multi-gpu.yaml
+++ b/.github/workflows/test-fabric-multi-gpu.yaml
@@ -48,7 +48,12 @@ jobs:
         run: pip install cmake
 
       - name: Install Isaac Lab
-        run: ./isaaclab.sh --install
+        # ``--install none`` = core submodules only (no mimic, no teleop, no
+        # auto-extra features).  The FabricFrameView cuda:1 tests only need
+        # isaaclab core + isaaclab_physx + Isaac Sim, all of which are core.
+        # This avoids building robomimic's egl_probe wheel, which requires
+        # libEGL/X11 headers that the multi-GPU runner image lacks.
+        run: ./isaaclab.sh --install none
 
       - name: Verify multi-GPU availability
         # Fail loud if the runner regressed to a single GPU.  Without this, the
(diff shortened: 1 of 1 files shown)


.github/workflows/test-fabric-multi-gpu.yaml
```diff
@@ -37,13 +30,43 @@ jobs:
   test-fabric-multi-gpu:
     name: FabricFrameView multi-GPU unit tests
     runs-on: [self-hosted, linux, x64, gpu, multi-gpu]
```
nv-apoddubny: Please remove GPU tag

hujc7:
Moved to https://github.com/isaac-sim/IsaacLab/pull/5823

PR #5104, opened 2026-03-24
Fixes windows batch command inline comment syntax

sellai-youcef:
The Windows batch commands in the installation docs had inline comments 
using `::` which cause errors when copy-pasted:

isaaclab.bat --install :: or "isaaclab.bat -i"

The `::` is not valid as an inline comment in batch — it breaks the command.

Fix: removed the inline comments, keeping only the clean commands:

isaaclab.bat --install
isaaclab.bat --install rl_games

Fixes #5086

PR #5740, opened 2026-05-22
docs: fix Windows batch inline comment syntax and add shorthand command

### LINKED ISSUE #5086: [Proposal] Installation docs update for Windows pre-built binaries
### Proposal

To improve the Isaac Lab installation for **Windows** users, the installation guide for both Isaac Sim as well as Isaac Lab might need some fixes and consistency. 

### Motivation

The current setup process for Isaac Lab alongside Isaac Sim is quite difficult to follow. In particular, installing the **pre-built binaries** on Windows is error-prone and not clearly documented.

I have noticed especially following issues in the installation process:
- The [Isaac Sim installation guide](https://docs.isaacsim.omniverse.nvidia.com/5.1.0/installation/quick-install.html#quick-install) assumes installation at `C:\isaac-sim` (please notice the -) while the Isaac Lab instructions expects a symlink at `C:\isaacsim`. 

https://github.com/isaac-sim/IsaacLab/blob/f4aa17f87e2e5db5484f0b5974918573e8918ce2/docs/source/setup/installation/binaries_installation.rst?plain=1#L26-L34

(This inconsistency is resolved in the installation guide for the Isaac Sim 6.0.0 early developer release, but remains a critical issue for users installing currently stable versions 5.1.0 and below.)

- The listed Windows shell commands in the documentation are prone to failure when copy-pasted (e.g., with the clipboard functionality on the right side of the code cell). For example:
```shell
isaaclab.bat --install :: or "isaaclab.bat -i"
```
results in an error because the inline code comment will be treated as an extra command. The correct syntax should include `&` before the inline comment:

```shell
isaaclab.bat --install &:: or "isaaclab.bat -i"
```

### Alternatives

- In my opinion, it might be clearer to just simply use `isaaclab.bat --install` without the inline comment (which is redundant and might introduce confusion and copy-paste errors)

### Build Info

- Isaac Lab Version: 2.3.2
- Isaac Sim Version: 5.1.0

### Additional context

--

### Checklist

### Acceptance Criteria

--

sellai-youcef:
The Windows batch command in the installation docs had an inline 
comment using :: which breaks the command when
copy-pasted:

isaaclab.bat --install :: or "isaaclab.bat -i"

In Windows batch, :: only works at the start of a line and not inline.

Fix: removed the inline comment and listed the shorthand as a separate line instead:

isaaclab.bat --install
isaaclab.bat -i

This is an updated version of #5104, following @kellyguo11's feedback to list the shorthand command separately rather than removing it entirely.

Fixes #5086

### DIFF
--- a/docs/source/setup/installation/include/src_build_isaaclab.rst
+++ b/docs/source/setup/installation/include/src_build_isaaclab.rst
@@ -34,7 +34,8 @@ Installation
 
          .. code:: batch
 
-            isaaclab.bat --install :: or "isaaclab.bat -i"
+            isaaclab.bat --install
+            isaaclab.bat -i
 
 
    All core submodules are **always** installed regardless of what is passed to ``-i``.


StafaH:
This is now stale after the install docs re-write

PR #5750, opened 2026-05-22
Adapt quadcopter.py For Dual Backends

YizeWang:
Description to be updated to be compliant with IsaacLab conventions.

# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/scripts/demos/default_physics_cfg.py
+++ b/scripts/demos/default_physics_cfg.py
@@ -1,0 +1,21 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Default physics backend presets for the demo scripts."""
+
+from isaaclab_newton.physics import NewtonCfg
+from isaaclab_physx.physics import PhysxCfg
+
+from isaaclab.utils.configclass import configclass
+from isaaclab_tasks.utils.hydra import PresetCfg
+
+
+@configclass
+class DefaultPhysicsCfgs(PresetCfg):
+    """Default physics backend presets for the demo scripts."""
+
+    default: PhysxCfg = PhysxCfg()
+    physx: PhysxCfg = PhysxCfg()
+    newton_mjwarp: NewtonCfg = NewtonCfg()
--- a/scripts/demos/quadcopter.py
+++ b/scripts/demos/quadcopter.py
@@ -11,22 +11,29 @@ This script demonstrates how to simulate a quadcopter.
     # Usage
     ./isaaclab.sh -p scripts/demos/quadcopter.py
 
+    # Run with the Newton backend
+    ./isaaclab.sh -p scripts/demos/quadcopter.py physics=newton_mjwarp -visualizer=newton
+
 """
 
-"""Launch Isaac Sim Simulator first."""
+"""Parse command line arguments."""
 
 import argparse
-
+import sys
 from isaaclab.app import AppLauncher
+
+from isaaclab_tasks.utils import add_launcher_args,fold_preset_tokens,launch_simulation,setup_preset_cli
 
 # add argparse arguments
 parser = argparse.ArgumentParser(description="This script demonstrates how to simulate a quadcopter.")
 # append AppLauncher cli args
-AppLauncher.add_app_launcher_args(parser)
+add_launcher_args(parser)
 # demos should open Kit visualizer by default
 parser.set_defaults(visualizer=["kit"])
 # parse the arguments
-args_cli = parser.parse_args()
+args_cli, hydra_args = setup_preset_cli(parser)
+sys.argv = [sys.argv[0]] + fold_preset_tokens(hydra_args)
+
 
 # launch omniverse app
 app_launcher = AppLauncher(args_cli)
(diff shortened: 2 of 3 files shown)


scripts/demos/default_physics_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+
+from isaaclab_newton.physics import NewtonCfg
+from isaaclab_physx.physics import PhysxCfg
+
+from isaaclab.utils.configclass import configclass
+from isaaclab_tasks.utils.hydra import PresetCfg
+
+
+@configclass
+class DefaultPhysicsCfgs(PresetCfg):
```
YizeWang, line 16: Does it make sense to define this `DefaultPhysicsCfgs`? If so, can I move this file to the same directory as `hydra.py`?

scripts/demos/quadcopter.py
```diff
@@ -39,18 +46,24 @@
...
 
 
-def main():
-    """Main function."""
-    # Load kit helper
-    sim_cfg = sim_utils.SimulationCfg(dt=0.005, device=args_cli.device)
+def main() -> None:
+    """Run the quadcopter demo."""
+    sim_cfg = sim_utils.SimulationCfg(dt=0.005, device=args_cli.device, physics=DefaultPhysicsCfgs())
+    global_presets, preset_sel, *_ = parse_overrides(sys.argv[1:], {"env": collect_presets(sim_cfg)})
```
YizeWang, line 61: Any better way to do line 60 and 61?

scripts/demos/quadcopter.py
```diff
@@ -39,18 +46,24 @@
...
+    sim_cfg = resolve_presets(sim_cfg, {*global_presets, *(name for _, _, name in preset_sel)})
+
     sim = SimulationContext(sim_cfg)
+    # launch_simulation(sim, args_cli)
```
YizeWang, line 65: I saw `launch_simulation` was implemented recently. Should I use this API instead?

scripts/demos/quadcopter.py
```diff
@@ -85,7 +98,7 @@ def main():
     sim_time = 0.0
     count = 0
     # Simulate physics
-    while simulation_app.is_running():
+    while sim.has_alive_visualizers():
```
YizeWang, line 101: `simulation_app` seems to refer to kit. In case of Newton visualizer, we cannot use `simulation_app`. Does it make sense to use `has_alive_visualizers` here? Ideally, I want to check if both the simulation and visualizer are running. For example, if simulation crashes but visualizer is alive and idle, we should break.

Another issue is that, if the user passes `--visualizer none`, we will never enter the while loop.

scripts/demos/default_physics_cfg.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Default physics backend presets for the demo scripts."""
+
+from isaaclab_newton.physics import NewtonCfg
```
YizeWang, line 8: If the user did not install newton, this could be a problem. Should I move these imports into `DefaultPhysicsCfgs` and guard them with `try`?

scripts/demos/default_physics_cfg.py
```diff
@@ -0,0 +1,21 @@
...
+from isaaclab_tasks.utils.hydra import PresetCfg
+
+
+@configclass
+class DefaultPhysicsCfgs(PresetCfg):
+    """Default physics backend presets for the demo scripts."""
+
+    default: PhysxCfg = PhysxCfg()
+    physx: PhysxCfg = PhysxCfg()
+    newton_mjwarp: NewtonCfg = NewtonCfg()
```
YizeWang, line 21: I am not sure how the default `NewtonCfg` works in different scenarios. Do we expect large changes on this configure across different tests?

scripts/demos/quadcopter.py
```diff
@@ -39,18 +46,24 @@
...
+    """Run the quadcopter demo."""
+    sim_cfg = sim_utils.SimulationCfg(dt=0.005, device=args_cli.device, physics=DefaultPhysicsCfgs())
+    global_presets, preset_sel, *_ = parse_overrides(sys.argv[1:], {"env": collect_presets(sim_cfg)})
+    sim_cfg = resolve_presets(sim_cfg, {*global_presets, *(name for _, _, name in preset_sel)})
+
     sim = SimulationContext(sim_cfg)
+    # launch_simulation(sim, args_cli)
+
     # Set main camera
     sim.set_camera_view(eye=[0.5, 0.5, 1.0], target=[0.0, 0.0, 0.5])
```
YizeWang, line 68: The camera view seems not properly set, in both kit and newton. Any ideas?

scripts/demos/quadcopter.py
```diff
@@ -11,22 +11,29 @@
     # Usage
     ./isaaclab.sh -p scripts/demos/quadcopter.py
 
+    # Run with the Newton backend
+    ./isaaclab.sh -p scripts/demos/quadcopter.py physics=newton_mjwarp -visualizer=newton
```
hujc7, line 15: I might slightly prefer to put hydra at the end so split between command line argument vs hydra

scripts/demos/quadcopter.py
```diff
@@ -11,22 +11,29 @@
...
 # demos should open Kit visualizer by default
 parser.set_defaults(visualizer=["kit"])
 # parse the arguments
-args_cli = parser.parse_args()
+args_cli, hydra_args = setup_preset_cli(parser)
+sys.argv = [sys.argv[0]] + fold_preset_tokens(hydra_args)
+
 
 # launch omniverse app
 app_launcher = AppLauncher(args_cli)
```
hujc7, line 39: what are all the conditions that app launcher is needed? seems not needed when newton physics + viz is used?

scripts/demos/quadcopter.py
YizeWang: Agreed.

PR #5770, opened 2026-05-25
Decouple backend-specific cloning from InteractiveScene

ooctipus:
## Summary

This PR removes backend-specific cloning logic from `InteractiveScene` and
replaces the previous implicit `replicate_session_defaults` /
`replicate_session` machinery with three orthogonal, explicit primitives that
compose identically whether you drive cloning through `InteractiveScene` or
by hand in a `DirectRLEnv` or standalone script.

The result is that `InteractiveScene` no longer knows anything about USD vs.
PhysX vs. Newton — it just enters a `ReplicateSession` and lets each asset's
constructor register the backend(s) it needs. This is foundational for two
follow-on capabilities the project has wanted for a while: **flexible backend
cloning** and **skip-cloning workflows**.

### New core primitives in `isaaclab.cloner`

- **`REPLICATION_QUEUE`** — module-level list that asset constructors append
  `(cfg, BackendCtxCls)` pairs to via tiny per-backend helpers
  (`queue_usd_replication`, `queue_physx_replication`,
  `queue_newton_replication`). Backends are no longer special-cased inside
  `InteractiveScene`; each one self-registers.
- **`ClonePlan`** — self-contained dataclass describing the world layout
  (`sources`, `destinations`, `clone_mask`, `env_ids`, `positions`,
  `cfg_rows`). Stage-agnostic by design; the USD stage is now passed
  explicitly to consumers so the same plan can be replayed, inspected, or
  serialized.
- **`replicate(plan, *, stage)`** — free function that drains
  `REPLICATION_QUEUE` against a plan, groups queued cfgs by backend context
  class, runs each context in ascending `replicate_priority` order (physics
  before USD), publishes the plan to `SimulationContext`, and clears the
  queue. The queue is snapshotted and cleared up front so a backend failure
  cannot leak stale entries into the next call.
- **`ReplicateSession`** is now a thin context manager that calls
  `make_clone_plan` in `__enter__` and `replicate` in `__exit__`. The
  state-bag version with `plan` / `stage` / `cfg_rows` /
  `replicate_on_exit` fields is gone.
- **`ClonePlan.from_env_0`** — classmethod that builds the single-source
  homogeneous plan most direct envs need by auto-populating `cfg_rows`
  from `REPLICATION_QUEUE` filtered by env-root prefix.
- **`CloneCfg.clone_regex`** (default `"/World/envs/env_.*"`) — single
  source of truth for the env-namespace convention. `InteractiveScene`
  reads it directly when expanding `{ENV_REGEX_NS}` cfg macros.

### Two equivalent invocation paths

```python
# InteractiveScene path (what the scene runs under the hood)
with cloner.ReplicateSession(cfgs, num_clones=N, env_spacing=2.0,
                             device=device, stage=stage):
    for cfg in cfgs:
        cfg.class_type(cfg)

# Direct env / script path
plan = cloner.ClonePlan.from_env_0(src, dest, num_envs, device, positions)
cloner.replicate(plan, stage=scene.stage)
```

Both end in the same `cloner.replicate(plan, stage=...)` call. The only
difference is how the plan was built and how asset construction was
interleaved.

### What got removed from `InteractiveScene`

- `clone_environments(...)` deprecated shim. The scene now replicates
  inside `__init__` via `ReplicateSession`.
- `env_ns` / `env_regex_ns` properties (used only internally).
- `_build_clone_plan_from_cfg` and `_default_env_origins` internals.
  Cfg-driven plan construction now lives in `make_clone_plan`; per-env
  positions are read from the published `ClonePlan`.
- `InteractiveScene.env_origins` now reads from the plan published to
  `SimulationContext`, making the plan the single source of truth for
  env placement.

### Why this matters (the actual point of the PR)

This refactor is foundational for two capabilities the current scene
coupling blocks:

- **Flexible backend cloning.** Backends now plug in by shipping a
  `<Backend>ReplicateContext` class + a one-line queue helper. Swapping
  PhysX ↔ Newton no longer requires `InteractiveScene` to change; cfgs
  and user code stay untouched, and a third-party backend can register
  itself without modifying core.
- **Skip-cloning workflows.** Because plan construction, asset
  registration, and drain are three independent primitives, callers
  that want to author env-0 prims by hand and skip the cloner — or
  drive replication out-of-band from a visualizer, replay tool, or
  test fixture — can do so without fighting `InteractiveScene`.

## Migration notes

- `with cloner.ReplicateSession():` (no-arg) →
  `cloner.replicate(cloner.ClonePlan.from_env_0(...), stage=...)`.
- `InteractiveScene.clone_environments(...)` → removed; the scene
  replicates inside `__init__`.
- `make_clone_plan(sources, destinations, ...)` →
  `make_clone_plan(cfgs, num_clones, env_spacing, device, ...)`.
- Pass `stage=...` explicitly to `replicate()` and `ReplicateSession()`.
- Read `CloneCfg.clone_regex` if you previously used
  `InteractiveScene.env_ns` / `env_regex_ns`.

About 17 direct envs were migrated to the new pattern in this PR.

## Test plan

- [ ] `./isaaclab.sh -p -m pytest source/isaaclab/test/sim/test_cloner.py`
- [ ] `./isaaclab.sh -p -m pytest source/isaaclab/test/scene/test_interactive_scene.py`
- [ ] `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/sim/test_cloner.py`
- [ ] Migrated direct envs (`cartpole`, `anymal_c`, `franka_cabinet`,
      `factory`, `humanoid_amp`, `inhand_manipulation`, `locomotion`,
      `quadcopter`, `shadow_hand_*`, `automate/*`, `cart_double_pendulum`,
      `cartpole_warp`, `inhand_manipulation_warp`, `locomotion_warp`)
      spawn and step on PhysX
- [ ] Same envs spawn and step on Newton
- [ ] `./isaaclab.sh -f` passes
- [ ] Docs build (`cd docs && make html`)


docs/source/how-to/cloning.rst
```diff
@@ -5,347 +5,361 @@ Cloning Environments
...
 
-Isaac Lab creates many parallel environments by spawning representative source prims and
-then cloning them to the remaining environment paths. This guide starts with direct cloning
-so the primitive contract is clear, then shows how :class:`~isaaclab.cloner.ClonePlan` and
-:class:`~isaaclab.scene.InteractiveScene` build on top of that contract.
+Parallel simulation at scale needs many environments stepping side by side —
+hundreds, sometimes tens of thousands per GPU — and authoring each of those envs
+by hand would be hopelessly slow. Cloning is Isaac Lab's answer: you author a
+small representative scene under ``/World/envs/env_n`` and the cloner expands it
+across the rest of the env population for you, optionally with per-env variation.
```
StafaH, line 12: Might be worth adding the wording "heterogenous" when discussing per-env cloning variations (and maybe some details on how that work?)

docs/source/how-to/cloning.rst
```diff
@@ -5,347 +5,361 @@ Cloning Environments
...
+     - Bool tensor ``[len(sources), num_envs]``; ``True`` when env ``j`` comes from row ``i``.
+   * - ``env_ids``
+     - Long tensor of target env ids.
+   * - ``positions``
+     - Optional per-env world positions [m], shape ``[num_envs, 3]``.
+
+The plan is stage-agnostic by design — the same instance can be replayed against a
+different stage, inspected by tooling, or serialized.
+
+When every env is a copy of env_0:
```
StafaH, line 137: Use the terminology "homogenous" here and other places where possible.

docs/source/how-to/cloning.rst
```diff
@@ -5,347 +5,361 @@ Cloning Environments
...
-    destinations = [destination_0, destination_1, ...]
-    clone_mask   = bool tensor, shape [len(sources), num_envs]
+    sources      = ("/World/envs/env_0",)
+    destinations = ("/World/envs/env_{}",)
+    clone_mask   = [[True, True, ..., True]]
 
-``clone_mask[i, j]`` is ``True`` when environment ``j`` should receive source row ``i``.
-The same plan can be passed to USD replication, physics replication, and scene-data
-providers.
+When envs differ — say a cartpole in every env plus a 2-variant obstacle (box into
```
StafaH, line 145: This reads a little bit confusing. Maybe re-write to something like:
"We can build heterogenous envs through ..., as an example we can build clone a cartpole environment that has two variants, half the environments use a box obstacle, the other half use sphere"

docs/source/how-to/cloning.rst
```diff
@@ -5,347 +5,361 @@ Cloning Environments
...
+    clone_mask   = [[1, 1, 1, 1],
+                    [1, 1, 0, 0],
+                    [0, 0, 1, 1]]
+
+A plan is the *what*. Putting one together and handing it to the backends is
+the *how*, and Isaac Lab exposes three idiomatic ways to do that. All three end
+in the same ``cloner.replicate(plan, stage=...)`` call, so the choice between
+them is purely about ergonomics:
+
+* The first wraps both phases in a context manager and is what
```
StafaH, line 165: Seems duplicated since below is the breakdown for all three, we can just remove these bullet points and maybe list the 3 replication methods, and then jump into below.

docs/source/how-to/cloning.rst
```diff
@@ -5,347 +5,361 @@ Cloning Environments
...
+  scene config.
+* The second spells the same flow out as plain function calls, leaving a moment
+  between the build and the drain where you can inspect or mutate the plan.
+  Reach for it when you are assembling a scene outside
+  :class:`~isaaclab.scene.InteractiveScene` or want fine control over timing.
+* The third is a one-shot shortcut for the case where every env is just a copy
+  of env_0. Reach for it in :class:`~isaaclab.envs.DirectRLEnv` and standalone
+  scripts that hand-build the env-0 prototype prim by prim.
+
+``ReplicateSession``
```
StafaH, line 177: We should use 1. 2. 3. for each of the methods, so that users can follow easily

docs/source/how-to/cloning.rst
```diff
@@ -359,39 +373,4 @@ per-environment isolation is not already provided by the cloning backend:
...
-authored. Fabric-backed stage data must be inspected through USDRT rather than normal USD
-APIs.
-
-
-See Also
---------
-
-* :doc:`multi_asset_spawning` -- configuring multi-asset and multi-USD spawners.
-* :doc:`optimize_stage_creation` -- Fabric cloning and stage-in-memory optimizations.
+Newton isolates envs through its world system and does not need this pass.
```
StafaH, line 376: Newton envs are inherently isolated, they don't truly live in the same world, since state is batched.

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -12,22 +12,62 @@
...
+        subsequent :func:`~isaaclab.cloner.replicate` call.
+
+        Args:
+            source: Source prim path (typically ``/World/envs/env_0``).
+            destination: Destination template with ``"{}"`` for the env id.
+            num_clones: Number of target envs.
+            device: Torch device for the mask and env id buffers.
+            positions: Optional per-env world positions [m], shape ``[num_clones, 3]``.
+        """
+        from .replicate_session import REPLICATION_QUEUE  # noqa: PLC0415
```
StafaH, line 64: Why not import at top?

source/isaaclab/isaaclab/cloner/replicate_session.py
```diff
@@ -0,0 +1,135 @@
...
+        row_list = sorted(row_set)
+        ctx.queue_mapping(
+            [plan.sources[i] for i in row_list],
+            [plan.destinations[i] for i in row_list],
+            plan.env_ids,
+            plan.clone_mask[row_list],
+            positions=plan.positions,
+        )
+
+    for ctx in sorted(backend_ctxs.values(), key=lambda c: getattr(c, "replicate_priority", 0)):
```
StafaH, line 60: For env cloning, why is the sorting important?

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -144,209 +150,67 @@ def __init__(self, cfg: InteractiveSceneCfg):
...
-            # subtrees are layered on TOP of the physics-side ``physx.clone()``
-            # replicas -- PhysX is indifferent to additional USD content and
-            # the two layers don't conflict.  Probing whether this assumption
-            # holds in practice; revert to ``not startswith("ovphysx")`` if
-            # ``physx.clone()`` errors on already-populated targets.
-            clone_usd=True,
-        )
+        self.cloner_cfg = cloner.CloneCfg(device=self.device)
+        env_root = self.cloner_cfg.clone_regex.rsplit("/", 1)[0]
+        self.env_prim_paths = [f"{env_root}/env_{i}" for i in range(self.cfg.num_envs)]
```
StafaH, line 161: Is this still needed? Does not seem used anymore since clone plan uses the {} pattern

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/cartpole/cartpole_warp_env.py
```diff
@@ -258,8 +259,10 @@ def _setup_scene(self) -> None:
         self.cartpole = Articulation(self.cfg.robot_cfg)
         # add ground plane
         spawn_ground_plane(prim_path="/World/ground", cfg=GroundPlaneCfg())
-        # clone and replicate
-        self.scene.clone_environments(copy_from_source=False)
+        src, dest = "/World/envs/env_0", "/World/envs/env_{}"
```
StafaH, line 209: Should we create a helper called `cloner.homogenous_cloning_from_env_0(...)` that makes these 4 lines into 1 line? Or even just `cloner.homogenous_cloning`

StafaH (approved):

docs/source/how-to/cloning.rst
ooctipus: I dont know, I think I want to try avoid using the world heterogenous/homogeneous. Newton doesn't use it at all

docs/source/how-to/cloning.rst
ooctipus: for the same reason as mentioned above, but happy to be convinced otherwise

docs/source/how-to/cloning.rst
ooctipus: this is a good suggestion Ill try to incorporate that.

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: some ai sloppyness...

source/isaaclab/isaaclab/cloner/replicate_session.py
ooctipus: in the case there is usd_cloning, its usually cloned at last to avoid other engine reads the entire scene.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/cartpole/cartpole_warp_env.py
ooctipus: I still want to make string explicit enough that user sees key information from the, I have implemented cloner.homogenous_cloning_from_env_0(...), but ditched this option because it feels a bit too magical.

PR #5771, opened 2026-05-25
Support Demo Scripts With Multiple Backends

YizeWang:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -12,7 +12,7 @@ from contextlib import contextmanager
 
 
 @contextmanager
-def resolve_backend_and_visualizer(args, newton_cfg=None):
+def resolve_backend_and_visualizer(args, physx_cfg=None, newton_cfg=None):
     """Resolve physics + visualizer cfgs from ``--physics`` / ``--visualizer``.
 
     Yields ``(physics_cfg, visualizer_cfg)``. Kit is launched when required by
@@ -23,7 +23,7 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
     if args.physics == "physx":
         from isaaclab_physx.physics import PhysxCfg
 
-        physics_cfg = PhysxCfg()
+        physics_cfg = physx_cfg or PhysxCfg()
     elif args.physics == "newton_mjwarp":
         from isaaclab_newton.physics import MJWarpSolverCfg, NewtonCfg
 
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -14,7 +14,8 @@ from contextlib import contextmanager
 def resolve_backend_and_visualizer(args, newton_cfg=None):
     """Resolve physics + visualizer cfgs from ``--physics`` / ``--visualizer``.
 
-    Yields ``(physics_cfg, visualizer_cfg)``. Kit is closed automatically on exit.
+    Yields ``(physics_cfg, visualizer_cfg)``. Kit is launched when required by
+    either the visualizer or the physics backend, and closed automatically on exit.
     """
 
     # Resolve physics cfg
@@ -42,8 +43,9 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
         raise ValueError(f"Unsupported --physics value: {args.physics}")
 
     # Resolve visualizer cfg
-    DEFAULT_VISUALIZER = "kit"
-    viz_type = args.visualizer or DEFAULT_VISUALIZER
+    if not isinstance(args.visualizer, list) or len(args.visualizer) != 1:
+        raise ValueError("Demos support exactly one --visualizer value: kit or newton.")
+    viz_type = args.visualizer[0]
     if viz_type == "kit":
         from isaaclab_visualizers.kit import KitVisualizerCfg
 
@@ -55,9 +57,10 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
     else:
         raise ValueError(f"Unsupported --visualizer value: {viz_type}")
 
-    # Kit needs to be closed by explicitly calling AppLauncher.close()
+    # PhysX requires Isaac Sim Kit extensions even when rendering through a
+    # standalone visualizer such as Newton.
     close_fn = None
-    if viz_type == "kit":
+    if viz_type == "kit" or args.physics == "physx":
         from isaaclab.app import AppLauncher
 
         args.visualizer = [viz_type]
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -69,4 +69,5 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
 
 def has_no_alive_visualizer_window(sim) -> bool:
     """Check if there are no alive visualizer windows."""
-    return bool(sim.visualizers and not any(v.is_running() and not v.is_closed for v in sim.visualizers))
+    visualizers = sim.visualizers or ()
+    return not any(v.is_running() and not v.is_closed for v in visualizers)
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -7,6 +7,7 @@
 This script contains helper functions for the demos.
 """
 
+import textwrap
 from contextlib import contextmanager
 
 
@@ -78,3 +79,13 @@ def has_no_alive_visualizer_window(sim) -> bool:
     """Check if there are no alive visualizer windows."""
     visualizers = sim.visualizers or ()
     return not any(v.is_running() and not v.is_closed for v in visualizers)
+
+
+def get_usage_examples(doc: str) -> str:
+    """Return usage examples from the module docstring."""
+    marker = ".. code-block:: bash"
+    if marker not in doc:
+        return ""
+
+    examples = textwrap.dedent(doc.split(marker, maxsplit=1)[1]).strip()
+    return "Examples:\n" + textwrap.indent(examples, "  ")
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -20,9 +20,11 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
     # Resolve physics cfg
     if args.physics == "physx":
         from isaaclab_physx.physics import PhysxCfg
+
         physics_cfg = PhysxCfg()
     elif args.physics == "newton_mjwarp":
         from isaaclab_newton.physics import MJWarpSolverCfg, NewtonCfg
+
         DEFAULT_NEWTON_CFG = NewtonCfg(
             solver_cfg=MJWarpSolverCfg(
                 njmax=70,
@@ -44,9 +46,11 @@ def resolve_backend_and_visualizer(args, newton_cfg=None):
     viz_type = args.visualizer or DEFAULT_VISUALIZER
     if viz_type == "kit":
         from isaaclab_visualizers.kit import KitVisualizerCfg
+
         visualizer_cfg = KitVisualizerCfg()
     elif viz_type == "newton":
         from isaaclab_visualizers.newton import NewtonVisualizerCfg
+
         visualizer_cfg = NewtonVisualizerCfg()
     else:
         raise ValueError(f"Unsupported --visualizer value: {viz_type}")
--- a/scripts/demos/demo_helper.py
+++ b/scripts/demos/demo_helper.py
@@ -1,0 +1,72 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+This script resolves the physics + visualizer cfgs from ``--physics`` / ``--visualizer``.
+"""
+
+from contextlib import contextmanager
+
+
+@contextmanager
+def resolve_backend_and_visualizer(args, newton_cfg=None):
+    """Resolve physics + visualizer cfgs from ``--physics`` / ``--visualizer``.
+
+    Yields ``(physics_cfg, visualizer_cfg)``. Kit is closed automatically on exit.
+    """
+
+    # Resolve physics cfg
+    if args.physics == "physx":
+        from isaaclab_physx.physics import PhysxCfg
+        physics_cfg = PhysxCfg()
+    elif args.physics == "newton_mjwarp":
+        from isaaclab_newton.physics import MJWarpSolverCfg, NewtonCfg
+        DEFAULT_NEWTON_CFG = NewtonCfg(
+            solver_cfg=MJWarpSolverCfg(
+                njmax=70,
+                nconmax=70,
+                ls_iterations=40,
+                cone="elliptic",
+                impratio=100,
+                ls_parallel=False,
+                integrator="implicitfast",
+            ),
+            num_substeps=2,
+        )
+        physics_cfg = newton_cfg or DEFAULT_NEWTON_CFG
+    else:
+        raise ValueError(f"Unsupported --physics value: {args.physics}")
+
+    # Resolve visualizer cfg
+    DEFAULT_VISUALIZER = "kit"
+    viz_type = args.visualizer or DEFAULT_VISUALIZER
+    if viz_type == "kit":
+        from isaaclab_visualizers.kit import KitVisualizerCfg
+        visualizer_cfg = KitVisualizerCfg()
+    elif viz_type == "newton":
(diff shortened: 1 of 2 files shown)


scripts/demos/quadrupeds.py
```diff
@@ -65,6 +61,12 @@ def define_origins(num_origins: int, spacing: float) -> list[list[float]]:
 
 def design_scene() -> tuple[dict, list[list[float]]]:
     """Designs the scene."""
+    import isaaclab.sim as sim_utils
+
+    from isaaclab_assets.robots.anymal import ANYMAL_B_CFG, ANYMAL_C_CFG, ANYMAL_D_CFG  # isort:skip
+    from isaaclab_assets.robots.spot import SPOT_CFG  # isort:skip
+    from isaaclab_assets.robots.unitree import UNITREE_A1_CFG, UNITREE_GO1_CFG, UNITREE_GO2_CFG  # isort:skip
```
hujc7, line 77: nit: what might be specific reason to change the import location?

scripts/demos/quadrupeds.py
```diff
@@ -124,14 +132,17 @@ def design_scene() -> tuple[dict, list[list[float]]]:
...
     # Define simulation stepping
     sim_dt = sim.get_physics_dt()
     sim_time = 0.0
     count = 0
     # Simulate physics
-    while simulation_app.is_running():
+    while True:
+        # Exit when every visualizer window has been closed (works for kit and newton)
+        if has_no_alive_visualizer_window(sim):
+            break
```
hujc7: Can this be used as loop condition?

scripts/demos/quadrupeds.py
```diff
@@ -124,14 +132,17 @@ def design_scene() -> tuple[dict, list[list[float]]]:
     return scene_entities, origins
 
 
-def run_simulator(sim: sim_utils.SimulationContext, entities: dict[str, Articulation], origins: torch.Tensor):
```
hujc7, line 127: Any reason to remove the type hint here?

scripts/demos/quadrupeds.py
```diff
@@ -79,37 +81,43 @@ def design_scene() -> tuple[dict, list[list[float]]]:
     # Origin 1 with Anymal B
     sim_utils.create_prim("/World/Origin1", "Xform", translation=origins[0])
     # -- Robot
-    anymal_b = Articulation(ANYMAL_B_CFG.replace(prim_path="/World/Origin1/Robot"))
```
hujc7, line 82: I thought if simulation context is set, articulation does the dispatch. Did that work?

scripts/demos/quadrupeds.py
```diff
@@ -175,24 +186,27 @@ def run_simulator(sim: sim_utils.SimulationContext, entities: dict[str, Articula
...
-    sim.reset()
-    # Now we are ready!
-    print("[INFO]: Setup complete...")
-    # Run the simulator
-    run_simulator(sim, scene_entities, scene_origins)
+    with resolve_backend_and_visualizer(args_cli) as (physics_cfg, visualizer_cfg):
+        import isaaclab.sim as sim_utils
+
+        # define simulation time step
+        dt = 1 / 200
```
hujc7: nit: can fold into dt below.

hujc7 (commented):
Maybe need to add some smoke test generally for demos?

scripts/demos/quadrupeds.py
YizeWang: Reverted, as discussed in https://nvidia.slack.com/archives/C0AV4B102HG/p1779865797697429?thread_ts=1779368084.742639&cid=C0AV4B102HG.

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,72 @@
...
+    try:
+        yield physics_cfg, visualizer_cfg
+    finally:
+        # No-op for the newton visualizer
+        if close_fn is not None:
+            close_fn()
+
+
+def has_no_alive_visualizer_window(sim) -> bool:
+    return bool(sim.visualizers and not any(v.is_running() and not v.is_closed for v in sim.visualizers))
```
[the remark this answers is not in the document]
YizeWang: Resolved in the "Resolve Comments" commit.

scripts/demos/quadrupeds.py
YizeWang: Resolved in the "Resolve Comments" commit.

scripts/demos/quadrupeds.py
YizeWang: Sure. Resolved in the "Resolve Comments" commit.

scripts/demos/quadrupeds.py
YizeWang: Resolved in the "Fold dt Into Function Call" commit.

scripts/demos/quadrupeds.py
```diff
@@ -8,41 +8,40 @@
...
     ./isaaclab.sh -p scripts/demos/quadrupeds.py
 
+    # Usage with the kit-less Newton (MJWarp) backend (launches Isaac Sim Kit by default).
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics newton_mjwarp
+
+    # Usage with the Newton (MJWarp) backend without Kit (launches Newton visualizer).
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics newton_mjwarp --visualizer newton
+
+    # Usage with the PhysX backend without Kit (launches Newton visualizer).
+    # TODO(yizew@nvidia.com): Not supported yet. Investigation needed.
```
YizeWang: Is `physx` + `newton` expected to work by design? `physx` seems dependent on `IsaacSim`. I need to properly resolve this dependency if we want this pair to work.

scripts/demos/quadrupeds.py
```diff
@@ -8,41 +8,40 @@
...
 
 # add argparse arguments
-parser = argparse.ArgumentParser(description="This script demonstrates different legged robots.")
+parser = argparse.ArgumentParser(description="This script demonstrates different legged robots.", conflict_handler="resolve")
 # append AppLauncher cli args
 AppLauncher.add_app_launcher_args(parser)
-# demos should open Kit visualizer by default
-parser.set_defaults(visualizer=["kit"])
+parser.add_argument("--physics", default="physx", choices=["physx", "newton_mjwarp"], help="Physics backend.")
+parser.add_argument("--visualizer", default="kit", choices=["kit", "newton"], help="Visualizer backend.")
```
YizeWang: `--visualizer` is added again to limit options inside `["kit", "newton"]`, to avoid dealing with `rerun` or `vizer` and to minimize the branches in `demo_helper.py`. Please let me know if these 2 visualizers are also supposed to be supported.

scripts/demos/quadrupeds.py
```diff
@@ -8,41 +8,40 @@
...
-app_launcher = AppLauncher(args_cli)
-simulation_app = app_launcher.app
-
-"""Rest everything follows."""
-
 import numpy as np
 import torch
 
-import isaaclab.sim as sim_utils
-from isaaclab.assets import Articulation
```
YizeWang, line 41: [Correct me if I am wrong] `isaaclab.sim` `Articulation` must be imported after `AppLauncher` if `kit` is the chosen visualizer. At this point, we are not sure if we want to launch `AppLauncher` yet and thus we need to move these imports below.

One alternative way could be, determine whether `kit` is to be launched here and do the same as before.
```
if kit is the visualizer:
  app_launcher = AppLauncher(args_cli)
  simulation_app = app_launcher.app
```
By doing so, we can keep these imports as where they were. However, this would requires the `simulation_app` to be passed into `resolve_backend_and_visualizer` so that we refer to the same instance.

scripts/demos/quadrupeds.py
YizeWang: Same reason as my last comment: https://github.com/isaac-sim/IsaacLab/pull/5771/changes#r3310474409. Let me know if my understanding in the previous comment is wrong.

scripts/demos/quadrupeds.py
YizeWang: Same reason as my last comment: https://github.com/isaac-sim/IsaacLab/pull/5771/changes#r3310474409. These types are not defined yet (these packages not imported). Let me know if my understanding in the previous comment is wrong.

scripts/demos/quadrupeds.py
hujc7: Not entirely sure. Is there a specific problem with the import?

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,77 @@
...
+                integrator="implicitfast",
+            ),
+            num_substeps=2,
+        )
+        physics_cfg = newton_cfg or DEFAULT_NEWTON_CFG
+    else:
+        raise ValueError(f"Unsupported --physics value: {args.physics}")
+
+    # Resolve visualizer cfg
+    DEFAULT_VISUALIZER = "kit"
```
hujc7: this seems redundant if default is already provided in argpase.

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,77 @@
...
+    elif viz_type == "newton":
+        from isaaclab_visualizers.newton import NewtonVisualizerCfg
+
+        visualizer_cfg = NewtonVisualizerCfg()
+    else:
+        raise ValueError(f"Unsupported --visualizer value: {viz_type}")
+
+    # Kit needs to be closed by explicitly calling AppLauncher.close()
+    close_fn = None
+    if viz_type == "kit":
```
hujc7: physx should also need kit?

scripts/demos/quadrupeds.py
hujc7: Was it just the kit launch? or there are other issues?

hujc7: nit: ideally this help message should also show up when run with -h/--help

scripts/demos/demo_helper.py
YizeWang: Resolved in the "Support Physx+NewtonViz" commit.

scripts/demos/demo_helper.py
YizeWang: Simplified the code in the "Support Physx+NewtonViz" commit.

scripts/demos/quadrupeds.py
YizeWang: Added help messages in the "Append Usage Examples to Help Menu" commit.

scripts/demos/quadrupeds.py
YizeWang: The failure shown in the terminal is not straight-forward.

Codex gives the following analysis: _"PhysxCfg / visualizer config imports can pull in pxr before SimulationApp starts. Kit then tries to initialize its own USD Python bindings and sees partially initialized/foreign pxr modules, which leads to the cascade of TfNotice, UsdAPISchemaBase, GfVec3f, and omni.usd failures."_

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+This script contains helper functions for the demos.
+"""
+
+import textwrap
+from contextlib import contextmanager
+
+
+@contextmanager
+def resolve_backend_and_visualizer(args, newton_cfg=None):
```
YizeWang: The option is limited inside `["kit", "newton"]`, to avoid dealing with `rerun` or `vizer `and to minimize the branches in `demo_helper.py`. Please let me know if these 2 visualizers are also supposed to be supported.

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,91 @@
...
+            num_substeps=2,
+        )
+        physics_cfg = newton_cfg or DEFAULT_NEWTON_CFG
+    else:
+        raise ValueError(f"Unsupported --physics value: {args.physics}")
+
+    # Resolve visualizer cfg
+    if not isinstance(args.visualizer, list) or len(args.visualizer) != 1:
+        raise ValueError("Demos support exactly one --visualizer value: kit or newton.")
+    viz_type = args.visualizer[0]
```
kellyguo11, line 49: why do we only parse the first element? is there a limitation to supporting multiple visualizers like the training scripts?

scripts/demos/demo_helper.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Resolve physics cfg
+    if args.physics == "physx":
+        from isaaclab_physx.physics import PhysxCfg
+
+        physics_cfg = PhysxCfg()
+    elif args.physics == "newton_mjwarp":
+        from isaaclab_newton.physics import MJWarpSolverCfg, NewtonCfg
+
+        DEFAULT_NEWTON_CFG = NewtonCfg(
```
kellyguo11, line 30: how generalizable is this cfg? do we expect this to work for most scripts? if not, it may be better to just keep this as a default MjWarpSolverCfg() object.

scripts/demos/demo_helper.py
kellyguo11: I think it'd be nice to also support rerun and viser so that we can be consistent across all the scripts

scripts/demos/quadrupeds.py
```diff
@@ -124,14 +135,15 @@ def design_scene() -> tuple[dict, list[list[float]]]:
...
+def run_simulator(sim, entities: dict, origins):
     """Runs the simulation loop."""
     # Define simulation stepping
     sim_dt = sim.get_physics_dt()
     sim_time = 0.0
     count = 0
     # Simulate physics
-    while simulation_app.is_running():
+    # Exit when every visualizer window has been closed (works for kit and newton)
+    while not has_no_alive_visualizer_window(sim):
```
kellyguo11, line 146: the double negative can be a bit confusing. would it make sense to make the utility has_alive_visualizer_window instead?

scripts/demos/quadrupeds.py
kellyguo11: yes i think we do need to handle some of the imports after running AppLauncher, but I'm not a big fan of having to move all the imports to be within design_scene. Is there a better way to handle the kit app lifecycle instead of wrapping everything in a context?

can we try splitting up the resolve_backend_and_visualizer into a more functional approach where we first resolve the cfgs, then launch the backend, and then we can handle all the other imports the same way as before?

I think it's actually also ok to not squeeze the kit app launcher code into the utils. so each script can handle the app launcher directly similar to previous workflows, something like

```
if requires_kit():
    simulation_app = AppLauncher(args_cli).app
else:
    simulation_app = None
```

scripts/demos/quadrupeds.py
kellyguo11: I think we can use `from __future__ import annotations` and type checking for this. shouldn't be necessary to remove the type hints

hujc7:
@greptileai review

scripts/demos/quadrupeds.py
```diff
@@ -8,45 +8,50 @@
...
+    # Usage with the kit-less Newton (MJWarp) backend (launches Isaac Sim Kit by default).
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics newton_mjwarp
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics newton_mjwarp --visualizer kit
+
+    # Usage with the Newton (MJWarp) backend without Kit (launches Newton visualizer).
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics newton_mjwarp --visualizer newton
+
+    # Usage with the PhysX backend and Newton visualizer.
+    # PhysX still launches Isaac Sim Kit headless because it depends on Kit extensions.
+    ./isaaclab.sh -p scripts/demos/quadrupeds.py --physics physx --visualizer newton
```
hujc7, line 26: this could be simplified a bit to avoid duplicates

scripts/demos/demo_helper.py
YizeWang: This currently works for most scripts while the default one barely works.

scripts/demos/demo_helper.py
YizeWang: I did not that we expect multiple visualizer. Let me take a look at training scripts and get back to you.

scripts/demos/quadrupeds.py
YizeWang: Okay. Tried to exhaust all combinations. I can reduce to the minimal.

scripts/demos/quadrupeds.py
YizeWang: Will update this.

scripts/demos/quadrupeds.py
YizeWang: Sure.

PR #4551, opened 2026-02-05
Migrating assets from torch to warp

AntoineRichard:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

Issue #289, opened 2024-03-15
Using assets from Isaac sim (USD)


SantiDiazC:
Hi!
currently I want to create a simulation with a UR5 robot arm and I have the urdf file so I think I can use the urdf importer.
However, I want to know how to import some of the assets that come with Isaac sim like the robot arms, warehouse etc. 

Thank you in advance!

Mayankm96:
Plase check the how-to guide for this information: https://isaac-orbit.github.io/orbit/source/how-to/index.html

PR #5776, opened 2026-05-26
[Fix] add_new_robot tutorial wheel-velocity action on GPU device

hujc7:
## 1. Summary

- ``scripts/tutorials/01_assets/add_new_robot.py`` errored at runtime: the wheel-velocity action was a CPU ``torch.Tensor`` and ``set_joint_velocity_target`` dispatches to a GPU Warp kernel that rejects CPU tensors at type check.
- Allocate the two wheel-velocity templates once on ``sim.device``, reuse across the loop, and call ``set_joint_velocity_target_index`` (also silences the deprecation warning visible in the bug report's first line).
- Scope: tutorial only. 1 file, +12 −3 plus a changelog fragment.

## 2. Why it broke

- The tutorial was added in #289 (2025-03-12) with ``torch.Tensor([[10.0, 10.0]])`` — CPU literal. The asset API at the time accepted CPU torch tensors.
- #4551 (2026-02-13) migrated the assets layer from torch to Warp kernels; CPU torch tensors silently became unsupported because Warp's kernel-argument check requires GPU-accessible data.
- #4911 (2026-03-10) followed up with a sweep that fixed several tutorials. It touched the reset block of this file (added ``wp.to_torch`` and ``_index`` variants for ``write_*_to_sim``), but did not touch the per-step action block — that remained CPU.

The bug has therefore been latent since #4551 and surfaced now that more users are running the tutorial on v3.0.0-beta.

## 3. Why only this tutorial

A grep across ``scripts/tutorials/**.py`` for tensor literals fed into ``set_joint_*_target*`` / ``write_*_to_sim_index`` shows ``add_new_robot.py`` is the only file constructing tensors without a ``device=`` kwarg. Every other tutorial either reads from a ``.torch`` accessor (already on device) or uses ``torch.tensor(..., device=sim.device)``.

## 4. Test plan

- [x] Reproduced the customer's error in isolation: CPU ``torch.Tensor`` → ``wp.launch`` → ``argument 'in_data' expects an array of type array(ndim=2, dtype=float32), but passed value has type Tensor``.
- [x] Same launch with a GPU tensor (``.cuda()``) succeeds.
- [x] Pre-commit clean on changed files.
- [ ] Manual smoke: ``./isaaclab.sh -p scripts/tutorials/01_assets/add_new_robot.py --num_envs 1 --headless`` runs through several reset cycles.

## 5. Out of scope

- The underlying API in ``isaaclab_physx`` silently accepts CPU torch tensors at the Python boundary and only fails inside ``wp.launch``. Auto-moving CPU tensors to ``self.device`` in ``set_joint_*_target_index`` would prevent users from hitting this from other scripts. Not in scope here per the bug-report preference for a tutorial-only fix; flagging for follow-up.

Refs NVBug 6156303

### DIFF
--- a/scripts/tutorials/01_assets/add_new_robot.py
+++ b/scripts/tutorials/01_assets/add_new_robot.py
@@ -103,6 +103,11 @@ def run_simulator(sim: sim_utils.SimulationContext, scene: InteractiveScene):
     sim_time = 0.0
     count = 0
 
+    # wheel-velocity templates allocated once on the simulation device; the joint
+    # target setters dispatch to GPU Warp kernels and reject CPU tensors.
+    straight_action = torch.tensor([[10.0, 10.0]], device=sim.device)
+    turn_action = torch.tensor([[5.0, -5.0]], device=sim.device)
+
     while simulation_app.is_running():
         # reset
         if count % 500 == 0:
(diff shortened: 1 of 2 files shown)


kellyguo11 (approved):

kellyguo11:
please also create a cherry-pick PR to the release branch

PR #5780, opened 2026-05-26
Install ovphysx wheel by default in `--install`

AntoineRichard:
# Description

Promote the publicly available `ovphysx` PyPI wheel out of the manual `ov[ovphysx]` extra and into the default `./isaaclab.sh --install` flow, mirroring how the newton wheel is treated. The OVRTX renderer wheel stays opt-in via `--install 'ov[ovrtx]'` or `--install 'ov[all]'`.

Concretely:

- `MANUAL_EXTRA_FEATURES` drops `"ov"` (it now only contains `"contrib"`), so the `ov` feature is part of the automatic `-i` / `-i all` set alongside `newton`, `rl`, and `visualizer`.
- `_install_ov_extra_dependencies("")` no longer prints a help message and returns; it now installs `isaaclab_ovphysx[ovphysx]`. Explicit `ov[ovrtx]`, `ov[ovphysx]`, and `ov[all]` behavior is unchanged.
- CLI `--install` help text and the `command_install` docstring are updated to reflect the new default.
- With the wheel guaranteed in every standard install, the `pytest.importorskip("ovphysx.types", reason="ovphysx wheel not installed")` guards (9 sites) and one inner `importorskip("isaaclab_ovphysx.tensor_types")` call are removed from the `isaaclab_ovphysx` test suite. The `isaaclab_ov` CI job (whose `filter-pattern: "isaaclab_ov"` already collects these tests) now exercises them for real instead of silently skipping.

No new required dependency is added; `source/isaaclab_ovphysx/setup.py` already declared `EXTRAS_REQUIRE = {"ovphysx": ["ovphysx"]}`. The extra remains unpinned for now — pinning a minimum version is a sensible follow-up once we lock in a release.

Fixes # (n/a)

## Type of change

Strictly, the default install footprint grows by one PyPI wheel and previously-skipped tests now run, so downstream consumers of `--install` see new behavior — but no public API is removed or renamed and all existing `ov[...]` selectors continue to work as before.

## Screenshots

N/A — install-CLI and test-gating change.

## Checklist

- [x] I have made corresponding changes to the documentation (CLI `--help` text and `command_install` docstring updated)
- [x] I have added tests that prove my fix is effective or that my feature works (new `test_all_installs_ov_with_default_ovphysx_selector` plus updates to `test_manual_extra_features` and `test_all_does_not_install_manual_extra_dependencies`; 51/51 pass in `test_install_command_parsing.py`)
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package — `source/isaaclab/changelog.d/antoiner-ovphysx-default-install.rst` and `source/isaaclab_ovphysx/changelog.d/antoiner-ovphysx-default-install.rst`


kellyguo11 (commented):
is making ovphysx installable by default mostly for the CI tests? I feel like it could be kept as an optional dependency from user perspective, but we can force an install of it for our CI jobs?

AntoineRichard:
> is making ovphysx installable by default mostly for the CI tests? I feel like it could be kept as an optional dependency from user perspective, but we can force an install of it for our CI jobs?

Yes it's mostly for CI tests. Sure we could leave that optional and handle it similarly to OvRTX in the CI.

AntoineRichard:
@myurasov-nv @hujc7 any opinions on the auto split cpu/gpu device.

hujc7:
> @myurasov-nv @hujc7 any opinions on the auto split cpu/gpu device.

This is great. The only concern is that it sits more in the test logic as well as the retries, which is not always obvious at test level (I think it's mostly for the process lock issue?). I also have something in https://github.com/isaac-sim/IsaacLab/pull/5823 which more relies on `@pytest.mark.parametrize("device", test_devices())`. 

Btw, I saw ovphysx has newer version. Is the process-global device lock still a problem?

AntoineRichard:
> > @myurasov-nv @hujc7 any opinions on the auto split cpu/gpu device.
> 
> This is great. The only concern is that it sits more in the test logic as well as the retries, which is not always obvious at test level (I think it's mostly for the process lock issue?). I also have something in #5823 which more relies on `@pytest.mark.parametrize("device", test_devices())`.
> 
> Btw, I saw ovphysx has newer version. Is the process-global device lock still a problem?

Thanks for taking a look! About 5823 the difference here is the device lock, we cannot run the cpu and gpu tests in the same run, we need to launch two separate test instances to test on all the devices. This is why it sits there. It could also be hard-coded, but that sounds like a bad idea?

Yeah, the new version will still have the device lock. I don't think they are planning on removing the lock at all.

kellyguo11 (approved):

source/isaaclab_ovphysx/changelog.d/antoiner-ovphysx-device-split-ci.rst
```diff
@@ -0,0 +1,9 @@
+Fixed
+^^^^^
+
+* Re-enabled both CPU and GPU coverage in CI for
+  :file:`test/sim/test_views_xform_prim_ovphysx.py` and
+  :file:`test/sensors/test_contact_sensor.py` by tagging them with the new
```
marcodiiga: nit: some more files were tagged with that

tools/conftest.py
```diff
@@ -400,282 +402,393 @@ def _retry_failed_test_in_fresh_process(
...
+
+    for test_file in test_files:
+        print(f"\n\n🚀 Running {test_file} independently...\n")
+        file_name = os.path.basename(test_file)
+        env = os.environ.copy()
+        env["PYTHONFAULTHANDLER"] = "1"
+
+        timeout = test_settings.PER_TEST_TIMEOUTS.get(file_name, test_settings.DEFAULT_TIMEOUT)
+
+        # Read the test file once for cold-cache and device-split detection.
```
marcodiiga, line 747: I think test_content  is read below and also in `is_device_split_file`, could be worth passing the contents to save one read

.github/workflows/build.yaml
```diff
@@ -543,6 +543,7 @@ jobs:
         isaacsim-base-image: ${{ needs.config.outputs.isaacsim_image_name }}
         isaacsim-version: ${{ needs.config.outputs.isaacsim_image_tag }}
         filter-pattern: "isaaclab_ov"
+        extra-pip-packages: "ovrtx ovphysx"
```
marcodiiga, line 546: uhm the installs could fail but CI might keep going with those `importorskip`, no? and quietly telling a green outcome. Just confirming this is the intended behavior.

source/isaaclab_ovphysx/changelog.d/antoiner-ovphysx-device-split-ci.rst
AntoineRichard: Updated in c746eaa: the changelog now lists all five OVPhysX test files tagged with device_split.

tools/conftest.py
AntoineRichard: Done in c746eaa: is_device_split_file now accepts preloaded source text, and tools/conftest.py passes the test_content it already read. Added a unit test for that path.

.github/workflows/build.yaml
AntoineRichard: Confirmed: the extra package install runs inside the Docker test command under set -e, and .github/actions/run-tests/action.yml returns the container exit code after copying reports. So if pip install ovphysx fails, the job fails before pytest/importorskip can produce a green result. The importorskip guards are only for local or partial installs where the optional wheel was not requested.

PR #5785, opened 2026-05-26
Clean up sub module packaging and remove setup.py

StafaH:
# Description

Migrates all IsaacLab packages from `setup.py` to `pyproject.toml` as the sole build system declaration, and introduces `uv` as the recommended package manager for development installs.

**What changed:**

- Removed `setup.py` from all 15 source packages (`isaaclab`, `isaaclab_assets`, `isaaclab_contrib`, `isaaclab_experimental`, `isaaclab_mimic`, `isaaclab_newton`, `isaaclab_ov`, `isaaclab_ovphysx`, `isaaclab_physx`, `isaaclab_ppisp`, `isaaclab_rl`, `isaaclab_tasks`, `isaaclab_tasks_experimental`, `isaaclab_teleop`, `isaaclab_visualizers`). All metadata (dependencies, extras, entry points, classifiers) is now declared exclusively in each package's `pyproject.toml`.
- Added a root `pyproject.toml` (`isaaclab-dev`) that wires all source packages together as editable path dependencies via `[tool.uv.sources]`, enabling a single `uv sync` to install the full development environment. `isaaclab-ppisp` is included in both the default dependencies and the `[all]` extra so that `pip install isaaclab[all]` and `uv sync` install the same set of packages.
- Replaced the third-party `toml` import in `simulation_context.py` and its test with stdlib `tomllib` (available since Python 3.11), removing an undeclared dependency. Module-level constants (`ISAACLAB_EXT_DIR`, `ISAACLAB_METADATA`, `ISAACLAB_ASSETS_EXT_DIR`, etc.) are preserved and continue to read from `config/extension.toml` via `tomllib`.
- Normalized extra names in `isaaclab_rl` to their PEP 685 canonical hyphenated forms — the underscore aliases `rsl_rl` and `rl_games` are removed. **Migration:** use `pip install isaaclab-rl[rsl-rl]` and `pip install isaaclab-rl[rl-games]` going forward.
- Fixed `isaaclab_ovphysx/pyproject.toml` declaring `requires-python = ">=3.11"` — corrected to `>=3.12` to match the rest of the project.
- Added a version-parity assertion to `tools/changelog/cli.py`: `current_version()` now raises immediately if `config/extension.toml` and `pyproject.toml` disagree, preventing silent version drift from corrupting future changelog bumps.
- Simplified `__init__.py` version detection across packages — the `importlib.metadata` lookup is now a clean try/except with a `"0.0.0"` fallback, removing boilerplate that was previously spread across every package.
- Moved test dependencies (`pytest`, `pytest-mock`, `junitparser`, `flatdict`, `flaky`) out of base `install_requires` and into a new `test` optional extra (`pip install isaaclab[test]`). The `run-tests` CI action now installs these unconditionally before invoking pytest, fixing the `No module named pytest` failure that occurred when running against the Isaac Sim bundled Python.
- Fixed the `isaacsim` and `all` extras in `source/isaaclab/pyproject.toml` pinning `isaacsim==5.1.0` — updated to `>=6.0.0` to match the version targeted by `isaaclab.sh -i` and `install.py`.
- Updated the `install.py` CLI command and benchmark test utilities to match the new package naming conventions.

Fixes # (issue)

## Type of change

## Breaking changes

- `isaaclab_rl[rsl_rl]` and `isaaclab_rl[rl_games]` extras no longer exist. Use `isaaclab_rl[rsl-rl]` and `isaaclab_rl[rl-games]` respectively.

## Screenshots

N/A — build system change only.

## Checklist


source/isaaclab/pyproject.toml
```diff
@@ -1,3 +1,103 @@
...
+]
+
+[project.urls]
+Homepage = "https://github.com/isaac-sim/IsaacLab"
+Repository = "https://github.com/isaac-sim/IsaacLab"
+
+[project.scripts]
+isaaclab = "isaaclab.cli:cli"
+play = "isaaclab.cli:play"
+train = "isaaclab.cli:train"
```
diegoferigo-rai, line 70: Would it be possible to use less generic names for these executables? In larger virtual environments, binaries like `play` and `train` could both confuse users and potentially clash with executables provided by other packages.

StafaH: We like the simple naming, what would be the alternative? `isaaclab_train`?

ooctipus (approved):

source/isaaclab/pyproject.toml
diegoferigo-rai: Let's consider the extreme case where IsaacLab eventually becomes part of the official Ubuntu repositories as a `python3-isaaclab` deb package. In that scenario, users would end up with generic `play` and `train` executables in their `PATH`, which would be highly ambiguous and likely conflict-prone.

While this may sound hypothetical, it is already a practical concern in ecosystems such as conda-forge. Those environments can grow to the scale of a Linux distribution and may contain IsaacLab alongside other frameworks like mjlab and many additional ML/RL toolchains. Generic executable names do not scale well in such shared environments.

I would strongly recommend addressing this upstream rather than leaving it to downstream packagers. Otherwise, distributions will likely introduce their own ad-hoc renaming conventions, leading to fragmentation and inconsistent UX across ecosystems.

Two approaches seem reasonable here:

1. Rename the executables to something namespaced, e.g.:

   * `isaaclab-play`
   * `isaaclab-train`

2. Introduce a single top-level `isaaclab` CLI with subcommands:

   * `isaaclab train [...]`
   * `isaaclab play [...]`

Personally, I think option 2 is the cleanest long-term design, especially since implementing subcommands with `argparse` is straightforward and scales better as the CLI surface grows. However, even option 1 would already avoid the namespace collision problem.

PR #5656, opened 2026-05-16
[Fix] Exclude numpy 2.3.5 from every IsaacLab install path

hujc7:
## 1. Summary

Adds `numpy!=2.3.5` at every install site that pulls numpy (10 touchpoints: 6 `setup.py`, 2 `install.py` pip argvs, 2 Dockerfile pip lines) so the broken release whose vendored OpenBLAS atfork handler crashes Kit's `libomni.platforminfo` `fork()` never lands in any pip resolve. Final image ends on `numpy 2.3.4`; supersedes #5642.

## 2. Sites covered

`!=2.3.5` is the *only* delta added at each site — no `>=2` constraints introduced where none existed.

| File | Site | Why this site sees numpy |
|---|---|---|
| `source/isaaclab/setup.py` | `INSTALL_REQUIRES` | direct numpy dep |
| `source/isaaclab_tasks/setup.py` | `INSTALL_REQUIRES` | direct numpy dep |
| `source/isaaclab_rl/setup.py` | `INSTALL_REQUIRES` | direct numpy dep |
| `source/isaaclab_visualizers/setup.py` | `INSTALL_REQUIRES` | direct numpy dep |
| `source/isaaclab_teleop/setup.py` | new `INSTALL_REQUIRES` entry | transitive: `dex-retargeting` → `pin` → cmeel-boost |
| `source/isaaclab_mimic/setup.py` | new `INSTALL_REQUIRES` entry | transitive: `h5py` |
| `source/isaaclab/.../install.py` | `_ensure_pink_ik_dependencies_installed` pip argv | force-reinstall of `pin pin-pink daqp` |
| `source/isaaclab/.../install.py` | `_maybe_preinstall_arm_nlopt` pip argv | ARM nlopt source-build prep |
| `docker/Dockerfile.base` | ARM nlopt prep `RUN pip install` | mirrors `_maybe_preinstall_arm_nlopt` |
| `docker/Dockerfile.curobo` | ARM nlopt prep + `nvidia-curobo` install | both can pull numpy under the cmeel cap |

`isaaclab_teleop`'s post-`--install` reinstall in `Dockerfile.curobo` is covered by the teleop `setup.py` entry — no separate Dockerfile argv needed.

### 3.1 Why declared at every site, not just `source/isaaclab/setup.py`

[...]

PR #5655, opened 2026-05-16
[Diag][DO NOT MERGE] Negative arm — numpy 2.3.5 atfork crash (no 5656)

hujc7:
## Purpose

Negative arm of the single-variable A/B that validates https://github.com/isaac-sim/IsaacLab/pull/5656.

Companion positive arm: https://github.com/isaac-sim/IsaacLab/pull/5684.

## Setup

- Branch base: develop (`a9b62101ca6`), no other workarounds.
- One commit on top: `source/conftest.py` that prints the resolved numpy version and bundled OpenBLAS .so filename at pytest session start.
- No `isaaclab.sh --install` overrides, no force-reinstall, no action.yml edits.

`git diff jichuanh/validate-numpy-pin-5642..jichuanh/validate-numpy-2-3-4-worstcase` between this branch and the positive arm equals exactly 5656's 10-site `numpy!=2.3.5` exclusion — nothing else.

## Expected result

- Install resolves to numpy 2.3.5 (cmeel-boost cap funnels to 2.3.x; pip picks the highest available).
- `[dep-manifest]` log line shows `numpy 2.3.5` + `libscipy_openblas64_-fdde5778.so`.
- Canary jobs SIGSEGV in `libomni.platforminfo` fork():
  - `isaaclab_physx`
  - `isaaclab_newton`
  - `isaaclab_rl`

If any canary passes, the negative test is invalid (something else on develop is already deferring/masking the bug — investigate before trusting the positive arm).

## Lifecycle

Diagnostic-only. **Do not merge.** Closes once evidence is captured.

## Related

- https://github.com/isaac-sim/IsaacLab/pull/5656 — the fix being validated
- https://github.com/isaac-sim/IsaacLab/pull/5684 — positive arm
- https://github.com/numpy/numpy/issues/30092
- https://github.com/OpenMathLib/OpenBLAS/pull/5520

PR #5800, opened 2026-05-27
[Diag][DO NOT MERGE] Negative arm — revert #5633 torch-defer to validate #5656 numpy!=2.3.5 fix

hujc7:
## 1. Purpose

Negative arm of the validation for PR #5656 (the ``numpy!=2.3.5`` exclusion).

Develop currently carries **two** independent mitigations for the OpenBLAS atfork SIGSEGV:

1. **PR #5633** (Antoine Richard, merge ``a5eb9add4c3``) which carried PR #5620's content — defers ``import torch`` in ``AppLauncher`` until **after** ``SimulationApp`` starts, so numpy/OpenBLAS isn't loaded in pytest's master process before the Kit fork.
2. **PR #5656** — excludes ``numpy 2.3.5`` from every install path, so even an early numpy import resolves to a non-buggy version.

If either mitigation alone is sufficient, the other is redundant. This PR reverts (1) so we can see whether (2) holds on its own.

## 2. Setup

- Branch base: ``develop`` (latest).
- Commit 1: surgical revert of ``source/isaaclab/isaaclab/app/app_launcher.py`` to its state just before ``a5eb9add4c3`` (#5633). Other files from #5633 left alone — the defer-torch logic in ``AppLauncher.__init__`` is the only piece relevant to the atfork bug.
- Commit 2: ``source/isaaclab/test/conftest.py`` that imports numpy and prints version + bundled OpenBLAS hash at pytest session start.

The conftest is deliberately placed at ``source/isaaclab/test/`` rather than ``source/`` to avoid the namespace-package shadowing that broke PR #5655 — anchoring pytest's rootdir at ``source/`` prepends it to ``sys.path``, making every ``source/<pkg>/`` (no ``__init__.py``) shadow the real pip-installed package and turn ``from isaaclab_teleop import IsaacTeleopCfg`` into an ImportError before any test runs.

## 3. Expected result

- Install resolves to numpy 2.3.4 (highest 2.3.x that satisfies ``!=2.3.5`` under cmeel-boost's cap from #5656).
- ``[dep-manifest]`` log line shows ``numpy 2.3.4`` + ``libscipy_openblas64_-8fb3d286.so``.
- ``import torch`` happens in ``AppLauncher.__init__`` BEFORE ``SimulationApp`` fork, but resolves to non-buggy numpy 2.3.4, so canary jobs (``isaaclab_physx``, ``isaaclab_newton``, ``isaaclab_rl``) pass.

If any canary still SIGSEGVs in ``libomni.platforminfo`` fork, the ``numpy!=2.3.5`` exclusion alone is **not** sufficient and #5633's torch-defer is also necessary — keep both.

If all canaries pass, the ``numpy!=2.3.5`` exclusion alone is sufficient and #5633's torch-defer is redundant — could be reverted in a follow-up cleanup.

## 4. Lifecycle

Diagnostic-only. **Do not merge.** Closes once evidence is captured.

## 5. Related

- PR #5656 — the ``numpy!=2.3.5`` exclusion being validated
- PR #5633 — Antoine's merge of Piotr's #5620 torch-defer mitigation (the one we're temporarily reverting)
- PR #5620 — Piotr's original (closed) version of the same mitigation
- PR #5655 — earlier diagnostic attempt; superseded by this PR (the ``source/conftest.py`` placement there broke test discovery before the bug could surface)
- numpy/numpy#30092
- OpenMathLib/OpenBLAS#5520

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,31 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Dep-manifest diagnostic: prints numpy version + bundled OpenBLAS hash at pytest session start.
+
+Located at the repo root so every subprocess pytest (driven by
+``tools/conftest.py``) discovers and loads it, regardless of which package's
+tests are running. The repo root has no ``isaaclab_*`` subdirectories, so
+``importmode=prepend`` placing the repo root on ``sys.path`` does NOT shadow
+the real pip-installed IsaacLab packages — unlike ``source/conftest.py``,
+where ``source/<pkg>/`` (no ``__init__.py``) would be promoted to a namespace
+package and break ``from isaaclab_teleop import IsaacTeleopCfg`` style imports.
+
+Importing numpy here registers its vendored OpenBLAS ``pthread_atfork``
+handler in the same process that later calls ``fork()`` via
+``SimulationApp()``. The print output identifies which numpy + OpenBLAS bundle
+actually landed in each CI test container.
+"""
+
+import os
+
+import numpy
+
+print(f"\n[dep-manifest] numpy {numpy.__version__}", flush=True)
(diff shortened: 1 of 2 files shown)


hujc7:
CI all green with latest. Closing this one.

PR #5823, opened 2026-05-28
[CI] Cross-platform — Part 4: Multi-GPU pytest workflow with cuda:1 coverage

hujc7:
## 1. Summary

- **CI:** Adds a dedicated multi-GPU pytest workflow that runs one shard per non-default `cuda:N`, with a shared atomic work queue, per-file reports, and end-of-run reconciliation.
- **isaaclab:** Adds composable `DeviceScope` flags and `test_devices()` for `scope ∩ runtime` parametrization while retaining exact custom string masks.
- **isaaclab:** Uses `ISAACLAB_TEST_DEVICES` as the single source for both test parametrization and explicit Kit launch-device selection; `AppLauncher` has no implicit test-environment override.
- **isaaclab:** Migrates 19 device-parametrized test modules across the core, Newton, OV PhysX, and PhysX suites while preserving the existing cpu + cuda:0 single-GPU behavior.

## 2. Device selection

```python
from isaaclab.test.utils import DeviceScope, test_devices

test_devices()                                                  # cpu + every GPU
test_devices(DeviceScope.CUDA)                                  # every GPU
test_devices(DeviceScope.CPU | DeviceScope.NON_DEFAULT_CUDA)    # composed scope
test_devices("101X")                                            # exact custom mask
```

- `ISAACLAB_TEST_DEVICES` limits which devices a run may use; unset preserves the historical cpu + cuda:0 runtime.
- The multi-GPU lane discovers non-default-capable scopes and narrows each shard to one concrete device.

## 3. Dependencies

All prerequisite changes have merged into `develop`, so this PR no longer carries them — it is now scoped to the device-selection infrastructure and the multi-GPU pytest workflow only:

- #5695 — Cross-platform Part 1 (base pytest markers / shared scaffolding).
- #5881 — Sim honors the device kwarg over `sim_cfg.device` in `build_simulation_context`.
- #5933 — Kit renderer defaults to one GPU + adds `ISAACLAB_FABRIC_USE_GPU_INTEROP`; the multi-GPU lane sets `ISAACLAB_FABRIC_USE_GPU_INTEROP=0`.
- #6322 — Newton multi-GPU initialization on non-default CUDA devices (`cuda:1` and higher).

## 4. Test plan

- [x] Device-selection unit suite passes.
- [x] Changed Python modules compile; multi-GPU shell scripts pass `bash -n`.
- [x] Full documentation build (`./isaaclab.sh -d`).
- [x] Full pre-commit suite (`./isaaclab.sh -f`).
- [x] GitHub Actions green on latest `develop` (one `rendering-correctness-kitless` flake cleared on rerun; fixed upstream on `develop` by #6431).


source/isaaclab/isaaclab/test/utils/devices.py
```diff
@@ -0,0 +1,210 @@
...
+
+    The order is ``[cpu]`` followed by ``cuda:0, cuda:1, ...`` for each CUDA
+    device torch reports as visible. ``cpu`` is always included even when
+    torch has no CUDA backend; downstream callers can filter via the mask.
+
+    Returns:
+        Ordered list of device strings as torch would address them.
+    """
+    devices = ["cpu"]
+    if torch.cuda.is_available():
```
pbarejko, line 253: Why are we using `torch` api here? This smells like tech debt. Shouldn't we switch to use `warp`?  i.e.:

```python
>>> import warp
>>> warp.init()
Warp 1.12.0 initialized:
   CUDA Toolkit 12.9, Driver 13.0
   Devices:
     "cpu"      : "x86_64"
     "cuda:0"   : "NVIDIA RTX PRO 6000 Blackwell Workstation Edition" (95 GiB, sm_120, mempool enabled)
   Kernel cache:
     /home/pbarejko/.cache/warp/1.12.0
>>> warp.get_cuda_devices()
['cuda:0']
```

source/isaaclab/test/sim/test_newton_model_utils.py
```diff
@@ -177,7 +178,7 @@ def _run_scatter_shape_color_rows_kernel(
         pytest.param((-0.25, 1.75, 0.5), id="oob_clamps_pow"),
     ],
 )
-@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
+@pytest.mark.parametrize("device", cuda_test_devices())
```
pbarejko: Based on function name `cuda_test_devices` have we started to exclude `cpu` device tests. If so - why?

source/isaaclab/isaaclab/test/utils/devices.py
```diff
@@ -0,0 +1,210 @@
...
+import torch
+
+_DEFAULT_MASK = "110"
+"""Default mask when ``ISAACLAB_TEST_DEVICES`` is unset.
+
+``110`` means cpu + cuda:0, matching the historical single-GPU CI device list
+so adopting this helper has zero impact on single-GPU runs.
+"""
+
+_ENV_VAR = "ISAACLAB_TEST_DEVICES"
```
pbarejko: Name of this global variable makes no sense, besides it's used in one place only. This unnecessary indirection makes code harder to read.

source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py
```diff
@@ -308,11 +300,7 @@ def test_fabric_cuda1_no_usd_writeback(device, view_factory):
     )
 
 
-@pytest.mark.skipif(
-    not os.environ.get("ISAACLAB_TEST_MULTI_GPU"),
-    reason="Multi-GPU tests disabled (set ISAACLAB_TEST_MULTI_GPU=1 to enable)",
-)
-@pytest.mark.parametrize("device", ["cuda:1"])
+@pytest.mark.parametrize("device", cuda_test_devices(mask="001", strict=False))
```
pbarejko: I don't understand the premise behind _strictness_ in this function. I read it as _maybe run on GPU 1, if unavailable then run on what's available_. This introduces testing ambiguity, Just by looking at this line I don't know what you are trying to test.

In my opinion we would be better without `cuda_test_devices` function and keep list explicit, instead of resolved at runtime. In testing you want to keep constraints explicit.

source/isaaclab/isaaclab/test/utils/devices.py
hujc7: Checked. using warp requires initializing runtime which might have side-effect to kit. torch does not have that.

source/isaaclab/test/sim/test_newton_model_utils.py
hujc7: renamed to test_devices

source/isaaclab/isaaclab/test/utils/devices.py
hujc7: Renamed.

source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py
hujc7: Strict is removed. 
The goal of this is to dynamically configure the parameterization at test time so we run it where suitable. On regular test, `cpu` and `cuda:0`; and `cuda:X` for multi-gpu. The overall goal is to mark all tests needing device parameterization with this.

source/isaaclab/isaaclab/test/utils/devices.py
kellyguo11: we would always have warp loaded for our APIs, right? what are the side effects that impact kit?

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1003,6 +1003,17 @@ def _resolve_device_settings(self, launcher_args: dict):
         device = launcher_args.get("device", AppLauncher._APPLAUNCHER_CFG_INFO["device"][1])
 
         device_explicitly_passed = launcher_args.pop("device_explicit", False)
+        # When the caller didn't pin a device, allow ISAACLAB_SIM_DEVICE to
+        # override the default. Used by the multi-GPU CI workflow to boot
+        # Kit with active_gpu=1 without editing every test's AppLauncher()
+        # call site; Kit's active_gpu is process-global and locked after
+        # SimulationApp init, so per-test selection cannot retarget it.
+        if not device_explicitly_passed:
+            env_device = os.environ.get("ISAACLAB_SIM_DEVICE")
```
kellyguo11: Could we avoid reading ISAACLAB_SIM_DEVICE directly inside _resolve_device_settings? The behavior makes sense for the multi-GPU workflow, but this makes AppLauncher feel like it is accumulating unrelated CI env vars. I think we are already doing too much of that inside the AppLauncher code, I'd really like to avoid all the CI-specific branching if we can.

source/isaaclab/isaaclab/test/utils/devices.py
```diff
@@ -0,0 +1,169 @@
...
+        :func:`pytest.mark.parametrize`. Empty means the test is skipped for
+        this run.
+
+    Raises:
+        ValueError: When the run explicitly set ``ISAACLAB_TEST_DEVICES`` to
+            devices that are not available on this host (a misconfigured run that
+            would otherwise skip everything and pass vacuously). A scope that
+            merely does not intersect this run's devices skips, it does not raise.
+    """
+    requested = os.environ.get(_RUNTIME_DEVICES_ENV_VAR)
```
kellyguo11, line 189: why do we need to reference CI env variables in the core code?

source/isaaclab/isaaclab/test/utils/devices.py
```diff
@@ -0,0 +1,169 @@
...
+
+_RUNTIME_DEVICES_ENV_VAR = "ISAACLAB_TEST_DEVICES"
+"""Env var naming the run's devices: the devices a run may use (see module docstring)."""
+
+_DEFAULT_RUNTIME = "110"
+"""Runtime devices when :data:`_RUNTIME_DEVICES_ENV_VAR` is unset: cpu + cuda:0,
+i.e. the historical single-GPU device set, so non-default GPUs are opt-in per run."""
+
+
+def test_devices(scope: str = "11X", *, skip: dict[str, str] | None = None) -> list:
```
kellyguo11: will this conflict with pytests's test_* search for test cases? could this become more of a test-only module that's not exposed as part of the isaaclab core APIs?

source/isaaclab/isaaclab/test/utils/devices.py
```diff
@@ -0,0 +1,169 @@
...
+
+    ISAACLAB_TEST_DEVICES=0001 ISAACLAB_SIM_DEVICE=cuda:2 \\
+        ./isaaclab.sh -p -m pytest path/to/test.py
+"""
+
+from __future__ import annotations
+
+import os
+
+import torch
```
kellyguo11: is this guaranteed to be imported after launching kit?

source/isaaclab/test/utils/test_device_selection.py
```diff
@@ -0,0 +1,148 @@
...
+    assert resolve_devices("11X", skip={"cuda:3": "n/a"}) == ["cpu", "cuda:0"]
+
+
+# ---------------------------------------------------------------------------
+# _expand mask grammar
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "mask, count, expected",
```
kellyguo11, line 194: I think the mask idea is clever, but it's a burden on all test developers that I feel might not be very intuitive. If we expect users from the community to implement tests, they'll need to understand the masks and make sure they specify the correct values if their tests cannot support some of the options. perhaps we can have some more user friendly naming for CPU, GPU, and default cuda device options? maybe something like

```
test_devices(DeviceScope.ALL)
test_devices(DeviceScope.CPU_AND_DEFAULT_CUDA)
test_devices(DeviceScope.NON_DEFAULT_CUDA)
```

source/isaaclab/isaaclab/test/utils/devices.py
hujc7: This is for configuration the device at CI level. e.g. for multi-gpu instance, it can dispatch the test to one of the non-default gpus.

source/isaaclab/isaaclab/test/utils/devices.py
hujc7: Correction: it's the warp behavior on non-default devices such that torch.cuda.set_device must be called first before wp.init. The same issue was discovered by Octi before.

source/isaaclab/isaaclab/test/utils/devices.py
hujc7: it does not. Was it for the kit crash issue due to openblast? The issue should be resolved. I can move this to function local import.

source/isaaclab/test/utils/test_device_selection.py
hujc7: Yes, I will add enum name support

kellyguo11 (approved):

PR #5825, opened 2026-05-28
Adds tests for `uv pip` installation path

myurasov-nv:
# Description

Adds install_ci tests for the `uv pip` install path and standardizes the existing install_ci tests (naming, docstrings, markers, fold/dedupe).

- New `install_ci/uv_pip/` tests covering documented isaaclab wheel install paths: bare core (`uv pip install <wheel>`), `[isaacsim]` only, `[rl,tasks]` without isaacsim, and `[all,isaacsim]` + cartpole training.
- Renamed `install_ci/cli` tests to `test_cli_<install>_in_<env>_<modifier>.py`; folded mimic+newton into the smoke file; dropped the physx duplicate.
- Switched install_ci top docstrings to a `Setup:`/`Tests:` command-list format that mirrors the documented install commands.
- Renamed test classes to snake-style `Test_<Filename_Stem>` and methods to `test_<install_cmd>_<outcome>`.
- Added `install_path_cli` and `install_path_uv_pip` markers (replacing `cli`); install paths are now filterable with `pytest -m install_path_uv_pip`.

## Type of change

## Checklist


myurasov-nv:
Dear 🤖 Isaac Lab Review Bot!

Addressing the findings from the first review:

1. Wheel cleanup after test: not applying. install_ci tests run in ephemeral Docker containers, so the wheel under `tools/wheel_builder/build/dist/` does not accumulate across runs. Also, `test_wheel_builder_smoke.py` consumes the same artifact path; cleaning it in this test's teardown could interfere when both tests are collected.

2. Two-step wheel count assertion: not applying. The current `assert len(wheels) == 1, f"... found: {wheels}"` already covers the 0-case (message reads "found: []") and the >1 case with the same diagnostic, so splitting it adds no information.

3. PEP 8 class naming (`Test_Wheel_UV_Install_Trains_Cartpole` vs `TestWheelUVInstallTrainsCartpole`): keeping snake-style. This is the chosen convention for install_ci test classes (now `Test_Cli_Install_<...>` across the directory) so the class name mirrors the filename for long compound names. Acknowledged the bot flagged this as stylistic.

Q1: The new convention is introduced by this PR. Earlier commits in this branch did the initial rename to `test_cli_<action>_<thing>_in_<envname>.py`; later commits refined it to `test_cli_<install>_in_<env>_<modifier>.py`. Not a dependency on a prior merge on develop.

Q2: No `__init__.py` needed. `install_ci/pytest.ini` sets `--import-mode=importlib`, which does not require package init files for collection. Sibling `cli/` and `misc/` directories also have no init files.

The inline cosmetic suggestion (extra space in `[all, isaacsim]`) was self-dismissed in the follow-up review; not applying.

myurasov-nv:
Dear 🤖 Isaac Lab Review Bot!

Confirmed. The `feba08fb` commit was an attempt to apply your earlier "remove space in `[all, isaacsim]`" suggestion that landed in the wrong place: it inserted `f"{wheel}[all,isaacsim]"` as a stray positional arg before the list and removed the opening `[`, leaving the closing `]` orphaned.

Fixed in `ef4838ad`: restored the `run_in_uv_env([..., f"{wheel}[all,isaacsim]", ...], ...)` list structure with the no-space form applied in the proper position, and dropped the space from the assertion message too. `python -c "import ast; ast.parse(...)"` parses clean.

source/isaaclab/test/install_ci/cli/test_cli_install_in_uvenv_smoke.py
```diff
@@ -0,0 +1,198 @@
...
+            # ./isaaclab.sh -i newton
+            result = self.run_in_uv_env([str(self.cli_script), "-i", "newton"], cwd=isaaclab_root)
+            assert result.returncode == 0, f"isaaclab -i newton failed:\n{result.stdout}\n{result.stderr}"
+
+            # Run isaaclab_newton test suite
+            test_dir = str(isaaclab_root / "source" / "isaaclab_newton" / "test")
+            result = self.run_in_uv_env(
+                ["python", "-m", "pytest", test_dir, "-sv", "--tb=short"],
+                cwd=isaaclab_root,
+            )
```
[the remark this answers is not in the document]
myurasov-nv: Dear Greptile! Fixed in 0a25b789. Set explicit `timeout=3200` on the inner `run_in_uv_env([..., 'python', '-m', 'pytest', ...])` call here and on the matching `isaaclab_physx` pytest subprocess in `test_cli_install_core_in_uvenv_correctness.py` so the nested run gets the full window granted by the outer `@pytest.mark.timeout(3600)`.

source/isaaclab/test/install_ci/uv_pip/test_uv_pip_install_isaaclab_imports_isaaclab.py
```diff
@@ -0,0 +1,83 @@
...
+    @pytest.mark.slow
+    @pytest.mark.timeout(900)
+    def test_install_makes_isaaclab_version_match_wheel(self):
+        """``isaaclab.__version__`` equals the version segment in the wheel filename."""
+        result = self.run_in_uv_env(["python", "-c", "from isaaclab import __version__; print(__version__)"])
+        imported_version = result.stdout.strip()
+        expected_version = self._wheel.split("/")[-1].split("-")[1]
+        assert imported_version == expected_version, (
+            f"isaaclab.__version__ mismatch: expected {expected_version}, got {imported_version}"
+        )
```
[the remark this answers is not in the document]
myurasov-nv: Dear Greptile! Applied your suggested fix in 0a25b789. Added the `assert result.returncode == 0` guard before parsing stdout so a failed import surfaces the real `stderr` instead of an empty-version mismatch.

source/isaaclab/test/install_ci/uv_pip/test_uv_pip_install_isaaclab_all_isaacsim_trains_cartpole.py
```diff
@@ -0,0 +1,127 @@
...
+def _assert_training_passed(result) -> None:
+    output = result.stdout + (result.stderr or "")
+    assert result.returncode == 0, f"Training failed (rc={result.returncode}):\n{output}"
+    assert "Traceback (most recent call last):" not in output, f"Training produced a traceback:\n{output}"
+    assert "Training time:" in output, f"Training did not report completion:\n{output}"
+
+
+@pytest.mark.install_path_uv_pip
+class Test_Uv_Pip_Install_Isaaclab_All_Isaacsim_Trains_Cartpole(UV_Mixin):
+    """Build the wheel, ``uv pip install <wheel>[all]``, verify cartpole training."""
```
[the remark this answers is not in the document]
myurasov-nv: Dear Greptile! Fixed in 0a25b789. Updated the class docstring to `uv pip install <wheel>[all,isaacsim]` to match the module-level Setup block and the actual install command.

source/isaaclab/test/install_ci/cli/test_cli_install_in_uvenv_smoke.py
```diff
@@ -0,0 +1,198 @@
...
+            self.create_uv_env(isaaclab_root)
+
+            # ./isaaclab.x -i core — core set only, no optional extras
+            result = self.run_in_uv_env([str(self.cli_script), "-i", "core"], cwd=isaaclab_root)
+            assert result.returncode == 0, f"isaaclab -i core failed:\n{result.stdout}\n{result.stderr}"
+
+            # All core packages should be importable.
+            for pkg in ("isaaclab_assets", "isaaclab_tasks", "isaaclab_rl", "isaaclab_physx"):
+                result = self.run_in_uv_env(["python", "-c", f"import {pkg}; print('{pkg} ok')"])
+                assert result.returncode == 0, f"import {pkg} failed:\n{result.stdout}\n{result.stderr}"
```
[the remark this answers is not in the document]
myurasov-nv: Dear Greptile! Fixed in 0a25b789. Replaced all five `./isaaclab.x` occurrences in the smoke docstrings with `./isaaclab.sh`.

myurasov-nv:
@greptile-apps

myurasov-nv:
@greptile-apps please re-review.

All four findings from the previous review are addressed in the current HEAD; the inline replies below each comment summarize the fix:

- P1 subprocess timeout: explicit `timeout=3200` on the inner `pytest` subprocess in both `test_install_newton_passes_isaaclab_newton_test_suite` and `test_install_core_passes_isaaclab_physx_test_suite`.
- P1 missing returncode guard: `assert result.returncode == 0` added before parsing stdout in `test_install_makes_isaaclab_version_match_wheel`.
- P2 class docstring contradiction: class docstring now says `uv pip install <wheel>[all,isaacsim]`.
- P2 `isaaclab.x` typos: all five occurrences in `test_cli_install_in_uvenv_smoke.py` replaced with `isaaclab.sh`.

The branch was rebased on develop after those fixes, so the comments are anchored to older commit SHAs (`388a4d9` / `f5d5f51`); the current HEAD is `f5d5f51` with the fixes applied.

AGENTS.md
```diff
@@ -168,6 +168,138 @@ Follow conventional commit message practices.
 
 - **Always verify regression tests fail without the fix.** When writing a regression test for a bug fix, temporarily revert the fix and run the test to confirm it fails. Then reapply the fix and verify the test passes. This ensures the test actually covers the bug.
 
+### Install CI tests (`source/isaaclab/test/install_ci/`)
```
StafaH, line 171: Let's reduce the scope of what's added here, adding things specific to CLI tests should not be needed here. Agents are pretty good at figuring out directory structure and tests naming based on the existing tests, they won't need the guidelines.

.github/actions/run-tests/action.yml
```diff
@@ -295,7 +295,7 @@ runs:
                 esac
               fi
               echo 'Starting pytest with path: $test_path'
-              ./isaaclab.sh -p -m pytest --ignore=tools/conftest.py --ignore=source/isaaclab/test/install_ci $test_path $pytest_options -v --junitxml=tests/$result_file
```
StafaH, line 298: Why are we removing this? we still want to not collect the install tests since there is a seperate action for that (comment in tools/conftest.py mentions this)

source/isaaclab/test/install_ci/uv_pip/test_uv_pip_install_isaaclab_all_isaacsim_trains_cartpole.py
```diff
@@ -0,0 +1,140 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+Setup:
+    - bash tools/wheel_builder/build.sh
```
StafaH: Here and in other places, we have isolated tests that all build wheels. Can we consolidate and only build the wheel once, and then re-use the build. Also can we consolidate the test files, do we need one test per file?

Along this line, the AGENTS.md now has these rules as well, which I think we should remove/relax.

StafaH (commented):
Hi @myurasov-nv, thanks for the PR, the better coverage will help alot.

Left some comments inline, but as an overall have 3 concerns which are detailed in the in-line comment:
1. The new rules in AGENTS.md are strict and very specific for CLI tests, I think we should remove
2. There is a discrepancy with the action.yml change, it directly contradicts that the CLI tests should be isolated from the rest of the tests
3. The new formatting creates excessive test files with repeated behavior that will make this CI long

AGENTS.md
myurasov-nv: Hence, there were cases of misplaced tests that do not belong in Install CI.

.github/actions/run-tests/action.yml
myurasov-nv: tools/conftest.py does its own collection that bypasses pytest's --ignore, so it had no effect.

source/isaaclab/test/install_ci/uv_pip/test_uv_pip_install_isaaclab_all_isaacsim_trains_cartpole.py
myurasov-nv: I considered this, but I think it's important to keep those tests 100% self-contained and not depending on the way they are called. Complete reproducibility can be achieved only this way.

StafaH (approved):

PR #5826, opened 2026-05-28
[Fix] Defer pxr loading in isaaclab.sim, .assets, .scene, and .sim/utils

hujc7:
## 1. Summary

- Closes NVBug 6121889: the direct Cartpole env (Warp backend) + ``--visualizer kit`` crashes during ``SimulationApp.startup`` with a ``TfNotice`` / ``UsdAPISchemaBase`` / ``GfVec3f`` converter cascade.
- Root cause: module-import-time ``from pxr import …`` in IsaacLab library code. Env-cfg parse runs *before* ``AppLauncher`` and transitively pulls ``pxr`` into ``sys.modules``, so Kit's USD-binding registration fails against an already partially-initialized ``pxr``.
- Fix: defer ``pxr`` into the function bodies that use it across the 9 modules on the env-cfg parse path. **0** ``pxr``/``omni``/``carb``/``isaacsim`` modules now leak across all **197** ``Isaac-*`` env-cfg parses (the Cartpole path previously pulled in 63 ``pxr``/``omni``/``isaacsim`` modules).
- No public API change: function bodies are otherwise unchanged; the diff is import placement only.

### 2.1 Per-file pattern

Each module-top ``from pxr import …`` is removed. The runtime import moves into each function that uses USD; type hints stay under ``TYPE_CHECKING`` so static checkers and IDE autocomplete keep resolving ``Usd.Stage``, ``Sdf.Path``, etc. against the real classes.

```python

if TYPE_CHECKING:
    from pxr import Sdf, Usd  # noqa: F401


def some_fn(prim: Usd.Prim) -> Sdf.Path:
    from pxr import Sdf, Usd  # noqa: PLC0415
    ...
```

``# noqa: PLC0415`` matches the function-local import sites already present across the repo; the ``TYPE_CHECKING`` guard matches the 200+ files already using ``if TYPE_CHECKING: from X import Y``.

### 2.2 Affected modules

| Module | pxr names deferred to function bodies |
|---|---|
| ``isaaclab.sim.simulation_context`` | ``Gf, UsdGeom, UsdPhysics, UsdUtils`` |
| ``isaaclab.assets.asset_base`` | ``Usd`` (type-only — no runtime pxr use) |
| ``isaaclab.scene.interactive_scene`` | ``Sdf`` |
| ``isaaclab.sim.spawners.from_files.from_files`` | ``Gf, Sdf, UsdGeom`` |
| ``isaaclab.sim.utils.prims`` | ``Sdf, Usd, UsdGeom, UsdPhysics, UsdShade, UsdUtils`` |
| ``isaaclab.sim.utils.queries`` | ``Sdf, Usd, UsdPhysics`` |
| ``isaaclab.sim.utils.semantics`` | ``Usd, UsdGeom, UsdSemantics`` |
| ``isaaclab.sim.utils.stage`` | ``Sdf, Usd, UsdUtils`` |
| ``isaaclab.sim.utils.transforms`` | ``Gf, Sdf, Usd, UsdGeom`` |

## 3. Verification

- ``source/isaaclab_tasks/test/test_env_cfg_no_forbidden_imports.py`` (existing integration guard): all 197 ``Isaac-*`` tasks parse their env-cfg with **0** ``pxr``/``omni``/``carb``/``isaacsim``/``scipy`` imports.
- Static check: every runtime ``pxr`` symbol use in the 9 modules is covered by a function-local ``from pxr import …``; type-only uses remain under ``TYPE_CHECKING`` (no module-scope ``pxr`` reference remains).
- ``./isaaclab.sh -f`` clean.

## 4. Test plan

- [x] ``./isaaclab.sh -f`` clean.
- [x] ``test_env_cfg_no_forbidden_imports.py`` — 197 tasks, no backend leak.
- [x] Direct Cartpole (Warp) + ``--visualizer kit``: env-cfg parse loads 0 ``pxr`` modules; ``SimulationApp.startup`` proceeds without the ``TfNotice`` / ``GfVec3f`` cascade.
- [x] PhysX backend tests unaffected — function bodies unchanged, ``pxr`` still imports on first use.

## 5. Out of scope / follow-ups

~32 more files in ``isaaclab/``, ``isaaclab_newton/``, ``isaaclab_physx/`` still ``from pxr import …`` at module top. They are not on the Cartpole-Warp env-cfg parse path so they don't block this fix, but should be converted the same way as the bug class will resurface for new envs that touch them:

- **Tier A — likely on env-cfg chains (~14 files):** ``cloner/cloner_utils``, ``sim/spawners/{lights,materials,meshes,shapes,sensors,wrappers}``, ``sim/schemas/schemas``, ``sim/utils/{legacy,newton_model_utils}``, ``sim/views/usd_frame_view``, ``isaaclab_newton/{assets,cloner,sensors,sim/views}``.
- **Tier B — PhysX equivalents (~13 files):** ``isaaclab_physx/*``.
- **Tier C — feature-gated, narrow trigger (~7 files):** ``sensors/{camera,ray_caster}``, ``envs/mdp/actions/task_space_actions``, ``envs/ui/base_env_window``, ``ui/xr_widgets``, ``sim/converters/mesh_converter``, ``utils/mesh``.


kellyguo11:
what's the main reason why we can't import pxr before these modules?

hujc7:
> what's the main reason why we can't import pxr before these modules?

Kit registers a specialized pxr such that, if pxr is imported before appLaunch, the later kit registration will have problems. The crashing behavior could be different but it seems the root cause is the same.

AntoineRichard (approved):
LGTM

PR #5834, opened 2026-05-28
Add generic Newton coupled-solver with Proxy + ADMM variants

mmichelis:
# Description

## Summary

* **New `isaaclab_contrib.coupling` submodule.** It exposes `NewtonCoupler` and a typed `Coupler*Cfg` API for Newton's experimental coupled solvers:
    * `CouplerProxyCfg` → `SolverCoupledProxy`: lagged-impulse virtual-proxy coupling. This works well when only a small set of rigid bodies, such as a gripper, needs to interact with a deformable while the rest of the articulation remains entirely in the rigid solver.
    * `CouplerAdmmCfg` → `SolverCoupledADMM`: linearized ADMM coupling with tunable `rho`, `gamma`, and `baumgarte` penalties plus cross-solver joint stiffness and damping.
* **N-way named solver entries.** `CouplerEntryCfg` partitions the model into named entries, each with its own typed `solver_cfg`. Each entry constructs its nested solver through that config's `class_type`, so combinations such as MJWarp + VBD, Featherstone + VBD, XPBD + Kamino, and implicit MPM work without coupler-specific solver registries.
* **Implicit MPM entries.** Entries forward `substeps` and `in_place` to Newton, reuse the MPM builder-registration and kinematic-collider finalization hooks, request FK before stepping, and prepare coupled contacts before CUDA graph capture.
* **Explicit external-contact routing.** Proxy source entries configured to consume Newton contacts receive the shared collision pipeline, including MJWarp with `use_mujoco_contacts=False`. Proxy destinations continue to use their entry-local collision pipeline.
* **Flexible ownership selectors.** Body selectors accept raw Newton body-label regex strings alongside `SceneEntityCfg`. Bodies, particles, joints, and shapes may be assigned to at most one entry; model elements may intentionally remain unassigned. When using `SceneEntityCfg`, assign `CouplerCfg.scene_cfg` from the environment config, typically in `__post_init__`.
* **`Isaac-Lift-Soft-Franka-v0` defaults to proxy-coupled MJWarp + VBD** through the `newton_mjwarp_vbd_proxy` preset.
* **`NewtonShapeCfg` gains `ke`, `kd`, and `mu` contact-material fields**, forwarded to Newton's `ModelBuilder.default_shape_cfg` for shapes without an explicit per-asset material.
* **Reusable Newton solver construction.** Newton solver managers expose a common nested-solver construction hook used by the coupler instead of duplicating config-to-constructor plumbing.
* **Documentation and tests.** The coupling API page, VBD guide, focused pure-Python tests, and existing changelog fragment cover the final `NewtonCoupler` / `Coupler*Cfg` API.

Builds on Newton Coupling in [gdaviet/coupled-solver-framework](https://github.com/gdaviet/newton/tree/gdaviet/coupled-solver-framework).

## How to use

Partition the model into named entries, then select a coupling algorithm with its typed coupler config:

```python
from isaaclab_contrib.coupling import (
    CouplerAdmmCfg,
    CouplerEntryCfg,
    CouplerProxyCfg,
    CouplerProxyMappingCfg,
)

# Proxy coupling: rigid MJWarp entry + soft VBD entry, with the gripper exposed as proxies.
solver_cfg = CouplerProxyCfg(
    entries=[
        CouplerEntryCfg(
            name="rigid",
            solver_cfg=MJWarpSolverCfg(...),
            bodies=["/World/envs/env_.*/Robot"],
        ),
        CouplerEntryCfg(
            name="soft",
            solver_cfg=VBDSolverCfg(iterations=10),
            all_particles=True,
            include_static_shapes=True,
        ),
    ],
    proxies=[
        CouplerProxyMappingCfg(
            source="rigid",
            destination="soft",
            bodies=[
                "/World/envs/env_.*/Robot/panda_hand",
                "/World/envs/env_.*/Robot/panda_(left|right)finger",
            ],
            collide_interval=5,
        )
    ],
    iterations=1,
)

# ADMM coupling: the same entries with symmetric contact interfaces instead of proxies.
solver_cfg = CouplerAdmmCfg(
    entries=[...],
    iterations=10,
    rho=1.0,
    gamma=0.0,
    baumgarte=0.0,
    joint_stiffness=1.0e4,
    joint_angular_stiffness=1.0e4,
)

# Selectors may also use SceneEntityCfg, for example:
# SceneEntityCfg("robot", body_names=["panda_hand"]).
# Attach the scene in the environment config so those selectors can resolve.
solver_cfg.scene_cfg = self.scene
```

An implicit MPM entry uses the same entry API:

```python
CouplerEntryCfg(
    name="media",
    solver_cfg=MPMSolverCfg(grid_type="fixed"),
    all_particles=True,
    substeps=4,
    in_place=True,
)
```

## Type of change

- Documentation update

## Test plan

- [x] `./isaaclab.sh -f` (pre-commit)
- [x] `./isaaclab.sh -p -m pytest source/isaaclab_contrib/test/coupling/test_coupler.py -q` — 24/24 passed
- [x] `./isaaclab.sh -d` — full documentation build passed
- [x] Changelog fragment validation passed
- [x] `Isaac-Lift-Soft-Franka-v0` with the `newton_mjwarp_vbd_proxy` preset runs stably

## Checklist

- [x] I have added my name to `CONTRIBUTORS.md`, or my name already exists there


mmichelis:
@greptile

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_cloth_env_cfg.py
```diff
@@ -166,12 +183,8 @@ def __post_init__(self) -> None:
         self.sim.dt = 1 / 60.0
         self.sim.render_interval = self.decimation
 
-        # viewer settings
-        self.viewer.origin_type = "asset_root"
-        self.viewer.asset_name = "robot"
-        self.viewer.env_index = 0
-        self.viewer.eye = (1.25, -1.5, 0.6)
-        self.viewer.resolution = (1920, 1080)
+        view = dict(eye=(1.25, -1.5, 0.6), lookat=(0.0, 0.0, 0.0), window_width=1920, window_height=1080)
```
ooctipus: do you meant viewerCfg?

source/isaaclab_contrib/isaaclab_contrib/coupling/coupled_manager.py
```diff
@@ -0,0 +1,510 @@
...
+
+    @dataclass
+    class _ResolvedProxy:
+        """Proxy configuration with source selectors resolved to indices."""
+
+        config: CoupledProxyCfg
+        bodies: list[int]
+        particles: list[int]
+
+    _SOLVER_CLASS_BY_CFG_TYPE: ClassVar[dict[type[NewtonSolverCfg], type[SolverBase]]] = {
```
ooctipus: we shouldn't Need this the class_type with resolvable string should handle it naturally

source/isaaclab_contrib/isaaclab_contrib/coupling/coupled_manager.py
```diff
@@ -0,0 +1,510 @@
...
+            rigid_contact_matching=solver_cfg.rigid_contact_matching,
+            contact_matching_pos_threshold=solver_cfg.contact_matching_pos_threshold,
+            contact_matching_normal_dot_threshold=solver_cfg.contact_matching_normal_dot_threshold,
+            contact_matching_force_scale=float(solver_cfg.contact_matching_force_scale),
+            contact_pairs=contact_pairs,
+        )
+        return SolverCoupledADMM(model=model, entries=entries, coupling=coupling)
+
+    @classmethod
+    def _validate_solver_cfg(
```
ooctipus: the validation code is a bit overly verbose maybe its remove-able?

source/isaaclab_contrib/isaaclab_contrib/coupling/coupled_manager.py
```diff
@@ -0,0 +1,510 @@
...
+    ) -> SolverCoupledADMM:
+        contact_pairs = (
+            SolverCoupledADMM.auto_detect_contact_pairs(entries)
+            if solver_cfg.contact_pairs is None
+            else [
+                SolverCoupledADMM.ContactPair(source=source, destination=destination)
+                for source, destination in solver_cfg.contact_pairs
+            ]
+        )
+        coupling = SolverCoupledADMM.Config(
```
ooctipus: I would make sure the cfg has the matching d_type then this buid just becomes SolverCoupledADMM.Config(**solver_cfg)

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,510 @@
...
+            bodies=entry.bodies,
+            particles=entry.particles,
+            joints=entry.joints,
+            shapes=entry.shapes,
+            substeps=entry_cfg.substeps,
+            in_place=entry_cfg.in_place,
+        )
+
+    @classmethod
+    def _build_proxy_coupled_solver(
```
ooctipus, line 348: again , the colyd looks like a lot of data piping, if we just ensure the type is given correctly at Cfg, leys construct just by   proxy_cfg.class_type(proxy_cfg)   or  NewtonConfig(**config)

source/isaaclab_contrib/isaaclab_contrib/coupling/coupled_manager.py
```diff
@@ -0,0 +1,510 @@
...
+        solver_cls = cls._resolve_solver_class(entry_cfg.solver_cfg)
+        solver_kwargs = (
+            {}
+            if isinstance(entry_cfg.solver_cfg, MPMSolverCfg)
+            else cls._filter_solver_kwargs(solver_cls, entry_cfg.solver_cfg)
+        )
+
+        def solver_factory(model_view):
+            if isinstance(entry_cfg.solver_cfg, MPMSolverCfg):
+                return solver_cls(
```
ooctipus: return entry_cfg.solver_cfg.class_type(entry_cfg.solver_cfg)
is the pattern we are using for lab

AntoineRichard (commented):
High-level comment. Should we promote VBD and the coupler to isaaclab rather than contrib?

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_soft_env_cfg.py
```diff
@@ -128,29 +104,89 @@ class DeformableCfg(PresetCfg):
...
         ),
+        default_shape_cfg=NewtonShapeCfg(ke=4e4, kd=1e-5, mu=5.0),
         num_substeps=10,
-        use_cuda_graph=True,
     )
 
     physx: PhysxCfg = PhysxCfg()
 
-    default = newton_mjwarp_vbd
+    default = newton_mjwarp_vbd_proxy
```
AntoineRichard, line 185: LEt's make sure the difference between the two is well documented.

AntoineRichard: (with pros & cons)

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+    CouplerEntryCfg,
+    CouplerProxyCfg,
+    CouplerProxyMappingCfg,
+)
+
+if TYPE_CHECKING:
+    from isaaclab.scene import InteractiveSceneCfg
+
+
+class NewtonCoupler(NewtonVBDManager):
```
AntoineRichard: Why does it inherit from VBDManager?

AntoineRichard:
I reviewed the current head with a particular focus on `source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py`. A few issues look worth addressing before merge:

1. **Nested entries bypass manager-specific lifecycle behavior** (`coupler.py:263-264`). `_build_entry()` calls only `solver_cfg.class_type._create_solver()`. This skips behavior implemented outside the constructor, including:
   - `MPMSolverCfg.project_outside_colliders`, which is implemented by `NewtonMPMManager._step_solver()` and is therefore silently ignored by the coupled solver.
   - Kamino's `max_contacts_per_world`, automatic `use_fk_solver` selection, and articulation validation, which are implemented in `NewtonKaminoManager._build_solver()` before `_create_solver()`.

   I think the generic entry contract needs a nested-solver lifecycle/step adapter, or unsupported manager-only options should be rejected explicitly.

2. **Coupled Kamino entries lose the required pre-step FK behavior** (`coupler.py:101-103`). The outer manager sets `_needs_fk_before_step` only for MPM, while standalone Kamino sets it unconditionally because resets update joint state and Kamino consumes authoritative body state. Kamino entries using internal collision detection, or used as proxy destinations, can therefore step from stale body transforms. Including `KaminoSolverCfg` in this predicate is necessary, although closed-loop Kamino also needs an entry-specific FK/reset hook rather than the outer manager's generic `eval_fk`.

3. **A proxy collision factory returning `None` can leave the destination with no contacts** (`coupler.py:84-88`, `coupler_cfg.py:118`). The public type permits `Callable[[ModelView], CollisionPipeline | None]`, and Newton interprets a `None` result as "use the outer contacts." However, proxy destinations are unconditionally excluded when deciding whether to create the outer collision pipeline. If the remaining entries use internal collision handling, no outer contact buffer exists. The outer pipeline should remain enabled when this fallback is possible, or the factory contract should be narrowed and validated.

4. **Incorrect intermediate type** (`coupler.py:287`). `_build_proxy_coupled_solver()` declares `proxies: list[SolverCoupledProxy.Proxy]`, but its callers pass `list[NewtonCoupler._ResolvedProxy]`, and the method accesses `proxy.config`. This should be `list[_ResolvedProxy]`.

5. **MPM is classified as requiring external contacts** (`coupler.py:64-70`). Implicit MPM handles collision internally, but `_requires_external_contacts()` falls through to `True`, causing an unnecessary global collision pipeline/pass when MPM is a proxy source.

The focused tests pass, but they replace both real coupled-solver implementations with recorder classes. A minimal headless proxy construct/reset/step test (and ideally an ADMM equivalent) would cover model-view construction, contact preparation, reset behavior, and the actual pinned Newton constructor contract; those paths can currently regress while all new tests remain green.

I also ran `./isaaclab.sh -f` successfully and the focused coupler suite passed (`24 passed`) using the Newton commit pinned by this PR.

mmichelis:
> High-level comment. Should we promote VBD and the coupler to isaaclab rather than contrib?

VBD I would absolutely agree to move into core isaaclab, though I would do this after we have deformables/cables outside of contrib so we have concrete examples in core isaaclab to test for VBD. For the coupling I would probably give a similar reasoning, if we moved this into core isaaclab instead of contrib now, we don't have examples to show that aren't contrib-only so far. My ordering for moving from contrib to isaaclab: deformables/cables -> VBD -> Coupling

source/isaaclab_tasks/changelog.d/feat-proxycoupledsolver.minor.rst
```diff
@@ -0,0 +1,26 @@
...
+  ``lift_franka_soft`` from ``deformable`` to ``object`` to align with the
+  rigid lift task. Affects ``Isaac-Lift-Soft-Franka-v0`` and the cloth
+  variant: scene entry ``scene.deformable`` -> ``scene.object``, command
+  ``deformable_pose`` -> ``object_pose``, and MDP functions
+  ``deformable_ee_distance``, ``deformable_lifted``,
+  ``deformable_com_goal_distance``, ``deformable_com_in_robot_root_frame``,
+  ``deformable_com_below_minimum``, ``deformable_outside_table_bounds``,
+  ``DeformableSampledPointsInRobotRootFrame`` -> ``object_*`` /
+  ``ObjectSampledPointsInRobotRootFrame``. Update env configs, checkpoints,
+  and RL configs accordingly.
```
AntoineRichard: This entry documents a rename that isn't in the PR:

> Renamed the deformable scene entity and its MDP terms in ``lift_franka_soft`` from ``deformable`` to ``object``

`franka_soft_env_cfg.py` still declares `deformable: DeformableCfg = DeformableCfg()` (line 216), all eight `SceneEntityCfg("deformable")` uses and the `mdp.deformable_*` terms are unchanged, and `object_pose` / `scene.object` appear nowhere on the branch. As written, the compiled changelog would ship false migration guidance for a breaking change that never happened. Either the rename is missing from the diff, or this bullet is stale from an earlier iteration and should be deleted.

It also cites `Isaac-Lift-Soft-Franka-v0` — the registered id is `Isaac-Lift-Soft-Franka` (`franka_soft/__init__.py:15`).

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
```diff
@@ -207,6 +204,114 @@ The rigid solver parameters still matter. For example, MJWarp's ``nconmax`` and
...
+(rigid, e.g. MJWarp) and a destination entry (soft, e.g. VBD), then exposes
+selected source bodies to the destination solver as *proxies* — virtual copies
+that the destination collides against. Contact feedback is returned to the
+source solver as lagged impulses. This typically scales better than the
+alternating coupling when only a small set of rigid bodies (e.g. the fingers of
+a gripper) actually needs to touch the deformable, since the bulk of the
+articulation is solved purely by MJWarp without seeing the particle contacts.
+
+The Franka soft-body task ships a ``newton_mjwarp_vbd_proxy`` preset (the new
+default for ``Isaac-Lift-Soft-Franka-v0``) that demonstrates the typical
```
AntoineRichard: There is no registered task with a `-v0` suffix — the gym id is `Isaac-Lift-Soft-Franka` (`franka_soft/__init__.py:15`); the rest of this doc (lines 42, 91, 97) uses the correct id. Same issue on the two demo commands below (lines 306, 312), which fail as written.

```suggestion
default for ``Isaac-Lift-Soft-Franka``) that demonstrates the typical
```

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
```diff
@@ -207,6 +204,114 @@ The rigid solver parameters still matter. For example, MJWarp's ``nconmax`` and
...
+    :start-at: self.sim.physics.newton_mjwarp_vbd_proxy.solver_cfg.scene_cfg
+    :end-at: self.sim.physics.default.solver_cfg.scene_cfg
+    :dedent: 8
+
+Try the demo:
+
+.. code-block:: bash
+
+    # zero-agent visual smoke test (default preset is now the proxy-coupled one)
+    ./isaaclab.sh -p scripts/environments/zero_agent.py --task Isaac-Lift-Soft-Franka-v0 --num_envs 1 --visualizer kit
```
AntoineRichard:
```suggestion
    ./isaaclab.sh -p scripts/environments/zero_agent.py --task Isaac-Lift-Soft-Franka --num_envs 1 --visualizer kit
```

AntoineRichard (commented):
I ran a multi-pass review of this PR (guidelines/correctness, docstring accuracy, pure-Python test coverage, silent-failure analysis, and API/type design), with `coupler.py` as the focus. Docstrings and config fields were cross-checked against the pinned Newton commit (`c7ae7c7`). Detailed findings are inline with the relevant lines; the short version:

**Must fix**
- The `isaaclab_tasks` changelog fragment documents a breaking `deformable` -> `object` rename that is not on the branch (inline).
- Docs and changelog reference `Isaac-Lift-Soft-Franka-v0`; the registered id has no `-v0` (inline, with suggestions).

**Main themes in the inline comments**
- *Silent resolution gaps*: joints with exactly one owned endpoint vanish from every entry; an entry that owns nothing is accepted; there is no diagnostic surface for unassigned model elements.
- *Nested-entry holes*: a nested `CouplerCfg` silently constructs a plain `SolverVBD`; a nested `model_cfg` is silently ignored; Kamino entries lose `_build_solver`-only behavior (`max_contacts_per_world`, FK-before-step).
- *Unchecked forwarding*: `_matching_config_values` reintroduces the silent-drift class `checked_apply` was built to kill, and roughly half the physical fields' forwarding is unasserted in the tests.
- Assorted docstring accuracy items and two test-quality fixes.

**Verified strengths** — things I checked that hold up well:
- Newton at the pin loudly validates ownership overlap, duplicate entry names, proxy/pair name references, and scalar ranges, so the cfg docstring contracts *are* enforced upstream (though the repo's build tests stub Newton construction, so nothing here pins that contract — see inline).
- Every `CouplerAdmmCfg` field docstring matches `SolverCoupledADMM.Config` at the pin exactly, and units are consistently annotated.
- The Proxy/ADMM-by-subclass config design, the `_create_solver`/`_filter_solver_kwargs` refactor, and the pure-Python test harness (recording stubs reusing the real Newton dataclasses) are all excellent; the `NewtonShapeCfg` default-parity regression test is exactly the right guard against upstream drift.
- Pre-commit passes on all files, and the late "coupler" rename left no stale references anywhere I could find.

Overall the structure is right — the requests above harden the existing shape rather than change it.

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
```diff
@@ -207,6 +204,114 @@ The rigid solver parameters still matter. For example, MJWarp's ``nconmax`` and
...
+
+    # scripted pick-and-lift via state machine
+    ./isaaclab.sh -p scripts/environments/state_machine/lift_franka_soft.py --num_envs 1
+
+    # explicitly select the alternating-substep preset instead
+    ./isaaclab.sh -p scripts/environments/zero_agent.py --task Isaac-Lift-Soft-Franka-v0 --num_envs 1 presets=newton_mjwarp_vbd
```
AntoineRichard:
```suggestion
    ./isaaclab.sh -p scripts/environments/zero_agent.py --task Isaac-Lift-Soft-Franka --num_envs 1 presets=newton_mjwarp_vbd
```

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+
+        joints: list[int] = []
+        if entry_cfg.include_child_joints and int(model.joint_count):
+            body_set = set(bodies)
+            parents = model.joint_parent.numpy()
+            joints = [
+                joint
+                for joint, child in enumerate(model.joint_child.numpy())
+                if int(child) in body_set and (int(parents[joint]) < 0 or int(parents[joint]) in body_set)
+            ]
```
AntoineRichard, line 242: A joint whose child is owned here but whose parent is owned by *no* entry falls through both this filter and `_validate_no_cross_entry_proxy_joints` (which only raises when the two endpoints are owned by *different* entries). Such a joint is owned by nobody and never integrated — a partially selected articulation silently loses joints (e.g. selecting only the gripper bodies detaches the gripper as a free body), in both Proxy and ADMM modes, with no diagnostic.

Could this raise (proxy mode) or at least warn for any joint with exactly one owned endpoint? Relatedly, the `include_child_joints` docstring in `coupler_cfg.py` documents only the "child selected" half of this condition — see comment there.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+        particles: list[int]
+
+    @staticmethod
+    def _requires_external_contacts(solver_cfg: NewtonSolverCfg) -> bool:
+        """Return whether a sub-solver expects contacts from Newton's collision pipeline."""
+        if isinstance(solver_cfg, MJWarpSolverCfg):
+            return not solver_cfg.use_mujoco_contacts
+        if isinstance(solver_cfg, KaminoSolverCfg):
+            return not solver_cfg.use_collision_detector
+        return True
```
AntoineRichard, line 69: The fall-through `return True` covers `MPMSolverCfg`, but implicit MPM never consumes external pipeline contacts — standalone `NewtonMPMManager._build_solver` sets `_needs_collision_pipeline = False`, and `NewtonCfg.collision_cfg`'s docstring states implicit MPM treats rigid geometry as colliders internally. A proxy config with a non-destination MPM entry therefore enables a shared collision pipeline its solver ignores.

Either add an explicit `MPMSolverCfg → False` branch, or amend the docstring to say unknown solver types conservatively return `True`.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @classmethod
+    def _build_solver(cls, model: Model, solver_cfg: CouplerCfg) -> None:
+        """Resolve ownership and construct the selected coupled solver."""
+        scene_cfg = solver_cfg.scene_cfg
+
+        resolved_entries = [cls._resolve_entry(model, entry, scene_cfg) for entry in solver_cfg.entries]
+        entries = [cls._build_entry(entry) for entry in resolved_entries]
```
AntoineRichard: The module currently has no diagnostic surface at all (`logging` isn't even imported), and two silent shapes fall out of resolution:

1. **An entry that owns nothing is accepted.** Forget `all_particles=True` on the soft entry and the VBD sub-solver is constructed over an empty view — the cloth never moves and nothing complains (Newton at the pinned `c7ae7c7` doesn't reject empty ownership either, and the per-spec no-match errors only fire *inside* the loop over specs, so an empty selector list bypasses them).
2. **Partially matching selectors leave the remainder unassigned and unsimulated.** Since non-assignment is documented as intentional on `CouplerCfg`, an accidental one is indistinguishable from a deliberate one.

Suggest raising when an entry resolves to no bodies, no particles, no joints, and no shapes, plus an info-level per-entry ownership summary (bodies/particles/joints/shapes counts and a count of unassigned elements) at build time. Optionally a `strict` flag that raises when anything is unassigned.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+
+        if isinstance(solver_cfg, CouplerProxyCfg):
+            proxies = [cls._resolve_proxy(model, proxy, scene_cfg) for proxy in solver_cfg.proxies]
+            cls._validate_no_cross_entry_proxy_joints(model, {entry.config.name: entry for entry in resolved_entries})
```
AntoineRichard: Two related points:

1. Duplicate entry names silently collapse in this dict comprehension before Newton's own duplicate-name validation runs at construction one line below, so `_validate_no_cross_entry_proxy_joints` can miss a cross-entry joint for the duplicated name. A cheap cross-check at the top of `_build_solver` (entries non-empty, names unique, every proxy/pair `source`/`destination` and `joint_proximal_destination_entries` entry resolves, `source != destination`) would give IsaacLab-quality error messages instead of Newton-internal ones.
2. Newton at the pinned `c7ae7c7` does validate all of these plus ownership overlap — I checked — but the build tests monkeypatch Newton construction, so nothing in this repo pins that contract across Newton bumps. Either the IsaacLab-side check or one test exercising the real Newton validation would close that.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+        NewtonManager._supports_contact_sensors = False
+        NewtonManager._needs_collision_pipeline = needs_collision_pipeline
+        NewtonManager._needs_fk_before_step = any(
+            isinstance(entry.config.solver_cfg, MPMSolverCfg) for entry in resolved_entries
+        )
+        if NewtonManager._report_contacts:
+            raise NotImplementedError(
+                "Newton contact sensors are not yet supported by coupled solvers because contact forces live "
+                "in per-entry buffers. Remove the contact sensor."
+            )
```
AntoineRichard, line 123: These five process-global class attributes are assigned before the `_report_contacts` guard can raise, so an abort leaves `NewtonManager` partially configured. The guard depends on nothing computed in this method — moving it to the top makes it fail before any global mutation (and before the expensive solver construction).

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+                f"CouplerCfg subclass {type(solver_cfg).__name__!r} is not supported; "
+                "use CouplerProxyCfg or CouplerAdmmCfg."
+            )
+
+        NewtonManager._use_single_state = False
+        NewtonManager._supports_contact_sensors = False
+        NewtonManager._needs_collision_pipeline = needs_collision_pipeline
+        NewtonManager._needs_fk_before_step = any(
+            isinstance(entry.config.solver_cfg, MPMSolverCfg) for entry in resolved_entries
+        )
```
AntoineRichard: Nested Kamino entries lose behavior that lives in `KaminoManager._build_solver` rather than `_create_solver`:

- `KaminoSolverCfg.max_contacts_per_world` is never applied to `model.rigid_contact_max`, so contact overflow silently drops contacts at runtime;
- `_needs_fk_before_step` is set only for MPM entries here, while the standalone Kamino manager always requires the FK pass;
- the one-articulation-per-env `RuntimeError` check is skipped.

Since Kamino entries are explicitly anticipated (`_requires_external_contacts` special-cases `KaminoSolverCfg`, and `test_coupler.py` parametrizes it), these should either be forwarded or Kamino entries rejected loudly until supported.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+            if isinstance(spec, str):
+                matched, _ = resolve_matching_names(f"(?:{spec})(?:/.*)?", labels, raise_when_no_match=False)
+                if not matched:
+                    raise ValueError(f"CouplerCfg {field}: body-label regex {spec!r} matched no Newton bodies.")
+                body_ids.extend(matched)
+                continue
+
+            asset_cfg = getattr(scene_cfg, spec.name, None) if scene_cfg is not None else None
+            if asset_cfg is None or not hasattr(asset_cfg, "prim_path"):
+                raise ValueError(f"CouplerCfg {field}: scene entity {spec.name!r} is not on the attached scene cfg.")
```
AntoineRichard: When `scene_cfg is None` (the user forgot the `__post_init__` wiring — easy, since it must be repeated on both `default` and each named preset), this message blames the scene entity: "scene entity 'robot' is not on the attached scene cfg" — but nothing is attached. Suggest splitting the branch: `scene_cfg is None` → "`CouplerCfg.scene_cfg` is unset; assign the environment scene cfg (typically in the env cfg's `__post_init__`)".

Two adjacent gaps in the same resolver:
- the non-`str` arm *assumes* `SceneEntityCfg` — `bodies=[1]` (which the PR's own tests pass to `CouplerProxyMappingCfg`) would die here with `AttributeError: 'int' object has no attribute 'name'`. An explicit `isinstance(spec, SceneEntityCfg)` check with a `TypeError` naming the two accepted forms would be friendlier;
- `SceneEntityCfg.body_ids` / `joint_names` / `preserve_order` are silently ignored — worth documenting that only `name` and `body_names` are honored (or warning when the others are set).

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+            body_ids.extend(asset_body_ids[index] for index in local_body_ids)
+
+        return list(dict.fromkeys(body_ids))
+
+    @classmethod
+    def _build_entry(cls, entry: _ResolvedEntry) -> SolverCoupled.Entry:
+        entry_cfg = entry.config
+
+        def solver_factory(model_view):
+            return entry_cfg.solver_cfg.class_type._create_solver(model_view, entry_cfg.solver_cfg)
```
AntoineRichard, line 334: Two silent holes in nested-entry construction:

1. **A nested `CouplerCfg` silently builds a `SolverVBD`.** `CouplerEntryCfg.solver_cfg: NewtonSolverCfg` admits a `CouplerCfg` (it *is* one), and `NewtonCoupler` inherits `_create_solver` from `NewtonVBDManager` (`vbd_manager.py:238`) without overriding it — so `class_type._create_solver` here returns a plain `SolverVBD` built from a coupler cfg. `CouplerProxyCfg.iterations` even name-matches a `SolverVBD` parameter and would flow through `_filter_solver_kwargs`. Suggest overriding `_create_solver` on `NewtonCoupler` to raise ("nested couplers are not supported"), or rejecting `isinstance(entry_cfg.solver_cfg, CouplerCfg)` in `_resolve_entry`.
2. **A nested `model_cfg` is silently ignored.** `CouplerEntryCfg(name="soft", solver_cfg=VBDSolverCfg(model_cfg=NewtonModelCfg(soft_contact_ke=...)))` — the natural transcription of the old standalone config — does nothing: `_apply_model_cfg` reads only the top-level `PhysicsManager._cfg.solver_cfg`, and `_filter_solver_kwargs` drops `model_cfg`. Contact stiffness silently stays at defaults. Raising/warning on a non-`None` `model_cfg` on an entry's solver cfg (it's model-global and can't be honored per entry) would save the first user a long debugging session.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+            joints=entry.joints,
+            shapes=entry.shapes,
+            substeps=entry_cfg.substeps,
+            in_place=entry_cfg.in_place,
+        )
+
+    @staticmethod
+    def _matching_config_values(target_type: type, config) -> dict:
+        """Return config values whose names match a target dataclass."""
+        return {field.name: getattr(config, field.name) for field in fields(target_type) if hasattr(config, field.name)}
```
AntoineRichard: This is intersection-based forwarding: any cfg field the Newton dataclass doesn't have is silently dropped. That's the exact drift class `checked_apply` (`isaaclab/utils/configclass.py:652`) was introduced for (the Newton `contact_margin` → `margin` rename), and the `NewtonShapeCfg.ke/kd/mu` forwarding added in this same PR uses the checked pattern correctly.

I verified every `Coupler*Cfg` field lands on the corresponding Newton dataclass at the pinned `c7ae7c7` — but the next Newton bump that renames, say, `mass_scale` or `rho` silently reverts every user's setting to Newton's default, with no error and changed physics. Suggest asserting completeness: every cfg field must land on the target unless it's on an explicit handled-elsewhere exclusion list (`bodies`, `particles`, `collision_pipeline`, `shape_material_*`, `entries`, `scene_cfg`, `class_type`, `contact_pairs`, ...).

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @classmethod
+    def _build_proxy_coupled_solver(
+        cls,
+        model: Model,
+        entries: list[SolverCoupled.Entry],
+        proxies: list[SolverCoupledProxy.Proxy],
```
AntoineRichard: The body reads `proxy.config` / `proxy.bodies`, and every caller passes resolved proxies — the annotation describes the *output* element type, not the input. The sibling `_apply_proxy_shape_overrides` is annotated correctly.

```suggestion
        proxies: list[_ResolvedProxy],
```

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -0,0 +1,354 @@
...
+        for proxy in proxies:
+            body_set = set(proxy.bodies)
+            shape_ids = [shape for shape, body in enumerate(shape_bodies) if int(body) in body_set]
+            for name in ("shape_material_ke", "shape_material_kd", "shape_material_mu", "shape_margin", "shape_gap"):
+                value = getattr(proxy.config, name)
+                array = getattr(model, name, None)
+                if value is not None and shape_ids and array is not None:
+                    values = array.numpy()
+                    values[np.asarray(shape_ids, dtype=np.int32)] = float(value)
+                    array.assign(values)
```
AntoineRichard: Two things here:

1. **Scope.** These writes land in the parent model's *global* arrays before entry views are built, so unless the `ModelView` copies them, the overrides also change these shapes' contact behavior inside the **source** entry's own solve. The cfg docstrings ("override for shapes on selected proxy bodies") read as destination-view-scoped — especially next to `mass_scale`'s explicit "in the destination view". If the global effect is intended, the docstrings should say so.
2. **Silence.** `if value is not None and shape_ids and array is not None` swallows two misconfigurations: an override set on a particles-only proxy (empty `shape_ids`) is silently ignored, and a missing/renamed model attribute (the `getattr(..., None)`) silently no-ops — same drift class as `_matching_config_values`. Raising for value-set-but-no-shapes and for a missing attribute would make both loud.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+
+    solver_cfg: NewtonSolverCfg = MISSING
+    """Configuration used to construct this entry's Newton solver."""
+
+    bodies: list[SceneEntityCfg | str] = field(default_factory=list)
+    """Bodies owned by this entry.
+
+    A :class:`~isaaclab.managers.SceneEntityCfg` selects bodies below the
+    corresponding scene asset and may narrow them with ``body_names``. A
+    string is treated as a regex matched against full Newton body labels.
```
AntoineRichard: The resolver wraps this as `(?:{spec})(?:/.*)?` (`coupler.py:229`), so a regex matching an ancestor path also selects every body beneath it. That's the intended prim-path semantics, but "matched against full Newton body labels" doesn't convey it — suggest adding "a match also selects all bodies below the matched path".

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+    particles: list[int] = field(default_factory=list)
+    """Parent-model particle indices owned by this entry."""
+
+    all_particles: bool = False
+    """Whether this entry owns every particle in the parent model."""
+
+    include_child_joints: bool = True
+    """Whether joints whose child body is selected are owned by this entry."""
```
AntoineRichard: > Whether joints whose child body is selected are owned by this entry.

The implementation (`coupler.py:148-155`) additionally requires the joint's parent to be the world (`< 0`) **or** also selected by the same entry; a joint whose child is selected but whose parent isn't ends up owned by nobody (see comment in `coupler.py`). Suggest documenting the full condition: "joints whose child body is selected and whose parent body is either the world or also selected".

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+    particles: list[int] = field(default_factory=list)
+    """Source particle indices exposed as proxies in the destination entry."""
+
+    mode: Literal["lagged", "staggered"] = "lagged"
+    """Proxy transfer mode passed to Newton's coupled-proxy solver."""
+
+    mass_scale: float = 1.0
+    """Scale applied to proxy effective mass and inertia in the destination view."""
```
AntoineRichard: Upstream, `SolverCoupledProxy.Proxy.mass_scale` also scales proxied *particle* mass (and this mapping supports `particles`) — suggest "Scale applied to proxy effective body mass/inertia and particle mass in the destination view".

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+
+    collide_interval: int | None = None
+    """Collision refresh interval when :attr:`collision_pipeline` is set.
+
+    ``None`` refreshes contacts on every proxy pass. Explicit values must be
+    positive integers.
+    """
```
AntoineRichard: "when :attr:`collision_pipeline` is set" is misleading: `NewtonCoupler` always installs a default pipeline when the factory is `None` (`CollisionPipeline(model_view, broad_phase="explicit")`, `coupler.py:297-298`) — the shipped preset relies on it (`collide_interval=5` with no pipeline set). So `collide_interval` is always live, and Newton's own documented no-pipeline behavior (destination consumes outer-level contacts) is unreachable through this API.

Suggest documenting the default on `collision_pipeline`, dropping the "when set" qualifier here, and — if the Newton default should stay reachable — a sentinel to opt out. Also: "Explicit values must be positive integers" is enforced only inside Newton; fine, but worth knowing that's where the error comes from.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+    """Optional factory for the proxy destination's collision pipeline."""
+
+    shape_material_ke: float | None = None
+    """Optional contact-stiffness override for shapes on selected proxy bodies [N/m]."""
+
+    shape_material_kd: float | None = None
+    """Optional contact-damping override for shapes on selected proxy bodies [N*s/m]."""
+
+    shape_material_mu: float | None = None
+    """Optional friction override for shapes on selected proxy bodies."""
```
AntoineRichard: Every other physical field here and the ADMM dimensionless params annotate units; `NewtonShapeCfg.mu` uses `[dimensionless]`.

```suggestion
    """Optional friction override for shapes on selected proxy bodies [dimensionless]."""
```

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
```diff
@@ -0,0 +1,234 @@
...
+
+    scene_cfg: InteractiveSceneCfg | None = None
+    """Scene cfg used to resolve :class:`~isaaclab.managers.SceneEntityCfg` selectors
+    to Newton bodies at solver-build time. Typically assigned from the env cfg's
+    ``__post_init__``."""
+
+
+@configclass
+class CouplerProxyCfg(CouplerCfg):
+    """Configuration for Newton's lagged-impulse virtual-proxy coupling."""
```
AntoineRichard: Upstream `SolverCoupledProxy.__init__` raises "Proxy coupling currently supports at most two solver entries" — worth a note in this docstring so users don't discover the limit at build time.

source/isaaclab_contrib/test/coupling/test_coupler.py
```diff
@@ -0,0 +1,650 @@
...
+        MJWarpSolverCfg(),
+        XPBDSolverCfg(),
+        FeatherstoneSolverCfg(),
+        KaminoSolverCfg(),
+        MPMSolverCfg(),
+        VBDSolverCfg(),
+    ],
+)
+def test_solver_config_manager_exposes_nested_factory(solver_cfg):
+    assert callable(solver_cfg.class_type._create_solver)
```
AntoineRichard: `callable(solver_cfg.class_type._create_solver)` passes even when a manager never overrides the hook — the abstract base's raising method is also callable, so this is a tautology. `assert solver_cfg.class_type._create_solver is not NewtonManager._create_solver` (the pattern already used in `test_newton_manager_abstraction.py:492`) would make it meaningful. As written, this is the only pure-Python check backing the MJWarp+VBD / Featherstone+VBD / XPBD+Kamino / implicit-MPM combination claim in the PR description.

source/isaaclab_contrib/test/coupling/test_coupler.py
```diff
@@ -0,0 +1,650 @@
...
+    )
+    monkeypatch.setattr(coupler.NewtonManager, "_solver", solver)
+    monkeypatch.setattr(coupler.NewtonManager, "_contacts", contacts)
+
+    NewtonCoupler._initialize_contacts()
+
+    assert events == [("initialize", None), ("prepare", contacts)]
+
+
+def test_algorithms_select_expected_outer_collision_pipeline(monkeypatch):
```
AntoineRichard: Two issues with this test:

1. `_build_solver` sets `NewtonManager._use_single_state = False` (`coupler.py:98`), which is not in the save/restore list here — it leaks into every later test in the session. Using `monkeypatch.setattr` for the class attributes (it already auto-restores the six patched methods) removes the hand-rolled `try/finally` entirely.
2. Five scenarios share mutable state in one test (`resolved_entries[0].config.solver_cfg` is reassigned mid-test), so a failure in scenario 2 masks scenarios 3-5. Parametrizing keeps the signal per case.

source/isaaclab_contrib/test/coupling/test_coupler.py
```diff
@@ -0,0 +1,650 @@
...
+    def auto_detect_contact_pairs(cls, entries):
+        return SolverCoupledADMM.auto_detect_contact_pairs(entries)
+
+    def __init__(self, *, model, entries, coupling):
+        self.model = model
+        self.entries = entries
+        self.coupling = coupling
+
+
+def test_admm_build_forwards_multiple_pairs_matching_and_proximal_options(monkeypatch):
```
AntoineRichard, line 720: This asserts `iterations`, the `joint_proximal_*` options, and the contact-matching options — but not `rho`, `gamma`, `baumgarte`, `joint_stiffness`, `joint_damping`, `joint_angular_stiffness`, `joint_angular_damping`. Because `_matching_config_values` drops non-matching names silently, an upstream Newton rename would make any of these silently fall back to Newton's default — exactly the regression this recording-stub pattern exists to catch, and it's seven more assertions in this test.

Same on the proxy side: `mode`, `mass_scale`, and `collide_interval` forwarding into `SolverCoupledProxy.Proxy` is never asserted (only `iterations` and `collision_pipeline` are).

source/isaaclab_newton/isaaclab_newton/physics/newton_manager_cfg.py
```diff
@@ -75,6 +75,28 @@ class NewtonShapeCfg:
     gap: float = 0.01
     """Default per-shape contact gap [m]. Newton's upstream default is ``None``."""
 
+    # Defaults mirror Newton's ShapeConfig defaults so an unspecified field is a no-op.
+    ke: float = 2.5e3
```
AntoineRichard, line 79: The class docstring above still says "Only fields Isaac Lab actually overrides are declared here; unspecified fields keep Newton's upstream default" — but `ke`/`kd`/`mu` are deliberately declared *at* Newton's defaults (per this comment and the new parity test). Suggest rewording to "fields Isaac Lab overrides or exposes for user overrides". The parity test itself is exactly the right guard — nice addition.

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
```diff
@@ -18,27 +18,64 @@
...
+
+    soft_contact_kd: float = 1.0e-2
+    """Body-particle contact damping [N*s/m]."""
+
+    soft_contact_mu: float = 0.5
+    """Body-particle contact friction coefficient [dimensionless].
+
+    Effective per-contact friction is ``sqrt(soft_contact_mu * shape_mu)``, where
+    ``shape_mu`` is the rigid shape's own friction coefficient (from its per-asset
+    material or :attr:`NewtonShapeCfg.mu` default), not set by this config.
```
AntoineRichard: `:attr:`NewtonShapeCfg.mu`` is referenced from `isaaclab_contrib.deformable`, but the class lives in `isaaclab_newton.physics` — the short target won't resolve across packages. Use `:attr:`~isaaclab_newton.physics.NewtonShapeCfg.mu``.

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
```diff
@@ -68,144 +103,65 @@ class VBDSolverCfg(NewtonSolverCfg):
...
+    """Manager class for the coupled MJWarp + VBD solver."""
 
     rigid_solver_cfg: MJWarpSolverCfg = MJWarpSolverCfg()
-    """Rigid-body sub-solver configuration for :class:`MJWarpSolverCfg`."""
+    """Rigid-body sub-solver configuration."""
 
     soft_solver_cfg: VBDSolverCfg = VBDSolverCfg(integrate_with_external_rigid_solver=True)
     """VBD sub-solver configuration for cloth/particle dynamics."""
 
     coupling_mode: str = "two_way"
```
AntoineRichard: `coupling_mode` would benefit from `Literal["one_way", "two_way"]` here (and `Literal["one_way", "two_way", "kinematic"]` at line 162) — the managers already enumerate and reject at runtime, so the `Literal` costs nothing and moves the invariant into the type. Relatedly, the managers validate the mode only at step time (`coupled_mjwarp_vbd_manager.py:85-87`); checking in `_build_solver` would fail at construction instead of potentially inside graph capture. (The new explicit `ValueError` instead of silently defaulting to two-way is a real improvement either way.)

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_soft_env_cfg.py
```diff
@@ -433,9 +471,11 @@ def __post_init__(self) -> None:
...
-        self.viewer.origin_type = "asset_root"
-        self.viewer.asset_name = "robot"
-        self.viewer.env_index = 0
-        self.viewer.eye = (1.25, -1.5, 0.75)
-        self.viewer.resolution = (1920, 1080)
+        # The coupled proxy solver resolves SceneEntityCfg selectors against the scene at
+        # solver-build time. Wire the scene on the named preset (used when selected by name)
+        # and on the ``default`` alias the resolver falls back to when no preset is given.
+        self.sim.physics.newton_mjwarp_vbd_proxy.solver_cfg.scene_cfg = self.scene
+        self.sim.physics.default.solver_cfg.scene_cfg = self.scene
```
AntoineRichard: This wiring works, but it's worth noting what it relies on: `self.scene` here is a `PresetCfg`, not the `InteractiveSceneCfg` that `CouplerCfg.scene_cfg` is annotated with — it only becomes valid because the preset resolver later walks into `sim.physics.*.solver_cfg.scene_cfg` and swaps the aliased instance. And it must be repeated for `default` plus every named preset (as the comment above documents); forgetting one leaves that preset resolving with `scene_cfg=None` until a selector fails at solver build.

Since resolution only ever reads `prim_path` + `body_names` from the scene, a self-contained selector (prim-path pattern + body names) would remove the post-hoc wiring entirely — might be worth considering before this API hardens.

source/isaaclab_newton/changelog.d/feat-proxycoupledsolver.rst
```diff
@@ -0,0 +1,10 @@
+Changed
+^^^^^^^
+
+* Extracted the repeated solver-kwargs filtering pattern from
+  :class:`~isaaclab_newton.physics.NewtonFeatherstoneManager`,
+  :class:`~isaaclab_newton.physics.NewtonMJWarpManager`, and
+  :class:`~isaaclab_newton.physics.NewtonXPBDManager` into a shared
+  :meth:`~isaaclab_newton.physics.NewtonManager._filter_solver_kwargs` helper,
```
AntoineRichard: Changelog guideline: entries shouldn't reference private symbols users can't call (`_filter_solver_kwargs`). Describing the user-visible effect — nested solver construction is now reusable by subclasses and the coupler, via the `_create_solver` hook that is the actual extension point — would fit better.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6, stacked on #5. #5 now forwards the proxy mapping config directly into Newton; #6 keeps that direct path but checks every authored field against the pinned Newton dataclass, and applies the same completeness check to ADMM so a renamed field fails loudly instead of silently falling back to a default.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. Resolution now rejects a joint with exactly one owned body endpoint for both Proxy and ADMM, tests both parent/child orientations, and documents the complete child-plus-parent ownership condition.

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
maxkra15: Fixed in mmichelis/IsaacLab#6: the zero-agent command now uses `Isaac-Lift-Soft-Franka`.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: this is a suboptimal part of the current design: the inheritance intentionally reuses VBD’s deformable lifecycle, but it conflates generic coupling with VBD-specific behavior even though solver selection remains entry-driven. A cleaner follow-up would have `NewtonCouplerManager` inherit directly from `NewtonManager` and introduce an extensible per-solver `CouplerEntrySupport` API, allowing current and future solvers to declare their builder, contact, and runtime requirements without changing the coupler

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_soft_env_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The VBD guide now has a decision table comparing alternating MJWarp+VBD and proxy coupling across interaction model, advantages, trade-offs/limits, and when to choose each.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupled_manager.py
maxkra15: The old `coupled_manager.py` implementation is now outdated. #5 removes the proxy-resolution argument piping, while mmichelis/IsaacLab#6 consolidates the remaining checks into construction-time validation because those checks cover lifecycle holes that Newton cannot validate (nested `model_cfg`, Kamino/MPM manager hooks, partial joints, and contact fallback).

source/isaaclab_tasks/changelog.d/feat-proxycoupledsolver.minor.rst
maxkra15: Addressed in mmichelis/IsaacLab#6: the stale deformable-to-object migration bullet is deleted, and the valid task name is used everywhere. The task remains `deformable`-named, so no migration is claimed.

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
maxkra15: Fixed in mmichelis/IsaacLab#6: the prose now uses the registered `Isaac-Lift-Soft-Franka` ID without `-v0`.

docs/source/overview/core-concepts/physical-backends/newton/using-vbd-solver.rst
maxkra15: Fixed in mmichelis/IsaacLab#6: the alternating-preset command now uses `Isaac-Lift-Soft-Franka`.

source/isaaclab_tasks/isaaclab_tasks/manager_based/manipulation/lift_franka_soft/mdp/observations.py
```diff
@@ -11,98 +11,96 @@
 from typing import TYPE_CHECKING
 
 import torch
-import warp as wp
```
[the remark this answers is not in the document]
maxkra15: This thread is outdated: the private cross-module helper import was removed from the current #5834 head. The coupling follow-up in mmichelis/IsaacLab#6 does not reintroduce it.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. `_requires_external_contacts(MPMSolverCfg)` now returns `False`; the docstring also states that unknown solver configs conservatively opt in, and the MPM behavior is covered by a focused test.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. `NewtonCouplerManager._create_solver` now rejects recursive nesting, `_validate_config` rejects nested `CouplerCfg`, and any entry-level non-`None` `model_cfg` raises with guidance to put model-global parameters on the outer coupler.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. An entry resolving to no bodies, particles, joints, or shapes now raises. Build-time info logs report each entry's ownership counts plus aggregate unassigned counts; unassigned elements remain supported intentionally, so this does not turn deliberate partitioning into an error.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed by #5. The proxy-local material/margin/gap override fields and parent-model writes are removed entirely. Proxy shapes use the source shapes' authored material values (per-asset or `NewtonShapeCfg` defaults); any future per-shape override API can be reviewed separately from coupling.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. Isaac Lab now validates non-empty/unique names, references, directions, ranges, and ownership overlap before construction. The new CPU runtime tests also construct the real pinned Proxy and ADMM implementations, so the Newton-side contract is no longer hidden behind recorder stubs.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed by #5. The intermediate resolved-proxy wrapper was removed; resolved selectors are written into `CouplerProxyMappingCfg` and the config is forwarded directly to `SolverCoupledProxy.Proxy`. mmichelis/IsaacLab#6 adds completeness checking around that direct path.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6 by rejecting Kamino entries at construction. The error names the missing manager-specific FK/reset lifecycle; this avoids silently skipping `max_contacts_per_world`, articulation validation, FK, and closed-loop reset behavior until a nested lifecycle adapter exists.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The `bodies` docs now state that a full-label regex matching an ancestor also selects descendant body labels below that path.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The intersection filter is replaced by checked dataclass forwarding: every Isaac Lab config field must exist on the target Newton dataclass or be on an explicit handled-elsewhere set. Proxy and ADMM schema drift now fails loudly.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The contact-sensor guard now runs before selector resolution, solver construction, or any `NewtonManager` global mutation; a regression test asserts every relevant class slot remains unchanged on rejection.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
maxkra15: Addressed in #5 and mmichelis/IsaacLab#6. Raw integer proxy body IDs are accepted and range-checked; unknown selector types raise a clear `TypeError`; an unset `scene_cfg` has its own actionable error; `preserve_order` is honored; and unsupported asset-local `SceneEntityCfg` fields are rejected rather than ignored.

source/isaaclab_contrib/test/coupling/test_coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. All touched `NewtonManager` slots use `monkeypatch` for automatic restoration, and the Proxy cases are parametrized independently (MJWarp internal/external, MPM, and destination outer-contact fallback).

source/isaaclab_newton/isaaclab_newton/physics/newton_manager_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The class docstring now says Isaac Lab declares fields that it overrides or exposes for user overrides, while unrepresented fields retain Newton defaults.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The two-entry limit is documented on `CouplerProxyCfg` and validated by Isaac Lab before invoking Newton.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. `mass_scale` now explicitly covers proxy body mass/inertia and particle mass in the destination view.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The docstring now gives the full condition (selected child and world-or-same-entry parent), and construction rejects a joint with exactly one owned endpoint so it cannot silently disappear.

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6 with the fully qualified public target `~isaaclab_newton.physics.NewtonShapeCfg.mu`; the full Sphinx build succeeds.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: This field is removed by #5 together with the other proxy-local material overrides, so the missing-unit issue no longer applies. Friction is configured on the source shape/per-asset material or through the dimensionless `NewtonShapeCfg.mu` default.

source/isaaclab_contrib/test/coupling/test_coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The test now compares each manager's `_create_solver.__func__` against the abstract base implementation, and construction performs the same concrete-factory check. Kamino was removed from the supported list and is rejected until its lifecycle is preserved.

source/isaaclab_contrib/test/coupling/test_coupler.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The ADMM test sets and asserts all seven fields listed here plus every proximal/matching option. The Proxy test now asserts non-default `mode`, `mass_scale`, `collide_interval`, factory behavior, and iterations; a schema-drift test checks completeness.

source/isaaclab_contrib/isaaclab_contrib/coupling/coupler_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. The default local explicit-broadphase factory is documented; setting the field itself to `None` now deliberately selects shared outer contacts; `collide_interval` requires a local factory; and a configured factory returning `None` is rejected.

source/isaaclab_newton/changelog.d/feat-proxycoupledsolver.rst
maxkra15: Addressed in mmichelis/IsaacLab#6. The fragment now describes the user-visible effect—reusable nested solver construction from typed configs—and no longer names the private `_filter_solver_kwargs` helper.

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6. Both config fields use the requested `Literal` types, and both managers reject invalid modes at the first executable line of `_build_solver`, before constructors or shared-state mutation. Regression tests verify that ordering.

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_soft_env_cfg.py
maxkra15: Addressed in mmichelis/IsaacLab#6 using the self-contained option you suggested. The shipped entry uses `/World/envs/env_.*/Robot`; the proxy bodies use full hand/finger label regexes; and both post-hoc `scene_cfg` assignments are removed.

maxkra15:
I put the full review-hardening follow-up in mmichelis/IsaacLab#6 (stacked on #5 and targeting the same `mym/coupledsolver` branch).

The top-level lifecycle findings are covered as follows:

- Kamino entries now fail at construction until their manager-specific build/FK/reset contract can be preserved.
- MPM `project_outside_colliders=True` is rejected, MPM entries must be `in_place`, MPM no longer requests external contacts, and only fixed-grid MPM is graph-capturable.
- CPU MJWarp is also rejected because its global reset state needs the standalone manager's reset-mask guard.
- Proxy collision factories are narrowed/checked; explicit `collision_pipeline=None` retains shared outer contacts, while a callable returning `None` fails loudly.
- The real pinned Newton Proxy and ADMM implementations now have CPU construct/contact/reset/step tests rather than recorder-only coverage.

The follow-up also covers the ownership, reference, selector, checked-forwarding, task wiring, coupling-mode, documentation, and changelog findings. I replied on each inline thread with the specific change and left the threads unresolved for reviewer verification.

Local validation: 69 focused tests passed, full pre-commit passed, the full Sphinx build passed, and the finalized feature diff passes the changelog gate against `develop`. The draft PR description includes the full issue-by-issue scope and the stacked-fragment caveat.

AntoineRichard (approved):

PR #5839, opened 2026-05-28
Fix bin_packing.py and Refactor with Wrapper APIs

YizeWang:
# Description

This PR updates the `scripts/demos/bin_packing.py` demo to use the current `RigidObjectCollection` body pose and velocity APIs instead of manipulating flat PhysX root-view tensors directly.

The demo now keeps body poses and body velocities in separate env-major buffers, avoiding the deprecated combined state pattern. It also switches mass updates to the collection-level `set_masses_index()` API and uses `body_link_pos_w`, `body_link_pose_w`, and `body_com_vel_w` data accessors for out-of-bounds recovery.

Fixes `bin_packing.py`.

## Type of change

- Documentation update

## Screenshots

bin_packing_10s

## Testing

- `./isaaclab.sh -p scripts/demos/bin_packing.py --num_envs 32`
- `./isaaclab.sh -f`

## Checklist

### DIFF
--- a/scripts/demos/bin_packing.py
+++ b/scripts/demos/bin_packing.py
@@ -159,7 +159,7 @@ def reset_object_collections(
         noise: If True, apply pose and velocity noise before writing.
 
     Returns:
-        None: This function updates ``states`` and the underlying PhysX view in-place.
+        None: This function updates ``view_states`` and the underlying PhysX view in-place.
     """
     rigid_object_collection: RigidObjectCollection = scene[asset_name]
     flat_states = view_states.view(-1, view_states.shape[-1])
@@ -283,7 +283,7 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
     cached_spawn_poses[..., :3] += scene.env_origins.view(-1, 1, 3)
     spawn_w = default_state_w.clone()
 
-    groceries_mask_helper = torch.arange(num_objects, device=device).view(1, num_objects)
+    groceries_mask_helper = torch.arange(num_objects * num_envs, device=device) % num_objects
     # Precompute a helper mask to toggle objects between active and cached sets.
     # Precompute XY bounds [[x_min,y_min],[x_max,y_max]]
     bounds_xy = torch.as_tensor(BIN_XY_BOUND, device=device, dtype=spawn_w.dtype)
@@ -295,7 +295,7 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
             count = 0
             # Randomly choose how many groceries stay active in each environment.
             num_active_groceries = torch.randint(MIN_OBJECTS_PER_BIN, num_objects, (num_envs, 1), device=device)
-            groceries_mask = (groceries_mask_helper < num_active_groceries).unsqueeze(-1)
+            groceries_mask = (groceries_mask_helper.view(num_envs, -1) < num_active_groceries).unsqueeze(-1)
             spawn_w[..., :7] = cached_spawn_poses * (~groceries_mask) + active_spawn_poses * groceries_mask
             # Retrieve positions
             with Timer("[INFO] Time to reset scene: "):
--- a/scripts/demos/bin_packing.py
+++ b/scripts/demos/bin_packing.py
@@ -145,13 +145,13 @@ def reset_object_collections(
     scene: InteractiveScene,
     asset_name: str,
     poses: torch.Tensor,
-    vel: torch.Tensor,
+    vels: torch.Tensor,
     view_ids: torch.Tensor,
     noise: bool = False,
 ) -> None:
     """Apply poses and velocities to a subset of a collection, with optional noise.
 
-    Updates ``poses`` and ``vel`` in-place for ``view_ids`` and writes them
+    Updates ``poses`` and ``vels`` in-place for ``view_ids`` and writes them
     to the PhysX view for the collection ``asset_name``. When ``noise`` is True, adds
     uniform perturbations to pose (XYZ + Euler) and velocities using ``POSE_RANGE`` and
     ``VELOCITY_RANGE``.
@@ -160,16 +160,16 @@ def reset_object_collections(
         scene: Interactive scene containing the collection.
         asset_name: Key in the scene (e.g., ``"groceries"``) for the RigidObjectCollection.
         poses: Env-major body poses [m, rad], shape ``(num_envs, num_bodies, 7)``.
-        vel: Env-major body velocities [m/s, rad/s], shape ``(num_envs, num_bodies, 6)``.
-        view_ids: 1D tensor of env-major flattened indices into ``poses`` and ``vel`` to update.
+        vels: Env-major body velocities [m/s, rad/s], shape ``(num_envs, num_bodies, 6)``.
+        view_ids: 1D tensor of env-major flattened indices into ``poses`` and ``vels`` to update.
         noise: If True, apply pose and velocity noise before writing.
 
     Returns:
-        None: This function updates ``poses``, ``vel``, and the underlying PhysX view in-place.
+        None: This function updates ``poses``, ``vels``, and the underlying PhysX view in-place.
     """
     rigid_object_collection: RigidObjectCollection = scene[asset_name]
     flat_poses = poses.view(-1, poses.shape[-1])
-    flat_velocities = vel.view(-1, vel.shape[-1])
+    flat_velocities = vels.view(-1, vels.shape[-1])
     selected_poses = flat_poses[view_ids]
     positions = selected_poses[:, :3]
     orientations = selected_poses[:, 3:7]
--- a/scripts/demos/bin_packing.py
+++ b/scripts/demos/bin_packing.py
@@ -144,14 +144,14 @@ class MultiObjectSceneCfg(InteractiveSceneCfg):
 def reset_object_collections(
     scene: InteractiveScene,
     asset_name: str,
-    body_poses: torch.Tensor,
-    body_velocities: torch.Tensor,
+    poses: torch.Tensor,
+    vel: torch.Tensor,
     view_ids: torch.Tensor,
     noise: bool = False,
 ) -> None:
     """Apply poses and velocities to a subset of a collection, with optional noise.
 
-    Updates ``body_poses`` and ``body_velocities`` in-place for ``view_ids`` and writes them
+    Updates ``poses`` and ``vel`` in-place for ``view_ids`` and writes them
     to the PhysX view for the collection ``asset_name``. When ``noise`` is True, adds
     uniform perturbations to pose (XYZ + Euler) and velocities using ``POSE_RANGE`` and
     ``VELOCITY_RANGE``.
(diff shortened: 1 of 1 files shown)


scripts/demos/h1_locomotion.py
```diff
@@ -85,6 +86,7 @@ def __init__(self):
         """Initializes environment config designed for the interactive model and sets up the environment,
         loads pre-trained checkpoints, and registers keyboard events."""
         agent_cfg: RslRlOnPolicyRunnerCfg = cli_args.parse_rsl_rl_cfg(TASK, args_cli)
+        agent_cfg = handle_deprecated_rsl_rl_cfg(agent_cfg, metadata.version("rsl-rl-lib"))
```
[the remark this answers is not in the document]
YizeWang: The following scripts use the same pattern as here.
- scripts/reinforcement_learning/rsl_rl/play_rsl_rl.py
- scripts/reinforcement_learning/rsl_rl/play.py
- scripts/reinforcement_learning/rsl_rl/train_rsl_rl.py
- scripts/reinforcement_learning/rsl_rl/train.py

scripts/demos/bin_packing.py
```diff
@@ -280,11 +293,12 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
     # Offset poses into each environment's world frame.
     active_spawn_poses[..., :3] += scene.env_origins.view(-1, 1, 3)
     cached_spawn_poses[..., :3] += scene.env_origins.view(-1, 1, 3)
-    active_spawn_poses = groceries.reshape_data_to_view(active_spawn_poses)
-    cached_spawn_poses = groceries.reshape_data_to_view(cached_spawn_poses)
-    spawn_w = groceries.reshape_data_to_view(default_state_w).clone()
+    active_spawn_poses = reshape_data_to_view_3d(groceries, active_spawn_poses, 7, device)
```
kellyguo11: `reshape_data_to_view_3d` - is this an API that would still make sense as part of RigidObjectCollection? the previous `reshape_data_to_view` seems to be part of the collection class. +@AntoineRichard

scripts/demos/bin_packing.py
```diff
@@ -296,16 +310,19 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
...
             spawn_w[:, :7] = cached_spawn_poses * (~groceries_mask) + active_spawn_poses * groceries_mask
             # Retrieve positions
             with Timer("[INFO] Time to reset scene: "):
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[~groceries_mask.view(-1)])
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[groceries_mask.view(-1)], noise=True)
                 # Vary the mass and gravity settings so cached objects stay parked.
                 random_masses = torch.rand(groceries.num_instances * num_objects, device=device) * 0.2 + 0.2
-                groceries.root_view.set_masses(random_masses.cpu(), view_indices.cpu())
-                groceries.root_view.set_disable_gravities((~groceries_mask).cpu(), indices=view_indices.cpu())
+                random_masses_wp = wp.from_torch(random_masses.cpu().contiguous(), dtype=wp.float32)
```
kellyguo11: are all the .contiguous() calls necessary? do we know if they impact performance?

scripts/demos/bin_packing.py
```diff
@@ -189,8 +199,11 @@ def reset_object_collections(
     view_states[view_ids, :7] = torch.concat((positions, orientations), dim=-1)
     view_states[view_ids, 7:] = new_velocities
 
-    rigid_object_collection.root_view.set_transforms(view_states[:, :7], indices=view_ids)
-    rigid_object_collection.root_view.set_velocities(view_states[:, 7:], indices=view_ids)
+    view_states_wp = wp.from_torch(view_states.contiguous(), dtype=wp.float32)
+    view_ids_wp = wp.from_torch(view_ids.contiguous().to(dtype=torch.int32), dtype=wp.int32)
+
+    rigid_object_collection.root_view.set_transforms(view_states_wp[:, :7], indices=view_ids_wp)
```
kellyguo11: if we switch from using the `root_view` directly to using APIs from RigidObjectCollection to set transforms like `write_body_link_pose_to_sim_index()`, would we be able to avoid a lot of the torch/warp interop?

scripts/demos/bin_packing.py
```diff
@@ -141,6 +142,15 @@ class MultiObjectSceneCfg(InteractiveSceneCfg):
     )
 
 
+def reshape_data_to_view_3d(
+    rigid_object_collection: RigidObjectCollection, data: torch.Tensor, data_dim: int, device: str
+) -> torch.Tensor:
+    """Reshape env-major ``(num_envs, num_bodies, data_dim)`` data to body-major view order."""
+
+    data_wp = wp.from_torch(data.contiguous(), dtype=wp.float32)
+    return wp.to_torch(rigid_object_collection.reshape_data_to_view_3d(data_wp, data_dim, device=device))
```
AntoineRichard: This should not be needed

scripts/demos/bin_packing.py
```diff
@@ -189,8 +199,11 @@ def reset_object_collections(
     view_states[view_ids, :7] = torch.concat((positions, orientations), dim=-1)
     view_states[view_ids, 7:] = new_velocities
 
-    rigid_object_collection.root_view.set_transforms(view_states[:, :7], indices=view_ids)
-    rigid_object_collection.root_view.set_velocities(view_states[:, 7:], indices=view_ids)
+    view_states_wp = wp.from_torch(view_states.contiguous(), dtype=wp.float32)
+    view_ids_wp = wp.from_torch(view_ids.contiguous().to(dtype=torch.int32), dtype=wp.int32)
```
AntoineRichard: Do not rely on states, this is outdated

scripts/demos/bin_packing.py
AntoineRichard: I agree it should work

scripts/demos/bin_packing.py
AntoineRichard: They should not be needed anymore

AntoineRichard: You shouldn't need to convert to warp

scripts/demos/bin_packing.py
```diff
@@ -296,16 +310,19 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
...
             with Timer("[INFO] Time to reset scene: "):
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[~groceries_mask.view(-1)])
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[groceries_mask.view(-1)], noise=True)
                 # Vary the mass and gravity settings so cached objects stay parked.
                 random_masses = torch.rand(groceries.num_instances * num_objects, device=device) * 0.2 + 0.2
-                groceries.root_view.set_masses(random_masses.cpu(), view_indices.cpu())
-                groceries.root_view.set_disable_gravities((~groceries_mask).cpu(), indices=view_indices.cpu())
+                random_masses_wp = wp.from_torch(random_masses.cpu().contiguous(), dtype=wp.float32)
+                view_indices_cpu_wp = wp.from_torch(view_indices.cpu().contiguous().to(torch.int32), dtype=wp.int32)
+                groceries.root_view.set_masses(random_masses_wp, view_indices_cpu_wp)
```
AntoineRichard: We have a built in method to set the mass

scripts/demos/bin_packing.py
```diff
@@ -314,7 +331,8 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
         sim.step()
 
         # Bring out-of-bounds objects back to the bin in one pass.
-        xy = groceries.reshape_data_to_view(groceries.data.object_pos_w - scene.env_origins.unsqueeze(1))[:, :2]
+        body_pos_b = groceries.data.body_link_pos_w.torch - scene.env_origins.unsqueeze(1)
+        xy = reshape_data_to_view_3d(groceries, body_pos_b, 3, device)[:, :2]
```
AntoineRichard: Consider using the built on method

AntoineRichard (changes requested):

scripts/demos/bin_packing.py
YizeWang: Yes, I found `set_masses_index()` instead, which takes in `torch.Tensor`.

scripts/demos/bin_packing.py
YizeWang: Yes, I found `set_masses_index()` instead. However, I did not find a built-in method for `set_disable_gravities`. Is there one?

scripts/demos/bin_packing.py
YizeWang: `reshape_data_to_view_3d` requires `data` to be `wp.array`, which is actually `torch.Tensor` at the call locations.

scripts/demos/bin_packing.py
```diff
@@ -296,16 +295,15 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
...
+            groceries_mask = (groceries_mask_helper.view(num_envs, -1) < num_active_groceries).unsqueeze(-1)
+            spawn_w[..., :7] = cached_spawn_poses * (~groceries_mask) + active_spawn_poses * groceries_mask
             # Retrieve positions
             with Timer("[INFO] Time to reset scene: "):
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[~groceries_mask.view(-1)])
                 reset_object_collections(scene, "groceries", spawn_w, view_indices[groceries_mask.view(-1)], noise=True)
                 # Vary the mass and gravity settings so cached objects stay parked.
-                random_masses = torch.rand(groceries.num_instances * num_objects, device=device) * 0.2 + 0.2
-                groceries.root_view.set_masses(random_masses.cpu(), view_indices.cpu())
-                groceries.root_view.set_disable_gravities((~groceries_mask).cpu(), indices=view_indices.cpu())
```
YizeWang, line 308: `set_disable_gravities` does not have a wrapper API yet. This line can be removed without an end-to-end behavioral impacts because the position and velocity will be reset in the out-of-bounds process. I argue we should avoid this low-level data operation to make the extension to other backends cleanly.

scripts/demos/bin_packing.py
```diff
@@ -154,15 +154,16 @@ def reset_object_collections(
...
+        view_states: Flat env-major tensor (N, 13) with [x, y, z, qx, qy, qz, qw, lin(3), ang(3)] in world frame.
+        view_ids: 1D tensor of env-major flattened indices into ``view_states`` to update.
         noise: If True, apply pose and velocity noise before writing.
 
     Returns:
         None: This function updates ``view_states`` and the underlying PhysX view in-place.
     """
     rigid_object_collection: RigidObjectCollection = scene[asset_name]
-    sel_view_states = view_states[view_ids]
+    flat_states = view_states.view(-1, view_states.shape[-1])
```
YizeWang: Previous `view_states` was reshaped to 2D data before passing into `reset_object_collections`. This refactor PR takes `view_states` of `[num_env, num_body, 13]` instead.

scripts/demos/bin_packing.py
AntoineRichard: No there is none, we don't provide wrappers for everything.

scripts/demos/bin_packing.py
AntoineRichard: It's ok to keep it this way

scripts/demos/bin_packing.py
AntoineRichard: Can we avoid states altogether? Doesn't solved the 2 / 3d thing, but let's stop using states. It's teaching users to do things incorrectly!

scripts/demos/bin_packing.py
```diff
@@ -314,11 +312,13 @@ def run_simulator(sim: SimulationContext, scene: InteractiveScene) -> None:
         sim.step()
 
         # Bring out-of-bounds objects back to the bin in one pass.
-        xy = groceries.reshape_data_to_view(groceries.data.object_pos_w - scene.env_origins.unsqueeze(1))[:, :2]
-        out_bound = torch.nonzero(~((xy >= bounds_xy[0]) & (xy <= bounds_xy[1])).all(dim=1), as_tuple=False).flatten()
+        xy = (groceries.data.body_link_pos_w.torch - scene.env_origins.unsqueeze(1))[..., :2]
+        out_bound = torch.nonzero((~((xy >= bounds_xy[0]) & (xy <= bounds_xy[1])).all(dim=-1)).view(-1)).flatten()
         if out_bound.numel():
             # Teleport stray objects back into the active stack to keep the bin tidy.
-            reset_object_collections(scene, "groceries", spawn_w, out_bound)
```
YizeWang, line 321: It could be an issue in the original implementation - passing `spawn_w` into `reset_object_collections`, because `spawn_w` is only updated at every other 250 frame, not frame-by-frame. It used to reset a wrong object!

scripts/demos/bin_packing.py
YizeWang: Done.

scripts/demos/bin_packing.py
```diff
@@ -142,29 +142,37 @@ class MultiObjectSceneCfg(InteractiveSceneCfg):
...
 def reset_object_collections(
-    scene: InteractiveScene, asset_name: str, view_states: torch.Tensor, view_ids: torch.Tensor, noise: bool = False
+    scene: InteractiveScene,
+    asset_name: str,
+    poses: torch.Tensor,
+    vels: torch.Tensor,
+    view_ids: torch.Tensor,
+    noise: bool = False,
 ) -> None:
-    """Apply states to a subset of a collection, with optional noise.
```
YizeWang, line 147: According to @AntoineRichard's suggestion, we should not deprecate states because they are not contiguous anymore in the latest IsaacLab. We want to instruct users to modify positions and velocities separately instead of combining them as states.

scripts/demos/bin_packing.py
YizeWang: Thanks for confirming this!

AntoineRichard (approved):
LGTM thanks for the changes. Still would be good to track the issue with `reshape_data_to_view` only supporting warp.

PR #5840, opened 2026-05-28
Adds public v1.0 benchmark schema for Odin (and others)

AntoineRichard:
# Description

Promote the JSON bundle schema produced by the standalone benchmark scripts under `scripts/benchmarks/` into a real public-API module, `isaaclab.benchmark.schema`. Until now there was no single place in lab that defined the shape of `training.json` / `startup.json`, even though three lab scripts emit it and downstream tooling (e.g. the in-tree Odin evaluation harness, but also any future regression-comparison or CI-perf tool) consumes it.

This PR is the **schema foundation** for a follow-up that rewrites `benchmark_rsl_rl.py`, `benchmark_skrl.py`, and `benchmark_startup.py` to emit schema-v1 bundles via this module. Kept separate so the type surface lands first and the script rewrite reads as a pure switch-over.

## What's in here

**New public package `isaaclab.benchmark`:**

- `isaaclab/benchmark/schema.py` — frozen dataclasses for `TrainingBundle`, `StartupBundle`, and their building blocks (`Versions`, `Hardware`, `Runtime`, `Resources`, `Learning`, `GpuDeviceInfo`, `MeanStd`, `MeanStdPeak`, `StartupPhase*`, `CProfileFunction`, `RunIdentity`, `StartupRunIdentity`). Plus `SCHEMA_VERSION = "1.0"` and `write_bundle_file()`.
- `isaaclab/benchmark/__init__.py` — re-exports the public surface, so callers can write `from isaaclab.benchmark import TrainingBundle`.

**Recorder peak fields:**

- `GPUInfoRecorder` now reports per-device peak memory and peak GPU utilisation alongside the existing mean/std rows (`"GPU Memory Used peak"`, `"GPU Utilization peak"`).
- `MemoryInfoRecorder` reports peak RSS / VMS / USS the same way.
- Existing mean/std rows are unchanged.
- Peak rows are always emitted (initialised to `0.0`) so consumers see a consistent key set regardless of whether any sample was recorded.

**Tests:**

- `source/isaaclab/test/benchmark/test_schema.py` — round-trip tests for `TrainingBundle` and `StartupBundle`, plus a sanity check that the package re-exports stay in sync with `schema.py`.
- `source/isaaclab/test/benchmark/test_recorders.py` — extended with peak-tracking and "peak is zero before any record" cases for both GPU and memory recorders.

**Changelog fragment:** `source/isaaclab/changelog.d/antoiner-feat-benchmark-schema.rst`.

## Compatibility

- Strictly additive. No public API removed, no existing recorder rows renamed or removed.
- The legacy benchmark scripts (`scripts/benchmarks/benchmark_*.py`) are untouched in this PR — they keep emitting their existing per-backend output format until the follow-up rewires them.

Fixes # (no issue)

## Type of change

## Screenshots

N/A — types + recorder fields, no UI surface.

## Checklist

- [x] I have made corresponding changes to the documentation (`benchmarking.rst` covering the public schema ships with the script-rewrite follow-up where it's user-facing; this PR is the type-only foundation)


zoctipus (approved):

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+def write_bundle_file(bundle: Any, path: str) -> None:
+    """Write a bundle dataclass to disk as schema-v1 JSON.
+
+    Creates the parent directory if missing. Uses ``indent=2`` for readability;
+    payloads are small (~10 KB training.json, ~50 KB startup.json).
+
+    Args:
+        bundle: A dataclass instance to serialise. Typically
+            :class:`TrainingBundle` or :class:`StartupBundle`; any frozen
+            dataclass tree composed of primitives, lists, and dicts works.
```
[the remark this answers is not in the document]
AntoineRichard: The asymetry is done on purpose. The gpu_util_pct peak value is not interesting on its own (max is 100% regardless). Whereas the memory is it directly dictates the amount of memory required to run something.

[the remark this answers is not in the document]
AntoineRichard: Can you update my grade then? I don't like to have 3/5. Or something else is wrong @greptile

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+    series_per_iter: list[float] | None
+
+
+@dataclass(frozen=True)
+class Learning:
+    """Learning curves for a training run, plus their EMA smoothing factor."""
+
+    ema_alpha: float
+    reward: LearningCurve
+    ep_length: LearningCurve
```
yts-nv: i would like to also put "success_rate" in final Training result, since it is used for out correctness result.

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+class Runtime:
+    """Aggregated runtime metrics for a training run."""
+
+    startup_phase_times_s: StartupPhaseTimes
+    iterations_completed: int
+    total_wall_time_s: float
+    steps_per_iteration: int
+    iteration_time_s: MeanStd
+    env_steps_per_s: MeanStd
+    iterations_per_s: MeanStd
```
yts-nv: here i think we need to track FPS as well?

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+import dataclasses
+import json
+import os
+from dataclasses import dataclass
+from typing import Any, Literal
+
+SCHEMA_VERSION = "1.0"
+
+Framework = Literal["rsl_rl", "skrl"]
+Backend = Literal["physx", "newton"]
```
StafaH: physics backend vs rendering backend?

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+    git_dirty: bool
+
+
+@dataclass(frozen=True)
+class RunIdentity:
+    """Identity of a training run."""
+
+    run_id: str
+    framework: Framework
+    backend: Backend
```
StafaH: Similar comment to above, backend is too generic and will need to split between physics and rendering

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+    num_envs: int
+    max_iterations: int
+    start_time_utc: str
+    end_time_utc: str
+    duration_s: float
+    status: RunStatus
+
+
+@dataclass(frozen=True)
+class StartupPhaseTimes:
```
StafaH: simplify to -> StartupTime?

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+
+SCHEMA_VERSION = "1.0"
+
+Framework = Literal["rsl_rl", "skrl"]
+Backend = Literal["physx", "newton"]
+RunStatus = Literal["completed", "interrupted", "crashed"]
+
+
+@dataclass(frozen=True)
+class MeanStd:
```
StafaH: Can we simplify below and this class into just one? peak can remain empty when it is not recorded?

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+@dataclass(frozen=True)
+class StartupConfig:
+    """CLI configuration captured in a :class:`StartupBundle`."""
+
+    top_n: int
+    whitelist: str | None
+
+
+@dataclass(frozen=True)
+class StartupRunIdentity:
```
StafaH: Why do we need a seperate class instead of just using the original but not filling in the fields not needed (None of 0)

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -0,0 +1,258 @@
...
+    task: str
+    seed: int
+    start_time_utc: str
+    end_time_utc: str
+    duration_s: float
+    status: RunStatus
+
+
+@dataclass(frozen=True)
+class StartupBundle:
```
StafaH: Similar comment to StartupRunIdentity

StafaH (approved):

StafaH:
Can we also add:
- Path to checkpoint?
- Path to video/gif?

source/isaaclab/isaaclab/benchmark/schema.py
fatimaanes: +1. And sensor datatype (rgb/ depth / albedo) and resolution (64 / 128 /  256). These help with the product primary grouping. Maybe a small config block on RunIdentity that holds all four? gives every consumer one table place to key on.

source/isaaclab/isaaclab/benchmark/schema.py
fatimaanes: +1

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — added `success_rate` to `TrainingBundle` (the final training result). Heads-up: the schema moved to `isaaclab.test.benchmark`.

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — `Runtime` now carries `collection_fps` and `total_fps` (the Collection/Total FPS the scripts emit); for no-learning runtime runs they are equal.

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Good call — replaced the single `backend` with a `RunConfig` block on `RunIdentity`: `physics_backend`, `rendering_backend`, `preset`

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — same `RunConfig` change as the thread above; the single `backend` is now physics + rendering + sensor fields.

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — renamed `StartupPhaseTimes` → `StartupTime`.

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — merged into a single `MeanStd` with an optional `peak` (left empty where a peak is not meaningful, e.g. utilization).

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Done — removed `StartupRunIdentity`; `StartupBundle.run` now reuses `RunIdentity` with `num_envs`/`max_iterations`/`framework` left unset.

source/isaaclab/isaaclab/benchmark/schema.py
AntoineRichard: Kept `StartupBundle` separate, intentionally. Unlike the identity (just two optional scalars), the bundle bodies do not overlap — startup carries `phases`/`config` and has no `runtime`/`resources`/`learning`, and vice-versa. Merging would make all of those Optional and force consumers to None-check to learn the run kind — the same optional-field shape we deliberately avoided by keeping `RuntimeBundle`/`TrainingBundle` separate. Happy to revisit if you would prefer one unified bundle.

AntoineRichard:
Re: checkpoint + video/gif paths — done, added `checkpoint_path` and `video_path` to `TrainingBundle`.

fatimaanes (approved):

PR #5849, opened 2026-05-28
Fix cartpole-camera frame-stacking performance regression

jmart-nv:
# Description

Frame-stacking was added to cartpole-camera by #5574 to make the task solvable on the Newton+Warp backend. Single-frame RGB observations don't carry pole velocity, so a single-image policy can only recover ~100/300 mean episode reward. On PhysX+RTX, implicit damping in the dynamics plus temporal antialiasing in the renderer leak enough adjacent-frame correlation that a single-frame policy still hits ~296. Newton+Warp has neither, so explicit frame stacking is required for the task to be solvable on this backend.

The implementation regressed throughput by 42% (54.6 → 31.7 FPS) and inflated peak VRAM 2.7× (3.13 → 8.37 GB) at R256, 1024 envs. PR #5574 added a `CircularBuffer` ring + a `stacked_image` MDP term that:

1. Stored frames as **float32** even when the camera output was uint8 → 4× ring memory and 4× shift bandwidth.
2. Ran the per-frame normalize (`x/255 - mean(x)`) **before** the buffer, forcing the buffer to be float32.
3. Re-rolled the buffer with `torch.roll` on every append → ring-sized temporary per env.step.
4. Built the channel-stacked policy obs via `permute + reshape + clone` → fresh `(B, H, W, K*C)` float32 tensor per env.step.
5. Inlined the normalize math at 5 call sites across DirectRLEnv and ManagerBasedEnv paths.

This PR rewrites the frame-stacking hot path; the result runs **24% faster than the pre-framestack baseline** with identical PPO convergence.

| Change | Why |
|---|---|
| **uint8 ring buffer.** Defer the `x/255 − mean` normalize past the frame-stack buffer for RGB-like data types. | Ring is 4× smaller; shift bandwidth per env.step drops 4×. |
| **Free contiguous stacked output.** New `CircularBuffer.stack_dim` arg + `.stacked` property arrange storage so the channel-stacked output is a contiguous view of internal storage. | Eliminates one `(B, H, W, K*C)`-sized allocation and a permute kernel per env.step. |
| **Drop `torch.roll` on append.** Replace with a front-to-back memmove. | One fewer ring-sized allocation and one fewer kernel launch per env.step. |
| **Fused Warp normalize kernel** (`normalize_image_uint8`). One kernel: `(uint8 / 255) − per-(batch, channel) mean → float32`. | Replaces upcast + reduce + subtract (three PyTorch passes) with one fused launch over the K-stacked input. |
| **Tiled int32 partial-sum reduction kernel** (`spatial_sum_uint8_tiled`). Warp kernel writes `(B, NUM_TILES, C)` int32 partials along H; PyTorch collapses the partial dim. | int32 accumulator is safe up to H·W·255 (128× headroom at R256) and ~3× faster than int64 / float32 on PyTorch's auto-tuned reduce path. C-innermost launch shape gives stride-1 reads on src's contiguous trailing dim. |
| **Shared `isaaclab.utils.images.normalize_camera_image` dispatch.** One helper used by both `image()` (DirectRLEnv) and `stacked_image.__call__` (ManagerBasedEnv); routes uint8 contiguous inputs through the Warp fast path and falls back to PyTorch otherwise. | Removes inline normalize math duplicated across 5 sites; the fast path picks itself up automatically. |
| **`image(clone=False)` opt-out.** Callers that immediately copy into their own storage skip the defensive clone. | Saves one obs-sized allocation per env.step at the consumer site (frame-stack buffer). |
| **Allocator-managed obs lifetime.** `_get_observations` allocates the normalize output fresh each call (no persistent `out=` buffer). | The RL trainer holds a reference to the previous-iteration observation until `record_transition` returns; reusing one buffer across env.step boundaries aliases the trainer's prior obs. Fresh allocation lets PyTorch's caching allocator hand out a different block while the prior is still referenced. |

## Type of change

## Results

R256 (256×256), 1024 envs, PPO via skrl 2.1.0, seed 42, on NVIDIA L40. Perf via non_rl measured over n=3 warm trials:

| Stage | FPS | step_ms | Peak VRAM | Mean reward |
|---|---|---|---|---|
| Pre-framestack baseline | 54.62 | 18.67 | 3.13 GB | ~102 (unsolvable — no temporal info) |
| Frame stacking from #5574 | 31.72 | 32.07 | 8.37 GB | 297.68 |
| This PR (pre-rebase) | 67.87 | 15.32 | 4.41 GB | 297.60 |
| **This PR (post-rebase)** | **69.50** | **14.94** | **4.57 GB** | **293.43** |

## Screenshots

pr_figure_v2

## Environment

NVIDIA L40 (48 GB), driver 570.158.01. Newton+Warp backend (`presets=newton_mjwarp,newton_renderer`). skrl 2.1.0 trainer.

## Checklist

- [ ] I have made corresponding changes to the documentation *(N/A - frame stacking already documented, these changes are implementation details)*

### DIFF
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -20,7 +20,7 @@ class CircularBuffer:
     The shape of the appended data is expected to be (batch_size, ...), where the first dimension is the
     batch dimension. Correspondingly, the shape of the ring buffer is (max_len, batch_size, ...).
 
-    When :paramref:`stack_dim` is set, the internal layout is rearranged so that :attr:`stacked`
+    When ``stack_dim`` is set, the internal layout is rearranged so that :attr:`stacked`
     returns the K frames merged into the chosen dim as a free contiguous view; :meth:`__getitem__`
     is disabled in this mode.
     """
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -39,7 +39,7 @@ class CircularBuffer:
                 :meth:`append`. Defaults to ``None`` (legacy layout).
 
         Raises:
-            ValueError: If ``max_len < 1`` or ``stack_dim == 0``.
+            ValueError: If the buffer size is less than one, or ``stack_dim == 0``.
         """
         if max_len < 1:
             raise ValueError(f"The buffer size should be greater than zero. However, it is set to {max_len}!")
@@ -52,7 +52,9 @@ class CircularBuffer:
 
         # CPU mirror of max_len; avoids a GPU sync via ``.item()`` on every property access.
         self._max_len_int: int = max_len
+        # max length tensor for comparisons
         self._max_len = torch.full((batch_size,), max_len, dtype=torch.int, device=device)
+        # number of data pushes passed since the last call to :meth:`reset`
         self._num_pushes = torch.zeros(batch_size, dtype=torch.long, device=device)
         # CPU gate; lets ``append`` skip a ``torch.any`` GPU sync on the steady-state path.
         self._need_reset: bool = True
@@ -140,6 +142,8 @@ class CircularBuffer:
         self._num_pushes[batch_ids_resolved] = 0
         self._need_reset = True
         if self._buffer is not None:
+            # set buffer at batch_id reset indices to 0.0 so that the buffer() getter returns
+            # the cleared circular buffer after reset.
             if self._stack_dim_internal is None:
                 self._buffer[:, batch_ids_resolved] = 0.0
             else:
@@ -242,5 +246,7 @@ class CircularBuffer:
 
         # Clamp to [0, ..] so batches with _num_pushes == 0 return the zeroed slot.
         valid_keys = torch.clamp(torch.minimum(key, self._num_pushes - 1), min=0)
+        # The buffer is stored oldest->newest along dimension 0, so the most
+        # recent item lives at the last index.
         index_in_buffer = (self._max_len_int - 1 - valid_keys).to(dtype=torch.long)
         return self._buffer[index_in_buffer, self._ALL_INDICES]
--- a/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
+++ b/source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
@@ -39,12 +39,14 @@ class CircularBuffer:
         self._device = device
         self._ALL_INDICES = torch.arange(batch_size, device=device)
 
-        # max length tensor for comparisons
+        # CPU mirror of max_len; avoids a GPU sync via ``.item()`` on every property access.
+        self._max_len_int: int = max_len
         self._max_len = torch.full((batch_size,), max_len, dtype=torch.int, device=device)
-        # number of data pushes passed since the last call to :meth:`reset`
         self._num_pushes = torch.zeros(batch_size, dtype=torch.long, device=device)
-        # the actual buffer for data storage
-        # note: this is initialized on the first call to :meth:`append`
+        self._pointer: int = -1
+        # CPU gate; lets ``append`` skip a ``torch.any`` GPU sync on the steady-state path.
+        self._need_reset: bool = True
+        # Lazily allocated on the first ``append``.
         self._buffer: torch.Tensor = None  # type: ignore
 
     """
@@ -64,7 +66,7 @@ class CircularBuffer:
     @property
     def max_length(self) -> int:
         """The maximum length of the ring buffer."""
-        return int(self._max_len[0].item())
+        return self._max_len_int
 
     @property
     def current_length(self) -> torch.Tensor:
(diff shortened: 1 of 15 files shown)


source/isaaclab/isaaclab/utils/images.py
```diff
@@ -0,0 +1,87 @@
...
+_RGB_LIKE_PREFIXES: tuple[str, ...] = ("rgb", "albedo", "simple_shading")
+_DEPTH_LIKE_PATTERNS: tuple[str, ...] = ("depth", "distance_to")
+_NORMALS_PREFIXES: tuple[str, ...] = ("normals",)
+
+
+def is_rgb_like(data_type: str) -> bool:
+    """Whether ``data_type`` is one of the RGB-like camera outputs (rgb, albedo, simple_shading_*).
+
+    Args:
+        data_type: The camera data-type string from ``sensor.data.output`` / ``TiledCameraCfg``.
```
pbarejko: `TiledCameraCfg` is deprecated, it should be `CameraCfg`.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -19,34 +19,52 @@ class CircularBuffer:
...
             device: The device used for processing.
+            stack_dim: If set, the buffer arranges its internal storage so :attr:`stacked` returns
+                the K stored frames merged into ``data.shape[stack_dim]`` of the appended data
+                as a free contiguous view. Negative indexing supported; ``0`` is invalid (batch
+                dim). Validation against the actual data rank is deferred to the first
+                :meth:`append`. Defaults to ``None`` (legacy layout).
 
         Raises:
-            ValueError: If the buffer size is less than one.
+            ValueError: If the buffer size is less than one, or ``stack_dim == 0``.
```
pbarejko, line 45: Few lines below you throw only when `is not None and stack_dim == 0`. `-1` means you are concatenating based on `W`, and to my understanding that is a legit axis to concatenate.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -19,34 +19,52 @@ class CircularBuffer:
...
         # number of data pushes passed since the last call to :meth:`reset`
         self._num_pushes = torch.zeros(batch_size, dtype=torch.long, device=device)
-        # the actual buffer for data storage
-        # note: this is initialized on the first call to :meth:`append`
+        # CPU gate; lets ``append`` skip a ``torch.any`` GPU sync on the steady-state path.
+        self._need_reset: bool = True
+        # Lazily allocated on the first ``append``.
         self._buffer: torch.Tensor = None  # type: ignore
 
+        self._stack_dim_arg: int | None = stack_dim
```
pbarejko, line 67: It took me some time to understand stacking terminology here. Why this is called stacking? This is like gluing frames side-by-side. Initially, I was under the impression that you actually have n-buffers represent stack, but what happens here is frame concatenation based on stack dimension?

source/isaaclab/isaaclab/utils/images.py
jmart-nv: Good catch! Fixed

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
jmart-nv: Right, any non-zero dim index is valid here, we only reject `0` at construction because we don't know `data.ndim` yet. Full range validation happens on the first call to `append()`. Clarified the docstring a bit and added an example.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
jmart-nv: Yes. "Frame stacking" is the conventional naming used for this technique. Under the hood, you're right that it's more of a concatenation. If we truly "stacked" on a new dimension, that would change the CNN's input shape, forcing a different architecture with higher complexity. Concatenating along the channel axis (and keeping K adjacent in ring buffer memory) allows us to avoid an expensive reshape when the policy asks for the obs.

pbarejko (approved):

jmart-nv:
Moving this back to draft while I rebase onto the cartpole refactor

jmart-nv:
After rebasing and refactoring to support the changes in #5930 which moved the channel dim to `BCHW` instead of `BHWC`.
- FPS shows further improvement at 69 fps vs 54 fps baseline
- VRAM shows marginal regression, up to 4.57 GB post-rebase vs 4.41GB pre-rebase (due to extra `permute()` each step)
- Still converges around ~295 reward

Also addressed @greptile's concerns and profiled the fixes.

Updated description with latest numbers and plot. Moving this back to PR.

source/isaaclab/isaaclab/utils/buffers/circular_buffer.py
```diff
@@ -83,7 +104,28 @@ def buffer(self) -> torch.Tensor:
...
+        the internal storage; callers must not mutate it.
+
+        Returns:
+            View of shape ``(batch_size, *frame_shape)`` with ``frame_shape[stack_dim]`` multiplied by ``max_length``.
+
+        Raises:
+            RuntimeError: If ``stack_dim`` was not set at construction.
+        """
+        if self._stack_dim_internal is None:
+            raise RuntimeError("stacked is only available when CircularBuffer was created with stack_dim set.")
```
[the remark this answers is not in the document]
jmart-nv: Made this more clear in the latest commit

source/isaaclab_tasks/isaaclab_tasks/direct/cartpole/cartpole_camera_env.py
```diff
@@ -94,30 +95,28 @@ def _pre_physics_step(self, actions: torch.Tensor) -> None:
     def _apply_action(self) -> None:
         self._cartpole.set_joint_effort_target_index(target=self.actions, joint_ids=self._cart_dof_idx)
 
-    def _get_observations(self) -> dict:
+    def _get_observations(self, clone: bool = True, normalize: bool = True) -> dict:
```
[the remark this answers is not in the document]
jmart-nv: Resolved during the rebase/refactor

PR #5896, opened 2026-06-01
Fix memory leak in ManagerBasedEnv.close

cvolkcvolk:
# Description

`ManagerBasedEnv.close()` does not release two categories of per-instance state, both of which then survive `close()` because gymnasium's wrapper chain (`OrderEnforcing` plus the `make()` registry) keeps the wrapped env reachable:

1. **`self.obs_buf`** — the dict of cached observation tensors populated by every `step()` / `reset()`. For image sensors at `(num_envs, H, W, C)` this is tens of MB per camera.
2. **`self.observation_space` / `self.action_space`** (and the vectorised-API `single_*_space` variants set by `ManagerBasedRLEnv`). Each `gym.spaces.Box` for an image observation at `(4, 720, 1280, 3)` float32 pins ~110 MB of bounds arrays (`low`, `high` as float32 + `bounded_below`, `bounded_above` as bool).

Net effect: ~one env's worth of `gym.spaces.Box` bounds arrays plus the last observation buffer are retained for the lifetime of the process even after `close()` returns. In evaluation loops that rebuild envs across tasks — or in any context where module reloads orphan multiple env instances against type-equality replacement — the leak compounds into multi-GB territory and ultimately OOMs the process.

No associated issue (found while debugging an evaluation-sweep OOM in a downstream project).

Fixes # (no associated issue)

## Repro & regression test

Two commits in this PR:

1. **The fix** in `ManagerBasedEnv.close` — clears `obs_buf`, nulls space attributes.
2. **A regression test** at `source/isaaclab/test/envs/test_manager_based_env_close_leak.py` that builds `Isaac-Cartpole-v0` via `gym.make`, runs reset + one step, calls `close()`, and asserts `obs_buf` is empty and the space attributes are `None`. Parametrised over `cuda:0` and `cpu`.

Verified the test:
- **Fails** on the pre-fix `close()` with `AssertionError: obs_buf was not cleared on close`.
- **Passes** with the fix applied.

A standalone runnable repro that demonstrates the leak magnitude visually (live `gym.spaces.Box` count and total bounds-array MB across cycles, with a `--patched` toggle) is in companion draft PR #5898. Run it as::

    ./isaaclab.sh -p source/isaaclab/test/envs/check_manager_based_env_close_leak.py
    # pre-fix: live Boxes=4, bounds-arrays held: 131.8 MB
    ./isaaclab.sh -p source/isaaclab/test/envs/check_manager_based_env_close_leak.py --patched
[...]

PR #5898, opened 2026-06-01
Add repro script for ManagerBasedEnv.close memory leak

cvolkcvolk:
# Description

Adds a runnable check script that reproduces the `ManagerBasedEnv.close` memory leak fixed by #5896. Useful for verifying the fix is in place and as a starting point for a future regression test.

The script builds the smallest manager-based RL task with an image observation (`Isaac-Cartpole-RGB-v0`), bumps its camera to 720×1280 so each `gym.spaces.Box` bounds array is large enough to be visible above RSS noise, and runs a tight construct → step → close loop. After each close it counts live `gym.spaces.Box` instances and sums the bytes pinned by their `low` / `high` / `bounded_below` / `bounded_above` numpy attributes.

Default invocation monkey-patches `ManagerBasedEnv.close` back to its pre-#5896 form, so the leak reproduces even when the installed IsaacLab is already patched. `--patched` leaves the installed `close()` in place.

Depends on (but does not block) #5896. The script works on either branch — the `--patched` flag toggles the close() body the script uses.

Fixes # (no associated issue — diagnostic script, no behaviour change)

## Repro output

Pre-fix (default):
```
--- cycle 04 ---
   after cycle 04  RSS=  5491.4 MB   live Boxes=  4   bounds-arrays held:   131.8 MB
```

Post-fix (`--patched`):
```
--- cycle 04 ---
   after cycle 04  RSS=  5438.0 MB   live Boxes=  0   bounds-arrays held:     0.0 MB
```

## Type of change

- Bug fix (non-breaking change which fixes an issue) — diagnostic / test asset for #5896

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A
- [ ] I have added tests that prove my fix is effective or that my feature works — This *is* a runnable repro; happy to convert to a pytest under `source/isaaclab/test/envs/` if maintainers prefer
- [x] I have updated the changelog and the corresponding version — `.skip` fragment added (test-only, no version bump)
- [ ] I have added my name to the `CONTRIBUTORS.md`

### DIFF
--- a/source/isaaclab/test/envs/check_manager_based_env_close_leak.py
+++ b/source/isaaclab/test/envs/check_manager_based_env_close_leak.py
@@ -1,0 +1,162 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Reproduces the ``ManagerBasedEnv.close`` memory leak fixed by upstream PR #5896.
+
+Builds the ``Isaac-Cartpole-RGB-v0`` task in a tight construct→step→close loop,
+then counts live :class:`gymnasium.spaces.Box` instances and sums the bytes they
+hold in their ``low`` / ``high`` / ``bounded_below`` / ``bounded_above`` numpy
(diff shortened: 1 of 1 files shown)


cvolkcvolk:
Closing — the formal regression test landed on #5896 itself (`test_manager_based_env_close_leak.py`). The standalone runnable repro was useful for diagnosing the leak internally but is redundant alongside the pytest.

PR #5915, opened 2026-06-02
Decouple physics-engine event terms into per-backend modules

ooctipus:
## Summary

- Splits the physics-property randomization event terms out of `isaaclab.envs.mdp.events` into a new `isaaclab.envs.mdp.physics_events` module: `randomize_rigid_body_material`, `randomize_rigid_body_scale`, `randomize_rigid_body_mass`, `randomize_rigid_body_inertia`, `randomize_rigid_body_com`, `randomize_rigid_body_collider_offsets`, `randomize_physics_scene_gravity`, `randomize_joint_parameters`, `randomize_fixed_tendon_parameters`, `randomize_actuator_gains` (plus the shared `randomize_prop_by_op` / `validate_scale_range` helpers). `events.py` now only holds backend-agnostic terms (resets, force/velocity, visual material/color).
- Relocates the backend-specific implementations into `isaaclab_<backend>.envs.mdp.physics_events` for PhysX, Newton, and OVPhysX. The core public terms resolve the active backend's implementation at runtime via a small private resolver, so `isaaclab.envs.mdp.events`/`physics_events` no longer import `pxr` or any backend.
- OVPhysX reuses the PhysX implementations for every term except material randomization, which has a dedicated no-op (the OVPhysX material tensor binding is not yet available).
- Deprecates `randomize_rigid_body_scale`: it remains PhysX-only (Newton raises `NotImplementedError`) and emits a `DeprecationWarning` recommending multi-asset spawning (`MultiAssetSpawnerCfg`).
- New extension `envs/mdp` packages follow the lazy `__init__.py` + `.pyi` stub convention.

## Test plan

- [ ] `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/assets/test_newton_actuators_physx.py`
- [ ] `./isaaclab.sh -p -m pytest source/isaaclab_newton/test/assets/test_newton_actuators_newton.py`
- [ ] Run an env that uses `randomize_rigid_body_material` / `randomize_joint_parameters` on both the PhysX and Newton backends and confirm parity with the previous behavior.
- [ ] Confirm `randomize_rigid_body_scale` raises on Newton and warns (deprecated) on PhysX.

> Note: changelog fragments are not yet included; happy to add per-package fragments if desired.


source/isaaclab/isaaclab/envs/mdp/physics_events.py
```diff
@@ -0,0 +1,989 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
hujc7, line 1: Is the rename better? following the existing convention, I think a events folder needs to be created within with it's can be more granularly structured.

source/isaaclab/isaaclab/envs/mdp/physics_events.py
```diff
@@ -0,0 +1,989 @@
...
+    try:
+        impl_cls = getattr(module, impl_name)
+    except AttributeError as exc:
+        raise ValueError(
+            f"The {backend!r} backend does not implement event term {impl_name!r} (looked in '{module.__name__}')."
+        ) from exc
+    return impl_cls(*args, **kwargs)
+
+
+def randomize_rigid_body_scale(
```
hujc7, line 78: does not seem to use backend  specific impl.

hujc7 (commented):
I am wondering how much work or if it would be possible to unify at backend level for the mdps needing specific implementation. If there's behavior difference, I think it could be explained in mdp description?

source/isaaclab/isaaclab/envs/mdp/physics_events.py
```diff
@@ -0,0 +1,989 @@
...
+        self.asset: RigidObject | Articulation = env.scene[self.asset_cfg.name]
+
+        if not isinstance(self.asset, (BaseRigidObject, BaseArticulation)):
+            raise ValueError(
+                f"Randomization term 'randomize_rigid_body_material' not supported for asset: '{self.asset_cfg.name}'"
+                f" with type: '{type(self.asset)}'."
+            )
+
+        # construct the backend-specific implementation via the converter
+        self._impl = _resolve_backend_event("RandomizeRigidBodyMaterial", cfg, env, self.asset, self.asset_cfg)
```
AntoineRichard, line 260: Why do we need a resolve_backend_event? Couldn't users just pick the event from the backend they want? I.e. we no longer keep this in IsaacLab and we leave it to the users to pick the right event. THis could make sense since these terms may not have the same contracts / definitions depending on the backend and trying to unify them may be more work than what it's worth?

PR #5916, opened 2026-06-02
Warp-first experimental envs implements lazy export + cfg implementation split

ooctipus:
## Summary

- Adds Warp-first experimental environments and MDP terms under `isaaclab_experimental` and `isaaclab_tasks_experimental` (direct + manager-based cartpole/humanoid/ant/reach/locomotion), with Warp-kernel implementations of actions, observations, rewards, terminations, and events.
- Adds lazy `__init__.py` + `.pyi` stubs (with explicit `__all__`) across the new/updated packages.

## Test plan

- [ ] `./isaaclab.sh -p -m pytest source/isaaclab_tasks_experimental/test`
- [ ] Train the experimental Warp cartpole/humanoid/ant configs on the Newton backend and confirm parity with the stable envs.
- [ ] `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/assets/test_newton_actuators_physx.py source/isaaclab_newton/test/assets/test_newton_actuators_newton.py`
- [ ] Verify `randomize_rigid_body_scale` raises on Newton and warns (deprecated) on PhysX.


source/isaaclab_experimental/isaaclab_experimental/managers/scene_entity_cfg.py
```diff
@@ -41,7 +44,7 @@ def resolve(self, scene: InteractiveScene):
         entity = scene[self.name]
 
         # -- Warp joint mask / ids for articulations
-        if isinstance(entity, BaseArticulation):
+        if hasattr(entity, "num_joints"):
```
hujc7: Is this also because of importing baseArticulation pulls extra dependencies? The only concern is that it's not very obvious when things can be done and cannot. Wondering if there's a way to generalize or referred in agents.md

source/isaaclab_experimental/isaaclab_experimental/managers/__init__.py
```diff
@@ -7,18 +7,14 @@
 
 This package is intended for experimental forks of manager implementations while
 keeping stable task configs and the stable `isaaclab.managers` package intact.
-"""
 
-from isaaclab.managers import *  # noqa: F401,F403
```
hujc7, line 12: This fallback is needed I think like that in mdp in .pyi?

source/isaaclab_experimental/isaaclab_experimental/envs/__init__.py
```diff
@@ -42,14 +42,8 @@
...
-from .manager_based_env_warp import ManagerBasedEnvWarp  # noqa: F401
-from .manager_based_rl_env_warp import ManagerBasedRLEnvWarp  # noqa: F401
-
-__all__ = [
-    "DirectRLEnvWarp",
-    "InteractiveSceneWarp",
-    "ManagerBasedEnvWarp",
-    "ManagerBasedRLEnvWarp",
-]
+from . import mdp
```
hujc7: Why this is different from others? agent suggests it should be in __all__.

source/isaaclab/isaaclab/envs/utils/camera_view.py
```diff
@@ -215,10 +218,6 @@ def prim_world_positions(
     Uses ``FrameView`` first so PhysX/Fabric-backed transforms are current; falls
     back to USD only if the backend view cannot be constructed.
     """
-    from pxr import UsdGeom
```
hujc7, line 228: Would it be an extra safeguard to keep those? Or with lazy import properly setup, it won't be a problem any more?

hujc7 (commented):
Thanks! 
Does it mean all of #5826 can be reverted?

source/isaaclab/isaaclab/envs/utils/camera_view.py
ooctipus: it won't be a problem if its top imported. Let's try not local import any packages or module

ooctipus:
> Thanks! Does it mean all of #5826 can be reverted?

yes that PR can be reverted

source/isaaclab_experimental/isaaclab_experimental/envs/__init__.py
ooctipus: nice catch it should be removed

source/isaaclab_experimental/isaaclab_experimental/managers/__init__.py
ooctipus: it seems we covered all managers? but just in case ill add it back

source/isaaclab_experimental/isaaclab_experimental/managers/scene_entity_cfg.py
ooctipus: you could check against cfg, for example 
if isinstance(entity.Cfg, BaseArticulation):

hujc7 (approved):

hujc7:
@ooctipus Is there a cherry-picked to release branch? I think it's needed.

kellyguo11:
> @ooctipus Is there a cherry-picked to release branch? I think it's needed.

looks like it was cherry picked in https://github.com/isaac-sim/IsaacLab/pull/5916

kellyguo11:
ah nvm was looking at the wrong branch. thanks @hujc7 let's get the cherry pick on merged

PR #5923, opened 2026-06-02
Fix pre-trained policy action grad leak

AntoineRichard:
# Description

Fixes a latent autograd leak in `PreTrainedPolicyAction.apply_actions`. When environments are stepped outside `torch.inference_mode()`, the low-level policy output can require gradients; copying that output into `self.low_level_actions` makes the persistent action buffer require gradients as well. Warp later rejects that tensor when accessing `__cuda_array_interface__`.

This runs the low-level policy under `torch.no_grad()` and detaches its output before copying it into the action buffer. No new dependencies are required.

@pascal-roth FYI for upstream review.

Fixes: N/A

## Type of change

## Screenshots

N/A.

## Checklist

## Verification

- `./isaaclab.sh -f` was run first. All non-LFS hooks passed, but `check-git-lfs-pointers` failed because `git-lfs` is not installed in this environment.
- `SKIP=check-git-lfs-pointers ./isaaclab.sh -f` passed.
- Focused inline check with normal grad mode enabled confirmed the policy output requires grad before the action term runs, while `low_level_actions.requires_grad`, `low_level_actions.grad_fn`, and the downstream action tensor remain detached after `apply_actions()`.
- No automated test was added for this change.

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/contrib/navigation/mdp/pre_trained_policy_action.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/contrib/navigation/mdp/pre_trained_policy_action.py
@@ -96,8 +96,7 @@ class PreTrainedPolicyAction(ActionTerm):
     def apply_actions(self):
         if self._counter % self.cfg.low_level_decimation == 0:
             low_level_obs = self._low_level_obs_manager.compute_group("ll_policy")
-            with torch.no_grad():
-                self.low_level_actions[:] = self.policy(low_level_obs).detach()
+            self.low_level_actions[:] = self.policy(low_level_obs)
             self._low_level_action_term.process_actions(self.low_level_actions)
             self._counter = 0
         self._low_level_action_term.apply_actions()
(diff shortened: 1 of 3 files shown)


source/isaaclab_tasks/isaaclab_tasks/contrib/navigation/mdp/pre_trained_policy_action.py
```diff
@@ -96,7 +96,8 @@ def process_actions(self, actions: torch.Tensor):
     def apply_actions(self):
         if self._counter % self.cfg.low_level_decimation == 0:
             low_level_obs = self._low_level_obs_manager.compute_group("ll_policy")
-            self.low_level_actions[:] = self.policy(low_level_obs)
+            with torch.no_grad():
```
pascal-roth: Are we stepping the simulation at any point that is not already wrapped in torch.no_inference() or torch.no_grad() ? IMO this should not be necessary

AntoineRichard:
@pascal-roth for approval

pascal-roth (approved):
Amazing thanks

PR #5925, opened 2026-06-02
Fix memory leak in ManagerBasedEnv.close

cvolkcvolk:
# Description

`ManagerBasedEnv.close()` leaves two kinds of per-instance state alive after it returns. Gymnasium's wrapper chain (`OrderEnforcing` plus the `make()` registry) keeps the wrapped env reachable, so anything the env still references at close stays in memory:

1. **`self.obs_buf`**: the dict of cached observation tensors from the last `step()` / `reset()`. For image sensors this is tens of MB per camera.
2. **`self.observation_space` / `self.action_space`** (plus the `single_*_space` variants on `ManagerBasedRLEnv`). Each image-observation `gym.spaces.Box` pins ~110 MB of bounds arrays (`low`, `high`, `bounded_below`, `bounded_above`) at shape `(4, 720, 1280, 3)`.

So roughly one env's worth of bounds arrays and the last observation buffer survive for the process lifetime. Evaluation loops that rebuild envs across tasks accumulate these and eventually OOM.

The fix clears `obs_buf` and nulls the space attributes in `close()`.

Found while debugging an evaluation-sweep OOM in a downstream project. No associated issue.

Fixes # (no associated issue)

## Testing

Regression test at `source/isaaclab/test/envs/test_manager_based_env_close_leak.py` builds `Isaac-Cartpole-v0` via `gym.make`, runs reset and one step, calls `close()`, then asserts `obs_buf` is empty and the space attributes are `None`. Parametrised over `cuda:0` and `cpu`. It fails on the pre-fix `close()` and passes with the fix.

## Type of change

## Checklist

- [ ] I have made corresponding changes to the documentation (N/A, internal lifecycle method)
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file (happy to add a fragment if maintainers want a patch-bump entry)
- [ ] I have added my name to the `CONTRIBUTORS.md` (happy to add if maintainers want)


source/isaaclab/test/envs/test_manager_based_env_close_leak.py
AntoineRichard, line 1: Do we need a test for this? I feel like it will never be exhaustive enough and would need constant updating.

AntoineRichard (commented):
Thanks for the PR! Could you fix the fragment failure. Usually claude/codex is pretty good at this :). Also can you port these fixes to the other envs?

source/isaaclab/test/envs/test_manager_based_env_close_leak.py
cvolkcvolk: It was meant as a repro test to tick the box for `I have added tests that prove my fix is effective or that my feature works` :) 
Agree that its not probably not the right test to merge. Will remove it again.

AntoineRichard:
@cvolkcvolk the direct and multi agent environments are likely also afflicted by this. Could you look into this? :D

cvolkcvolk:
@AntoineRichard PR covers both `DirectRLEnv` and `DirectMARLEnv`  now as well

AntoineRichard (approved):

PR #5933, opened 2026-06-03
[MGPU] App: default Kit renderer to one GPU

hujc7:
## 1. Summary

- **isaaclab**: makes the base Kit experiences use one renderer GPU by default; XR explicitly retains multi-GPU rendering.
- Keeps single-process renderer multi-GPU available through existing `kit_args`, avoiding overlap with training `--distributed`.
- Adds `ISAACLAB_FABRIC_USE_GPU_INTEROP=0|1` as an independent temporary CI workaround; when unset, the Kit default is preserved.

## 2. Opt-in and workaround

Enable single-process renderer multi-GPU explicitly:

```text
--kit_args "--/renderer/multiGpu/enabled=true --/renderer/multiGpu/autoEnable=true --/renderer/multiGpu/maxGpuCount=16"
```

The multi-GPU CI lane can temporarily disable Fabric GPU interop with `ISAACLAB_FABRIC_USE_GPU_INTEROP=0`. Explicit raw Kit settings and `kit_args` take precedence over the environment override. Remove the workaround after the underlying Kit/PhysX problem is fixed.

## 3. Validation

- [x] Merged current `develop` without conflicts.
- [x] Full `./isaaclab.sh -f` passed after the merge.
- [x] Five focused Fabric override and explicit-precedence tests passed.
- [x] Exact four-A10G diagnostics reproduced instability without the Fabric override while five explicit-false controls completed cleanly: https://github.com/isaac-sim/IsaacLab/pull/6300#issuecomment-4841829935


source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1060,6 +1060,32 @@ def _resolve_device_settings(self, launcher_args: dict):
...
+        # = true``, so each Kit process enumerates every visible GPU at
+        # startup. Under concurrent multi-GPU CI shards (``--gpus all`` per
+        # container, one Kit per non-default cuda device), that produces a
+        # shared GPU-interop context across sibling processes -- surfacing as
+        # ``[Error] [omni.physx.plugin] Stage X already attached`` mid-test and
+        # ``SimulationApp.close`` hanging in teardown (see
+        # https://github.com/isaac-sim/IsaacLab/issues/3475). The mitigation is
+        # to set ``renderer.multiGpu.enabled = false`` + ``maxGpuCount = 1`` so
+        # each Kit only touches its assigned GPU.
+        if os.environ.get("ISAACLAB_PIN_KIT_GPU", "0").lower() not in {"", "0", "false", "no", "off"}:
```
kellyguo11: is there a good reason why we shouldn't always have these set? this feels like could be necessary for multi GPU training as well.

AntoineRichard: Wouldn't that explicitely prevent multi-gpu rendering? But I agree with the default, maybe we should make this a default, and for people that want to do single process multi-gpu tell them to enable that flag?

AntoineRichard: Or we could add a flag? Might be  more visible than an environment variable?

kellyguo11: ya I think for our current workflows, having this as the default makes more sense

hujc7: I will test on if making this default will break anything

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -523,6 +532,12 @@ def add_app_launcher_args(parser: argparse.ArgumentParser) -> None:
             default=AppLauncher._APPLAUNCHER_CFG_INFO["device"][1],
             help='The device to run the simulation on. Can be "cpu", "cuda", "cuda:N", where N is the device ID',
         )
+        arg_group.add_argument(
+            "--multi_gpu",
```
kellyguo11: I think this might be confusing to users since it's not the same multi GPU flag we have for training (--distributed). users should be able to always set this through the `--kit_args` flag already

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1169,11 +1187,32 @@ def _resolve_anim_recording_settings(self, launcher_args: dict):
...
-        self._kit_args = []
-        if "kit_args" in launcher_args:
-            self._kit_args = [arg for arg in launcher_args["kit_args"].split()]
-            sys.argv += self._kit_args
+        self._kit_args = launcher_args.get("kit_args", "").split()
+
+        default_args = []
+        if not launcher_args["multi_gpu"]:
+            default_args = [
+                "--/renderer/multiGpu/enabled=false",
```
kellyguo11: we can probably put default settings directly in the .kit files?

hujc7: updated

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -33,6 +33,8 @@
 # import logger
 logger = logging.getLogger(__name__)
 
+_FABRIC_GPU_INTEROP_ENV = "ISAACLAB_FABRIC_USE_GPU_INTEROP"
```
kellyguo11, line 36: were we able to figure out why this was needed for CI? ideally I think this flag should always be on.

hujc7: Mainly the performance impact (agent claiming ~5% but needs to be confirmed). The problem currently only repros on CI instances when running tests in parallel, which makes it hard to investigate. The goal is to eventually remove those so currently it's setup as a WAR.

kellyguo11: ok let's create a ticket before merging this so that we keep track of the issue and remind us to remove this workaround.

kellyguo11 (approved):

PR #5885, opened 2026-05-30
Adds CI tests for template generator

kellyguo11:
# Description

Adds CI tests for the template generator workflow

Coverage includes:

- algorithm discovery for single-agent, multi-agent, and combined workflow inputs
- external and internal generation paths
- single-agent registry entries for rl_games, rsl_rl, skrl, and sb3
- multi-agent skrl registry entries for IPPO and MAPPO
- the important default skrl PPO key: skrl_cfg_entry_point
- generated VS Code launch configs passing --algorithm for every skrl agent

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #5941, opened 2026-06-03
[Fix] Make generated template projects import without Kit (no eager omni.ext)

hujc7:
## 1. Summary

- Generated template projects could not be imported headless (e.g. Gym registration before `SimulationApp`): `import <project>` failed with `ModuleNotFoundError: No module named 'omni.ext'`. The package root no longer imports the example UI extension, so it stays omni-free. (NVBug 6251247)
- Hardened the generator to fail loudly when a requested agent-config template is missing, instead of silently emitting a project whose Gym registration points at a non-existent config.
- Self-contained on `develop`. Includes the template-generator test scaffolding (also proposed in #5885, commit preserved) plus new regression coverage — 15 tests.

### 2.1 Generator (`tools/template`)
- `templates/extension/__init__ext`: drop the eager `from .ui_extension_example import *` from the package root.
- `templates/extension/config/extension.toml`: register `<name>.ui_extension_example` as its own `[[python.module]]` so Kit still loads the UI extension (Kit imports each `python.module` by name and scans it for `omni.ext.IExt`).
- `templates/extension/ui_extension_example.py`: add the missing `import omni.ui`.
- `generator.py`: raise (instead of silently `continue`-ing) when an `agents/<rl_library>_<algorithm>_cfg` template is missing.

### 2.2 Tests
- Existing scaffolding: algorithm discovery, single/multi-agent registry entry points, launch configs.
- New regressions: omni/pxr import-safety of the generated package, `extension.toml` UI-module registration, generated-`.py` `ast`/`compile` validity, no top-level backend imports in generated env modules, per-library agent-config completeness.

## 3. Design note: why not lazy-import `omni.ext`
- `omni.ext` / `ExampleExtension` are consumed only by Kit's extension manager (via `[[python.module]]`), never on the headless path. The correct fix is to not import it at the package root, not to defer it.

## 4. Out of scope / known limitations
- AMP on the generated single-agent template (NVBug 6257353) is intentionally left as a known limitation (tracked on the ticket), not changed here: AMP requires task-specific reference motion data (a `MotionLoader` + `collect_reference_motions` + AMP observations) that a generic generated env cannot provide — see `contrib/humanoid_amp`. PPO trains as-generated.
- The `free()` / `omni.UsdMdl TfNotice` startup crashes (NVBug 6257159, 6251258) are a separate core issue already fixed by deferring pxr (#5826) and cherry-picked to `release/3.0.0-beta2`.

## 5. Test plan
- [x] `./isaaclab.sh -p -m pytest source/isaaclab_rl/test/test_template_generator.py` → 15 passed
- [x] Import-safety + UI-module tests fail without the `__init__ext`/`extension.toml` fix (verified by reverting)
- [x] Generator raises on a missing agent template (verified)
- [x] `./isaaclab.sh -f` (pre-commit) clean


kellyguo11 (approved):

kellyguo11:
pls also cherry pick to the release branch

PR #5944, opened 2026-06-03
Validate typed presets resolve against their backend during Hydra resolution

ooctipus:
## Summary

alternative to #5939

- Selecting a typed preset (e.g. `presets=newton_mjwarp`) on a task that does **not** declare that name on a `PhysicsCfg`/`RendererCfg` previously created a silent mixed configuration: the name would match an unrelated scalar/sensor preset that happens to share the name, tuning those fields while leaving the real backend (e.g. PhysX) untouched. This adds a post-resolution check that a selected *reserved* typed name actually landed on a config of its own type at least once, and raises a clear `ValueError` otherwise.
- The set of reserved canonical names is declared per typed target in `PresetTarget` (`preset_target.py`) — `PHYSICS = {newton_mjwarp, newton_kamino}`, plus the known renderer backends. New solvers/renderers opt in by adding their name there; the resolver hardcodes nothing. `physx`/`ovphysx` are intentionally **not** reserved (PhysX is the default backend, so selecting it on a task without a physics preset is harmless rather than a silent mismatch).
- Implementation is minimal: one `typed_hits: set[str]` threaded through `_pick_alternative`/`_resolve_active_presets` records when a reserved name resolves to a value of its target type; `_validate_typed_presets` then flags `reserved-selected ∩ consumed − typed_hits`. Truly-unknown names continue to flow to the existing "Unknown preset(s)" error path.

## Behavior

| Invocation | Result |
|---|---|
| `Isaac-Velocity-Flat-Anymal-C-v0 presets=newton_mjwarp` | resolves (lands on `NewtonCfg`) |
| `Isaac-Navigation-Flat-Anymal-C-v0 presets=newton_mjwarp` | `ValueError` (no Newton physics; only matched a sensor/scalar) |
| `Isaac-Navigation-Flat-Anymal-C-v0 presets=physx` | resolves (PhysX not reserved, benign) |
| `Isaac-Navigation-Flat-Anymal-C-v0 presets=newton_kamino` | "Unknown preset(s)" (unchanged path) |
| legacy alias `presets=newton` on navigation | `ValueError` (canonicalized to `newton_mjwarp`) |

## Test plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab_tasks/test/core/test_hydra.py` (added 7 regression tests for the validator + end-to-end resolution; full file passes)
- [x] `test_preset_cli.py`, `test_newton_solver_preset_names.py`, `test_preset_kit_decision.py` pass
- [x] Config-resolution probes for the table above
- [x] `pre-commit` (ruff/format/codespell/rst/license) green
- [ ] CI

A changelog fragment is included under `source/isaaclab_tasks/changelog.d/`.

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
@@ -15,9 +15,11 @@ presets and their paths automatically, including inside dict-valued fields.
 
 Override categories (applied in order):
     1. Global presets: ``presets=inference,newton_mjwarp`` -- apply everywhere matching
-    2. Path presets: ``env.backend=newton_mjwarp`` -- REPLACE specific section
-    3. Preset-path scalars: ``env.backend.dt=0.001`` -- handled by us
-    4. Global scalars: ``env.decimation=10`` -- handled by Hydra
+    2. Typed selectors: ``physics=newton_mjwarp`` / ``renderer=NAME`` -- like a
+       global preset, but must resolve against a config of that type or it errors
+    3. Path presets: ``env.backend=newton_mjwarp`` -- REPLACE specific section
+    4. Preset-path scalars: ``env.backend.dt=0.001`` -- handled by us
+    5. Global scalars: ``env.decimation=10`` -- handled by Hydra
 
 Example usage::
 
@@ -279,7 +281,7 @@ def _pick_alternative(
     path: str = "",
     explicit_name: str | None = None,
     consumed_selected: set[str] | None = None,
-    typed_hits: set[str] | None = None,
+    typed_hits: dict[str, set[PresetTarget]] | None = None,
 ):
     """Choose the best alternative from a PresetCfg.
 
@@ -316,10 +318,11 @@ def _pick_alternative(
             consumed_selected.add(raw_name)
             consumed_selected.add(name)
         if typed_hits is not None:
-            # mark a reserved typed name (e.g. newton_mjwarp) that landed on its own backend type
-            target = PresetTarget.reserved_preset_targets().get(name)
-            if target is not None and isinstance(val, target.base_classes):
-                typed_hits.add(name)
+            # record which typed targets (physics/renderer) this name landed on
+            targets = {t for t in PresetTarget if t.base_classes and isinstance(val, t.base_classes)}
+            if targets:
+                typed_hits.setdefault(raw_name, set()).update(targets)
+                typed_hits.setdefault(name, set()).update(targets)
         if match_name is not None:
             if match_value is not val and match_value != val:
                 raise ValueError(
@@ -344,7 +347,7 @@ def _resolve_active_presets(
     *,
     strict_explicit: bool = True,
     consumed_selected: set[str] | None = None,
-    typed_hits: set[str] | None = None,
+    typed_hits: dict[str, set[PresetTarget]] | None = None,
     consumed_explicit: set[str] | None = None,
 ):
     """Resolve presets by walking only the currently active tree.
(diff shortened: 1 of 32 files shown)


source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -52,10 +58,16 @@ class PresetTarget(enum.Enum):
...
-    PHYSICS = ("physics", (PhysicsCfg,), {"newton": "newton_mjwarp", "kamino": "newton_kamino"})
+    # Members. Tuple values are (label, base_classes, legacy_aliases,
+    # preset_names); the enum metaclass collects the whole namespace before
+    # constructing members, so ``__new__`` below unpacks each tuple regardless
+    # of declaration order.
+    PHYSICS = (
+        "physics",
+        (PhysicsCfg,),
+        {"newton": "newton_mjwarp", "kamino": "newton_kamino"},
+        {"newton_mjwarp", "newton_kamino"},
```
hujc7: I think this can reuse the values for legacy alias assuming it's 1 to 1. @AntoineRichard also suggested the legacy ones could be dropped as they might not be used yet?

hujc7: To some extent, I don't think we can rely on fixed naming. It might need to be discovery based on PhysicsCfg(PresetCfg). So consuming typed presets might be needed. Otherwise, we would need to hardcode all names here. Legacy names does not enforce name but provides a alias.

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -551,21 +592,32 @@ def register_task(task_name: str, agent_entry: str) -> tuple:
...
         if "=" not in arg:
             hydra_args.append(arg)
             continue
         key, val = arg.split("=", 1)
-        if key.lstrip("-") == "presets":
+        token = key.lstrip("-")
+        if token == PresetTarget.DOMAIN.value:
             global_presets.extend(v.strip() for v in val.split(",") if v.strip())
+        elif token in typed_labels:
+            for name in (v.strip() for v in val.split(",") if v.strip()):
```
hujc7, line 612: so it looks like multiple targets are accepted for typed presets, but it will fail later?

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
```diff
@@ -551,21 +592,32 @@ def register_task(task_name: str, agent_entry: str) -> tuple:
...
     hydra_args: list[str] = []
     for arg in sys.argv[1:]:
         if "=" not in arg:
             hydra_args.append(arg)
             continue
         key, val = arg.split("=", 1)
-        if key.lstrip("-") == "presets":
+        token = key.lstrip("-")
+        if token == PresetTarget.DOMAIN.value:
             global_presets.extend(v.strip() for v in val.split(",") if v.strip())
```
AntoineRichard, line 610: This still lets `presets=newton_mjwarp` / `presets=newton` act as untyped broadcasts, so they can hit only scalar or sensor presets and leave PhysX active. Could we carry typed intent for the legacy Newton names even when they arrive through `presets=`?

That keeps ordinary domain broadcasts (`presets=depth`, `presets=inference`, etc.) unchanged, but makes the deprecated aliases and their canonical replacements go through the existing `_validate_typed_presets()` check.

```suggestion
    typed_labels = {target.value: target for target in PresetTarget if target.base_classes}
    legacy_typed_targets = {
        name: target
        for target in PresetTarget
        if target.base_classes
        for alias, replacement in target.legacy_aliases.items()
        for name in (alias, replacement)
    }
    global_presets: list[str] = []
    requested_targets: dict[PresetTarget, set[str]] = {}
    override_items: list[tuple[str, str, str]] = []
    hydra_args: list[str] = []
    for arg in sys.argv[1:]:
        if "=" not in arg:
            hydra_args.append(arg)
            continue
        key, val = arg.split("=", 1)
        token = key.lstrip("-")
        if token == PresetTarget.DOMAIN.value:
            for name in (v.strip() for v in val.split(",") if v.strip()):
                global_presets.append(name)
                target = legacy_typed_targets.get(name)
                if target is not None:
                    requested_targets.setdefault(target, set()).add(name)
```

I'd also add regression coverage for `presets=newton_mjwarp` and `presets=newton` on a config where the name only appears on an unrelated scalar preset; both should raise the same way `physics=newton_mjwarp` does.

AntoineRichard (changes requested):
Just tried `./isaaclab.sh -p scripts/reinforcement_learning/skrl/train.py --task Isaac-Navigation-Flat-Anymal-C-v0 presets=newton_mjwarp` on this branch and it does not raise as expected. `./isaaclab.sh -p scripts/reinforcement_learning/skrl/train.py --task Isaac-Navigation-Flat-Anymal-C-v0 --physics=newton_mjwarp` raised as expected. `./isaaclab.sh -p scripts/reinforcement_learning/skrl/train.py --task Isaac-Navigation-Flat-Anymal-C-v0 --physics=newton` raised.  `./isaaclab.sh -p scripts/reinforcement_learning/skrl/train.py --task Isaac-Navigation-Flat-Anymal-C-v0 presets=newton`  did not raise. See suggested change

source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py
hujc7: I discussed with @ooctipus yesterday. I think the main problem to also check the `presets` path is that we need to know the exact names that corresponds to the type, e.g. newton_mjwarp -> physics, which will need to be hardcoded. It's the standard convention in the repo, but might not be something used by users, where they can define anything. 

The existing way tried to avoid that by going the discovery path: for the target class, the typed presets target must be resolved properly, then it can be broadcasted to the rest of the cfg tree.

ooctipus: you have two options there:
1.hard code the preset string then `presets=` behave the same way as `physics=`

2.zero hard code string, accept `presets=` as the dangerous path, and recommend user to use `physics=` for typing protected path.

current implementation does 2. which I think makes sense. 1. could also make sense then you would need to maintain a set of strings, and those set of strings won't protect the user's custom physics engine either.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
ooctipus: I agree we should not hard code any naming

AntoineRichard (approved):

PR #5956, opened 2026-06-04
Deprecate MJWarp parallel linesearch

AntoineRichard:
## Summary
- Kept `MJWarpSolverCfg.ls_parallel` as a deprecated compatibility option instead of removing the public field.
- Setting `ls_parallel=True` now emits a warning, resets to iterative line search, and is not forwarded to `SolverMuJoCo`.
- Removed task/test call sites that selected parallel line search.
- Updated user-facing solver docs to describe `ls_iterations` as an iterative line-search cap that stops early on convergence, without documenting the deprecated option.

Review requested: @adenzler-nvidia

## Test Plan
- Checked that no source/docs call sites pass `ls_parallel=`.
- Checked that the physical-backend and migration docs no longer mention `ls_parallel`, parallel line search, or the old “keep low for performance” phrasing.
- Ran a temporary source-level compatibility check confirming `MJWarpSolverCfg.ls_parallel` is accepted, warns via `DeprecationWarning`, and is excluded from MuJoCo solver kwargs.
- `./isaaclab.sh -p -m compileall -q source/isaaclab_newton/isaaclab_newton/physics/mjwarp_manager_cfg.py source/isaaclab_newton/isaaclab_newton/physics/mjwarp_manager.py`
- `SKIP=check-git-lfs-pointers ./isaaclab.sh -f`

Full `./isaaclab.sh -f` was also run, but the local environment does not have `git-lfs` installed, so only `check-git-lfs-pointers` failed with `git-lfs is not installed`; the remaining hooks passed.


AntoineRichard:
Tagging @adenzler-nvidia for review on the MuJoCo Warp linesearch removal.

docs/source/migration/migrating_to_isaaclab_3-0.rst
```diff
@@ -672,7 +672,8 @@ Key Newton solver parameters:
    * - ``nconmax``
      - Max contacts per env
    * - ``ls_iterations``
-     - Linear solver iterations (higher = more stable, slower)
+     - Iterative line search cap; keep low for performance. Parallel line
+       search is not configurable.
```
adenzler-nvidia: Let's just drop any mention of parallel here. Also, the low for performance is unclear, because it's terminating early on convergence anyway, and usually you pay in more outer solver iterations if your linesearch does not converge.

adenzler-nvidia (commented):
Thanks, I'd just drop any mention of parallel linesearch. Or if you don't want the breaking change for users, make ls_parallel a noop for a while and always use iterative linesearch behind the scenes.

adenzler-nvidia (approved):
Looks good from my side

docs/source/migration/migrating_to_isaaclab_3-0.rst
AntoineRichard: Updated in 0612907. The migration table now drops the deprecated option text and describes `ls_iterations` as a cap that stops early on convergence; the solver docs use the same wording.

AntoineRichard:
> Looks good from my side

Thanks for the review! It's odd we didn't add you to the reviewer list.

source/isaaclab_newton/isaaclab_newton/physics/mjwarp_manager_cfg.py
```diff
@@ -89,7 +90,12 @@ class MJWarpSolverCfg(NewtonSolverCfg):
     """
 
     ls_parallel: bool = False
```
ooctipus, line 92: can we just remove it?  this is introduced in lab 3.0 and its not even GA yet

ooctipus (approved):

source/isaaclab_newton/isaaclab_newton/physics/mjwarp_manager_cfg.py
AntoineRichard: The rational behind this is not screw the early adopters!

PR #5504, opened 2026-05-05
[WARP] Cleanup Part 1: Add frontend bridge for stable manager-based tasks

hujc7:
## Review Map

- A stacked PR’s page shows the cumulative diff of its dependency chain; the "Exact changes" link pins that PR’s own commit range.
- SHA links can go stale after a branch update; the copy on #5504 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| 📌 #5504 Part 1: frontend bridge (stable ids on the warp runtime) |  | — | — |
| #6607 Part 2: mask-first frontend (draft) |  | #5504 | [link](https://github.com/isaac-sim/IsaacLab/pull/6607/changes/00ae695ac63e67ca678b91a837d3737529d2ec91..f68d03f40854b31a36063c5068f2933d32ea5db9) |
| #6646 Part 3: scene reset mask-native + capture-ready (draft, base rewritten — pending rebase onto #6607) |  | #6607 | [link](https://github.com/isaac-sim/IsaacLab/pull/6646/changes/4360622bde446c52c3fc85924639f9d1119d6e95..6d49e32f11e15894a4a0e49161a356ec79679ac6) |
| #6580 Part 4: OVPhysX backend (draft) |  | #5504 | [link](https://github.com/isaac-sim/IsaacLab/pull/6580/changes/fc37fa65ad95d1ac6e3f32e721329ef715899d37..716cef20979ca6c578a6e595812291c7b1a282fc) |
| #6611 Part 5: persistent-launch speedups (draft, absorbs #6551) |  | #6646 | [link](https://github.com/isaac-sim/IsaacLab/pull/6611/changes/6d49e32f11e15894a4a0e49161a356ec79679ac6..6e3da794215da03745a9bc8d35790e22262b099a) |

## 1. Summary

- `--frontend {torch,warp}` on the shared RL CLI (train **and play**) runs **stable** task ids on the warp runtime — no duplicated configs, no parallel warp task registrations at all (end state pinned by test). The `rlinf` integration constructs environments inside the external framework and is not frontend-routable (known limitation).
- **1.24–1.70× faster end-to-end** than torch on the same tasks, identical Newton physics (§3.1).
- Manager-based: `WarpFrontend` adapts the stable cfg in place; MDP twins resolve by name on the mirrored experimental package tree — deterministic, registration-free, missing twins reported in one aggregated failure.
- Direct: the stable registration declares its warp class via `warp_entry_point`; only the env class swaps, the cfg is shared.
[...]

PR #5974, opened 2026-06-04
[Newton] Declare physics presets for warp-only experimental envs

hujc7:
## 1. Summary

- 18 warp env configs in `isaaclab_tasks_experimental` assigned a bare `NewtonCfg` straight to `sim.physics`, bypassing the `PresetCfg` mechanism — so they declared no `default` and did not participate in `presets=` resolution like the migrated `core` envs.
- Each now wraps physics in a module-level `PhysicsCfg(PresetCfg)` with a `newton_mjwarp` variant and `default` aliasing it (warp-only, so no PhysX default is invented), matching the `core/velocity` convention.
- Direct envs additionally had stray `solver_cfg` / `newton_cfg` class attributes (confirmed unreferenced) lifted into the preset class.
- No behavior change: with no preset and with `presets=newton_mjwarp`, resolution returns the same `NewtonCfg` these envs used before.

## 2. Scope

- Locomotion velocity flat (9): `a1, anymal_b, anymal_c, anymal_d, cassie, g1, go1, go2, h1`
- Manipulation reach (2): `franka, ur_10` (`SimulationCfg` `dt` preserved)
- Classic manager-based (3): `ant, cartpole, humanoid` (humanoid builds physics in `__post_init__`)
- Direct warp (4): `allegro_hand_warp_env_cfg`, and the split cfg files `ant/ant_env_warp_cfg`, `cartpole/cartpole_warp_env_cfg`, `humanoid/humanoid_warp_env_cfg` (allegro's `physics_material` preserved)

## 3. Pattern

```python
@configclass
class PhysicsCfg(PresetCfg):
    newton_mjwarp = NewtonCfg(
        solver_cfg=MJWarpSolverCfg(...),
        num_substeps=1,
        debug_mode=False,
    )
    default = newton_mjwarp
```

## 4. Test plan

- [x] Rebased onto latest `develop` (picks up the direct-env cfg split from #5916 and the `ls_parallel` deprecation from #5956)
- [x] `./isaaclab.sh -f` — clean pass
- [x] `py_compile` on all 18 files
- [x] Preset resolution exercised on every distinct `PhysicsCfg`: both `default` (no preset) and `presets=newton_mjwarp` resolve to a `NewtonCfg`

### DIFF
--- a/source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/allegro_hand/allegro_hand_warp_env_cfg.py
+++ b/source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/allegro_hand/allegro_hand_warp_env_cfg.py
@@ -16,7 +16,30 @@ from isaaclab.sim.spawners.materials.physics_materials_cfg import RigidBodyMater
 from isaaclab.utils.assets import ISAAC_NUCLEUS_DIR
 from isaaclab.utils.configclass import configclass
 
+from isaaclab_tasks.utils import PresetCfg
+
 from isaaclab_assets.robots.allegro import ALLEGRO_HAND_CFG
+
+
+@configclass
+class PhysicsCfg(PresetCfg):
+    newton_mjwarp = NewtonCfg(
+        solver_cfg=MJWarpSolverCfg(
+            solver="newton",
+            integrator="implicitfast",
+            njmax=80,
+            nconmax=70,
+            impratio=10.0,
+            cone="elliptic",
+            update_data_interval=2,
+            iterations=100,
+            ls_iterations=15,
+            # save_to_mjcf="AllegroHand.xml",
+        ),
+        num_substeps=2,
+        debug_mode=False,
+    )
+    default = newton_mjwarp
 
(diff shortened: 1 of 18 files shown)


kellyguo11:
are presets necessary if they only support one backend?

hujc7:
> are presets necessary if they only support one backend?

So there are two ways I think

1. I have https://github.com/isaac-sim/IsaacLab/pull/5504 which bridges warp env to torch env, which reuses the torch cfg.
2. There's a QA bug where they are trying to run warp envs following the same preset pattern as torch envs and it complains that newton_mjwarp does not exist, which was then reported as a bug. To some extent, aligning the usage could be beneficial. @AntoineRichard mentioned kamino is coming so this will need preset later.

hujc7:
Superseded by https://github.com/isaac-sim/IsaacLab/pull/5504: it removes all 18 duplicated warp env configurations this PR patches. The warp frontend now runs stable task ids with the stable configurations, whose `PresetCfg` physics blocks already provide `presets=` resolution — the bypass this PR fixed no longer exists.

PR #5983, opened 2026-06-05
Fix Newton deferred graph warmup

### LINKED ISSUE #5961: [Bug Report] Kit Visualizer + Newton MJWarp Drops First-Step External Wrench
### Describe the bug

When running `scripts/demos/quadcopter.py` with the Kit visualizer and Newton MJWarp backend, the quadcopter falls instead of hovering. The same script behaves as expected in the other visualizer/backend combinations.

The thrust force is set to match gravity, so the quadcopter should remain approximately stationary. Runtime debugging shows that on the first Kit+Newton step, the vertical velocity becomes -0.04905[m/s], matching one step of gravity-only integration. This suggests the first-step external wrench is not consumed by the actual physics step.

This appears related to Newton’s deferred relaxed CUDA graph capture path when RTX/Kit is active. The log also says "Newton CUDA graph capture deferred until first step() (RTX active)".

### Steps to reproduce

- Run: `./isaaclab.sh -p scripts/demos/quadcopter.py --physics newton_mjwarp`
- Expected: The quadcopter should hover in place.
- Observed: With Kit visualizer + Newton MJWarp, the quadcopter falls.

Image

### System Info

  - Commit: 271f9304818
  - Isaac Sim Version: 6.0.0-alpha.212
  - OS: Ubuntu 22.04.5 LTS
  - GPU: NVIDIA RTX A5000
  - CUDA: 13.0
  - GPU Driver: 580.159.03

### Additional context

Likely root cause: in Kit+Newton, CUDA graph capture is deferred until the first `sim.step()` because RTX is active. The relaxed capture path performs a warm-up simulation before graph capture. That warm-up appears to consume and clear the `state_0.body_f` external wrench written by `robot.write_data_to_sim()`, so the first visible graph launch advances under gravity without the quadcopter thrust.

### Checklist

### Acceptance Criteria

The quadcopter should remain stationary in the air with Kit visualization and Newton MJWarp backend, as in other visualizer+backend combinations.

AntoineRichard:
# Description

Fixes #5961.

When Kit/RTX is active, CUDA graph capture is deferred until the first `NewtonManager.step()` because initialization may overlap RTX background-stream work. Before capture, `_capture_relaxed_graph()` performs an eager warmup simulation to allocate Newton solver buffers.

The warmup is not a user-visible IsaacLab step, but it was mutating the live state: it can advance positions and velocities, clear a staged `state.body_f` wrench, and swap the manager’s double-buffered states. The first graph replay could therefore begin from the warmup state rather than the caller-prepared state.

This PR snapshots both state buffers before warmup and restores their references and contents in a `finally` block before capture begins. Restoring both buffers is required because a solver may ping-pong them during warmup; synchronizing after restoration ensures capture sees the restored data.

The regression test forces deferred relaxed capture, stages a wrench that cancels gravity, and checks that the first visible replay has zero vertical velocity. The test uses native Warp arrays and adds no dependencies.

Dependencies: none.

## Type of change

## Screenshots

N/A — backend-only physics change.

## Checklist

- [x] I have made corresponding changes to the documentation (not applicable; no user-facing documentation changed)

## Test Plan

- `./isaaclab.sh -f`
- `git diff --check`
- The focused CUDA regression needs an IsaacLab Python environment with Warp and a CUDA runtime. The local wrapper resolved to an environment without Warp, so CI/GPU validation must run it.


source/isaaclab_newton/changelog.d/yizew-deferred-cuda-wrench.rst
```diff
@@ -0,0 +1,4 @@
+Fixed
+^^^^^
+
+* Fixed deferred Newton CUDA graph capture with Kit/RTX so first-step external wrenches are preserved.
```
pbarejko: What I really find ambiguous. The changes are related to newton manager, however the comment explains CUDA graph capture with Kit/RTX. 

Also, I have been wondering what does this really fix? You create snapshots, you stash them, and after sim you recover them.

Why do we need to call wp.synchronize_stream twice?

AntoineRichard: Yeah, we were seeing an issue with defered capture.

The warmup steps the simulation without calling lab internals so on the very first step we can get incorrect behaviors. So we snapshot the state before warm-up and restore them after the warmup that's it.

We may not need the wp.sync twice.

AntoineRichard:
@YizeWang is this still needed?

AntoineRichard:
Confirmed with @YizeWang that this is needed, @pbarejko @ooctipus for review.

pbarejko (approved):

source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py
```diff
@@ -784,3 +785,50 @@ def test_reset_lands_in_state_0_after_odd_kamino_steps_without_cuda_graph(num_st
...
+        body_f = wp.from_numpy(
+            np.array([[0.0, 0.0, 9.81, 0.0, 0.0, 0.0]], dtype=np.float32), dtype=wp.spatial_vector, device="cuda:0"
+        )
+        NewtonManager._state_0.body_f.assign(body_f)
+
+        sim.step(render=False)
+        wp.synchronize_device("cuda:0")
+
+        joint_qd = NewtonManager._state_0.joint_qd.numpy()
+        assert np.isclose(joint_qd[2], 0.0, atol=1e-3, rtol=0.0)
```
pbarejko, line 1543: You observe here effect of gravity, why not additionally observe initial conditions from `joint_q`? They should be all initial values?

source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py
```diff
@@ -1508,6 +1508,53 @@ def test_reset_lands_in_state_0_after_odd_kamino_steps_without_cuda_graph(num_st
...
+        sim.reset()
+
+        NewtonManager._graph = None
+        NewtonManager._graph_capture_pending = True
+        body_f = wp.from_numpy(
+            np.array([[0.0, 0.0, 9.81, 0.0, 0.0, 0.0]], dtype=np.float32), dtype=wp.spatial_vector, device="cuda:0"
+        )
+        NewtonManager._state_0.body_f.assign(body_f)
+
+        sim.step(render=False)
```
hujc7: **[P2] Agent found bug: this test never reaches `_capture_relaxed_graph`, so it passes with or without the fix.**

`build_simulation_context` leaves `NewtonManager._usdrt_stage` as `None`, so `step()` takes the deferred **standard** branch (`newton_manager.py:970-975`, plain `wp.ScopedCapture`). That branch has no warmup and therefore cannot exhibit the bug — the patched relaxed path is only reached when `_usdrt_stage is not None`.

Measured on an RTX A6000, running this test body verbatim:

| tree | `joint_qd[2]` after the step | |
|---|---|---|
| `develop`, no fix | `0.0` | test passes |
| `develop` + this PR | `0.0` | test passes |

The assertion itself is fine — dropping the staged wrench gives `-0.04905`, so it does detect a lost wrench. The problem is only that the capture path under test is the wrong one. `visualizers=["kit"]` does not help either: `build_simulation_context` writes `/isaaclab/visualizer/types` after `SimulationContext.__init__` has already computed `_clone_physics_only`, and `resolve_visualizer_types()` ignores that setting without `/isaaclab/visualizer/explicit`.

Driving the relaxed path directly does reproduce it:

```python
NewtonManager._graph = None
NewtonManager._state_0.body_f.assign(body_f)

graph = NewtonManager._capture_relaxed_graph("cuda:0")
assert graph is not None
wp.capture_launch(graph)
wp.synchronize_device("cuda:0")

joint_qd = NewtonManager._state_0.joint_qd.numpy()
assert np.isclose(joint_qd[2], 0.0, atol=1e-3, rtol=0.0)
```

With that shape I measure `body_f == [0, 0, 0]` and `joint_qd[2] == -0.04905` on `develop`, and `[0, 0, 9.81]` / `0.0` with this PR applied — fails without the fix, passes with it.

This also covers @pbarejko's question below about asserting on `joint_q`: once the test drives the real path, checking that `joint_q` is still at its initial values is a stronger statement than the velocity alone.

- [ ] The regression test exercises `_capture_relaxed_graph` and fails on `develop`.

source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py
```diff
@@ -1508,6 +1508,53 @@ def test_reset_lands_in_state_0_after_odd_kamino_steps_without_cuda_graph(num_st
...
+            solver_cfg=MJWarpSolverCfg(use_mujoco_contacts=True),
+            num_substeps=1,
+            use_cuda_graph=True,
+        ),
+    )
+
+    with build_simulation_context(sim_cfg=sim_cfg) as sim:
+        builder = NewtonManager.create_builder()
+        body = builder.add_body(mass=1.0)
+        builder.add_joint_free(child=body)
```
hujc7: **[P2] Agent found bug: `add_body()` already creates a FREE joint, so this line adds a second one in parallel.**

Measured: `joint_qd` comes out 12 wide instead of 6, and Newton/MuJoCo warns twice —

```
UserWarning: Adding a FREE joint between parent -1 and child 0 (label: 'body_0'), but another
joint already connects these bodies. A FREE joint parallel to another joint is inconsistent.

UserWarning: Loop joint 1 (type FREE, 3 linear + 3 angular DOFs) has no supported MuJoCo
equality constraint mapping. Skipping loop closure for this joint.
```

`builder.joint_q[-7:] = ...` on the next line then writes the *second* joint's coordinates rather than the body's. Deleting this line gives a clean single free body: `joint_qd` 6 wide, no warnings, and the gravity check still reads `joint_qd[2]`.

source/isaaclab_tasks/changelog.d/pr-5983-rendering.skip
```diff
@@ -0,0 +1,2 @@
+Internal: updated Franka cloth golden images to match the corrected Newton deferred graph warmup
```
hujc7: **[P2] Agent found bug: this fragment describes a change that is not in the PR.**

Head has no golden-image diff against `develop` — `git diff develop..HEAD -- source/isaaclab_tasks/test/golden_images/` is empty, and `isaaclab_tasks` has no source change here either.

`7ea8934f3` did refresh 12 `franka_cloth` baselines, with the rationale "The corrected deferred warmup rollback removes a hidden Newton step and intentionally shifts the cloth trajectory". The merge at `1d7c0f2d3` then resolved those files to `develop`'s side, so the refresh is gone but this fragment stayed.

One of the two is wrong, and I can't tell which without an RTX run:

- the fragment — the trajectory does not actually shift, so drop it (and with it the only reason for an `isaaclab_tasks` fragment at all); or
- the baselines — it does shift, so they need re-refreshing from the CI runner at the current head.

Worth noting `7ea8934f3` also refreshed the `physx-*` baselines, which a Newton-side fix should not move; that suggests the original refresh picked up unrelated runner drift and is worth redoing rather than restoring.

- [ ] The Franka-cloth rendering test passes at head, and this fragment matches what the PR actually contains.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -2332,13 +2358,28 @@ def _capture_relaxed_graph(cls, device: str, capture_target: Callable[[], None]
 
         # Warmup: pre-allocate all solver scratch buffers so the capture window has
         # no new cudaMalloc calls (which are forbidden inside graph capture).
+        # Warmup also advances Newton outside the normal IsaacLab step bookkeeping,
+        # so snapshot/restore state to keep its writes out of the first real replay.
+        # LIMITATION: Only Newton State buffers are restored below. If this warmup
```
hujc7: **[P2] Agent review: the rollback also misses arrays that are *created* on the `State` during warmup — worth adding to this list.**

`_snapshot_state_arrays` records a fixed set of `(owner, name)` pairs *before* warmup. That lets `_restore_state_arrays` follow an array that gets **rebound** during warmup — XPBD does `state_out.particle_q = new_particle_q` (`newton/_src/solvers/xpbd/solver_xpbd.py:268-269`) and resolving through `getattr(owner, name)` handles it correctly. But it has no way to reach an attribute that gets **added**, because the name was not in the snapshot.

Kamino does exactly that: `StateKamino.from_newton()` sets `state.body_f_total`, `state.joint_q_prev` and `state.joint_lambdas` on the Newton `State` when they are absent (`newton/_src/solvers/kamino/_src/core/state.py:262-288`) — which is what the comments at `newton_manager.py:980` and `:2259` describe. After warmup those three hold warmup-derived values and sit outside the rollback, and `joint_q_prev` is previous-step state.

Pre-existing rather than introduced here, and I have not measured whether it changes Kamino's result. Is it worth naming in this LIMITATION comment, or taking a second snapshot pass after warmup to catch newly added names?

kellyguo11:
run-ci

kellyguo11:
this probably needs a rebase, but I don't seem to have access to do that

PR #5984, opened 2026-06-05
Fix kamino reset for closed-loop systems.

rubengrandia:
# Description

Newton's generic `eval_fk` is an *articulated* (tree) forward kinematics and is
incorrect for the Kamino maximal-coordinate solver, especially for closed-loop
mechanisms. However, the base `NewtonManager.forward()` / `step()` called
`eval_fk` directly. The Kamino manager previously worked around this by
overriding the entire `step()` method, but `forward()` — reached from
`ArticulationData._ensure_fk_fresh()` via `SimulationManager.forward()` after a
manual joint/root write — still ran the wrong articulated FK for Kamino.

This PR refactors `NewtonManager` to allow per-solver specialization of forward
kinematics, removes the Kamino-specific `step()` override, and fixes a
double-buffer state-swap bug that caused environment resets to be written into
the wrong state buffer.

**What changed:**

- **FK specialization hook.** Added a solver-specialized `_eval_fk(world_mask, fk_mask)`
  classmethod plus a `_reconcile_fk` hook (bound in `initialize_solver`) and two
  flags, `_reconcile_fk_before_step` and `_forward_full_reconcile`. Both
  `forward()` and `step()` now dispatch through this hook, so the correct FK runs
  regardless of whether the call enters through the base class or a subclass. The
  initial body-state reconcile is deferred to the tail of `initialize_solver()`.
- **Removed the Kamino-specific `step()` override** (and the duplicated
  `_capture_or_defer_cuda_graph`). `NewtonKaminoManager` now only specializes
  `_eval_fk` / `_forward_kamino` and sets its flags in `_build_solver`.
- **Correct Kamino FK reset.** `_forward_kamino` extracts the actuated joint
  state and gathers the per-world base (root) pose/twist from the base joint's
  coordinates in `joint_q`/`joint_qd`, feeding `actuator_q`/`actuator_u`/`base_q`/`base_u`
  into `SolverKamino.reset`. After the reset it recomputes the full joint state
  from the reconciled body poses so reads of `joint_pos`/`joint_vel` reflect
  passive/loop-closure joints. `KaminoSolverCfg.use_fk_solver=False` falls back to
  articulated `eval_fk` (the faster path for tree systems).
- **Bug fix — `use_cuda_graph` gating the double-buffer swap.** In
  `_run_solver_substeps`, `need_copy_on_last` was gated on `cfg.use_cuda_graph`.
  With double-buffered state (Kamino uses `_use_single_state=False`) and an odd
  substep count, this left the canonical state in `_state_1` while reset writes
  (which target the `_state_0`-bound data layer) landed in the stale buffer when
  CUDA graphs were disabled. The gate is removed so the copy-on-last always runs
  for odd substep counts.

## Type of change

- Bug fix (non-breaking change which fixes an issue) — includes an internal
  refactor of `NewtonManager`'s FK dispatch

## Screenshots

N/A — no visual/UI change.

## Checklist

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/kamino_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/kamino_manager.py
@@ -11,15 +11,74 @@ import logging
 
 import warp as wp
 from newton import Model, eval_fk
+from newton._src.solvers.kamino._src.core.joints import JointDoFType
+from newton._src.solvers.kamino._src.core.types import vec6f
+from newton._src.solvers.kamino._src.kinematics.joints import (
+    extract_actuators_state_from_joints,
+)
+from newton._src.solvers.kamino._src.kinematics.resets import _reset_joints_of_select_worlds
 from newton.solvers import SolverKamino
 
 from isaaclab.physics import PhysicsManager
-from isaaclab.utils.timer import Timer
 
 from .kamino_manager_cfg import KaminoSolverCfg
 from .newton_manager import NewtonManager
 
 logger = logging.getLogger(__name__)
+
+# Kamino joint DoF type for a 6-DoF free joint (a floating base's root joint).
+_FREE_DOF_TYPE = wp.constant(int(JointDoFType.FREE))
+
+
+@wp.kernel(enable_backward=False)
+def _gather_base_state_from_joints(
+    base_joint_index: wp.array(dtype=wp.int32),
+    joint_dof_type: wp.array(dtype=wp.int32),
+    joint_coords_offset: wp.array(dtype=wp.int32),
+    joint_dofs_offset: wp.array(dtype=wp.int32),
+    joint_q: wp.array(dtype=wp.float32),
+    joint_qd: wp.array(dtype=wp.float32),
+    # outputs
+    base_q: wp.array(dtype=wp.transformf),
+    base_u: wp.array(dtype=vec6f),
+):
+    """Gather each world's base pose/twist from the base joint coordinates in the joint state.
+
+    For each world ``wid``, when its base joint (``base_joint_index[wid]``) is a free joint,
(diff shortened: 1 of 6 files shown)


source/isaaclab_newton/isaaclab_newton/physics/kamino_manager.py
```diff
@@ -30,87 +89,149 @@ class NewtonKaminoManager(NewtonManager):
...
+            outputs=[
+                scratch.p_j,
+                scratch.r_j,
+                scratch.dr_j,
+                cls._state_0.joint_q,
+                cls._state_0.joint_qd,
+                scratch.lambda_j,
+            ],
+            device=model_kamino.device,
+        )
```
AntoineRichard, line 235: Why not doing it now? IsaacLab can be binded to latest newton. @vastsoun

source/isaaclab_newton/isaaclab_newton/physics/kamino_manager.py
```diff
@@ -30,87 +89,149 @@ class NewtonKaminoManager(NewtonManager):
...
-        cls._log_solver_debug()
+        # Run FK
+        cls._solver.reset(
+            state_out=cls._state_0,
+            actuator_q=solver_kamino._actuators_q,
+            actuator_u=solver_kamino._actuators_u,
+            base_q=cls._base_q,
+            base_u=cls._base_u,
+            world_mask=world_mask,
+        )
```
AntoineRichard, line 192: Should the `_gather_base_state_from_joints` also be offloaded to Kamino's internals?

AntoineRichard (commented):
LGTM, but I'm wondering if some of things that are currently in IsaacLab should not be put in Kamino now.

AntoineRichard:
@aserifi @rubengrandia should we close this one?

rubengrandia:
Closing, will open an new PR that builds on recent changes in the Kamino reset, which should make the changes here smaller.

PR #6000, opened 2026-06-05
Frlai/fix rnn leapp tracing

frlai:
# Description
Fixes leapp lstm policy export. Previously it hooked onto structures in rslrl=3.1.2. The update now makes it work for rslrl=5.0.1, same version as the one in the pyproject.toml. Added a regression test to test if rslrl rnns work new test generates a dummy policy with rnn and tests the export. Previous it was assumed that there was one policy with a pretrained checkpoint that had an rnn but turns out there isn't any that by default run rnns so this error slipped through.

## Type of change

## Checklist


scripts/reinforcement_learning/leapp/rsl_rl/export.py
```diff
@@ -174,17 +174,37 @@ def _load_runtime_dependencies() -> None:
...
+    memory = get_actor_memory_module(policy_nn)
+    if memory is not None:
+        memory.hidden_state = actor_hidden
+
+
 def ensure_actor_hidden_state_initialized(policy_nn, batch_size: int, device, dtype):
     """Initialize and return the actor hidden state when a recurrent policy has not created it yet."""
-    actor_state, _ = policy_nn.get_hidden_states()
+    torch_module = torch
+    assert torch_module is not None
```
StafaH: This pattern seems odd, why do we need to do it this way? If torch is required just do the common:

```
try:
   import torch
except:
   raise ...
```

But even that might not be needed, if torch is not installed the error is quire obvious to the user to install torch

scripts/reinforcement_learning/leapp/rsl_rl/export.py
```diff
@@ -298,7 +318,7 @@ def export_rsl_rl_agent(
         runner.load(resume_path)
 
         policy = runner.get_inference_policy(device=env.unwrapped.device)
-        policy_nn = getattr(policy, "__self__", None)
+        policy_nn = policy
```
StafaH: Can we remove this line? Seems like a no-op, just use policy everywhere below?

source/isaaclab_rl/test/export/test_rsl_rl_export_recurrent_state.py
```diff
@@ -0,0 +1,104 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: It might be better to just an add a RNN test to the previously existing test file, this would allow re-use and reduce code duplication, and it would fit in with the other set of tests

StafaH (approved):
Left some minor comments to fix, otherwise LGTM

PR #6003, opened 2026-06-06
Fix Newton mesh approximation import controls

ooctipus:
## Summary
- add the missing nested mesh-collision property field to collision configs so collision props can dispatch authored mesh approximation settings
- update Newton replication to honor per-collider `physics:approximation` tokens instead of approximating every imported mesh as a convex hull
- add a Newton schema regression test for nested mesh-collision config dispatch

## Tests
- `git diff --cached --check` before commit
- syntax parsed modified Python files with `ast.parse`
- Not run: `./isaaclab.sh -p -m pytest source/isaaclab_newton/test/sim/test_newton_schemas.py -k test_newton_mesh_collision_nested_cfg_written -q` cannot collect locally because Python is missing `lazy_loader`

### DIFF
--- a/source/isaaclab/changelog.d/fix-nested-mesh-collision-cfg.rst
+++ b/source/isaaclab/changelog.d/fix-nested-mesh-collision-cfg.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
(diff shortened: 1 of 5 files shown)


StafaH:
Fix landed in #6892

PR #6006, opened 2026-06-06
Fix termination manager reset comments

NeoZng:
# Description

Corrects stale terminology in `TerminationManager.reset()` while preserving the original documentation fix:

- Replaces the stale references to reward terms with termination terms.
- Clarifies that the logged per-term metric is a mean activation across environments, rather than an episodic sum or count.

## Type of change

- Documentation update

## Validation

- `./isaaclab.sh -f`
- `git diff --check`


AntoineRichard (approved):
Thanks @NeoZng for the contribution! I retargeted the branch to develop since this will land in a short while.

PR #6009, opened 2026-06-06
Update the central pyproject, remove subpackage specific dependencies

StafaH:
# Description

Consolidates all dependencies into the central pyproject. This should make it easier to maintain a ground truth.

## Checklist


AntoineRichard:
Yes let's do this :D

.github/workflows/license-check.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
       run: |
         uv sync --extra all
         # Isaac Sim isn't in the dev pyproject; sync first so it isn't pruned.
-        uv pip install 'isaacsim[all,extscache]==${{ vars.ISAACSIM_BASE_VERSION || '6.0.0' }}'
+        uv pip install 'isaacsim[all,extscache]==${{ vars.ISAACSIM_BASE_VERSION || '6.0.0.1' }}'
```
pbarejko: Can we have one common place where we define version of IsaacSim? I can see this going out of sync very quickly. 

In the past, in OV projects what worked well is VERSION file. That file in the root of the repo had definitions of all most important dependencies.

StafaH: Added a single source of truth for isaacsim version in the pyproject.toml

docs/_extensions/isaaclab_docs.py
```diff
@@ -160,10 +175,10 @@ def _quickstart_isaacsim(branch: str, platform: str) -> str:
    uv venv --python 3.12 --seed env_isaaclab
    source env_isaaclab/bin/activate
    uv pip install --upgrade pip
-   uv pip install "isaacsim[all,extscache]==6.0.0.1" \\
+   uv pip install "isaacsim[all,extscache]=={isaacsim_version}" \\
      --extra-index-url https://pypi.nvidia.com \\
      --index-strategy unsafe-best-match --prerelease=allow
-   uv pip install -U torch==2.10.0 torchvision==0.25.0 \\
+   uv pip install -U torch==2.11.0 torchvision==0.26.0 \\
```
hujc7: Is there a way to also centralized this where possible? Or it's fine as we might not update it often?

docs/source/overview/core-concepts/renderers.rst
```diff
@@ -133,11 +133,11 @@ Install via the Isaac Lab CLI using the ``ov[ovrtx]`` token:
    packages are already part of the core install). Use ``ov[ovrtx]`` (or ``ov[all]``)
    to pull in the ``ovrtx`` dependency.
 
-Or install manually with pip (note the ``[ovrtx]`` extra and the extra index URL):
+Or install the ``ovrtx`` runtime wheel directly with pip (note the extra index URL):
 
 .. code-block:: bash
 
-   pip install --extra-index-url https://pypi.nvidia.com -e "source/isaaclab_ov[ovrtx]"
+   pip install --extra-index-url https://pypi.nvidia.com "ovrtx>=0.3.0,<0.4.0"
```
hujc7: This seems to be another maintenance point, can we rely on one path?

.github/workflows/license-exceptions.json
```diff
@@ -507,5 +512,15 @@
...
     "comment": "MPL-2.0 / OSRB"
+  },
+  {
+    "package": "distlib",
+    "license": "Python Software Foundation License",
+    "comment": "PSFL / OSRB"
+  },
+  {
+    "package": "ovrtx",
+    "license": null,
```
pbarejko: We need to find out this information.

docs/source/setup/installation/isaaclab_pip_installation.rst
```diff
@@ -63,21 +63,21 @@ Installing dependencies
 
          .. code-block:: bash
 
-            pip install -U torch==2.10.0 torchvision==0.25.0 --index-url https://download.pytorch.org/whl/cu128
```
hujc7, line 75: apply the same as sim version?

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -491,6 +577,15 @@ def _install_isaacsim() -> None:
     "teleop": ("isaaclab_teleop",),
 }
 
+# Root pyproject optional-dependency groups that carry the third-party runtime
+# requirements for each optional submodule (the submodules themselves no longer
+# declare dependencies). Mirrors OPTIONAL_ISAACLAB_SUBMODULES: ``mimic`` pulls in
+# the teleop stack as well, matching the editable-install behavior it replaces.
+OPTIONAL_SUBMODULE_ROOT_EXTRAS: dict[str, tuple[str, ...]] = {
+    "mimic": ("mimic", "teleop"),
+    "teleop": ("teleop",),
```
hujc7: I don't full understand install.py yet, but could this also live in pyproject?

.github/workflows/license-exceptions.json
StafaH: Updated

docs/_extensions/isaaclab_docs.py
StafaH: Centralized

docs/source/overview/core-concepts/renderers.rst
StafaH: Centralized

docs/source/setup/installation/isaaclab_pip_installation.rst
StafaH: Centralized

source/isaaclab/pyproject.toml
```diff
@@ -16,49 +16,10 @@ authors = [{name = "Isaac Lab Project Developers"}]
...
-    # OpenUSD (kit-less mode)
-    "usd-core>=25.0.0 ; platform_machine in 'x86_64 AMD64'",
-    "usd-exchange>=2.2 ; platform_machine in 'x86_64 AMD64 aarch64 arm64'",
-    # avoid broken hf-xet pre-release cached on NVIDIA Artifactory
-    "hf-xet>=1.4.1,<2.0.0 ; platform_machine in 'x86_64 AMD64 aarch64 arm64'",
-]
+# Third-party requirements are centralized in the root pyproject.toml
+# (single source of truth for the uv workspace, the wheel builder, and the
+# ./isaaclab.sh -i install CLI).
+dependencies = []
```
diegoferigo-rai, line 22: From a packaging perspective, I don't think this approach is viable.

I understand the need for an `isaaclab-dev` catch-all metapackage, but that should not influence the packaging semantics of the individual packages. Each package should remain valid as a standalone distribution, with a complete and accurate declaration of its own dependencies.

If stricter version pinning is needed for development or repository-wide compatibility, that can be enforced at the top-level `isaaclab-dev` package. The individual packages, however, should retain dependency specifications that reflect their actual requirements and support independent installation and distribution.

source/isaaclab/pyproject.toml
```diff
@@ -70,36 +31,6 @@ play = "isaaclab.cli:play"
 train = "isaaclab.cli:train"
 train_multigpu = "isaaclab.cli:train_multigpu"
```
diegoferigo-rai: As I already commented in the previous PR, these generic executable names are problematic in larger downstream environments.

I would suggest to refactor these play/train commands as subcommands of `isaaclab`, such as:
- isaaclab [...]
- isaaclab play [...]
- isaaclab train [...]
- isaaclab train --multi-gpu [...]

diegoferigo-rai (dismissed):
Thanks @StafaH for the contribution. I'm quite concerned about the direction this PR is taking, as it makes downstream consumption significantly more difficult.

In my view, the packages under `src/` (which are the ones that could eventually be published independently to PyPI) should be self-contained. Each package should define all of its own metadata and dependencies in its `pyproject.toml`, so that it remains functional and installable outside the context of this repository.

I do like the idea of having a repository-wide lock file, but I would expect the dependency declarations at the top level to remain relatively slim and primarily aggregate the requirements of the packages in `src/`.

One potential benefit of the top-level `pyproject.toml` would be to centralize the stricter version constraints required for IsaacSim, while keeping the package-level dependency specifications in `src/` more relaxed and suitable for standalone use.

Happy to discuss this further.

pyproject.toml
```diff
@@ -8,63 +8,174 @@ name = "isaaclab-dev"
...
-    "isaaclab-rl[rsl-rl]",
+# Isaac Sim (PhysX physics backend). The pinned version mirrors
+# ``[tool.isaaclab.versions].isaacsim`` (the single source of truth read by docs
+# and CI). Co-resolves with the base install under uv via the conflicts table in
+# [tool.uv]; isaacsim narrows newton to its ``==`` pin.
+isaacsim = ["isaacsim[all,extscache]==6.0.0.1"]
+# Omniverse renderer / physics backends. Installable separately or together via
+# the ``ov[ovphysx]`` / ``ov[ovrtx]`` / ``ov[all]`` install selectors. Pins mirror
+# ``[tool.isaaclab.versions]``.
+ov = ["ovphysx==0.4.13"]
```
kellyguo11: should this be ovphysx instead?

docs/source/experimental-features/bleeding-edge.rst
```diff
@@ -28,11 +28,11 @@ only the dependencies needed for the features you use:
...
-   # Install the base contrib package
-   uv pip install -e "source/isaaclab_contrib"
-
-   # Install with optional extras (e.g., for RLinf VLA post-training)
-   uv pip install -e "source/isaaclab_contrib[rlinf]"
+   # The base contrib package is installed with the core Isaac Lab packages.
+   # Install its optional dependencies (e.g., for RLinf VLA post-training) with pip:
+   pip install "ray[default]>=2.47.0" "av>=12.3.0" "numpydantic>=1.7.0" \
+       "albumentations>=1.4.18" decord2 "dm_tree>=0.1.8" "diffusers>=0.35.0" \
+       "timm>=1.0.14" "peft>=0.17.0" pandas
```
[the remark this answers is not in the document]
kellyguo11: ^

docs/source/experimental-features/bleeding-edge.rst
```diff
@@ -28,11 +28,11 @@ only the dependencies needed for the features you use:
 
 .. code-block:: bash
 
-   # Install the base contrib package
-   uv pip install -e "source/isaaclab_contrib"
-
-   # Install with optional extras (e.g., for RLinf VLA post-training)
-   uv pip install -e "source/isaaclab_contrib[rlinf]"
+   # The base contrib package is installed with the core Isaac Lab packages.
+   # Install its optional dependencies (e.g., for RLinf VLA post-training) with pip:
```
kellyguo11: this feels a lot less user friendly. are we able to reference the `rlinf` option from pyproject directly somehow?

source/isaaclab/isaaclab/cli/__init__.py
```diff
@@ -105,7 +105,7 @@ def cli() -> None:
             "    contrib[rlinf]\n"
             "    ov[ovrtx|ovphysx|all]\n"
             "    rl[rsl-rl|skrl|sb3|rl-games]  (default: all)\n"
-            "    visualizer[kit|newton|rerun|viser]  (default: all)\n"
+            "    visualizer[kit|rerun|viser]  (default: all)\n"
```
kellyguo11, line 108: is newton still a valid option? it still seems to be listed in the selective install docs

source/isaaclab/pyproject.toml
kellyguo11: that's a valid point. I think part of what we wanted to achieve here is that we can have a central location that guards the version pinning for our dependencies so that we don't end up with multiple packages specifying the same dependency, but could easily misalign in versioning because someone forgot to update it. 

perhaps there's a better way to handle both of these needs. ideally we can have a central place that acts as the source of truth for all version pinning, and each package will hold its own list of dependencies and pull the versioning from the central location. maybe the per-package pyproject.toml files could be constructed from the central one?

source/isaaclab/pyproject.toml
kellyguo11: is the concern here around the global definition of train/play in an environment?

diegoferigo: Yes, the main concern is the global definition of these commands in an environment.

`isaaclab` might be installed in a virtualenv or a larger conda environment together with many other packages, including ML/RL frameworks such as PyTorch, Lightning, Ray, etc. In these cases, the environment's `bin/` directory contains entry points from all installed packages. Adding generic executables such as `play` or `train` makes the namespace ambiguous and can easily lead to conflicts with other projects.

This also creates problems for downstream packagers. For example, if someone wanted to create a `python3-isaaclab` package for Ubuntu/Debian, generic binaries like `train` or `play` would likely not be acceptable in `/usr/bin` and would need to be renamed or wrapped anyway.

This is admittedly an extreme example, but it highlights the underlying issue: these names do not uniquely identify IsaacLab functionality. Since you already have a natural namespace available through an `isaaclab` CLI, I think it would be cleaner and more future-proof to expose these as subcommands:

```
isaaclab play [...]
isaaclab train [...]
isaaclab train --multi-gpu [...]
```

Given the growing adoption of IsaacLab, having a single, well-defined command namespace would also make it easier for users and downstream maintainers to integrate it into larger environments.

source/isaaclab/pyproject.toml
diegoferigo: I think the individual packages should remain self-contained and define a sufficiently relaxed set of dependencies so that their package metadata is complete and accurate. Moving dependency definitions elsewhere would result in incomplete metadata. For example, if `isaaclab-newton` is published on PyPI, its metadata should explicitly declare `newton` as a dependency.

> maybe the per-package pyproject.toml files could be constructed from the central one?

Regarding the central location, I see that as something serving repository development rather than package distribution. It is useful for maintaining a consistent development environment, but it should not dictate the structure of the individual packages. The individual packages should also not depend on `isaaclab-dev`, since that package is not intended for downstream distribution.

Instead, `isaaclab-dev` could collect the individual packages and provide a resolved development environment (e.g., through a workspace/lock file). This would allow you to:

1. keep package metadata correct and independent
2. avoid unnecessary strict pins in individual packages
3. introduce strict pinning here, e.g. for isaacsim interoperability
4. provide developers with a known-good environment

My preference would be:

* Individual packages specify dependencies with `>=` constraints (and upper bounds only when strictly required, e.g. a new major release that already broke assumptions).
* `isaaclab-dev` is structured as a [workspace](https://docs.astral.sh/uv/concepts/projects/workspaces) that collects all packages and generates the lock file (workspaces seems the perfect use case here).
* If you want to ensure shared dependencies use exactly matching constraints across packages, a small validation python script in pre-commit / CI could parse the pyproject files and check for inconsistencies.

In general, I think it is better for individual packages to be under-constrained rather than over-constrained. Downstream users can always add stricter pins in their environment, but overly strict package requirements are difficult to override and can prevent otherwise valid installations.

source/isaaclab/pyproject.toml
kellyguo11: I agree, that's a good point, we should not pollute global namespace in any environment

docs/source/experimental-features/bleeding-edge.rst
StafaH: Fixed with a better option that uses deps form root pyproject

source/isaaclab/pyproject.toml
StafaH: Fixed

kellyguo11 (approved):

PR #5546, opened 2026-05-08
docs: fix swapped preserve_order descriptions in resolve_matching_names docstrings

dashitongzhi:
## Description

Fixes #5146

The docstrings for both `resolve_matching_names` and `resolve_matching_names_values` in `source/isaaclab/isaaclab/utils/string.py` had the descriptions for `preserve_order=True` and `preserve_order=False` swapped.

**Current (incorrect) docstring:**
- `preserve_order=True` → "ordering is dictated by the order of the target strings"
- `preserve_order=False` → "ordering is the same as the order of the provided list of query regular expressions"

**Actual code behavior (and examples):**
- `preserve_order=False` → target list order (default, no reordering)
- `preserve_order=True` → query keys order (reordering occurs inside `if preserve_order:`)

The fix swaps the prose descriptions to match the actual code behavior and the provided examples.

## Type of change

- [x] Documentation fix (non-breaking change which fixes a docstring)

## Checklist

- [x] I have checked that there is no similar PR
- [x] The code examples in the docstrings are correct and unchanged

PR #6016, opened 2026-06-07
docs: clarify preserve_order descriptions and fix comma typo in string utils

### LINKED ISSUE #4493: Docstring error in resolve_matching_names and resolve_matching_names_values: preserve_order description is inverted

Ottohere-Mourn:
## Summary

- Clarified the `preserve_order` parameter descriptions in `resolve_matching_names` and `resolve_matching_names_values`
- The original wording ("the ordering is the same as the order of") could be misread as describing the opposite behavior
- Changed to explicit language: False = iteration order of target strings, True = reordered to follow query key order
- Fixed a redundant phrase in the second function's docstring example text

Fixes #4493

## Test Plan

- [ ] Docstrings now clearly describe the actual function behavior shown by the existing examples
- [ ] No code logic changes — documentation only

### DIFF
--- a/source/isaaclab/isaaclab/utils/string.py
+++ b/source/isaaclab/isaaclab/utils/string.py
@@ -290,7 +290,7 @@ def resolve_matching_names_values(
     appear first, then those matching the second key, and so on).
 
     For example, consider the dictionary is {"a|d|e": 1, "b|c": 2}, the list of strings is ['a', 'b', 'c', 'd', 'e'].
-    If :attr:`preserve_order` is False, then the function will return the indices of the matched strings and the values as: ([0, 1, 2, 3, 4], ['a', 'b', 'c', 'd', 'e'], [1, 2, 2, 1, 1]). When
+    If :attr:`preserve_order` is False, then the function will return the matched indices, names, and values as: ([0, 1, 2, 3, 4], ['a', 'b', 'c', 'd', 'e'], [1, 2, 2, 1, 1]). When
     :attr:`preserve_order` is True, it will return them as:
     ([0, 3, 4, 1, 2], ['a', 'd', 'e', 'b', 'c'], [1, 1, 1, 2, 2]).
 
--- a/source/isaaclab/isaaclab/utils/string.py
+++ b/source/isaaclab/isaaclab/utils/string.py
@@ -183,12 +183,13 @@ def resolve_matching_names(
     When a list of query regular expressions is provided, the function checks each target string against each
     query regular expression and returns the indices of the matched strings and the matched strings.
 
-    If the :attr:`preserve_order` is False, the ordering of the matched indices and names is the same as the order
-    of the provided list of strings. This means that the ordering is dictated by the order of the target strings
-    and not the order of the query regular expressions.
+    If :attr:`preserve_order` is False, the matched indices and names are returned in the iteration order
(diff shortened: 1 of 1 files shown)


StafaH:

PR #6039, opened 2026-06-08
Fix AutoMate SoftDTW import fallback

ooctipus:
## Summary
- Make the AutoMate SoftDTW module importable when the optional `numba.cuda` backend fails to import.
- Fall back to the existing CPU SoftDTW implementation when `SoftDTW(use_cuda=True)` is requested but the CUDA backend is unavailable.
- Add a focused regression test that simulates the reported `numba.cuda.types.NPDatetime` import failure.

## Reproduction / Verification
- Checked the provided failure path: `assembly_env.py` imports `SoftDTW` from `soft_dtw_cuda.py`, which previously imported and decorated `numba.cuda` kernels at module import time.
- In an existing beta2 virtualenv with broken Numba imports, verified the patched `soft_dtw_cuda.py` imports successfully and a small `SoftDTW(use_cuda=True)` call falls back to CPU.

## Tests
- `git diff --cached --check`
- Parsed modified Python files with `ast.parse`
- `/home/zhengyuz/Projects/IsaacLab.wt/develop/.venv/bin/python -m pytest source/isaaclab_tasks/test/contrib/test_automate_soft_dtw_cuda.py -q`
- `/home/zhengyuz/Projects/IsaacLab.wt/beta2-automate-collision-stack/.venv/bin/python -m pytest source/isaaclab_tasks/test/contrib/test_automate_soft_dtw_cuda.py -q`
- Direct patched-module import/fallback check in `beta2-automate-collision-stack/.venv`

### DIFF
--- a/source/isaaclab_tasks/changelog.d/fix-automate-softdtw-numba-cuda-import.rst
+++ b/source/isaaclab_tasks/changelog.d/fix-automate-softdtw-numba-cuda-import.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed AutoMate assembly environment imports when the optional SoftDTW CUDA
+  backend cannot be imported from the installed Numba/Numba-CUDA stack.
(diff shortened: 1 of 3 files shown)


ooctipus:
Closing this implementation. The runtime fallback was too broad; the root cause is the uv dependency override allowing an incompatible pre-release Numba/NumPy stack. Replacing with a narrower dependency-constraint fix.

PR #6040, opened 2026-06-08
Remove AutoMate Numba SoftDTW dependency

ooctipus:
## Summary
- Remove `numba` from `isaaclab_tasks` dependencies.
- Replace AutoMate's Numba CUDA/CPU-JIT SoftDTW helper with a Torch implementation that runs on the input tensor device.
- Add a no-grad anti-diagonal SoftDTW path plus `forward_with_lengths(...)` so the AutoMate reward evaluates padded variable-length reference segments in one batched call instead of one SoftDTW call per environment.
- Clean up the autograd SoftDTW path to use a Torch DP table instead of Python row lists, with a clearer docstring.
- Remove the Numba CUDA warning environment variable from `run_w_id.py`.
- Add focused SoftDTW tests for the no-Numba path, hard DTW, normalized SoftDTW, variable-length padded SoftDTW, and finite backward gradients.

## Rationale
The original failure is not a sustainable place to solve with a NumPy pin. AutoMate only needs SoftDTW values for reward computation; it does not require the copied differentiable Numba implementation as a package-level dependency. Keeping Numba also exposes a second failure mode on RTX 5090: the old Numba CUDA kernel can fail at compile time with `CUDA_ERROR_UNSUPPORTED_PTX_VERSION` / unsupported PTX version.

This removes the dependency instead of constraining global NumPy resolution.

## Verification
- Focused tests pass in the develop venv: `python -m pytest source/isaaclab_tasks/test/contrib/test_automate_soft_dtw.py -q` (`5 passed`).
- Focused tests pass in the beta2 venv where Numba import is broken (`5 passed`).
- `git diff --check` passes.
- `py_compile` passes for the touched Python files.
- Old-vs-new SoftDTW CPU forward parity: 594 finite cases across `gamma={0.01,0.1,1.0}`, normalized/non-normalized valid cases, bandwidth `{None,2,20}`, and sequence lengths up to `B=8,N=10,M=100`; max absolute difference was `1.526e-04`.
- Mustafa's row/column Torch DP variant matched the current no-grad implementation exactly in direct forward checks; for `B=128,N=10,M=100`, it measured `82.497 ms` on CUDA versus `14.038 ms` for the anti-diagonal no-grad SoftDTW path, so this PR keeps the anti-diagonal path for reward inference while using the cleaner Torch DP-table style for autograd.
- For `gamma=0`, the old implementation returns `nan` on a simple hard-DTW case; the new implementation returns the expected hard-DTW value `1.0`.
- New SoftDTW autograd smoke test produces finite gradients.
- AutoMate reward parity: optimized length-aware reward path matches the original per-env reward loop on synthetic AutoMate-shaped data (`128` envs, `10` robot waypoints, `ref_len=100`, `gamma=0.01`); max absolute error was `0.0` on CPU and CUDA.

## Performance
Synthetic AutoMate-shaped reward benchmark on RTX 5090 with Torch `2.10.0+cu128`, `128` envs, `10` robot waypoints, `ref_len=100`, `gamma=0.01`, `no_grad`:

| Path | CPU median | CUDA median | CUDA peak allocated delta |
| --- | ---: | ---: | ---: |
| Per-env Torch reward loop | `141.617 ms` | `355.131 ms` | `15.372 MB` |
| Batched length-aware Torch reward | `13.483 ms` | `25.824 ms` | `15.372 MB` |

The peak CUDA allocation in this reward benchmark is dominated by the closest-state `torch.cdist` calculation, not by the SoftDTW table.

The previous Numba CUDA path could not be timed on this RTX 5090 because it fails locally with `CUDA_ERROR_UNSUPPORTED_PTX_VERSION`; the performance comparison above is against the direct per-env Torch replacement path that this PR would otherwise have used.

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/contrib/automate/soft_dtw_cuda.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/contrib/automate/soft_dtw_cuda.py
@@ -66,7 +66,7 @@ def _soft_dtw_no_grad(D: torch.Tensor, gamma: float, bandwidth: float) -> torch.
         i = torch.arange(max(1, diag - len_y), min(len_x, diag - 1) + 1, device=D.device)
         j = diag - i
 
-        if 0 < bandwidth:
+        if bandwidth > 0:
             keep = torch.abs(i - j) <= bandwidth
             i = i[keep]
             j = j[keep]
@@ -100,7 +100,7 @@ def _soft_dtw_variable_y_no_grad(
         i = torch.arange(max(1, diag - len_y), min(len_x, diag - 1) + 1, device=D.device)
         j = diag - i
 
-        if 0 < bandwidth:
+        if bandwidth > 0:
             keep = torch.abs(i - j) <= bandwidth
             i = i[keep]
             j = j[keep]
--- a/source/isaaclab_tasks/isaaclab_tasks/contrib/automate/soft_dtw_cuda.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/contrib/automate/soft_dtw_cuda.py
@@ -29,8 +29,8 @@
 import torch
 
 
-def _soft_dtw(D: torch.Tensor, gamma: float, bandwidth: float) -> torch.Tensor:
-    """Compute SoftDTW from a batched pairwise distance matrix using Torch ops."""
+def _soft_dtw_autograd(D: torch.Tensor, gamma: float, bandwidth: float) -> torch.Tensor:
+    """Compute SoftDTW using Torch ops that preserve autograd."""
     batch_size, len_x, len_y = D.shape
     inf = torch.full((batch_size,), float("inf"), device=D.device, dtype=D.dtype)
     prev_row = [inf] * (len_y + 2)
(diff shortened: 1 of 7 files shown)


source/isaaclab_tasks/test/contrib/test_automate_soft_dtw.py
```diff
@@ -0,0 +1,54 @@
...
+        / "soft_dtw_cuda.py"
+    )
+    spec = importlib.util.spec_from_file_location("automate_soft_dtw_under_test", module_path)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_soft_dtw_use_cuda_does_not_require_numba():
```
StafaH, line 22: Seems like this and below test can be combined, and use parameterized ('cpu', 'cuda')

source/isaaclab_tasks/isaaclab_tasks/contrib/automate/soft_dtw_cuda.py
```diff
@@ -26,427 +26,84 @@
...
-            1.0 / gamma.item(),
-            bandwidth.item(),
-            N,
-            M,
-            n_passes,
-            cuda.as_cuda_array(E),
-        )
-        E = E[:, 1 : N + 1, 1 : M + 1]
-        return grad_output.view(-1, 1, 1).expand_as(E) * E, None, None
+def _soft_dtw(D: torch.Tensor, gamma: float, bandwidth: float) -> torch.Tensor:
```
StafaH: How about the below (using torch only instead of python lists, better docstring)?

```
def _soft_dtw(D: torch.Tensor, gamma: float, bandwidth: float = -1) -> torch.Tensor:
    """
    Compute batched SoftDTW from a pairwise distance tensor.

    D: Tensor of shape (batch, len_x, len_y)
    gamma: SoftDTW smoothing parameter. If gamma == 0, computes hard DTW.
    bandwidth: Optional Sakoe-Chiba bandwidth. If <= 0, no band constraint.
    """
    batch_size, len_x, len_y = D.shape

    inf = torch.tensor(float("inf"), device=D.device, dtype=D.dtype)

    R = torch.full(
        (batch_size, len_x + 2, len_y + 2),
        inf,
        device=D.device,
        dtype=D.dtype,
    )

    R[:, 0, 0] = 0

    use_band = bandwidth > 0
    bandwidth = int(bandwidth) if use_band else max(len_x, len_y)

    for i in range(1, len_x + 1):
        j_start = max(1, i - bandwidth)
        j_end = min(len_y, i + bandwidth) + 1

        for j in range(j_start, j_end):
            r0 = R[:, i - 1, j - 1]
            r1 = R[:, i - 1, j]
            r2 = R[:, i, j - 1]

            if gamma == 0:
                softmin = torch.minimum(torch.minimum(r0, r1), r2)
            else:
                softmin = -gamma * torch.logsumexp(
                    torch.stack((-r0 / gamma, -r1 / gamma, -r2 / gamma), dim=0),
                    dim=0,
                )

            R[:, i, j] = D[:, i - 1, j - 1] + softmin

    return R[:, len_x, len_y]
```

StafaH (approved):

PR #5618, opened 2026-05-14
Make Isaac Lab Docker images run as non-root

sheikh-nv:
# Description

Makes the Isaac Lab base, ROS 2, and cuRobo Docker images run as a non-root runtime user by creating an `isaaclab` user after root-only setup and switching the final images to `USER isaaclab`.

The ROS 2 Dockerfile temporarily switches back to `USER root` for apt-based ROS setup, then restores `USER isaaclab` for runtime. The `installci` Dockerfile is intentionally unchanged.

This also updates deprecated Isaac Sim Dockerfile comments to point to the NGC Isaac Sim container page, removes the default root-allowance compose setting, updates Docker documentation, and adds CI coverage to verify the built base and cuRobo images do not run as root by default. ROS 2 is covered by the static Dockerfile regression test because this workflow does not build a ROS 2 image.

Fixes: N/A

Validation:
- `./isaaclab.sh -f`
- `git diff --check`
- `docker run ... /isaac-sim/python.sh -m pytest docker/test/test_dockerfile_nonroot.py -q` -> `7 passed, 1 skipped`
- Manual cuRobo runtime check confirmed non-root `uid=1000`

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

N/A

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (not applicable; no `source/<pkg>/` package touched)
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6061, opened 2026-06-09
Document non-root Docker volume fix for XR teleop startup

rwiltz:
# Description

Since the non-root Docker migration (https://github.com/isaac-sim/IsaacLab/pull/5618, first shipped in 3.0.0-beta2), the
container runs as user isaaclab (uid/gid 1000) with HOME=/root. Persistent
mounts at /root/.local/share/ov/data that were created by an older root-based
image (stale named volumes) or by Docker as auto-created bind-mount dirs are
owned by root, so the runtime user cannot write the extension-registry cache.
For XR teleop this aborts startup with a confusing, seemingly-unrelated error:
PermissionError: [Errno 13] Permission denied: '/root/.local/share/ov/data/exts'
...
No versions of omni.kit.xr.bundle.generic that satisfies: isaaclab.python.xr.openxr-3.0.0 ...
Exiting app because of dependency solver failure...
The XR bundle isn't actually missing — the registry never synced because its
cache dir couldn't be created. This PR documents the cause and fix:

docs/source/how-to/cloudxr_teleoperation.rst: adds an admonition to the
"Run with Docker" section explaining the failure and the fix (recreate/chown
volumes for Compose, pre-create/chown host dirs for single-container).
docs/source/deployment/docker.rst: warns that non-root prebuilt images
need bind-mount host dirs pre-created and chowned to uid/gid 1000, with a
copy-paste snippet.
Docs-only; no code changes and no changelog fragment (no source/<pkg>/ package
touched).

Fixes # (issue)

## Type of change

- Documentation update

## Checklist

### DIFF
--- a/docs/source/deployment/docker.rst
+++ b/docs/source/deployment/docker.rst
@@ -324,6 +324,21 @@ To pull the minimal Isaac Lab container, run:
 
   docker pull nvcr.io/nvidia/isaac-lab:3.0.0-beta1
 
+.. attention::
+
+  If the pre-built image you use runs as a **non-root** user (uid/gid 1000) -- as Isaac Lab
+  3.0.0-beta2 and later do -- the bind-mounted host directories below must be writable by that
+  user. Docker creates any missing bind-mount source directory as ``root``, which the non-root
+  runtime user cannot write to, leading to startup errors such as
+  ``PermissionError: [Errno 13] Permission denied: '/root/.local/share/ov/data/exts'``.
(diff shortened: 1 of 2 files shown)


AntoineRichard:
This is not in develop. Is there any reason why it would be in beta 2 but not develop? Should we bring this to develop? I'm approving this, @rwiltz can you make sure this is also pushed to develop?

AntoineRichard (approved):

PR #6068, opened 2026-06-09
Fix sensor ancestor suffix trimming

AntoineRichard:
# Description

Fixes rigid-body ancestor expression resolution for sensors mounted under a
non-rigid child when the terminal child suffix also appears earlier in the path.
The old implementation used an unbounded ``str.replace()``, which could remove
more than the sensor-relative suffix.

Simple example:

```text
target_expr   = /World/envs/env_.*/Robot/link/link
relative_path = link
old result    = /World/envs/env_.*/Robot
new result    = /World/envs/env_.*/Robot/link
```

The fix verifies that the relative path is a terminal suffix and slices only that
end segment.

No issue filed.

Validation:

- `./isaaclab.sh -p -m pytest source/isaaclab/test/sensors/test_sensor_base.py::test_rigid_body_ancestor_expr_trims_only_terminal_suffix -q`
- `PATH=/tmp/git-lfs-v3.7.1:$PATH ./isaaclab.sh -f`

## Type of change

## Screenshots

Not applicable.

## Checklist

- [x] I have made corresponding changes to the documentation, or documentation is not required for this change

### DIFF
--- a/source/isaaclab/isaaclab/sensors/sensor_base.py
+++ b/source/isaaclab/isaaclab/sensors/sensor_base.py
@@ -472,6 +472,12 @@ class SensorBase(ABC):
             return target_expr, None, None
 
         relative_path = prim.GetPath().MakeRelativePath(ancestor_prim.GetPath()).pathString
-        rigid_parent_expr = target_expr.replace("/" + relative_path, "")
+        suffix = "/" + relative_path
+        if not target_expr.endswith(suffix):
+            raise RuntimeError(
+                f"Failed to build rigid body ancestor expression: target expression {target_expr!r} does not end "
+                f"with relative path {relative_path!r}."
+            )
+        rigid_parent_expr = target_expr[: -len(suffix)]
         fixed_pos_b, fixed_quat_b = resolve_prim_pose(prim, ancestor_prim)
         return rigid_parent_expr, fixed_pos_b, fixed_quat_b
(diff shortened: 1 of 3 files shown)


source/isaaclab/isaaclab/sensors/sensor_base.py
```diff
@@ -472,6 +472,12 @@ def _resolve_rigid_body_ancestor_expr(
...
 
         relative_path = prim.GetPath().MakeRelativePath(ancestor_prim.GetPath()).pathString
-        rigid_parent_expr = target_expr.replace("/" + relative_path, "")
+        suffix = "/" + relative_path
+        if not target_expr.endswith(suffix):
+            raise RuntimeError(
+                f"Failed to build rigid body ancestor expression: target expression {target_expr!r} does not end "
+                f"with relative path {relative_path!r}."
+            )
+        rigid_parent_expr = target_expr[: -len(suffix)]
```
mmichelis, line 481: I might not be fully familiar with the sensor code, but here it feels like ancestor_prim.GetPath() already gets you the rigid_parent_expr, since the ancestor prim has been checked for rigid body api and is the closest/first ancestor that matches this.

mmichelis (commented):
Nice fix! I added a quick point that could be nice to discuss.

source/isaaclab/isaaclab/sensors/sensor_base.py
AntoineRichard: The difference here is that `ancestor_prim.GetPath()` will give you the concrete path of one instance (say env 0) but not the regex that this method should return,

mmichelis: Ahhh right the wildcard expressions, makes sense, thanks for the clarification!

mmichelis (approved):

ooctipus (approved):

PR #6073, opened 2026-06-09
Fix Newton Fabric body binding initialization

ooctipus:
## Summary

- initialize Fabric world matrix attributes before assigning `newton:index` during Newton USD/Fabric body binding
- skip missing Fabric body prims and invalid index attributes instead of writing through invalid attributes
- update the Fabric hierarchy once after body binding, matching the existing Fabric frame view initialization pattern

## Motivation

Dexsuite startup with Newton + Isaac RTX rendering can hang after contact sensor initialization with:

```
Currently unsupported type (unknown) for UsdAttribute.Set() at <invalid UsdAttribute>
```

The body binding loop was creating and setting `newton:index` before Fabric hierarchy world matrix state was initialized for each body prim. This mirrors the safer setup order used by `FabricFrameView`.

## Validation

- `pre-commit run --files source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py source/isaaclab_newton/changelog.d/fix-newton-fabric-body-binding.rst`
- `python3 tools/changelog/cli.py check develop`
- `python3 -m py_compile source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py`
- `git diff --check`

I attempted the reported training startup locally with `--max_iterations 1`, but this checkout has no active Isaac Lab Python environment and `isaaclab.sh` selected `/usr/bin/python3.12`, which is missing `gymnasium`.

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -11,7 +11,7 @@ import contextlib
 import ctypes
 import logging
 from abc import abstractmethod
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from typing import TYPE_CHECKING
 
 import warp as wp
@@ -284,6 +284,7 @@ class NewtonManager(PhysicsManager):
     _LocalSite = tuple[None, list[list[int]]]
     _SiteEntry = _GlobalSite | _LocalSite
     _cl_site_index_map: dict[str, _SiteEntry] = {}
+    _cl_fabric_body_bindings: list[tuple[str, int]] | None = None
     _world_xforms: list[wp.transform] | None = None
 
     @classmethod
@@ -692,6 +693,7 @@ class NewtonManager(PhysicsManager):
         NewtonManager._scene_data_backend = None
         NewtonManager._cl_pending_sites = {}
         NewtonManager._cl_site_index_map = {}
+        NewtonManager._cl_fabric_body_bindings = None
         NewtonManager._world_xforms = None
         NewtonManager._pending_extended_state_attributes = set()
         NewtonManager._pending_extended_contact_attributes = set()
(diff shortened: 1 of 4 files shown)


ooctipus:
Validation update:

* Created a clean detached worktree at `refs/remotes/upstream/develop` (`5124aa17e22`) and installed the pip Isaac Sim workflow there (`./isaaclab.sh --uv`, `./isaaclab.sh -i 'newton,ov[ovrtx],rl[rsl-rl]'`, `uv pip install ... isaacsim[all,extscache]==6.0.0.1`, then reran `./isaaclab.sh -i ...`).
* Baseline command with `--max_iterations 1` reproduced the hang: it initialized all four contact sensors, emitted the post-contact-sensor USD warning at the same point, then made no progress to solver initialization or environment setup for more than five minutes. SIGINT/SIGTERM did not unwind the process; I had to SIGKILL the repro child.
* Fixed branch command with the same installed stack and `--max_iterations 1` exits 0. It reaches `Initialize solver`, `Completed setting up the environment`, runs learning iteration `0/1`, reports `Training time`, and clears `SimulationContext`.
* Focused checks passed: `python3 -m py_compile source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py`, `git diff --check`, `pre-commit run --files source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py source/isaaclab_newton/changelog.d/fix-newton-fabric-body-binding.rst`, and `python3 tools/changelog/cli.py check develop`.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1027,17 +1027,55 @@ def start_simulation(cls) -> None:
...
+                        continue
+
+                    xformable_prim = usdrt.Rt.Xformable(prim)
+                    has_attr = (
+                        xformable_prim.HasFabricHierarchyWorldMatrixAttr()
+                        if hasattr(xformable_prim, "HasFabricHierarchyWorldMatrixAttr")
+                        else False
+                    )
+                    if not has_attr:
+                        xformable_prim.CreateFabricHierarchyWorldMatrixAttr()
```
pbarejko: Why do we need to create this attribute? I understand it's not there, but my question is why? 

The `omni:fabric:worldTransform` attribute  is created of all `UsdGeomImageable`-style prims during the Fabric population.

Why do we have USD and Fabric primitives out of sync?

ooctipus:
Updated after Piotr’s review: the `CreateFabricHierarchyWorldMatrixAttr()` part was not necessary for this bug.

I retested a narrower patch that leaves Fabric world-transform population alone and only guards the Newton body-index binding:

* skip missing/invalid Fabric prims from `body_label` before creating/setting `newton:index`
* validate the created `newton:index` attr before calling `Set()`
* keep the existing `HasWorldXform()` / `SetWorldXformFromUsd()` behavior for valid prims

That narrower version still fixes the reproduced hang: the same `newton_mjwarp,isaacsim_rtx_renderer,rgb64,single_camera` command with `--max_iterations 1` reaches solver initialization, completes environment setup, runs learning iteration `0/1`, and exits 0.

So I agree with the concern: we should not create Fabric hierarchy world-matrix attributes here. The root issue is the body binding loop assuming every Newton `body_label` resolves to a valid Fabric prim under replicated RTX startup. Some labels do not resolve to concrete valid Fabric prims at that point, and the old code then called `Set()` on an invalid attribute, matching the `UsdAttribute.Set() at <invalid UsdAttribute>` warning and hang.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1027,10 +1027,25 @@ def start_simulation(cls) -> None:
                 NewtonManager._usdrt_stage = None
             else:
                 NewtonManager._usdrt_stage = get_current_stage(fabric=True)
+                missing_body_count = 0
```
pbarejko: Why do we have missing bodies? I understand there are not there, but my question is why? Why do we have USD and Fabric primitives out of sync?

ooctipus:
Updated after the clone-plan review: the previous USD-replication approach and the helper-style Fabric-label fallback have been replaced.

Current fix:

- Newton body labels remain deterministic from the existing clone-plan label rewrite path (`_rename_builder_labels`).
- While `_rename_builder_labels` is already iterating source rows and clone-plan mappings, it now records `(source_body_path, destination_body_path, newton_body_index)` binding records for body labels.
- Newton assets still do not queue USD replication.
- During RTX/Fabric setup, `NewtonManager` consumes those binding records directly. For destination body prims absent from Fabric, it materializes a lightweight Fabric-only prim from the corresponding source body path, then binds `newton:index` and `omni:fabric:worldMatrix` for the existing transform sync path.
- This avoids reverse-resolving destination labels back to sources and avoids cloning USD specs for each environment.

Validation:

- `./isaaclab.sh train --rl_library rsl_rl --task Isaac-Dexsuite-Kuka-Allegro-Lift-v0 presets=newton_mjwarp,isaacsim_rtx_renderer,rgb64,single_camera --seed 42 --num_envs 4 --max_iterations 10` completed successfully.
- `pre-commit run --files ...` passed on touched files.
- `python3 tools/changelog/cli.py check develop` passed.
- Focused pytest for `source/isaaclab_newton/test/cloner/test_rename_builder_labels.py` could not run in this local env because `pytest` is not installed.

pbarejko (approved):

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1017,30 +1019,38 @@ def start_simulation(cls) -> None:
...
-                NewtonManager._usdrt_stage = get_current_stage(fabric=True)
-                for i, prim_path in enumerate(body_paths):
-                    prim = cls._usdrt_stage.GetPrimAtPath(prim_path)
-                    prim.CreateAttribute(cls._newton_index_attr, usdrt.Sdf.ValueTypeNames.UInt, True)
-                    prim.GetAttribute(cls._newton_index_attr).Set(i)
-                    # Tag with PhysicsRigidBodyAPI so cubric's eRigidBody mode
-                    # applies Inverse propagation (preserves Newton's world
-                    # transforms and derives local) instead of Forward.
-                    prim.AddAppliedSchema("PhysicsRigidBodyAPI")
+            body_paths = list(cls._model.body_label)
```
huidongc, line 1022: I am curious whether we should extract this code block as a helper function that we can add some more unit tests?

huidongc (approved):

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
huidongc: Could use agent to do this work.

PR #6074, opened 2026-06-09
Remove obsolete OVPhysX bootstrap hacks

marcodiiga:
# Description

Removes obsolete OVPhysX bootstrap compatibility code for the pinned
`ovphysx==0.5.9` runtime and hardens normal lifecycle cleanup:

- uses the public bootstrap and schema APIs without hiding loaded `pxr` modules
  or probing private wheel layout;
- registers only missing public schema providers;
- replaces forced-success process exit with guarded `atexit` cleanup and
  explicit runtime/view release;
- completes STOP and context cleanup even when a listener fails;
- preserves the one required STOP on same-context re-warm and the public,
  idempotent bootstrap call;
- isolates view cleanup between tests.

Dependency: `ovphysx==0.5.9`.

Fixes N/A

## Type of change

## Screenshots

N/A

## Test Plan

- `uv run --frozen python tools/changelog/cli.py check develop`
- focused lifecycle and view tests: `122 passed`
- Sphinx warning-as-error documentation build
- full 22-file OVPhysX backend sweep: 21 files passed; the unchanged CUDA
  articulation file hit intermittent assertions. An identical property-binding
  failure reproduced on exact `develop` merge-base `5322a1bdcff` (1/5 control
  runs and 1/4 PR runs).

## Checklist

- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` - CI handles that)

The OVPhysX backend tests still emit existing deprecation and runtime warnings
from surrounding test coverage.


source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,129 @@
...
+        "_physx": manager._physx,
+        "_locked_device": manager._locked_device,
+    }
+    try:
+        manager._physx = None
+        manager._locked_device = None
+        yield module
+    finally:
+        for name, value in saved.items():
+            setattr(manager, name, value)
```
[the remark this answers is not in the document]
marcodiiga: Fixed in d9056d570cf. The fixture now snapshots/restores the relevant OvPhysxManager class-level lifecycle fields, including _warmup_done, _usd_handle, _stage_path, _tmp_dir, _pending_clones, _scene_data_backend, and _physx_schemas_registered. It also copies _pending_clones and cleans up any test-created temporary directory before restoring the original state. Verified with the targeted lifecycle/scene-data/runtime-import tests: 16 passed.

marcodiiga:
Stalling this PR for now and marking it as draft. The cleanup intentionally targets the next ovphysx line via `ovphysx>=0.5,<0.6`, but the public package indexes currently only resolve `ovphysx<=0.4.13`. That makes dependency-resolution CI fail before this can be mergeable.

We should mark this ready again only after a public Python ovphysx wheel satisfying `>=0.5,<0.6` is available on the configured indexes, then rerun CI and the existing smoke matrix.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -265,36 +262,23 @@ def register_clone(
...
-        if any(plugin.name == "physxSchema" for plugin in registry.GetAllPlugins()):
-            cls._physx_schemas_registered = True
-            return
-        plugin_root = os.path.join(os.path.dirname(ovphysx.__file__), "plugins", "usd")
-        for sub in ("PhysxSchema/resources", "PhysxSchemaAddition/resources"):
-            path = os.path.join(plugin_root, sub)
-            if os.path.isdir(path):
-                registry.RegisterPlugins(path)
+        schema_paths = [str(path) for path in ovphysx.codeless_schema_paths()]
+        Plug.Registry().RegisterPlugins(schema_paths)
```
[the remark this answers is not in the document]
marcodiiga: Good catch. Fixed by case-insensitively comparing the provider name derived
from each path returned by `ovphysx.codeless_schema_paths()` with the names
already present in `Plug.Registry().GetAllPlugins()`, then registering only
missing provider paths. The dependency is pinned to `ovphysx==0.5.9`, which
exposes the documented `codeless_schema_paths()` helper. Tests cover one
provider already present, both providers present, and repeated calls.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -354,63 +335,62 @@ def step(cls) -> None:
...
-        cls._stage_usda = None
-        cls._warmup_done = False
-        # Drop the SceneDataBackend singleton: its cached ``TensorBinding`` handles
-        # point into the wheel's prior scene which we just cleared.
-        # The next :class:`SimulationContext` re-creates the backend in
-        # :meth:`initialize`. Matches Newton's lifecycle.
-        cls._scene_data_backend = None
-
-        super().close()
+        # Dispatch STOP while the runtime is still live. Asset and sensor callbacks
```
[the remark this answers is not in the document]
marcodiiga: Fixed. `SimulationContext` now stores the resolved manager class, so active
manager detection and STOP delivery work with string-based `class_type`
configuration. `PhysicsManager.close()` snapshots STOP listeners in callback
order, gives every listener a chance to run, clears callbacks and the active
manager's shared simulation state, and then raises a `RuntimeError` chained
from the first listener failure. `OvPhysxManager.close()` releases its runtime
from `finally`, and `SimulationContext.clear_instance()` continues its normal
remaining teardown before re-raising the listener failure. Regression coverage
includes a failed listener between live listeners, a collected weak listener,
shared-state cleanup, runtime release, and outer context cleanup.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -440,6 +420,24 @@ def _destroy_ovstage(cls) -> None:
...
+
+        OvPhysxView._close_all_for(physx)
+
+    @classmethod
+    def _prepare_physx_for_stage_reuse(cls) -> None:
+        """Invalidate stage-bound handles before resetting a cached runtime."""
+        physx = cls._physx
+        if physx is None:
+            return
+        cls.dispatch_event(PhysicsEvent.STOP, payload={})
```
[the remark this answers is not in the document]
marcodiiga: I kept the STOP dispatch on this path. `SimulationContext` is a singleton, so
the two-unclosed-context scenario is not reachable through the supported API:
constructing another context while one is active returns the existing instance.
The reachable cached-runtime path is a same-context re-warm. It dispatches STOP
once before closing views backed by that runtime, and the following
PHYSICS_READY rebuilds listener-owned views. The CPU re-warm test exercises this
path, while the GPU deformable regression verifies binding replacement.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -636,65 +630,22 @@ def _warmup_and_load(cls) -> None:
...
-
-        _hidden_pxr = {k: _sys.modules.pop(k) for k in list(_sys.modules) if k == "pxr" or k.startswith("pxr.")}
-        try:
-            _ovphysx_bootstrap = import_ovphysx()
-            _ovphysx_bootstrap.bootstrap()
-        finally:
-            _sys.modules.update(_hidden_pxr)
-
         ovphysx = import_ovphysx()
+        ovphysx.bootstrap()
```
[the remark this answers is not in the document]
marcodiiga: Kept the repeated call intentionally. This PR pins `ovphysx==0.5.9`; its public
`bootstrap()` docstring states that the function is idempotent and native
loading occurs at most once. `_construct_physx()` therefore invokes it before
each new runtime construction, including reconstruction after an explicit
release. The manager docstring records that contract, and a fresh-process
regression calls the public function twice.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -636,65 +630,22 @@ def _warmup_and_load(cls) -> None:
...
+            # Globally retained environments may otherwise keep TensorBinding DLPack
+            # caches alive until Python module finalization. Normal atexit cleanup runs
+            # before that phase and preserves the process's real exit status.
+            atexit.register(cls.close)
             cls._atexit_registered = True
 
     @staticmethod
     def _create_physx_instance(ovphysx: Any, ovphysx_device: str, gpu_index: int) -> Any:
-        """Create a PhysX instance for the declared or current OVPhysX runtime API.
+        """Create a PhysX instance through the pinned OVPhysX runtime API.
```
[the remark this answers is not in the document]
marcodiiga: Fixed. `atexit` now registers a guarded `_close_at_exit()` wrapper. It is a
no-op when no OVPhysX runtime exists, performs the full manager close only while
OVPhysX is active, otherwise releases the retained OVPhysX runtime without
touching another backend's shared callbacks, and logs and swallows exit-cleanup
exceptions. Tests cover the post-close no-op state, another active backend, and
failing exit cleanup.

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
```diff
@@ -229,6 +231,10 @@ class OvPhysxView:
             Defaults to ``False`` (lazy: bindings are created on first access).
     """
 
+    # Keep views reachable until manager shutdown so cached TensorBinding
+    # DLPack capsules can be destroyed before the OVPhysX runtime is released.
+    _live_views: ClassVar[set[OvPhysxView]] = set()
```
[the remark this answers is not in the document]
marcodiiga: I kept the strong registry because `binding_for()` exposes a raw binding that
callers can retain after dropping their local `OvPhysxView` reference; a
`WeakSet` could let the owner disappear before manager shutdown destroys that
binding. I addressed repeated PHYSICS_READY initialization by closing
`OvPhysxFrameView`'s previous root view before replacement. Same-context stage
reuse also closes every view backed by the cached runtime before rebuilding. A
regression test verifies that the previous frame root view is closed before
reinitialization.

source/isaaclab_ovphysx/test/sim/test_ovphysx_view.py
```diff
@@ -88,18 +95,27 @@ def __init__(self, n: int = 3, unavailable: set | None = None, all_unavailable:
...
 
 
 def _make_view(n: int = 3, unavailable: set | None = None, device: str = "cpu") -> OvPhysxView:
     return OvPhysxView(_FakePhysX(n=n, unavailable=unavailable), pattern="/World/env_*/body", device=device)
 
 
+@pytest.fixture(autouse=True)
+def _close_live_views():
+    yield
+    for view in tuple(OvPhysxView._live_views):
```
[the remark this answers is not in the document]
marcodiiga: Good catch. Fixed the fixture to snapshot the views that existed before each
test and close only the views added by that test, matching the sibling scene
backend fixture.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,260 @@
...
+        [sys.executable, "-c", script],
+        text=True,
+        capture_output=True,
+        timeout=120,
+        check=False,
+    )
+    output = completed.stdout + completed.stderr
+
+    assert completed.returncode == 0, output[-8000:]
+    assert "NORMAL_ATEXIT" in output, output[-8000:]
```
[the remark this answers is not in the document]
marcodiiga: Fixed. The child installs a `sys.unraisablehook` sentinel, requires the normal
atexit and STOP markers, and rejects the known atexit error markers. A
negative-control child raises an uncaught `RuntimeError` and verifies exit
status 1 while the normal atexit cleanup still runs.

source/isaaclab/isaaclab/physics/physics_manager.py
```diff
@@ -428,18 +432,42 @@ def after_visualizers_render(cls) -> None:
...
-            cls.dispatch_event(PhysicsEvent.STOP)  # notify listeners before cleanup
+            matching = [
+                (callback, order)
+                for event, callback, order, _name, _subscription in cls._callbacks.values()
+                if event == PhysicsEvent.STOP
+            ]
+            matching.sort(key=lambda item: item[1])
+            for callback, _order in matching:
+                try:
+                    callback(None)
```
AntoineRichard: **Important — real PhysX STOP failures still bypass this aggregation.** Asset and sensor callbacks call `PhysicsManager.safe_callback_invoke(..., physics_manager=PhysxManager)`. That helper stores the exception through `PhysxManager.store_callback_exception()` and returns, so this `try` sees no exception and `close()` succeeds. I reproduced this with a manager exposing `store_callback_exception`: the `ValueError` remained stored after `close()` returned. Route STOP through one failure-aware dispatcher and drain any backend-stored callback exception before clearing state. The regression test should exercise the `safe_callback_invoke` path, not only a directly raising lambda.

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -833,7 +836,11 @@ def clear_instance(cls) -> None:
         if cls._instance is not None:
             # Close physics manager FIRST to detach PhysX from the stage
             # This must happen before clearing USD prims to avoid PhysX cleanup errors
-            cls._instance.physics_manager.close()
+            physics_error: Exception | None = None
+            try:
+                cls._instance.physics_manager.close()
+            except Exception as exc:
+                physics_error = exc
```
AntoineRichard: **Important — delaying only the physics exception does not guarantee teardown.** If any subsequent `viz.close()`, stage close, or cache cleanup raises, execution exits before the remaining phases and before `_instance = None`; the saved STOP failure is also lost. Thus the claimed "finish cleanup" behavior holds only while every later operation succeeds. Run each teardown phase independently while collecting failures, clear the singleton in a `finally`, then report errors with deterministic precedence. Extend this test with both a physics failure and a later visualizer failure.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+        from isaaclab_ovphysx.sim.views import OvPhysxView
+
+        sim = SimulationContext(SimulationCfg(physics=OvPhysxCfg(), device="cpu", dt=1.0 / 60.0))
+        assert sim.physics_manager is OvPhysxManager
+        OvPhysxManager.register_callback(
+            lambda _payload: print("OVPHYSX_STOP", flush=True),
+            PhysicsEvent.STOP,
+            wrap_weak_ref=False,
+        )
+        obj = RigidObject(
```
AntoineRichard: **Important — this unmarked CPU test has an accidental CUDA dependency.** `RigidObject` allocates pinned CPU buffers, which calls Warp's CUDA allocator even with `device="cpu"`. On a machine without a CUDA-capable device, both new subprocess tests fail here before reaching the atexit assertions (`2 failed, 118 passed, 2 skipped` in the focused run). Build a minimal USD rigid body and raw tensor binding for this lifecycle probe, or mark the tests as GPU-only; the minimal fixture is preferable because shutdown semantics are not GPU-specific.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+
+    manager._construct_physx("cpu", 0)
+
+    assert sys.modules["pxr"] is host_pxr
+    assert sys.modules["pxr.Usd"] is host_usd
+    assert registrations == [manager._close_at_exit]
+
+
+def test_pinned_bootstrap_is_idempotent_in_fresh_process():
+    completed, output = _run_child("import ovphysx\novphysx.bootstrap()\novphysx.bootstrap()\n")
```
AntoineRichard: **Moderate — this tests the pinned dependency, not IsaacLab.** It invokes `ovphysx.bootstrap()` directly and never exercises project code, so no IsaacLab regression can make it fail. The exact `ovphysx==0.5.9` pin owns this contract, and a fresh subprocess adds cost. Remove this test; if `_construct_physx()` call behavior needs a guard, use the existing fake module and assert how IsaacLab invokes `bootstrap()` across constructions.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+
+    assert completed.returncode == 0, output[-8000:]
+
+
+def test_release_destroys_bindings_before_runtime_and_stage(monkeypatch, manager_module):
```
AntoineRichard: **Moderate — consolidate the duplicated release-order test.** `test_manager_attaches_and_releases_owned_ovstage` in `test_ovphysx_scene_data_backend.py` already verifies `reset_stage -> wait_op -> release -> OVStage.destroy`. This repeats the same fake lifecycle only to add `close_views`. Extend the existing ownership test with that event, or keep this test and remove the older duplicate, so teardown order has one authoritative regression.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+    return completed, completed.stdout + completed.stderr
+
+
+def _assert_no_atexit_errors(output: str) -> None:
+    assert "ATEXIT_UNRAISABLE" not in output, output[-8000:]
+    assert "Exception ignored in atexit callback" not in output, output[-8000:]
+    assert "Error in atexit._run_exitfuncs" not in output, output[-8000:]
+
+
+def test_retained_binding_exits_through_normal_atexit():
```
AntoineRichard: **Moderate — the stronger failure-status case below subsumes this process test.** It executes the same script and checks `NORMAL_ATEXIT`, `OVPHYSX_STOP`, and the absence of atexit errors, then additionally proves an uncaught failure retains exit status 1. This case adds only the clean exit-code assertion at the cost of a second full runtime process. Keep the stronger case and cover any remaining clean no-op behavior with the existing unit-level atexit test.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -209,11 +207,10 @@ class OvPhysxManager(PhysicsManager):
...
     _requires_full_stage: ClassVar[bool] = False
-    # Device the process is locked to once :meth:`_warmup_and_load` constructs the
-    # ``ovphysx.PhysX`` instance for the first time.  ``ovphysx<=0.3.7`` enforces
-    # a process-global device-mode lock at the C++ layer (see HACK note on
-    # :meth:`_release_physx`); we mirror it here so a clear Python error is raised
-    # if a later :class:`~isaaclab.sim.SimulationContext` requests a different device.
+    # First device selected by IsaacLab for this process. OVPhysX 0.5.9 makes
+    # CPU-only mode sticky but does not symmetrically lock GPU mode. IsaacLab
+    # nevertheless pins the first choice so all contexts follow one predictable
+    # process lifecycle and a CPU-first process is never asked to switch modes.
```
AntoineRichard: **Comment hygiene:** this is a version-specific narrative for a policy choice, and the same rationale is repeated in `_warmup_and_load()` and its error. Keep the comment functional and durable, for example: `# Device mode is process-wide; later contexts must reuse the first selected device.` The runtime error already provides the user-facing remediation.

source/isaaclab_ovphysx/changelog.d/remove-ovphysx-bootstrap-hacks.minor.rst
```diff
@@ -0,0 +1,18 @@
...
+^^^^^
+
+* Fixed kitless OVPhysX shutdown to destroy cached tensor bindings before the
+  runtime is released. Normal process-exit cleanup now preserves Python
+  handlers and the real exit status instead of forcing success.
+* Fixed startup to use the public OVPhysX 0.5.9 bootstrap and all codeless
+  schema paths without hiding host USD modules, probing private wheel paths, or
+  replacing a schema plugin already supplied by the host.
+* Fixed repeated frame-view initialization to release the previous tensor view.
+* Fixed process-exit cleanup to avoid leaking atexit exceptions or disturbing
```
AntoineRichard: **Documentation duplication:** the first `Fixed` entry already describes normal process-exit cleanup and preserved handlers/status. This repeats the same fix in implementation terms. Fold active-backend isolation into that first entry or remove this bullet so the changelog states each user-visible change once.

AntoineRichard (changes requested):
Requesting changes for three functional lifecycle/test issues and several targeted hygiene problems. The inline comments include concrete failure paths and focused remediations.

Verification at `fda86746`:
- Focused 122-test lifecycle/view suite: 118 passed, 2 skipped, 2 failed. Both failures are the new process-exit tests, which reach a CUDA-dependent pinned allocation despite configuring CPU physics.
- Direct reproduction confirmed that a STOP callback routed through `safe_callback_invoke(..., physics_manager=<manager with store_callback_exception>)` returns from `close()` normally with the exception merely stored.

The main code paths are close, but teardown must preserve failures consistently across backends and the regression coverage should be smaller, portable, and non-duplicative.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+    @classmethod
+    def set_cpu_mode(cls, enabled):
+        pass
+
+    def __init__(self, active_cuda_gpus=None, config=None):
+        self.active_cuda_gpus = active_cuda_gpus
+        self.config = config
+
+
+@pytest.fixture
```
AntoineRichard, line 37: **Test compaction:** This fixture duplicates pytest restoration machinery and infers state from attribute-name suffixes. Define explicit test-state defaults and install fresh values with `monkeypatch.setattr(manager, name, value)`; pytest will restore the original class attributes automatically. That removes `saved`, the copy logic, `try/finally`, and the duplicate corrections on lines 66–69—roughly twenty lines of brittle bookkeeping—while making every required initial value explicit.

source/isaaclab_ovphysx/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -0,0 +1,414 @@
...
+
+    assert events == [
+        ("close_views", physx),
+        "reset",
+        ("wait", 9),
+        "destroy_stage",
+    ]
+
+
+def test_frame_view_reinitialization_closes_previous_root_view(monkeypatch):
```
AntoineRichard: **Keep this module manager-focused.** This test exercises `OvPhysxFrameView._on_physics_ready()` and never touches `OvPhysxManager`. Move it alongside the existing FrameView coverage in `test_views_xform_prim_ovphysx.py`. That keeps this file from becoming a lifecycle catch-all and makes the view-specific regression discoverable.

AntoineRichard (commented):
Follow-up on test scope: this new 414-line file mixes unrelated concerns and carries more fixture/process scaffolding than the lifecycle contract needs. Please compact it using these inline suggestions together with the previously requested redundant-test removals and minimal process fixture. The goal is focused coverage and less maintenance surface, not an arbitrary line limit.

kellyguo11 (approved):

AntoineRichard (approved):

kellyguo11:
let's look into why the rendering correctness tests are failing on this one

PR #5700, opened 2026-05-20
[CI] Cross-platform — Part 3: Windows workflow

hujc7:
## Summary

Part 3 of the cross-platform CI series. Adds `.github/workflows/windows-ci.yaml` — CI pipeline for Windows GPU self-hosted runners. Same shape as `arm-ci.yaml` (Part 2, #5698) but native install path instead of Docker.

- Tier 1 (smoke + install): `general-windows`, `install-windows` (uv install + wheel build + reinstall), `kit-launch-windows`.
- Tier 2 (meaningful, marker-driven): `path-io-windows`, `perception-windows`, ...
- All jobs `continue-on-error: true` while runners stabilize.
- All pytest invocations use `--timeout=N` + `--timeout-method=signal` so hung tests fail fast (fixes the previous perception-windows pattern where Vulkan failures hung the job instead of erroring).
- Marker-driven discovery: `pytest <root> -m windows_ci --continue-on-collection-errors`. Adding a new Windows-safe test requires only tagging it `windows_ci`, no yaml edit.

## Series

PRs prefixed with `[CI] Cross-platform —`. Current siblings:

- Part 1 (base) — #5695 — Markers and shared scaffolding
- Part 2 — #5698 — ARM/Spark workflow
- Part 3 (this PR) — Windows workflow

Depends on Part 1 (#5695). Independent of Part 2 (#5698).

## Test plan

- [x] `./isaaclab.sh -f` (pre-commit) passes.
- [ ] windows-ci.yaml triggers on PR push; all jobs run on Windows runners.
- [ ] perception-windows fails fast on Vulkan/runtime errors instead of hanging.

PR #6086, opened 2026-06-09
[CI] Cross-platform — Part 3: Windows workflow

hujc7:
## Summary

Same-repo version of #5700, opened from a branch in `isaac-sim/IsaacLab` (not a fork) so the Windows CI job receives the org secrets it needs. GitHub does not deliver workflow secrets to `pull_request` runs from forks, so the fork-based #5700 cannot exercise the authenticated Isaac Sim install — this PR can.

Adds `.github/workflows/windows-ci.yaml` — CI pipeline for Windows GPU self-hosted runners, native (non-Docker) install path.

- Installs the **develop-aligned** Isaac Sim from the internal Artifactory index (via `tools/resolve_isaacsim_develop.py`), so native Windows tests the same Sim build as the Linux/ARM develop containers instead of the older public pip release.
- Authenticates the internal index with the `ISAACSIM_ARTIFACTORY_READONLY_USERNAME` / `ISAACSIM_ARTIFACTORY_READONLY_PASSWORD` org secrets (anonymous Artifactory access was removed).
- Single consolidated job: setup → deps smoke → path-io → Kit headless boot → cartpole training smoke (state + perception) → cartpole-camera perception → wheel build + reinstall.
- All pytest invocations use `--timeout` so hung tests fail fast instead of hanging the job.

## TEMP (revert before final review)

To save runner time/cost while iterating, the heavy non-Windows PR workflows are force-skipped (each marked `TEMP`): Docker + Tests, Installation Tests, Build PIP Wheel, License Check, Multi-GPU.

## Notes

- The durable answer for fork/contributor PRs is to bake the read-only Artifactory creds into the self-hosted runner image (per the platform security guidance), removing the need for workflow secrets entirely. This PR validates the wiring works with the secrets in the meantime.
- Supersedes #5700.

## Test plan

- [ ] windows-ci.yaml triggers on this same-repo PR; setup step authenticates and installs develop-aligned Isaac Sim.
- [ ] perception steps fail fast on Vulkan/runtime errors instead of hanging.


hujc7:
Closing. The same-repo PR was to test whether fork-PR secret withholding was the blocker — it isn't: the Windows CI guard still reports the secrets empty on this in-repo run, so the issue is that `ISAACSIM_ARTIFACTORY_READONLY_*` isn't scoped to isaac-sim/IsaacLab at all (not the fork rule). Continuing on the original fork PR #5700.

PR #6093, opened 2026-06-10
Fix RMPFlow teleoperation related bugs

rebeccazhang0707:
# Description

The lula-backed RMPFlow controller re-converted the end-effector orientation target to (x, y, z, w) before building its rotation matrix, even though IsaacLab quaternions are already in that order. Drop the extra conversion so the already-xyzw target feeds matrix_from_quat directly. This makes all RMPFlow tasks hold their reset pose (Agibot right arm total joint drift over 60 zero-action steps drops from ~12.5 to ~0.2 rad) with no regression on the Galbot tasks. Fixes bug (6288451)

Add a parametrized regression test that resets each RMPFlow task and asserts the arm settles and stays close to its reset pose under a zero command. Verified the test fails on the affected tasks when the bug is reintroduced.

Also adjust task defaults made possible by the fix: set sensible default viewer poses for the Agibot upright-mug and Galbot cube-stack tasks, and remove now-redundant zero-valued body offsets. For bug (6288480)

Add documentation for humanoid teleoperation instruction with space mouse/keyboard. For bug (6288401)

## Screenshots
Screenshot from 2026-06-10 13-08-25

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Documentation update

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6099, opened 2026-06-10
Fix RMPFlow teleoperation related bugs (develop)

rebeccazhang0707:
## Description

Cherry-pick of #6093 ("Fix RMPFlow teleoperation related bugs") onto `develop`.

Fixes the RMPFlow end-effector orientation quaternion convention so a zero
relative command holds the reset pose instead of diverging, adds a regression
test pinning that behavior, and updates the humanoid teleop docstrings.

## Dependency / stacking

**Depends on #6025** (Newton-backend migration for stack/place RMPFlow tasks),
which must merge into `develop` first.

This branch is built on top of #6025's branch, so until #6025 merges this PR's
diff will also show #6025's commits. Once #6025 lands in `develop`, those commits
drop out of the diff automatically (the branch will be rebased onto `develop` if
needed). Please review/merge #6025 before this one.

## Type of change


AntoineRichard (approved):
Thanks for cherry picking this!

PR #6150, opened 2026-06-11
Fixes backend asset state invalidation + provides a skip

AntoineRichard:
# Description

Fix stale cached pose and velocity state after backend asset state writers push data into simulation.

This updates the Newton, PhysX, and OV PhysX backend asset writers to invalidate the affected cached
data through shared data-class reset helpers.

Fixes # (issue): N/A

## Type of change

## Screenshots

N/A

## Validation

- `./isaaclab.sh -f` (all hooks passed except `check-git-lfs-pointers`, which failed because `git-lfs` is not installed locally)
- `SKIP=check-git-lfs-pointers ./isaaclab.sh -f`

## Checklist


source/isaaclab_newton/isaaclab_newton/assets/rigid_object_collection/rigid_object_collection.py
```diff
@@ -740,24 +746,21 @@ def write_body_link_velocity_to_sim_index(
...
             device=self.device,
         )
-        # Invalidate dependent timestamps
-        self.data._body_link_state_w.timestamp = -1.0
-        self.data._body_state_w.timestamp = -1.0
-        self.data._body_com_state_w.timestamp = -1.0
-        SimulationManager.invalidate_fk(env_ids=env_ids, articulation_ids=self._root_view.articulation_ids)
+        # Invalidate dependent timestamps. The link velocities were just written, so they must not be
+        # invalidated. ``reset_velocity`` already triggers the simulation-side FK invalidation.
+        if not skip_forward:
```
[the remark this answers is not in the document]
AntoineRichard: Should be fixed now?

ooctipus (approved):

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -126,10 +127,89 @@ def update(self, dt: float) -> None:
...
+        """
+        # Reset the root com pose only if needed.
+        if from_link:
+            self._root_com_pose_w.timestamp = -1.0
+        # Always reset the body poses
+        self._body_link_pose_w.timestamp = -1.0
+        self._body_com_pose_w.timestamp = -1.0
+        # Force refresh on all the root states
+        if self._root_state_w is not None:
+            self._root_state_w.timestamp = -1.0
```
hujc7: as this pattern is very common, wondering if we can create a helper function for individual reset. Or one takes a list input and reset them all. One of the benefit might be to avoid typos, e.g. checking on property but actually resetting a different one.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
```diff
@@ -162,22 +163,83 @@ def update(self, dt: float) -> None:
...
+        next read of any property that depends on body poses or velocities. The physics instance is
+        absent under the mocked-interface tests, in which case the refresh is skipped.
+        """
+        if self._fk_timestamp < self._sim_timestamp:
+            physx_instance = OvPhysxManager.get_physx_instance()
+            if physx_instance is not None:
+                physx_instance.update_articulations_kinematic()
+            self._fk_timestamp = self._sim_timestamp
+
+    def reset_pose(self, from_link: bool = True) -> None:
```
hujc7: nit: the correspondence is consistent, reset_pose -> from_link, reset_v -> from_from. But I am wondering if it be generalized one lever further, e.g. from: ResetSource = ResetSource.LINK, or from: ResetSource = "link", so it's more consistent. Maybe a little too much.

source/isaaclab_newton/isaaclab_newton/assets/rigid_object/rigid_object_data.py
```diff
@@ -123,6 +124,75 @@ def update(self, dt: float) -> None:
...
+        ``-1.0`` to force a refresh on the next read of any property that depends on
+        body poses (``body_link_pose_w``, ``body_com_pose_w`` and the composite body
+        state buffers).
+        """
+        if self._fk_timestamp < self._sim_timestamp:
+            SimulationManager.forward()
+            self._fk_timestamp = self._sim_timestamp
+
+    def reset_pose(
+        self, env_ids: wp.array | torch.Tensor | None = None, env_mask: wp.array | None = None, from_link: bool = True
```
hujc7: Is this needed? Seems an outlier.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
```diff
@@ -345,6 +345,7 @@ def write_root_pose_to_sim_index(
         *,
         root_pose: torch.Tensor | wp.array,
         env_ids: Sequence[int] | torch.Tensor | wp.array | None = None,
+        skip_forward: bool = False,
```
hujc7, line 348: nit: Curious on this. The usage is commonly `not skip_forward`, would e.g. "force_forward", or something positive also serve the purpose and avoids double negation? At the end, would skipping be more common?

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -134,6 +134,86 @@ def _ensure_fk_fresh(self) -> None:
...
+        if self._root_com_state_w is not None:
+            self._root_com_state_w.timestamp = -1.0
+        # Force refresh on all the body com states
+        if self._body_state_w is not None:
+            self._body_state_w.timestamp = -1.0
+        if self._body_link_state_w is not None:
+            self._body_link_state_w.timestamp = -1.0
+        if self._body_com_state_w is not None:
+            self._body_com_state_w.timestamp = -1.0
+        # Why do we have _fk_timestamp and invalidate_fk? _fk_timestamp is on the data-side,
```
hujc7: nit: maybe consider adding a  `# NOTE:` here and make it more declarative, e.g. "the reason to have....".  I was thinking it might be a todo when first reading it.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/rigid_object/rigid_object_data.py
```diff
@@ -138,6 +138,44 @@ def update(self, dt: float) -> None:
         # since we do finite differencing.
         self.body_com_acc_w
 
+    def reset_pose(self, from_link: bool = True) -> None:
```
hujc7: As the function signature differs and there's no base class declaration, I think those could be declared as _private_function to indicate they are internal.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -134,6 +134,86 @@ def _ensure_fk_fresh(self) -> None:
             SimulationManager.forward()
             self._fk_timestamp = self._sim_timestamp
 
+    def reset_pose(
```
hujc7: Is some positive/negative test needed to catch the change from the old behavior?

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
AntoineRichard: I followed Genesis here, they have skip_forward flag set to false by default. But that's fair point.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
AntoineRichard: It's mostly a clean-up, it does not change the behavior except for the velocities, the pose were already covered in the standard tests. I'll see if we can add the velocities.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/rigid_object/rigid_object_data.py
AntoineRichard: Agreed, thanks for catching the missing base class implementations.

source/isaaclab_newton/isaaclab_newton/assets/rigid_object/rigid_object_data.py
AntoineRichard: Yes, at least that's what can flow inside of this.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
AntoineRichard: I felt like a boolean check would have been faster that's why I went with the from_link. I'll double check for homogeneity.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Initially I had a getattr / setattr path, but in the end you need to write the string anyway, I'm not sure it makes that big of a difference, it's a bit more compact for sure, but then it's a bunch of getattr / setattr not sure it's better.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
AntoineRichard: Going to keep `skip_forward` as-is. It mirrors Genesis's rigid solver convention exactly — `RigidSolver.set_qpos` / `set_base_links_pos` / `set_base_links_quat` take `skip_forward: bool = False` and gate the forward-kinematics pass on `if not skip_forward:` ([rigid_solver.py](https://github.com/Genesis-Embodied-AI/Genesis/blob/main/genesis/engine/solvers/rigid/rigid_solver.py)). Our `write_*_to_sim` methods sit at that same level, and the flag is now propagated uniformly across all three backends plus the base classes, so a rename would diverge from that precedent and churn the whole write surface. Defaulting to `False` keeps "forward runs unless explicitly skipped" as the safe default.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
AntoineRichard: Keeping the `from_link`/`from_com` booleans — there are only the two states and they are documented per-method, so a `ResetSource` enum would add indirection without much benefit here. (Agreed it would be a little too much.)

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/rigid_object/rigid_object_data.py
AntoineRichard: Done — renamed `reset_pose`/`reset_velocity` to `_reset_pose`/`_reset_velocity` across all three backends and declared them as abstract methods on `BaseArticulationData`/`BaseRigidObjectData`/`BaseRigidObjectCollectionData`, so they are internal but keep a discoverable contract.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Added `isaaclab.utils.buffers.reset_timestamps(buffers)` — it takes a list and invalidates each non-`None` buffer in one call. Every backend's `_reset_pose`/`_reset_velocity` now uses it, so each buffer is named exactly once at the call site and the repeated `if self._x is not None: self._x.timestamp = -1.0` blocks (and the check-one-reset-another typo risk you flagged) are gone.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
AntoineRichard: Reworded to a declarative `# NOTE:` block in both `_reset_pose` and `_reset_velocity`.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
AntoineRichard: Added the velocity coverage: a new `test_write_root_state_functions_data_consistency` for the articulation, plus priming in the rigid-object `test_write_state_functions_data_consistency`. The existing test did not actually exercise the invalidation — the derived caches recompute on first access unless they are primed at the current timestamp before the write — so I added that priming. Verified both tests fail if the `reset_velocity` invalidation is removed. The `body_com_pose_w` staleness greptile flagged is fixed and covered as well.

PR #6164, opened 2026-06-12
Fix beta2 docs deploy selection

ooctipus:
## Summary
- include release/3.0.0-beta2 in the multi-version docs deploy branch whitelist
- exclude prerelease tags from the deploy build so the broken v3.0.0-beta tag does not block beta2 publication
- add lazy_loader to docs requirements to match package imports used by newer docs

## Context
Manual Docs workflow run 27398053912 failed while building the v3.0.0-beta tag before deploy, so release/3.0.0-beta2 docs were not published.

## Testing
- python3 -c 'import yaml; yaml.safe_load(open(".github/workflows/docs.yaml")); print("yaml ok")'
- git diff --check

### DIFF
--- a/.github/workflows/check-links.yml
+++ b/.github/workflows/check-links.yml
@@ -75,6 +75,8 @@ jobs:
           --exclude 'user@'
           --exclude 'helm\.ngc\.nvidia\.com'
           --exclude 'slurm\.schedmd\.com'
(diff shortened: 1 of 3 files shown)


AntoineRichard (approved):
LGTM

PR #6185, opened 2026-06-14
Preserve LEAPP metadata for manual deploy inputs

ashwinvkNV:
# Description

This PR adds a metadata-preserving manual LEAPP input path for deploy observation tensors.

Some deploy workflows need to expose a higher-level observation tensor as the LEAPP input boundary instead of tracing every simulator-side dependency used to compute that observation. For the gear assembly ROS inference task, the deploy runtime provides arm joint state and shaft pose inputs directly, while the Isaac Lab environment may compute those values from simulator-only state during training/export.

The new `leapp_input_tensor(...)` helper is a no-op outside LEAPP export, but during export it can annotate a computed tensor as a LEAPP input with `kind`, `element_names`, and `isaaclab_connection` metadata. This keeps exported YAML compatible with Isaac ROS Deploy input routing, which selects converters using the input `kind`.

This PR also updates the gear assembly deploy observations so the exported policy inputs are:

- `robot_joint_pos`: 7 arm joints, `state/joint/position`
- `robot_joint_vel`: 7 arm joints, `state/joint/velocity`
- `gear_shaft_pos`: `state/body/position`
- `gear_shaft_quat`: `state/body/rotation`

The action output remains arm-only and keeps its existing `target/joint/position` metadata.

Fixes # 

## Type of change

## Screenshots

Not applicable.

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/utils/leapp/manual_inputs.py
+++ b/source/isaaclab/isaaclab/utils/leapp/manual_inputs.py
@@ -1,0 +1,114 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Helpers for manually exposing computed tensors as LEAPP inputs."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import torch
+
+
+class _ManualTensorView:
+    """Small adapter matching the ``data.<property>.torch`` access pattern."""
+
+    def __init__(self, tensor: torch.Tensor):
+        self._tensor = tensor
+
+    @property
+    def torch(self) -> torch.Tensor:
+        return self._tensor
+
+
+def _cache_manual_data_proxy(data_proxy: Any, real_data: Any, property_name: str, tensor: torch.Tensor) -> None:
+    """Cache a manual input tensor on a LEAPP data proxy."""
+    manual_cache = object.__getattribute__(data_proxy, "_manual_cache")
+    manual_cache[(id(real_data), property_name)] = _ManualTensorView(tensor)
+
+
+def leapp_real_env(env: Any) -> Any:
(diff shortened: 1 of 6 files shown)


source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/mdp/observations.py
```diff
@@ -180,8 +250,19 @@ def __call__(
         Returns:
             Gear shaft orientation tensor of shape (num_envs, 4)
         """
-        # Get base quaternion
-        base_quat = self.asset.data.root_quat_w.torch
+        # Get the raw shaft/base quaternion. During LEAPP export, read it from
+        # the real env and annotate it before canonicalization so the ONNX graph
+        # owns the ``qw >= 0`` sign convention.
+        real_env = leapp_real_env(env)
+        base_quat = real_env.scene[self.asset_cfg.name].data.root_quat_w.torch
```
frlai, line 257: yeah here can just be
`base_quat = annotate.input_tensor('policy', TensorSemantics('gear_shaft_quat', base_quat, kind=InputKindEnum.BODY_ROTATION, element_names=QUAT_XYZW_ELEMENT_NAMES)`

similar to everything above

source/isaaclab/isaaclab/utils/leapp/manual_inputs.py
frlai, line 1: why was this needed?
this seems like a re-implementation of leapp tensors.

this seems to also hook onto the env proxy which I don't think is needed. That is only for automatic annotation at the source.

frlai (commented):
I think the AI was trying to do this to preserve the ability to run everything in simulation using deploy.py. I'm not sure if that's a goal for us but if you're able to run everything in deploy.py using this method that would actually be pretty interesting and might be a useful change. But for a MVP, most of these changes can be cut down i believe.

AntoineRichard:
@ashwinvkNV @frlai should we move this to draft?

AntoineRichard:
If this is not needed anymore can this be closed?

frlai:
@ashwinvkNV I believe you are going with directly using leapp api for special observations right? We shouldn't need this anymore.

PR #6219, opened 2026-06-21
Expose Same-Step terminal observations

### LINKED ISSUE #3095: [Proposal] All RL environment should use `gymnasium.VectorEnv` rather than `gymnasium.Env`

ooctipus:
## Summary
- Preserve Isaac Lab existing Same-Step autoreset behavior without adding autoreset mode config.
- Declare `metadata["autoreset_mode"] = gym.vector.AutoresetMode.SAME_STEP` on RL envs.
- Capture terminal observations before the Same-Step reset and expose them through `extras["final_obs"]` for Direct RL, Manager-based RL, and Direct MARL envs. The captured observation has the same observation noise applied as the returned observation.
- Gate the capture behind an opt-in `compute_final_obs` flag (default `False`) on `DirectRLEnvCfg`, `DirectMARLEnvCfg`, and `ManagerBasedRLEnvCfg`. With the default the previous behavior is preserved: nothing is captured, `extras["final_obs"]` is not populated, and the extra observation computation is skipped.
- Update `Sb3VecEnvWrapper` to use `extras["final_obs"]` as SB3 `terminal_observation` when envs reset on the same step (falls back to the previous behavior when `final_obs` is absent).

## Related
- Addresses the Same-Step autoreset semantics and terminal-observation exposure required by the migration to `gymnasium.vector.VectorEnv` in #3136 (which fixes #3095). This PR declares the autoreset mode in environment metadata and surfaces the pre-reset terminal observation via `extras["final_obs"]` — the groundwork a `VectorEnv`-based implementation needs to report `final_obs` / `final_info` correctly under Same-Step autoreset.

## Validation
- `./isaaclab.sh -f` (pre-commit) on all touched files.
- `py_compile` on the touched env and wrapper modules.
- `git diff --check`.


kellyguo11:
let's add the flag to make this optional as we discussed

kellyguo11 (approved):

PR #6224, opened 2026-06-22
Add OvPhysxView: a string-keyed binding manager for OVPhysX tensor bindings (view series, part 1)

AntoineRichard:
> **📦 Part 1 of a 5-part series** adding an OvPhysX binding view and migrating the assets onto it:
> 1. **This PR** — add `OvPhysxView` (the API) + tests.
> 2. Migrate the OvPhysX **Articulation** to `root_view`.
> 3. Migrate the **RigidObject**.
> 4. Migrate the **RigidObjectCollection** (uses the fused `prim_paths`+`key_aliases` path).
> 5. Migrate the **sensors** (pose tracking; contact forces use a separate `ContactBinding` API and may not fully adopt the view).
>
> Parts 2–5 each depend only on this PR and can be reviewed independently.

---

# Description

Adds **`OvPhysxView`**, a string-keyed binding-management layer over the OVPhysX tensor bindings, in `isaaclab_ovphysx`.

**Why.** OVPhysX is the odd backend out: it exposes physics attributes as a loose `dict[TensorType, TensorBinding]` with **no view object**, whereas Newton has `selection.ArticulationView` and PhysX has typed tensor views. Today callers must hold the right `TensorType` enum member, manage a destination buffer, and create bindings against a USD glob — and the `create_tensor_binding` calls are scattered across the asset classes behind a private `_get_binding`. `OvPhysxView` gives OVPhysX a single, discoverable surface as pleasant as Newton's selection API, and one owner of binding creation/caching that the asset/data classes can delegate to. It is the narrow, independently-useful slice carved out of the (shelved) cross-backend `get_property`/`set_property` effort; the full rationale and decision log live in `docs/superpowers/specs/2026-06-17-ovphysx-view-design.md`.

**What it provides**
- String-keyed access by lowercased `TensorType` name (auto-derived from the wheel enum — no hand-maintained table); a `TensorType` member is also accepted.
- `get_attribute(name, out=)` — fresh-allocates a **typed** array (`wp.transformf` for poses, `wp.spatial_vectorf` for velocities, flat `float32` otherwise; via a maintained `_ATTR_DTYPE` map), no aliasing; **`read_into(name, dst)`** — zero-copy fill of a caller-owned, possibly structured-dtype buffer (`wp.transformf`, …) via a `float32` reinterpret view that is **cached per destination buffer** so the wheel's object-identity read cache stays warm even when callers hand a structured buffer each step; `set_attribute(name, values, indices=/mask=)` (structured sources reinterpreted); and a raw `binding_for(name)` accessor for adoption.
- The fused multi-prim form (`prim_paths=[...]` + `key_aliases`) so `RigidObjectCollection` can use it.
- Discoverability (`attribute_names`, `available_attributes`, `has_attribute`), metadata passthrough, and a nested error hierarchy (`UnknownAttribute` / `ReadOnlyAttribute` / `AttributeUnavailable` / `ShapeMismatch` / `DeviceMismatch`).

**Device policy:** no implicit CPU↔GPU conversion — CPU-resident property types are read/written on CPU, and a buffer on the wrong device raises `DeviceMismatch` rather than being staged.

**Scope / status:** this PR adds the view plus 48 mock-based unit tests (no live sim). The asset adoption that replaces `_get_binding`/`_binding_read`/`_binding_write` lives in the follow-up PRs (Articulation #6225, RigidObject #6226), which dogfood the view against a live sim on CPU and GPU and have fed fixes back here (e.g. `try_binding_for`, the read-view cache, typed `get_attribute`).

Fixes # — N/A (internal design tracking; surviving slice of the shelved cross-backend view effort).

## Type of change

## Screenshots

N/A — no user-facing visual change.

## Checklist

- [x] I have added tests that prove my feature works (`test/sim/test_ovphysx_view.py`, 48 tests)

### DIFF
--- a/source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
+++ b/source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
@@ -370,6 +370,20 @@ class OvPhysxView:
         """
         return self._binding(self._resolve(name))
 
+    def try_binding_for(self, name: str | Any) -> _BindingLike | None:
+        """Like :meth:`binding_for`, but return ``None`` instead of raising when the attribute
+        is valid yet **not available for this view's prims** (e.g. tendon types on a
+        tendon-less articulation, or a not-yet-created optional binding).
+
+        An invalid *name* still raises :class:`UnknownAttribute` -- that is a programming
+        error, not an availability question. Use this for the asset's ``binding or None``
+        pattern over optional bindings.
+        """
+        try:
+            return self._binding(self._resolve(name))
+        except OvPhysxView.AttributeUnavailable:
+            return None
+
     # -- discoverability -------------------------------------------------------
 
     @property
--- a/source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
+++ b/source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
@@ -39,7 +39,7 @@ from __future__ import annotations
 
 import logging
 import math
-from typing import Any, Protocol, runtime_checkable
+from typing import Any, Protocol
 
 import warp as wp
 
@@ -79,9 +79,13 @@ _READ_ONLY_NAMES: frozenset[str] = frozenset(
 _CPU_ONLY_NAMES: frozenset[str] = frozenset(tt.name.lower() for tt in _CPU_ONLY_TYPES)
 
 
-@runtime_checkable
 class _BindingLike(Protocol):
-    """Structural type of an ovphysx ``TensorBinding`` as used by this view."""
+    """Structural type of an ovphysx ``TensorBinding`` as used by this view.
+
+    ``read`` fills the passed array in place; ``write`` consumes ``indices``/``mask``
+    for partial writes. ``shape`` is the binding's flat tensor shape; ``count`` is the
+    number of matched prims.
+    """
 
     shape: tuple[int, ...]
     count: int
@@ -379,26 +383,32 @@ class OvPhysxView:
 
     @property
     def dof_count(self) -> int:
+        """Number of DOFs per articulation (articulation views only)."""
         return self._sample().dof_count
 
     @property
     def body_count(self) -> int:
+        """Number of bodies (links) per articulation (articulation views only)."""
         return self._sample().body_count
 
     @property
     def joint_count(self) -> int:
+        """Number of joints per articulation (articulation views only)."""
         return self._sample().joint_count
 
     @property
     def is_fixed_base(self) -> bool:
+        """Whether the articulation has a fixed base (articulation views only)."""
         return self._sample().is_fixed_base
 
     @property
     def fixed_tendon_count(self) -> int:
+        """Number of fixed tendons per articulation (articulation views only)."""
         return self._sample().fixed_tendon_count
 
     @property
     def spatial_tendon_count(self) -> int:
+        """Number of spatial tendons per articulation (articulation views only)."""
         return self._sample().spatial_tendon_count
 
     # -- internals -------------------------------------------------------------
@@ -407,7 +417,11 @@ class OvPhysxView:
         """Resolve a string name or a ``TensorType`` member to a ``TensorType``."""
         if isinstance(name, str):
             return resolve_tensor_type(name)
-        return name
+        if isinstance(name, TensorType):
+            return name
+        raise OvPhysxView.UnknownAttribute(
+            f"Attribute key must be a str name or a TensorType member, got {type(name).__name__}."
+        )
 
     def _binding(self, tensor_type: Any) -> Any:
         """Return the cached ``TensorBinding`` for ``tensor_type``, creating it on first use."""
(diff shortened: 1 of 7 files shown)


source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
```diff
@@ -0,0 +1,674 @@
...
+
+
+# -----------------------------------------------------------------------------
+# Pure helpers (no native simulation required; testable against ``ovphysx.types``)
+# -----------------------------------------------------------------------------
+
+
+def attribute_vocabulary() -> list[str]:
+    """Return every valid attribute name (sorted lowercased ``TensorType`` members)."""
+    return sorted(t.name.lower() for t in TensorType if t.name != "INVALID")
```
marcodiiga, line 157: API concern: this advertises every `TensorType` as an `OvPhysxView` attribute, but the implementation is still largely float32-oriented. The tensor surface includes non-float bindings such as `rigid_body_disable_simulation` (uint8/bool), so `attribute_names` / `has_attribute` can imply support for names that `get_attribute` will allocate/read with the wrong dtype. I think this should either filter to the currently supported subset, or wait on wheel-exposed dtype metadata before exposing the full `TensorType` vocabulary as supported attributes.

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
```diff
@@ -0,0 +1,674 @@
...
+# Pure-Python enum (no native dependency); safe to import regardless of USD state.
+TensorType = import_ovphysx("ovphysx.types").TensorType
+
+# Tensor types that cannot be written. The first group is read-only by PhysX
+# convention (accelerations, inverse mass/inertia, projected joint force); the
+# computed-dynamics group (jacobian, mass matrix, coriolis, gravity) is read-only
+# in practice. The wheel enum exposes no writability flag today, so this set is
+# hand-maintained.
+# TODO(ovphysx): source this from a wheel writability query (design doc §7 ask 3).
+_READ_ONLY_NAMES: frozenset[str] = frozenset(
```
marcodiiga, line 75: Can we keep this local read-only list clearly temporary? The wheel has at least three access modes in practice: read/write, read-only, and write-only control tensors. A local `_READ_ONLY_NAMES` table will drift as `TensorType` grows, and a future wheel API should probably expose an `access_mode` enum rather than a boolean `is_writable` flag.

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
```diff
@@ -0,0 +1,674 @@
...
+    Args:
+        physx: The OVPhysX ``PhysX`` instance exposing ``create_tensor_binding``.
+        pattern: An fnmatch glob selecting the prims this view addresses. Mutually
+            exclusive with ``prim_paths``.
+        device: Simulation device (e.g. ``"cuda:0"`` or ``"cpu"``). State bindings are
+            read/written on this device; CPU-only property bindings always use ``"cpu"``.
+        prim_paths: An explicit list of fnmatch globs for the fused multi-prim binding
+            form (``create_tensor_binding(prim_paths=[...])``). Mutually exclusive with
+            ``pattern``.
+        key_aliases: Optional mapping ``requested_type -> created_type`` so a binding can
```
marcodiiga, line 216: `key_aliases` makes sense for the collection/fused-binding adapter path, but it is a subtle semantic escape hatch: the requested key and the created binding type can differ. That is fine as an internal IsaacLab adapter, but I would avoid presenting it as the long-term public API without descriptor metadata that records the requested key, source tensor type, shape, native device, and access mode. Otherwise callers can reason from the visible key and get different runtime semantics.

source/isaaclab_ovphysx/test/sim/test_ovphysx_view.py
```diff
@@ -0,0 +1,582 @@
...
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Unit tests for the :class:`OvPhysxView` string-keyed binding manager.
+
+These exercise the pure-Python name<->enum logic and the view's get/set/read-into
+dispatch (including the float32 reinterpret of structured buffers and the
+no-implicit-conversion device policy) against a fake ``PhysX`` + fake ``TensorBinding``.
+Full read/write round-trips on a live sim are covered by the asset integration tests.
```
marcodiiga: This PR itself appears to add mock/fake-binding coverage only. That is good for the API mechanics, but before this becomes the hot-path `root_view` implementation I would like the follow-up/live coverage to explicitly exercise CPU-only properties on a GPU sim, write-only/read-only failures, and at least one structured `read_into` path against a real ovphysx binding. If that is intentionally covered by later asset-integration PRs, maybe make that expectation explicit here so the current test scope is not overstated.

marcodiiga (commented):
A few API-level comments. Overall this looks useful as an IsaacLab-side consolidation layer; my main concern is keeping the public contract narrow until dtype/access metadata comes from the wheel instead of local tables.

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
AntoineRichard: Good point. Note that in 0.4.13 all 55 `TensorType` members are float32 — `rigid_body_disable_simulation` isn't in the wheel yet — so nothing is mis-read today; this is forward-looking. Rather than guess a subset with another local table, I narrowed the *documented* contract instead: the module docstring + `attribute_names`/`has_attribute` now state they report name-validity, **not** dtype support, and that the view is float32-only pending wheel dtype metadata. Filed `TensorBinding.dtype` as the upstream ask (design §7 ask 2); once it lands we can filter `attribute_names` to a dtype-aware subset and drop the guesswork. Confirmed offline you're planning to expose this.

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
AntoineRichard: Agreed. Marked `_READ_ONLY_NAMES` explicitly **temporary** and reframed the upstream ask around an `access_mode` enum (read/write, read-only, write-only) rather than a boolean `is_writable`, so write-only control tensors stay distinguishable. It replaces the hand-maintained table (design §7 ask 3).

source/isaaclab_ovphysx/isaaclab_ovphysx/sim/views/ovphysx_view.py
AntoineRichard: Agreed — documented `key_aliases` as an **internal** IsaacLab adapter for the fused-collection binding path, explicitly not general public API (requested key vs created type intentionally differ), and noted that a public form should carry descriptor metadata (requested key, source tensor type, shape, native device, access mode), deferred to wheel metadata.

source/isaaclab_ovphysx/test/sim/test_ovphysx_view.py
AntoineRichard: Made the scope explicit in the test module docstring: these cover the view's API mechanics against mock bindings; live coverage — CPU-only properties on a GPU sim, read-only/write-only failures, and structured `read_into` against real bindings — is in the asset-integration tests (`test_articulation.py` / `test_rigid_object.py` / `test_rigid_object_collection.py`) that adopt this as `root_view`.

marcodiiga (approved):

kellyguo11 (approved):

AntoineRichard:
For the record (post-merge): the no-`out` `get_attribute` growth flagged here was fixed before merge — that path bypasses `_read_view` and calls `_as_binding_view` directly, with `test_get_attribute_no_out_does_not_grow_read_cache` asserting `_read_views` stays empty. The cross-attribute / eviction concerns later raised on the stacked PRs were verified as latent and unreachable by shipped code (replies on #6229 / #6231 / #6232): the wheel validates tensor shape natively before any copy and raises on mismatch, and every consumer uses one persistent buffer per attribute, so the cache stays correct and bounded.

PR #6229, opened 2026-06-22
Migrate OVPhysX PVA sensor onto OvPhysxView (view series, part 5)

AntoineRichard:
> Branched off #6224 as a sibling of the asset migrations (#6225/#6226/#6227). Targets `develop`; until #6224 merges, the diff below includes the Part 1 `OvPhysxView` commits. **Please review #6224 first.**

# Description

Part 5 of the OVPhysX view-migration series — route the **PVA** sensor's tensor-binding management through `OvPhysxView`. Internal refactor; **no public API change** (sensors do not expose a `root_view`).

* `_initialize_impl` builds one `OvPhysxView` instead of separate `create_tensor_binding` calls; binding handles come from `binding_for`.
* Pose/state reads route through `read_into` (cached `float32` reinterpret, typed off the binding shape), replacing the manual `wp.array(ptr=...)` reinterpret views.
* The CPU-only `RIGID_BODY_COM_POSE` keeps its pinned-host staging + `wp.copy` to the sim device.
* Numeric behavior unchanged.

## Type of change

- [x] This change requires no public API change (internal refactor)

## Checklist

- [x] I have added a changelog fragment (`.skip` — internal refactor, no user-facing change / no bump)

## Testing

`test_pva.py` — cpu: **13 passed**, cuda: **13 passed**.


AntoineRichard:
We dug into the read-cache "silent corruption" concern closely, including the wheel's native code, and it's a **latent** gap rather than a silent-corruption bug:

- The wheel's `read`/`write` validate the tensor shape (dtype + ndim + every dim) **before** any copy (`validateTensorShape` in `ovphysxTensorBinding.cpp`) and return a non-SUCCESS status that the Python wrapper raises as `RuntimeError` (api.py:472). A cross-attribute reuse therefore **raises**, it does not silently corrupt.
- It is unreachable by shipped code: each asset/sensor owns one persistent buffer per attribute (never reused across two `TensorType`s), and a binding's flat shape is constant for a session — it can only change on a full sim reset, which rebuilds the view from scratch.

So we deliberately did **not** add a per-hit shape guard (it would add hot-path cost to defend an unreachable, already-raising case); a dtype-aware cache key is tracked as a wheel-metadata follow-up (the wheel exposes no dtype today). We did make the reset boundary explicitly tidy — every consumer now clears `_root_view` in `_invalidate_initialize_callback`.

kellyguo11 (approved):

PR #6231, opened 2026-06-22
Migrate OVPhysX frame-transformer sensor onto OvPhysxView (view series, part 5)

AntoineRichard:
> Branched off #6224 as a sibling of the asset migrations (#6225/#6226/#6227). Targets `develop`; until #6224 merges, the diff below includes the Part 1 `OvPhysxView` commits. **Please review #6224 first.**

# Description

Part 5 of the OVPhysX view-migration series — route the **frame-transformer** sensor's tensor-binding management through `OvPhysxView`. Internal refactor; **no public API change** (sensors do not expose a `root_view`).

* `_initialize_impl` builds one `OvPhysxView` instead of separate `create_tensor_binding` calls; binding handles come from `binding_for`.
* Pose/state reads route through `read_into` (cached `float32` reinterpret, typed off the binding shape), replacing the manual `wp.array(ptr=...)` reinterpret views.
* All pose bindings are GPU-resident state; no CPU staging. One view per tracked body.
* Numeric behavior unchanged.

## Type of change

- [x] This change requires no public API change (internal refactor)

## Checklist

- [x] I have added a changelog fragment (`.skip` — internal refactor, no user-facing change / no bump)

## Testing

`test_frame_transformer.py` — cpu: **12 passed**, cuda: **12 passed**.


AntoineRichard:
Re the read-view cache's missing shape re-check (and the untested cross-attribute reuse): verified **latent, not a live bug**. The wheel validates tensor shape in native code before copying and raises `RuntimeError` on a mismatch (not silent), and no shipped caller reuses one buffer across two attributes — each buffer owns a single attribute, and a binding's flat shape is fixed for a session (it only changes on a full reset, which rebuilds the view). We opted not to add a per-hit guard (hot-path cost for an unreachable case); dtype-aware narrowing is a wheel-metadata follow-up. We did add reset-boundary tidiness (clearing the view on `_invalidate_initialize_callback`) across the consumers — frame_transformer already clears `_body_views`.

kellyguo11 (approved):

PR #6232, opened 2026-06-22
Migrate OVPhysX contact-sensor pose binding onto OvPhysxView (view series, part 5)

AntoineRichard:
> Branched off #6224 as a sibling of the other migrations. Targets `develop`; until #6224 merges the diff includes the Part 1 `OvPhysxView` commits. **Please review #6224 first.**

# Description

Part 5 of the OVPhysX view-migration series — the **contact sensor**. This is the *partial* one: only the `RIGID_BODY_POSE` pose binding (the `track_pose` path) is a `TensorBinding` the view can manage, so only it is migrated. The **`ContactBinding`** (net forces via `read_net_forces`, force matrix via `read_force_matrix`) is a separate ovphysx wheel API that `OvPhysxView` does not wrap — it is left exactly as-is.

* `_initialize_impl` builds an `OvPhysxView` for the pose binding (when `track_pose`) and obtains it via `binding_for(TT.RIGID_BODY_POSE)`; the count-mismatch check and the `pose_binding` accessor are unchanged.
* The per-step pose read goes through `read_into(TT.RIGID_BODY_POSE, ...)` (cached reinterpret).
* The `ContactBinding` create/read/destroy paths are untouched.
* Internal refactor; no public API change.

## Type of change

- [x] Internal refactor (non-breaking)

## Checklist

- [x] Changelog fragment (`.skip` — internal refactor, no bump)

## Testing

`test_contact_sensor.py` — cpu: **9 passed, 4 skipped**; cuda: **9 passed, 4 skipped** (exercises both the ContactBinding force path and the migrated pose-tracking path).


AntoineRichard:
Fixed — `_invalidate_initialize_callback` now clears `self._root_view` alongside `_pose_binding` (commit b8453bdcd9f), so the destroyed pose binding the view caches is no longer reachable after a sim stop; `_initialize_impl` rebuilds a fresh view on the next play. The separate read-cache concern doesn't apply here: `_poses_flat_buf` is flat `float32`, so it's passed straight through to the binding and never cached.

kellyguo11 (approved):

PR #6238, opened 2026-06-23
Gate Isaac RTX per-env scene partitioning behind environment variable

rilei-nvidia:
# Description

Add ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION environment variable to opt in to authoring primvars:omni:scenePartition and omni:scenePartition on the USD stage. Previously this was always on; now it is off by default.  This is to address an OOM in one of our IsaacSim RTX benchmark suites related to primvar inheritance of the scene partition attributes while the root cause/fix is under investigation.

## Testing

We have a separate issue which only visualizes env_0 in kit when scene partitioning is enabled...  so I tested with:
```bash
./isaaclab.sh train --rl_library rsl_rl \
    --task Isaac-Lift-KukaAllegro-Camera \
    presets=newton_mjwarp,isaacsim_rtx_renderer,rgb64,single_camera \
    --seed 42 \
    --num_envs 4 \
    --max_iterations 1000 \
    --visualizer kit
```
... and it shows all the envs - checked all the env Xforms and camera and they don't have the scenePartition attribute/primvars:
image

Conversely, tested with the environment variable enabled:
```bash
ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION=1 ./isaaclab.sh train --rl_library rsl_rl \
    --task Isaac-Lift-KukaAllegro-Camera \
    presets=newton_mjwarp,isaacsim_rtx_renderer,rgb64,single_camera \
    --seed 42 \
    --num_envs 4 \
    --max_iterations 1000 \
    --visualizer kit
```

And we get the old behavior of only env_0 showing.

@fatimaanes keen to hear if this helps mitigate the benchmark OOM you are seeing.

## Checklist

### DIFF
--- a/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
+++ b/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import json
 import logging
 import math
-import os
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, NoReturn
 
@@ -23,6 +22,7 @@ from pxr import Sdf, Usd, UsdGeom
 from isaaclab.app.settings_manager import get_settings_manager
 from isaaclab.renderers import BaseRenderer, RenderBufferKind, RenderBufferSpec
 from isaaclab.renderers.camera_render_spec import CameraRenderSpec
+from isaaclab.utils.renderers import isaac_rtx_per_env_scene_partition_enabled
 from isaaclab.utils.version import get_isaac_sim_version
 from isaaclab.utils.warp.kernels import reshape_tiled_image
 from isaaclab.utils.warp.warp_math import clamp_depth_to_inf_wp, replace_inf_depth_wp
@@ -52,16 +52,6 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
     if exc.name != "isaaclab_ppisp" and not (exc.name and exc.name.startswith("isaaclab_ppisp.")):
         raise exc
     raise ModuleNotFoundError(_PPISP_IMPORT_ERROR_MESSAGE, name="isaaclab_ppisp") from exc
-
-
-def isaac_rtx_per_env_scene_partition_enabled() -> bool:
-    """Return whether per-environment RTX scene partitioning is enabled.
-
-    Partitioning is opt-in: set
-    ``ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION=1`` to enable authoring of
-    ``primvars:omni:scenePartition`` and ``omni:scenePartition`` on the USD stage.
-    """
-    return os.environ.get("ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION", "0") == "1"
 
 
 # RTX simple-shading constants.
--- a/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
+++ b/source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
@@ -57,12 +57,11 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
 def isaac_rtx_per_env_scene_partition_enabled() -> bool:
     """Return whether per-environment RTX scene partitioning is enabled.
 
-    Partitioning is opt-in: the presence of the
-    ``ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION`` environment variable (regardless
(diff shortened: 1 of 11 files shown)


source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -182,12 +194,28 @@ def supported_output_types(self) -> dict[RenderBufferKind, RenderBufferSpec]:
     def prepare_stage(self, stage: Usd.Stage, num_envs: int) -> None:
         """Author per-env ``omni:scenePartition`` attributes for RTX cull-by-env rendering.
 
-        For each ``/World/envs/env_{i}`` root, writes the inheriting primvar
+        Authoring is only performed when the environment variable
```
huidongc: Maybe make a while list of valid values for the env var, like 1, ON, True etc?

I am a bit concerned about the semantic meaning of using any value, e.g. 0, False, empty string, to enable the code, would appear confusing. 

Maybe have a search whether IL has any similar env var toggle and whether there is any established convention.

rilei-nvidia: In IsaacLab there is only the usage of the DEBUG_TIMERS env var which defaults to "0" and is enabled with "1".

I would stray away from having a list of accepted values like "ON", "True", etc as it can introduce more ambiguity (casing, spelling mistake).  

Would be happy to change it to:
- "0" == off, and the default value if the env var is not set.
- "1" == on
which seems quite standard around the industry and easy to type and read

Thoughts?

huidongc: sounds good. this is a just a temporary toggle tmk.

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -513,7 +513,23 @@ def _apply_viewport_camera_scene_partition(self, usd_stage: Usd.Stage, num_envs:
...
+
+        logger.debug(
+            "[KitVisualizer] Per-environment Isaac RTX scene partitioning is enabled"
+            " (ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION)."
+            " Authoring omni:scenePartition attribute onto viewport camera '%s'.",
+            self._controlled_camera_path,
+        )
+
         if num_envs <= 0 or self._controlled_camera_path is None:
             return
```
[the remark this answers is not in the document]
rilei-nvidia: Good call

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
rilei-nvidia: Fixed!

huidongc (approved):

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -513,9 +513,26 @@ def _apply_viewport_camera_scene_partition(self, usd_stage: Usd.Stage, num_envs:
         ``omni:scenePartition`` token. Interactive viewport cameras live outside
         ``/World/envs`` and are created by Kit, so they do not inherit the env-root
         primvar authored by :class:`~isaaclab.scene.InteractiveScene`.
+
+        This method is a no-op unless ``ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION=1``,
+        matching the opt-in behaviour of
+        :meth:`~isaaclab_physx.renderers.IsaacRtxRenderer.prepare_stage`.
         """
+        from isaaclab_physx.renderers.isaac_rtx_renderer import isaac_rtx_per_env_scene_partition_enabled
```
fatimaanes: One concern here: this puts an `isaaclab_physx` import on the always-run visualizer initialization path (`_apply_viewport_camera_scene_partition()` is called unconditionally from `initialize()`), before we check whether the feature is enabled.

Since `isaaclab_visualizers` doesn't declare `isaaclab_physx` as a dependency (its `pyproject.toml` only lists `isaaclab`, which doesn't pull in `isaaclab_physx` either), this could break lighter/non-RTX installs with `ModuleNotFoundError`, even when the feature is disabled.

The existing `isaaclab_physx` imports in this file are behind camera-specific code paths (e.g. `_setup_camera_sensor_view()`), so they aren't exercised for a plain viewport. This import is different because it runs on every initialization.

Could we check the env var inline before importing from `isaaclab_physx`, or alternatively move the helper into a shared lower-level module?

rilei-nvidia: Great catch @fatimaanes, I will move the helper to isaaclab utils.

PR #6247, opened 2026-06-23
[IsaacLab CI] Add light weight post-merge integration tests

mataylor-nvidia:
# Description

All other tests are treated as a part of the full suite and do not get run on push only the following:

[rendering-correctness](https://github.com/isaac-sim/IsaacLab/actions/runs/27923944023/job/82630331274#logs) (~20m)
[rendering-correctness-kitless](https://github.com/isaac-sim/IsaacLab/actions/runs/27923944023/job/82630331226#logs) (~20m)

A subset of these rendering tests were added to the toml which are stable.

## Type of change

## Checklist


.github/workflows/build.yaml
```diff
@@ -67,7 +72,7 @@ permissions:
 
 env:
   NGC_API_KEY: ${{ secrets.NGC_API_KEY }}
-  CI_IMAGE_TAG: isaac-lab-ci:${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || github.ref_name }}-${{ github.sha }}
+  CI_IMAGE_TAG: isaac-lab-ci:${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || format('run-{0}', github.run_id) }}-${{ github.sha }}
```
mataylor-nvidia: Change was made because `github.ref_name` can contain characters Docker tags reject, especially / from branches like release/foo. The old non-PR tag could become invalid:
isaac-lab-ci:release/foo-<sha>

mataylor-nvidia: `github.run_id` is a unique numeric identifier assigned to a specific execution of a workflow.

mataylor-nvidia:
@greptileai

.github/workflows/build.yaml
nvsekkin: ~~this was put in place to let nightly build skip the tags if there were no changes to the branch history as seen here: https://github.com/isaac-sim/IsaacLab/actions/runs/27999168967/job/82867606478#step:7:136~~

~~this change may regress the publish behavior~~

EDIT: this is a separate publish step - I was wrong initially about this message.

nvsekkin: this is how the name is sanitized in the publish step: https://github.com/isaac-sim/IsaacLab/blob/develop/.github/workflows/publish-images.yaml#L130

nvsekkin: the tricky thing here is that the sanitization has to be a shell step, so we may want to drop the variable `CI_IMAGE_TAG` from this top level `env` and then need to push this section into a `run` block in the `config` job output

mataylor-nvidia: I wanted to try to get around it, but it is looking like I will need to do it this way

nvsekkin: we can add a new step here: https://github.com/isaac-sim/IsaacLab/pull/6247/changes#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R199

```
    - id: tag
      env:
        EVENT_NAME: ${{ github.event_name }}
        PR_NUMBER: ${{ github.event.pull_request.number }}
        REF_NAME: ${{ github.ref_name }}
        SHA: ${{ github.sha }}
      run: |
        set -euo pipefail
        if [ "$EVENT_NAME" = "pull_request" ]; then
          prefix="pr-${PR_NUMBER}"
        else
          prefix=$(printf '%s' "$REF_NAME" | sed 's/[^a-zA-Z0-9._-]/-/g')
        fi
        echo "ci_image_tag=isaac-lab-ci:${prefix}-${SHA}" >> "$GITHUB_OUTPUT"
```

add `ci_image_tag: ${{ steps.tag.outputs.ci_image_tag }}` to `outputs:`

then we would also need to change it everywhere: `env.CI_IMAGE_TAG` to `needs.config.outputs.ci_image_tag`

mataylor-nvidia: I have added something like this here:

https://github.com/isaac-sim/IsaacLab/pull/6247/changes#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R206

.github/workflows/build.yaml
```diff
@@ -46,12 +47,19 @@
...
+      - develop
+      - 'release/**'
   pull_request:
     types: [opened, synchronize, reopened]
     branches:
       - main
       - develop
       - 'release/**'
+  merge_group:
+    types: [checks_requested]
```
mataylor-nvidia: helps to fix situations where multiple different MRs were passing tests, but once put together onto a branch were breaking a test.

nvsekkin: this is good..
question: once the merge queue is added with this then we almost guarantee a green push-triggered pipeline - is this the correct assumption?

nvsekkin: my only concern here is that this feels like a separate, higher-level decision rather than part of a lightweight post-merge tests PR. 

Could we split the merge-queue adoption into its own PR? @huidongc @rilei-nvidia  what do you think?

rilei-nvidia: Agreed @nvsekkin - merge-queue adoption should be a separate discussion & PR

.github/workflows/build.yaml
```diff
@@ -667,7 +737,7 @@ jobs:
     name: "rendering-correctness"
     runs-on: [self-hosted, gpu]
     timeout-minutes: 120
-    continue-on-error: true
+    continue-on-error: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
```
nvsekkin, line 739: with the known flakiness this will show up as red. are we okay with this @mataylor-nvidia ?

.github/workflows/build.yaml
```diff
@@ -693,7 +763,7 @@ jobs:
     name: "rendering-correctness-kitless"
     runs-on: [self-hosted, gpu]
     timeout-minutes: 120
-    continue-on-error: true
+    continue-on-error: ${{ github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' }}
```
nvsekkin, line 767: same comment as https://github.com/isaac-sim/IsaacLab/pull/6247/changes#r3471125967

.github/workflows/build.yaml
mataylor-nvidia: Yes, I originally added as per Piotr's suggestion lets keep this PR short and remove this

.github/workflows/build.yaml
mataylor-nvidia: I will create a yaml which contains the individual list of tests which aren't flaky and only run these slowly adding more as others are fixed

.github/workflows/install-ci.yml
```diff
@@ -7,6 +7,8 @@ name: Installation Tests
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+  merge_group:
```
nvsekkin: looks like a leftover from dropping merge-queue scope; remove for consistency?

nvsekkin (approved):

PR #6255, opened 2026-06-24
OMPE-96696: Using ISAAC_LAB_SAVE_STAGES when running tests corrupts http:// asset URLs in saved USDs

mataylor-nvidia:
# Description

Preserve URI asset paths in save_stage

Avoid resolving asset paths with explicit URI schemes as filesystem paths
during stage save path remapping. This keeps remote assets such as
DomeLight texture URLs from being rewritten into invalid relative paths.

Fixes # ([OMPE-96696](https://jirasw.nvidia.com/browse/OMPE-96696))

## Type of change

## Screenshots

## Checklist


mataylor-nvidia:
@greptileai

pbarejko (approved):

source/isaaclab/isaaclab/sim/utils/stage.py
```diff
@@ -105,9 +116,11 @@ def resolve_paths(
     dst_dir = os.path.dirname(dst_layer.realPath or dst_layer.identifier)
 
     def _modify_path(asset_path: str) -> str:
-        if not asset_path:
+        if not asset_path or _is_uri_path(asset_path):
             return asset_path
         resolved = src_layer.ComputeAbsolutePath(asset_path)
+        if resolved and _is_uri_path(resolved):
```
rilei-nvidia, line 122: Does this second _is_uri_path need to be called?  If we determine that asset_path is not a URI, would ComputeAbsolutePath ever return a URI?

source/isaaclab/isaaclab/sim/utils/stage.py
```diff
@@ -59,6 +59,17 @@ def _check_ancestral_node(node) -> bool:
...
 
+def _is_uri_path(asset_path: str) -> bool:
+    """Return whether an asset path has an explicit URI scheme."""
+    scheme = asset_path.split("://", 1)[0]
+    return (
+        scheme != asset_path
+        and len(scheme) > 1
+        and scheme[0].isalpha()
+        and all(c.isalnum() or c in "+-." for c in scheme[1:])
+    )
```
rilei-nvidia, line 70: urllib.parse.urlparse is a utility from the python standard library that can aid with parsing the scheme from a URI.

source/isaaclab/isaaclab/sim/utils/stage.py
mataylor-nvidia: asset_path being non-URI only tells us about the authored value. src_layer.ComputeAbsolutePath(asset_path) anchors that value relative to the source layer. If the source layer lives at a URI-backed location, for example omniverse://server/project/scene.usd, then a relative authored path like:
textures/albedo.png
can resolve to something like:
omniverse://server/project/textures/albedo.png

The goal is to detect URI-backed resolved assets when this is the case

rilei-nvidia: I see - thanks for the explanation.  What could be helpful is adding this as code comments or a docstring of what _modify_paths is trying to achieve - it was not immediately clear with all the conditionals

source/isaaclab/isaaclab/sim/utils/stage.py
mataylor-nvidia: Yes I wanted to try something like this.

This helper is doing a stricter policy check: “treat this asset path as already URI-like only if it has an explicit :// scheme.”

That matters because urlparse will report schemes for things that are not remote/URI asset paths in this context, for example:

```
urlparse("C:/tmp/model.usd").scheme      # "c"
urlparse("anon:0x123:World.usd").scheme  # "anon"
```

The custom checks avoid those false positives by requiring:

```
scheme != asset_path   # the string contained ://
len(scheme) > 1        # avoid Windows-drive-ish C:// cases
scheme[0].isalpha()
all(c.isalnum() or c in "+-." for c in scheme[1:])
```

The reason I ended up not using it in this code is intentionally is because some paths may be URI-style so we should not relativize or resolve it as a filesystem path. Note: IsaacLab support for windows

rilei-nvidia: OK great to know you dug deeper into this.  Similar to the comment above I think it would be helpful to add a docstring with your findings so that in the future a developer (or AI) doesn't go and refactor this into using `urlparse` thinking that would handle these specific situations!

PR #6256, opened 2026-06-24
Add explicit FrameTransformer prim-path regex API

ooctipus:
## Summary
- add mutually exclusive `prim_path` / `prim_path_regex` fields for FrameTransformer source and target frames
- normalize explicit regexes as `(?:<regex>)`
- normalize deprecated `prim_path` values as `(?:<path>)(?:/.*)?`
- parse that normalized expression inside `resolve_matching_prims_from_source`
- keep PhysX and OV PhysX on one resolver call; Newton consumes the same final expression

## Matching semantics
- `prim_path_regex="/World/.../Robot/pelvis"`: filter only the directly matched pelvis prim
- `prim_path_regex="/World/.../Robot/.*"`: filter prims selected by that expression
- `prim_path="/World/.../Robot"` (deprecated): preserve recursive descendant lookup
- specifying both public fields, or neither, raises when the sensor is constructed

Existing unwrapped callers of `resolve_matching_prims_from_source` retain their previous recursive predicate behavior.

## Validation
- path selection guard: 4 passed
- normalized resolver tests: 7 passed
- PhysX FrameTransformer suite: 11 passed
- Newton FrameTransformer suite: 11 passed
- OV PhysX module reaches its expected capability-gated skip
- changed-file pre-commit: passed
- changelog fragment check: passed

### DIFF
--- a/docs/source/overview/core-concepts/sensors/frame_transformer.rst
+++ b/docs/source/overview/core-concepts/sensors/frame_transformer.rst
@@ -15,7 +15,12 @@ Frame Transformer
 
 One of the most common operations that needs to be performed within a physics simulation is the frame transformation: rewriting a vector or quaternion in the basis of an arbitrary euclidean coordinate system. There are many ways to accomplish this within Isaac and USD, but these methods can be cumbersome to implement within Isaac Lab's GPU based simulation and cloned environments. To mitigate this problem, we have designed the Frame Transformer Sensor, that tracks and calculate the relative frame transformations for rigid bodies of interest to the scene.
 
-The sensory is minimally defined by a source frame and a list of target frames.  These definitions take the form of a prim path (for the source) and list of regex capable prim paths the rigid bodies to be tracked (for the targets).
+The sensor is minimally defined by a source frame and a list of target frames.
+Use ``prim_path_regex`` for new configurations. It filters only the prims directly
+matched by the prim-path expression: ``/Robot/pelvis`` stops at the pelvis, while
+``/Robot/.*`` can select rigid bodies at the next path level below ``/Robot``.
+The deprecated ``prim_path`` field remains available for compatibility and
+recursively searches below each matched prim.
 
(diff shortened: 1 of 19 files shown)


ooctipus:
Reworked this in `bbdee4c022f` so the follow-up is a generic prim-path resolution fix. Relative to the original feature commit (`7001771bf2f`), the corrective delta now touches only:

- `source/isaaclab/isaaclab/sim/utils/queries.py`
- `source/isaaclab/test/sim/test_utils_queries.py`

There are no additional FrameTransformer, SensorBase, or backend changes in that delta. FrameTransformer keeps only the original compatibility shim: deprecated `prim_path` becomes a descendant regex, while `prim_path_regex` remains the authored full-path regex.

The resolver now full-matches the regex before applying the predicate, including clone-plan projections. Coverage includes authored regex quantifiers, actual environment-scope detection, non-contiguous clone IDs, nested clone ownership, heterogeneous predicate coverage, full-USD/source-only clone deduplication, and global paths.

Local verification: query suite `16 passed`; all-bodies integration passed on PhysX, Newton, and OVPhysX CPU; all staged pre-commit hooks passed.

ooctipus:
Correction to my previous update: e7de73b4a47 restores the source contract in resolve_matching_prims_from_source.

- Clone/source ownership is selected first.
- The full path regex filters paths projected from that selected source.
- The prim predicate runs only after the regex match.
- A clone-owned regex cannot fall through to scene-wide authored-stage traversal; partial-environment matches are rejected.
- Nearest nested clone ownership is preserved even when the nearer source lacks the requested prim.
- Ownership resolution is performed once per source-prim/template, not once per environment.

The corrective delta from the original API commit touches only generic prim-query code and tests; it adds no FrameTransformer, SensorBase, or backend-specific fix.

Verified with 23 generic query tests, the all-bodies FrameTransformer integration test on PhysX, Newton, and OVPhysX CPU, and the repository pre-commit hooks.

PR #6260, opened 2026-06-25
Optimize local-only wrench composition

ooctipus:
## Summary
- Track whether a WrenchComposer has ever received world-frame inputs since its last full reset.
- Fast-path local-only composition by copying local force/torque buffers into the body-frame outputs only when dirty.
- Preserve per-call pose refresh for world-frame positioned forces, which remain pose-dependent.

## Tests
- ./isaaclab.sh -p -m pytest source/isaaclab/test/utils/test_wrench_composer.py -k "local_only_composition_does_not_query_pose or global_positioned_composition_queries_pose_each_call or lazy_composition_tracks_dirty_flag or compose_is_idempotent" -q
- ./isaaclab.sh -p -m pytest source/isaaclab/test/utils/test_wrench_composer.py -q
- ./isaaclab.sh -p -m py_compile source/isaaclab/isaaclab/utils/wrench_composer.py source/isaaclab/test/utils/test_wrench_composer.py
- git diff --check

Ruff was not run because the local IsaacLab venv does not have the ruff module installed.

### DIFF
--- a/source/isaaclab/isaaclab/utils/wrench_composer.py
+++ b/source/isaaclab/isaaclab/utils/wrench_composer.py
@@ -66,6 +66,7 @@ class WrenchComposer:
         self._asset = asset
         self._active = False
         self._dirty = False
+        self._uses_global_frame = False
         if hasattr(self._asset.data, "body_com_pos_w"):
             self._get_com_pos_fn = lambda a=self._asset: a.data.body_com_pos_w.warp
         else:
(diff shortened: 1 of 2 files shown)


AntoineRichard (changes requested):
This is not the solution to the problem. THe problem is uncached access to a CPU property which is crazy expensive in PhysX.

PR #6269, opened 2026-06-25
OMPE-96097: Add kitless visual color randomization for Newton-Warp and OVRTX renderers

jmart-nv:
# Description

randomize_visual_color only worked on the Kit/Replicator (IsaacRTX) path, so the kitless renderers had no way to randomize visual color at reset. Add per-prim AND per-env color randomization that honors an env_ids subset across all three backends, dispatched by sniffing the scene's renderer types.

Each backend implements a uniform writer interface (constructed once from env + mesh_prim_path, then write_colors per reset):

- NewtonShapeColorWriter: writes rows of model.shape_color directly (the live-array, no-notify path the Newton-Warp render context binds at load). Matches env-agnostically so a single source env_0 match recolors every cloned env, and skips /collisions/ shapes.
- OVRTXVisualColorWriter: injects one OmniPBR material per target and rebinds each mesh via the renderer's write_attribute channel. Authoring is deferred to the first write_colors (post-bake, since the bake resets the runtime stage), and Renderer.reset() flushes the path-traced accumulator after every write.
- Replicator path retained unchanged for Kit/IsaacRTX.

Add a pre-PHYSICS_READY hook (EventManager.initialize_pre_physics_ready_terms) so writers can author USD before the OVRTX bake: OVRTX strips asset-bundled material:binding opinions there, otherwise the post-bake rebind is silently shadowed. Newton/Replicator are no-ops. The per-env target selection and shape validation shared by all backends live in isaaclab.utils.visual_color.

Tests: one render-readback contract test per renderer (color lands, per-prim/per-env distinct, env_ids subset honored) driven from a shared scene helper and a bundled-OmniPBR cube fixture, plus writer unit tests and a pre-physics-ready-terms test.

## Type of change

## Screenshots

color_dr_3backends

## Checklist


StafaH:
Resolved by #6299

PR #5976, opened 2026-06-04
Adding RigidBody USD data classes and writers 

vidurv-nvidia:
# Description

## Summary (PR 1 of 2 — additive, no behavior change)

Introduces single-namespace schema **fragments** so a prim can carry rigid-body properties from multiple USD namespaces at once (`physics:`, `physxRigidBody:`, `mjc:`), with core naming no backend. **Purely additive** — existing cfgs and the deprecated `RigidBodyPropertiesCfg` class are untouched.

- **core:** `SchemaFragment`, `RigidBodyFragment`, `UsdPhysicsRigidBodyCfg`; `apply_namespaced` (generic per-fragment writer) and `apply_rigid_body_properties` (applies a fragment list with `UsdPhysics.RigidBodyAPI` as the implicit anchor).
- **spawner:** `rigid_props` accepts a fragment list; legacy single cfgs keep working via a transition bridge (shapes/meshes/from_files/mesh_converter).
- **backends:** `PhysxRigidBodyCfg` (physx), `MujocoRigidBodyCfg` (newton).
- new `test_schema_fragments.py` + exports.

**PR 2** (stacked) flips `RigidBodyPropertiesCfg` to a deprecation factory and migrates all call sites: vidurv-nvidia/IsaacLab#4 (review after this lands).

## Test Plan
- [x] `test_schema_fragments.py` — 7 passed
- [x] pre-commit clean · `./isaaclab.sh -d` docs build clean
- [ ] CI

First slice of the staged physics schema-cfg refactor (collision / mass / joint-drive / … follow later). Draft for review of the fragment API shape.

## Type of change

- New feature (non-breaking change which adds functionality)

## Screenshots

N/A — no user-facing visual changes.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the pre-commit checks (`ruff` + `ruff-format`) on the changed files
- [ ] I have made corresponding changes to the documentation (the public API reference is auto-generated via `./isaaclab.sh -d`; not separately authored)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [ ] I have added my name to `CONTRIBUTORS.md`

PR #5276, opened 2026-04-14
Refactors the doc on Schema for Lab 3.0, and adds MuJoCo gravcomp

vidurv-nvidia:
# Description

Adds Newton-native and MuJoCo-specific schema cfg classes to `isaaclab_newton.sim.schemas`,
following the base/subclass framework from #5275. All new cfgs use the per-declaring-class
MRO routing in `_apply_namespaced_schemas` — no backend-specific branching in any writer.

Depends on #5275.

### MuJoCo (Newton MuJoCo kernel, `mjc:*` namespace)

| Class | Field | USD attribute | Applied schema |
|---|---|---|---|
| `MujocoRigidBodyPropertiesCfg` | `gravcomp` | `mjc:gravcomp` | None (raw attr) |
| `MujocoJointDrivePropertiesCfg` | `actuatorgravcomp` | `mjc:actuatorgravcomp` | `MjcJointAPI` |

Body-level `gravcomp` must be set for joint-level `actuatorgravcomp` to have any effect.
The spawner auto-enables `MujocoRigidBodyPropertiesCfg(gravcomp=1.0)` when joint-level
actuator gravcomp is requested without body-level gravcomp.

### Newton-native (`newton:*` namespace)

| Class | Fields | USD attributes | Applied schema |
|---|---|---|---|
| `NewtonCollisionPropertiesCfg` | `contact_margin`, `contact_gap` | `newton:contactMargin`, `newton:contactGap` | `NewtonCollisionAPI` |
| `NewtonMeshCollisionPropertiesCfg` | `max_hull_vertices` | `newton:maxHullVertices` | `NewtonMeshCollisionAPI` |
| `NewtonMaterialPropertiesCfg` | `torsional_friction`, `rolling_friction` | `newton:torsionalFriction`, `newton:rollingFriction` | `NewtonMaterialAPI` |
| `NewtonArticulationRootPropertiesCfg` | `self_collision_enabled` | `newton:selfCollisionEnabled` | `NewtonArticulationRootAPI` |

## Design constraints

Same single-cfg-per-spawner-slot rule as #5275. Newton cfgs subclass the same base classes
as PhysX cfgs; each declares `_usd_namespace`/`_usd_applied_schema` (ClassVar) and fields
that auto-camelCase to their USD attr names. Per-declaring-class MRO routing handles mixed
PhysX+Newton cfg hierarchies correctly.

## Field renames (with deprecation aliases through 5.0)

| Old | New | Reason |
|---|---|---|
[...]

PR #6272, opened 2026-06-26
Adding Articulation Root USD data classes and writers

vidurv-nvidia:
# Description

Adds the **articulation-root** schema-fragment API.

- `ArticulationRootFragment` marker (core) plus the PhysX / Newton articulation-root fragments — `PhysxArticulationCfg` (solver iteration counts, sleep / stabilization thresholds, self-collision toggles) and `NewtonArticulationCfg` (native self-collisions), each in its own namespace. There is no core neutral fragment: `UsdPhysics.ArticulationRootAPI` carries no attributes, so the writer applies it as a presence-gated anchor and each backend fragment contributes its own namespaced schema.
- `apply_articulation_root_properties` family writer — resolves the existing `UsdPhysics.ArticulationRootAPI` root within the prim subtree and tunes *that* prim; only defines a fresh root on the target prim when the subtree has none. It never duplicates the root (exactly-one-root invariant). The `ArticulationRootAPI` anchor is applied by the writer, not by any fragment.
- The non-USD `fix_root_link` flag is a **spawner-cfg field** (`UsdFileCfg.fix_root_link`), passed to the writer as a keyword argument — fragments carry USD attributes only. This is additive; the legacy `articulation_props.fix_root_link` path is untouched.
- Fixing an articulation base is a capability of the active physics backend: `PhysicsManager.fix_articulation_root(prim, stage)`, resolved from `cfg.physics.class_type`, authors the world-to-root fixed joint and returns the resulting root prim. The base implementation authors a backend-neutral fixed joint and returns the same prim; PhysX overrides it to relocate the articulation root to the parent (parser workaround) and returns the parent. The writer fixes the root **before** writing fragments, so every fragment lands on the single resulting root regardless of backend — a composed PhysX + Newton fragment list with `fix_root_link=True` leaves exactly one root. Dispatch is resolved through the normal method-resolution order, so cfg subclasses and every backend inherit the correct behaviour without any per-type registration or import-order dependency.
- The spawner `articulation_props` slot now also accepts a fragment list (transition bridge routes legacy single cfgs to the existing `define_`/`modify_` writers). The slot is routed by type even when the list is empty, and `fix_root_link` is honored independently of whether schema properties were supplied.

Purely **additive** and self-contained: builds only on the schema-fragment base already in `develop`, existing call sites untouched, depends on no other open PR. Local `test_articulation_fragments.py`: 20/20 passing (includes PhysX + Newton root-fixing single-root, manager-subclass capability inheritance, empty/topology-only spawn, and end-to-end `_spawn_from_usd_file` regressions).

## Type of change

## Screenshots

N/A — non-visual API change.

## Checklist


source/isaaclab/isaaclab/sim/schemas/_backend_hooks.py
```diff
@@ -16,6 +16,44 @@
...
+
+def _resolve_fixed_root_joint_creator(physics_cfg: type | None) -> Callable | None:
+    """Return the creator registered for the active backend's physics-cfg type, or None.
+
+    Args:
+        physics_cfg: The type of the active simulation's ``cfg.physics`` (or None if no simulation).
+    """
+    if physics_cfg is None:
+        return None
+    return _FIXED_ROOT_JOINT_CREATORS.get(physics_cfg)
```
ooctipus: Blocking: this exact-type lookup is not complete for the supported backend/config surface. `OvPhysxCfg` is a sibling of `PhysxCfg`, and nothing in `isaaclab_ovphysx` registers a creator, so fragment mode with `fix_root_link=True` reaches the RuntimeError. It also misses the in-tree `DeformableNewtonCfg(NewtonCfg)` and any user subclass. Hook availability additionally depends on importing the matching `isaaclab_*.sim.schemas` module, rather than on activating that backend, so a Newton run using only the PhysX fragment for `articulation_enabled` does not register the Newton creator.

The inversion-of-control direction is consistent with #6273, but the cfg class is a brittle dispatch key. `SimulationContext.physics_manager` already identifies the active implementation and supports inherited backend capabilities; a classmethod/capability there would avoid exact-type and import-order failures. If the registry remains, please resolve through the MRO, register OVPhysX explicitly, make registration occur with backend activation, and test each supported manager plus a cfg subclass.

source/isaaclab_physx/isaaclab_physx/sim/schemas/__init__.py
```diff
@@ -5,12 +5,92 @@
...
+        aname = attr.GetName()
+        if aname.startswith("physxArticulation:"):
+            parent_attr = parent_prim.GetAttribute(aname)
+            if not parent_attr:
+                parent_attr = parent_prim.CreateAttribute(aname, attr.GetTypeName())
+            parent_attr.Set(attr.Get())
+
+    # remove the api from the (former) root
+    articulation_prim.RemoveAppliedSchema("PhysxArticulationAPI")
+    articulation_prim.RemoveAPI(UsdPhysics.ArticulationRootAPI)
```
ooctipus: Blocking property-completeness bug: this relocation moves only the USD/PhysX pieces. For the documented composition `[PhysxArticulationCfg(...), NewtonArticulationCfg(...)]` with `fix_root_link=True`, `NewtonArticulationRootAPI` and `newton:*` remain on the old child. Because `NewtonArticulationRootAPI` itself includes `PhysicsArticulationRootAPI`, removing the directly applied USD API here does not remove the composed root API: the parent and child both remain articulation roots.

That violates the exactly-one-root invariant and leaves `newton:selfCollisionEnabled` on the wrong root. The current composition test misses it because it never combines fragments with root fixing. Please make the backend operation return/identify the resulting root and let the family writer preserve or reapply every supplied articulation fragment, then add a PhysX + Newton + `fix_root_link=True` test asserting one root and the Newton value on that root. This is also important for URDF/MJCF assets that already carry the Newton API.

source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
```diff
@@ -369,7 +369,16 @@ def _spawn_from_usd_file(
 
     # modify articulation root properties
     if cfg.articulation_props is not None:
-        schemas.modify_articulation_root_properties(prim_path, cfg.articulation_props)
+        # transition shim, remove later: new fragment list -> apply_*; legacy single cfg -> modify_*
+        articulation_frags = (
+            cfg.articulation_props if isinstance(cfg.articulation_props, (list, tuple)) else [cfg.articulation_props]
+        )
+        if articulation_frags and all(isinstance(f, schemas.SchemaFragment) for f in articulation_frags):
```
ooctipus: The field has moved out of the USD fragment, which is the right ownership, but the behavior is still coupled to a non-empty property list here. `articulation_props=[]` fails this guard and is sent to the legacy writer as a cfg, where `dataclasses.fields([])` raises `TypeError`. With `articulation_props=None`, the spawner-level `fix_root_link` is silently ignored. That means a caller needs a dummy backend fragment just to request a topology-only operation; there is no core articulation fragment to use instead. The direct writer tests at lines 268 and 298 already treat `[]` as valid, so the two entry points disagree.

Please route fragment collections by type even when empty, and preferably process the topology flag independently from whether schema properties were supplied. An end-to-end `_spawn_from_usd_file` regression test is important here; unlike #5976 and #6254, all new tests bypass this transition bridge.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -275,6 +276,158 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+    for cfg in fragments:
+        func = cfg.func if callable(cfg.func) else string_to_callable(cfg.func)
+        func(cfg, root_path, stage)
+
+    # fix root link based on input
+    # we do the fixed joint processing later to not interfere with setting other properties.
+    # this logic is reproduced from the legacy ``modify_articulation_root_properties`` writer.
+    if fix_root_link is not None:
+        # check if a global fixed joint exists under the resolved root prim
+        existing_fixed_joint_prim = find_global_fixed_joint_prim(root_path)
```
ooctipus: Please pass `stage=stage` here. The root and fragment writes above honor the explicit stage argument, but this lookup falls back to the current global stage. When the supplied stage is not current, this can either raise because `root_path` is absent there or toggle a same-path joint on the wrong stage. The current tests all make the supplied stage current, so they do not cover the advertised explicit-stage behavior.

ooctipus (changes requested):
I reviewed this against the runtime schemas and the same author history used for the earlier fragment reviews:

- [#5275](https://github.com/isaac-sim/IsaacLab/pull/5275) established the core/base versus backend-package split. This PR follows that ownership correctly.
- [#5276](https://github.com/isaac-sim/IsaacLab/pull/5276) established the explicit schema-property audit standard. Against Kit 110.1.11, the inventory here is complete: OpenUSD ArticulationRootAPI has no attributes; `PhysxArticulationCfg` has all six PhysxArticulationAPI attributes; `NewtonArticulationCfg` has the sole Newton articulation attribute.
- [#5976](https://github.com/isaac-sim/IsaacLab/pull/5976) established fragment composition and tested the real spawner transition. The PhysX relocation here breaks that composition, and this PR has no end-to-end spawner test.
- [#6273](https://github.com/isaac-sim/IsaacLab/pull/6273) is the closest precedent for moving a non-USD behavior flag to `UsdFileCfg` and using backend hooks. Moving `fix_root_link` out of schema fragments is correct; the remaining plumbing is not yet clean because it is still conditional on a non-empty property list and dispatches through an exact cfg type/import side effect.

I am requesting changes for two blocking correctness/completeness issues:

1. PhysX root relocation leaves `NewtonArticulationRootAPI` on the old child. Since that schema includes `PhysicsArticulationRootAPI`, a composed PhysX + Newton asset ends with two articulation roots after `fix_root_link=True`, with `newton:selfCollisionEnabled` on the wrong root.
2. Backend dispatch is incomplete: OVPhysX has no registration, cfg subclasses miss the exact lookup, and the matching creator may be absent solely because its schema subpackage was not imported.

The empty-list/spawner behavior and explicit-stage bug are also called out inline. I would add a real `UsdFileCfg` spawn test covering the fragment bridge, each supported physics manager, a cfg subclass, empty fragments/topology-only use, and combined PhysX + Newton fragments with root fixing.

The cleanest design in the current architecture is to make fixed-root authoring a capability of the active `SimulationContext.physics_manager`, and have that operation return the resulting articulation-root prim. Core can then retain ownership of fragment application/migration across every namespace, while each backend owns only its topology/parser difference. This removes cfg-type and registration-order coupling. A smaller acceptable fix is MRO-aware lookup, activation-time registration for every backend including OVPhysX, and a hook contract that lets core preserve all composed fragments.

Non-blocking cleanup: the PR description still promises a core `UsdPhysicsArticulationRootCfg` that is not present, and the new public symbols are absent from the API RST. The two failing CI jobs appear unrelated: OVPhysX failed while downloading a wheel due to a SHA mismatch, and rendering timed out in the Shadow Hand test.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -275,6 +276,158 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+            existing_fixed_joint_prim.GetJointEnabledAttr().Set(fix_root_link)
+        elif fix_root_link:
+            logger.info(f"Creating a fixed joint for the articulation: '{root_path}'.")
+            # Creating a world<->root fixed joint to fix the base is backend-specific. Look up the
+            # creator registered for the active simulation's physics-cfg type so core carries no backend
+            # logic and never touches a non-active backend; see register_fixed_root_joint_creator.
+            from isaaclab.sim import SimulationContext
+
+            sim = SimulationContext.instance()
+            creator = _resolve_fixed_root_joint_creator(type(sim.cfg.physics) if sim is not None else None)
```
ooctipus: there is one usage of this, maybe remove the shim helper

source/isaaclab/isaaclab/sim/schemas/_backend_hooks.py
ooctipus: To make the alternative concrete: I do not think we need another factory here, and I would avoid putting this on `physics_cfg`. The cfg should remain declarative data; `SimulationContext` has already used `cfg.physics.class_type` to select the active runtime implementation as `sim.physics_manager`.

The call could therefore be:

```python
sim = SimulationContext.instance()
final_root = sim.physics_manager.fix_articulation_root(articulation_prim, stage)
```

with a classmethod capability on `PhysicsManager`, implemented by each backend. Newton would author the fixed joint and return the same prim; PhysX would author the joint, perform its parser-specific relocation, and return the parent/new root; OVPhysX would provide its own implementation. Returning the final root is important so the family writer can apply every fragment to the actual resulting root.

```python
final_root = sim.physics_manager.fix_articulation_root(articulation_prim, stage)
for fragment in fragments:
    apply_fragment(fragment, final_root, stage)
```

The relocation path must also migrate/remove pre-existing articulation-root family schemas so the former child cannot remain a second root. This keeps runtime USD mutation on the manager, keeps property ownership in the family writer, inherits naturally for manager subclasses, and removes the parallel cfg-type registry plus its import-order dependency.

vidurv-nvidia: Implemented as suggested in b0dcdd86ca4. Fixing the base is now a capability on the manager:

```python
final_root = sim.physics_manager.fix_articulation_root(articulation_prim, stage)
for cfg in fragments:
    func(cfg, final_root.GetPath().pathString, stage)
```

`PhysicsManager.fix_articulation_root` (base) authors the backend-neutral fixed joint and returns the same prim; `PhysxManager` overrides it to relocate the root to the parent and return the parent. The cfg stays declarative — dispatch is resolved from `cfg.physics.class_type`, so it inherits through the MRO with no parallel registry. The former child is no longer left as a second root because fragments are applied to the returned root (see the property-completeness thread).

One deviation: the enable/disable of an *existing* fixed joint stays in the writer (a backend-neutral `JointEnabledAttr` toggle, no relocation); only the create/relocate path is delegated to the manager.

vidurv-nvidia: Fixed in b0dcdd86ca4. The exact-type cfg registry is gone, replaced by `PhysicsManager.fix_articulation_root` resolved through `cfg.physics.class_type`. Because dispatch is by manager class and inheritance is by MRO, a subclassed cfg like `DeformableNewtonCfg(NewtonCfg)` resolves `NewtonManager` and inherits the base capability, `OvPhysxManager` inherits the base default instead of hitting a RuntimeError, and there is no import-order dependency (the active manager class always exists). Added `test_fix_articulation_root_capability_inherited_by_manager_subclass` covering a subclass inheriting an overriding backend, and a base-default test for the Newton/OVPhysx path.

source/isaaclab_physx/isaaclab_physx/sim/schemas/__init__.py
vidurv-nvidia: Fixed in b0dcdd86ca4. Root fixing now happens **before** any fragment is written, and each fragment is applied to the root returned by `fix_articulation_root`. So after PhysX relocates the root to the parent, both the PhysX and the Newton fragments land on that parent — nothing is stranded on the former child, and there is exactly one root. The PhysX override no longer copies attributes at all (there are none to copy at relocation time). Added `test_physx_and_newton_fragments_fix_root_link_keeps_single_root` asserting one root with the Newton value on it; verified it fails on the old ordering (`2 == 1`, the composed `NewtonArticulationRootAPI` kept the child a root).

source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
vidurv-nvidia: Fixed in b0dcdd86ca4. `articulation_props` is now routed by type: a list (including `[]`) goes to the fragment writer, only a legacy single cfg goes to `modify_*`, so an empty list no longer reaches `dataclasses.fields([])`. `fix_root_link` is read as a spawner-level flag independently of whether properties were supplied. Added end-to-end `_spawn_from_usd_file` regressions for both `articulation_props=[]` + `fix_root_link=True` and `articulation_props=None` + `fix_root_link=True`.

source/isaaclab/isaaclab/sim/schemas/schemas.py
vidurv-nvidia: Fixed in b0dcdd86ca4 — `find_global_fixed_joint_prim(root_path, stage=stage)`. Added `test_apply_articulation_root_properties_honors_explicit_stage`, which authors on a non-current in-memory stage and asserts nothing leaks onto the current stage.

source/isaaclab/isaaclab/sim/schemas/schemas.py
vidurv-nvidia: Removed in b0dcdd86ca4 — the shim (`_resolve_fixed_root_joint_creator`) and the whole creator registry are gone with the move to the `PhysicsManager.fix_articulation_root` capability.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -267,6 +268,117 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+            :attr:`~isaaclab.sim.spawners.UsdFileCfg.fix_root_link` for the semantics. Defaults
+            to None, in which case the root link is not modified.
+
+    Returns:
+        True if the properties were successfully set.
+
+    Raises:
+        ValueError: When the prim path is not valid.
+        NotImplementedError: When the root prim is not a rigid body and a fixed joint is to be created.
+    """
```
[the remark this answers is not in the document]
vidurv-nvidia: Addressed in b0dcdd86ca4. The per-backend "no creator registered" case is gone with the registry; the remaining `RuntimeError` is raised only when a fixed joint must be created but there is no active `SimulationContext` to resolve the backend from, and the `Raises` section now documents exactly that plus the propagated `NotImplementedError`.

vidurv-nvidia:
Addressed the two non-blocking notes from the review summary:

- **Description mismatch:** corrected. There is no core `UsdPhysicsArticulationRootCfg` — `UsdPhysics.ArticulationRootAPI` has no attributes, so core contributes the `ArticulationRootFragment` marker + the writer (which applies the anchor) and the concrete fragments are `PhysxArticulationCfg` / `NewtonArticulationCfg`. The PR description now reflects this and the new `PhysicsManager.fix_articulation_root` capability design.
- **API RST:** the fragment API is currently absent from the RST framework-wide (the base `SchemaFragment`, every `apply_*` writer, and the sibling `Physx*`/`Newton*` fragment classes already merged are all undocumented; the hand-curated schema pages list only the multi-namespace `*PropertiesCfg` names). Rather than half-document the module here, the full fragment-API RST sweep will land in the docs PR #6286. Happy to pull it into this PR instead if you would prefer it co-located.

The two failing CI jobs (OVPhysX wheel SHA mismatch, Shadow Hand render timeout) are unrelated to this change, as you noted.

source/isaaclab_physx/isaaclab_physx/sim/schemas/__init__.py
vidurv-nvidia: Follow-up in 3b9cf1cebf0: the write-fragments-after-fix reorder covered composed *fragments* but not the URDF/MJCF case you mentioned — an asset that ships with `NewtonArticulationRootAPI` already authored on the root link would still have kept a second root through schema composition, and pre-authored `physxArticulation:*` values were dropped instead of moving with the root as the legacy writer did. The PhysX relocation now migrates any directly authored API schema that the USD schema registry reports as composing `PhysicsArticulationRootAPI` (backend-agnostic, no backend names) plus the `PhysxArticulationAPI` companion, together with their authored attributes. Two regression tests added (`test_physx_fix_root_link_migrates_preauthored_newton_root_api`, `..._preauthored_physx_attrs`); both verified to fail without the migration (the Newton one with the same 2-roots signature).

source/isaaclab/isaaclab/physics/physics_manager.py
```diff
@@ -99,6 +99,64 @@ def provides_implicit_damping(cls) -> bool:
...
+        Raises:
+            NotImplementedError: When the root prim is not a rigid body and the first rigid-body link
+                cannot be determined to anchor the fixed joint.
+        """
+
+        from isaaclab.sim.utils import create_fixed_root_joint  # noqa: PLC0415
+
+        cls._require_rigid_body_root(articulation_prim)
+        create_fixed_root_joint(articulation_prim, stage)
+        return articulation_prim
```
ooctipus, line 152: Blocking: this base implementation is not sufficient for OVPhysX. `OvPhysxManager` inherits it, but OVPhysX uses the PhysX parser and does not recognize the resulting topology as fixed-base when `ArticulationRootAPI` remains on the rigid link. I reproduced this with the official `ovphysx==0.4.13` parser: root API on the parent gives `is_fixed_base=True`, while the topology produced here gives `is_fixed_base=False`.

The new “Newton/OVPhysX” test only invokes `PhysicsManager` directly and checks authored USD; it never runs the OVPhysX parser. Please give OVPhysX the PhysX-style relocation/migration behavior (a shared pure-USD helper would avoid duplication) and add a real backend test asserting `is_fixed_base`.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -275,6 +276,131 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+            logger.warning(
+                f"No articulation root found under '{prim_path}': ignoring fix_root_link={fix_root_link}."
+                " The flag only applies to articulations."
+            )
+        else:
+            root_path = articulation_prim.GetPath().pathString
+            # honor an explicit stage so the lookup does not fall back to the global current stage
+            existing_fixed_joint_prim = find_global_fixed_joint_prim(root_path, stage=stage)
+            if existing_fixed_joint_prim is not None:
+                # a joint already exists: just enable/disable it in place (no relocation)
```
ooctipus: Blocking: an existing joint does not make relocation unnecessary. If a pre-authored fixed joint exists while `ArticulationRootAPI` is still on the rigid link, this branch enables the joint but bypasses `physics_manager.fix_articulation_root`; PhysX/OVPhysX therefore retain the same floating/maximal-coordinate topology.

The backend capability should own the complete “ensure fixed” operation, including normalizing root placement when the joint already exists (without creating a duplicate). Please cover at least disabled-existing-joint -> `fix_root_link=True` with the real parser; the current test only covers True -> False.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -275,6 +276,131 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+            raise TypeError(
+                "apply_articulation_root_properties: expected ArticulationRootFragment instances in"
+                f" 'fragments', got '{type(frag).__name__}'. Pass a legacy single cfg to"
+                " modify_articulation_root_properties instead."
+            )
+    if stage is None:
+        stage = get_current_stage()
+    # resolve the existing root (it may live on a child prim in USD assets); instance proxies can't be
+    # authored on, so don't traverse them
+    articulation_prim = get_first_matching_child_prim(
```
ooctipus: Blocking: resolving only the first matching child silently skips sibling articulations. I reproduced `/World/A` and `/World/B`, both carrying `ArticulationRootAPI`: the fragment was authored on A and absent from B. The legacy `@apply_nested` writer processes every non-nested sibling root, and multi-articulation USD files are valid inputs.

Please resolve and process all top-level/non-nested articulation roots, creating an anchor on `prim_path` only when no root exists, and aggregate the per-root result.

source/isaaclab_physx/isaaclab_physx/physics/physx_manager.py
```diff
@@ -303,6 +303,93 @@ def initialize(cls, sim_context: SimulationContext) -> None:
...
+                    a.GetName() for a in old_root.GetAttributes() if prefix and a.GetName().startswith(prefix)
+                ]
+            for prop_name in prop_names:
+                attr = old_root.GetAttribute(prop_name)
+                if not attr or not attr.HasAuthoredValue():
+                    continue
+                new_attr = new_root.GetAttribute(prop_name)
+                if not new_attr:
+                    new_attr = new_root.CreateAttribute(prop_name, attr.GetTypeName())
+                new_attr.Set(attr.Get())
```
ooctipus: Blocking: this does not preserve valid authored USD data. The source schema is removed before this read, and `attr.Get()` reads only the default time. I reproduced `physxArticulation:sleepThreshold` authored only at times 1 and 2: relocation fails with `Type mismatch ... expected 'float', got 'void'`. If a default is present, the operation succeeds but silently drops the time samples.

Please capture/copy before removing the schema and preserve the authored default, all time samples, metadata, and connections; property/spec-level copying is safer than reducing each attribute to one `Get()`/`Set()`.

source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
```diff
@@ -368,8 +368,36 @@ def _spawn_from_usd_file(
...
+        and not isinstance(articulation_props, (list, tuple))
+        and not isinstance(articulation_props, schemas.SchemaFragment)
+    ):
+        if articulation_fix_root_link is not None:
+            logger.warning(
+                f"Ignoring the spawner-level 'fix_root_link={articulation_fix_root_link}' because"
+                " 'articulation_props' is a legacy cfg, which owns its own 'fix_root_link' field. Set"
+                " it on that cfg instead."
+            )
+        schemas.modify_articulation_root_properties(prim_path, articulation_props)
```
ooctipus, line 391: Blocking for the advertised legacy transition: `NewtonArticulationRootPropertiesCfg(articulation_enabled=False, self_collision_enabled=True)` routed here raises `ValueError` instead of authoring both fields. The root cause is `NewtonArticulationRootPropertiesCfg._usd_field_exceptions = {}`, which shadows the inherited mapping required for `articulation_enabled`.

This breaks the base-field-through-backend-subclass invariant established in the same author's merged #5275. Please inherit/merge the base exception mapping and add a regression comparing this legacy cfg with the equivalent PhysX + Newton fragment list.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -275,6 +276,131 @@ def apply_namespaced(cfg: schemas_cfg.SchemaFragment, prim_path: str, stage: Usd
...
+                        " base."
+                    )
+                final_root = sim.physics_manager.fix_articulation_root(articulation_prim, stage)
+
+    # dispatch each fragment to the resolved (possibly relocated) root via its own applier
+    if fragments:
+        final_root_path = final_root.GetPath().pathString
+        for cfg in fragments:
+            func = cfg.func if callable(cfg.func) else string_to_callable(cfg.func)
+            func(cfg, final_root_path, stage)
```
ooctipus: The fragment result is discarded, so the family writer returns `True` even when an applier returns `False`; I reproduced that with a failing stub `cfg.func`. This contradicts the documented return contract and the aggregate-result semantics established by the same author's merged #5976 follow-up (`2a8b8e7`). Please accumulate `bool(func(...))` across fragments/roots and add the same failing-applier regression used by the other family writers.

ooctipus (changes requested):
I re-reviewed the latest head against the same author's merged schema-fragment precedents. The field split itself is complete (PhysX 6/6, Newton 1/1, and UsdPhysics.ArticulationRootAPI has no configurable attributes), and manager-based dispatch is the right separation boundary. I am requesting changes for the independently reproduced correctness gaps below. All 24 targeted tests pass locally, which means these cases need explicit regression coverage.

ooctipus (approved):
Re-reviewed head d55e0ed8170 against the same property-inventory, backend-ownership, and fragment-composition precedents used in the earlier review. Approved: the fragment inventory remains complete (PhysX 6/6, Newton 1/1), and the direct fixes now cover every top-level root, instance proxies, return-value propagation, existing disabled joints, PhysX/OVPhysX parent-root normalization, lossless authored USD property migration, Newton legacy routing, and explicit-stage isolation. The joint authoring helper is private and the shared relocation boundary carries no backend imports. Local verification: pre-commit and changelog gates passed; 229 focused/neighboring tests passed. The OVPhysX result was also parsed directly as one fixed-base articulation; full local OV reset remains blocked by the pre-existing pinned-wheel set_cpu_mode mismatch, unrelated to this PR.

vidurv-nvidia:
CI triage for the failures on 280c20a02e5 (`isaaclab_tasks [2/3]`, `isaaclab_rl`):

Both are the config-import purity tests (`test_env_cfg_no_forbidden_imports`, `test_template_generator`). Root cause: the module-level `from pxr import ...` that came in with the import hoisting — `manager_base` imports `PhysicsManager` while environment configs load, so the hoist makes every registered env cfg pull `pxr` before the app starts, which those tests forbid. Reproduced locally on this head (Isaac-Ant fails with the `pxr` violation) and fixed in f6fd754ea74 by re-deferring only the `pxr` imports into the two methods that use them, mirroring the local-import pattern the last commit already used for the `SimulationContext` cycle.

The module-level `from isaaclab.sim.utils.stage import get_current_stage` is fine to keep now: I merged latest `develop`, and #6275 defers the Isaac Sim stage-context integration, so that path no longer reaches `isaacsim`/`carb` at import time (it was the second half of the CI failure — the July 4 run predated #6275). After the merge + re-deferral: `test_env_cfg_no_forbidden_imports` 96/96 and `test_articulation_fragments` 26/26 locally.

`test-curobo` and `rendering-correctness` look unrelated — the rendering failure is in the golden-image comparison path (`rendering_test_utils.py:638`, flaky-retry involved) and develop just churned goldens; worth a re-run now that develop is merged in.

vidurv-nvidia:
Install-CI note: the `Installation Tests (x86)` failure on 10ecaf909f4 is develop-wide, not from this branch. Both of develop's own Installation Tests runs today fail with the identical signature — the Cartpole camera smoke hits `AssertionError: render width must be a multiple of tile_width` inside `newton/_src/sensors/warp_raytrace/render_context.py:277`. It appeared with the Newton pin bump in #6366 (`newton==1.4.0.dev0 @ c7ae7c76`), whose raytrace context now asserts tile alignment that the camera env render width does not satisfy. This branch touches nothing in the renderer/camera stack, and the same job was green on the previous head before today's develop merge.

PR #6279, opened 2026-06-27
Remove RenderCfg from SimulationCfg and rely on new Global Settings

pbarejko:
# Description

When refactoring rendering back-ends we forgot `SimulationCfg` holds `RenderCfg` that carries global settings for Isaac Sim RTX. To follow separation of concerns the `RenderCfg` has been removed from Simulation configuration, and all global settings were moved to IsaacRTXRenderer configuration. 

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


rilei-nvidia (approved):
Hey @pbarejko I read through the code changes and they look good.  

Please add a PR description with the context motivating these changes and this should be OK to merge!

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -11,218 +11,14 @@
...
+from typing import Literal
 
 from isaaclab.physics import PhysicsCfg
 from isaaclab.sim.spawners.materials.physics_materials_cfg import RigidBodyMaterialCfg
 from isaaclab.utils.configclass import configclass
 from isaaclab.visualizers import VisualizerCfg
 
 
-@configclass
-class RenderCfg:
```
huidongc, line 23: Since we take a more aggressive practice here - removing the class directly vs. deprecating it for a period, we should better still provide some doc how to transition from the old class to the new class.

docs/source/how-to/configure_rendering.rst
```diff
@@ -19,15 +19,18 @@ Selecting a Rendering Mode
 
 Rendering modes can be selected in 2 ways.
 
-1. using the ``rendering_mode`` input class argument in :class:`~sim.RenderCfg`
+1. using the ``rendering_mode`` field in
+   :class:`~isaaclab_physx.renderers.IsaacRtxRendererGlobalSettingsCfg`
```
huidongc, line 23: Maybe add a note here for existing users of `sim.RenderCfg`. We don't use deprecation approach but directly remove it, so better have a doc to guide transitioning.

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -46,6 +46,20 @@ def __init__(self) -> None:
...
+            return
+        for stored_cfg, _renderer in self._renderer_entries:
+            if getattr(stored_cfg, "renderer_type", None) != "isaac_rtx" or not hasattr(stored_cfg, "global_settings"):
+                continue
+            if stored_cfg.global_settings != cfg.global_settings:
+                raise ValueError(
+                    "Isaac RTX global settings differ across camera renderer configs. "
+                    "These settings are process-global; configure the same "
+                    "IsaacRtxRendererCfg.global_settings for every Isaac RTX camera."
+                )
```
huidongc, line 61: Since the global settings of each renderer must be always the same, why do we have to own the global settings in each IsaacRTXRendererCfg? Can they separate the cfg classes with a clear scope?

- IsaacRTXRendererGlobalSettingsCfg: process global
- IsaacRTXRendererCfg: no ownership of the global, just owning its renderer-specific data

source/isaaclab/isaaclab/sim/simulation_cfg.py
huidongc: Oh, I see the changelog. NVM.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
```diff
@@ -26,6 +54,99 @@
...
+
+    preset_filename = os.path.join(isaaclab_app_exp_path, f"rendering_modes/{rendering_mode}.kit")
+    if os.path.exists(preset_filename):
+        with open(preset_filename, "rb") as file:
+            _apply_nested_preset(settings, tomllib.load(file))
+    else:
+        logger.warning("[isaac_rtx] Render preset file not found: %s", preset_filename)
+
+
+def apply_isaac_rtx_global_settings(global_settings: Any, settings: Any) -> None:
```
huidongc: Could use default value of None for `settings`, in which case it will use get_setting_manager().

This should save a bit of boilerplate code across the repo:

```py
        settings = get_settings_manager()
        apply_isaac_rtx_global_settings(self.cfg.global_settings, settings)
```

huidongc: Is `global_settings` the source to read from, `settings` is the destination to apply new settings to? The doc can explain it a bit.

docs/source/how-to/configure_rendering.rst
pbarejko: I see where you are coming from, however I am taking advantage of the major version bump and knowingly this is a breaking change. I removed ambiguity such as - you can not set render settings for IsaacSimRTX when you render with Newton Warp or OVRTX. If we keep this field for backwards compatibility it will do us more harm than good leading us into corner cases as the ambiguity as mentioned above.

huidongc: I see. What I am asking for is maybe a note to guide existing users what this new class replaces with, but maybe the changelog itself has explained it well enough.

source/isaaclab/isaaclab/renderers/render_context.py
pbarejko: > why do we have to own the global settings in each IsaacRTXRendererCfg

There are few reasons for this:

* Because those settings have to be consumed by the renderer through Render configuration - this is a single argument with definition how renderer is instantiated.
* The renderer configuration must be serializable, by keeping it outside of the renderer configuration we leave some of the code not serialized. Certain settings might vanish. 

I didn't know how to make it better, than what it is now.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
pbarejko: This is a good suggestion, thanks! Also I will add the documentation!

source/isaaclab/isaaclab/renderers/render_context.py
huidongc: I see now. Thanks for explanation.

docs/source/how-to/configure_rendering.rst
pbarejko: updated the docs with information how to transition from RenderCfg to global settings. thanks!

huidongc (approved):

PR #6281, opened 2026-06-27
Fix Newton marker world rendering

ooctipus:
## Summary

- add `NewtonVisualizerCfg.world_spacing` for visual-only separation of co-located simulation worlds
- render dense environment-major marker batches using the same selected worlds and offsets as model geometry
- select marker state along its environment axis and broadcast offsets without flattened gather or repeated-offset tensors
- cache marker prototype conversion and keep filtering/offset application in one typed render path
- resolve and deduplicate visible environment IDs once, then pass the exact selection to the Newton viewer
- document partial Newton visualization and its dense environment-major marker convention

Non-divisible marker batches remain global and are neither filtered nor offset.

## Testing

- `22 passed`: `source/isaaclab_visualizers/test/test_newton_adapter.py`
- `10 passed`: scoped Newton marker render and partial-update tests
- CPU marker state with CUDA viewer offsets: selected worlds produced `[12, 13, 36, 37]`
- all-world offset smoke: `[0, 1, 12, 13, 24, 25, 36, 37]`
- non-divisible global marker smoke remained `[0, 1, 2]`
- fail-before-fix verification: old filtering produced `[1, 3, 5, 7]` instead of `[12, 13, 36, 37]`; the old allocation path also invoked the forbidden `torch.arange` regression guard
- dynamic reordered selections `[3, 1]` and `[1, 3]` preserve transform, scale, and source tensor state
- `./isaaclab.sh -d` passed with warnings treated as errors
- scoped pre-commit passed, including the LFS hook
- repository-wide pre-commit passed with only the known unrelated golden-image LFS hook skipped
- changelog fragment check passed

### DIFF
--- a/docs/source/overview/core-concepts/visualization.rst
+++ b/docs/source/overview/core-concepts/visualization.rst
@@ -163,6 +163,22 @@ There are 3 fields exposed in the ``VisualizerCfg`` for selecting environments f
 
 Also, there is a CLI arg ``--max_visible_envs`` that overrides ``VisualizerCfg.max_visible_envs`` for the run.
 
+Newton environments can share simulated coordinates, for example when ``scene.env_spacing=0``.
+Use :attr:`~isaaclab_visualizers.newton.NewtonVisualizerCfg.world_spacing` to arrange selected
+worlds visually without changing their simulated poses:
+
+.. code-block:: python
+
+
+    from isaaclab_visualizers.newton import NewtonVisualizerCfg
+    NewtonVisualizerCfg(
+        visible_env_indices=[0, 1, 2, 3],
+        world_spacing=(2.0, 2.0, 0.0),
+    )
+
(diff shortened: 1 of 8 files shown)


docs/source/overview/core-concepts/visualization.rst
```diff
@@ -163,6 +163,22 @@ There are 3 fields exposed in the ``VisualizerCfg`` for selecting environments f
 
 Also, there is a CLI arg ``--max_visible_envs`` that overrides ``VisualizerCfg.max_visible_envs`` for the run.
 
+Newton environments can share simulated coordinates, for example when ``scene.env_spacing=0``.
```
kellyguo11, line 166: might be clearer if we start with explaining that `env_spacing` controls the visual spacing/location of the envs and how differs from the spacing/coordinates in physics. "share simulated coordinates" might be a bit harder to grasp for users who are not as familiar with the concept.

kellyguo11 (approved):

PR #6287, opened 2026-06-28
Adding Rigid Body Material USD data classes and writers

vidurv-nvidia:
## Description

Converts **rigid-body physics materials** to the single-namespace "fragment" model already used by the rigid-body / collision / mass / mesh / tendon / joint-drive / articulation families. Additive and backward-compatible — the legacy inheritance cfgs remain as deprecated shims.

The one material-specific twist vs. the schema families: a physics material is **spawned as its own `UsdShade.Material` prim and bound** (not applied onto the body prim). So the family writer both spawns the prim + applies the anchor *and* dispatches the fragment list.

### Added
- **Core** (`isaaclab.sim.spawners.materials`):
  - `RigidBodyMaterialFragment` — marker base typing the `physics_material` slot.
  - `UsdPhysicsRigidBodyMaterialCfg` — solver-common `physics:*` friction/restitution (anchor `UsdPhysics.MaterialAPI`).
  - `spawn_rigid_body_material_from_fragments(prim_path, fragments, stage)` — spawns the `UsdShade.Material` prim, applies the `MaterialAPI` anchor, dispatches each fragment's `func` (default `apply_namespaced`).
  - `spawn_physics_material(prim_path, material, stage)` — slot dispatcher: fragment list → fragment writer, else legacy cfg via its own `func`.
- **PhysX** (`isaaclab_physx.sim.spawners.materials`):
  - `PhysxMaterialCfg` — single-namespace `physxMaterial:*` (`PhysxMaterialAPI`): compliant-contact spring + combine-mode tokens.

### Changed
- Spawner `physics_material` slots (`shapes`, `meshes`, `from_files`) now accept `RigidBodyMaterialFragment | list[...]` in addition to the legacy material cfg; consume sites route through `spawn_physics_material`. **Legacy single-cfg path unchanged.**

### Scope
- **Rigid-body materials only.** Deformable-body materials are deferred — they pair with the deferred deformable-body family (multi-inherited OmniPhysics + PhysX material APIs).

## Tests
- New `test_material_fragments.py` (6): fragment metadata; spawn-from-fragments composes `physics:*` + `physxMaterial:*` with the `MaterialAPI` anchor + `PhysxMaterialAPI`; single-fragment; partial-update (None left unauthored); slot dispatcher handles both fragment and legacy forms.
- Regression: existing `test_spawn_materials.py` (6) and `test_spawn_shapes.py` (12) pass — legacy path intact.

## Checklist
- [x] Ran `./isaaclab.sh --format`
- [x] Added changelog fragments (core + physx, minor)


source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -89,6 +90,45 @@ class RigidBodyMaterialBaseCfg(PhysicsMaterialCfg):
...
+    _usd_namespace: ClassVar[str | None] = "physics"
+    _usd_applied_schema: ClassVar[str | None] = None  # MaterialAPI applied by the family writer
+
+    static_friction: float | None = None
+    """The static friction coefficient. Writes ``physics:staticFriction``."""
+
+    dynamic_friction: float | None = None
+    """The dynamic friction coefficient. Writes ``physics:dynamicFriction``."""
+
+    restitution: float | None = None
```
ooctipus, line 137: Blocking: this fragment omits `physics:density`. The configured Kit 110.1.2 schema registry reports four `UsdPhysics.MaterialAPI` attributes: dynamic friction, static friction, restitution, and density. Material density participates in mass computation through material binding, so it is not equivalent to placing density on `MassCfg`. Please add `density: float | None` and a round-trip test; a schema-parity assertion would also prevent future drift.

source/isaaclab_physx/isaaclab_physx/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -203,6 +204,43 @@ class PhysxRigidBodyMaterialCfg(RigidBodyMaterialBaseCfg):
...
+
+    Irrelevant if compliant contacts are disabled (``compliant_contact_stiffness`` is zero). Writes
+    ``physxMaterial:compliantContactDamping``.
+    """
+
+    friction_combine_mode: Literal["average", "min", "multiply", "max"] | None = None
+    """How friction is combined during collisions. Writes ``physxMaterial:frictionCombineMode``."""
+
+    restitution_combine_mode: Literal["average", "min", "multiply", "max"] | None = None
+    """How restitution is combined during collisions. Writes ``physxMaterial:restitutionCombineMode``."""
```
ooctipus, line 269: Blocking: the target `PhysxMaterialAPI` has six attributes, while this fragment exposes four. It is missing `damping_combine_mode` and `compliant_contact_acceleration_spring`; both affect compliant-contact behavior. Please add both fields and test them, and update the 104.2 schema link because it no longer represents the Isaac Sim 6 / Kit 110 surface.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -89,6 +90,45 @@ class RigidBodyMaterialBaseCfg(PhysicsMaterialCfg):
...
 
 
+@configclass
+class RigidBodyMaterialFragment(SchemaFragment):
+    """Marker base for rigid-body physics-material fragments; types the ``physics_material`` slot.
+
+    A rigid-body physics material is a single ``UsdShade.Material`` prim that carries one or more
+    physics-material schemas. The fragments author single namespaces onto that prim: the solver-common
+    ``physics:*`` friction/restitution (:class:`UsdPhysicsRigidBodyMaterialCfg`) and any backend-specific
+    namespace (e.g. PhysX ``physxMaterial:*`` via :class:`~isaaclab_physx.sim.spawners.materials.PhysxMaterialCfg`).
```
ooctipus: Blocking: #5276 already added `NewtonMaterialPropertiesCfg` with `newton:torsionalFriction` and `newton:rollingFriction`, but this family adds only the PhysX backend fragment. Because the new dispatcher rejects a legacy cfg inside a fragment list, Newton properties cannot compose with `UsdPhysicsRigidBodyMaterialCfg`. That diverges from the backend-symmetric pattern in #5976, #6254, #6264, and #6273. Please add a Newton material fragment under `isaaclab_newton.sim.spawners.materials`, retaining the existing class as the legacy path.

source/isaaclab/isaaclab/sim/spawners/meshes/meshes_cfg.py
```diff
@@ -57,9 +57,17 @@ class MeshCfg(RigidObjectSpawnerCfg, DeformableObjectSpawnerCfg):
     This parameter is ignored if `physics_material` is not None.
     """
 
-    physics_material: materials.PhysicsMaterialCfg | None = None
+    physics_material: (
+        materials.PhysicsMaterialCfg
+        | materials.RigidBodyMaterialFragment
+        | list[materials.RigidBodyMaterialFragment]
```
ooctipus, line 63: Blocking: this newly advertised fragment-list form is still rejected by `_spawn_mesh_geom_from_mesh` whenever `rigid_props` is set. Its guard checks `isinstance(cfg.physics_material, RigidBodyMaterialCfg)`, so a lone fragment or fragment list raises before reaching `spawn_physics_material`. Please make that validation backend-neutral and add a real rigid-mesh spawner test; direct dispatcher tests do not exercise this path.

source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
```diff
@@ -211,7 +212,7 @@ def spawn_ground_plane(
 
     # Create physics material
     if cfg.physics_material is not None:
-        cfg.physics_material.func(f"{prim_path}/physicsMaterial", cfg.physics_material)
+        spawn_physics_material(f"{prim_path}/physicsMaterial", cfg.physics_material)
```
ooctipus: This call now supports fragments at runtime, but `GroundPlaneCfg.physics_material` remains annotated and documented only as `materials.RigidBodyMaterialCfg`. Please widen that field to the same legacy-or-fragment union and add a ground-plane binding test so the public type contract matches this consumer.

ooctipus (commented):
Focused review against the property-completeness and separation conventions established in #5275, #5276, #5976, #6254, #6263, #6264, and #6273. The core/backend boundary is directionally clean, but I found the following completeness and integration gaps.

vidurv-nvidia:
Thanks @ooctipus — all five addressed in a8bc6aab4f2 (schema facts verified against the installed Kit 110 schemas, not guessed):

1. **`physics:density`** — added `density: float | None` to `UsdPhysicsRigidBodyMaterialCfg`, plus a **schema-parity test** asserting the fragment covers exactly the four `UsdPhysics.MaterialAPI` attributes (static/dynamic friction, restitution, density) so future drift fails loudly. Density round-trip test added.
2. **`PhysxMaterialAPI` completeness** — added `damping_combine_mode` (token) and `compliant_contact_acceleration_spring` (bool); both tested. Updated the stale 104.2 doc link to the Kit 110 schema surface.
3. **Newton material fragment** — added `NewtonMaterialCfg(RigidBodyMaterialFragment)` under `isaaclab_newton.sim.spawners.materials` (`newton:torsionalFriction` / `newton:rollingFriction`), with `NewtonMaterialPropertiesCfg` retained as the legacy path. Test confirms it composes in a fragment list alongside `UsdPhysicsRigidBodyMaterialCfg` (both `newton:*` and `physics:*` authored on one prim).
4. **Mesh spawner guard** — `_spawn_mesh_geom_from_mesh` rigid-vs-deformable split is now backend-neutral: it accepts a lone fragment or fragment list (not just `RigidBodyMaterialCfg`) before dispatching to `spawn_physics_material`. Added a real rigid-mesh-with-fragment-material spawner test (exercises the path direct dispatcher tests missed).
5. **`GroundPlaneCfg.physics_material`** — widened to the same legacy-or-fragment union `FileCfg` uses. Its spawner already routed through `spawn_physics_material`, so only the annotation was stale; added a ground-plane fragment-binding test.

Tests: `test_material_fragments.py` 12/12, `test_newton_schemas.py` 18/18, mesh/collision/from-files regressions clean (the one `test_spawn_urdf` failure is pre-existing and unrelated — a `merge_fixed_joints` env mismatch this PR does not touch). Changelog fragments added for all three packages.

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -77,3 +78,37 @@ class NewtonSurfaceDeformableBodyMaterialCfg(SurfaceDeformableBodyMaterialBaseCf
...
+    torsional_friction: float | None = None
+    """Torsional friction coefficient (resistance to spinning at a contact point) [dimensionless].
+
+    Writes ``newton:torsionalFriction``. Range: [0, inf).
+    """
+
+    rolling_friction: float | None = None
+    """Rolling friction coefficient (resistance to rolling motion) [dimensionless].
+
+    Writes ``newton:rollingFriction``. Range: [0, inf).
```
ooctipus, line 120: Blocking: the registered-schema audit is 2/2, but this fragment is only 2/6 against the Newton runtime that IsaacLab actually pins. At commit `79e95bf` from `isaaclab_newton/pyproject.toml`, [`SchemaResolverNewton.mapping[PrimType.MATERIAL]`](https://github.com/newton-physics/newton/blob/79e95bf5571d70a0a46c8eaedc80644531d27368/newton/_src/usd/schemas.py#L164-L170) also consumes `newton:contactStiffness`, `newton:contactDamping`, `newton:contactFrictionGain`, and `newton:contactAdhesion`. Newton uses these as the canonical material-side replacements for the deprecated per-shape `contact_ke/kd/kf/ka` fields, and its pinned import tests exercise them on a bound `NewtonMaterialAPI` material.

Please add `contact_stiffness`, `contact_damping`, `contact_friction_gain`, and `contact_adhesion` as `float | None` fields, with an authoring/parser regression test. `apply_namespaced` can create these float attrs while the generated schema plugin catches up. This follows the #5276 precedent of auditing the actual backend reader, rather than treating a lagging generated schema as the whole property contract.

source/isaaclab/isaaclab/sim/spawners/meshes/meshes.py
```diff
@@ -367,7 +373,15 @@ def _spawn_mesh_geom_from_mesh(
         if not isinstance(cfg.physics_material, DeformableBodyMaterialBaseCfg):
             raise ValueError("Deformable properties require a deformable physics material.")
     if cfg.rigid_props is not None and cfg.physics_material is not None:
-        if not isinstance(cfg.physics_material, RigidBodyMaterialCfg):
+        # accept anything spawn_physics_material accepts for the rigid case: a legacy rigid-body
+        # material cfg, a single fragment, or a list/tuple of fragments
+        physics_material_frags = (
+            cfg.physics_material if isinstance(cfg.physics_material, (list, tuple)) else [cfg.physics_material]
+        )
+        is_rigid_material = isinstance(cfg.physics_material, RigidBodyMaterialCfg) or all(
```
ooctipus: This only partially addresses the earlier mesh-spawner finding. `RigidBodyMaterialCfg` is the deprecated PhysX alias and is a *subclass* of the recommended `PhysxRigidBodyMaterialCfg`, so the canonical PhysX cfg is not an instance of it. The same check rejects #5276 `NewtonMaterialPropertiesCfg`, even though both derive from `RigidBodyMaterialBaseCfg` and both are accepted by `spawn_physics_material`. Thus a rigid `MeshCfg` still fails before the supposedly backend-neutral dispatcher for valid legacy cfgs.

Please import/check the core `RigidBodyMaterialBaseCfg` here and add a real mesh regression using at least `PhysxRigidBodyMaterialCfg` (ideally Newton legacy too). The new fragment-list test cannot catch this transition path, and retaining this alias check keeps a PhysX-specific dependency in core.

source/isaaclab/isaaclab/sim/spawners/from_files/from_files_cfg.py
```diff
@@ -259,5 +267,11 @@ class GroundPlaneCfg(SpawnerCfg):
     size: tuple[float, float] = (100.0, 100.0)
     """The size of the ground plane. Defaults to 100 m x 100 m."""
 
-    physics_material: materials.RigidBodyMaterialCfg = materials.RigidBodyMaterialCfg()
-    """Physics material properties. Defaults to the default rigid body material."""
+    physics_material: (
+        materials.PhysicsMaterialCfg | materials.RigidBodyMaterialFragment | list[materials.RigidBodyMaterialFragment]
```
ooctipus: The fragment case and binding test are fixed, but this legacy side is now broader than the old rigid-only contract: `PhysicsMaterialCfg` also includes deformable and surface-deformable material cfgs. `spawn_physics_material` will happily spawn one and the ground-plane path will bind it to a rigid collision plane.

Please use the backend-neutral rigid base here: `RigidBodyMaterialBaseCfg | RigidBodyMaterialFragment | list[RigidBodyMaterialFragment]`. That preserves #5275 backend subclassing (including Newton) without advertising deformable materials on a rigid-only spawner.

source/isaaclab_physx/changelog.d/vidurv-schema-frag-materials.minor.rst
```diff
@@ -0,0 +1,12 @@
...
+
+* Added :class:`~isaaclab_physx.sim.spawners.materials.PhysxMaterialCfg`, a single-namespace
+  ``physxMaterial`` rigid-body physics-material fragment (compliant-contact spring stiffness/damping
+  and the friction/restitution combine-mode tokens) backing ``PhysxMaterialAPI``.
+* Added :attr:`~isaaclab_physx.sim.spawners.materials.PhysxMaterialCfg.damping_combine_mode` (writes
+  ``physxMaterial:dampingCombineMode``) and
+  :attr:`~isaaclab_physx.sim.spawners.materials.PhysxMaterialCfg.compliant_contact_acceleration_spring`
+  (writes ``physxMaterial:compliantContactAccelerationSpring``), completing the fragment's coverage
+  of ``PhysxMaterialAPI``. Also added the same two fields to the legacy
+  :class:`~isaaclab_physx.sim.spawners.materials.PhysxRigidBodyMaterialCfg`.
```
ooctipus, line 12: This compatibility claim is not implemented: `PhysxRigidBodyMaterialCfg` still stops at `compliant_contact_stiffness`, `compliant_contact_damping`, `friction_combine_mode`, and `restitution_combine_mode`; the two new fields exist only on `PhysxMaterialCfg`. Please either add them to the supported legacy cfg with coverage, as stated here, or remove/correct this changelog sentence. Adding them is the cleaner transition because the existing metadata writer already handles them and it keeps legacy/new property parity during the deprecation window.

ooctipus (changes requested):
Follow-up review of `a8bc6aab4f2` against my prior review at `bd0f0632790`.

The response materially improves the PR:

- `UsdPhysicsRigidBodyMaterialCfg` now covers all 4 registered `MaterialAPI` attributes, with a useful exact-parity test.
- `PhysxMaterialCfg` now covers all 6 registered Kit 110 `PhysxMaterialAPI` attributes.
- A Newton fragment was added in the correct backend package and composes with the USD fragment.
- The fragment-list mesh path and ground-plane path now have real spawner/binding tests.
- Current CI is fully green and the diff passes whitespace validation.

I still need changes for two substantive gaps:

1. The new Newton fragment matches the lagging generated schema (2 fields), but not the actual Newton reader pinned by this repository (6 material fields). The pinned resolver additionally consumes contact stiffness, damping, friction gain, and adhesion from the bound Newton material. Per the runtime-reader audit precedent in #5276, those four canonical material attributes need to be represented and tested.
2. The mesh integration fix still checks the deprecated PhysX `RigidBodyMaterialCfg` alias. It therefore rejects the recommended `PhysxRigidBodyMaterialCfg` and `NewtonMaterialPropertiesCfg`, despite both deriving from the core rigid-material base and being accepted by the dispatcher. Using `RigidBodyMaterialBaseCfg` restores the clean #5275 core/backend boundary.

Two narrower inconsistencies are inline as well: `GroundPlaneCfg` now advertises generic/deformable material cfgs despite being rigid-only, and the PhysX changelog claims two fields were added to the legacy cfg when they were only added to the fragment.

So the original five comments are largely addressed, but property completeness must include the active backend consumer, and the remaining legacy transition guard is not yet backend-neutral.

vidurv-nvidia:
Thanks @ooctipus — round 2 addressed in 2d70ad4a37d. Two of these were real bugs in my prior fix; good catches.

- **Mesh guard (B):** you're right — `RigidBodyMaterialCfg` is the deprecated *leaf* alias, so the canonical `PhysxRigidBodyMaterialCfg` and `NewtonMaterialPropertiesCfg` (both ancestors, both accepted by `spawn_physics_material`) failed the check. Switched the guard to the core base `RigidBodyMaterialBaseCfg` and dropped the PhysX-specific import from core. Added a mesh regression spawning a rigid mesh with a legacy `PhysxRigidBodyMaterialCfg` physics_material (verified it failed before the fix, passes after).
- **GroundPlaneCfg (C):** narrowed the legacy member from `PhysicsMaterialCfg` (which also admits deformable/surface-deformable) to `RigidBodyMaterialBaseCfg | RigidBodyMaterialFragment | list[...]`, preserving backend subclassing without advertising deformables on a rigid-only plane. Corrected the changelog wording too.
- **Legacy PhysX parity (D):** implemented the changelog claim — added `damping_combine_mode` and `compliant_contact_acceleration_spring` to legacy `PhysxRigidBodyMaterialCfg`; the metadata-driven writer authors them with no other changes. Test asserts both `physxMaterial:*` attrs round-trip.
- **Newton 2/6 (A):** audited the pinned runtime reader (`SchemaResolverNewton.mapping[PrimType.MATERIAL]` at `79e95bf5`) rather than the lagging generated schema, and added `contact_stiffness`→`newton:contactStiffness`, `contact_damping`→`newton:contactDamping`, `contact_friction_gain`→`newton:contactFrictionGain`, `contact_adhesion`→`newton:contactAdhesion` to `NewtonMaterialCfg`. `apply_namespaced` authors them as raw float attrs while the generated plugin catches up; regression test asserts all six `newton:*` attrs round-trip.

Tests: `test_material_fragments.py` 15/15, `test_newton_schemas.py` 19/19, 0 failures/errors.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
```diff
@@ -37,9 +37,18 @@ class ShapeCfg(RigidObjectSpawnerCfg):
     If the path is relative, then it will be relative to the prim's path.
     This parameter is ignored if `physics_material` is not None.
     """
-    physics_material: materials.PhysicsMaterialCfg | None = None
+    physics_material: (
+        materials.PhysicsMaterialCfg
```
ooctipus: Blocking: `ShapeCfg` is a rigid-object spawner, but this legacy arm remains `PhysicsMaterialCfg`, which also admits volume and surface deformable material cfgs. `spawn_physics_material()` will spawn and bind those rather than reject them. Please mirror the now-correct `GroundPlaneCfg` contract here: `RigidBodyMaterialBaseCfg | RigidBodyMaterialFragment | list[RigidBodyMaterialFragment] | None`, and add acceptance/rejection coverage. This keeps backend legacy subclasses admissible without advertising deformable materials on a rigid-only spawner.

source/isaaclab/changelog.d/vidurv-schema-frag-materials.minor.rst
```diff
@@ -0,0 +1,26 @@
+Added
+^^^^^
+
+* Added the rigid-body physics-material "fragment" classes
+  :class:`~isaaclab.sim.spawners.materials.RigidBodyMaterialFragment` (marker base) and
+  :class:`~isaaclab.sim.spawners.materials.UsdPhysicsRigidBodyMaterialCfg` (solver-common
+  ``physics:*`` friction/restitution), plus the family writer
+  :func:`~isaaclab.sim.spawners.materials.spawn_rigid_body_material_from_fragments` and the slot
+  dispatcher :func:`~isaaclab.sim.spawners.materials.spawn_physics_material`. Spawner
+  ``physics_material`` slots now accept a list of single-namespace fragments in addition to the
```
ooctipus: Blocking completeness gap: the public terrain material slot still bypasses this dispatcher. Plane terrain now reaches `GroundPlaneCfg`, but generated terrain reaches `terrains/utils.py:create_prim_from_mesh`, which still calls `physics_material_cfg.func(...)`. On this head I reproduced a fragment list failing there with `AttributeError: 'list' object has no attribute 'func'`. Please widen `TerrainImporterCfg.physics_material`, route the generated-mesh utility through `spawn_physics_material`, and add a generator-terrain regression test.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -89,6 +90,52 @@ class RigidBodyMaterialBaseCfg(PhysicsMaterialCfg):
...
+    static_friction: float | None = None
+    """The static friction coefficient. Writes ``physics:staticFriction``."""
+
+    dynamic_friction: float | None = None
+    """The dynamic friction coefficient. Writes ``physics:dynamicFriction``."""
+
+    restitution: float | None = None
+    """The restitution coefficient. Writes ``physics:restitution``."""
+
+    density: float | None = None
```
ooctipus, line 140: Blocking under the transition-parity convention: this makes the fragment complete at 4/4 `UsdPhysics.MaterialAPI` properties, but `RigidBodyMaterialBaseCfg` above remains 3/4 and cannot author material density. Please also add `density: float | None = None` to the legacy base and cover its authoring path. That is backward-compatible because `None` is unauthored, and it matches the fragment/legacy parity established by #6263 and the PhysX legacy parity fix in this revision.

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -77,3 +78,73 @@ class NewtonSurfaceDeformableBodyMaterialCfg(SurfaceDeformableBodyMaterialBaseCf
...
+        The generated ``NewtonMaterialAPI`` USD schema currently only declares the two friction
+        attributes; the four contact attributes are still authored as raw ``newton:*`` USD
+        attributes and are read directly by Newton's schema resolver.
+
+    Composes with other rigid-body material fragments (e.g.
+    :class:`~isaaclab.sim.spawners.materials.UsdPhysicsRigidBodyMaterialCfg`) in the same fragment
+    list passed to
+    :func:`~isaaclab.sim.spawners.materials.spawn_rigid_body_material_from_fragments`. For the
+    legacy (non-fragment) equivalent, see
+    :class:`~isaaclab_newton.sim.schemas.NewtonMaterialPropertiesCfg`.
```
ooctipus, line 105: Blocking: this calls `NewtonMaterialPropertiesCfg` the legacy equivalent, but the fragment and pinned runtime reader now have six properties while that legacy cfg still has only torsional/rolling friction. Please add `contact_stiffness`, `contact_damping`, `contact_friction_gain`, and `contact_adhesion` as optional fields to the legacy cfg and test them. This is the same transition parity used by #6254/#6273 and already applied to the PhysX material cfg in this revision.

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -77,3 +78,73 @@ class NewtonSurfaceDeformableBodyMaterialCfg(SurfaceDeformableBodyMaterialBaseCf
...
+
+    contact_damping: float | None = None
+    """Contact normal-force damping coefficient [N·s/m].
+
+    Writes ``newton:contactDamping``. Replaces the deprecated per-shape ``kd`` contact parameter;
+    used by the SemiImplicit, Featherstone, MuJoCo, and VBD solvers.
+    """
+
+    contact_friction_gain: float | None = None
+    """Friction-force stiffness gain used by the tangential (friction) contact response [N/m].
```
ooctipus: Nonblocking documentation correction: the pinned Newton runtime documents `Model.shape_material_kf` as a tangential friction response gain in `[N·s/m]`, not `[N/m]`. Please update this unit so the newly exposed property's contract matches its consumer.

ooctipus (commented):
Re-reviewed the latest head. The previously reported schema-coverage issues are fixed: the USD, PhysX, and Newton fragments now cover their full active property surfaces, and the core/backend separation is clean. I am keeping my existing changes-requested state because four integration/transition gaps remain in the inline comments below. I also confirmed the existing explicit-stage thread with a two-stage local reproduction and replied there rather than duplicating it. Nonblocking test note: the Newton six-attribute test proves USD authoring, but still does not exercise SchemaResolverNewton parsing as previously requested.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials.py
```diff
@@ -12,10 +12,87 @@
...
+        stage: The stage to spawn on. Defaults to None, in which case the current stage is used.
+
+    Returns:
+        The spawned material prim.
+    """
+    fragments = material if isinstance(material, (list, tuple)) else [material]
+    if fragments and all(isinstance(f, physics_materials_cfg.RigidBodyMaterialFragment) for f in fragments):
+        return spawn_rigid_body_material_from_fragments(prim_path, list(fragments), stage)
+    # legacy single-cfg path (rigid or deformable material cfg with its own spawner ``func``)
+    return material.func(prim_path, material)
```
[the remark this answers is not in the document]
ooctipus: Confirmed on current head `2d70ad4a37d` with a two-stage reproduction: `spawn_physics_material(..., legacy_cfg, stage=other_stage)` creates the material on the global current stage, while `other_stage` remains empty. The explanatory comment documents the limitation but the public `stage` contract is still branch-dependent. Please either forward `stage` through the built-in legacy material writers or explicitly remove/narrow that contract.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
vidurv-nvidia: Fixed in c84801e75db. ShapeCfg.physics_material is now `RigidBodyMaterialBaseCfg | RigidBodyMaterialFragment | list[RigidBodyMaterialFragment] | None` — backend legacy subclasses stay admissible, deformables are no longer advertised. For the record, the broad `PhysicsMaterialCfg` arm predates this PR (develop has it today); tightened here since this PR defines the fragment-era slot contract. Coverage is structural rather than per-slot: `test_material_slot_unions_match_fragment_kind`-style enforcement in 1983a592a46 asserts every rigid-only slot (ShapeCfg, GroundPlaneCfg, TerrainImporterCfg, SimulationCfg) excludes `PhysicsMaterialCfg` and admits base+fragments, and that the mixed slots (FileCfg, MeshCfg) keep it — so a future slot with the wrong union fails CI.

source/isaaclab/changelog.d/vidurv-schema-frag-materials.minor.rst
vidurv-nvidia: Fixed in 435240ed3f4. `create_prim_from_mesh` now routes through `spawn_physics_material` (your `AttributeError` repro is the regression test — verified failing before the fix with exactly that error), and `TerrainImporterCfg.physics_material` is widened to `RigidBodyMaterialBaseCfg | RigidBodyMaterialFragment | list[RigidBodyMaterialFragment]`. The generated-terrain test asserts both `physics:*` and `newton:*` attrs authored end-to-end through the fragment list.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
vidurv-nvidia: Fixed in bb3226f2d21 — `density: float | None = None` on `RigidBodyMaterialBaseCfg` (None default = unauthored, backward compatible; the metadata-driven writer picks it up under `physics:*`). Authoring test covers the set and unset cases. 1983a592a46 additionally adds an introspective parity test (fragment fields == legacy fields per backend, minus `func`) so this class of asymmetry fails CI instead of waiting for review.

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
vidurv-nvidia: Fixed in 85592afc4fe. The four contact fields are on the legacy `NewtonMaterialPropertiesCfg` with docstrings word-identical to the fragment, and an authoring test asserts all four `newton:contact*` attrs. The structural parity test in 1983a592a46 pins fragment == legacy-minus-base for Newton going forward.

source/isaaclab_newton/isaaclab_newton/sim/spawners/materials/physics_materials_cfg.py
vidurv-nvidia: Fixed in 85592afc4fe — `[N·s/m]` in both the fragment and the (new) legacy field docstrings, matching the pinned runtime doc for `Model.shape_material_kf`.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials.py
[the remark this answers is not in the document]
vidurv-nvidia: Resolved by narrowing the contract in ffa19f3c12f: the legacy path now raises `ValueError` when given an explicit stage that is not the current stage (your two-stage repro is the regression test, incl. asserting nothing leaks onto the current stage). Passing the current stage explicitly remains valid — the in-tree spawners pass `stage=stage` unconditionally with legacy cfgs, so raising on any explicit stage would have broken them. Truly forwarding the stage would require making the `@clone` path-resolution stage-aware, which is out of scope for the transition bridge; the fragment path honors explicit stages, and the docstring now states the legacy path is current-stage-bound. Flagged as **Breaking:** in the changelog with migration guidance.

vidurv-nvidia:
While addressing the latest review I audited the full material-slot surface (every `physics_material` cfg field and every consumer that spawns from one) rather than fixing the flagged sites alone. That audit found three more sites with the same issues, fixed in the same push:

- **`SimulationCfg.physics_material`** was typed as the deprecated leaf and its consumer (the PhysX default-material spawn) bypassed `spawn_physics_material` — a fragment list there would have crashed the same way the terrain path did. Both fixed together in e0e647d78b2 (slot widened + routed + end-to-end test on `{physics_prim_path}/defaultMaterial`).
- **`spawn_from_usd_with_compliant_contact_material`** called the material cfg's `func` directly; routed through the dispatcher in 31abf2c31ed (behavior-preserving, pinned by test). The ground-plane call now also passes `stage=` like its siblings.
- **Default instances** of `SimulationCfg` / `TerrainImporterCfg` / `GroundPlaneCfg` `physics_material` constructed the deprecated leaf class; swapped to `RigidBodyMaterialBaseCfg` after proving byte-identical authored USD (equivalence test included). Called out as **Breaking:** in the changelog with migration guidance, since PhysX-only fields no longer exist on the default object for in-place mutation.

To keep these from recurring, 1983a592a46 adds two structural tests: an introspective fragment↔legacy field-parity check per backend, and a slot-union check asserting each `physics_material` slot's accepted types match what its spawner can honor. The mesh-spawner material check fixed earlier also gets an explicit `Fixed` changelog entry, since it rejects canonical rigid material classes on develop today independent of this PR.

Test state: `test_material_fragments.py` 23/23, `test_newton_schemas.py` 20/20; crash-class fixes verified failing before each fix.

ooctipus (approved):

PR #6299, opened 2026-06-30
Heterogeneous visual materials with GPU attribute randomization

ooctipus:
## Summary

This PR rebases visual-material randomization onto current `develop` and keeps the original declarative scene API while moving runtime writes onto renderer-native GPU paths.

### Architecture and ownership

`VisualMaterialCfg` remains a first-class `AssetBaseCfg` in `InteractiveSceneCfg`:

1. The material is declared beside the robot and other assets.
2. Its source prim is spawned before assets that bind it.
3. It participates in the same clone plan as the rest of the scene. `{ENV_REGEX_NS}` creates one material clone per environment; an absolute path creates one shared material.
4. Normal asset initialization resolves the cloned material/shader paths and registers the material with `RenderContext`.
5. `RenderContext` concatenates each numeric channel into one flat device buffer and asks active renderer/visualizer consumers for their writer.
6. Reset/runtime events sample directly on the target device and use one generic Warp scatter to update selected material/environment rows.

There is no owner-asset hierarchy, plan merging/reindexing, queued-material adapter, runtime relationship rewrite, or SimulationContext material service. `SimulationContext` only owns the shared `RenderContext` lifecycle.

### Runtime backend paths

- **Kit / Isaac RTX:** cached Fabric shader selections and batched Warp writes.
- **OVRTX:** cached native addresses for both legacy and OVStage APIs; dirty channel buffers are published directly from CUDA. Native cloning preserves relative material bindings.
- **Newton:** material bindings are compiled once into flat material-to-shape rows, then color updates scatter into `model.shape_color` with one Warp launch.

Runtime material writes do not author USD, walk the stage, transfer environment IDs to the CPU, convert tensors to lists, or loop over environment rows in Python. USD work is restricted to scene construction/import. OVRTX currently publishes the full dirty channel because its write API does not expose indexed row updates; the Lab-side partial reset remains a device scatter.

### Public behavior

`VisualMaterialCfg`, `VisualMaterial.write_channels`, `read_channel`, `randomize_visual_material`, `randomize_visual_color`, and the Newton-only `randomize_visual_shape` event remain available.

Numeric channels are:

- Preview Surface: color, roughness, metallic, emissive color, opacity.
- OmniPBR: color, roughness, metallic, specular, emissive color/intensity, opacity, UV scale/offset/rotation.
- OmniGlass: color, frosting roughness, and IOR.

Texture asset paths remain construction-time USD data. One static texture may be preloaded and its numeric UV channels randomized, but runtime texture-path selection is intentionally unsupported by this GPU-only pipeline. `prestartup` material writes are also rejected because renderer writers exist only after scene cloning/finalization.

Material and shape granularity stay separate:

- `randomize_visual_material` samples per declared material and environment on supported renderers.
- `randomize_visual_shape` samples per selected link and environment against Newton's live shape storage. Other renderers should declare one material per independently randomized part.

### Demo

`scripts/demos/visual_color_randomization.py` keeps the original heterogeneous five-style ANYmal scene:

1. Original authored appearance.
2. Textured body, legs, and feet.
3. Glass body, legs, and feet.
4. Solid body, legs, and feet.
5. Mixed textured body, solid legs, and glass feet.

Each family uses scene-declared per-environment materials for body, legs, and feet. The demo defaults to 512 environments and resets random subsets so partial writes are visible.

```bash
# PhysX + Kit/Fabric
uv run --extra isaacsim python scripts/demos/visual_color_randomization.py

# Newton physics + Newton GL
uv run python scripts/demos/visual_color_randomization.py \
    --physics newton_mjwarp --visualizer newton_gl
```

## Validation

- Repository formatting, Ruff, whitespace, conflict, codespell, license, RST, changelog, secret, and Git LFS gates passed.
- Core material/event, scene lifecycle, USD relationship, PhysX/Fabric, Newton material/shape, OVRTX clone/material, and visualizer hook suites passed.
- Broad cloner suite: 43 passed.
- Local CI-equivalent core `[2/3]`: 78/78 files, 2,212 tests, 0 failures/errors, 16 skipped.
- SimulationContext suite: 51 passed.
- Fresh golden-image rendering passed for Kit/Fabric, Newton Warp, OVRTX legacy, and OVRTX OVStage.
- `git diff --check` passed and `uv.lock` is unchanged.

Synchronized RTX 5090 measurements for 4096 environments:

| Path | Full reset | 512-env partial reset |
|---|---:|---:|
| Core flat-buffer scatter | 0.041 ms | 0.044 ms |
| Newton material events | 0.351 ms | 0.368 ms |
| Kit/Fabric material events | 1.008 ms | 1.047 ms |
| OVRTX legacy publish | 0.216 ms | 0.233 ms |
| OVRTX OVStage publish | 8.682 ms | 8.329 ms |

The production implementation is 862 net lines smaller than the previous PR revision (about 43%) while preserving the declarative asset/event/demo boundaries.


ruziniuuuuu:
I audited this PR's OVRTX seam against a real kitless Newton/OVRTX task using
cloned `UsdPreviewSurface` fixtures. The result supports the architecture here
and adds a useful regression case:

- Runtime: Isaac Lab `v3.0.0-beta2.patch1`, `isaaclab_ov==0.4.2`,
  `ovrtx==0.3.0.312915`.
- The installed adapter snapshots USD once and only publishes body/camera
  transforms afterward, so current reset-time Preview Surface writes never
  reach OVRTX.
- Pre-export randomization does work. Fresh-process seed 17 produced table
  color `(0.3964694, 0.4153489, 0.3775899)`, roughness `0.5688838`, and settled
  RGB mean `221.975`; seed 18 produced
  `(0.3911828, 0.4098105, 0.3725550)`, roughness `0.7043907`, and RGB mean
  `202.699`. Image hashes differed.

Could the OVRTX contract test cover a cloned `UsdPreviewSurface` material,
write both `color` and `roughness` after stage ingestion, and verify settled
pixels change? That would protect the exact gap observed here without adding an
application-owned material bridge.

For partial reset semantics, the documented bucket-global behavior is clear:
`env_ids` cannot imply row-local OVRTX appearance when rows share one bucket.
A test that fires the event with a subset and proves one global bucket update
would make that intentional limitation explicit and prevent callers from
mistaking run/reset timing for per-row diversity.

Lighting remains a separate capability because the current adapter has no
light-update lane and a single dome light is scene-global; I am not proposing
to fold it into this material PR.

ooctipus:
Root cause of the OVRTX failures found and fixed (933925f2437 + efb6acf5c43), reproduced and verified locally against the pinned `ovrtx` 0.3:

- **Per-env colors never reached clones**: `clone_usd` copies a subtree without remapping `material:binding` targets, so every cloned geometry kept rendering the *source* env's material — `test_per_env_visual_color_contract_ovrtx`'s `env1 got ('red','blue')` was env_1's cube bound to env_0's material. Fix: clone-plan rows that hold or bind a per-environment material bypass OVRTX cloning; their USD clones (already correctly wired per env by scene replication) stay in the exported stage.
- **Missing shape validation**: `notify_visual_material_written` documented one value row per shader prim but never enforced it (`test_notify_validates_color_shape`). Writes are now validated batch-first, then applied.

Local results on ovrtx 0.3: `test_ovrtx_visual_material_sync.py` 8/8, `test_color_randomization_contract_ovrtx.py` 2/2. The kitless `franka_cloth` golden failures plausibly share the binding root cause — let's see this CI round. Still open from the previous run: the `test_robot_load_performance` 65 s vs 60 s budget (possible spawn-time cost of authoring channel defaults — needs a local A/B) and the visualizers job docker-125 infra fail.

docs/source/how-to/cloning.rst
```diff
@@ -391,34 +440,11 @@ The dispatch shape is roughly:
...
+Every context receives the full canonical plan plus the row indices it owns and
+executes only those rows; contexts do not rebuild or accumulate a second mapping.
+Contexts run in a priority order that puts physics ahead of visuals, and the plan
+is published to :class:`~isaaclab.sim.SimulationContext` so the rest of the
 framework can read the per-env layout back.
-
-Collision Filtering
--------------------
-
-PhysX models per-env isolation through collision groups, so PhysX scenes need a
```
rilei-nvidia, line 404: Just wanted to check you meant to remove this block of documentation?

source/isaaclab/isaaclab/assets/visual_material/visual_material.py
```diff
@@ -0,0 +1,139 @@
...
+    def initialize(self) -> None:
+        """Compile stable device-write addresses after native asset cloning."""
+        entries = sorted(
+            (
+                env_id,
+                material_path := path.rebase(source_path, source_root, destination.format(env_id)),
+                material_path + shader_suffix,
+                input_name,
+                color,
+            )
```
rilei-nvidia: Minor suggestion - using a named tuple here can help readability.  In this case it's not so bad since it's consumed immediately after, but i've seen a couple of places around the code base with many-membered tuples being indexed by 0, 1, 2, etc and it can be confusing

source/isaaclab/isaaclab/assets/visual_material/visual_material.py
```diff
@@ -0,0 +1,139 @@
...
+    def write_colors(
+        materials: Sequence[VisualMaterial], colors: torch.Tensor, env_ids: torch.Tensor | None = None
+    ) -> None:
+        """Write colors for aligned materials and selected environments.
+
+        Materials must share clone coverage. ``colors`` has shape ``[M, E, 3]`` or
+        ``[M, K, 3]`` for selected ``env_ids``, which must belong to that coverage.
+        """
+        colors = colors.detach()
+        expected = (len(materials), len(materials[0]._env_ids) if env_ids is None else len(env_ids), 3)
```
rilei-nvidia: What are situations where env_ids is `None`?

source/isaaclab/isaaclab/assets/visual_material/visual_material_cfg.py
```diff
@@ -0,0 +1,35 @@
...
+
+if TYPE_CHECKING:
+    from .visual_material import VisualMaterial
+
+
+@configclass
+class VisualMaterialCfg:
+    """Configuration for a runtime-writable visual material scene entity.
+
+    Place the material below the cloned asset subtree containing every target. The scene
```
rilei-nvidia: @ooctipus I understand the rest of the paragraph, but can you clarify this part: "Place the material below the cloned asset subtree containing every target"

source/isaaclab/isaaclab/assets/visual_material/visual_material_cfg.py
```diff
@@ -0,0 +1,35 @@
...
+    authors one material and binding in each asset prototype, then the asset's native clone
+    operation carries them to its destination environments.
+    """
+
+    class_type: type["VisualMaterial"] | str = "{DIR}.visual_material:VisualMaterial"
+    """The associated scene-entity class."""
+
+    prim_path: str = MISSING
+    """Environment-scoped path below the target asset's cloned subtree."""
```
rilei-nvidia: Environment-scoped path below the target asset's cloned subtree **to spawn the material**

?

source/isaaclab/isaaclab/assets/visual_material/visual_material_cfg.py
```diff
@@ -0,0 +1,35 @@
...
+
+    spawn: PreviewSurfaceCfg | PbrMdlCfg | GlassMdlCfg = MISSING
+    """Supported spawner that authors the material source before cloning."""
+
+    target_prim_paths: tuple[str, ...] = MISSING
+    """Environment-scoped prim paths to bind to this material before cloning."""
```
rilei-nvidia: Do we need to mention the "before cloning" part in this class documentation?  I realize that it is what happens in our cloning implementation.  But to a user they only see the end result in that there is a material in every env bound to the target prim paths

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -48,19 +48,18 @@ class ClonePlan:
...
+    """Long tensor ``[num_clones]`` of target environment ids, aligned with the mask columns."""
 
-    Optional for plans used only with :func:`~isaaclab.cloner.query.iter_sources` or
-    :func:`~isaaclab.cloner.query.path_to_source`; required by :func:`~isaaclab.cloner.replicate`.
-    """
-
-    positions: torch.Tensor | None = None
-    """Per-env world positions [m], shape ``[num_clones, 3]``, or ``None``."""
+    env_poses: torch.Tensor
+    """Per-env root poses ``[x, y, z, qx, qy, qz, qw]``, shape ``[num_clones, 7]``."""
```
rilei-nvidia: Out of curiosity why did you need orientation introduced for this Visual material spawning?  Is this for memory layout alignment?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -46,6 +57,10 @@ def __init__(self) -> None:
         self._prepared_renderer_ids: set[int] = set()
         self._prepared_num_envs: int | None = None
         self._last_scene_state_step: int | None = None
+        self._visualizers: list[BaseVisualizer] = []
+        self._visual_material_color_batch: _VisualMaterialColorBatch | None = None
+        self._visual_material_writers: tuple[Callable[[], None], ...] = ()
+        self._is_complete = False
```
rilei-nvidia: _is_initialized?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -93,6 +110,84 @@ def ensure_initialize(self) -> None:
...
+        """Scene-owned material addresses and colors shared with rendering backends."""
+        return self._visual_material_color_batch
+
+    def set_visual_material_color_batch(self, batch: _VisualMaterialColorBatch | None) -> None:
+        """Publish the scene-owned material batch to rendering backends."""
+        if self._is_complete:
+            raise RuntimeError("Visual materials must be published before the first simulation reset.")
+        self._visual_material_color_batch = batch
+
+    def complete_initialization(self) -> None:
```
rilei-nvidia: There's a few methods on this class.  ensure_initialize, complete_initialization, initialize. To me they mean the same thing.  Can we clarify them

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -93,6 +110,84 @@ def ensure_initialize(self) -> None:
...
+        consumers = (*self._visualizers, *(renderer for _cfg, renderer in self._renderer_entries))
+        unsupported = [
+            type(consumer).__name__ for consumer in consumers if not consumer.supports_visual_material_color_writes
+        ]
+        if unsupported:
+            raise NotImplementedError(
+                f"Runtime visual-material color writes are not supported by {', '.join(unsupported)}. "
+                "Use Kit/Isaac RTX, OVRTX, Newton Warp, Newton GL, Rerun, or Viser."
+            )
+        writers = (*self._visual_material_writers, *(consumer.visual_material_writer for consumer in consumers))
```
rilei-nvidia: Out of curiousity - why does this class keep track of a separate list of visual material writers beyond the consumesr?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -93,6 +110,84 @@ def ensure_initialize(self) -> None:
...
+        written: set[int] = set()
+        for writer in writers:
+            if writer is None or id(writer) in written:
+                continue
+            writer()
+            written.add(id(writer))
+
+    def create_visual_shape_color_writer(self, asset: BaseArticulation, body_names: tuple[str, ...]):
+        """Create an articulation-link writer from the active rendering consumers."""
```
rilei-nvidia: Articulation-link?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -93,6 +110,84 @@ def ensure_initialize(self) -> None:
...
+        consumers = (*self._visualizers, *(renderer for _cfg, renderer in self._renderer_entries))
```
rilei-nvidia: consider a helper method _get_rendering_consumers (duplicate logic from above method)

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -93,6 +110,84 @@ def ensure_initialize(self) -> None:
...
+            raise NotImplementedError(
+                f"Per-shape visual color randomization is not supported by {', '.join(unsupported)}. "
+                "Declare one VisualMaterialCfg per independently colored link inside the cloned target-asset "
+                "subtree and use material-level randomization instead."
+            )
+        if not consumers:
+            raise NotImplementedError(
+                "Per-shape visual color randomization requires a Newton GL, Warp, Rerun, or Viser consumer."
+            )
+        return consumers[0].visual_shape_color_writer_factory(asset, body_names)
```
rilei-nvidia: It is OK to always pick the first consumer?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -181,12 +276,24 @@ def close(self) -> None:
...
                 logger.error("Error closing renderer %s: %s", type(renderer).__name__, exc)
                 errors.append(exc)
+        for writer in self._visual_material_writers:
+            close = getattr(writer, "close", None)
+            if callable(close):
+                try:
+                    close()
+                except Exception as exc:  # noqa: BLE001 - reported after all resources are closed
+                    logger.error("Error closing shared visual-material writer: %s", exc)
+                    errors.append(exc)
```
rilei-nvidia: Do we actually need this guard or should we let the error raise?

docs/source/how-to/cloning.rst
```diff
@@ -240,14 +240,68 @@ When envs need to differ across the population, use
...
+  scene entities. It samples one color per selected material and environment; every
+  render shape bound to the same material shares that color. Kit and Isaac RTX update
+  through Fabric, OVRTX uses its direct API, and Newton Warp, Newton GL, Rerun, and Viser
+  update the live Newton model. Newton RTX reads the material state when its scene is
+  built, but does not support reset-time material writes.
+* :class:`~isaaclab.envs.mdp.randomize_visual_shape` selects bodies on one asset. It
+  samples one color per selected body and environment, then writes every Newton render
+  shape on that body. Only Newton Warp, Newton GL, Rerun, and Viser consume this live
+  Newton shape-color buffer. For other renderers, declare one ``VisualMaterialCfg`` per
+  independently colored link inside the cloned target-asset subtree.
```
rilei-nvidia: So randomize_visual_shape is also for randomizing color?  When I read it at face value it sounds like you're randomizing the geometry (shape)

docs/source/how-to/cloning.rst
ooctipus: yes I folded another pr that remove this section to this pr

source/isaaclab/isaaclab/assets/visual_material/visual_material.py
ooctipus: noted will address in follow up pr

source/isaaclab/isaaclab/assets/visual_material/visual_material.py
ooctipus: when user wants apply to everything, they usually don't specify env_ids

source/isaaclab/isaaclab/assets/visual_material/visual_material_cfg.py
ooctipus: noted will address in follow up pr

docs/source/how-to/cloning.rst
ooctipus: thats actually a good call, randomize_visual_shape has been newton idiom to refers to visual attributes including color, maybe a better name is randomize_visual_properties or randomize_body_visual_attributes

source/isaaclab/isaaclab/renderers/render_context.py
ooctipus: yes probably remove these guard!

source/isaaclab/isaaclab/renderers/render_context.py
ooctipus: it is ok right now, since we only have one renderer consumer today

source/isaaclab/isaaclab/assets/visual_material/visual_material_cfg.py
ooctipus: yup we don't need!

pbarejko (approved):
Approved from perspective of stale golden images.

PR #6194, opened 2026-06-15
Propagate Newton shape colors before cloning

rilei-nvidia:
# Description

This PR integrates @ooctipus's proposed fix for newton shape color replacement to occur pre-clone.  Original PR: https://github.com/rilei-nvidia/IsaacLab/pull/1  

Replace the post-finalize `replace_newton_shape_colors` pass on the Newton model with `replace_newton_builder_shape_colors`, applied to each ModelBuilder before clone replication. This ensures all cloned environments automatically inherit the correct USD material and displayColor values.  This not only improves performance but also addresses the regression that @huidongc discovered last week where only env_0 (the clone source) had the right color for the robot arm, and the other cloned envs got the fallback colors

- Add replace_newton_builder_shape_colors to newton_model_utils.  
- Removed replace_newton_shape_colors and any un-used functionality
- Call it in NewtonManager (flat, env, and proto-source builders),
  newton_clone_utils, and replicate
- Update tests to exercise the new replace_newton_builder_shape_colors
- Update dexsuite_kuka_hetero golden images for correctness 

# Validation

Validated for correctness in the Newton visualizer with (all robot arms should have the same color):
```bash
./isaaclab.sh train --rl_library rsl_rl \
    --task Isaac-Lift-KukaAllegro-Camera \
    presets=newton_mjwarp,newton_renderer,rgb64,single_camera \
    --seed 42 \
    --num_envs 4 \
    --max_iterations 2 \
    --visualizer newton
```
image

Rendering test golden images for the dexsuite_kuka_hetero kitless suite have also been updated to reflect fix to the regression
```bash
./isaaclab.sh -p -m pytest -v -rA source/isaaclab_tasks/test/core/test_rendering_dexsuite_kuka_hetero_kitless.py::test_rendering_dexsuite_kuka_hetero_kitless[newton-newton_warp-rgb]
```
image

# Performance

Testing on a AMD Ryzen Threadripper PRO 7965WX 24-Cores (48) @ 5.36 GHz

For testing the performance impact I used the benchmark script:
```bash
./isaaclab.sh -p scripts/benchmarks/benchmark_non_rl.py \
    --task Isaac-Lift-KukaAllegro-Camera \
    presets=newton_mjwarp,ovrtx_renderer,rgb64,single_camera \
    --seed 42 \
    --num_envs "8000" \
    --benchmark_backend json \
    --output_path "hdc/benchmarks/20260611/heterogeneous_8000"
```

[...]

PR #6300, opened 2026-06-30
[STRESS / DO NOT MERGE] Validate Fabric GPU interop necessity on multi-GPU CI

hujc7:
## 1. Summary

- Diagnostic-only PR; do not merge.
- Validates whether `--/physics/fabricUseGPUInterop=false` is necessary for multi-GPU CI stability.
- Reuses the production multi-GPU container, shard runner, queue reconciliation, and test selection.

## 2. Experiment

- Run five negative trials first with the Fabric interop argument completely omitted.
- Run five controls with `--/physics/fabricUseGPUInterop=false`.
- Run all 17 currently eligible multi-GPU test files in every trial; ignore all `MULTI_GPU_SKIP_REASON` exclusions. The former 18th file, `test_newton_model_utils.py`, no longer has multi-GPU device parametrization after #6194.
- Hold the three renderer pinning arguments and all other workflow settings constant.
- Log the resolved Carb setting and preserve launcher logs, JUnit XML, aggregate summaries, and queue reconciliation for every trial.
- Gate unrelated Docker and installation test jobs so this PR only consumes test capacity for this diagnostic.

## 3. Decision criteria

- Evidence that the flag is needed: at least one negative trial reproduces the target crash, timeout, or queue-orphan failure while controls complete cleanly.
- An invalid setting resolution, incomplete file set, unclaimed test, or orphaned test invalidates that trial.
- If the first 5 + 5 trials are inconclusive, run another 5 + 5 before drawing a conclusion.


hujc7:
## Fabric GPU interop diagnostic result

The negative condition reproduced the multi-GPU CI instability.

- **Attempt 1 (5 omitted + 5 false):** all 10 final trials passed with 17/17 files, zero JUnit failures, zero unclaimed files, and zero orphans. Negative trial 4 had a 120 s `test_simulation_context.py` startup hang on its first attempt and passed only after the built-in retry. All five explicit-false controls were clean. Requested/resolved setting markers were correct in every trial.
- **Attempt 2 (same SHA/image/runner):** the first negative trial again omitted `/physics/fabricUseGPUInterop`, seeded all 17 files, then the entire A/B step was terminated after 526 s with **exit 143 (SIGTERM)**. Runner cleanup found and terminated the orphaned Docker process. There was no workflow cancellation, timeout, or lost-runner message. The termination prevented reconciliation, artifact upload, and the remaining trials.

**Conclusion:** keep `--/physics/fabricUseGPUInterop=false` in the multi-GPU CI lane. The exact negative condition produced one recovered startup hang and then one fatal signal termination, while the five same-SHA controls completed cleanly. This validates the flag as a necessary practical CI mitigation; it does not establish that Fabric interop is the sole root cause or justify disabling it globally.

Runs: [attempt 1](https://github.com/isaac-sim/IsaacLab/actions/runs/28424700108/attempts/1), [attempt 2](https://github.com/isaac-sim/IsaacLab/actions/runs/28424700108/attempts/2).

PR #6308, opened 2026-07-01
Add shared kinematic rigid-object renderer contract

ooctipus:
# Description

Revives #6308 on current `develop` and supersedes #3728 with a single backend-neutral contract for kinematic rigid-object rendering.

## Architecture

`source/isaaclab/test/renderers/rigid_object_rendering_contract.py` is the composition root. It owns the cloned scene, kinematic pose sequence, depth measurements, and assertions. Package-local adapters own only availability checks, simulation/renderer selection, and backend cleanup:

- Isaac RTX + PhysX on CPU and CUDA, with and without a coexisting articulation;
- Newton Warp + PhysX on CUDA;
- OVRTX + OVPhysX on CUDA through both legacy and OVStage scene ownership (OVStage runs when installed).

The dependency direction is adapter -> shared test contract -> public Isaac Lab APIs. An AST architecture gate rejects backend imports in the shared contract and rejects scene, asset, sensor, or class ownership in adapters.

The contract creates two cloned instanceable DexCubes with root-level nonuniform scale, verifies their depth silhouettes, moves both kinematic bodies through the public rigid-object tensor API, verifies the physics poses, and requires opposite rendered centroid displacement.

## Current-develop audit

Most production changes in the old PR have since landed through newer ownership boundaries: Isaac RTX render-product lifetime in #6729, Newton shadow-state copying in #6773, OVRTX scale-aware transform writes in #7010, and Newton Fabric scale preservation in #7481. This revival removes those stale patches rather than carrying duplicate implementations.

The revived contract exposed one remaining OVRTX bug: composed scale was captured only for clone-plan source paths, while OVRTX creates non-source destinations after exporting the host USD stage. Those destinations therefore defaulted to unit scale. As a deliberately temporary bridge, this PR projects only captured non-unit scales through the existing `isaaclab.cloner.query.path_env_ids` and `path_to_clone` boundary using the already-validated `ClonePlan`; real destination scales take precedence. It adds no plan fields, query APIs, renderer configuration, or per-body fallback, and the bridge can be deleted as one unit when SDP supplies composed scale aligned with canonical rigid-body paths.

Historical context: [Isaac Sim forum report](https://forums.developer.nvidia.com/t/rigidbody-prim-is-not-updated-in-rendering-pipeline-if-set-to-kinematic/346608).

## Type of change

- Bug fix
- Shared regression coverage

## Testing

- Isaac RTX contract: 4 passed (CUDA/CPU x articulation absent/present).
- Newton Warp contract: 1 passed.
- OVRTX contract: 2 passed (legacy and OVStage).
- OVRTX renderer unit surface: 167 passed.
- Core architecture and Newton visualization suites: 25 passed.
- OVRTX clone-plan suite: 17 passed.
- Cloner query and rendering-contract architecture suites: 87 passed.
- Controlled OVRTX regression: failed before the production fix with clone silhouettes of 264 vs. 36 pixels; passed after the fix.
- Incoming #7462 NumPy-backed `ClonePlan` query-boundary smoke check: passed unchanged, including non-dense environment ids.
- `uv run isaaclab -f`: all hooks passed against the exact upstream `develop` base, including changelog validation.

## Checklist

- [x] Documentation changes are not applicable
- [x] I have added unit and integration regression coverage


ooctipus:
run-ci

ooctipus:
I would make #6308 a deliberately temporary, deletable bridge—using today’s ClonePlan without adding any new plan fields or APIs.

  Use the existing query boundary and pass the already-required plan explicitly:

  from isaaclab.cloner import ClonePlan
  from isaaclab.cloner import query as clone_query


  def _capture_object_scales(self, stage: Any, plan: ClonePlan) -> None:
      ...
      for source_path, scale in tuple(self._object_scales_by_path.items()):
          for env_id in clone_query.path_env_ids(plan, source_path):
              clone_path = clone_query.path_to_clone(plan, source_path, env_id)
              assert clone_path is not None
              self._object_scales_by_path.setdefault(clone_path, scale)

  Call it only after prepare_stage() has validated the plan:

  self._capture_object_scales(stage, self._clone_plan)

  Why this transition is clean:

  - It works with the current Torch-backed plan and the incoming NumPy plan unchanged; /home/zhengyuz/Projects/IsaacLab.wt/northstar-pr3a/source/isaaclab/isaaclab/cloner/query.py:120 hide that representation.
  - Nested, heterogeneous, partial-environment, and non-dense layouts retain the cloner’s nearest-owner semantics.
  - setdefault() preserves a destination scale already captured from the real stage.
  - There is no silent missing-plan fallback.
  - It adds no ClonePlan.scales, ScaleMap, new query function, or renderer configuration.
  - The extra work remains sparse: scenes without non-unit scales perform no projection.

  I would not resolve every object path through path_to_source() inside _create_object_scale_array(). Although visually flatter, it queries every rigid body in every environment. A host-only 4,096×48-body check cost roughly one second, including completely unscaled bodies. Source-side
  projection only touches captured non-unit scales.

  For testing, keep the cross-backend rendered contract. Change the focused OVRTX unit test to assert the resulting scale array for source and destination bodies, rather than asserting the private dictionary’s exact contents. That lets the temporary cache disappear later without
  preserving its implementation.

  The later SDP cutover should replace the whole bridge:

  Plan completion: exact rigid-body paths + composed scales
  Physics:         native pose pointer + dirty
  OVRTX:           requests TransposedMatrix44d
  SDP:             pose + composed scale → matrix in one conversion
  OVRTX:           writes that pointer directly

  Importantly, future FrameLayout.scale currently records only local xformOp:scale; the final conversion needs composed scale aligned with the canonical rigid-body paths. Once that exists, delete _capture_object_scales, _object_scales_by_path, and the destination projection together.

  That version can merge before or after #7462 and should transition without another mechanical rewrite.

ooctipus:
run-ci

ooctipus:
run-ci

kellyguo11 (approved):

PR #6321, opened 2026-07-02
Raise instead of swallowing OVRTXRenderer failures

rilei-nvidia:
# Description

Several methods in OVRTXRenderer (scene/camera/object binding setup, scene partition writes, and render) caught broad exceptions and only logged a warning, letting simulation or training continue on a renderer failure that should have stopped it. Let these propagate so callers can decide how to handle the failure, matching the existing pattern already used by _clone_environments.

Object binding setup for Newton now explicitly skips when no SimulationContext is active yet, rather than crashing on an assertion deep in NewtonManager.

## Type of change

- Refactor to fail hard on real errors

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -584,7 +584,8 @@ class OVRTXRenderer(BaseRenderer):
         from isaaclab_newton.physics import NewtonManager
 
         newton_state = NewtonManager.get_state()
-        assert newton_state is not None, "Newton state should not be None"
+        if newton_state is None:
+            raise RuntimeError("Newton state should not be None")
 
         body_q = getattr(newton_state, "body_q", None)
         if body_q is None:
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -368,64 +368,62 @@ class OVRTXRenderer(BaseRenderer):
             raise RuntimeError(f"Expected camera prim under '{env_0_prefix}', got '{first_cam_path}'")
         self._camera_rel_path = spec.camera_path_relative_to_env_0
 
-        if self._exported_usd_string is not None:
-            logger.info("Injecting camera definitions...")
+        logger.info("Injecting camera definitions...")
 
-            render_product_string, render_product_path = build_render_product_as_string(
-                width=width,
-                height=height,
-                num_envs=num_envs,
-                data_types=data_types,
-                minimal_mode=_resolve_rtx_minimal_mode(data_types),
-                camera_rel_path=self._camera_rel_path,
-            )
-            self._render_product_paths.append(render_product_path)
+        assert self._exported_usd_string is not None, "Expected an exported USD string from stage"
 
-            combined_usd_string = self._exported_usd_string + "\n\n" + render_product_string
-            self._exported_usd_string = None  # Free memory
+        render_product_string, render_product_path = build_render_product_as_string(
+            width=width,
+            height=height,
+            num_envs=num_envs,
+            data_types=data_types,
+            minimal_mode=_resolve_rtx_minimal_mode(data_types),
+            camera_rel_path=self._camera_rel_path,
+        )
+        self._render_product_paths.append(render_product_path)
 
-            # If temp_usd_dir is set, write the combined USD stage to a temporary file.
-            if self.cfg.temp_usd_dir is not None:
-                _write_file(Path(self.cfg.temp_usd_dir), "ovrtx_renderer_stage.usda", combined_usd_string)
+        combined_usd_string = self._exported_usd_string + "\n\n" + render_product_string
+        self._exported_usd_string = None  # Free memory
 
-            logger.info("Loading USD into OvRTX...")
-            try:
-                self._renderer.open_usd_from_string(combined_usd_string)
-                logger.info("OVRTX loaded USD from string successfully")
-            except Exception as e:
-                logger.exception("Error loading USD: %s", e)
-                raise
+        # If temp_usd_dir is set, write the combined USD stage to a temporary file.
+        if self.cfg.temp_usd_dir is not None:
+            _write_file(Path(self.cfg.temp_usd_dir), "ovrtx_renderer_stage.usda", combined_usd_string)
 
-            if num_envs > 1:
-                self._clone_sources_in_ovrtx()
-                self._update_scene_partitions_after_clone(num_envs)
+        logger.info("Loading USD into OvRTX...")
+        try:
+            self._renderer.open_usd_from_string(combined_usd_string)
+            logger.info("OVRTX loaded USD from string successfully")
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -361,64 +361,62 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
...
-                data_types=data_types,
-                minimal_mode=_resolve_rtx_minimal_mode(data_types),
-                camera_rel_path=self._camera_rel_path,
-            )
-            self._render_product_paths.append(render_product_path)
+        logger.info("Injecting camera definitions...")
 
-            combined_usd_string = self._exported_usd_string + "\n\n" + render_product_string
-            self._exported_usd_string = None  # Free memory
+        assert self._exported_usd_string is not None, "Expected an exported USD string from stage"
```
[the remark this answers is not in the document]
rilei-nvidia: The assertion is enough here.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -361,64 +361,62 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
...
+        if self.cfg.temp_usd_dir is not None:
+            _write_file(Path(self.cfg.temp_usd_dir), "ovrtx_renderer_stage.usda", combined_usd_string)
 
-            self._initialized_scene = True
+        logger.info("Loading USD into OvRTX...")
+        try:
+            self._renderer.open_usd_from_string(combined_usd_string)
+            logger.info("OVRTX loaded USD from string successfully")
+        except Exception as e:
+            logger.exception("Error loading USD: %s", e)
```
pbarejko: Why do we capture this? The logged exception isn't very helpful?

rilei-nvidia: Removed try except

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -361,64 +361,62 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
...
+        self._renderer.write_attribute(
+            prim_paths=camera_paths,
+            attribute_name="omni:resetXformStack",
+            tensor=np.full(num_envs, True, dtype=np.bool_),
+        )
+
+        if self._camera_binding is not None:
+            logger.info("Camera binding created successfully")
+        else:
+            logger.warning("Camera binding is None")
```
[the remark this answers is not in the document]
rilei-nvidia: @pbarejko / @huidongc - is there _always_ going to be at least one camera in each of the environments?  If so, we can assert that the camera binding must be successful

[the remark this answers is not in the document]
pbarejko: That's what I would assume, we can't do sensor simulation without camera.

pbarejko (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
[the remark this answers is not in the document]
huidongc: From my perspective, the point raised by the review bot is legit. If the optimization flag is used then assert will be ineffective.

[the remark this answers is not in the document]
huidongc: I'd raise an exception inside `prepare_stage()` if `self._exported_usd_string` is null, so that errors can be reported as early as possible.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -461,76 +455,71 @@ def _update_scene_partitions_after_clone(self, num_envs: int):
...
-        except Exception as e:
-            logger.warning("Error setting up object bindings: %s", e)
+        self._renderer.write_attribute(
+            prim_paths=object_paths,
+            attribute_name="omni:resetXformStack",
+            tensor=np.full(len(object_paths), True, dtype=np.bool_),
+        )
+
+        if self._object_binding is None:
+            raise RuntimeError("Failed to create OVRTX object bindings for Newton")
```
huidongc: The error message "for Newton" is probably inaccurate.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -579,26 +568,26 @@ def update_transforms(self) -> None:
...
-            from isaaclab_newton.physics import NewtonManager
+        # If self._object_newton_indices is not None, then Newton's the current physics backend
 
-            newton_state = NewtonManager.get_state()
-            if newton_state is None:
-                return
-            body_q = getattr(newton_state, "body_q", None)
-            if body_q is None:
-                return
+        from isaaclab_newton.physics import NewtonManager
```
huidongc, line 582: I think the import here should still be wrapped with `try/except` because ovrtx may be combined with other sim backend (not Newton backend).

huidongc (changes requested):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: I thought that is what the above self._object_newton_indices is None: guards against?  This method seems specific to the newton use case.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Is it not?  I thought this entire method is to create object bindings only when the physics back-end is newton.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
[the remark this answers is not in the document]
rilei-nvidia: Applied!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: The message appears like we are doing it "for Newton". Newton produces the data after physics simulation and here the function passes the simulation data to the render backend.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Aah, okay!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Ah got it - will change wording

rilei-nvidia: Removed the "for Newton" part

huidongc (approved):

PR #6248, opened 2026-06-23
[DO NOT MERGE] Add articulation ordering controls

AntoineRichard:
# Description

Adds major cross-backend articulation ordering support so public joint/body order can be kept independent from the active backend's native storage order.

This introduces optional `ArticulationCfg.joint_ordering` and `ArticulationCfg.body_ordering` controls, shared name-map utilities, and Warp reorder kernels. Newton, PhysX, and OVPhysX articulation data paths are wired to translate between public and backend order for reads and writes. Built-in symbolic conventions currently include `physx` and `mjwarp`, with automatic cross-backend name resolution.

Fixes # N/A

## Type of change

- New feature (non-breaking change which adds functionality)
- Major change (cross-backend articulation API and data-path update; changelog fragments use `.major.rst`)

## Screenshots

N/A. Validation included local playback videos for Anymal-D and Franka policies under alternate ordering settings.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/assets/test_articulation_ordering.py source/isaaclab_newton/test/assets/test_articulation.py::test_mjwarp_ordering_resolver_matches_newton_backend_names -q` -> 26 passed
- [x] `./isaaclab.sh -d` -> documentation build succeeded; reported 6 warnings from `docs/superpowers/plans/2026-06-23-articulation-ordering.md`
- [x] `./isaaclab.sh -f` -> all hooks passed
- [x] Trained Newton/MJWarp Anymal-D flat policy with 4096 envs for 300 iterations; replayed in Newton/PhysX ordering variants
- [x] Trained Newton/MJWarp Franka reach joint-position policy with 4096 envs for 300 iterations; replayed with default and PhysX articulation ordering

PR #6335, opened 2026-07-03
Articulation Reordering Part 2/8: Core ordering foundation

AntoineRichard:
# Description

Introduces the public joint/body ordering machinery in `isaaclab` core: the `ArticulationNameMap` type and builder, the ordering preset helper, the Warp reorder kernels with an explicit visibility contract, the symbolic convention resolvers (`physx`, `mjwarp`, `robot_schema`), the `ArticulationCfg.joint_ordering`/`body_ordering` fields, the `BaseArticulation`/`BaseArticulationData` plumbing that backends wire into, and the API reference entries.

**Transitional semantics:** no backend consumes the machinery yet, so configuring an ordering in this intermediate state is a silent no-op until Parts 3/4/6 land. This is called out in the cfg field docstrings and resolves quickly as the stack merges.

Part 2/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- **Part 2/8 (this PR)**: #6335 — Core ordering foundation

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p1...antoiner%2Farticulation-reordering-p2).

## Type of change

## Checklist

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/assets/test_articulation_ordering.py` -> 67 passed
- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/assets/test_articulation_ordering_kernels.py` -> 24 passed

### DIFF
--- a/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
+++ b/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
@@ -8,6 +8,7 @@
 
 from __future__ import annotations
 
+import logging
 import warnings
 from abc import abstractmethod
 from collections.abc import Sequence
@@ -29,6 +30,8 @@ if TYPE_CHECKING:
 
     from .articulation_cfg import ArticulationCfg
     from .base_articulation_data import BaseArticulationData
+
+logger = logging.getLogger(__name__)
 
 
 class BaseArticulation(AssetBase):
@@ -248,9 +251,10 @@ class BaseArticulation(AssetBase):
     def joint_ordering(self) -> ArticulationNameMap | None:
         """Bidirectional map between backend and public joint order.
 
-        The map is ``None`` for the default ``None`` configuration, a non-``None``
-        identity map when an explicit order resolves to backend order, and a
-        nonidentity map for any actual permutation.
+        The map is ``None`` whenever the public and backend orders coincide:
+        either no ordering is configured, or the configured ordering resolved
+        to the backend's native order. A non-``None`` map always denotes an
+        actual permutation.
         """
         return getattr(self.data, "joint_ordering", None)
 
@@ -258,9 +262,10 @@ class BaseArticulation(AssetBase):
     def body_ordering(self) -> ArticulationNameMap | None:
         """Bidirectional map between backend and public body order.
 
-        The map is ``None`` for the default ``None`` configuration, a non-``None``
-        identity map when an explicit order resolves to backend order, and a
-        nonidentity map for any actual permutation.
+        The map is ``None`` whenever the public and backend orders coincide:
+        either no ordering is configured, or the configured ordering resolved
+        to the backend's native order. A non-``None`` map always denotes an
+        actual permutation.
         """
         return getattr(self.data, "body_ordering", None)
 
@@ -273,9 +278,9 @@ class BaseArticulation(AssetBase):
         that pick joints with public indices (for example event terms) must
         convert those indices before addressing backend arrays.
 
-        When :attr:`joint_ordering` is ``None`` or an identity permutation, the
-        public and backend orders coincide and :paramref:`joint_ids` is returned
-        unchanged without any per-index lookup.
+        When :attr:`joint_ordering` is ``None`` the public and backend orders
+        coincide and :paramref:`joint_ids` is returned unchanged without any
+        per-index lookup.
 
         Args:
             joint_ids: Joint indices in public :attr:`joint_names` order.
@@ -285,7 +290,7 @@ class BaseArticulation(AssetBase):
             order, or :paramref:`joint_ids` unchanged when the orders coincide.
         """
         ordering = self.joint_ordering
-        if ordering is None or ordering.is_identity:
+        if ordering is None:
             return joint_ids
         return [ordering.user_to_backend_indices[joint_id] for joint_id in joint_ids]
 
@@ -298,9 +303,9 @@ class BaseArticulation(AssetBase):
         that pick bodies with public indices (for example event terms) must
         convert those indices before addressing backend arrays.
 
-        When :attr:`body_ordering` is ``None`` or an identity permutation, the
-        public and backend orders coincide and :paramref:`body_ids` is returned
-        unchanged without any per-index lookup.
+        When :attr:`body_ordering` is ``None`` the public and backend orders
+        coincide and :paramref:`body_ids` is returned unchanged without any
+        per-index lookup.
 
         Args:
             body_ids: Body indices in public :attr:`body_names` order.
@@ -310,7 +315,7 @@ class BaseArticulation(AssetBase):
             or :paramref:`body_ids` unchanged when the orders coincide.
         """
         ordering = self.body_ordering
-        if ordering is None or ordering.is_identity:
+        if ordering is None:
             return body_ids
         return [ordering.user_to_backend_indices[body_id] for body_id in body_ids]
 
@@ -330,7 +335,12 @@ class BaseArticulation(AssetBase):
         raise NotImplementedError()
 
     def _resolve_and_install_ordering_maps(self) -> None:
-        """Resolve configured articulation name orderings and store maps on :attr:`data`."""
+        """Resolve configured articulation name orderings and store maps on :attr:`data`.
+
+        Orderings that resolve to the backend's native order are normalized to
+        ``None`` here, so an installed map always denotes an actual permutation
+        and ``is not None`` checks alone decide whether reordering is active.
+        """
         had_ordering = self.data.joint_ordering is not None or self.data.body_ordering is not None
 
         if self.cfg.joint_ordering is None:
@@ -344,13 +354,21 @@ class BaseArticulation(AssetBase):
                 active_backend_name=self.__backend_name__,
                 articulation=self,
             )
-            self.data.joint_ordering = build_articulation_name_map(
+            joint_ordering = build_articulation_name_map(
                 kind="joint",
                 backend_names=self.backend_joint_names,
                 user_names=joint_user_names,
                 device=self.device,
             )
-            self.data.joint_names = list(self.data.joint_ordering.user_names)
+            if joint_ordering.is_identity:
+                logger.info(
+                    "Configured joint_ordering for '%s' resolves to the backend's native order; no reordering"
+                    " will be applied.",
+                    self.cfg.prim_path,
+                )
+                joint_ordering = None
+            self.data.joint_ordering = joint_ordering
+            self.data.joint_names = list(joint_user_names)
 
         if self.cfg.body_ordering is None:
             self.data.body_ordering = None
@@ -380,8 +398,15 @@ class BaseArticulation(AssetBase):
                         f"{requested_index}. Put '{root_body_name}' first; all remaining bodies may be reordered "
                         "freely."
                     )
+            if body_ordering.is_identity:
+                logger.info(
+                    "Configured body_ordering for '%s' resolves to the backend's native order; no reordering"
+                    " will be applied.",
+                    self.cfg.prim_path,
+                )
+                body_ordering = None
             self.data.body_ordering = body_ordering
-            self.data.body_names = list(body_ordering.user_names)
+            self.data.body_names = list(body_user_names)
 
         has_ordering = self.data.joint_ordering is not None or self.data.body_ordering is not None
         if had_ordering or has_ordering:
--- a/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
+++ b/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
@@ -2435,12 +2435,13 @@ class BaseArticulation(AssetBase):
     def _joint_user_to_backend_map(self) -> wp.array:
         """Device public-to-backend joint map, or the identity map when no ordering is active.
 
-        The identity fallback reads the backend-allocated ``_ALL_JOINT_INDICES``
-        index buffer (the same backend-owned contract as
-        :attr:`_ordering_joint_staging_names`), so unconditional launch sites --
-        for example in-graph publish kernels that always gather -- get a valid map
-        either way. Conditional sites should read :attr:`data.joint_ordering`
-        directly and skip the launch when it is ``None``.
+        The identity fallback reads ``_ALL_JOINT_INDICES``, an index buffer that
+        backends allocate without a base-class default (a backend-owned contract,
+        like the staging attributes named by :attr:`_ordering_joint_staging_names`),
+        so unconditional launch sites -- for example in-graph publish kernels that
+        always gather -- get a valid map either way. Conditional sites should read
+        :attr:`~BaseArticulationData.joint_ordering` directly and skip the launch
+        when it is ``None``.
         """
         ordering = self.data.joint_ordering
         return ordering.user_to_backend if ordering is not None else self._ALL_JOINT_INDICES
@@ -2506,7 +2507,7 @@ class BaseArticulation(AssetBase):
 
         Avoid this modify-then-reorder helper on per-step hot paths: it costs a
         full extra kernel launch over the buffer. Prefer the fused
-        ``write_*_user_to_backend`` kernels from
+        ``write_*_user_to_backend`` write paths from
         :mod:`isaaclab.assets.articulation.ordering_kernels`, which write the
         public and backend buffers in a single launch. This helper exists for
         infrequent property writers where the extra launch is irrelevant.
--- a/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
+++ b/source/isaaclab/isaaclab/assets/articulation/base_articulation.py
@@ -220,7 +220,14 @@ class BaseArticulation(AssetBase):
         The inherited compatibility fallback emits :class:`DeprecationWarning`
         and returns :attr:`joint_names`. A subclass relying on that fallback
         therefore receives public order and cannot expose a distinct solver order.
+
+        Raises:
+            NotImplementedError: If the subclass overrides neither
+                :attr:`joint_names` nor this property, since the two inherited
+                fallbacks delegate to each other and cannot produce names.
         """
+        if type(self).joint_names is BaseArticulation.joint_names:
+            raise NotImplementedError(f"{type(self).__name__} must override joint_names or backend_joint_names.")
         warnings.warn(
             f"{type(self).__name__} must override backend_joint_names before it becomes abstract in a future release.",
             DeprecationWarning,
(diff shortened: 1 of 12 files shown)


AntoineRichard:
Design change from review discussion: **identity orderings now normalize to `None` at the install choke point** (4b4c2f91d48). Previously a configured ordering that resolved to the backend's native order installed an identity `ArticulationNameMap`, giving consumers two sentinels for the same no-op state (`None` vs `is_identity`). Now: `joint_ordering`/`body_ordering` `is not None` ⇔ `_has_*_ordering` ⇔ reordering actually happens — one sentinel, checked with plain `is None`. `ArticulationNameMap.is_identity` and the total builder remain (identity maps stay legal values of the *type*, e.g. hand-built in tests); only assets normalize them away, with an info log recording the resolution. Later parts get their small ripple (guide examples, one events test) when the stack rebases after this review.

AntoineRichard:
Addressed the ordering.py review points in 18ef656afcc — all ten drafts trace to one root cause (a coercion layer serving inputs we control), now deleted:

- **`coerce` naming** → renamed `_validate_articulation_names` / `_validate_articulation_indices`; they validate and return the canonical tuple.
- **Defensive layers** (str/bytes/bytearray special-cases, `try: tuple(iterable)`) → gone; hard `type(x) in (list, tuple)` check, per your one-contract call: list-or-tuple at the builder/cfg boundary (canonicalized to tuple), exact `tuple` on the frozen dataclass fields.
- **`operator.index` + the bool branch** → gone. Bools only needed rejecting because `operator.index(True)` succeeds; with exact `type(v) is int`, `True` fails the ordinary check (bool is not int by exact type) with the standard message. NumPy ints are now rejected too — typed inputs exist for a reason.
- **`object.__setattr__`** → existed solely to write coerced values back into frozen fields; validation-only means no mutation, so it's deleted outright.
- Element checks are now exact (`type(name) is str`), one line each.

Net: both helpers are ~7 lines, `_validate` no longer reassigns anything, and the tests were flipped to pin the strict contract (generic sequences and NumPy integral types now assert rejection; fail-then-pass preserved). 68/68 green, pre-commit clean. Your pending drafts should all be resolvable.

source/isaaclab/isaaclab/assets/articulation/ordering.py
```diff
@@ -0,0 +1,369 @@
...
+    ``None`` selects the active backend order by default and is not a member of
+    this enum.
+    """
+
+    PHYSX = "physx"
+    MJWARP = "mjwarp"
+    ROBOT_SCHEMA = "robot_schema"
+
+
+@dataclass(frozen=True)
```
ooctipus, line 66: Could we keep ArticulationNameMap private and make it represent only a real permutation?

Condensing the current representation, it stores:

    kind
    backend_names / user_names
    two CPU permutations
    two device permutations
    is_identity

Several fields duplicate state already present on the articulation, while identity maps are immediately normalized to None. Public construction also requires a large validation surface, including copying the Warp maps back to the host, and frozen dataclass semantics do not make the Warp arrays immutable.

Could this instead be factory-controlled and private?

    @dataclass(frozen=True, slots=True)
    class _AxisPermutation:
        public_to_backend_indices: tuple[int, ...]
        public_to_backend: wp.array
        backend_to_public: wp.array

    def _make_axis_permutation(...) -> _AxisPermutation | None:
        if public_names == backend_names:
            return None
        ...

The existing joint_names/body_names remain the public result, and map_joint_ids_to_backend/map_body_ids_to_backend can provide the raw-view escape hatch without exposing the representation. This leaves one identity state, removes duplicated names and is_identity, and avoids committing to a public constructor and its direct-construction tests before an external use case needs it.

source/isaaclab/isaaclab/assets/articulation/ordering_kernels.py
```diff
@@ -0,0 +1,870 @@
...
+        backend_data: Destination array shaped [num_envs, num_items] in backend
+            order, with the same units as user_data. It must not alias user_data
+            for a nonidentity permutation.
+    """
+    env_id, backend_id = wp.tid()
+    user_id = backend_to_user[backend_id]
+    backend_data[env_id, backend_id] = user_data[env_id, user_id]
+
+
+@wp.kernel
```
ooctipus, line 72: Could we separate the fundamental permutation/scatter primitives from semantic and fused kernels, and see whether more of the writer family can share generalized primitives?

The current module grows through a cross-product of concerns:

    reorder_2d / reorder_3d
    indices / masks
    scalar / array inputs
    full / compact inputs
    ordinary writes / velocity history / complete joint state
    dtype-specific Torch overloads

I agree that some specializations are real. Jacobians and mass matrices permute multiple axes while preserving floating-base DoFs, and joint-state writes must update velocity history and acceleration. I am not suggesting forcing those into one magical generic kernel.

For the remaining property writers, could the layering be closer to:

    # generalized private primitives
    _reorder_2d_*
    _reorder_3d_*
    _scatter_2d_with_indices_or_mask
    _scatter_3d_with_indices_or_mask

    # semantic/fused kernels only where required or measured hot
    _write_joint_state_and_history
    _reorder_joint_targets
    _reorder_jacobian
    _reorder_mass_matrix

For example, normalizing Torch/Warp input representation once in the Python wrapper may let the generic kernels handle more dtypes without a separate public overload name for every call shape. Where index and mask launch shapes prevent useful unification, keeping two concrete variants is still clearer than adding abstraction for its own sake.

Since this foundation PR has no backend call sites yet, another option is to introduce each semantic/fused kernel with its first consumer while keeping it physically in core when it is shared. That would let us review the axis rules and any fusion/performance justification beside the launch site, and would substantially reduce the review surface here without discarding legitimate optimized kernels.

ooctipus: actually I am not knowledge enough to comment on what is the best practice on Kernel organization, @StafaH can you comment?

source/isaaclab/isaaclab/assets/articulation/ordering.py
hujc7: Suggest also storing the result only on `ArticulationData` (asset delegates) and collapsing the install into one factory call resolving both axes in one discovery pass:

```python
data.joint_ordering, data.body_ordering = ArticulationOrdering.resolve(
    cfg, backend_names, device, is_fixed_base, staging_specs)
```

This removes `_install_ordering_flags` / `_cache_ordering_maps` / `_reset_and_cache_ordering_maps` / `_ordering_maps_cached` and the mirrored `_has_*`/map attributes on both classes — one source of truth, so the copies can't disagree.

source/isaaclab/isaaclab/assets/articulation/ordering_kernels.py
hujc7: One concrete reduction: the direction pairs are the same kernel body — keep one gather per rank and name the map by its contract:

```python
def gather_2d(src, dst_to_src, dst):
    env_id, dst_id = wp.tid()
    dst[env_id, dst_id] = src[env_id, dst_to_src[dst_id]]
```

`reorder_2d_backend_to_user` / `reorder_2d_user_to_backend` (and the 3d pair) become aliases of this; `dst_to_src` makes the map direction unambiguous by construction.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -0,0 +1,833 @@
...
+
+_ROBOT_SCHEMA_RELATIONSHIP_NAMES: dict[Literal["joint", "body"], str] = {
+    "joint": "isaac:physics:robotJoints",
+    "body": "isaac:physics:robotLinks",
+}
+
+_ROBOT_SCHEMA_NAME_OVERRIDE_ATTRS: dict[Literal["joint", "body"], tuple[str, ...]] = {
+    "joint": ("isaac:NameOverride", "isaac:nameOverride"),
+    "body": ("isaac:nameOverride", "isaac:NameOverride"),
+}
```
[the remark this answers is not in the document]
AntoineRichard: Good catch — and digging into it, the divergence is worse than a style issue: neither ordering was authoritative. The Isaac Sim robot schema (`isaacsim.robot.schema`) defines exactly one spelling, `isaac:nameOverride` (`RobotSchema.usda`: `string isaac:nameOverride = ""`); the capitalized `isaac:NameOverride` does not exist in any schema — it was speculative tolerance introduced here. Fixed in cb749f4ae15 by reading only the schema-defined attribute for both joints and bodies, which removes the dual-spelling ambiguity entirely rather than just unifying the priority.

source/isaaclab/isaaclab/assets/articulation/ordering.py
AntoineRichard: Largely agree, and adopted most of this in 7b0c9db5c34: `build_articulation_name_map` is now the only construction path and returns `None` for identity (your `_make_axis_permutation | None` shape), so the direct-construction validation surface — `__post_init__`, `_prevalidated`, the device-to-host comparison copies, and their tests — is gone. `kind`, `user_names`/`backend_names`, and `is_identity` are dropped from the type (names live on `data.joint_names`/`backend_joint_names`; identity is represented by `None`), leaving exactly the two CPU tuples and two device permutations.

Two deliberate deviations: (1) the type itself stays public (not underscored) because it is the return type of the public `articulation.joint_ordering`/`body_ordering` properties, and raw-view interop — launching the public gather kernels on solver-view arrays — needs read access to the device maps; `map_*_ids_to_backend` covers index translation but not device-side reordering. It is now a read-only result type with no public constructor (the builder is also de-exported from `isaaclab.assets`). (2) Kept the `user_*` field naming over `public_*`: it matches every kernel parameter and docstring across the series, and the rename would touch dozens of files for no semantic change.

source/isaaclab/isaaclab/assets/articulation/ordering_kernels.py
AntoineRichard: Adopted the wrapper idea in 0c0dae23c45 — it is strictly better than the overload table: each generic `write_*` kernel now has a thin Python launch wrapper that adapts torch inputs once (`wp.from_torch` with explicit dtype, including the vec3/transform trailing-dim folding) and derives the launch dim, so the 8 dtype-suffixed overload names are gone and, since call sites go through wrappers, the generic kernels are now properly module-private (`_write_*`) — which also resolves the awkward "non-underscored but internal" caveat in the visibility contract. hujc7's gather unification is folded in as well (54f10b3f0d3).

On relocating the fused kernels to their first-consumer PRs: pushing back there. The kernels plus their unit tests are a self-contained reviewable unit here, with the axis conventions documented in one place; splitting them across the backend PRs would fragment those conventions and put core-package churn inside backend diffs that are already inflated by fork-stacking. The backend PRs (Parts 3/4/6) show every launch site if you want to review the fusion justification beside its consumer. Happy to revisit the file organization if @StafaH lands a different convention — at that point it is a mechanical reshuffle.

source/isaaclab/isaaclab/assets/articulation/ordering.py
AntoineRichard: Agreed — implemented in 321e9b3410d. The maps live only on `ArticulationData` (`joint_ordering`/`body_ordering`), the asset properties delegate, and `_install_ordering_flags` / `_cache_ordering_maps` / `_reset_and_cache_ordering_maps` / `_ordering_maps_cached` plus every mirrored `_has_*`/map attribute on both classes are deleted; read paths check `data.joint_ordering is not None` directly (one attribute hop per setter call — noise next to the kernel launch). The install-once guard went with the machinery: it only existed because the mirrors could go stale, and with one home there is nothing to desynchronize.

Two small deviations from the sketch: resolution stays two per-axis calls behind the one install site (joint and body discovery read different USD relationships, so a fused pass shares nothing), and backend staging keeps its existing allocation hook, now invoked directly by backends when they (re)create buffers instead of through the cache guard.

source/isaaclab/isaaclab/assets/articulation/ordering_kernels.py
AntoineRichard: Done in 54f10b3f0d3 — one `gather_2d`/`gather_3d` per rank with the map named `dst_to_src` (unambiguous by construction, as you say), and the four directional names kept as aliases of the same kernel objects so launch sites keep reading as intent (`reorder_2d_backend_to_user`) and no call site or test changes. The alias block documents which map to pass for each direction.

AntoineRichard:
Addressed this review round as six commits:

- cb749f4ae15 — robot-schema name override: read only the schema-defined `isaac:nameOverride` (greptile)
- 54f10b3f0d3 — `gather_2d`/`gather_3d` unification with directional aliases (@hujc7)
- 0c0dae23c45 — write-kernel launch wrappers replace the 8 dtype-suffixed overloads; generic kernels are now module-private (@ooctipus)
- 7b0c9db5c34 + 321e9b3410d — `ArticulationNameMap` slimmed to a factory-built permutation type (builder returns `None` for identity, de-exported) and single-homed on `ArticulationData`; the flag/map mirrors and cache-guard machinery are gone (@ooctipus, @hujc7)
- 932fe0e18a3 — docstring clarifications on the new map helpers

Net −130 lines. Replies with specifics — including the two partial declines (type stays publicly readable for raw-view interop; fused kernels stay in this PR) and why — are on the threads.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
```diff
@@ -162,21 +198,212 @@ def spatial_tendon_names(self) -> list[str]:
...
+        The order follows :attr:`ArticulationCfg.body_ordering` when configured
+        and otherwise matches :attr:`backend_body_names`.
+        """
+        ordering = self.data.body_ordering
+        if ordering is not None:
+            return list(self.data.body_names)
+        return list(self.backend_body_names)
+
+    @property
+    def backend_joint_names(self) -> list[str]:
```
hujc7, line 217: **Agent-found bug:** if a backend overrides neither `joint_names` nor `backend_joint_names`, the two fallbacks recurse into each other (previously the abstract property failed at instantiation). Suggest guarding the fallback:

```python
if type(self).joint_names is BaseArticulation.joint_names:
    raise NotImplementedError("override joint_names or backend_joint_names")
```

Same for the body pair.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
```diff
@@ -162,21 +198,212 @@ def spatial_tendon_names(self) -> list[str]:
...
+        """Bidirectional map between backend and public body order.
+
+        The map is ``None`` whenever the public and backend orders coincide:
+        either no ordering is configured, or the configured ordering resolved
+        to the backend's native order. A non-``None`` map always denotes an
+        actual permutation.
+        """
+        return self.data.body_ordering
+
+    def map_joint_ids_to_backend(self, joint_ids: Sequence[int]) -> Sequence[int]:
```
hujc7: **Agent-found bug:** `slice(None)` passes through under identity ordering but raises `TypeError` under a real permutation — a config-dependent split on the common "all joints" selector. Suggest handling `slice` on both paths, plus a 3-cycle test for both methods (currently untested; a swapped-direction implementation would pass the suite).

source/isaaclab/isaaclab/assets/articulation/ordering.py
hujc7: A preference note, no behavior concern: I personally find this easier to follow when the functions live with the data — state is now on `data.joint_ordering`/`body_ordering`, but resolve/install and the staging/reorder helpers sit on `BaseArticulation`, gather paths on `BaseArticulationData`, so reading "ordering" still spans two classes and three modules. Was keeping them separate deliberate (asset context? churn in parts 3–8)?

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
AntoineRichard: Good catch, thanks — fixed in 3da1e81164a with exactly your guard, on both axes, placed ahead of the deprecation warning so a backend overriding neither gets `NotImplementedError` instead of `RecursionError` (and a backend overriding only the public property keeps the deprecated-fallback behavior unchanged). The regression test asserts the error from all four entry points (`joint_names`, `body_names`, and both `backend_*` fallbacks) on a subclass that overrides neither.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
AntoineRichard: Fixed in 3da1e81164a. The permutation path now expands slices through tuple slicing (`ordering.user_to_backend_indices[joint_ids]`), which handles any slice — start/stop/step — not just `slice(None)`; the identity path keeps returning the selector unchanged, so both paths accept the common "all joints" selector. Also added the 3-cycle tests you suggested for both methods (`user_to_backend == (1, 2, 0)` vs `backend_to_user == (2, 0, 1)`, so a direction-swapped implementation fails), plus slice coverage on both paths.

source/isaaclab/isaaclab/assets/articulation/ordering.py
AntoineRichard: Deliberate, though the readability cost is real. The split follows what each piece needs and which buffers it fills: resolution/install sits on the asset because it consumes asset-scope context (cfg, the backend-overridable `backend_*_names`, `__backend_name__`, `is_fixed_base`, and — for the robot-schema/USD resolvers — the articulation object itself plus its convention cache); the staging configure lives there because the write-path staging buffers are asset attributes feeding `write_data_to_sim`; the gather paths live on data because they fill data-owned timestamped caches. Moving resolve onto data would just mean passing the asset in as an argument, and parts 3–8 backends override the asset-side staging and the data-side hooks independently, so co-locating them now would churn every backend part mid-review. Happy to add a short "where ordering lives" map to the `ordering.py` module docstring if that would help navigation.

hujc7 (approved):

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -0,0 +1,836 @@
...
+        TypeError: If convention or discovered names are malformed.
+        ValueError: If kind or convention is invalid, a provider rejects source
+            metadata, or an authored robot-schema relationship targets the
+            same prim more than once.
+        NotImplementedError: If no supported source provides a complete ordering.
+            The message identifies the corresponding configuration field, the
+            explicit-name fallback, and a short reason the attempted
+            resolution strategy did not apply.
+    """
+    kind = _validate_ordering_kind(kind)
```
hujc7, line 631: I feel this should be encapsualted into kind.

AntoineRichard: Good call — agreed. Ill encapsulate the element kind into a small internal type (a frozen `_ArticulationElementKind` carrying its `backend_*_names` attribute, robot-schema relationship + name-override attrs, the `joint_ordering`/`body_ordering` config-field label, and whether backend-spelling matching applies), resolved once at the public boundary so the `Literal["joint","body"]` entry contract is unchanged. That collapses the scattered `kind == "joint"` ternaries and per-kind module dicts into one place. Landing it in this PR so it merges alongside Part 3.

AntoineRichard: Follow-up: Part 2 merged before this landed, so the encapsulation is up as a small standalone PR against develop — #6507. Same shape described above (frozen `_ArticulationElementKind`, string boundary preserved, no behavior change).

PR #6336, opened 2026-07-03
Articulation Reordering Part 3/8: Newton backend

AntoineRichard:
# Description

Wires ordering into the Newton backend read/write paths. Non-identity orderings register an unconditional post-step publish that reorders the core state buffers inside the stepped (and CUDA-graph-captured) region, keeping the documented Tier-1 capture-safety contract without Python read guards; `NewtonManager` gains the matching `register`/`unregister_post_step_callback` API. Actuator torque telemetry reads live backend state through the ordering map. Also carries the shared branching USD test fixture used by cross-backend convention tests here and in later parts.

Part 3/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- **Part 3/8 (this PR)**: #6336 — Newton backend

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p2...antoiner%2Farticulation-reordering-p3).

## Type of change

## Checklist

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab_newton/test/assets/test_articulation_ordering_kernels.py` -> 2 passed
- [x] `./isaaclab.sh -p -m pytest source/isaaclab_newton/test/assets/test_newton_actuators_newton.py` -> 76 passed
- [x] `./isaaclab.sh -p -m pytest source/isaaclab_newton/test/assets/test_articulation.py` -> 257 passed, 8 skipped, 1 xfailed


AntoineRichard:
Addressed the first review batch in 470b4b7ee94 — per-comment map:

- **Unify into a single launch (both write_data_to_sim sites):** done, and better than asked — those were the *identity* branches (the ordered path already used the fused kernel). Since the asset-side joint map is an identity arange when no ordering is configured, both branch pairs collapsed into **one unconditional fused launch** (no branch at all): identity path goes from 4 memcpys (direct-drive) / 3 (Lab) to 1 launch; `_assign_backend_ordered_joint_buffer` lost its last callers and is deleted.
- **Unified property-write kernels:** audited every `wp.launch` in the file — none remain to migrate. All joint property writers (stiffness/damping/armature/friction/limits, index+mask) already route through the fused `write_*_user_to_backend` kernels, and position limits have their own fused vec2f kernel. The anchored hunks predate the fused migration that's already on this branch.
- **`_write_actuator_param` torch routing / no mask variant:** correct on both counts — gains are per-actuator torch views over arbitrary joint subsets, so single-launch needs the actuator-side buffer layout rework. Marked with a TODO deferring to the actuator rework built on this series, as suggested.
- **Write-to-sim mega kernel:** evaluated, recommending no: the wrench scatter iterates `(envs, bodies)` vs the target gather's `(envs, joints)` (max-dim launch with divergent guarded phases, zero memory saving), and on the Lab path `_apply_actuator_model()` runs *between* them producing the gather's inputs — they can't share a launch. Reasons now stated in a comment at the wrench launch.
- **Review-trace comment:** reduced to the functional sentence.
- **Runtime validation:** the configure step allocates every checked field, so the per-call validations on sync/rebind paths were unreachable — removed; the validators now run exactly once at the end of configure as a structural guard.
- **"Is this used?" quartet:** full caller-map audit — everything is production-used, nothing deletable: `_refresh_user_order_state` *is* the registered post-step callback (every step under ordering, in-graph); the joint-state pairs in the sync method are the resolve/rebind **seeds** that make first pre-step reads valid, since the post-step publish only fires inside a step — now documented in that docstring.

Tests: baseline-exact (257 passed/8 skipped/1 xfailed + 76 passed), pre-commit clean.

source/isaaclab_newton/isaaclab_newton/actuators/kernels.py
```diff
@@ -176,23 +178,31 @@ def sync_torque_telemetry(
...
         pd_clipped = wp.clamp(pd, -limit, limit)
-        total = pd_clipped + sim_bind_joint_effort[i, j]
+        total = pd_clipped + sim_bind_joint_effort[i, backend_j]
         computed[i, j] = total
         applied[i, j] = total
     else:
-        computed[i, j] = actuator_computed_effort[i, j]
-        applied[i, j] = sim_bind_joint_effort[i, j]
+        computed[i, j] = actuator_computed_effort[i, backend_j]
+        applied[i, j] = sim_bind_joint_effort[i, backend_j]
```
hujc7, line 206: What's the mental model to differentiate when to use j vs backend_j just from the var name? Suggest making it mechanical: index variables carry their order space as a prefix (`user_j` / `backend_j`), and an index only subscripts arrays of its own space — the array names already carry `_backend` suffixes, so `joint_pos_backend[i, backend_j]` vs `joint_pos_target[i, user_j]` becomes self-checking. Same convention core adopted for the map argument (`dst_to_src`).

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -1146,17 +1246,19 @@ def write_joint_position_to_sim_mask(
...
+        ordering_kernels.write_float_user_to_backend_with_mask(
+            position,
+            env_mask,
+            joint_mask,
+            self._joint_user_to_backend_map(),
+            has_joint_ordering,
+            joint_pos_user,
+            self.data._sim_bind_joint_pos,
             device=self.device,
         )
```
hujc7, line 1260: nit: minor consistency issue — some ordering_kernels symbols are called as Python wrappers, others wp.launch'ed as raw kernels. The split has a rationale (wrappers where torch→warp adaptation is needed in setters; raw launches on hot per-step paths with Warp-native inputs) but it's undocumented. Suggest stating the two-tier contract in the ordering_kernels module docstring (which names are wrappers vs launchable kernels) or a naming cue, rather than unifying — all-wrappers costs Python overhead per step, all-raw pushes dtype adaptation to call sites.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -3890,6 +3815,19 @@ def _create_lab_actuator(
         j_ids = actuator.joint_indices
         if j_ids == slice(None):
             j_ids = self._ALL_JOINT_INDICES
+        elif isinstance(j_ids, torch.Tensor):
+            # The ordering write wrappers require Warp int32 selector arrays and no
+            # longer accept torch tensors (the old ``wp.launch`` auto-converted them).
+            if j_ids.dtype == torch.int64:
+                j_ids = j_ids.to(torch.int32)
+            j_ids = wp.from_torch(j_ids, dtype=wp.int32)
+        has_joint_ordering = self.data.joint_ordering is not None
```
hujc7: `has_joint_ordering = self.data.joint_ordering is not None` is re-derived at several call sites — if it's commonly used, suggest a derived property on the data container:

```python
@property
def has_joint_ordering(self) -> bool:
    return self.joint_ordering is not None
```

Property (derivation), deliberately not a stored flag — the stored-copy variant is what the ordering rework removed for staleness.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -167,27 +171,41 @@ def num_bodies(self) -> int:
...
 
+    @property
+    def _backend_num_shapes_per_body(self) -> list[int]:
+        """Number of collision shapes for each backend-order body."""
+        if not hasattr(self, "_num_shapes_per_body_backend"):
+            self._num_shapes_per_body_backend = [len(shapes) for shapes in self._root_view.body_shapes]
+        return self._num_shapes_per_body_backend
+
     @property
     def num_shapes_per_body(self) -> list[int]:
```
hujc7, line 175: **Agent-found bug:** `num_shapes_per_body` now returns public body order, but the `_RandomizeRigidBodyMaterialNewton` consumer in events.py still prefix-sums it (`start_idx = sum(num_shapes_per_body[:body_id])`) against shape-material bindings laid out in backend shape order — nonidentity `body_ordering` + specific `body_ids` silently randomizes the wrong bodies' shapes. `backend_num_shapes_per_body` was added for this and has zero callers; the consumer needs `map_body_ids_to_backend` + backend counts.

source/isaaclab_newton/isaaclab_newton/assets/articulation/kernels.py
```diff
@@ -37,6 +37,27 @@ def compute_soft_joint_pos_limits_func(
 """
 
 
+@wp.kernel
+def update_wrench_array_with_force_and_torque_ordered(
```
hujc7, line 43: The user→backend wrench routing is encoded twice: core's fused `reorder_body_wrench_user_to_backend` (consumed by PhysX) gathers over backend ids, while this kernel re-implements the routing as a masked scatter over user ids with spatial-vector packing inline. The output layouts differ (spatial vectors vs split arrays), but the direction/masking logic could exist once — extend the core kernel family (or share the routing wp.func) and keep only the packing Newton-specific.

source/isaaclab_newton/isaaclab_newton/assets/articulation/kernels.py
```diff
@@ -624,7 +635,12 @@ def concat_joint_pos_limits_lower_and_upper(
 def gather_jacobian_rows(
```
hujc7, line 633: The fixed-base Jacobian row convention is encoded twice: core `_make_jacobian_body_user_to_backend` pre-shifts rows by −1 (correct for PhysX, whose source already omits the root row), and this kernel re-adds `link_offset` because Newton's source keeps all rows. Building the map per source layout — for Newton it's just `user_to_backend[1:]`, no shift — lets `link_offset` and `has_body_ordering` drop out of the kernel, leaving one encoding owned by the builder.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -167,27 +171,41 @@ def num_bodies(self) -> int:
         """Number of bodies in articulation."""
         return self.root_view.link_count
 
+    @property
+    def _backend_num_shapes_per_body(self) -> list[int]:
```
hujc7: This private property is returned verbatim by the public `backend_num_shapes_per_body` (:198) — one layer too many. Suggest folding the cache into the public property and deleting the private one; also the `hasattr`-based lazy init is off-convention — declare the cache attribute in `_create_buffers` like the sibling caches.

source/isaaclab_newton/isaaclab_newton/actuators/adapter.py
```diff
@@ -239,19 +248,20 @@ def build_newton_actuator_defaults(
     dof_offset: int,
     env_stride: int,
     device: str,
+    joint_user_to_backend_indices: Sequence[int] | None = None,
```
hujc7, line 342: The Newton-actuator fast-path init is assembled piecemeal in articulation.py (~:3578-3670): `build_implicit_dof_mask` + `build_newton_actuator_defaults` as free functions with ~8 threaded params each, plus a separate `computed_effort_2d` slice bind — all consuming the same adapter state. Suggest one method on the owner:

```python
binding = adapter.bind_articulation(dof_offset, num_joints, joint_user_to_backend, device)
# .stiffness / .damping / .joint_indices / .implicit_dof_mask / .computed_effort_view
```

Same-purpose functions move into the namespace whose data they read; the parameter threading disappears.

source/isaaclab_newton/isaaclab_newton/actuators/adapter.py
```diff
@@ -302,6 +332,19 @@ def build_newton_actuator_defaults(
...
+        user_to_backend_tensor = torch.tensor(user_to_backend, dtype=torch.long, device=device)
+        stiffness = stiffness.index_select(1, user_to_backend_tensor)
+        damping = damping.index_select(1, user_to_backend_tensor)
+        if not isinstance(joint_indices, slice):
+            backend_to_user = [0] * num_joints
+            for user_index, backend_index in enumerate(user_to_backend):
+                backend_to_user[backend_index] = user_index
+            joint_indices = torch.tensor(
+                sorted(backend_to_user[index] for index in managed_local),
+                dtype=torch.int32,
```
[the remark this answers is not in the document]
AntoineRichard: Clarified — renamed the gather index (`backend_column_indices`) so the backend→user-order direction is explicit at the call site. Commit "Clarify Newton actuator gain-reorder gather index" in this PR.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: Documented the two-tier contract at the call site (commit "Document the two-tier ordering-kernel launch contract"): raw launches on hot per-step paths with Warp-native inputs, Python wrappers where torch→warp adaptation is needed in setters. Left the core `ordering_kernels` module docstring untouched since Part 2 is now merged.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: Added `has_joint_ordering` / `has_body_ordering` as derived properties on the base articulation data class (commit "Add ordering-presence properties and adopt them in Newton") — a derivation, not a stored flag, per your note. Adopted at the Newton sites here; the additional sites introduced by the identity-fast-path work are updated in Part 8.

source/isaaclab_newton/isaaclab_newton/assets/articulation/kernels.py
AntoineRichard: Extracted a Newton-local `resolve_backend_index(...)` `wp.func` shared across the ordered kernels (commit "Share Newton user-to-backend index resolution via a wp.func"), so the direction/masking logic isn't re-implemented per kernel. Full sharing with core's `reorder_body_wrench_user_to_backend` would mean editing that now-merged (Part 2) kernel, and the two are different shapes anyway — core does a backend-iterating gather via `backend_to_user`, Newton a user-iterating masked scatter with spatial-vector packing — so I kept the packing Newton-specific and left core untouched. Can file a follow-up if you'd like the core family extended.

source/isaaclab_newton/isaaclab_newton/assets/articulation/kernels.py
AntoineRichard: Done — the row layout is now encoded in the map builder per source convention (commit "Encode Jacobian body-row layout in the Newton map builder"): Newton's fixed base is `user_to_backend[1:]` with no shift, so `link_offset` and the `has_body_ordering` branch drop out of the gather kernel, leaving one encoding owned by the builder. This also retires the separate "None passed to kernel" note above, since the body-row map is now always built.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: Folded — the cache now lives directly in the public `backend_num_shapes_per_body` property with the private layer removed, and the cache attribute is declared in `_create_buffers` like the sibling caches (no more `hasattr` lazy-init). Commit "Fold Newton backend shape-count cache into its property".

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: This is a split-across-parts artifact rather than a live defect. `backend_num_shapes_per_body` has no caller *in this part*, but its consumer lives in Part 8: there `_RandomizeRigidBodyMaterialNewton` (events.py) uses `asset.backend_num_shapes_per_body` (backend order) together with `map_body_ids_to_backend`, which is exactly the fix you describe. `test_newton_material_randomization_automatically_converts_public_body_ids_to_backend_shape_range` covers it and passes, so there's no wrong-body randomization on the merged stack.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -925,7 +941,12 @@ def body_com_jacobian_w(self) -> ProxyArray:
             inputs=[
                 self._jacobian_buf,
                 self._jacobian_view_art_ids,
+                self._jacobian_body_user_to_backend,
+                self._joint_user_to_backend,
                 self._jacobian_link_offset,
+                self._num_base_dofs,
+                self._has_body_ordering,
+                self._has_joint_ordering,
             ],
```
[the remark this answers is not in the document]
AntoineRichard: Addressed by the Jacobian rework: the body-row map is now built unconditionally in the map builder (commit "Encode Jacobian body-row layout in the Newton map builder"), so `gather_jacobian_rows` / `gather_mass_matrix_rows` no longer receive a nullable array — the identity path passes a real map (`user_to_backend[1:]` for Newton's fixed base), not `None`. The flag/mirror machinery this note references was also removed in the ordering rework.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -308,17 +338,61 @@ def write_data_to_sim(self):
...
+                    self._has_implicit_actuators,
+                    self._has_implicit_actuators,
+                    False,
+                ],
+                outputs=[
+                    self.data._sim_bind_joint_effort,
+                    self.data._sim_bind_joint_position_target,
+                    self.data._sim_bind_joint_velocity_target,
+                    self.data._sim_bind_joint_effort,
+                ],
```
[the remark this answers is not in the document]
AntoineRichard: Good catch. This write launch is restructured by the identity-fast-path change that lands in Part 8 (split into identity/ordered branches); the fix — 4th output → `_sim_bind_joint_act`, matching the newton-actuator branch — is applied there. It'll be visible once Part 8 is refreshed onto the current base.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1772,6 +1790,8 @@ def _simulate_physics_only(cls) -> None:
             contacts = None
 
         cls._run_solver_substeps(contacts)
+        for cb in cls._post_step_callbacks:
+            cb()
         cls._update_sensors(contacts)
 
     # State accessors (used extensively by articulation/rigid object data)
```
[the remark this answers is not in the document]
AntoineRichard: Clarified — the docstring now states the callback fires once per `step()`, reflecting state after all decimation iterations complete (not once per substep). The change lands in Part 8, alongside the rest of the post-step-hook work.

source/isaaclab_newton/isaaclab_newton/actuators/kernels.py
AntoineRichard: Adopted the convention — user-space indices are now `user_j`, backend-space `backend_j`, each subscripting only arrays of its own space (self-checking against the `_backend` array suffixes). The telemetry-kernel changes land in Part 8 where the fast-path gather lives; visible once that part is refreshed.

source/isaaclab_newton/isaaclab_newton/actuators/adapter.py
AntoineRichard: Agreed — consolidating the piecemeal fast-path init (`build_implicit_dof_mask` + `build_newton_actuator_defaults` + the computed-effort slice) into a single `adapter.bind_articulation(...)` on the owner is the right shape. Since it reaches into actuator-init coupling, I'm doing it as a focused follow-up commit rather than folding it into this review-response batch, so it stays cleanly reviewable.

AntoineRichard: Done — added `NewtonActuatorAdapter.bind_articulation(...)` (commit "Encapsulate Newton actuator articulation binding" in this PR), returning an `ArticulationBinding` with `.stiffness` / `.damping` / `.joint_indices` / `.implicit_dof_mask` / `.implicit_dof_mask_owner` / `.computed_effort_view`. The piecemeal `build_implicit_dof_mask` + `build_newton_actuator_defaults` + manual computed-effort slice at the call site collapse into one `binding = adapter.bind_articulation(...)`; the genuinely per-articulation inputs (`lab_actuators`, `dof_offset`, `num_joints`, ordering map) are the only args — `num_envs`/`device`/`actuators`/`env_stride` come from the adapter. The two building-block functions stay public and unit-tested (the method calls them internally). The identical PhysX call site is updated the same way in Part 4.

source/isaaclab_newton/isaaclab_newton/assets/articulation/kernels.py
AntoineRichard: Follow-up: moved the shared user→backend routing helper `resolve_backend_index` (the direction + identity-fallback logic) into the core `ordering_kernels` module (commit "Move resolve_backend_index into the core ordering kernels"), so it's defined once and any backend can reuse it rather than re-implementing the routing inline.

I stopped short of unifying the wrench *kernels* themselves: core's `reorder_body_wrench_user_to_backend` is a backend-iterating gather (via `backend_to_user`) writing split force/torque arrays, while Newton's is a user-iterating masked scatter that packs into a single spatial vector. The direction, layout, and masking all differ, so beyond the shared index helper the two are genuinely different shapes — folding them into one core kernel would cost more than the one line of shared logic it saves. Happy to revisit if you'd prefer the full core-family extension.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -3541,13 +3443,34 @@ def _create_buffers(self):
...
         # asset named data
-        self.data.joint_names = self.joint_names
-        self.data.body_names = self.body_names
+        self._resolve_and_install_ordering_maps()
+        self._ordering_configure_backend_staging()
+        # Cache a torch ``long`` alias of the joint user-to-backend map so per-call actuator
+        # writes reuse it instead of re-wrapping and re-casting the Warp map every call.
+        self._joint_user_to_backend_torch = (
+            wp.to_torch(self.data.joint_ordering.user_to_backend).to(dtype=torch.long)
+            if self.data.joint_ordering is not None
```
hujc7: Two leftover sites re-derive the check the new property encapsulates: `self.data.joint_ordering is not None` here and `self.joint_ordering is not None` at the `bind_articulation` call — suggest `self.data.has_joint_ordering` (or bind the map to a local first) for consistency with the converted call sites.

hujc7 (approved):

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: Done — bound the map to a local at both sites (commit "Use bound ordering local at leftover derivation sites"): `joint_ordering = self.data.joint_ordering` / `self.joint_ordering` before the presence check, so the `is not None` isn't re-derived inline and the map access reuses the local, consistent with the converted call sites.

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
hujc7: I was actually thinking just use self.data.has_joint_ordering as check, and self.data.joint_ordering if check passes XD. Not needing to create a new var should be slightly better

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -1669,16 +1770,30 @@ def write_joint_position_limit_to_sim_mask(
         if isinstance(limits, float):
             raise ValueError("Joint position limits must be a tensor or array, not a float.")
         self.assert_shape_and_dtype_mask(limits, (env_mask, joint_mask), wp.vec2f, "limits")
+
+        _ = self.data.joint_pos_limits
+        has_joint_ordering = self.data.has_joint_ordering
```
hujc7, line 1775: nit: Ideally, should prefer using the existing one over creating new var as it does not contain any extra logic. At call site, it becomes obvious.

PR #6337, opened 2026-07-03
Articulation Reordering Part 4/8: PhysX backend

AntoineRichard:
# Description

Wires ordering into the PhysX backend read/write paths using the shared kernels: fused user-and-backend target writes, lazily refreshed timestamped `body_mass`/`body_inertia` (see the changelog `Changed` entry for the semantics change vs develop), and ordering maps resolved at initialization.

Part 4/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- **Part 4/8 (this PR)**: #6337 — PhysX backend

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p3...antoiner%2Farticulation-reordering-p4).

## Type of change

## Checklist

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/assets/test_articulation.py` -> 239 passed
- [x] `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/assets/test_newton_actuators_physx.py` -> 44 passed


source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -793,31 +801,106 @@ def root_com_vel_w(self) -> ProxyArray:
...
+        pipelines this refresh performs a host-to-device copy; this is acceptable because consumers
+        are init or episodic. Values written directly through the tensor view (``root_view.set_masses``)
+        become visible at the first read after the next simulation update; same-step view-write
+        visibility is not guaranteed.
         """
-        self._body_mass.assign(self._root_view.get_masses())
+        self._refresh_reordered_body_buffer(self._body_mass, self._body_mass_backend, self._root_view.get_masses)
         if self._body_mass_ta is None:
-            self._body_mass_ta = ProxyArray(self._body_mass)
+            self._body_mass_ta = ProxyArray(self._body_mass.data)
```
hujc7, line 903: I had this question before, is lazy initialization good for capturing?

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -1682,13 +1949,240 @@ def _create_buffers(self) -> None:
...
+        had_joint_ordering = self._joint_pos_backend is not None
+        had_body_ordering = self._body_com_pose_b_backend is not None
+        joint_ordering = self.joint_ordering
+        body_ordering = self.body_ordering
+        self._configure_ordering_buffers(had_joint_ordering, had_body_ordering)
+
+        if body_ordering is not None:
+            self._jacobian_body_user_to_backend = self._make_jacobian_body_user_to_backend()
+        elif joint_ordering is not None:
+            self._jacobian_body_user_to_backend = joint_ordering.user_to_backend
```
hujc7: Agent review: both Jacobian map attributes are sometimes filled with the *other axis's* map as a never-read launch placeholder (joints-only ordering parks the joint map in `_jacobian_body_user_to_backend`; bodies-only parks the row map in the joint attr). Safe today — the `has_*` flags gate all indexing — but it's the wrong axis and wrong length behind a trusted name. Suggest passing `None` for the unused axis instead: Warp accepts None for array params a kernel doesn't index (verified), and the attribute stays honest (real row map or None, nothing else).

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -1682,13 +1949,240 @@ def _create_buffers(self) -> None:
...
+        _inertias = self._root_view.get_inertias()
+        self._body_inertia = TimestampedBuffer(_inertias.shape, self.device, _inertias.dtype)
+        self._body_mass_backend: wp.array | None = None
+        self._body_inertia_backend: wp.array | None = None
         self._default_root_state = None
 
         # Initialize ProxyArray wrappers
         self._pin_proxy_arrays()
 
+    def _configure_ordering_buffers(self, had_joint_ordering: bool, had_body_ordering: bool) -> None:
```
hujc7: Agent review: the `had_joint_ordering`/`had_body_ordering` release/reseed branches are unreachable on PhysX — `_resolve_and_install_ordering_maps` runs exactly once per fresh data instance (~40 untested lines; same on OVPhysX), and the dead parameters mislead the signature's reader. Either wire a real re-resolve lifecycle in core or drop the `had_*` plumbing chain-wide.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -3978,6 +4149,49 @@ def _invalidate_initialize_callback(self, event):
...
+            shape = (self.num_instances, self.num_bodies)
+            if getattr(self, "_body_wrench_force_backend", None) is None:
+                self._body_wrench_force_backend = wp.zeros(shape, dtype=wp.vec3f, device=self.device)
+            if getattr(self, "_body_wrench_torque_backend", None) is None:
+                self._body_wrench_torque_backend = wp.zeros(shape, dtype=wp.vec3f, device=self.device)
+        else:
+            self._body_wrench_force_backend = None
+            self._body_wrench_torque_backend = None
+
+    def _get_user_ordered_joint_3d_buffer(
```
hujc7, line 4165: Agent suggestion (optional): two consolidations for whenever it's convenient — this part or the stack tail. `_get_user_ordered_joint_3d_buffer` is the read-direction mirror of core's `_get_backend_ordered_joint_buffer`; since OVPhysX needs the identical read path, hoisting it to `BaseArticulation` beside its mirror avoids a second copy (Newton simply inherits it unused, like the write-direction helper today). And the torch→wp.int32 selector conversion block is copy-pasted 5× — foldable into `_resolve_joint_ids`/`_resolve_body_ids`, or the write wrappers could adapt torch selectors the way they already adapt torch values.

source/isaaclab_physx/test/assets/test_articulation.py
```diff
@@ -29,7 +30,7 @@
 import isaaclab.utils.math as math_utils
 import isaaclab.utils.string as string_utils
 from isaaclab.actuators import ActuatorBase, IdealPDActuatorCfg, ImplicitActuatorCfg
-from isaaclab.assets import ArticulationCfg
+from isaaclab.assets import ArticulationCfg, get_articulation_name_ordering
```
hujc7, line 33: Agent review: the trim commit removes the two live ANYmal-D ordering tests as redundant — redundant with what? Everything else exercising PhysX ordering runs against hand-written mock views, so these were the only tests validating the feature against real PhysX on GPU (the anchor that catches mock drift). Suggest keeping one live test or noting the replacement explicitly.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -263,9 +267,32 @@ def write_data_to_sim(self):
             else:
                 composer = self._permanent_wrench_composer
             composer.compose_to_body_frame()
+            if self.data.body_ordering is not None:
```
hujc7: Agent suggestion: this gate re-derives what `data.has_body_ordering` (added in the Part 3 round) encapsulates. Could the pure boolean gates in this backend be swept to the properties? (~10 in this file, ~16 in the data class — sites that dereference the map via a local are fine as-is.)

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: For PhysX it doesn't matter we won't be able to capture anyway

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: makes sense

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
AntoineRichard: WOuld make sense

source/isaaclab_physx/test/assets/test_newton_actuators_physx.py
```diff
@@ -170,15 +213,241 @@ def _run_simulation(
...
         "applied_torque": recorded_applied,
+        "adapter_computed_effort": recorded_adapter_computed,
+        "adapter_applied_effort": recorded_adapter_applied,
         "target_pos": target_pos.clone(),
         "target_vel": target_vel.clone(),
+        "effort_target": None if effort_target is None else effort_target.clone(),
+    }
+
+
+_ORDERING_TRACE_FIELDS = (
```
ooctipus, line 251: Agent review: lines 237–343 (`_ORDERING_TRACE_FIELDS`, `_ORDERING_TRACE_TOLERANCES`, `_canonicalize_ordering_result`, the reversed-ordering rollout test) are byte-identical to `test_newton_actuators_newton.py:259-365` from Part 3 except one docstring word, and the ordering plumbing inside `_run_simulation` is copied line-for-line too. Since Part 5 exists to extract shared articulation test utilities and Part 6 will need the same machinery again, could this live there instead of a second (soon third) copy? Two riders if it moves: `is_identity` is hardcoded `False` by the harness (line 180) and asserted `not ...` by the test (line 305) — the index-map assertions right below already prove non-identity — and the `num_steps=0` probe at line 448 builds a full Kit/PhysX context just to read joint names the Newton tests pin as a constant.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -832,7 +915,17 @@ def body_link_pose_w(self) -> ProxyArray:
         if self._body_link_pose_w.timestamp < self._sim_timestamp:
```
ooctipus, line 933: Agent suggestion: `body_link_pose_w`, `body_com_vel_w`, and `body_com_acc_w` inline the same gate-and-gather block that this file already abstracts twice for the neighboring axes (`_refresh_joint_state_user`, `_refresh_reordered_body_buffer`) — only the map and dim differ. Either form is fine on its own, but the file currently carries both; whichever one wins is what OVPhysX will copy in Part 6.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -730,13 +737,14 @@ def root_link_vel_w(self) -> ProxyArray:
         relative to the world.
         """
         if self._root_link_vel_w.timestamp < self._sim_timestamp:
+            body_com_pose_b = self._backend_body_com_pose_b if self.body_ordering is not None else self.body_com_pose_b
```
ooctipus: Agent review: this ternary (also at `root_com_pose_w`) re-derives the branch `_backend_body_com_pose_b` already takes internally — the base property returns the public buffer under identity ordering. Both sites can use the property unconditionally.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -263,9 +267,32 @@ def write_data_to_sim(self):
...
+                force_backend = self._body_wrench_force_backend
+                torque_backend = self._body_wrench_torque_backend
+                if force_backend is None or torque_backend is None:
```
ooctipus: Agent review: these staging guards (also 2407, 2499, 2610, 4185, and `articulation_data.py:1271`) can't fire: `_create_buffers` installs the staging immediately after resolving ordering, and ordering resolves once per instance (as you noted on the `had_*` thread). If the dead release/reseed plumbing goes, these six can go with it — or collapse to one check at configure time.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -1682,13 +1949,240 @@ def _create_buffers(self) -> None:
...
+            self._body_link_pose_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.transformf, device=self.device
+            )
+            self._body_com_vel_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
+            )
+            self._body_com_acc_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
+            )
+        self._body_link_pose_w_ta = None
```
ooctipus: Agent suggestion: Newton's `_apply_ordering_maps_after_resolve` ends with a single `_pin_proxy_arrays()` instead of enumerating wrappers — it already nulls every `_ta`, and lazy re-wrap is harmless for untouched buffers. Same here would replace the 16 per-attribute resets (and the doubled `_body_com_pose_b` invalidation).

source/isaaclab_physx/test/assets/test_articulation.py
```diff
@@ -310,6 +363,246 @@ def sim(request):
...
+
+@pytest.mark.parametrize("device", ["cuda:0", "cpu"])
+@pytest.mark.parametrize("gravity_enabled", [False])
+@pytest.mark.parametrize("body_ordering", ["identity", "reversed"])
+def test_live_direct_view_mass_inertia_writes_become_visible(sim, device, gravity_enabled, body_ordering):
+    """Direct tensor-view writes to body masses/inertias become visible on the lazy read.
+
+    Regression for the timestamp-lazy body-property buffers. Shipped Factory/Forge envs write
+    masses/inertias straight through ``root_view.set_masses`` / ``set_inertias`` (bypassing the
+    asset setters). With the previous init-clone-only getters those writes were never seen. The
```
ooctipus: Agent review: this docstring inverts the change: the merge-base getters re-read the view on every access (`assign(get_masses())` in the old property body), so direct view writes were always visible — the new lazy gate is what could *hide* them, which is what this test actually pins.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -49,17 +52,21 @@ class ArticulationData(BaseArticulationData):
     .. note::
         **Pull-to-refresh model.** PhysX state properties are *not* automatically updated each
         simulation step. Each property getter pulls fresh data from the PhysX tensor API on first
-        access per timestamp, then caches the result until the next step. This differs from the
-        Newton backend, where buffers are refreshed automatically by the simulation.
+        access per timestamp, then caches the result until the next step. Newton's solver-owned,
+        backend-order buffers are refreshed automatically by the simulation; when Newton uses a
+        nonidentity public ordering, its public-order shadows are also refreshed lazily on first
```
ooctipus: Agent review: Newton's ordered shadows aren't "refreshed lazily on first access" — `_refresh_user_order_state` is a post-step callback that publishes every step (Part 3's changelog says the same). The merge-base one-sentence contrast was accurate.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -793,31 +801,106 @@ def root_com_vel_w(self) -> ProxyArray:
...
+        self,
+        buf: TimestampedBuffer,
+        backend_buffer: wp.array,
+        view_getter: Callable[[], wp.array],
+        *,
+        component_count: int | None = None,
+    ) -> None:
+        """Refresh a timestamp-lazy static body buffer from a CPU-only tensor view.
+
+        When stale for the current step, aliases the view directly (identity body
```
ooctipus: Agent review: "aliases the view directly" — the identity branch is `buf.data.assign(...)`, a copy, never an alias (necessarily, the views are CPU-only). Alias-vs-copy is the load-bearing distinction in this file's ProxyArray notes, so worth the one-word fix.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -793,31 +801,106 @@ def root_com_vel_w(self) -> ProxyArray:
...
+        buf.timestamp = self._sim_timestamp
+
     @property
     def body_mass(self) -> ProxyArray:
-        """Body mass in the world frame.
+        """Body mass [kg].
 
         Shape is (num_instances, num_bodies), dtype = wp.float32. In torch this resolves to (num_instances, num_bodies).
+
+        Refresh is timestamp-lazy: the buffer is refreshed from the tensor view at most once per
```
ooctipus: Agent suggestion: the timestamp-lazy/CPU-only contract paragraph appears in full four times (here, `body_inertia`, the `_make_*` init comment, the changelog), and the alias-vs-gather + no-`_ensure_fk_fresh` template three more times across the dynamics properties plus a fourth copy on `_refresh_generalized_joint_buffer`. Keeping the full statement once on the helper that implements it, with one-line pointers at the properties, would remove most of this file's added prose.

ooctipus (approved):

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Confirmed. No code change is needed here: PhysX tensor reads are not graph-captured, so lazy ProxyArray creation does not change capture support. The wrapper remains cached after first access.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. The inactive Jacobian axis map now remains `None`, and `test_unused_jacobian_ordering_map_is_none` covers joint-only and body-only ordering on both CPU and CUDA.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. `_configure_ordering_buffers()` is now explicitly a one-time install path; the `had_*` parameters, release/reseed branches, and rebind wording are removed.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
AntoineRichard: Agreed on the optional consolidation. I am deferring the base-class hoist until the later stack part has the additional shared consumer; keeping the helper local here avoids expanding the base API in the Part 4 correctness change.

source/isaaclab_physx/test/assets/test_articulation.py
AntoineRichard: There is no live-coverage gap: the removed ANYmal-D cases were replaced by `test_live_manual_root_preserving_ordering_reorders_backend_reads_and_writes`, `test_live_floating_root_writers_match_identity_after_body_reordering`, and the direct mass/inertia test. Those run against real PhysX on CPU/CUDA; the local USDA fixture only adds convention-specific coverage.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
AntoineRichard: Addressed in 7269f44a753. Pure boolean gates now use `has_joint_ordering` / `has_body_ordering`; direct map checks remain only where the map is immediately dereferenced.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -1682,13 +1949,240 @@ def _create_buffers(self) -> None:
...
+        if body_ordering is not None:
+            self._body_link_pose_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.transformf, device=self.device
+            )
+            self._body_com_vel_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
+            )
+            self._body_com_acc_w.data = wp.zeros(
+                (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
+            )
```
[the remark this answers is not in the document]
AntoineRichard: Addressed in 7269f44a753. Body-only ordering now leaves `_jacobian_joint_user_to_backend` as `None`, with axis-specific CPU/CUDA regression assertions.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -263,9 +267,32 @@ def write_data_to_sim(self):
...
+                force_data = composer.out_force_b.warp
+                torque_data = composer.out_torque_b.warp
             self.root_view.apply_forces_and_torques_at_position(
-                force_data=composer.out_force_b.warp.flatten().view(wp.float32),
-                torque_data=composer.out_torque_b.warp.flatten().view(wp.float32),
+                force_data=force_data.flatten().view(wp.float32),
+                torque_data=torque_data.flatten().view(wp.float32),
                 position_data=None,
                 indices=self._ALL_INDICES,
                 is_global=False,
```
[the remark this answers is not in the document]
AntoineRichard: Addressed in 7269f44a753. The gated-off `joint_act` output is now `None`; both reversed-ordering Newton-actuator regressions pass with that launch.

source/isaaclab_physx/test/assets/test_newton_actuators_physx.py
AntoineRichard: The shared test-utility extraction remains assigned to Part 5 so it lands once for all consumers. In this Part 4 feedback commit I removed the redundant hardcoded `is_identity` field/assertion and replaced the zero-step Kit probe with the canonical ANYmal-C PhysX name tuple.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. `_refresh_body_state_user` now owns the shared timestamp gate and body-axis gather for link pose, COM velocity, and COM acceleration while preserving the existing FK refresh points.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. Both root computations now use `_backend_body_com_pose_b` unconditionally and rely on its identity-order fallback.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
AntoineRichard: Addressed in 7269f44a753. The unreachable staging guards were removed; the code now relies on the one-time ordering configuration invariant established during initialization.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. `_apply_ordering_maps_after_resolve` now finishes with one `_pin_proxy_arrays()` call instead of maintaining a manual proxy-cache reset list.

source/isaaclab_physx/test/assets/test_articulation.py
AntoineRichard: Addressed in 7269f44a753. The regression now correctly states that develop rereads the tensor view and that the new timestamp-lazy cache is what could hide a direct write until the next eligible refresh.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. The note now states that Newton solver buffers and nonidentity public-order shadows are published automatically by the simulation.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. The helper documentation now says `assign()` copies the CPU-only view rather than aliases it.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Addressed in 7269f44a753. The common refresh contract now lives in the helpers, and the mass-matrix, gravity-compensation, initialization, and body-property prose has been trimmed to the property-specific behavior.

PR #6339, opened 2026-07-03
Articulation Reordering Part 6/8: OVPhysX backend

AntoineRichard:
# Description

Wires ordering into the OVPhysX backend read/write paths: fused user-and-backend writes with persistent target staging separated from transient applied-torque scratch, FK-aware body state refresh, and a fused ordered joint acceleration kernel.

Part 6/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- Part 1/8: #6334 — Extract rigid object interface test utilities
- Part 2/8: #6335 — Core ordering foundation
- Part 3/8: #6336 — Newton backend
- Part 4/8: #6337 — PhysX backend
- Part 5/8: #6338 — Extract articulation interface test utilities
- **Part 6/8 (this PR)**: #6339 — OVPhysX backend
- Part 7/8: #6340 — Cross-backend ordering interface tests
- Part 8/8: #6341 — Events, consumers, and documentation

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p5...antoiner%2Farticulation-reordering-p6).

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [x] My name exists in `CONTRIBUTORS.md`

## Test Plan

- [x] Mock-based coverage lands in Part 7 (233 passed there). The live suite is blocked in the local environment: all 105 failures share the single `PhysX.set_cpu_mode` AttributeError from the OVPhysX 0.4.13 wheel vs #6314's manager code (develop shows the identical signature on its 99 pre-existing live tests); the same tests passed on the pre-sync monolithic branch
- [x] CI with the 0.5.1 wheelhouse is the authoritative gate for this part
- [x] `./isaaclab.sh -f` -> all hooks pass

PR #6338, opened 2026-07-03
Articulation Reordering Part 5/8: Extract articulation interface test utilities

AntoineRichard:
# Description

Moves the articulation interface test factories into a shared utility module mirroring Part 1's rigid object extraction, with ordering-aware construction hooks used by Part 7's cross-backend suite.

Part 5/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- Part 1/8: #6334 — Extract rigid object interface test utilities
- Part 2/8: #6335 — Core ordering foundation
- Part 3/8: #6336 — Newton backend
- Part 4/8: #6337 — PhysX backend
- **Part 5/8 (this PR)**: #6338 — Extract articulation interface test utilities
- Part 6/8: #6339 — OVPhysX backend
- Part 7/8: #6340 — Cross-backend ordering interface tests
- Part 8/8: #6341 — Events, consumers, and documentation

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p4...antoiner%2Farticulation-reordering-p5).

## Type of change

- Test/documentation/refactor change

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [x] My name exists in `CONTRIBUTORS.md`

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/assets/test_articulation_iface.py` -> 4086 passed, 96 skipped, 32 xfailed (with Newton and PhysX wired, OVPhysX not yet — exercises the unwired-backend identity path)
- [x] `./isaaclab.sh -f` -> all hooks pass

PR #6340, opened 2026-07-03
Articulation Reordering Part 7/8: Cross-backend ordering interface tests

AntoineRichard:
# Description

Adds the shared interface test suite exercising ordering allocation, write parity, partial writes, and property reordering uniformly across PhysX, OVPhysX, and Newton through their mock views. Lands after all three backend parts because present-but-unwired backends error rather than skip.

Part 7/8 of the series splitting #6248 into independently reviewable units. #6248 remains open, untouched, as the reference implementation; this stack was cut from its final tree after a `develop` sync (Kamino Core, Newton bump, OVPhysX 0.5.1).

## Stack

- **Part 7/8 (this PR)**: #6340 — Cross-backend ordering interface tests

> [!NOTE]
> All parts target `develop`, so until the preceding parts merge this PR's diff includes them. The true delta of this part is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p6...antoiner%2Farticulation-reordering-p7).

## Type of change

- Test/documentation/refactor change

## Checklist

## Test Plan

- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/assets/test_articulation_ordering_iface.py` -> 233 passed

### DIFF
--- a/source/isaaclab/isaaclab/test/utils/articulation_ordering.py
+++ b/source/isaaclab/isaaclab/test/utils/articulation_ordering.py
@@ -5,9 +5,70 @@
 
 """Assertions for comparing articulation traces across joint orderings."""
 
+from pathlib import Path
 from typing import Any
 
 import torch
+
+ANYMAL_C_PHYSX_JOINT_NAMES = (
+    "LF_HAA",
+    "LH_HAA",
+    "RF_HAA",
+    "RH_HAA",
+    "LF_HFE",
+    "LH_HFE",
+    "RF_HFE",
+    "RH_HFE",
+    "LF_KFE",
+    "LH_KFE",
+    "RF_KFE",
+    "RH_KFE",
+)
+
+PANDA_JOINT_NAMES = (
+    "panda_joint1",
+    "panda_joint2",
+    "panda_joint3",
+    "panda_joint4",
+    "panda_joint5",
+    "panda_joint6",
+    "panda_joint7",
+    "panda_finger_joint1",
+    "panda_finger_joint2",
+)
+
+PANDA_BODY_NAMES = (
+    "panda_link0",
+    "panda_link1",
+    "panda_link2",
+    "panda_link3",
+    "panda_link4",
+    "panda_link5",
+    "panda_link6",
+    "panda_link7",
+    "panda_hand",
+    "panda_leftfinger",
+    "panda_rightfinger",
+)
+PANDA_ROOT_PRESERVING_REVERSED_BODY_NAMES = (PANDA_BODY_NAMES[0], *reversed(PANDA_BODY_NAMES[1:]))
+
+BRANCHING_PHYSX_JOINT_NAMES = ("left_shoulder", "right_shoulder", "left_elbow", "right_elbow")
+BRANCHING_MJWARP_JOINT_NAMES = ("left_shoulder", "left_elbow", "right_shoulder", "right_elbow")
+BRANCHING_PHYSX_BODY_NAMES = ("base", "left_upper", "right_upper", "left_tip", "right_tip")
+BRANCHING_MJWARP_BODY_NAMES = ("base", "left_upper", "left_tip", "right_upper", "right_tip")
+
+
+def articulation_ordering_branching_fixture_path() -> Path:
+    """Return the repository path to the shared branching articulation fixture."""
+    return (
+        Path(__file__).resolve().parents[4]
+        / "isaaclab_physx"
+        / "test"
+        / "assets"
+        / "data"
+        / "articulation_ordering_branching.usda"
+    )
+
 
 _ORDERING_TRACE_FIELDS = (
     "joint_pos",
--- a/source/isaaclab/isaaclab/test/utils/articulation_ordering.py
+++ b/source/isaaclab/isaaclab/test/utils/articulation_ordering.py
@@ -5,7 +5,6 @@
 
 """Assertions for comparing articulation traces across joint orderings."""
 
-from pathlib import Path
 from typing import Any
 
 import torch
@@ -56,18 +55,6 @@ BRANCHING_PHYSX_JOINT_NAMES = ("left_shoulder", "right_shoulder", "left_elbow", 
 BRANCHING_MJWARP_JOINT_NAMES = ("left_shoulder", "left_elbow", "right_shoulder", "right_elbow")
 BRANCHING_PHYSX_BODY_NAMES = ("base", "left_upper", "right_upper", "left_tip", "right_tip")
 BRANCHING_MJWARP_BODY_NAMES = ("base", "left_upper", "left_tip", "right_upper", "right_tip")
-
-
-def articulation_ordering_branching_fixture_path() -> Path:
-    """Return the repository path to the shared branching articulation fixture."""
-    return (
-        Path(__file__).resolve().parents[4]
-        / "isaaclab_physx"
-        / "test"
-        / "assets"
-        / "data"
-        / "articulation_ordering_branching.usda"
-    )
 
 
 _ORDERING_TRACE_FIELDS = (
--- a/source/isaaclab/test/assets/test_articulation_ordering_iface.py
+++ b/source/isaaclab/test/assets/test_articulation_ordering_iface.py
@@ -7,10 +7,6 @@
 
 """Mocked cross-backend articulation ordering interface tests."""
 
-import subprocess
-import sys
-import textwrap
-from pathlib import Path
 from unittest.mock import MagicMock
 
 import numpy as np
@@ -21,57 +17,6 @@ from _articulation_iface_test_utils import BACKENDS, get_articulation
 
 from isaaclab.utils.buffers import TimestampedBufferWarp
 from isaaclab.utils.wrench_composer import WrenchComposer
-
-
-def test_discoverable_backend_internal_import_error_propagates() -> None:
-    """Do not treat a discoverable backend with a broken internal import as absent."""
-    script = textwrap.dedent(
-        """
-        import builtins
-        import importlib.util
-        import os
-        import types
-
-        os.environ["LD_PRELOAD"] = "ovphysx-test"
-        real_find_spec = importlib.util.find_spec
-        real_import = builtins.__import__
-
-        def find_spec(name, package=None):
-            if name == "isaaclab_ovphysx":
-                return object()
-            return real_find_spec(name, package)
-
-        def guarded_import(name, globals=None, locals=None, fromlist=(), level=0):
-            if name == "isaaclab_ovphysx.assets.articulation.articulation":
-                return types.SimpleNamespace(Articulation=object)
-            if name == "isaaclab_ovphysx.assets.articulation.articulation_data":
-                raise ModuleNotFoundError(
-                    "broken backend dependency",
-                    name="broken_backend_dependency",
-                )
-            return real_import(name, globals, locals, fromlist, level)
-
-        importlib.util.find_spec = find_spec
-        builtins.__import__ = guarded_import
-
-        try:
-            import _articulation_iface_test_utils
-        except ModuleNotFoundError as exc:
-            if exc.name == "broken_backend_dependency":
-                raise SystemExit(0)
-            raise
-        raise SystemExit("discoverable backend internal import error was swallowed")
-        """
-    )
-
-    result = subprocess.run(
-        [sys.executable, "-c", script],
-        cwd=Path(__file__).parent,
-        capture_output=True,
-        text=True,
-    )
-
-    assert result.returncode == 0, result.stdout + result.stderr
 
 
 def _make_body_ordering_backend_data(num_instances: int, num_bodies: int) -> tuple[np.ndarray, ...]:
--- a/source/isaaclab/test/assets/test_articulation_ordering_iface.py
+++ b/source/isaaclab/test/assets/test_articulation_ordering_iface.py
@@ -1,0 +1,2536 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# pyright: reportPrivateUsage=none
+
+"""Mocked cross-backend articulation ordering interface tests."""
+
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import numpy as np
+import pytest
+import torch
+import warp as wp
+from _articulation_iface_test_utils import BACKENDS, get_articulation
+
+from isaaclab.utils.buffers import TimestampedBufferWarp
+from isaaclab.utils.wrench_composer import WrenchComposer
+
+
+def test_discoverable_backend_internal_import_error_propagates() -> None:
+    """Do not treat a discoverable backend with a broken internal import as absent."""
+    script = textwrap.dedent(
+        """
+        import builtins
+        import importlib.util
+        import os
+        import types
+
+        os.environ["LD_PRELOAD"] = "ovphysx-test"
+        real_find_spec = importlib.util.find_spec
+        real_import = builtins.__import__
+
+        def find_spec(name, package=None):
+            if name == "isaaclab_ovphysx":
+                return object()
+            return real_find_spec(name, package)
+
+        def guarded_import(name, globals=None, locals=None, fromlist=(), level=0):
+            if name == "isaaclab_ovphysx.assets.articulation.articulation":
+                return types.SimpleNamespace(Articulation=object)
+            if name == "isaaclab_ovphysx.assets.articulation.articulation_data":
+                raise ModuleNotFoundError(
+                    "broken backend dependency",
+                    name="broken_backend_dependency",
+                )
+            return real_import(name, globals, locals, fromlist, level)
+
+        importlib.util.find_spec = find_spec
+        builtins.__import__ = guarded_import
+
+        try:
+            import _articulation_iface_test_utils
+        except ModuleNotFoundError as exc:
+            if exc.name == "broken_backend_dependency":
+                raise SystemExit(0)
+            raise
+        raise SystemExit("discoverable backend internal import error was swallowed")
+        """
+    )
+
+    result = subprocess.run(
+        [sys.executable, "-c", script],
+        cwd=Path(__file__).parent,
+        capture_output=True,
+        text=True,
+    )
+
+    assert result.returncode == 0, result.stdout + result.stderr
+
+
+def _make_body_ordering_backend_data(num_instances: int, num_bodies: int) -> tuple[np.ndarray, ...]:
+    """Create deterministic backend-order body data with identity rotations."""
+    root_pose = np.zeros((num_instances, 7), dtype=np.float32)
+    root_pose[:, 6] = 1.0
+    root_vel = np.zeros((num_instances, 6), dtype=np.float32)
+    link_pose = np.zeros((num_instances, num_bodies, 7), dtype=np.float32)
+    com_pose_b = np.zeros((num_instances, num_bodies, 7), dtype=np.float32)
+    body_com_vel = np.zeros((num_instances, num_bodies, 6), dtype=np.float32)
+    body_acc = np.zeros((num_instances, num_bodies, 6), dtype=np.float32)
+
+    for env_index in range(num_instances):
+        root_pose[env_index, :3] = (10.0 + env_index, 0.0, 0.0)
+        root_vel[env_index, 0] = 3.0 + env_index
+        root_vel[env_index, 5] = 7.0 + env_index
+        for body_index in range(num_bodies):
+            link_pose[env_index, body_index, :3] = (10.0 * body_index, float(env_index), 0.0)
+            link_pose[env_index, body_index, 6] = 1.0
+            com_pose_b[env_index, body_index, :3] = (float(body_index + 1), 0.0, 0.0)
+            com_pose_b[env_index, body_index, 6] = 1.0
+            body_com_vel[env_index, body_index, 0] = 20.0 + body_index
+            body_com_vel[env_index, body_index, 5] = 30.0 + body_index
+            body_acc[env_index, body_index, 0] = 100.0 + body_index
+            body_acc[env_index, body_index, 3] = 200.0 + body_index
+
+    return root_pose, root_vel, link_pose, com_pose_b, body_com_vel, body_acc
+
+
+def _install_test_body_ordering(art) -> np.ndarray:
+    """Install a non-identity body ordering that preserves a fixed root body."""
+    if art.is_fixed_base:
+        body_ordering = (art.backend_body_names[0], *reversed(art.backend_body_names[1:]))
+    else:
+        body_ordering = tuple(reversed(art.backend_body_names))
+    art.cfg = art.cfg.replace(body_ordering=body_ordering)
+    art._resolve_and_install_ordering_maps()
+    # Mirror the backend initialization path, which resets the guard flag before
+    # rebuilding ordering maps on an already-initialized articulation.
+    art._ordering_maps_cached = False
+    art._cache_ordering_maps()
+    return np.asarray(art.body_ordering.user_to_backend_indices, dtype=np.int64)
+
+
+def _install_reversed_joint_ordering(art) -> np.ndarray:
+    """Install a reversed public joint ordering on an already constructed articulation."""
+    art.cfg = art.cfg.replace(joint_ordering=tuple(reversed(art.backend_joint_names)))
+    art._resolve_and_install_ordering_maps()
+    # Mirror the backend initialization path, which resets the guard flag before
+    # rebuilding ordering maps on an already-initialized articulation.
+    art._ordering_maps_cached = False
+    art._cache_ordering_maps()
+    return np.asarray(art.joint_ordering.user_to_backend_indices, dtype=np.int64)
+
+
+@pytest.mark.parametrize("backend", ["physx", "ovphysx", "newton"])
+def test_cache_ordering_maps_rejects_reinvocation_after_init(backend: str) -> None:
+    """Reject a second ordering-map cache without an explicit guard-flag reset."""
+    if backend not in BACKENDS:
+        pytest.skip(f"{backend} backend is not available")
+    art, _ = get_articulation(backend, num_instances=2, num_joints=4, num_bodies=4, device="cpu")
+
+    # Initialization already cached the maps; re-invoking without resetting the guard
+    # flag (as backends do before their single call) must be rejected.
+    assert art._ordering_maps_cached is True
+    with pytest.raises(RuntimeError, match="cached exactly once"):
+        art._cache_ordering_maps()
+
+
+def _joint_ordering_for_mode(mode: str, num_joints: int) -> tuple[str, ...] | None:
+    """Return the configured public joint ordering for a parity-test mode."""
+    backend_names = tuple(f"joint_{index}" for index in range(num_joints))
+    if mode == "none":
+        return None
+    if mode == "identity":
+        return backend_names
+    if mode == "reversed":
+        return tuple(reversed(backend_names))
+    raise ValueError(f"Unsupported joint ordering mode: {mode}")
+
+
+def _body_ordering_for_mode(mode: str, num_bodies: int) -> tuple[str, ...] | None:
+    """Return the configured public body ordering for a parity-test mode."""
+    backend_names = tuple(f"body_{index}" for index in range(num_bodies))
+    if mode == "none":
+        return None
+    if mode == "identity":
+        return backend_names
+    if mode == "reversed":
+        return tuple(reversed(backend_names))
+    raise ValueError(f"Unsupported body ordering mode: {mode}")
+
+
+def _ordering_shadow_shape(buffer: wp.array | TimestampedBufferWarp) -> tuple[int, ...]:
+    """Return the allocation shape for a raw or timestamped ordering shadow."""
+    if isinstance(buffer, TimestampedBufferWarp):
+        buffer = buffer.data
+    return tuple(buffer.shape)
+
+
+def _ordering_user_to_backend(ordering, item_count: int) -> np.ndarray:
+    """Return public-to-backend indices for an optional ordering map."""
+    if ordering is None:
+        return np.arange(item_count, dtype=np.int64)
+    return np.asarray(ordering.user_to_backend_indices, dtype=np.int64)
+
+
+def _make_backend_com_poses(num_instances: int, num_bodies: int) -> np.ndarray:
+    """Create distinct deterministic body COM transforms in backend order."""
+    coms = np.zeros((num_instances, num_bodies, 7), dtype=np.float32)
+    for env_index in range(num_instances):
+        for body_index in range(num_bodies):
+            value = float(100 * env_index + 10 * body_index)
+            coms[env_index, body_index, :3] = (value + 1.0, value + 2.0, value + 3.0)
+            coms[env_index, body_index, 6] = 1.0
+    return coms
+
+
+def _seed_backend_com_poses(backend: str, art, raw_backend, coms: np.ndarray) -> None:
+    """Seed backend-order COM transforms and invalidate both public and ordered caches."""
+    if backend == "physx":
+        raw_backend._coms = wp.array(coms, dtype=wp.float32, device="cpu")
+    elif backend == "ovphysx":
+        from isaaclab_ovphysx import tensor_types as TT
+
+        raw_backend.bindings[TT.BODY_COM_POSE]._data = coms.copy()
+    else:
+        raise AssertionError(f"Unsupported backend for COM parity test: {backend}")
+
+    art.data._body_com_pose_b.timestamp = -1.0
+    backend_staging = art.data._body_com_pose_b_backend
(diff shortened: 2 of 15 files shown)


source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+                shadow = getattr(owner, field_name)
+                if not expected_ordering:
+                    assert shadow is None, f"{backend} {ordering_mode} unexpectedly allocated {field_name}"
+                    continue
+                assert shadow is not None, f"{backend} reversed ordering did not allocate {field_name}"
+                item_count = num_bodies if "_body_" in field_name else num_joints
+                assert _ordering_shadow_shape(shadow)[:2] == (num_instances, item_count)
+
+    @pytest.mark.parametrize("ordering_mode", ["none", "identity", "reversed"])
+    def test_newton_allocates_shadows_only_for_nonidentity_ordering(self, ordering_mode: str) -> None:
```
ooctipus: Agent review: this Newton variant is a near-verbatim copy of the PhysX/OVPhysX allocation test above and the shadow tables already carry newton entries — the four allocation-shadow tests are one parametrized test. Relatedly, `test_identity_ordering_matches_none_behavior` and `test_physx_none_ordering_allocates_no_external_wrench_staging_buffers` are strict subsets of this family's identity/none cases and can fold in.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+        ],
+    )
+    def test_nonidentity_body_ordering_avoids_redundant_backend_staging(
+        self,
+        backend: str,
+        redundant_field: str,
+        retained_field: str,
+    ) -> None:
+        """Allocate one staging destination for each ordered body read path."""
+        if backend not in BACKENDS:
```
ooctipus: Agent suggestion: 27 tests open with an in-body backend-availability `pytest.skip(...)` like this one, where the module already defines a marker convention for exactly this — sweeping them to the marker trims ~2 lines each and makes the skips visible at collection time.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+            num_joints=1,
+            num_bodies=num_bodies,
+            device="cpu",
+            body_ordering=_body_ordering_for_mode(ordering_mode, num_bodies),
+        )
+        backend_seed = _make_backend_com_poses(num_instances, num_bodies)
+        _seed_backend_com_poses("ovphysx", art, raw_backend, backend_seed)
+        np.testing.assert_array_equal(_read_backend_com_poses("ovphysx", art, raw_backend), backend_seed)
+
+        art.data._body_com_pose_b.timestamp = -1.0
```
ooctipus: Agent review: this cache re-invalidation right after the seed helper is a no-op (the helper already leaves the cache invalidated), and the surrounding `user_to_backend` if/else is re-derived inline at 4 more sites past the file's own `_ordering_user_to_backend` helper.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+        raw_backend._dof_positions = wp.array(position, dtype=wp.float32, device=art.device)
+        raw_backend._dof_velocities = wp.array(velocity, dtype=wp.float32, device=art.device)
+
+        def set_backend_rows(storage: wp.array, values: wp.array, indices: wp.array | None) -> None:
+            if indices is None:
+                storage.assign(values)
+                return
+            env_ids = indices.numpy()
+            # PhysX writers pass global staging rows; the generic mock setter expects compact rows.
+            source_ids = env_ids if values.shape[0] == art.num_instances else np.arange(indices.shape[0])
```
ooctipus: Agent review: the PhysX partial-write tests validate against this test-local shape-sniffing setter instead of the shared mock's setter — and the two have already drifted. Validating against a private reimplementation is exactly the mock-drift failure mode this cross-backend file exists to catch; better to fix and use the shared mock setter.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+        import importlib.util
+        import os
+        import types
+
+        os.environ["LD_PRELOAD"] = "ovphysx-test"
+        real_find_spec = importlib.util.find_spec
+        real_import = builtins.__import__
+
+        def find_spec(name, package=None):
+            if name == "isaaclab_ovphysx":
```
ooctipus: Agent review: this subprocess scenario fakes `find_spec` only for `isaaclab_ovphysx`, but the utils gate the backend on two specs (`isaaclab_ovphysx` AND the external `ovphysx` SDK). On machines without that SDK — exactly where this file's other ovphysx tests politely skip — the guarded branch short-circuits and this test hard-fails instead of skipping. Faking both specs makes it hermetic.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2517 @@
...
+class TestArticulationOrderingWriteParity:
+    """Test that partial joint writes preserve unselected backend state."""
+
+    @pytest.mark.parametrize("backend", ["physx", "ovphysx", "newton"])
+    @pytest.mark.parametrize("ordering_mode", ["none", "reversed"])
+    @pytest.mark.parametrize("selection", ["index", "mask"])
+    @pytest.mark.parametrize("operation", ["position", "velocity", "state"])
+    @pytest.mark.parametrize(
+        "coverage",
+        ["one_env_one_item", "all_envs_one_item", "one_env_all_items"],
```
ooctipus: Agent review: 36 of the 108 cases in this joint-write cartesian run the coverage mode the block's own comment declares redundant — dropping that mode deletes a third of the matrix at no coverage cost.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in 5279c4c8a6. The allocation checks are now one all-backend parametrized test, and the separate identity-ordering and PhysX no-wrench allocation subsets were removed. This commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in 5279c4c8a6. Backend availability is now expressed with collection-time skip parameters/markers across the file; the in-body availability skips were removed. This commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in 5279c4c8a6. The redundant cache invalidation after `_seed_backend_com_poses` was removed, and the repeated inline maps now use `_ordering_user_to_backend`. This commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in 398358d328. The shared Torch and Warp mock DOF setters now support both compact selected rows and full global staging rows, with position/velocity regression coverage verified red before the fix and green afterward. The P7-local setter override was removed. This commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: This duplicate subprocess regression was removed earlier in 012aa6e42c, so the non-hermetic single-spec fake no longer exists. The remaining backend-discovery coverage owns this behavior. The rebased commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in 5279c4c8a6. The redundant `one_env_one_item` mode was dropped; the matrix retains `all_envs_one_item` and `one_env_all_items`, reducing it from 108 to 72 cases without losing the partial-item/full-item branches. This commit is awaiting the next authorized push.

source/isaaclab/isaaclab/test/utils/articulation_ordering.py
```diff
@@ -5,10 +5,71 @@
...
+BRANCHING_PHYSX_JOINT_NAMES = ("left_shoulder", "right_shoulder", "left_elbow", "right_elbow")
+BRANCHING_MJWARP_JOINT_NAMES = ("left_shoulder", "left_elbow", "right_shoulder", "right_elbow")
+BRANCHING_PHYSX_BODY_NAMES = ("base", "left_upper", "right_upper", "left_tip", "right_tip")
+BRANCHING_MJWARP_BODY_NAMES = ("base", "left_upper", "left_tip", "right_upper", "right_tip")
+
+
+def articulation_ordering_branching_fixture_path() -> Path:
+    """Return the repository path to the shared branching articulation fixture."""
+    return (
+        Path(__file__).resolve().parents[4]
```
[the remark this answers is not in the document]
AntoineRichard: Keeping this unchanged. These are source-checkout-only tests: the repository runner discovers tests under `source/`, while the backend asset test and USD fixture live under `source/isaaclab_physx/test` and are not wheel package resources. The helper intentionally resolves that shared repository fixture, and `test_shared_articulation_ordering_fixtures_are_consistent` already asserts that the resolved path is a file. An installed-wheel caller for this top-level backend test does not exist.

source/isaaclab_ovphysx/test/test_mock_ovphysx_bindings.py
```diff
@@ -0,0 +1,58 @@
...
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Tests for mock OVPhysX tensor bindings."""
+
+import numpy as np
+import pytest
+import warp as wp
+from isaaclab_ovphysx.test.mock_interfaces import MockTensorBinding
```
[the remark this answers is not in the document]
AntoineRichard: Addressed locally in 7ba5f665ad. The test now gates on `pytest.importorskip("ovphysx.types", ...)` before importing `MockTensorBinding`. A simulated missing-wheel collection now skips the module instead of erroring; the affected CPU-only tests pass with the wheel installed. This commit is awaiting the next authorized push.

source/isaaclab/test/utils/test_articulation_ordering.py
```diff
@@ -55,3 +55,31 @@ def test_assert_articulation_ordering_trace_matches_canonicalizes_public_and_ada
     )
 
     helper.assert_articulation_ordering_trace_matches(identity_trace, reordered_trace, ("knee", "hip"))
+
+
+def test_shared_articulation_ordering_fixtures_are_consistent() -> None:
+    helper = importlib.import_module("isaaclab.test.utils.articulation_ordering")
+
+    assert (
```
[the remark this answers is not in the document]
AntoineRichard: Addressed locally in 7ba5f665ad. The Panda root-preserving body order and ANYmal-C PhysX joint order are now pinned with explicit ordered tuple literals rather than recomputing the helper definition or comparing sets. This commit is awaiting the next authorized push.

source/isaaclab/test/utils/test_articulation_ordering.py
```diff
@@ -55,3 +55,31 @@ def test_assert_articulation_ordering_trace_matches_canonicalizes_public_and_ada
...
+        "LF_HFE",
+        "LH_HFE",
+        "RF_HFE",
+        "RH_HFE",
+        "LF_KFE",
+        "LH_KFE",
+        "RF_KFE",
+        "RH_KFE",
+    }
+    assert set(helper.BRANCHING_PHYSX_JOINT_NAMES) == set(helper.BRANCHING_MJWARP_JOINT_NAMES)
```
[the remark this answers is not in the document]
AntoineRichard: Addressed locally in 7ba5f665ad. The PhysX and MJWarp branching joint/body conventions are each pinned to explicit ordered tuple literals. This is stronger than inequality alone and detects either convention collapsing or being permuted. This commit is awaiting the next authorized push.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2328 @@
...
+    art.cfg = art.cfg.replace(joint_ordering=tuple(reversed(art.backend_joint_names)))
+    # Re-resolve and re-stage the ordering maps exactly as backend initialization
+    # does after a config change installs a new ordering on an already-initialized
+    # articulation.
+    art._resolve_and_install_ordering_maps()
+    art._ordering_configure_backend_staging()
+    return np.asarray(art.joint_ordering.user_to_backend_indices, dtype=np.int64)
+
+
+def _joint_ordering_for_mode(mode: str, num_joints: int) -> tuple[str, ...] | None:
```
mmichelis, line 76: The tested orderings are quite trivial and self-inverse. Adding another ordering test case like a cycled one (from [1,0,2] to [2,1,0]) can cover more errors potentially, like already done in e.g. source/isaaclab/test/assets/test_articulation_ordering.py:1248.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2328 @@
...
+        art, _ = get_articulation(
+            backend,
+            num_instances=2,
+            num_joints=1,
+            num_bodies=3,
+            device="cpu",
+            body_ordering=("body_2", "body_1", "body_0"),
+        )
+
+        assert getattr(art.data, redundant_field, None) is None
```
mmichelis: Agent: This assertion cannot fail: both `redundant_field` names cannot exist. `_body_link_vel_w_backend` has zero occurrences repo-wide, and `_body_com_acc_w_backend` exists only on the OVPhysX data class but is asserted here against Newton. With the `, None` default both reduce to `assert None is None`.

Newton names these shadows with a `_user` suffix (`isaaclab_newton/.../articulation_data.py:1794`), which `retained_field` correctly uses -- so the two halves of the same tuple disagree on convention. Dropping the `, None` default would surface stale names as `AttributeError`, like the sibling test at `:1280`.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2328 @@
...
+        expected = np.asarray([[patterns[name] for name in art.joint_names]], dtype=np.float32)
+        np.testing.assert_array_equal(art.data.default_joint_pos.warp.numpy(), expected)
+
+    @_requires_ovphysx
+    def test_ovphysx_mock_shell_mirrors_write_capabilities(self) -> None:
+        """The OVPhysX test shell mirrors the write bindings installed by initialization."""
+
+        art, _ = get_articulation("ovphysx", device="cpu")
+
+        assert art._can_write_effort is True
```
mmichelis: Agent: This asserts the fixture against itself. The only change this PR makes to `_articulation_iface_test_utils.py` is hardcoding these same three values at `:301-303`, so no product code is on the path.

The real derivation is `_get_binding(...) is not None` (`isaaclab_ovphysx/.../articulation.py:4015-4017`), and the mock already registers all three tensor types (`mock_ovphysx_bindings.py:391-393`) -- driving the real path would give the same values while actually exercising it.

Related: `_has_implicit_actuators` is `False` throughout this fixture and gates `write_pos`/`write_vel` (`articulation.py:266-267`), so the OVPhysX position- and velocity-target reorder paths are never exercised.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2328 @@
...
+            )
+        elif operation == "velocity":
+            art.write_joint_velocity_to_sim_index(
+                velocity=velocity,
+                env_ids=env_ids_wp,
+                joint_ids=joint_ids_wp,
+            )
+        elif backend == "ovphysx":
+            with pytest.deprecated_call():
+                art.write_joint_state_to_sim(
```
mmichelis: Agent: `write_joint_state_to_sim` is deprecated (`isaaclab_ovphysx/.../articulation.py:4698`) and just delegates to the position/velocity index writers already covered by `operation="position"`/`"velocity"`, so these `state`+`index` cases add no OVPhysX coverage.

OVPhysX does implement `write_joint_state_to_sim_index` with an identical keyword-only signature (`:872`), and that is the path with the fused write buffer. The mask branch at `:334` already treats all three backends uniformly, so this looks unintentional. Same at `:537-551`.

source/isaaclab_physx/test/test_mock_interfaces/test_mock_articulation_view.py
```diff
@@ -221,6 +221,23 @@ def test_set_dof_positions_with_indices(self, view):
...
         assert torch.allclose(result[2], new_data[1])
 
+    @pytest.mark.parametrize(
+        ("setter_name", "getter_name"),
+        [("set_dof_positions", "get_dof_positions"), ("set_dof_velocities", "get_dof_velocities")],
+    )
+    def test_set_dof_state_with_indices_uses_matching_global_rows(self, view, setter_name, getter_name):
+        """Test indexed writes from global inputs select the matching source rows."""
+        baseline = torch.zeros(4, 12)
+        getattr(view, setter_name)(baseline)
```
mmichelis: Agent: `baseline` aliases the mock storage: `set_dof_positions` keeps the tensor by reference when `indices is None` (`mock_articulation_view.py:409`). The indexed write then mutates `baseline`, and `expected = baseline.clone()` clones the already-mutated buffer, so the value assertions cannot fail.

I injected a bug that also clobbers unselected row 1 and this test still passed; passing `baseline.clone()` in makes it correctly fail. It does fail pre-fix (the old setter raised), so it pins that the write works, just not which rows. The warp twin at `..._warp.py:275` is sound because `wp.array(numpy)` copies.

source/isaaclab_physx/isaaclab_physx/test/mock_interfaces/views/mock_articulation_view_warp.py
```diff
@@ -533,7 +534,7 @@ def set_dof_positions(
             self._dof_positions = wp.zeros((self._count, self._num_dofs), dtype=wp.float32, device=self._device)
         if indices is not None:
             wp.launch(
-                scatter_floats_2d,
+                scatter_floats_2d_global_rows if positions.shape[0] == self._count else scatter_floats_2d,
```
mmichelis: Agent: The global-rows fix landed on the DOF setters only. Production passes full buffers to the root setters too (`isaaclab_physx/.../articulation.py:557,650,814,917`), but `set_root_transforms`/`set_root_velocities` here (`:485`, `:510`) still launch plain `scatter_floats_2d`, which with global rows plus indices silently writes the wrong rows. The torch twin (`mock_articulation_view.py:362,383`) raises instead. Both docstrings advertise `(N, 7) or (len(indices), 7)` but implement only the compact form.

Also, dispatching on `shape[0] == self._count` is ambiguous when a caller writes every env via an explicit index list (`env_ids = arange(n)`, a common reset path); committing to global-rows unconditionally would remove that.

mmichelis (dismissed):
Very thorough ordering tests! I added a few agent comments, but the one thing that could improve the testing is to add another test mode to cover cyclical permutations of body ordering, just so it is not the trivial reverse ordering tests which might not catch all errors.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in `a4cfde1b2d` (not pushed yet). The existing body-property read/write tests now run both reversed and cyclic ordering modes, with the cyclic case using a non-self-inverse mapping. I also mutation-checked the map direction: the reversed cases still passed with the intentional inversion, while every available cyclic case failed, confirming the added coverage catches that class of error.

AntoineRichard: Now pushed in `a4cfde1b2d`. The existing body-property read and partial-write matrices now run both reversed and cyclic modes; for three bodies the cyclic case is `(body_2, body_0, body_1)`, so the mapping is not self-inverse. The direction-inversion mutation remained invisible to the reversed cases and failed all available cyclic cases.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Confirmed as a test-quality observation: the `getattr(..., None)` default means a stale member name is indistinguishable from an intentionally absent staging allocation. Mike identified cyclic coverage as the requested change in this review, so I left this diagnostic assertion unchanged in the cycle-only update. A focused follow-up can tighten it by asserting the concrete allocation member without a default.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Confirmed. This shell bypasses `_initialize_impl`, so the three capability flags are fixture wiring rather than a test of binding discovery; the standalone assertion therefore only validates that wiring. Also, `_has_implicit_actuators = False` means this fixture does not cover position/velocity target forwarding through `write_data_to_sim`. Those are separate fixture/integration coverage improvements and were left unchanged in the cycle-only update requested by the review.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Confirmed. The OVPhysX special case calls the deprecated combined API, which delegates to the already-covered position and velocity index writers; switching both helper sites to `write_joint_state_to_sim_index` is the focused way to exercise the fused write path. This is orthogonal to the requested cyclic body-order coverage and was left unchanged in this update.

source/isaaclab_physx/test/test_mock_interfaces/test_mock_articulation_view.py
AntoineRichard: Confirmed from the mock storage behavior: the no-index setter retains `baseline` by reference, so the later indexed write mutates the expected-value source. Passing `baseline.clone()` into the initial setter is the minimal test repair. This is a mock-test robustness follow-up rather than a production ordering failure, and it was left unchanged in the cycle-only update.

source/isaaclab_physx/isaaclab_physx/test/mock_interfaces/views/mock_articulation_view_warp.py
AntoineRichard: Confirmed that the Warp root setters still use compact-row scatter while production can pass full global buffers. The suggested unconditional global-row interpretation is not safe by itself because these mock APIs also document and accept compact `(len(indices), width)` inputs; the repair needs an explicit source-row contract plus Torch/Warp parity coverage, especially for the ambiguous full explicit index list. This broader mock-setter follow-up was left unchanged in the cycle-only update.

AntoineRichard:
@mmichelis The non-self-inverse cyclic ordering coverage requested in your review is now pushed in `a4cfde1b2d`. I replied inline to each agent thread with its current disposition; those five broader test/mock-quality observations remain scoped follow-ups and were not silently folded into this cycle-only update. Could you re-review the requested change when you have a chance?

source/isaaclab/test/assets/test_articulation_ordering_iface.py
mmichelis: Agent: The `cyclic` permutation itself is correct — n=4 gives `u2b=(3,0,1,2)` and `b2u=(1,2,3,0)`, genuinely non-self-inverse. But it landed on the two tests that only read the forward map, so the gap is still open.

Re-ran the mutation on `a4cfde1b`: replacing `backend_to_user`/`backend_to_user_indices` with the forward map at `ordering.py:226,228` still leaves the suite fully green (127 passed, identical to clean).

Only three sites consume the inverse map — `:2024-2025`, `:2085-2086`, and the sanity check at `:1653` — and all three still build `reversed(...)` orderings. They also derive the expected value from `art.*_ordering.backend_to_user_indices`, i.e. the same map under test, so a cyclic ordering alone wouldn't catch it there either.

Two changes would close it: add `cyclic` to `_joint_ordering_for_mode` and route it into those sites, and compute the expected permutation from the declared ordering tuple rather than reading it back off the articulation. Self-check: with that mutation applied, the suite should go red.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in `ee9160715d` (not pushed yet). I removed the vacuous absence test rather than adding dummy fields. `TestArticulationOrderingAllocation` remains the canonical exact inventory: it accesses every concrete shadow without a default and verifies allocation/non-allocation for identity and nonidentity mappings.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in `ee9160715d` (not pushed yet). The shell now derives all three capability flags through `_get_binding(...)`, and the self-referential capability assertion was replaced by a behavioral test that enables implicit actuators, calls `write_data_to_sim`, and verifies reordered position and velocity targets in the real mock bindings.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
AntoineRichard: Addressed locally in `ee9160715d` (not pushed yet). Both OVPhysX deprecated special cases were removed; index-based state cases now call `write_joint_state_to_sim_index` uniformly across backends and exercise the fused OVPhysX buffer path. The focused OVPhysX state matrix passes 9 cases.

source/isaaclab_physx/test/test_mock_interfaces/test_mock_articulation_view.py
AntoineRichard: Addressed locally in `ee9160715d` (not pushed yet) by cloning the baseline before installing it in the Torch mock. Mutation check: the old aliased setup passed both cases despite an injected row-1 clobber; with the clone, both cases failed; after restoring clean code, both pass.

source/isaaclab_physx/isaaclab_physx/test/mock_interfaces/views/mock_articulation_view_warp.py
AntoineRichard: Addressed locally in `ee9160715d` (not pushed yet). Indexed root and DOF setters now consistently consume full global buffers in both Torch and Warp mocks, matching every production call site. Compact-only mock tests/docstrings were removed, root global-row regressions were added, and the full two-file mock suite passes 87 tests.

source/isaaclab/isaaclab/test/utils/articulation_ordering.py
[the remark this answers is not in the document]
hujc7: **[P1] Agent Review:** Reopening with a different argument than the bot's packaging one — this coupling was deliberately removed five days ago.

[#6675 — "Remove isaaclab_physx cross-package dependency from isaaclab_newton tests"](https://github.com/isaac-sim/IsaacLab/pull/6675) (merged 07-23) states: *"Two ordering tests resolved `articulation_ordering_branching.usda` via a hardcoded path that traversed into `isaaclab_physx/test/assets/data/`. The fixture is copied into `isaaclab_newton/test/assets/data/` so each package owns its test data."*

`articulation_ordering_branching_fixture_path()` reinstates that traversal and promotes it from a test-local helper into `isaaclab` core. The PhysX test regressed specifically — `test_branching_fixture_resolves_distinct_conventions` used `Path(__file__).parent / "data" / …` (correct, own package) and now round-trips through core to reach its own directory. Net: Newton follows #6675, PhysX and OVPhysX no longer do.

Suggested: drop the helper, restore local resolution in the PhysX test, copy the `.usda` into `isaaclab_ovphysx/test/assets/data/`. Name constants can stay shared — plain tuples carry no packaging coupling.

source/isaaclab/test/assets/test_articulation_ordering_iface.py
```diff
@@ -0,0 +1,2307 @@
...
+            "mass": _clone_backend_tensor(art.data._sim_bind_body_mass),
+            "inertia": _clone_backend_tensor(art.data._sim_bind_body_inertia),
+        }
+    raise AssertionError(f"Unsupported backend for body-property ordering test: {backend}")
+
+
+def _backend_param(backend: str, *values, **kwargs) -> ParameterSet:
+    """Build a backend parameter that skips unavailable plugins at collection time."""
+    marks = list(kwargs.pop("marks", ()))
+    marks.append(pytest.mark.skipif(backend not in BACKENDS, reason=f"{backend} backend is not available"))
```
hujc7: **[P2] Agent Review:** This reason string (and `_requires_physx`/`_requires_ovphysx`/`_requires_newton` at `:879-881`) reports "backend is not available" without saying why, and `_articulation_iface_test_utils.py:23-29` discards the cause with `except ImportError: pass`. Running kitless:

```
$ uv run --extra test python -m pytest source/isaaclab/test/assets/test_articulation_ordering_iface.py -q
53 passed, 146 skipped        # BACKENDS = ['Mock', 'newton']
```

PhysX is gone because `isaaclab_physx.assets.articulation.articulation` raises `ModuleNotFoundError: No module named 'carb'` — but the skip text suggests the package isn't installed, and it is. For a cross-backend parity suite, that's a green run that exercised one backend.

Capturing the exception in the utils and interpolating it here would make the skip self-explanatory:

```python
except ImportError as exc:
    _PHYSX_UNAVAILABLE = f"PhysX backend unavailable: {exc}"
```

source/isaaclab_ovphysx/test/assets/test_articulation.py
```diff
@@ -725,18 +686,16 @@ def test_branching_fixture_physx_ordering_is_identity_on_ovphysx(sim, device):
     assert articulation.is_initialized
 
     # Ground truth pinned by isaaclab_physx's test_branching_fixture_resolves_distinct_conventions.
```
hujc7: **[P3] Agent Review:** This comment and `:725` (plus `isaaclab_newton/test/assets/test_articulation.py:636`) now sit above a blank line — the locals they described moved into the shared module, and the ground truth is `isaaclab.test.utils.articulation_ordering`, not the PhysX test. Delete or repoint.

hujc7 (approved):

source/isaaclab/test/assets/test_articulation_ordering_iface.py
hujc7: **[P1] Agent Review:** Verified at HEAD (`ee9160715d3`) — still open on all three points, and the latest commit added a fourth instance of the pattern.

- `_joint_ordering_for_mode` (`:76-85`) still rejects `cyclic`; every joint ordering in the file remains self-inverse.
- `:2023` and `:2060` still build the expectation from `art.joint_ordering.backend_to_user_indices` — the map under test.
- `ee9160715d3` added `test_ovphysx_implicit_targets_are_written_in_backend_order` (`:2003`) with both properties: `reversed` at `:2013`, `position[:, backend_to_user]` at `:2026`. It is blind to the same mutation.

For n=3 the two defects compound: reversed gives `u2b == b2u == (2,1,0)`, so swapping the maps is a no-op; and because the expectation is sourced from the map under test, a corrupted map is mirrored into both sides of the assertion and cancels out regardless of ordering mode. Fixing either alone leaves the mutation alive.

Four edits:

1. `_joint_ordering_for_mode` — add the branch its body twin already has:
```python
if mode == "cyclic":
    return (backend_names[-1], *backend_names[:-1])
```
2. Generalise `_install_reversed_joint_ordering` (`:65`) to take a mode — `:2040` reaches the inverse map through it.
3. Add a helper deriving both maps from the declared tuple only, mirroring `ordering.py:220-222` but never reading the articulation:
```python
def _declared_ordering_maps(ordering, backend_names):
    count = len(backend_names)
    if ordering is None:
        identity = np.arange(count, dtype=np.int64)
        return identity, identity
    user_to_backend = np.asarray([backend_names.index(n) for n in ordering], dtype=np.int64)
    backend_to_user = np.empty_like(user_to_backend)
    backend_to_user[user_to_backend] = np.arange(count, dtype=np.int64)
    return user_to_backend, backend_to_user
```
4. Parametrise the four inverse-map sites (`:1608`, `:1972`, `:2023`, `:2060`) over `["reversed", "cyclic"]` and swap in the derived map:
```python
-backend_to_user = np.asarray(art.joint_ordering.backend_to_user_indices, dtype=np.int64)
+_, backend_to_user = _declared_ordering_maps(joint_ordering, tuple(art.backend_joint_names))
```

Two constraints: cyclic needs n >= 3 (for two items a cycle *is* a swap and degenerates to reversed — the joint sites are n=3 and n=4, so fine), and a mode-aware body variant needs `_install_test_body_ordering`'s `is_fixed_base` guard at `:52-55`, since cyclic moves body 0 out of the root slot.

Self-check as you suggested: with `ordering.py:226,228` mutated to the forward map, the suite should go red — it was green at 127 on `a4cfde1b`, and nothing in `ee9160715d3` touches the ordering-mode helpers.

ooctipus (approved):

kellyguo11 (approved):

PR #6346, opened 2026-07-03
[IsaacLab CI] Run only necessary tests as pre-merge, run full test as post merge

mataylor-nvidia:
# Description

This PR reduces pre-merge CI time by using `pytest-testmon` coverage data to select tests affected by Python source changes.

Read about how testmon works: https://www.testmon.org/

pytest-testmon 2.x only records coverage inside the pytest process, so tests that exercise code by spawning child Python processes never have those dependencies tracked and are not reselected when the child-only code changes.

Add tools/testmon_subprocess_coverage.py and its testmon_coverage_context coverage plugin, which install a process_startup .pth shim, tag child coverage with the active test's node id, and merge child coverage back into Testmon before each report. Wire the hooks in via a repository-root conftest.py; they are no-ops unless Testmon is collecting. Include a regression test that verifies a subprocess-only dependency edit reselects the test.

Post-merge CI runs the full test suite and refreshes the Testmon dependency database for future pull requests.

The previous manually maintained test-subset configuration and exact node-ID selection plumbing have been removed.

`.testmon.db` artifact is uploaded so we can findout when certain tests weren't run and why if things fail

Smoke tests should still be run on all PR pipelines, added additional smoke tests for completeness.

Report on test time decrease compiled by agent:

[testmon_report.md](https://github.com/user-attachments/files/29752545/testmon_report.md)

Dependencies:

- Adds `pytest-testmon`, `coverage` as a CI-only dependency.
- No new runtime dependencies.

Fixes # ([OMPE-99210](https://jirasw.nvidia.com/browse/OMPE-99210))

## Type of change

## Screenshots

## Checklist


mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

.github/actions/run-tests/action.yml
```diff
@@ -126,11 +125,11 @@ runs:
           local shard_count="${13}"
           local volume_mount_source="${14}"
           local extra_pip_packages="${15}"
-          local test_node_ids_file="${16}"
```
mataylor-nvidia, line 129: `test_node_ids_file` is removed because all tests run post merge

we no longer only run a subset of the tests post-merge all are run

mataylor-nvidia:
@greptile review

tools/testmon_subprocess_coverage.py
mataylor-nvidia, line 1: This is a shim for test coverage in subprocesses. kit is usually launched in a separate process so we need to support searching in these files

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

conftest.py
nvsekkin, line 1: nit: the marker PR also adds a conftest at the root level, this will most likely need to be rebased before we merge.

mataylor-nvidia:
@greptile review

.github/actions/run-tests/action.yml
```diff
@@ -273,6 +255,24 @@ runs:
             docker_env_vars="$docker_env_vars -e TEST_EXTRA_PIP_PACKAGES"
           fi
 
+          testmon_options=""
+          if [ -n "$testmon_data_dir" ]; then
+            mkdir -p "$testmon_data_dir"
+            docker_env_vars="$docker_env_vars -e TESTMON_DATAFILE=/workspace/isaaclab/$testmon_data_dir/.testmondata"
```
nvsekkin, line 260: suggestion (non-blocking): we may use `ISAACLAB_PATH_ARG` instead of the hardcoded `/workplace/isaaclab`.

Though this is unlikely to change, so we can leave it as hardcoded

conftest.py
```diff
@@ -0,0 +1,47 @@
...
+from collections.abc import Generator
+from pathlib import Path
+
+import pytest
+
+
+@pytest.hookimpl(wrapper=True, tryfirst=True)
+def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]) -> Generator[None, None, None]:
+    """Keep tests marked ``smoke`` selected when Testmon filters the collection."""
+    always_items = [item for item in items if item.get_closest_marker("smoke") is not None]
```
nvsekkin, line 46: note: this is good but today we do not have the `pytest.mark.smoke` marker, this marker is introduced in the pytest marker PR, which should land before this one. Also, we should consider marking a subset of the tests as `smoke`

mataylor-nvidia: note: smoke test marker landed on friday see below, which prompted the other markers PR

https://github.com/isaac-sim/IsaacLab/pull/6347

nvsekkin: great thank you for the heads up, I was looking at a stale local version 😅

.github/workflows/build.yaml
```diff
@@ -255,7 +254,6 @@ jobs:
     runs-on: [self-hosted, gpu]
     needs: [changes, config]
     if: >-
-      github.event_name != 'push' &&
```
nvsekkin, line 258: note: This is obviously the intent of this PR and the smoke tests/change detection will be used for the rest of the pull requests. Since by removing this trigger we will be using a lot more resources on push commit. (and less for each pull request)

It is still important to broadcast this change to a broad audience @mataylor-nvidia

mataylor-nvidia: Not all commits are pushed to develop so this is okay, this should in theory make the runners work less

mataylor-nvidia: I have included a markdown report where I have forced select mode to show perf implications and show deselection of tests is working.

hujc7:
Agent review: verified the machinery (shipped regression test 2/2; counterfactual with the shim removed confirms testmon does NOT reselect subprocess-only edits without it — the shim is genuinely load-bearing). Core design fails toward running more tests. Findings by severity:

1. **install-ci: the DB never persists, and selection there would be unsound even if it did.** On this PR's own runs (both arches) no `.testmondata` lands on the host — the upload warns `No files were found` and an empty cache is re-saved, so select mode permanently falls back to full baseline. Couldn't reproduce locally (same pytest/testmon versions write the DB fine), needs live debugging. Separately, rootdir inside the container is the install_ci dir, so the repo-root conftest (shim + smoke re-add) never loads and the real work runs in fresh coverage-less envs — if DB persistence alone gets fixed, Python-only PRs would deselect every install test. Suggest dropping testmon from install-ci here and reworking separately.
2. **Add fail-closed rules:** changed `.py` with zero recorded dependents in the DB (import-time-only code, e.g. `gym.register` in `__init__.py`) → collect; tests with home-file-only fingerprints (skipped at baseline) → never deselectable.
3. **Smoke re-add is too strong and too weak (reproduced):** with `-k normal` it ran `test_smoke` anyway (overrides `-k`; will misfire on device-split passes), and with a directory arg testmon skips whole files pre-collection so nothing is re-added ("no tests ran" despite a smoke test). Re-add only testmon-deselected items and assert the per-file invocation form.
4. **The new unit tests never run in CI** (mode selector, summary assembler, selection graph — 23 tests, pass locally, referenced by no workflow; the tools/ one needs `--noconftest`). Wire into a lightweight job.
5. **`VERSION` gap:** a VERSION-only PR yields `select` → zero tests, though install-ci's trigger list names `^VERSION$`. One-line regex fix.
6. **Cache hygiene:** no self-heal for a corrupt restored DB (re-restored until a hashed file changes); run-package-tests cache key omits `tools/testmon_coverage_context.py` (install-ci's includes it); noted risk: patch-level dep drift is invisible to both the key and testmon's env fingerprint.
7. **Rollout: measure the false-negative rate instead of assuming zero** — shadow mode first, post-merge escape attribution, a small random full-suite control group, and a `ci:full-suite` escape-hatch label.
8. **Description accuracy:** the report is a job summary + artifact (not a PR comment), and the `.testmon.db` upload only exists for install-ci — where it currently uploads nothing (#1).

PR #6351, opened 2026-07-04
Complete Newton state write boundaries

ooctipus:
# Description

This completes the authored rigid-state boundary contract on top of #6350.

State writers now separate two responsibilities:

- `skip_forward=True` preserves the caller-managed asset-cache behavior.
- The physics write is always published to the shared transaction, so FK and mandatory backend synchronization still happen before the next step or coherent read.

Partial `RigidObjectCollection` writes preserve their selected body columns when marking articulations, including a no-op fast path for empty selections. Direct consumers of live Newton body state—site frames, PVA, ray-caster site transforms, and joint-wrench conversion—flush pending synchronization before reading.

This PR does not add optional solver reset, particle-state publication, or coupled deformable synchronization.

Depends on #6350. The commit is intentionally isolated so the displayed diff will reduce to this boundary layer once the dependency merges.

## Type of change

- Refactor (non-breaking change)

## Testing

- Foundation + boundary unit suites — 40 passed.
- Newton site-frame contract suite — 46 passed.
- PVA and joint-wrench stale-state integrations — 2 passed.
- Targeted pre-commit hooks, `compileall`, and `git diff --check` pass.

## Checklist

- [x] I have added tests for the changed behavior

### DIFF
--- a/source/isaaclab_newton/changelog.d/state-boundaries.rst
+++ b/source/isaaclab_newton/changelog.d/state-boundaries.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Ensured rigid state writes are published even when asset cache invalidation
+  is skipped, and direct body-state readers flush pending synchronization.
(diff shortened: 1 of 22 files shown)


ooctipus:
Closing this implementation because it depends on the discarded authored-state transaction foundation. The proven NewtonSiteFrameView freshness issue will be handled separately as a minimal boundary fix.

PR #6352, opened 2026-07-04
Synchronize coupled Newton solver state

ooctipus:
## Summary

- make the real rigid and deformable solvers explicit owners of model-change notifications
- rebase rigid finite-difference history only for the authored articulation selection, while still covering global free bodies
- align coupled MuJoCo sparse handoff and CUDA graph policy with standalone operation
- clear the newly explicit solver and history references during teardown

## Dependencies

- Depends on #6350 and #6351.
- This layer is intentionally solver-reset-free and particle-state-free.

## Testing

- exact pinned Newton dependency: 51 state-pipeline, boundary, and coupled-state-sync tests passed
- targeted pre-commit checks passed
- compileall and diff checks passed

### DIFF
--- a/source/isaaclab_contrib/changelog.d/newton-state-sync.rst
+++ b/source/isaaclab_contrib/changelog.d/newton-state-sync.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Synchronized authored rigid state with coupled Newton solvers and rebased
+  finite-difference history only for selected articulations.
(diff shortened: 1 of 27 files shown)


ooctipus:
Closing this implementation because its solver-ownership and synchronization hooks depend on the discarded transaction foundation. The narrow two-way coupled state-history issue can be revisited independently with an application regression.

PR #6349, opened 2026-07-04
Add opt-in Newton solver reset

ooctipus:
# Description

Adds `NewtonCfg.solver_reset`, disabled by default, to clear solver-owned state after task-authored environment writes. At each coherent boundary, pending model changes are delivered first, forward kinematics and mandatory backend synchronization run next, and the optional reset then clears solver-owned state for dirty worlds with `flags=0`. Task-authored positions and velocities therefore remain unchanged.

Kamino keeps its mandatory solver-specific synchronization regardless of this option and skips the second generic reset. Solvers that inherit Newton's no-op reset hook are unaffected. The option is a direct `NewtonCfg` attribute; it does not introduce another configuration object.

## Dependencies

- Depends on #6350, #6351, and #6352.
- GitHub shows their cumulative diff here until those PRs land. The reset-policy commit itself is limited to the configuration, shared/Kamino manager policy, focused tests, and changelog.
- Particle-state synchronization is outside this PR.

## Type of change

- New feature (non-breaking change which adds functionality)

## Screenshots

Not applicable.

## Testing

- Exact pinned Newton dependency: 57 state-pipeline, boundary, coupled-state, and reset-policy tests passed.
- GPU MJWarp integration passed with reset both disabled and enabled.
- Kamino mandatory one-shot synchronization integration passed.
- Single-world CPU MuJoCo reset integration passed.
- Targeted pre-commit, compile, and diff checks passed.
- `./isaaclab.sh -f` passed source and RST checks; its checkout-wide LFS hook reports golden-image pointer metadata outside this diff.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6353, opened 2026-07-04
Synchronize authored Newton particle state

ooctipus:
## Summary

- add a world-only particle path to the shared authored-state transaction, preserving exact mask/index selection without requesting rigid forward kinematics
- publish every explicit MPM and contrib deformable state, position, and velocity write through that path
- re-arm particle rendering when device writes replay from a CUDA graph
- keep per-step deformable kinematic-target application on the control path, where it refreshes caches and rendering without creating optional-reset work

Current MPM and VBD reset hooks remain Newton no-ops; this PR does not claim to clear their deformation or warm-start history. It provides the correct selected worlds to owned solver hooks when they have meaningful reset behavior.

## Dependencies

- Depends on #6350, #6351, #6352, and #6349.
- GitHub shows the cumulative stack here until those PRs land. This PR's final commit is confined to particle transaction/rendering behavior, MPM/deformable writers, focused tests, and two package changelogs.

## Testing

- Exact pinned Newton dependency: 75 state-pipeline, boundary, coupled-state, reset-policy, and particle-state tests passed.
- Existing MPM object integration suite: 8 passed against the exact pin using process-local logging symbols expected by that pin's newer Warp API; no compatibility workaround is included in source.
- Targeted pre-commit checks passed for all eight files.
- Compile and diff checks passed.

## Checklist

- [x] I have run the [`pre-commit` checks](https://pre-commit.com/)

### DIFF
--- a/source/isaaclab_contrib/changelog.d/newton-state-sync.rst
+++ b/source/isaaclab_contrib/changelog.d/newton-state-sync.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Synchronized authored rigid state with coupled Newton solvers and rebased
+  finite-difference history only for selected articulations.
--- a/source/isaaclab_contrib/changelog.d/particle-state-sync.rst
+++ b/source/isaaclab_contrib/changelog.d/particle-state-sync.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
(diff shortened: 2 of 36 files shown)


ooctipus:
Closing this implementation because MPM and VBD already receive authored particle state in both Newton state buffers, while their reset hooks are no-ops. Any CUDA-replayed particle-rendering issue should be handled as a separate focused fix.

PR #4605, opened 2026-02-12
Adds support for IsaacTeleop and refactors/deprecates existing XR classes

rwiltz:
# Description
Adds new isaaclab_teleop extension providing IsaacTeleopDevice, a unified teleoperation device wrapping IsaacTeleop's pipeline-based retargeting engine, OpenXR session lifecycle, XR anchor synchronization, and action-tensor generation behind a context-manager interface.
Deprecates OpenXRDevice, OpenXRDeviceCfg, RetargeterBase, and RetargeterCfg in isaaclab.devices in favor of isaaclab_teleop.IsaacTeleopDevice / IsaacTeleopCfg.
Updates teleop_se3_agent.py and record_demos.py to auto-detect env_cfg.isaac_teleop and use the IsaacTeleop stack, with full backward compatibility for the native teleop stack.

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6358, opened 2026-07-05
Fix XrCfg.near_plane being ignored in AR/VR profile XR sessions

asierarranz:
## Description

In XR teleoperation you naturally lean in to look at things up close — and today anything closer than ~15 cm disappears, no matter what `near_plane` you configure. That breaks close-up inspection in CloudXR sessions.

**Cause:** Isaac Lab writes the generic `/persistent/xr/render/nearPlane`, but the profile-specific settings (`/persistent/xr/profile/{ar|vr}/render/nearPlane`) default to `0.15` (set in `apps/isaaclab.python.xr.openxr.kit`) and take precedence — the configured value never reaches profile-based sessions.

## Fix

- Write the `ar`/`vr` profile settings alongside the generic one, in `XrAnchorManager` and in the deprecated `OpenXRDevice` (same bug).
- `test_xr_anchor` now uses a non-default `near_plane=0.05` and asserts all three settings paths.
- Verified on a Quest 3 + CloudXR setup: near clipping now matches the configured value.

No restore-on-teardown, matching how this code path already handles its other persistent settings. Both profiles are written so behavior is deterministic regardless of the active profile.

## Type of change

## Checklist

- [x] I have made corresponding changes to the documentation (changelog fragment; no doc pages affected)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -56,6 +56,7 @@ Guidelines for modifications:
 * Anton Bjørndahl Mortensen
 * Antonin Raffin
 * Arjun Bhardwaj
+* Asier Arranz
 * Ashwin Varghese Kuruttukulam
 * Bikram Pandit
 * Bingjie Tang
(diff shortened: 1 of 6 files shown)


asierarranz:
Good catch, though half of it was already in: `openxr_device.py` got the profile-aware writes in 395fa13 (the review ran on 65729ab, right before that push). Just added the same two writes to `ManusVive` in af42b84, so every OpenXR device path handles `XrCfg.near_plane` the same way. Worth noting ManusVive has been deprecated since #4605 and nothing in the repo instantiates it anymore, but two lines are cheaper than an inconsistency.

PR #6360, opened 2026-07-05
Newton Warp renderer integration support for distance_to_camera and distance_to_image_plane

rilei-nvidia:
# Description

Add distance annotators to Newton Warp renderer

- Add `distance_to_camera` and `distance_to_image_plane` support to NewtonWarpRenderer. 
- Also fix the `depth` data type to return `distance_to_image_plane` to match the Isaac Lab camera contract

## Type of change

- Fix

## Checklist


source/isaaclab_tasks/test/golden_images/dexsuite_kuka_hetero/newton-newton_renderer-depth.png
huidongc, line 1: There might be an issue here that camera near/far clipping plane is not passed to the render backend so it always takes 1000.0 as the far clip and 0.0 as the near clip, making the depth of objects nondistinguishable.

@daniela-hase for vis.

huidongc: If I reduce the far clipping plane to 2.5 I get this image which looks much better:

Image

rilei-nvidia: Indeed we are not passing the clipping range to newton sensor's render config.  

I am able to pass the far clipping plane into the RenderConfig's max_distance and get the results similar to  your screenshot @huidongc:
image

However Newton sensor's RenderConfig does not take a near plane / min distance.  Is this a known limitation @daniela-hase?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -149,20 +205,53 @@ def set_outputs(self, output_data: dict[str, ProxyArray]):
...
+        Newton writes ``0.0`` for rays that miss all geometry or fall beyond the far plane
+        (``max_distance``), so ``"none"`` and ``"zero"`` both leave that ``0.0`` background. ``"max"``
+        replaces the background with the far clip [m] to mirror the RTX renderer's
+        :attr:`~isaaclab_physx.renderers.IsaacRtxRendererCfg.depth_clipping_behavior`. No-op when no
+        depth output was requested or the camera did not provide a clipping range.
+        """
+        if behavior != "max" or self.far_clip is None:
+            return
+        for dest in self._depth_dests.values():
+            replace_background_depth_wp(dest, self.far_clip, device=dest.device)
```
daniela-hase, line 253: If I understand correctly, this just replaces all depth values `<= 0.0` with `self.far_clip`?

This can probably be replaced by just passing a different clear-value for the depth output to the renderer:
```python
clear_data=newton.sensors.SensorTiledCamera.ClearData(clear_color=0xFFEEEEEE, clear_depth=self.far_clip),
```

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -301,7 +392,15 @@ def prepare_stage(self, stage: Any, num_envs: int) -> None:
...
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.create_render_data`."""
-        return RenderData(self.newton_sensor, spec)
+        render_data = RenderData(self.newton_sensor, spec)
+        # Honor the camera's far clipping plane (``spawn.clipping_range[1]``). ``max_distance`` is
+        # baked into the sensor at construction from the renderer cfg default; the per-camera far
+        # plane is the authoritative value, so push it onto the live render config (documented as
+        # mutable for subsequent ``update`` calls). Rays beyond it return no hit and the depth buffer
+        # keeps its ``0.0`` background sentinel.
+        if render_data.far_clip is not None:
+            self.newton_sensor.render_config.max_distance = render_data.far_clip
```
daniela-hase: `render_config` currently is a state-machine across all NewtonRenderers, so changing it here will change it for all other sensors as well if the user has more than one.

This is badly documented in Newton, and I'll probably go and fix it to not be a state machine anymore. But in the meantime, probably better to set this config on every render call.

rilei-nvidia: Noted, will fix, and thanks for updating the RenderConfig in Newton to not be a state machine!

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: Good to know about this.  I will update it to pass in the far_clip value to the clear_depth arg when rendering the `distance_to_camera` annotator.

When the user requests `distance_to_image_plane` however, we can not apply use this approach as the convert_plane_depth is called as a post-processs after newton_sensor.update.

If the Newton Warp renderer eventually supports distance-to-image-plane natively we can clean all this logic up!

daniela-hase: Why do we still need to do it for the `distance_to_image_plane`? My understanding is that his here just replaces all values `<= 0.0` with `self.far_clip` - so if we change the default depth clear value to `self.far_clip`, then all the depth values where no ray hit will be `self.far_clip`, not `0.0`.

So this wouldn't replace anything anymore anyway.. no? 🤔

rilei-nvidia: Not quite.  The convert to planar depth would end up modifying those far clip clear values whereas by default the clear_depth value of 0 would stay as 0 post-planar-conversion

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -346,9 +457,21 @@ def render(self, render_data: RenderData):
...
+                clear_color=0xFFEEEEEE,
+                **({"clear_depth": render_data.far_clip} if _use_depth_clear else {}),
+            ),
         )
 
+        if _depth_kinds & render_data._PLANE_DEPTH_KINDS:
+            # Derive planar depth (``depth`` / ``distance_to_image_plane``) from Newton's ray-hit
+            # distance, then apply far-plane clipping. We cannot use ``clear_depth`` here because
+            # the ray-depth buffer feeds ``convert_plane_depth``; pre-clearing to ``far_clip`` would
+            # produce ``far_clip * cos(θ)`` per pixel in the planar outputs instead of ``0.0``,
```
rilei-nvidia, line 481: See comment here @daniela-hase

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
daniela-hase: Sorry for the misunderstanding, I was missing the two lines further down 😅 

This is all good! :)

daniela-hase (approved):

PR #6374, opened 2026-07-06
[DO NOT MERGE] Probe multiple Python module selection

mataylor-nvidia:
Stacks on #6346 to test affected-test selection when two internal Python modules change together. The PR intentionally changes two module docstrings so CI output shows the union of dependent tests.


mataylor-nvidia:
Closing this develop-targeted probe because #6346's workflow changes force Testmon collect mode in the combined diff. Reopening the isolated probe against mataylor-nvidia/IsaacLab develop after its full baseline cache is populated.

PR #6377, opened 2026-07-07
[Fix] Apply rigid-body and mass properties to every link of nested-hierarchy assets

### LINKED ISSUE #5918: [Bug Report] URDF spawn properties only apply to the root link (disable_gravity, contact sensors)
### Describe the bug

When I spawn a URDF robot with `UrdfFileCfg`, the per-body spawn properties only reach the
root link. Every other link keeps the PhysX default. There is no warning.

Example: I set `RigidBodyPropertiesCfg(disable_gravity=True)`. My fixed-base arm has 29
links, but only the root link gets `disableGravity=True`. The other 28 links still fall
under gravity, so the arm droops about 0.2 m at the hand. `max_depenetration_velocity` has
the same problem.

I also set `activate_contact_sensors=True`. Only the root link gets the contact report API.
A `ContactSensor` on any other link then fails to start:

```
RuntimeError: Sensor at path '.../panda_link0/.../F1_TIP' could not find any bodies with contact reporter API.
```

### Why it happens

The Isaac Sim 6.0 URDF importer nests the links. Each child link prim sits under its parent
link prim, like `/robot/.../panda_link0/panda_link1/.../F1_TIP`.

IsaacLab applies the spawn properties with `apply_nested` in `isaaclab/sim/utils/prims.py`.
It stops walking down a path as soon as one prim takes the schema, because it assumes a
rigid body cannot contain another rigid body. `activate_contact_sensors` in
`isaaclab/sim/schemas/schemas.py` does the same. So it sets the root link and stops, and all
the nested links are skipped.

The importer side is fine. Isaac Sim's own `add_rigid_body_schemas`
(`isaacsim.asset.importer.utils`) walks the same tree with `stage.Traverse()` and reaches
every body. Only the IsaacLab helpers stop early.

### Steps to reproduce

Spawn any URDF that has a chain of links, then count how many bodies got the property:

```python
import argparse

args, _ = parser.parse_known_args()
args.headless = True
args.enable_cameras = True
app = AppLauncher(args).app

from isaaclab.sim import SimulationCfg, SimulationContext
from pxr import UsdPhysics

cfg = sim_utils.UrdfFileCfg(
    asset_path="<path to any URDF with several links in a chain>",
    fix_base=True,
    force_usd_conversion=True,
    rigid_props=sim_utils.RigidBodyPropertiesCfg(disable_gravity=True),
    activate_contact_sensors=True,
)
sim = SimulationContext(SimulationCfg(dt=1.0 / 200.0))
cfg.func("/World/robot", cfg)
sim.reset()

stage = sim_utils.get_current_stage()
total = grav = report = 0
for prim in stage.Traverse():
    if prim.HasAPI(UsdPhysics.RigidBodyAPI):
        total += 1
        attr = prim.GetAttribute("physxRigidBody:disableGravity")
        grav += int(bool(attr.Get()) if attr and attr.IsValid() else False)
        report += int("PhysxContactReportAPI" in prim.GetAppliedSchemas())
print(f"disable_gravity set on {grav}/{total} bodies")
print(f"contact report API on {report}/{total} bodies")
app.close()
```

I get `1/<N>` for both, not `<N>/<N>`.

### System Info

- Commit: `a4a7602f29` (`v3.0.0-beta`). Same problem on `release/3.0.0-beta2` (`8ef1bf7a8b`);
  `apply_nested` and `activate_contact_sensors` are unchanged there.
- Isaac Sim Version: 6.0.0-rc.22 (urdf importer `3.2.1`, `urdf-usd-converter` `0.1.0`)
- OS: Ubuntu 24.04
- GPU: RTX 5090
- CUDA: 13.0
- GPU Driver: 580.159.03

### Suggested fix

Let `apply_nested` (and the loop in `activate_contact_sensors`) keep walking into the
children of a rigid body when it is part of an articulation. Or walk the whole tree with
`stage.Traverse()`, the way the importer already does.

### Checklist

### Acceptance Criteria

- [ ] `disable_gravity` and `activate_contact_sensors` apply to every link of a URDF
  articulation, not only the root link.

hujc7:
# Description

Spawning an asset whose rigid bodies are nested (child link prims authored under their parent link prim, as produced by the URDF importer in Isaac Sim 6.0 and later) only applied `rigid_props` and `mass_props` to the first link: the `apply_nested` decorator stops descending into a prim's children once the wrapped function succeeds, assuming physics schemas cannot nest. On a 29-link arm, `disable_gravity=True` reached 1/29 bodies with no warning.

This completes the nested-hierarchy fix started in #6259 (which covered contact-sensor activation):

- `apply_nested` gains a `stop_on_success` flag (default `True`; the dual-form decorator keeps every existing bare `@apply_nested` usage working unchanged).
- `modify_rigid_body_properties` and `modify_mass_properties` opt out of the early stop, since their anchor prims are exactly the link prims that nest.
- Regression test with a 3-deep nested rigid-body chain verifying both properties land on every body (mirrors the #6259 test).

## Scope: why only the rigid-body and mass writers

An audit of the 40 asset stages referenced by `isaaclab_assets`/`isaaclab_tasks` informed the scoping:

1. Rigid-body nesting is the only *independent-anchor* nesting in use: `ShadowHandNewton` authors 24 of its 25 links body-under-body and is spawned with `rigid_props` by the shadow-hand task — this fix is required for those values to reach every link.
2. No joint-under-joint nesting and no nested articulation roots exist in any audited asset; keeping the early stop for those writers is safe.
3. Collider-under-collider exists only as duplicated same-intent anchors in a few props (e.g. the factory gear assets author `CollisionAPI` with SDF approximation on two ancestor Xforms of the same meshes, via layer composition). Those anchors describe the *same* geometry rather than independent colliders — the USD Physics spec ties `CollisionAPI` to leaf gprims, so nested Xform-level anchors have no defined precedence. Widening the traversal there would paper over an asset-level duplication instead of fixing schema application, so the collision writers keep the early stop; the asset duplication is being followed up separately.
4. Some flat assets (e.g. Franka) author `MassAPI` on both link prims and mesh prims beneath them. With this change `mass_props` now also authors the nested `MassAPI` prims; per the [USD Physics spec](https://openusd.org/release/wp_rigid_body_physics.html), a parent's explicit mass overrides mass properties authored further down the subtree, so effective per-body mass is unchanged.

## Behavior note

The Newton shadow-hand task cfg sets `disable_gravity=True` (and related rigid-body properties) that previously reached only 1 of 25 links. With this fix the cfg applies to all 25 links as intended. This is the bug being fixed, but baselines tuned against the old (partial) behavior may shift for tasks whose assets have nested links.

## Verification

- End-to-end through the spawner (`UsdFileCfg` → the same property-application path `UrdfFileCfg` uses) on a nested Geometry-scoped 3-link chain: `disable_gravity` and `mass` author on 1/3 bodies before the fix and 3/3 after; contact-report API 3/3 on both (via #6259).
- New regression test in `test_schemas.py` (nested chain, asserts both properties on every link).
- Full CI green (34/34 checks).

Fixes #5918

## Type of change

## Checklist


AntoineRichard:
AI-generated review comment:

Important — [`prims.py:561`](https://github.com/isaac-sim/IsaacLab/blob/09a61f2fa53880e49640ff5b54a4531f39647b5a/source/isaaclab/isaaclab/sim/utils/prims.py#L561): The new configured decorator form returns a bare `Callable`, losing the decorated function’s signature and failing to enforce its required boolean return. Use `ParamSpec` with overloads for bare and configured forms, and type the wrapper’s actual `None` return.

AntoineRichard (approved):
There is the AI comment otherwise LGTM

PR #6378, opened 2026-07-07
[Fix] Create ContactSensor PhysX views per body so nested-hierarchy assets initialize

### LINKED ISSUE #5126: [Bug] PhysX create_rigid_body_view fails to match bodies under Geometry/ scope from v3.0 URDF converter
## Description

The IsaacLab 3.0 URDF converter (on the `develop` branch targeting Isaac Sim 6.0) produces USD files with rigid bodies nested under a `Geometry/` scope:

```
# v3.0 URDF converter output (nested hierarchy):
/robot/Geometry/base_link
/robot/Geometry/base_link/torso_yaw_link
/robot/Geometry/base_link/torso_yaw_link/torso_pitch_link
...

# v2.x URDF converter output (flat siblings):
/robot/base_link
/robot/torso_yaw_link
/robot/torso_pitch_link
...
```

The `Geometry/` scope is created by the `urdf_usd_converter.Converter` in step 1 of the pipeline, even with `layer_structure=False`. It persists regardless of whether the asset transformer (step 6) runs.

## Impact

**`physx.SimulationView.create_rigid_body_view(pattern)`** fails to match any rigid bodies when the pattern uses flat body names (e.g., `(base_link|torso_yaw_link|...)`), because the actual prim paths are nested under `Geometry/base_link/...`.

This directly breaks **`isaaclab_physx.sensors.ContactSensor._initialize_impl()`** at line ~294, which uses `sim_utils.find_matching_prims()` + a flat `(body1|body2|...)` regex pattern for `create_rigid_body_view` and `create_rigid_contact_view`.

**`create_articulation_view` works correctly** — it takes the articulation root path and traverses the entire articulation internally. The issue is specific to `create_rigid_body_view` and `create_rigid_contact_view` which require explicit per-body path pattern matching.

## Steps to Reproduce

1. Use the `develop` branch of IsaacLab with Isaac Sim 6.0
2. Convert any URDF robot (e.g., a humanoid with 50+ links) using `isaaclab.sim.converters.UrdfConverter`
3. Spawn the robot in a scene with `activate_contact_sensors=True`
4. Create a `ContactSensor` pointing at the robot bodies
5. Observe: `"Pattern '...' did not match any rigid bodies"` from `omni.physx.tensors.plugin`

## Observed Behavior

```
[Error] [omni.physx.tensors.plugin] Pattern '/World/envs/env_*/Robot/(base_link|torso_yaw_link|...)' did not match any rigid bodies
[Error] [omni.physx.tensors.plugin] Provided pattern list did not match any rigid bodies
```

Even using full nested paths in the regex alternation fails:
```
[Error] [omni.physx.tensors.plugin] Pattern '/World/envs/env_*/Robot/(Geometry/base_link|Geometry/base_link/torso_yaw_link|...)' did not match any rigid bodies
```

## Expected Behavior

`create_rigid_body_view` should match rigid bodies at their actual prim paths, or the URDF converter should produce prim paths that are compatible with the existing pattern matching in `ContactSensor` and other consumers.

## Key Observations

- `create_articulation_view("/World/envs/env_*/Robot/Geometry/base_link")` → **works**, finds all 53 links
- `ArticulationView.shared_metatype.link_names` → returns flat names: `['base_link', 'torso_yaw_link', ...]`
- `create_rigid_body_view("/World/envs/env_*/Robot/Geometry/base_link/(base_link|torso_yaw_link|...)")` → **fails**, finds only 4 direct children
- The v2.x URDF converter produced flat siblings under `/robot/` — all existing contact sensor code assumes this flat structure

## Environment

- **IsaacLab**: `develop` branch (3.0.0)
- **Isaac Sim**: 6.0.0 (pip install)
- **Python**: 3.12
- **OS**: Ubuntu 24.04, Linux 6.17

## Suggested Fix

Either:
1. The URDF converter should produce a flat prim hierarchy (matching v2.x behavior) that `create_rigid_body_view` can pattern-match
2. `ContactSensor._initialize_impl()` should use `ArticulationView.link_paths` to get PhysX-resolved body paths instead of constructing them from USD prim names
3. `create_rigid_body_view` should support recursive path matching for articulated bodies

hujc7:
# Description

`ContactSensor` initialization fails on assets whose rigid bodies are nested (child link prims authored under their parent link prim, as produced by the URDF importer in Isaac Sim 6.0 and later). The PhysX views were created from a single parent-level name alternation derived from the first matched body's parent:

```
bodies found (USD side):   .../Robot/pelvis
                           .../Robot/pelvis/left_hip
                           .../Robot/pelvis/left_hip/left_knee
pattern built (view side): .../Robot/(pelvis|left_hip|left_knee)   ← assumes one shared parent
```

Nested bodies do not share a parent, so the pattern matched only the first-level links and initialization raised `Failed to initialize contact reporter for specified bodies`.

## Changes

The fix is two coupled changes — the second is a forced consequence of the first, not an independent choice:

1. **Create the views from one path expression per body** (`create_rigid_body_view` / `create_rigid_contact_view` take a pattern list), reusing the per-body destination expressions that `resolve_matching_prims_from_source` already returns — the same approach as the ray caster and `RigidObjectCollection`. A single pattern cannot express nested bodies at all, so the list form is the only representation the view API offers for them. The list overload also requires `filter_patterns` as one filter list per sensor pattern; the sensor replicates its uniform filter list accordingly (per-body filter semantics are unchanged, `filter_count` stays the per-body count).
2. **Adapt all raw-view consumers to the resulting body-major row order.** With a pattern list, the view enumerates matches grouped by pattern (`row = body_id * num_envs + env_id`) — the ordering is defined by the API's enumeration contract and is not selectable. The alternatives to adapting (permuting the flat arrays back to env-major on every simulation step, or expanding num_envs × num_bodies concrete paths to reconstruct env-major) add recurring hot-path or view-creation cost for no benefit, so the sensor adopts the same body-major convention `RigidObjectCollection` already documents. Every consumer of the raw flat data is updated: the three copy kernels, the `body_names` property (strides one path per body instead of slicing the first N), the debug-visualization pose fallback, and the raw-buffer indexing in the `_test_friction_forces` helper. **Data buffers and the public `data` API keep their `(num_envs, num_bodies, ...)` layout — no user-facing change.**

Also: the initialization error message now includes the actual view body counts to make future pattern mismatches diagnosable from logs.

## Regression test

Authors a chain of kinematic rigid bodies with URDF-importer-style nested link prims directly under each environment prim (no scene/cloner — the code under test only needs the prims on the stage), creates a standalone `ContactSensor` over them, and asserts: initialization succeeds (pre-fix: deterministic `RuntimeError`), all bodies resolve with exact `body_names` order, and each `(env, body)` cell of `data.pos_w` matches that body's authored world position — pinning the row-to-cell mapping end-to-end. Parametrized over 1 and 6 environments on CPU and GPU.

Existing multi-env coverage (`test_cube_stack_contact_filtering` with 6 and 24 envs, and the task suites) exercises the ordering change on flat assets.

Fixes #5126

## Type of change

## Checklist


source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
```diff
@@ -493,9 +501,9 @@ def _debug_vis_callback(self, event):
         if self.cfg.track_pose:
             frame_origins = self._data.pos_w.torch  # (N, B, 3)
         else:
-            pose = self.body_physx_view.get_transforms()  # (N*B, 7) float32
+            pose = self.body_physx_view.get_transforms()  # (B*N, 7) float32, body-major
             pose_torch = wp.to_torch(pose)
-            frame_origins = pose_torch.view(-1, self._num_sensors, 7)[:, :, :3]
+            frame_origins = pose_torch.view(self._num_sensors, -1, 7).transpose(0, 1)[:, :, :3]
         # visualize
         self.contact_visualizer.visualize(frame_origins.reshape(-1, 3), marker_indices=marker_indices.reshape(-1))
```
AntoineRichard: These changes seem unrelated?

source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/kernels.py
```diff
@@ -243,7 +248,7 @@ def update_net_forces_kernel(
     if timestamp[env] == 0.0:
         return
 
-    src_idx = env * num_sensors + sensor
+    src_idx = sensor * num_envs + env
 
     # Update net forces
     net_forces_w[env, sensor] = net_forces_flat[src_idx]
```
AntoineRichard, line 254: I don't see why we need to change these kernels?

AntoineRichard (commented):
I don't see why we make some changes to the kernels to adress the nested hierarchy. I feel like this should only be a change on the regex side? Then if we want to process flat arrays that's a different story. But I don't see why that'd be necessary.

source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
hujc7: the main reason is that the shape/order of the array is changed if we call the API, create_rigid_contact_view, with array instead of a string. I tried to avoid it, let me confirm again.

AntoineRichard (approved):

PR #6245, opened 2026-06-23
Rename MJWarp to VBD and Implement Kitless Path for Newton Deformables

mmichelis:
# Description

Deformable body spawning previously depended on Kit-only PhysX helpers from `omni.physx.scripts.deformableUtils` (`remove_deformable_body`, `create_triangle_mesh_square`). This made deformables unusable in a kitless Newton setup. This PR removes those dependencies and adds a working kitless path for deformables on the Newton backend, while keeping the existing PhysX behavior intact.

TODO: On the Newton path we author a temporary `PhysicsDeformableBodyAPI` token, which is not yet a registered USD applied schema.

A first attempt at fixing [#6207](https://github.com/isaac-sim/IsaacLab/issues/6207).

### Changes

- **`sim/utils/queries.py`** — Added `has_deformable_body_api`, a central helper that detects deformable-body schemas via `Usd.PrimTypeInfo.GetAppliedAPISchemas()`. This catches both the registered `OmniPhysicsDeformableBodyAPI` (PhysX) and the token-authored `PhysicsDeformableBodyAPI` (Newton). Exported from `sim.utils`.

- **`sim/schemas/schemas.py`** — `define_deformable_body_properties` now branches on the config's USD namespace (`newton` vs. Omni/PhysX) via `_usd_namespace`:
  - Drops the `omni.physx.scripts.deformableUtils.remove_deformable_body` call. A docstring note states the function authors a fresh setup and that callers must clear any previous deformable setup themselves.
  - Omni-only steps (`OmniPhysics*` sim APIs, rest-shape attributes, bind-pose `OmniPhysicsDeformablePoseAPI`) are gated behind the PhysX path; the Newton path applies `PhysicsSurfaceDeformableSimAPI` / `PhysicsVolumeDeformableSimAPI` tokens and syncs the visual mesh to the sim mesh.
  - The collision API and the sim-mesh render-hiding (guide purpose) are now applied once for both backends.
  - Authors the deformable body API on the root prim per backend: `OmniPhysicsDeformableBodyAPI` (PhysX) or a `PhysicsDeformableBodyAPI` token (Newton, placeholder until a registered ApplyAPI schema ships).
  - `modify_deformable_body_properties` now uses `has_deformable_body_api` for its guard.

- **`sim/utils/prims.py` / `sim/spawners/from_files/from_files.py`** — Deformable-body detection (physics-material binding and the modify-vs-define branch) now goes through `has_deformable_body_api`.

[...]

PR #6379, opened 2026-07-07
Add support for deformable rendering

huidongc:
# Description

Rendering support for deformable bodies is validated for the following combinations:

|                 | ovrtx_renderer | newton_warp_renderer | isaacsim_rtx_renderer |
| :-------------: | :------------: | :------------------: | :-------------------: |
| newton          |       ✅       |         ✅           |         ✅            |
| physx           |       --       |         ✅           |         ✅            |
| ovphysx         |       --       |         --           |         --            |

Note: we can't validate the ovphysx + ovrtx/newton_warp because Franka cloth and soft envs haven't defined the ovphysx preset in the env cfgs.

## Type of change

## Screenshots

See golden images.

## Checklist


source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1077,3 +1111,252 @@ class _DexsuiteKukaAllegroLiftCameraTestEnvCfg(DexsuiteKukaAllegroLiftCameraEnvC
...
+    physics_backend: str,
+    renderer: str,
+    data_type: str,
+    comparison_scores: list[dict],
+) -> None:
+    if physics_backend == "ovphysx":
+        pytest.skip("ovphysx is not supported yet.")
+
+    if physics_backend == "physx" and renderer == "newton_renderer":
+        pytest.skip("physx + newton_renderer is not supported yet.")
```
huidongc: Crash in the newton due to schema usage of `deformablePose:default:omniphysics:purposes`

@mmichelis for vis. not sure your PR #6245 can solve it or not.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1077,3 +1111,252 @@ class _DexsuiteKukaAllegroLiftCameraTestEnvCfg(DexsuiteKukaAllegroLiftCameraEnvC
...
+
+    if data_type == "motion_vectors":
```
huidongc: Garbage data. but this is not related to deformables. Will file a separate ticket about it because this is not related to deformable.

source/isaaclab_tasks/test/golden_images/franka_soft/newton-newton_renderer-rgb.png
huidongc, line 1: will investigate why there is one extra tiny object in env_0.

huidongc:
@greptile-apps

source/isaaclab_tasks/test/rendering_test_utils.py
huidongc: This is still reproducible after rebasing to latest of `develop` which has PR #6245 landed.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1079,3 +1113,264 @@ class _DexsuiteKukaAllegroLiftCameraTestEnvCfg(DexsuiteKukaAllegroLiftCameraEnvC
...
+            offset=CameraCfg.OffsetCfg(
+                pos=(0.85, -0.55, 0.42),
+                rot=(0.5080, 0.2114, 0.318, 0.7720),
+                convention="opengl",
+            ),
+            data_types=[data_type],
+            spawn=sim_utils.PinholeCameraCfg(clipping_range=(0.01, 3.0)),
+            width=128,
+            height=128,
+            renderer_cfg=MultiBackendRendererCfg(),
```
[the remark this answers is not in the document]
huidongc: Good catch!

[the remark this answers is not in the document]
huidongc: Let's keep it this way. The deinstancing call is a short-term workaround, and will be soon removed once the fix lands in upstream components.

source/isaaclab_tasks/test/golden_images/franka_soft/newton-newton_renderer-rgb.png
huidongc: The Newton Warp renderer (NewtonWarpRenderer → newton.sensors.SensorTiledCamera) partitions rigid shapes correctly per world (bvh_shapes_group_roots), so each tile only shows its own robot/table. But deformable geometry is raytraced from the flat, all-worlds state.particle_q triangle mesh, and is not restricted to the tile's own world — so any env's deformable that happens to land inside a camera's frustum gets drawn into that tile.

It appears only in env_0 simply because of geometry: env_0's camera at (2.35, -2.05, 0.42) looks diagonally across the 2×2 grid, and env_3's deformable at (-1.0, 1.5, 0.05) (~4.9 m away) falls almost exactly in that viewing direction, rendering as a small bar near top-center. The other cameras' diagonal deformables fall outside their frustums, so no visible bleed there. (This matches the known "deformable env mapping" area — fix_deformable_env_mapping plan and the "Validate deformable env mapping in OVRTX renderer" commit.)

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -324,6 +324,11 @@ def update_transforms(self):
         sim.physics_manager.forward()
         NewtonManager.update_visualization_state()
 
+    def update_geometries(self) -> None:
+        """No-op for Newton Warp - geometry is read directly from Newton state during render.
```
rilei-nvidia, line 419: Ah - so the Newton state is available for the renderer even when the Physics engine back-end is physx / ovphysx?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -535,11 +574,101 @@ def _setup_newton_object_bindings(self):
...
+            num_envs: Number of environments.
+        """
+        try:
+            from isaaclab_newton.physics import NewtonManager
+        except ImportError:
+            logger.debug("NewtonManager not available, skipping deformable body bindings")
+            return
+
+        # Early return if the deformable registry is empty.
+        deformable_registry = NewtonManager._deformable_registry
```
rilei-nvidia, line 605: Accessing a internal member is generally not recommended - is there a public facing way to check the registry / pending ticket for this?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -535,11 +574,101 @@ def _setup_newton_object_bindings(self):
...
+                self._deformable_particle_offsets.append(particle_offset)
+
+                vis_mesh_prim_paths.append(re.sub(r"(?<=[Ee]nv_)\.\*", str(idx), entry.vis_mesh_prim_path))
+
+        prim_count = len(vis_mesh_prim_paths)
+        if prim_count == 0:
+            logger.warning("No deformable visual prim paths collected, skipping deformable body bindings")
+            return
+
+        # World-space particle_q is written directly into mesh points. Reset the xform stack
```
rilei-nvidia, line 651: Ooo.  Won't we run into precision issues with the particles far from origin?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -608,6 +737,54 @@ def update_transforms(self) -> None:
                 device=self._device,
             )
 
+    def update_geometries(self) -> None:
+        """Sync geometries to OVRTX."""
```
rilei-nvidia, line 748: So the ovphysx + ovrtx combo is not supported at the moment right?  Can you include some of these high level details into the description of this PR?

source/isaaclab_tasks/test/golden_images/franka_soft/newton-newton_renderer-rgb.png
huidongc: Filed a ticket: https://github.com/newton-physics/newton/issues/3487

source/isaaclab_tasks/test/rendering_test_utils.py
huidongc: This is a known issue for newton: https://github.com/newton-physics/newton/issues/3228

I will try to request for a fix.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -608,6 +737,54 @@ def update_transforms(self) -> None:
...
+            raise RuntimeError("Newton state has no particle_q but deformable bindings exist")
+
+        for points_buffer, particle_offset in zip(
+            self._deformable_points_buffers,
+            self._deformable_particle_offsets,
+            strict=True,
+        ):
+            particle_count = points_buffer.shape[0]
+            particle_end = particle_offset + particle_count
+            wp.copy(points_buffer, particle_q[particle_offset:particle_end])
```
rilei-nvidia: I guess there's no way to pass a single array of the particle positions (basically particle_q) with the associated offsets directly to OVRTX right?  Wondering how we can avoid the N (env count) number of GPU-to-GPU copies that need to be made

rilei-nvidia (commented):
@huidongc thanks for the comprehensive tests, clear and simple code, and documentation.   Commented with some questions

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Good point, but I can't find a public facing API for this.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: I agree it will likely run into precision issues.

I feel the particle_q struct is still to be improved, currently deformable geometry is raytraced from the flat, all-worlds particle_q array, and is not restricted to the tile's own world — so any env's deformable that happens to land inside a camera's frustum gets drawn into that tile. See https://github.com/newton-physics/newton/issues/3487

For both correctness and precision reason, the flat global particle_q array should be partitioned into multiple per-world structs.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: At the point we can't validate this combo because the franka cloth and soft envs haven't defined ovphysx preset in the env cfgs.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
huidongc: To my understanding, NewtonWarpRenderer always renders from a Newton Model/State, and that works even when the active physics backend is PhysX or OVPhysX. This is achieved through a shadow Newton model built inside NewtonManager:

- Under the Newton backend, NewtonManager._model/_state_0 are the authoritative simulation state, advanced directly by step()/forward kinematics. update_visualization_state() is a no-op.

- Under the PhysX / OVPhysX backend, there is no native Newton state. So NewtonManager lazily builds a shadow Newton model by walking the USD stage (_ensure_visualization_model at newton_manager.py:1857), and refreshes its _state_0.body_q from the live physics scene every render frame.

Both cases are hidden behind NewtonManager.get_model() and NewtonManager.get_state(), so the renderer never needs to know which physics backend is running.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: No — with the current OVRTX API you cannot pass one particle_q buffer plus an offset table the way Newton's Fabric sync does. Array attributes like points require List[DLTensor], one tensor per prim, via write() / write_array_attribute(). map() is only for scalar per-prim attributes (e.g. omni:xform).

But your question pushed my agent further to avoid any copy. See the new commit - e887864, there should be no copy now. Thanks a lot for this question!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Added a validation matrix in the PR description.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: I see.  Good to know how it works with the different physics back-ends.  Thank you!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: cool thanks for the explanation - and I see it's now passing the slices of particle_q directly, nice!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: The table is very clear - thank you!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Good to know.   Do you know if the Newton developers are aware of:
> For both correctness and precision reason, the flat global particle_q array should be partitioned into multiple per-world structs.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -466,6 +493,18 @@ def _clone_sources_in_ovrtx(self):
...
+        # Restore the pre-clone offsets so hierarchy-positioned prims (e.g. static props)
+        # under /World/envs/env_* keep their intended per-env world placement.
+        self._renderer.write_attribute(
+            prim_paths=env_prim_paths,
+            attribute_name=xform_attr_name,
+            tensor=env_root_xforms,
+            semantic=Semantic.XFORM_MAT4x4,
+            prim_mode=PrimMode.MUST_EXIST,
+        )
+        logger.info("Restored per-env root transforms after cloning")
```
rilei-nvidia, line 516: Interesting!  How did you end up finding out about this issue and how come we didn't encounter before?

huidongc: I think it is just new envs exposing it. Without the snapshot and restoring, objects not included in `body_label` of the newton model will be mispositioned.

rilei-nvidia (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -425,22 +439,35 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
...
             raise RuntimeError("Clone plan is required when using OVRTX cloning")
 
-        logger.info("Cloning sources in OVRTX...")
-
         num_envs = clone_plan.clone_mask.shape[1]
+        env_prim_paths = [f"/World/envs/env_{i}" for i in range(num_envs)]
+        xform_attr_name = "omni:xform"
+
+        # Snapshot per-env root transforms before clone_usd overwrites them with the source env root.
+        env_root_xforms = np.empty((num_envs, 4, 4), dtype=np.float64)
```
pbarejko, line 473: Why do we have to do this? I thought the whole point of scene partitioning was that we don't need to do this. We can have them overlapping in the stage. This seems more like debugging feature?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -535,11 +574,98 @@ def _setup_newton_object_bindings(self):
...
         self._object_newton_indices = wp.array(newton_indices, dtype=wp.int32, device=self._device)
 
+    def _setup_deformable_bindings(self, num_envs: int):
+        """Setup OVRTX bindings for Newton deformable bodies.
+
+        Args:
+            num_envs: Number of environments.
+        """
+        try:
+            from isaaclab_newton.physics import NewtonManager
```
pbarejko, line 599: We are missing some parts of arch design here. The scene data provider should bridge simulation backends with rendering backends. This code instantiates concrete classes from Newton. When we add another simulation backend we will keep adding if statements? This isn't good design. We should have some abstraction/interface that allow us to get scene data without knowing simulation backend.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: we are relying on the objects to be in the list of body_label, for those objects we reset xform stack and then sync their transforms with the data in body_q, for these objects, the env root transforms don't matter to them.

however, for objects not in the body_label list, if we don't restore the env root transforms, those objects will be mis-positioned.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: I can see where you come from.

But to defend this PR, I don't make a new pattern into existing code, it is the same design in setup_xform_bindings() since the ovrtx renderer class was born. 

I can note these concerns and see how it can evolve from here.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Let's say some static rigid objects that are not in in the body_label list, we won't create xform bindings for these objects:

```py
        all_body_paths = getattr(newton_model, "body_label", None)
        if all_body_paths is None:
            logger.info("Newton model has no body_label, skipping object bindings")
            return

        object_paths = []
        newton_indices = []
        for idx, path in enumerate(all_body_paths):
            if "/World/envs/" in path and self._camera_rel_path not in path and "GroundPlane" not in path:
                object_paths.append(path)
                newton_indices.append(idx)

        if len(object_paths) == 0:
            logger.info("No dynamic objects found for binding")
            return

        self._object_xform_binding = self._renderer.bind_attribute(
            prim_paths=object_paths,
            attribute_name="omni:xform",
            semantic=Semantic.XFORM_MAT4x4,
            prim_mode=PrimMode.EXISTING_ONLY,
        )
```

Then because we don't create xform bindings for these objects, we don't reset their xform stack.

These means, the correct position of these objects will depend on each of its ancestors to have correct xform in the renderer data.

If the env root has no valid xform, then these objects will positioned close to world origin.
[2:06 PM]for example, the table could be placed at world origin without snapshot/restoring the env root xforms, thus missing in the corresponding tiles.

pbarejko: ok thank you for explaining this!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pbarejko: I wasn't attacking your PR :) Just pointing out the gaps. Yes I have seen what is happening in `setup_xform_bindings` and I was wondering why we don't use Scene Data Provider there?

 After we add physx deformables we would have to modify this code again?

pbarejko (approved):

PR #6385, opened 2026-07-07
ovrtx: expose semantic segmentation idToLabels

rilei-nvidia:
# Description

This PR brings our ovrtx renderer integration's `semantic_segmentation` annotator to full parity with Isaac RTX's.  This involves building the idToLabels dict mapping the semantic ID or color to the corresponding labels - which IsaacRTX supplies through Replicator.  

A new `ovrtx_annotator_utils.py` is introduced encapsulating the logic of decoding ovrtx SemanticIdMap raw array of C++ structs.  In the future we can remove this implementation in favor of logic exposed natively in ovrtx or through ovannotators.

While adding the tests to validate the consistency of idToLabels output - I also noticed that the IsaacRTX's pseudo-random colors generated from the semantic IDs was slightly different than ovrtx's.  So that's been fixed and golden images regenerated.

## Type of change

## Screenshots

See golden images update

## Checklist

- [x]  I have made corresponding changes to the documentation


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -718,6 +731,30 @@ def _process_id_segmentation_render_var(
...
+        BACKGROUND (ID 0) and UNLABELLED (ID 1) entries are always included.
+
+        Args:
+            render_data: OVRTX render data for the current frame.
+            frame: OVRTX frame holding the mapped render vars.
+        """
+        if "SemanticIdMap" not in frame.render_vars:
+            return
+
+        with frame.render_vars["SemanticIdMap"].map(device=Device.CPU) as mapping:
```
pbarejko, line 760: Why this mapping happens on CPU?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -250,6 +254,11 @@ def supported_output_types(self) -> dict[RenderBufferKind, RenderBufferSpec]:
             if self.cfg.colorize_instance_id_segmentation
             else RenderBufferSpec(1, wp.uint32)
         )
+        # Semantic segmentation: colorized RGBA (uint8), else raw int32 IDs (matches Isaac RTX, whose
+        # non-colorized per-pixel value is the semantic ID).
+        semantic_seg_spec = (
```
pbarejko, line 259: With colorization enabled, this PR transforms both parts of the semantic output:

1. Per-pixel semantic IDs become hashed RGBA colors.
2. semantic_id → labels becomes RGBA color → labels.

The original ids are lost. And to what I read there is no way to convert color to back original semantic_id. How good is the hashing function? Can we get collisions? 

Should we retain the raw per-pixel semantic-ID tensor and keep the label mapping keyed by semantic ID? The colorized image could be exposed separately as a visualization:

```
selected pixel (x, y)
  semantic_ids[y, x]
  semanticIdToLabels[semantic_id]
```

For example:
```
  output["semantic_segmentation"]          # colorized RGBA
  output["semantic_segmentation_ids"]      # raw semantic IDs
  info["semantic_segmentation"]["semanticIdToLabels"]
```

One tradeoff: retaining both adds four bytes per pixel. At 8,192 environments and 128×128 resolution, that is another 512 MiB, so an explicit optional output or on-demand colorization may be preferable.


Additionally, should idToTabel, which is semantic id to label dictionary be a shared dictionary across SEMANTIC_SEGMENTATION and INSTANCE_SEGMENTATION?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: The SemanticIdMap provides the array of C++ structs we need to decode into our host-side idToLabels python dictionary

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Good observations and questions here @pbarejko 

> With colorization enabled, this PR transforms both parts of the semantic output:
> Per-pixel semantic IDs become hashed RGBA colors.
> semantic_id → labels becomes RGBA color → labels.
> The original ids are lost. And to what I read there is no way to convert color to back original semantic_id. 

Previously, the ovrtx semantic_segmentation was already and always colorizing the output without an option to produce raw IDs -> labels - but as documented here https://github.com/isaac-sim/IsaacLab/blob/develop/docs/source/overview/core-concepts/sensors/camera.rst#semantic-segmentation - and to reach parity with the IsaacRTX integration, we should expose the flag for the user to output the raw IDs if desired.  

> How good is the hashing function? Can we get collisions?

There is definitely room for collisions as `2^32 - 2` (2^32 possible values, minus 2 reserved for unlabeled and background) is a much greater number than 256^3 (possible uint8_t per channel color combos).   Realistically for the `semantic_segmentation` annotator this would only come up if the scene has assigned a huge number of different semantic tags.  For instance_segmentation` this could definitely come up, however as I understand the colorize_outputs = True is only for visualization, and they have the option to turn that off and get the raw IDs.

> Should we retain the raw per-pixel semantic-ID tensor and keep the label mapping keyed by semantic ID? The colorized image could be exposed separately as a visualization:

We could if there is a solid use case for that - and that would be a change we would likely want to implement and align across the other renderer back-ends.   I'd like to understand how the use cases more before making this change.  Do you want me to ticket up a investigation for this?

> Additionally, should idToTabel, which is semantic id to label dictionary be a shared dictionary across SEMANTIC_SEGMENTATION and INSTANCE_SEGMENTATION?

Do you mean for example, exposing the idToLabels in:
```
tiled_camera.data.info['idToLabels']
```
as opposed to:
```
tiled_camera.data.info['semantic_segmentation']['idToLabels']
```
?

They're distinct in nature, as their keys are different. I think consolidating the idToLabels up a level would make it confusing to the user as `tiled_camera.data.info['semantic_segmentation']['idToLabels']` would communicate more exactly the data it's accessing in code.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pbarejko: > into our host-side idToLabels python dictionary

Perhaps this goes beyond scope of this PR but why do we need to put it into CPU? Have you traced how this data is being consumed?

Long term goal would to keep all of the data on GPU, and avoid syncing/copying data between CPU/GPU devices.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pbarejko: > Do you mean for example, exposing the idToLabels in:
> tiled_camera.data.info['idToLabels']

I would explore this as a possibility. Probably under different name `idToLabels` seems ambiguous. 

Based on what was shared with us semantic segmentation and instance segmentation use `SemanticIdMap`.  I wonder if we could pull it once and reuse for both.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Agreed on the general notion that we should keep the data on the GPU if possible.

As for consumption - I don't know how exactly IsaacLab users are using this dict for look-ups in an actual workflow.  The idToLabels host-side dictionary is what IsaacRTX aka Replicator outputs.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Ahhh I see what you mean.  Yes semantic segmentation & instance segmentation can absolutely share the same SemanticIdMap render var data output internally.  

For semantic segmentation's case, its user-facing idToLabels dict is more or less the equivalent to SemanticIdMap, _however_ for instance segmentation its idToLabels and idToSemantics dicts are a combination of the other maps, thus not the same!

source/isaaclab_tasks/test/golden_images/dexsuite_kuka_hetero/newton-ovrtx_renderer-semantic_segmentation.png
huidongc, line 1: Looks like some objects were missing? How do you fix it? Or is it another random missing issue?

huidongc: Maybe the same issue that @pbarejko had been investigating?

source/isaaclab_tasks/test/golden_images/shadow_hand/newton-isaacsim_rtx_renderer-semantic_segmentation.png
huidongc, line 1: What is the difference?

source/isaaclab_tasks/test/golden_images/dexsuite_kuka_hetero/newton-ovrtx_renderer-semantic_segmentation.png
rilei-nvidia: I didn't do anything to fix them - only deleted all the semantic_segmentation images and regenerated them.  The previous golden must have been saved when the GPU transform issue was still affecting our renders and we did not catch it.  Then the pixel threshold must have been high enough for it to pass the test whether or not those objects were present

source/isaaclab_tasks/test/golden_images/shadow_hand/physx-isaacsim_rtx_renderer-semantic_segmentation.png
huidongc, line 1: I don't know what happened and made it an executable file before. Thanks for fixing it.

source/isaaclab_tasks/test/golden_images/shadow_hand/newton-isaacsim_rtx_renderer-semantic_segmentation.png
rilei-nvidia: The green is slightly more saturated, and now matches the exact color that replicator (Isaac RTX renderer backend) generates through its colorization processs!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -213,6 +214,9 @@ def __init__(self, spec: CameraRenderSpec, device):
         self.num_cols = math.ceil(math.sqrt(self.num_envs))
         self.num_rows = math.ceil(self.num_envs / self.num_cols)
         self.warp_buffers: dict[str, wp.array] = {}
+        # Per-output metadata collected during render() and copied into CameraData.info by read_output().
+        # Currently only "semantic_segmentation" is populated (with an "idToLabels" mapping).
+        self.renderer_info: dict[str, Any] = {}
```
huidongc, line 219: this struct holds some extra metadata of annotators, I don't know why the name ends up being "renderer info". any idea?

huidongc: I know this name is copied from IsaacRTX, just wanted to be sure if it is a precise name that indicates its purpose.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -633,15 +642,19 @@ def read_output(
...
+        zero-copy warp array stored in ``render_data.warp_buffers``, and :meth:`render` writes the rendered
+        tiles directly into those warp arrays. The only work left is forwarding the metadata gathered in
+        ``render_data.renderer_info`` (e.g. the semantic ``idToLabels`` mapping).
 
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.read_output`.
         """
+        if camera_data.info is None:
+            camera_data.info = {}
+        for output_name, info in render_data.renderer_info.items():
+            camera_data.info[output_name] = info
```
huidongc: I wondered in case some keys in the source dict get dropped, the dest dict will have some stale entries. Not sure if it is practical for OVRTX in real life, but this function is brittle if that happens.

Is there a reason why you don't use shallow reference of the source dict? Or the renderer_info is not a dict but a traversable C++ struct?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Yeah, at this point, we are doing some expensive transfer that might not even be needed by users. It might warrant a ticket for us to consider lazy data transfer from GPU to CPU until the data needs to be accessed, not just for this specific annotator, but holistically.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: so `camera_data.info[output_name] = info` is already doing a shallow reference, thus we won't run into lifetime issues, I can make the docstring more clear!

source/isaaclab_tasks/test/golden_images/shadow_hand/physx-isaacsim_rtx_renderer-semantic_segmentation.png
rilei-nvidia: No worries!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Yeah, I am not sure of the history there.  We should talk about it tomorrow at Sync up.  If there is a stronger name, we can a follow-up PR to change this for all renderer integrations for consistency

rilei-nvidia: So i grepped the code base and the only results are in isaac rtx renderer, and now ovrtx renderer and in our rendering test utils.  I'm a little surprised that we don't have any other example scripts / code accessing that.   

While IsaacLab users could be relying on renderer_info, given 3.0 is a major version upgrade we can consider renaming it to something more precise!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Let me try to give a more specific example:

If `render_data.renderer_info` goes from `{"Output1": "foo", "Output2": "bar"}` to `{}`, `camera_data.info` will still contain the stale entries for `{"Output1": "foo", "Output2": "bar"}`.

huidongc (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: OK I understand now, yes indeed the staleness could be a thing if renderer_info happens to not include an output for a given frame.  

The reason why we shouldn't directly do:
```
camera_data.info = render_data.renderer_info 
```
is because camera_data.info is pre-seeded with the possible output names mapping to None (see CameraData.allocate) - this allows the camera to communicate all the output names it can supply, even if it has no data yet.  Also render_data is considered a "private" data structure since it is renderer-specific, and it would not be best to share its internals, even if in practice it would most likely work.  To your other comment about the renderer_info naming - we could definitely rename that to whatever we see fit, but will do in a separate PR.

I've pushed up a fix that addresses the staleness by looping over camera_info.info's keys and extracting the data from render_data.renderer_info instead with a .get() (thus will return None if there is no data, and clear the stale value)

PR #6388, opened 2026-07-07
[CI] Fetch OVPhysX wheelhouse through Docker

marcodiiga:
## Summary

- Replace OVPhysX NGC-resource ingestion with an optional Docker wheelhouse.
- Keep the Docker path disabled with `ovphysx_wheelhouse_image: ""`; CI continues to install public `ovphysx==0.5.9`.
- Require future wheelhouse images to be pinned by digest and restrict their use to trusted same-repository refs.
- Reuse the existing offline wheelhouse installation and runtime-version check.
- Materialize ECR dependency-cache hits for test callers while base, cuRobo, and multi-GPU build-only jobs opt out.
- Remove private-wheel handling from the license workflow.
- Align the public `ovstage` and `typing_extensions` license metadata exceptions.

## Activation

A separate reviewed change can set `ovphysx_wheelhouse_image` to an `@sha256:` image reference.

## Validation

- Focused ECR/wheelhouse regression check passed with a verified red/green boundary.
- Repository-wide `./isaaclab.sh -f` passed on the committed tree.


.github/workflows/build.yaml
```diff
@@ -598,7 +598,7 @@ jobs:
       github.event_name != 'push' &&
       needs.build.result == 'success'
     env:
-      USE_OVPHYSX_WHEELHOUSE: ${{ needs.config.outputs.ovphysx_wheelhouse_resource != '' && ((github.event_name == 'pull_request' && github.base_ref == 'develop' && github.event.pull_request.head.repo.full_name == github.repository) || (github.event_name != 'pull_request' && github.ref_name == 'develop')) }}
+      USE_OVPHYSX_WHEELHOUSE: ${{ needs.config.outputs.ovphysx_wheelhouse_image != '' && ((github.event_name == 'pull_request' && github.base_ref == 'develop') || (github.event_name != 'pull_request' && github.ref_name == 'develop')) }}
```
hujc7: Agent review: Dropping the `head.repo.full_name == github.repository` clause means fork PRs to `develop` now take the wheelhouse path (also line 781). Mechanically fine — no secret needed, and this PR's own green `isaaclab_ov` job proves it — but the old code explicitly excluded untrusted runs and the manifest says `internal-ci-only`. Is shipping these wheels into approved fork-PR runs signed off? A one-line comment documenting the intent would help the next reader.

.github/workflows/license-check.yaml
```diff
@@ -33,98 +31,82 @@ jobs:
...
+        WHEELHOUSE_IMAGE: ${{ steps.config.outputs.wheelhouse_image }}
       run: |
         set -euo pipefail
 
         if [ -z "${NGC_API_KEY:-}" ]; then
-          echo "::error::NGC_API_KEY is unavailable; cannot download the configured OVPhysX wheelhouse"
+          echo "::error::NGC_API_KEY is unavailable; cannot pull the configured OVPhysX wheelhouse"
+          exit 1
+        fi
+        if [[ "$WHEELHOUSE_IMAGE" != nvcr.io/nvidian/ovphysx-wheelhouse:*@sha256:* ]]; then
```
hujc7: Agent review: An empty pin fails this guard, so clearing the pin hard-fails license-check for every same-repo PR — unlike `build.yaml`, which degrades to the public-package path, and contrary to the PR description's "clearing the pin restores the normal public-package path". If degradation is intended, gate the step:

```yaml
    - name: Extract OVPhysX wheelhouse
      id: extract-wheelhouse
      if: steps.config.outputs.wheelhouse_image != ''
```

(and check `uv sync` tolerates an empty `UV_FIND_LINKS`). If fail-fast is intended, the description sentence should note license-check needs a matching edit when the pin is cleared.

.github/actions/run-package-tests/action.yml
```diff
@@ -158,99 +158,87 @@ runs:
...
-      if: inputs.wheelhouse-resource != '' && env.NGC_API_KEY != ''
-      uses: actions/cache@v4
-      with:
-        path: ${{ runner.temp }}/ngc-cli-cache/ngccli_linux.zip
-        key: ngc-cli-${{ runner.os }}-${{ runner.arch }}-4.20.0-5cf084c88998c58ad8abf7849d2d1b41d578423886eb03018df10194e341d35b
-
     - name: Extract ovphysx wheelhouse
       id: extract-wheelhouse
-      if: inputs.wheelhouse-resource != ''
+      if: inputs.wheelhouse-image != ''
```
hujc7, line 225: Agent suggestion: This ~100-line extract+validate block is hand-duplicated in `license-check.yaml`, and it's exactly the code that must not drift. Since this PR rewrites both copies anyway, consider factoring them into a local composite action (this file already nests `./.github/actions/run-tests`) or a shared script + validator `.py` beside `cleanup_docker_storage.sh`, with auth mode as an input. Non-blocking.

hujc7 (commented):
Agent review: Transport swap verified end to end — this PR's own [isaaclab_ov job](https://github.com/isaac-sim/IsaacLab/actions/runs/28883056613/job/85676875809) pulled the pinned digest, validated both wheels, and ran the OVPhysX tests green. Two notes:

1. The `exclude-pattern`/`test-k-expr` flips are behavioral fixes: `cond && '' || 'x'` always yields `'x'` in GitHub expressions, so OVPhysX tests never ran even on trusted runs before this PR. Worth a sentence in the description.
2. The license-check path is unexercised on GitHub (self-skips for fork PRs; none of the last 100 runs executed). `preserve_wheelhouse_root=true` fixes a real latent bug there, but the path rests on local validation — suggest exercising it once via a same-repo PR after merge.

AntoineRichard:
@hujc7 thanks for the careful review — all three points are addressed:

**1. Ternary flip is a behavioral fix, not just a rename.** Documented in the PR description under *Review follow-ups*: the old `${{ cond && '' || 'x' }}` always evaluates to `'x'` (empty string is falsy), so the OVPhysX tests never ran even on trusted runs; the inverted `${{ cond != 'true' && 'x' || '' }}` fixes that.

**2. Fork-gate on the test jobs.** Added a one-line comment in `build.yaml` documenting the intent: the image is pulled with the runner's ambient `nvcr.io` login (no `NGC_API_KEY` secret crosses the fork boundary) and the wheels are `internal-ci-only` rather than credentials. Whether fork-PR runs *should* receive these wheels is ultimately a maintainer call — flagging it explicitly rather than leaving it implicit.

**3. license-check empty-pin / unexercised path.** Rather than just gate the step, I removed the internal wheelhouse from `license-check` entirely. `uv sync --extra all` excludes the `ov` extra (it clashes with isaacsim), and `uv.lock` confirms `ovphysx` is gated behind `extra == 'ov'` only — so that job never actually resolved OVPhysX and the wheelhouse fetch was inert. OVPhysX now gets the same treatment as its sister `ovrtx` (plain public resolution). Both your concerns dissolve with the code: there's no empty-pin hard-fail and no locally-validated-only path left. Note this drops the `NGC_API_KEY` dependency and the same-repo gate, so **license-check now runs on fork PRs again** — worth a second look as a deliberate behavior change.

Two related cleanups while here:
- Renamed the private-source identifiers to `internal` (`USE_OVPHYSX_INTERNAL`, `ovphysx_internal_image`, the `internal-image` action input) to match the manifest's `internal-ci-only` purpose; `wheelhouse` now names only the extracted artifact.
- Factored the byte-identical manifest validator out of both workflows into a shared `validate_wheelhouse_manifest.py`.

The Docker wheelhouse path for the `build.yaml` test jobs (which need ovphysx ≥ 0.5.1) is unchanged.

.github/workflows/build.yaml
marcodiiga: The fork-PR path is intentional. The workflow YAML does not implement an
approval gate; whether an untrusted external fork workflow runs is governed
by the repository/organization approval policy. If such a run is approved,
the image is pulled with the runner's ambient nvcr.io login and no
`NGC_API_KEY` value is passed to fork code. Comments on both jobs document
the intended path and that the wheelhouse is internal-CI-only rather than
credential material.

.github/workflows/license-check.yaml
marcodiiga: Resolved by removing the wheelhouse dependency from `license-check`
entirely. The root `all` extra selects neither `ov` nor `rtx`, so this job
installs neither optional backend dependency. It no longer reads the
internal-image pin, `NGC_API_KEY`, or a same-repository gate, so the
empty-pin failure mode no longer exists.

.github/actions/run-package-tests/action.yml
marcodiiga: Addressed. `license-check.yaml` no longer extracts the wheelhouse.
`run-package-tests` delegates the remaining extraction path to
`extract_wheelhouse.sh`, with the standard-library helpers beside it for
image-reference, Docker-credential, and manifest validation. The same helper
suite is run from `license-check`, so the action logic is no longer duplicated
there.

.github/actions/run-package-tests/action.yml
```diff
@@ -158,99 +158,87 @@ runs:
...
+        source = pathlib.Path(sys.argv[1])
+        destination = pathlib.Path(sys.argv[2])
+        config = json.loads(source.read_text(encoding="utf-8"))
+        config["credsStore"] = ""
+        config.pop("credHelpers", None)
+        destination.write_text(json.dumps(config), encoding="utf-8")
+        PY
+        chmod 700 "$docker_config_dir"
+        chmod 600 "$docker_config_dir/config.json"
+        export DOCKER_CONFIG="$docker_config_dir"
```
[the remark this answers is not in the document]
marcodiiga: Fixed. Credential handling moved to `prepare_docker_config.py`. It
canonicalizes matching `auths` and `credHelpers` aliases (including
`https://nvcr.io/v2/`) to `nvcr.io`, rejects conflicting aliases, preserves
the applicable helper when static auth is absent, and fails with an
actionable message if no credential source exists. No credential value is
printed. Regression tests cover canonical static-auth and helper output,
deduplication, conflict rejection, global `credsStore`, and no-credential
failure.

kellyguo11:
for the python license check, you can update the license-exceptions.json file to include ovstage, and update the license for typing_extensions

.github/actions/run-package-tests/action.yml
```diff
@@ -174,130 +174,62 @@ runs:
...
       shell: bash
       env:
-        WHEELHOUSE_RESOURCE: ${{ inputs.wheelhouse-resource }}
+        WHEELHOUSE_IMAGE: ${{ inputs.wheelhouse-image }}
       run: |
         set -euo pipefail
 
-        if [ -z "${NGC_API_KEY:-}" ]; then
-          echo "::error::wheelhouse-resource is set but NGC_API_KEY is unavailable; cannot download configured wheelhouse resource"
+        if [[ "$WHEELHOUSE_IMAGE" != *@sha256:* ]]; then
```
marcodiiga, line 232: Blocking: the Docker transport now validates neither the image origin nor the extracted artifact. The `*@sha256:*` check accepts any repository and malformed digests; after extraction, `manifest.json`, source identity, the exact wheel set, and wheel hashes are never verified. The later OVPhysX version comparison is only self-consistency against the same image and happens after installation/import.

Please retain a narrow pre-install validator that allowlists the expected repository, requires a full lowercase SHA-256 digest, binds manifest version/source identity to repository-controlled expected values, and verifies regular wheel hashes/sizes with no missing or extra files. The current OVPhysX wheel installs OVStage transitively, but OVStage is now part of the required carrier contract, so the validator/runtime check should also require, import, and version-check both OVPhysX and OVStage. The earlier public validator is a useful reference, with its old repository/tag assumptions updated rather than restored unchanged.

.github/workflows/config.yaml
```diff
@@ -8,4 +8,4 @@
 isaacsim_image_name: nvcr.io/nvidian/isaac-sim
 isaacsim_image_tag: latest-develop
 isaaclab_image_name: nvcr.io/nvidian/isaac-lab
-ovphysx_wheelhouse_resource: "nvidian/ovphysx-wheelhouse:trunk-gf62c22207cfd"
+ovphysx_wheelhouse_image: ""
```
marcodiiga, line 11: Blocking test gap: leaving this pin empty for now is sensible, but it also means every check on this head follows the public-package fallback and none executes the new registry/extraction path. The focused helper/policy tests were removed too, so green CI cannot catch malformed references, missing/extra/tampered wheels, cleanup regressions, or broken trusted-ref fallback.

Please retain a reduced CPU-only mocked suite in an existing workflow. It should cover the dormant/enabled policy, strict image reference and manifest/file validation (including OVStage), and cleanup/error paths. The real image pull, offline OVPhysX+OVStage install/import checks, and OVStage-backed ingestion test can remain an explicit gate for the later activation change.

marcodiiga (commented):
Reviewed exact head after handoff `4e4656401d181da17fe86ab1c7fa5d191da81a02`. The Docker carrier and intentionally empty pin make sense. I found two blocking gaps in the trim: the carrier path no longer establishes artifact identity/integrity or the mandatory OVStage closure, and no persistent test exercises that dormant policy. I left one comment for each concern.

AntoineRichard (approved):

.github/workflows/license-check.yaml
```diff
@@ -15,145 +15,14 @@ concurrency:
 
 jobs:
   license-check:
-    # Fork PRs cannot access NGC_API_KEY, so they cannot resolve the immutable
-    # OVPhysX wheel required by the development extras.
-    if: github.event.pull_request.head.repo.full_name == github.repository
```
hujc7, line 20: Any reason this is dropped?

AntoineRichard: I may have cut a bit too aggressively.

hujc7 (approved):

.github/workflows/license-check.yaml
AntoineRichard: Looking again, dropping this guard was intentional. It only existed because the old license workflow required NGC_API_KEY to download the private NGC wheelhouse. This PR leaves the Docker carrier disabled and resolves the public packages instead, so the license check no longer needs secrets and can run on fork PRs. The refreshed license check passes. When the Docker carrier is activated, we can decide whether its CI-only wheels also need to participate in license scanning.

PR #6397, opened 2026-07-07
[IsaacLab Tests]: golden usd tests

mataylor-nvidia:
# Description

This PR adds **golden USD stage comparison** to the rendering correctness test suite. It extends `maybe_save_stage` so that, in addition to optionally dumping the composed USD stage when `ISAAC_LAB_SAVE_STAGES` is set, it can also validate the exported stage against a checked-in golden `.usda` baseline for selected rendering cases.

**Motivation:** Rendering correctness tests today compare rendered pixel output against golden images. Those catch visual regressions but not changes to the *scene composition* itself (prims, transforms, materials, camera/sensor setup). Golden USD stages let us detect unintended changes to the stage that a task builds — a stronger, earlier signal than pixel diffs alone.

**What's included:**

- **`maybe_save_stage` extension** (`rendering_test_utils.py`) — new `compare_golden` keyword. When enabled, it writes the stage to a temp file, flattens it (inlining sublayer references), and compares against `golden_stages/<test_name>/<backend>-<renderer>-<data_type>.usda` by checking prim structure exactly and world transforms with `numpy.allclose`. On a missing baseline it bootstraps the golden file and fails intentionally with `"A new baseline was written"`; on a mismatch it fails with a human-readable diff.

- **Unit tests** (`test_maybe_save_stage_golden.py`) covering bootstrap, match, mismatch, noise tolerance, sublayer result, and no-op paths with a mocked `save_stage`.

- Committed golden baselines under `golden_stages/`, a `.gitattributes` entry, and a changelog fragment.

Fixes # ([OMPE-98717](https://jirasw.nvidia.com/browse/OMPE-98717))

## Type of change

## Checklist


mataylor-nvidia:
@greptile review

.gitattributes
```diff
@@ -12,6 +12,7 @@
 *.hdf5 filter=lfs diff=lfs merge=lfs -text
 
 source/isaaclab_tasks/test/golden_images/**/*.png filter=lfs diff=lfs merge=lfs -text
+source/isaaclab_tasks/test/golden_stages/**/*.usda filter=lfs diff=lfs merge=lfs -text
```
huidongc: There is a `.usda` entry right above.

mataylor-nvidia: I will remove this and use the entry above

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
**Kitless golden baselines still need to be bootstrapped**

The latest commit enables `compare_golden=True` for the kit-less cartpole test, so `ovphysx+ovrtx`, `newton+ovrtx`, `ovphysx+newton_renderer`, and `newton+newton_renderer` will now each attempt a golden stage comparison on their `rgb` case — but no `.usda` baselines have been committed for those backends yet.

Until the baselines are committed, the kitless `rgb` cases will fail with "Golden stage not found" on every run (the bootstrap writes the file and intentionally fails). To fix this, run the following on a GPU machine, commit the generated files, and push:

```bash
# Bootstrap pass — writes the 4 missing baselines and fails (expected)
./isaaclab.sh -p -m pytest source/isaaclab_tasks/test/core/test_rendering_cartpole_kitless.py -k "rgb" -v

# Stage and commit the generated goldens
git add source/isaaclab_tasks/test/golden_stages/cartpole/ovphysx-ovrtx-rgb.usda \
        source/isaaclab_tasks/test/golden_stages/cartpole/newton-ovrtx-rgb.usda \
        source/isaaclab_tasks/test/golden_stages/cartpole/ovphysx-newton_renderer-rgb.usda \
        source/isaaclab_tasks/test/golden_stages/cartpole/newton-newton_renderer-rgb.usda
git commit -m "Add golden USD stage baselines for kitless backends"

# Verify pass — all 4 cases should now pass
./isaaclab.sh -p -m pytest source/isaaclab_tasks/test/core/test_rendering_cartpole_kitless.py -k "rgb" -v
```

(The exact filenames match the `{physics_backend}-{renderer}-{data_type}.usda` pattern used by `maybe_save_stage`; adjust if the `ovrtx` renderer string differs.)

source/isaaclab_tasks/test/golden_stages/cartpole/newton-isaacsim_rtx_renderer-rgb.usda
nvsekkin, line 1: question: where were these files generated?

source/isaaclab_tasks/test/golden_stages/registered_tasks_Isaac-Reorient-Cube-Shadow-Camera-Direct/default_physics-default_renderer-stage.usda
```diff
@@ -0,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:41c70aed3a5e6f53a27e9210ffe6d5fd6dde2ada543fe8e76805a46f4df500c3
+size 10647919
```
nvsekkin, line 3: question: related to the one above, why is this file much larger then the others?

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -187,19 +195,182 @@ def _make_sensor_data_type_params(
...
+            flat_layer = opened_stage.Flatten()
+
+            if not os.path.exists(golden_path):
+                if not flat_layer.Export(golden_path):
+                    pytest.fail(f"Failed to export the flattened golden baseline to {golden_path}.")
+                # Strip host-specific provenance/paths so the committed baseline is reproducible.
+                with open(golden_path, encoding="utf-8") as file:
+                    golden_text = file.read()
+                with open(golden_path, "w", encoding="utf-8", newline="\n") as file:
+                    file.write(_sanitize_golden_stage_text(golden_text))
```
nvsekkin, line 353: I see this generates the new set, was this codepath used for the inital set?

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -187,19 +195,182 @@ def _make_sensor_data_type_params(
...
+        ``transforms`` maps each :class:`~pxr.UsdGeom.Xformable` prim path to its 4x4 local-to-world
+        transform as a ``float64`` array.
+    """
+    from pxr import Usd, UsdGeom  # noqa: PLC0415
+
+    stage = Usd.Stage.Open(usd_path)
+    if stage is None:
+        raise RuntimeError(f"Failed to open USD stage at {usd_path}.")
+
+    structure: dict[str, str] = {}
```
nvsekkin, line 230: question: I see we use a mapping from prim paths to structure (type name), and transform for the comparison below. This is a good approach. 

What made you pick these two specifically for the comparison and not say meshes, textures (do we rely on the image comparison for those?)

nvsekkin:
the overall structure looks good to me I have a few non-blocking questions

source/isaaclab_tasks/test/golden_stages/cartpole/newton-isaacsim_rtx_renderer-rgb.usda
mataylor-nvidia: The golden .usda files are generated by the test itself as a bootstrap: when the golden doesn't exist, maybe_save_stage() in [rendering_test_utils.py](vscode-webview://0t1bnjqfgs5e1104e3du440s6afdflffegkmuvahkiqs7lfru4sh/source/isaaclab_tasks/test/rendering_test_utils.py) writes it and then intentionally fails the test so you can review and commit the file.

source/isaaclab_tasks/test/golden_stages/registered_tasks_Isaac-Reorient-Cube-Shadow-Camera-Direct/default_physics-default_renderer-stage.usda
mataylor-nvidia: The Reorient-Cube stage contains inlined mesh geometry.

Do you think I should look into minimizing these files. 10mb is a lot worse than 50kb but still much smaller than the renders luckily.

source/isaaclab_tasks/test/rendering_test_utils.py
mataylor-nvidia: If there is a big change in scene structure this should break the tests and cause an error so goldens can be regenerated. (possible changes to the task)

I mainly wanted to focus on transforms, this is where we can tell if the simulation data is correct.

Mesh vertex data is large and doesn't change unless assets are re-exported.
similarly for textures which are visual and tested by the goldens.

source/isaaclab_tasks/test/rendering_test_utils.py
mataylor-nvidia: yes, this is the initial generation path.

source/isaaclab_tasks/test/golden_stages/registered_tasks_Isaac-Reorient-Cube-Shadow-Camera-Direct/default_physics-default_renderer-stage.usda
nvsekkin: thanks for the explanation, makes sense. since it's on LFS, maybe this is OK to keep. also I am not sure what would minimizing the file would look like. (maybe write our the paths:type/transform into a dict/json?) but keeping the USD seems like a better approach.

nvsekkin (approved):

PR #6398, opened 2026-07-07
Adds DisplayPort cable insertion envs

shauryadNv:
# Description

This PR adds DisplayPort cable insertion training and ROS-inference environments for the Flexiv Rizon4s with Grav gripper.

The environment is added under `isaaclab_tasks.contrib.deploy.cable_insertion` and includes:

- DisplayPort plug/socket assets
- Flexiv Rizon4s + Grav gripper joint-space training environment
- Flexiv Rizon4s + Grav gripper task-space training environment
- ROS-inference environment (both joint and task space)
- no-joint-velocity policy variant (joint-space)
- play variants for visualization/debugging
- DisplayPort insertion rewards, termination terms, reset/curriculum events, and success metrics
- policy deployment documentation for training and inference
- LEAPP-exportable action terms
- LEAPP export support for task-space policies

The main trainable task variants are:

- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-v0`
- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-NoJointVel-v0`
- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-TaskSpace-v0`

The main ROS-inference task variants are:

- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-ROS-Inference-v0`
- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-NoJointVel-ROS-Inference-v0`
- `Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-TaskSpace-ROS-Inference-v0`

The DisplayPort environment uses:

- physics frequency: `240 Hz`
- decimation: `8`
- effective control frequency: `30 Hz`
- action scale 0.025 -- relative joint-position (rad) in joint-space; Cartesian pose-delta (0.025 m / 0.025 rad) in task-space
- plug/socket mate-point success threshold: `3 mm`

The environments also log insertion metrics:

- `Metrics/success_rate`
- `Metrics/plug_socket_pos_error_m`
- `Metrics/plug_socket_keypoint_dist_m`
- `Metrics/terminal_success_rate`

No new external dependencies are required.

Fixes # N/A

## Type of change

- Documentation update

## Screenshots

N/A

## Checklist


source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,364 @@
...
+from isaaclab.utils.noise import UniformNoiseCfg
+
+import isaaclab_tasks.contrib.deploy.mdp as mdp
+from isaaclab_tasks.contrib.deploy.mdp.noise_models import ResetSampledConstantNoiseModelCfg
+
+CABLE_INSERTION_DIR = os.path.dirname(os.path.abspath(__file__))
+DISPLAY_ASSETS_DIR = os.path.join(CABLE_INSERTION_DIR, "display_cable_insertion_assets")
+
+
+def _quat_rotate_vec(q_xyzw, v):
```
ashwinvkNV, line 40: can use utils from math.py or some other library thats used in isaaclab already? I don't think we should have the quat utils here in-line.

source/isaaclab/isaaclab/utils/leapp/export_annotator.py
```diff
@@ -644,18 +659,23 @@ def _collect_processed_action_fallbacks(self, action_manager) -> list[TensorSema
             if processed is None:
                 continue
             if isinstance(processed, torch.Tensor):
-                logger.warning(
-                    "Action term '%s' did not write to any asset directly. Falling back to processed_actions as the"
-                    " export output.\nIf you wish to add semantic data to this policy, you need to manually annotate it"
-                    " with output_tensors.",
-                    term_name,
-                )
+                element_names = getattr(term, "_leapp_processed_action_element_names", None)
```
frlai: I understand why you need to do this. But is there any way to add this without modifying library code?

source/isaaclab/isaaclab/utils/leapp/export_annotator.py
```diff
@@ -644,18 +659,23 @@ def _collect_processed_action_fallbacks(self, action_manager) -> list[TensorSema
...
+                kind = getattr(term, "_leapp_processed_action_kind", None)
+                extra = getattr(term, "_leapp_processed_action_extra", None)
```
frlai: extra should also never be extensively used in library code. It you are free to use it on your own but by injecting data into extra you are creating ad hoc fields that downstream deployment frameworks may not honor. Useful as a one off but not generic enough.

frlai (changes requested):
Hey @ashwinvkNV @shauryadNv ,

It would be best to not modify export.py. This file is considered generic user facing entry point similar to the original train.py and play.py. If you need to make modifications to the to enable cable insertion, it would be best if you can create your own export.py under cable_insertion to maintain consistency. you can most likely reuse the entire main function and just inject your own configuration functions. 


for the terms needed to generate semantics on processed output, I like that you embedded it with the action term but ideally you don't inject that directly into main library code as well. I think you should be able to patch these for your needs if you write your own export.

j3soon:
Reproduced this branch from a clean clone on latest commit 015b11331bd73b3fa23e4c65dbaccf947ff84c65, using the container workflow on an RTX PRO 6000 Blackwell. Just opened a PR https://github.com/shauryadNv/IsaacLab/pull/10 that fixes issues (such as `--headless` flag deprecation in 3.x) encountered during the reproduction. Feel free to merge directly or cherrypick useful commits.

Three additional suggestions/question:
- Training appears to take more than 5 days on a single RTX PRO 6000. Providing a pretrained checkpoint would make testing and reviewing this work much easier.
- I only trained for a few iterations to verify the workflow. Is 5+ GPU-days expected for this policy? Most IsaacLab RL examples seem to train within hours to about a day. It would be helpful to document the iterations needed for a usable policy, e.g. ~50% success, versus full training, e.g. ~95%.
- (Optional) Since there's no cable simulation in this, maybe rename to `DP plug/connector insertion` (instead of `DP cable insertion`) can better reflect the task.

| Default View | Zoomed In View |
|---|---|
| <video src="https://github.com/user-attachments/assets/9e44df73-ac37-4b0b-9e75-d93d53dcfddb"> | <video src="https://github.com/user-attachments/assets/847e3ecc-d24c-49b1-a7d1-1dcd940e60c2"> |

(tested on policy trained for only ~25 iterations for validating the inference code path)

> Note: I guided Claude to reproduce the walkthrough, draft the fixes and validate them. The remainder is drafted by Claude.

<hr>

## What I found

Every command producing a deployable artifact failed, three of them silently. The
environment itself is untouched here: all nine task ids instantiate, the observation and
action contracts match the tutorial, and a task-space policy trained on this branch inserts
to 1.94 mm in deterministic play.

**1. Task-space LEAPP export writes no model, and reports validation as passing.** The
recommended deployment path. Exits 0 with a success banner, writing no `.onnx` and
`model_path: null`:

```
[WARNING]: Model DisplayPortTaskSpace does not have a compiled model. Skipping validation.
[INFO]:   Passed: 1/1
```

Skipped validation still counts as passing, so the tutorial ships a package with no policy
and reports success. `export_displayport_insertion.py:320` forwards `args_cli.export_method`,
default `None`, so LEAPP compiles nothing. Line 225 already defaults it on the other path.

**2. Joint-space LEAPP export aborts on the default backend.**

```
DispatchError: No ONNX function found for <OpOverload(op='aten.masked_select', ...)>
```

`rigid_object_quat_w` canonicalises the socket quaternion with boolean-mask assignment;
`isaaclab.utils.math.quat_unique` does the same as a `torch.where`. Task space is
unaffected, using the 6D rotation term. Fixing only that, the graph is then rejected:

```
INVALID_GRAPH: Type 'tensor(int32)' of input parameter (val_204) of operator (GatherND)
```

`JointAction` stores `find_joints` output as warp-backed `int32`; ONNX `GatherND` needs
int64. Torch accepts int32 indexing, so it only appears once traced. Any manager-based task
exporting a joint action hits this.

**3. `DP_ROBOT_USD` is documented but never read.** Named twice, including in the
`important` block telling the reader to run `flexiv_calibration` and select the result with
it. No Python file references it. Fails silently: calibrate, export the variable, train for
days against the shipped `Rizon4s-063459`, discover it on hardware.

Related: the control spaces start from different robots. Joint space spawns the calibrated
asset, task space never set a `usd_path` and spawns the stock arm.

**4. The training commands do not run, twice over.**

```
train.py: error: unrecognized arguments: --headless
```

Headless is now the default; `--visualizer` opts into a GUI. Five occurrences, though Step 1
already uses the current flag. Removing it exposes `ModuleNotFoundError` for MoviePy, which
the container does not ship and the page never mentions.

**5. Smaller gaps.** Two of three station-pose definitions are documented, omitting
`task_space_ros_inference_env_cfg.py`, so the recommended control space is the one never
mentioned. The play examples claim to apply unchanged to task space while passing the
joint-space pose, 10 mm off in Z. And `save_interval = 50` puts the first trained checkpoint
at `model_50.pt` (`model_0.pt` is untrained), so the export half needs a long run first,
unmentioned.

## What this PR changes

| Commit | Change |
|---|---|
| `187bfd02` | Resolve the export backend once, so the task-space path stops forwarding `None` |
| `e3f59262` | Use `quat_unique`; normalise `JointAction` indices to int64 |
| `4af575ec` | Add `resolve_robot_usd()`, use it in both control spaces, document the defaults |
| `6a492f1b` | Remove `--headless`, add the MoviePy note and prerequisites |
| `b4bbb751` | Document the task-space station pose, correct the play example |
| `f0099780` | Note `agent.save_interval` for reaching the export step sooner |
| `318bb571` | Changelog fragments |
| `b7b766f6` | Make `play_displayport_insertion.py` pass the pre-commit hooks |
| `4fd7b33a` | Regression tests for the robot USD override |

Defaults are unchanged. `DP_ROBOT_USD` passes to the spawner untouched, so a local path,
Nucleus URL or HTTP(S) URL all work, and the spawner still raises on an unreachable one.

`b7b766f6` is unrelated, included because the file fails `ruff`, `ruff format` and
`codespell` as it stands, so `uv run isaaclab -f` errors on a clean checkout. Three needed a
decision, each following existing convention here: `contextlib.suppress` for SIM105,
`# noqa: C901` on `main` as five other entrypoints do, and renaming the local `rin` rather
than growing `ignore-words-list`.

## Validation

Isaac Sim 6.0.0, LEAPP 0.6.1, ONNX Runtime 1.29.0. Checkpoints were trained on this branch
and are not converged; they only need to export.

| Check | Before | After |
|---|---|---|
| Task-space export, documented command | exits 0, no `.onnx`, validation skipped but passing | `.onnx` written, metadata populated, validation runs |
| Joint-space export, default backend | `DispatchError` | completes, graph loads, passes |
| Joint-space export, quaternion fix alone | n/a | `INVALID_GRAPH`, so both halves are needed |
| `DP_ROBOT_USD` local / `omniverse://` / HTTP(S) | ignored | honoured in both spaces |
| `DP_ROBOT_USD` unset, empty, whitespace | n/a | falls back to each default |
| Documented training command | `--headless`, then MoviePy error | runs, records 200 frames |
| Nine task ids, 200-step play rollout | pass | unchanged |
| `check-changelog-fragments`, `play_*.py` hooks | fail | pass |

`quat_unique` is bit-identical to the mask assignment over 2000 random unit quaternions.

The tests are 18 parametrised cases, 1.5 s, no GPU or Kit. They were confirmed to fail
without the fix against two broken states: configs not calling the helper, and the helper
validating via `os.path.isfile`, which rejects the remote URLs, hence those cases. The
export-backend and ONNX fixes have no collected tests, being observable only by a real
export.

<hr>

Additional info, kept out of the docs as the figures are machine-specific.

At `--num_envs 256` an iteration takes ~315 s, of which 309.2 s is physics collection and
5.7 s the PPO update: ~416 env steps/s, 38 GB VRAM, 92 % utilisation at 110 W of 300 W. The
shipped `max_iterations = 1500` is therefore ~5.5 days on one GPU, unstated on a page that
presents single-GPU training as the default. The workload is contact-latency bound, so more
GPUs or cheaper contact settings help where a faster GPU does not.

## Not addressed here

The two deployment stations still disagree on position; that is a hardware question, so I
documented it rather than picking one.

shauryadNv:
run-ci

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/config/displayport_rizon_4s/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    "isaaclab_tasks.contrib.deploy.cable_insertion.insertion_env:DisplayportInsertionEnv"
+)
+
+##
+# Register Gym environments.
+##
+
+# Flexiv Rizon 4s - Joint space
+gym.register(
+    id="Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-v0",
```
kellyguo11: contrib tasks should start with IsaacContrib

docs/source/policy_deployment/index.rst
```diff
@@ -10,6 +10,7 @@ Below, you'll find detailed examples of various policies for training and deploy
 
     01_io_descriptors/io_descriptors_101
     02_gear_assembly/gear_assembly_policy
+    06_displayport_insertion/displayport_insertion_policy
```
kellyguo11, line 13: should this replace the above tutorial? is there value to keeping both?

scripts/reinforcement_learning/deploy/play_displayport_insertion.py
```diff
@@ -0,0 +1,2101 @@
...
+tools can overlay sim and real rollouts.
+
+This is intentionally separate from ``play.py`` so DisplayPort-specific pose and
+logging knobs stay out of the generic play path.
+
+Example (recommended shipping task, RSL-RL checkpoint):
+
+.. code-block:: bash
+
+    ./isaaclab.sh -p scripts/reinforcement_learning/deploy/play_displayport_insertion.py \\
```
kellyguo11: commands should use uv run instead

scripts/reinforcement_learning/deploy/play_displayport_insertion.py
```diff
@@ -0,0 +1,2101 @@
...
+)
+from isaaclab_tasks.utils import get_checkpoint_path, setup_preset_cli
+from isaaclab_tasks.utils.hydra import hydra_task_config
+
+from success_utils import SuccessTracker  # isort: skip
+
+_DEFAULT_TASK = "Isaac-Deploy-DisplayportInsertion-Rizon4s-Grav-NoJointVel-ROS-Inference-v0"
+
+# -- argparse ----------------------------------------------------------------
+parser = argparse.ArgumentParser(
```
kellyguo11, line 153: is this a common script for deployment or something very specific to the task? it's also an extremely long script...

scripts/reinforcement_learning/deploy/success_utils.py
```diff
@@ -0,0 +1,87 @@
...
+``deploy.cable_insertion.insertion_env.DisplayportInsertionEnv``). For other
+environments the tracker disables itself gracefully.
+"""
+
+from __future__ import annotations
+
+import torch
+
+
+class SuccessTracker:
```
kellyguo11, line 19: this also doesn't feel like should belong here?

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,369 @@
...
+
+import isaaclab_tasks.contrib.deploy.mdp as mdp
+from isaaclab_tasks.contrib.deploy.mdp.noise_models import ResetSampledConstantNoiseModelCfg
+
+CABLE_INSERTION_DIR = os.path.dirname(os.path.abspath(__file__))
+
+# DisplayPort assets are published on the Isaac staging asset server. They are not yet mirrored
+# to the production bucket that ``ISAAC_NUCLEUS_DIR`` resolves to, so the root is pinned here.
+# Switch to ``f"{ISAAC_NUCLEUS_DIR}/Props/Factory/display_port_cable_assets"`` once promoted.
+_DISPLAY_ASSETS_STAGING_ROOT = "https://omniverse-content-staging.s3-us-west-2.amazonaws.com/Assets/Isaac/6.0/Isaac"
```
kellyguo11, line 43: this path should not be here

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,369 @@
...
+    ty = 2.0 * (qz * vx - qx * vz)
+    tz = 2.0 * (qx * vy - qy * vx)
+    return (
+        vx + qw * tx + qy * tz - qz * ty,
+        vy + qw * ty + qz * tx - qx * tz,
+        vz + qw * tz + qx * ty - qy * tx,
+    )
+
+
+def _quat_mul(q1_xyzw, q2_xyzw):
```
kellyguo11: we should have these utilities already defined in utils, are these really necessary?

ooctipus: +1

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,369 @@
...
+    z_clearance=0.033,
+)
+
+##
+# Asset Configurations
+##
+
+
+@configclass
+class DisplayPortPlug(RigidObjectCfg):
```
ooctipus, line 126: I don't think this is standard way we define 

please do

DisplayPortPlug=RigidObjectCfg(....)

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,369 @@
...
+            max_contact_impulse=None,
+        ),
+        mass_props=sim_utils.MassPropertiesCfg(mass=0.03),
+        collision_props=sim_utils.CollisionPropertiesCfg(contact_offset=0.00001, rest_offset=-0.00005),
+    )
+    init_state = RigidObjectCfg.InitialStateCfg(pos=_PLUG_ROOT_POS, rot=_DEFAULT_PLUG_ROT)
+
+
+@configclass
+class DisplayPortSocket(RigidObjectCfg):
```
ooctipus, line 154: similar for this

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/displayport_insertion_env_cfg.py
```diff
@@ -0,0 +1,369 @@
...
+
+    replicate_physics = True
+
+    ground = AssetBaseCfg(
+        prim_path="/World/ground",
+        spawn=sim_utils.GroundPlaneCfg(),
+        init_state=AssetBaseCfg.InitialStateCfg(pos=(0.0, 0.0, -1.05)),
+    )
+
+    dp_plug = DisplayPortPlug()
```
ooctipus, line 198: please use dp_plug = RigidObjectCfg(...)

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/insertion_env.py
```diff
@@ -0,0 +1,124 @@
...
+from isaaclab.utils.math import combine_frame_transforms
+
+
+def _keypoint_offsets_6d(device: torch.device) -> torch.Tensor:
+    """Return the 7 unit keypoint offsets used by the keypoint reward."""
+    corners = torch.tensor([[0, 0, 0], [1, 0, 0], [0, 1, 0], [0, 0, 1]], device=device, dtype=torch.float32)
+    return torch.cat((corners, -corners[-3:]), dim=0)
+
+
+class DisplayportInsertionEnv(ManagerBasedRLEnv):
```
ooctipus, line 27: if you are using manager based environment please don't create subclass, manager base environment is meant to be fully cfg driven with no central envrionment implementation code, sublcass only apply to direct environment

scripts/reinforcement_learning/deploy/play_displayport_insertion.py
ooctipus, line 1: instead of do this I recommend writing all the play logic into the environment cfg and use the generic play script rather than writing a separate ply disport script

scripts/reinforcement_learning/deploy/success_utils.py
ooctipus, line 19: again this suppose to be in envrionment logic doesn't belong in scripts

scripts/reinforcement_learning/deploy/play_displayport_insertion.py
StafaH: Seconded, please remove this file and `success_utils.py`

scripts/reinforcement_learning/leapp/rsl_rl/export_displayport_insertion.py
```diff
@@ -0,0 +1,358 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: we moved all leapp logic out of scripts. In general we are trying to remove all scripts/ into module logic, so this logic has to somehow be folded in with leapp. If leapp is not providing the correct export logic, then work with @frlai to figure out how to allows users to create "custom" export logic and still use export.py. Then the code can be moved into environment logic and leapp export can pick up the custom logic during the export.

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/config/displayport_rizon_4s/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    disable_env_checker=True,
+    kwargs={
+        "env_cfg_entry_point": f"{__name__}.joint_pos_env_cfg:Rizon4sGravDisplayportInsertionNoJointVelEnvCfg",
+        "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:Rizon4sGravDisplayportInsertionRNNPPORunnerCfg",
+    },
+)
+
+# Flexiv Rizon 4s - Joint space without joint velocity Play
+gym.register(
+    id="IsaacContrib-Deploy-DisplayportInsertion-Rizon4s-Grav-NoJointVel-Play",
```
StafaH, line 53: We no longer recommend registering play and inference environments, and building seperate configs. You can override the`play_mode` function, and our play script will pick up your logic when play is called.

source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/cable_insertion/config/displayport_rizon_4s/__init__.py
```diff
@@ -0,0 +1,120 @@
...
+    disable_env_checker=True,
+    kwargs={
+        "env_cfg_entry_point": f"{__name__}.joint_pos_env_cfg:Rizon4sGravDisplayportInsertionEnvCfg",
+        "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:Rizon4sGravDisplayportInsertionRNNPPORunnerCfg",
+    },
+)
+
+# Flexiv Rizon 4s - Joint space Play
+gym.register(
+    id="IsaacContrib-Deploy-DisplayportInsertion-Rizon4s-Grav-Play",
```
StafaH, line 31: All of these environments could be one and use a preset over the action space.

StafaH (changes requested):
High level comments:
- We currently name things IsaacContrib-Action-Specific-Robot, where action = insert, specific = displayport, and robot = Rizon4s. This allows for in the future other robots, other specifics, and other actions (e.g. IsaacContrib-Unplug-Displayport-Rizon4s might be another task)
- Presets can be used to probably help collapse things down into less environments. More manageable
- Lots of old IsaacLab conventions are being used (e.g. _PLAY env) which will need to be removed.
- If LEAPP does not support something, you will need to work together to figure out how to cleanly implement customization, we won't be able to add custom scripts for one task

PR #6404, opened 2026-07-07
Fix double environment reset at episode boundaries in record_demos.py

asierarranz:
# Description

When recording teleop demos, every episode boundary resets the scene twice, which glitches in the headset; this makes it reset once.

**Root cause:** `handle_reset()` in `scripts/tools/record_demos.py` calls `env.sim.reset()` right before `env.reset()`. The `sim.reset()` stop/play cycle snaps every body back to its authored spawn pose, and then the managed reset moves everything again to the actual reset state. The pilot sees the scene jump twice, and the first recorded frames of the next episode can capture the stale spawn poses.

**Fix:** drop the `env.sim.reset()` call at the episode boundary. `env.reset()` alone restores the managers, events and recorder state. The initial `sim.reset()` before the recording loop starts is left as is, since that one runs before any episode exists and is needed to start the simulation.

Verified by flying XR recording sessions on a Quest 3: episode transitions now show a single clean reset and the first frames of each new episode match the reset state.

No changelog fragment: the change only touches `scripts/`, no `source/<pkg>` package.

## Type of change

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -57,6 +57,7 @@ Guidelines for modifications:
 * Antonin Raffin
 * Arjun Bhardwaj
 * Ashwin Varghese Kuruttukulam
+* Asier Arranz
 * Bikram Pandit
 * Bingjie Tang
 * Bocheng Zou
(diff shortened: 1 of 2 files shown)


StafaH:
Partially resolved by #6650 , the double reset is still there but more intentional

PR #6278, opened 2026-06-26
Removes rendering mode presets

nvsekkin:
# Description

Removes the three RTX rendering mode presets (`performance`, `balanced`, `quality`) and the machinery that processed them. With RTX Minimal mode now serving the high-performance path, the presets provide little benefit.

The former `quality` settings are retained as high-fidelity defaults, applied automatically when camera (RGB) rendering is enabled. Users who need high performance should switch to the RTX Minimal renderer.

## Breaking change
- `--rendering_mode` and `RenderCfg.rendering_mode` are removed.
- Drop `--rendering_mode` / `RenderCfg(rendering_mode=...)`. High-fidelity
  (former `quality`) rendering is now the default when cameras are
  enabled.
- Override individual settings via `RenderCfg` fields or `carb_settings`.
- For high-performance rendering, use the RTX Minimal renderer.

## Type of change

- Breaking change (existing functionality will not work without user modification)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6405, opened 2026-07-08
Removes rendering mode presets

nvsekkin:
# Description

Continues #6305 (includes @pbarejko's `Remove quality settings` commit) and supersedes the now-closed #6278. Branched on top of the post-#6279 changes.

Removes the three RTX rendering-mode presets (`performance`, `balanced`, `quality`) and the machinery that loaded them. With the RTX Minimal renderer now serving the high-performance path, the presets added little value. The former `quality` settings are retained as high-fidelity defaults baked into the camera rendering Kit experiences, applied automatically when RGB camera rendering is enabled.

Updated 18 rgb/rgba golden images (isaacsim_rtx renderer, both physx/newton) from the CI-rendered output, since baking the high-fidelity RTX defaults into kit changes the composited color: confirmed the shift is color-only, so no geometry/segmentation goldens were touched.

## Breaking change
- Removed the `--rendering_mode` launcher argument and `IsaacRtxRendererGlobalSettingsCfg.rendering_mode`, plus the `apps/rendering_modes/` presets.
- Baked high-fidelity RTX defaults into `apps/isaaclab.python.rendering.kit` and the headless variant.

## Type of change

## Checklist


docs/source/how-to/configure_rendering.rst
```diff
@@ -10,89 +10,31 @@ Configuring RTX Rendering Settings
    For the **Newton renderer** (used with the Newton backend or in kit-less mode), see
    :ref:`overview_renderers` for the pluggable renderer architecture and available backends.
 
-Isaac Lab's RTX renderer offers 3 preset rendering modes: performance, balanced, and quality.
-You can select a mode via a command line argument or from within a script, and customize settings as needed.
-Adjust and fine-tune rendering to achieve the ideal balance for your workflow.
+Isaac Lab's RTX renderer applies high-fidelity camera rendering defaults when
+running with ``--enable_cameras``. The former rendering mode presets
+(``performance``, ``balanced``, and ``quality``) have been removed; tune the
```
pbarejko: What is the reason to keep old, unsupported behavior as documentation? I would suggest to remove it, this note should be included in change log. Perhaps entry of performance vs quality should be outlined as RT2 vs Minimal mode with a link to documentation?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
```diff
@@ -143,8 +102,10 @@ def _apply_isaac_rtx_global_settings(
             import omni.replicator.core as rep
 
             rep.settings.set_render_rtx_realtime(antialiasing=antialiasing_mode)
-        except Exception:
-            pass
+        except ImportError:
+            logger.warning("omni.replicator.core is unavailable; skipping antialiasing_mode=%s.", antialiasing_mode)
```
pbarejko: This except is captures too broadly. Can you elaborate why you had to put a warning with hard coded omni.replicator.core?

nvsekkin: this was a `pass` before, I added following a [review ](https://github.com/isaac-sim/IsaacLab/pull/6278/changes#r3494681218) in the previous PR, since the try block has the import call [here](https://github.com/isaac-sim/IsaacLab/pull/6405/changes/BASE..764f31879b5b79ce194db12236621287ad5d28f5#diff-8540e4afa518425988e26ef03fe6917f99ac6629b26ccff994069de2ccd5f284R102) 

I agree I can wrap this in 2 separate try/except - one for import one for the call to settings?

source/isaaclab/test/sim/test_simulation_render_config.py
```diff
@@ -112,6 +141,10 @@ def test_isaac_rtx_global_settings(monkeypatch):
         enable_cached_raytracing=True,
         max_samples_per_launch=500000,
         view_tile_limit=500000,
+        carb_settings={
+            "rtx.raytracing.subpixel.mode": 2,
+            "/rtx/custom/setting": True,
+        },
```
rilei-nvidia, line 147: curious why you added this and how it relates to this PR - was it to validate that these carb settings eventually make it into the global RTX settings?

rilei-nvidia: ahh this is to test the overriding aspect right?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
rilei-nvidia: The warnings are certainly better than what it was before - and 2 separate exceptions sounds like a better approach

I'm curious why this try except is needed in the first place:
1. Why is omni.replicator.core import deferred and guarded by an exception?  Is it possible for an Isaac Sim installation to not have the replicator extension?   Since Replicator is also used as interface to actually do rendering, not being able to set the anti aliasing mode would be the least of the problems!
2. Why would `rep.settings.set_render_rtx_realtime(antialiasing=antialiasing_mode)` fail?  Does the API document that it can throw an exception?

rilei-nvidia (approved):
@nvsekkin great clean-up and refactor, I had some non-blocking questions, LGTM!

pbarejko (approved):

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
nvsekkin: good point, this was an existing pattern and I took it at face value. 

I think you are right, 1) it should be safe to remove this guard. I can remove the try/except block altogether. 2) same reason, this call was in the try/except block before.

source/isaaclab/test/sim/test_simulation_render_config.py
nvsekkin: yes this is to test the `carb_settings` override

PR #6413, opened 2026-07-08
[Task Clean-up] Dexterous Part 3/9: Add success-rate metrics to the reorientation Direct tasks

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution.
- Links pin specific SHAs and can go stale after a branch update — the table on #6324 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| #6411 Part 1/9: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/9: OVPhysX articulation + manager runtime |  | — | merged |
| 📌 #6413 Part 3/9: Reorient Direct, torch (this PR) |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6413/changes/f4895f0f9ee..d29afc75e71) |
| #6414 Part 4/9: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/d29afc75e71..b10a84948f8) |
| #6418 Part 5/9: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/d29afc75e71..707d37f8f99) |
| #6421 Part 6/9: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/707d37f8f99..b50ac8906fc) |
| #6410 Part 7/9: Environment overview docs |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6410/changes/b50ac8906fc..af259c0778d) |
| #6415 Part 8/9: Benchmark success-rate utilities |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/b50ac8906fc..c7f2f019d8b) |
| #6582 Part 9/9: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |
| #6324 [DO-NOT-MERGE] Lumped validation reference |  | ALL | — |

## Summary

[...]

PR #6412, opened 2026-07-08
[Task Clean-up][OVPhysX] Dexterous Part 2/8: Fix articulation and manager runtime

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution.
- Links pin specific SHAs and can go stale after a branch update — the table on #6324 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| #6410 [Docs] Environment overview regen |  | — | — |
| #6411 Part 1/8: Newton cloner/cubric/visualizer fixes |  | — | — |
| 📌 #6412 Part 2/8: OVPhysX articulation + manager runtime (this PR) |  | — | — |
| #6413 Part 3/8: Reorient Direct, torch |  | — | — |
| #6414 Part 4/8: Handover + camera Direct, torch |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/6e8a63e4e028b2d43676ea30c446b9dc9068c7b5..a468ed0b86d8f4b47d5c9809163d04428c2dd545) |
| #6418 Part 5/8: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/6e8a63e4e028b2d43676ea30c446b9dc9068c7b5..f8b36113f4917b2f9fb56b15fb7ea9e44c95f121) |
| #6421 Part 6/8: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/37f9ee65b64852068c3fca6e22357c759fe329e2..b27cb241d876820c14d0718d7266b5dadb272a24) |
| #6415 Part 7/8: Benchmark success-rate utilities + docs |  | #6413, #6414, #6418, #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/b27cb241d876820c14d0718d7266b5dadb272a24..df070af53a6f499e331d5972d3da21eb80a9debd) |
| #6582 Part 8/8: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/6e8a63e4e028b2d43676ea30c446b9dc9068c7b5..4ec232ae09283ddfc58dbaca1c6ec90ec2b6b4c3) |
[...]

PR #6411, opened 2026-07-08
[Task Clean-up][Newton] Dexterous Part 1/8: Fix cubric fallback and visualizer teardown

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution.
- Links pin specific SHAs and can go stale after a branch update — the table on #6324 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| #6410 [Docs] Environment overview regen |  | — | — |
| 📌 #6411 Part 1/8: Newton cloner/cubric/visualizer fixes (this PR) |  | — | — |
| #6412 Part 2/8: OVPhysX articulation + manager runtime |  | — | — |
| #6413 Part 3/8: Reorient Direct, torch |  | — | — |
| #6414 Part 4/8: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/6e8a63e4e028b2d43676ea30c446b9dc9068c7b5..5cb00e7cb5cc813b202521272e043007cd255194) |
| #6418 Part 5/8: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/79f87501ac4c81de93a71dab00dc443da62113aa..e7c9a9a3fae3a7972b0c5165ae683abffb7d0e0f) |
| #6421 Part 6/8: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/01c9f4d8c5c35a5688b2a5bb90209e16b8f81b99..835a5815ec49b11aada1d20a76c177054505e6e7) |
| #6415 Part 7/8: Benchmark success-rate utilities + docs |  | #6413, #6414, #6418, #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/e1abb6b1790ccc57af42551eebccf743633f1f13..d6348539aa8032d9668c20a8fea462c5d88d3af9) |
| #6582 Part 8/8: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/21dbb1769c4e30c8e9e5b0f563c2dae24c230349..83e1587cadd9712a60615ed2a3cb2d177c2ac24d) |
[...]

PR #6410, opened 2026-07-08
[Task Clean-up][Docs] Dexterous Part 7/10: Environment overview updates for the reorientation series

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution.
- Links pin specific SHAs and can go stale after a branch update — this table on #6324 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| #6411 Part 1/10: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/10: OVPhysX articulation + manager runtime |  | — | merged |
| #6413 Part 3/10: Reorient Direct, torch |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6413/changes/f4895f0f9ee..d29afc75e71) |
| #6414 Part 4/10: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/d29afc75e71..b10a84948f8) |
| #6418 Part 5/10: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/3ea6f7bbf6c..b31b04eaf71) |
| #6421 Part 6/10: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/b31b04eaf71..0cae3d87be4) |
| 📌 #6410 Part 7/10: Environment overview docs |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6410/changes/b50ac8906fc..af259c0778d) |
| #6954 Part 8/10: OpenAI Shadow variants → contrib (draft) |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6954/changes/0cae3d87be4..b0d2d8448af) |
| #6415 Part 9/10: Benchmark success-rate utilities |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/b50ac8906fc..c7f2f019d8b) |
| #6582 Part 10/10: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |
| #6324 [DO-NOT-MERGE] Lumped validation reference |  | ALL | — |

PR #6414, opened 2026-07-08
[Task Clean-up] Dexterous Part 4/9: Fix MARL-to-single-agent training and enable handover Direct RSL-RL

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution. This PR (#6414) is now rebased directly onto `develop`, so its own **Files changed** tab is its contribution.
- Links pin specific SHAs and can go stale after a branch update — the table on #6324 is refreshed first.

|---|---|---|---|
| #6411 Part 1/9: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/9: OVPhysX articulation + manager runtime |  | — | merged |
| #6413 Part 3/9: Reorient Direct, torch |  | — | merged |
| 📌 #6414 Part 4/9: MARL-to-single-agent fix + handover/camera Direct (this PR) |  | — (on develop) | Files changed tab |
| #6418 Part 5/9: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/d29afc75e71..707d37f8f99) |
| #6421 Part 6/9: Handover + camera manager counterparts |  | #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/707d37f8f99..b50ac8906fc) |
| #6410 Part 7/9: Environment overview docs |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6410/changes/b50ac8906fc..af259c0778d) |
| #6415 Part 8/9: Benchmark success-rate utilities |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/b50ac8906fc..c7f2f019d8b) |
| #6582 Part 9/9: Warp variants → experimental (draft; merges last) |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |

## Summary

Converts the two-hand Shadow Hand **handover** task from multi-agent (MARL) to **single-agent** (enabling RSL-RL training), fixes the underlying MARL→single-agent observation bridge, and consolidates the shared Shadow Hand definitions into the asset.

- **MARL → single-agent bridge (general, broader than this task).** Single-agent RL libraries train `DirectMARLEnv` tasks via `multi_agent_to_single_agent`; the bridge dropped the latest observations from the public buffer. Fixed generally — RSL-RL observations read from the env-owned `obs_buf`, stored by `reset` like `step` on all env bases (`DirectRLEnv`, `DirectMARLEnv`, the experimental warp base). Every MARL task + single-agent runner benefits; handover is the first consumer.
- **Handover Direct → single-agent + RSL-RL.** Single-agent conversion, RSL-RL runner config, shared identity in `handover_common`, success-rate reward metrics, and a fix for its Newton construction failure (see below).
- **Shadow Hand Newton robot → the asset.** Moved the Newton (MJWarp) robot cfg into `isaaclab_assets` as `SHADOW_HAND_NEWTON_CFG` (beside `SHADOW_HAND_CFG`); reorient and handover consume it as **siblings** (neither imports the other). Reorient uses the default gains; handover overrides only the finger stiffness/damping to `20/2` for its catch.
- **Shadow camera cleanup.** Renamed the camera Direct modules to `shadow_hand_direct_camera_env[_cfg]`; moved cube-keypoint math to the shared, unit-tested `reorient/mdp/observations.py` (replacing the per-env `compute_keypoints` shim); and added OVPhysX physics presets to the handover and camera Direct environments. The camera env config is otherwise unchanged from develop.

### Behavior changes

**Reorient-on-Newton is unchanged.** The Newton hand definition moves **verbatim** into `SHADOW_HAND_NEWTON_CFG` — same joint expressions, effort limits, gains, friction and armature — so its config diff vs `develop` is purely the move, and the Direct env config has no diff at all.

Handover changes in three ways:

- Its **default physics backend moves from PhysX to Newton** (MJWarp), matching the reorientation tasks. Pass `physics=physx` for the previous backend.
- Its `distal_passive` actuator group is **removed**. It targeted `robot0_(FF|MF|RF|LF)J0` — joints that **do not exist** on the Newton asset (its fingers are numbered `J1`–`J4`/`J5`), so constructing the environment raised `No joints found for actuator group`. The `fingers` group already drives those joints, and the comment justifying the group (USD-baked `stiffness=286/damping=57`) does not apply to this asset, which authors no drive stiffness at all.
- Two orientation fixes: the goal orientation was initialized to a 180-degree rotation instead of identity, and the Newton root rotation replaced the asset's baked rotation instead of composing with it, leaving both palms rotated 90 degrees.

Two pre-existing discrepancies were found while verifying against the asset and are left for a follow-up, since each is a behavior change needing its own validation: the config sets `armature=2e-3` where the asset authors `0.0002`, and the per-finger `J1`/`J2` pair is coupled by a fixed tendon that the MJWarp solver currently skips — the actuator gains are what hold that pair together today, which is why the `J4` knuckle-abduction joints cannot simply be added to the driven set.

## Notes

- Rebased onto **latest develop**; clean 3-commit history: `Add single-agent flattening for MARL Direct environments` / `Convert Shadow Hand handover to single-agent Direct with RSL-RL` / `Clean up Shadow reorient/camera tasks and consolidate the Newton hand`.

## Validation

- Kit-free suite green: handover, keypoint math, the MARL adapter, and new checks pinning the RSL-RL wrapper observation contract (verified to fail against the previous implementation).
- Camera Direct reaches training takeoff under `rsl_rl` (frame-verified videos); its env config now matches develop's established vision env.
- **Reorient-on-Newton vs `develop`** (`rsl_rl`, 2048 envs, 1500 iterations, 4 runs per config): this branch spans 524–664 reward, `develop` spans 654–822 — overlapping distributions whose per-config spreads (140–168) exceed the gap between them. The move is additionally verified at the config level: the Newton robot configuration is character-for-character identical to `develop`'s.
- **Handover-on-Newton** trains end-to-end on this branch (reward 1314), which the `develop` config cannot do at all — it fails during construction. Full evidence: #6324.


Toni-SM:
> Fixes the multi-agent-to-single-agent adapter to expose current observations through the Direct RL interface...

Such implementation should not be accepted as it is trying to fix a problem a RL library has (require a private method, that is not part of the standard API Gymnasium / Petting-Zoo Parallel API to initialize its logic) from Isaac Lab side. 

There is nothing to fix here. The fix must be implemented in the RL library to do not depend on the private `_get_observations` method that is not part of the standard API, but instead to rely on public properties and method of the standard environment API to initialize the library. 

@kellyguo11 for viz

source/isaaclab/isaaclab/envs/utils/marl.py
```diff
@@ -81,22 +81,38 @@ def __init__(self, env: DirectMARLEnv) -> None:
...
+                )
+            }
 
-            return obs, extras
+        def _get_observations(self) -> VecEnvObs:
+            """Return current observations through the single-agent interface."""
+            if self._state_as_observation:
+                # the state replaces the observations entirely; skip computing them
+                return {"policy": self.env.state()}
+            return self._convert_observations(self.env._get_observations())
```
Toni-SM: See https://github.com/isaac-sim/IsaacLab/pull/6414#issuecomment-4916444317

Toni-SM (dismissed):

source/isaaclab/isaaclab/envs/utils/marl.py
hujc7: Updated.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
```diff
@@ -72,14 +76,50 @@ def __init__(self, cfg: HandoverEnvCfg, render_mode: str | None = None, **kwargs
 
         # default goal positions
         self.goal_rot = torch.zeros((self.num_envs, 4), dtype=torch.float, device=self.device)
-        self.goal_rot[:, 0] = 1.0
+        self.goal_rot[:, 3] = 1.0  # identity quaternion in (x, y, z, w) layout
         self.goal_pos = torch.zeros((self.num_envs, 3), dtype=torch.float, device=self.device)
```
AntoineRichard, line 72: Why not use warp arrays directly?

AntoineRichard (commented):
My concern is that this PR is doing some real (important and needed) work on MARL and RSL-RL while hiding that under a RSL-RL only refactor.

Also same concerns around moving entirely to warp.

source/isaaclab_tasks/test/core/test_shadow_hand_camera_presets.py
```diff
@@ -180,6 +180,7 @@ def shadow_hand_camera_presets():
     ("default", ["rgb", "depth", "semantic_segmentation"]),
     ("full", ["rgb", "depth", "semantic_segmentation"]),
     ("rgb", ["rgb"]),
+    ("rgb_depth", ["rgb", "depth"]),
```
StafaH: rgbd?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
```diff
@@ -58,21 +58,21 @@
...
     kwargs={
-        "env_cfg_entry_point": f"{__name__}.shadow_hand_camera_env_cfg:ShadowHandCameraEnvCfg",
+        "env_cfg_entry_point": f"{__name__}.shadow_hand_direct_camera_env_cfg:ShadowHandCameraEnvCfg",
         "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:ShadowHandCameraFFPPORunnerCfg",
         "rl_games_cfg_entry_point": f"{agents.__name__}:rl_games_ppo_camera_cfg.yaml",
     },
 )
 
 gym.register(
     id="Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct",
```
StafaH, line 71: As discussed offline, we can move benchmark to contrib. (if this is handled in a future PR please ignore)

source/isaaclab/isaaclab/envs/utils/marl.py
```diff
@@ -81,22 +81,34 @@ def __init__(self, env: DirectMARLEnv) -> None:
...
-            # use environment state as observation
-            if self._state_as_observation:
-                obs = {"policy": self.env.state()}
-            # concatenate agents' observations
+        @episode_length_buf.setter
+        def episode_length_buf(self, value: torch.Tensor) -> None:
+            self.env.episode_length_buf = value
+
+        @property
+        def obs_buf(self) -> VecEnvObs:
```
StafaH, line 98: It might be cleaner to do something like: 

```
def reset(...):
    obs, extras = self.env.reset(seed, options)
    self._obs_buf = self._convert_observations(obs)
    return self._obs_buf, extras

def step(...):
    ...
    self._obs_buf = self._convert_observations(obs)
    return self._obs_buf, ...

@property
def obs_buf(self):
    return self._obs_buf
```

So obs buffer stays true to it's propery, instead of being a conversion mechanism

source/isaaclab/test/envs/test_marl_utils.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: AI Review: The PR changes RslRlVecEnvWrapper.get_observations() from actively computing observations to unconditionally reading env.obs_buf, but there is no wrapper-level regression test. Please add tests showing that get_observations() exactly matches the most recent reset and step outputs, including a noisy/non-idempotent observation case. We should also verify every supported environment family exposes obs_buf.

source/isaaclab/isaaclab/envs/utils/marl.py
```diff
@@ -81,22 +81,34 @@ def __init__(self, env: DirectMARLEnv) -> None:
...
+            """Episode lengths from the wrapped multi-agent environment."""
+            return self.env.episode_length_buf
 
-            # use environment state as observation
-            if self._state_as_observation:
-                obs = {"policy": self.env.state()}
-            # concatenate agents' observations
+        @episode_length_buf.setter
+        def episode_length_buf(self, value: torch.Tensor) -> None:
+            self.env.episode_length_buf = value
```
StafaH: This might need to be something like: `self.env.episode_length_buf.copy_(value)`? Not entirely sure though.

source/isaaclab/test/envs/test_marl_utils.py
```diff
@@ -0,0 +1,82 @@
...
+        pass
+
+
+def test_multi_agent_to_single_agent_reset_concatenates_agents():
+    """The adapter reset should concatenate the agents' observations."""
+    env = multi_agent_to_single_agent(_FakeMultiAgentEnv())
+
+    observations, _ = env.reset()
+
+    torch.testing.assert_close(observations["policy"], torch.tensor([[1.0, 2.0, 5.0], [3.0, 4.0, 6.0]]))
```
StafaH, line 61: AI Review: This only verifies that the property converts the initial dictionary. It does not verify “latest observation” behavior.

A stronger test would make reset() and step() return data different from env.obs_dict, or make state() return a different result on each call. That would catch unwanted recomputation.

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/__init__.pyi
```diff
@@ -0,0 +1,12 @@
...
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+__all__ = [
+    "handover_reward",
+    "evaluate_handover_success",
+]
+
+from .rewards import evaluate_handover_success, handover_reward
```
StafaH, line 11: let's use absolute imports, here and elsewhere

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
```diff
@@ -72,14 +68,18 @@ def __init__(self, cfg: HandoverEnvCfg, render_mode: str | None = None, **kwargs
...
-        self.goal_pos[:, :] = torch.tensor([0.0, -0.64, 0.54], device=self.device)
+        # goal = object default position + shared offset (mirrors HandoverCommand.__init__)
```
StafaH, line 73: Comments here and elsewhere need to be cleaned up, some are redundant and other are a bit too verbose

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
```diff
@@ -192,17 +192,21 @@ def _get_states(self) -> torch.Tensor:
 
     def _get_rewards(self) -> dict[str, torch.Tensor]:
         # compute reward
-        goal_dist = torch.linalg.norm(self.object_pos - self.goal_pos, ord=2, dim=-1)
-        rew_dist = 2 * torch.exp(-self.cfg.dist_reward_scale * goal_dist)
+        succeeded, goal_dist = evaluate_handover_success(
```
StafaH, line 195: The functions here are just abstracting away 1-2 lines of logic? Creating an MDP folder/module just to make a rewards.py just to abstract away 1 line seems a bit excessive (unless a future PR is doing something with that MDP)

source/isaaclab_assets/isaaclab_assets/robots/shadow_hand.py
```diff
@@ -81,7 +82,97 @@
...
+        # together: both ends are driven with identical gains and effort limits, which
+        # reproduces the 1:1 coupling. Keep them symmetric and in the same actuator group --
+        # driving one end while clamping the other makes the two fight, and an uncapped effort
+        # limit on either end diverges to NaN within a few hundred steps.
+        #
+        # Known limitation: the ``J4`` knuckle abduction joints (and ``LFJ5``) are left
+        # undriven, so the fingers cannot spread laterally. Correcting that requires resolving
+        # the skipped tendon first and is deferred to a follow-up.
+        "fingers": ImplicitActuatorCfg(
+            joint_names_expr=[
```
StafaH, line 129: Is WR wrist? If so shouldn't this be called "hand"

This might also be useful when creating subgroups for the actual "fingers" which might have different PD from other joints

source/isaaclab_experimental/isaaclab_experimental/envs/direct_rl_env_warp.py
```diff
@@ -455,8 +457,10 @@ def step(self, action: torch.Tensor) -> VecEnvStepReturn:
             self._post_step_visualize()
 
         # return observations, rewards, resets and extras
+        # store the returned buffer so RslRlVecEnvWrapper.get_observations() can read env.obs_buf
+        self.obs_buf = {"policy": self.torch_obs_buf.clone()}
```
StafaH, line 461: Here and above, .clone() is probably expensive. Is there a specific rationale for cloning rather than having a reference/view

source/isaaclab_rl/isaaclab_rl/rsl_rl/vecenv_wrapper.py
```diff
@@ -171,11 +171,7 @@ def reset(self) -> tuple[TensorDict, dict]:  # noqa: D102
 
     def get_observations(self) -> TensorDict:
         """Returns the current observations of the environment."""
-        if hasattr(self.unwrapped, "observation_manager"):
-            obs_dict = self.unwrapped.observation_manager.compute()
-        else:
-            obs_dict = self.unwrapped._get_observations()
-        return TensorDict(obs_dict, batch_size=[self.num_envs])
+        return TensorDict(self.unwrapped.obs_buf, batch_size=[self.num_envs])
```
StafaH, line 174: Let's add a test to make sure all envs implement obs_buf (unless that test is already there?)

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_common.py
```diff
@@ -0,0 +1,57 @@
...
+shared structural definitions so the two variants cannot drift apart. Every
+scalar task parameter reads as a flat ``cfg.<name>`` field on the env
+configuration; non-scalar items (joint/body-name lists, the goal offset, and
+marker templates) stay module-level constants.
+"""
+
+import isaaclab.sim as sim_utils
+from isaaclab.markers import VisualizationMarkersCfg
+
+from isaaclab_assets.robots.shadow_hand import (
```
StafaH, line 18: Cleaner to write:

```
from isaaclab_assets.robots.shadow_hand import (
    SHADOW_ACTUATED_JOINT_NAMES as ACTUATED_JOINT_NAMES,
    SHADOW_FINGERTIP_BODY_NAMES as FINGERTIP_BODY_NAMES,
)
```

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
```diff
@@ -72,14 +68,18 @@ def __init__(self, cfg: HandoverEnvCfg, render_mode: str | None = None, **kwargs
 
         # default goal positions
         self.goal_rot = torch.zeros((self.num_envs, 4), dtype=torch.float, device=self.device)
-        self.goal_rot[:, 0] = 1.0
+        self.goal_rot[:, 3] = 1.0  # identity quaternion in (x, y, z, w) layout
         self.goal_pos = torch.zeros((self.num_envs, 3), dtype=torch.float, device=self.device)
-        self.goal_pos[:, :] = torch.tensor([0.0, -0.64, 0.54], device=self.device)
+        # goal = object default position + shared offset (mirrors HandoverCommand.__init__)
+        self.goal_pos[:, :] = self.object.data.default_root_pose.torch[:, :3].to(self.device) + torch.tensor(
```
StafaH: Here and elsewhere, what's going on with this self.device stuff? Are the buffers on the wrong device? Everything should already be on the same device. I would remove all this .to(device) stuff, maybe add a test instead? to(device) could be expensive to always be calling in hot paths.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -274,69 +294,24 @@ class HandoverEnvCfg(DirectMARLEnvCfg):
...
-    # simulation
+    # simulation — values mirrored by the manager cfg (guarded by the value-parity test)
     sim: SimulationCfg = SimulationCfg(
         dt=1 / 120,
-        render_interval=decimation,
-        physics_material=RigidBodyMaterialCfg(
-            static_friction=1.0,
-            dynamic_friction=1.0,
-        ),
+        render_interval=2,
```
StafaH: here and elsewhere let's make the default for base envs be `render_interval = decimation`? Seems like default behaviour should be to render at every physics step?

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -124,61 +131,69 @@ class EventCfg:
...
+    newton_rot = tuple(
+        math_utils.quat_mul(
+            torch.tensor(init_rot, dtype=torch.float64),
+            torch.tensor(SHADOW_HAND_NEWTON_CFG.init_state.rot, dtype=torch.float64),
+        ).tolist()
+    )
+    # The Newton hand inherits the shared actuator from ``SHADOW_HAND_NEWTON_CFG``,
+    # overriding only the finger gains: the catch needs a higher nominal gain
+    # (20.0/2.0) than reorient's base. Per-hand prim path and init pose also differ here.
+    newton_cfg = SHADOW_HAND_NEWTON_CFG.replace(
```
StafaH: newton_mjwarp_cfg, since kamino is unknown right now

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -229,7 +244,8 @@ class ObjectCfg(PresetCfg):
         ),
         init_state=RigidObjectCfg.InitialStateCfg(pos=(0.0, -0.39, 0.54), rot=(0.0, 0.0, 0.0, 1.0)),
     )
-    default = physx
+    ovphysx = physx  # OvPhysX is PhysX-based; reuse the PhysX-tuned rigid sphere
```
StafaH: Comment not really necessary

StafaH (approved):

source/isaaclab/isaaclab/envs/utils/marl.py
hujc7: Done in `f142a22`, as sketched — stored in `reset()`/`step()`, `obs_buf` is now a plain property. Also removes the repeated conversion greptile flagged.

source/isaaclab/isaaclab/envs/utils/marl.py
hujc7: Changed in `6ec4220`. Rebinding left anything holding a reference to the wrapped buffer stale; the in-place write keeps them in sync. The test now covers values and buffer identity.

source/isaaclab/test/envs/test_marl_utils.py
hujc7: That test would have passed without the buffer ever updating. `f142a22` adds `step()` to the fake env and asserts the buffer matches both the reset and step returns, plus a state-as-observation variant.

source/isaaclab/test/envs/test_marl_utils.py
hujc7: There wasn't one. `41b7ed5` adds kit-free `test_rsl_rl_wrapper_observations.py`: the wrapper returns the env buffer, tracks updates, never calls the private method, and `DirectRLEnv` publishes it in `reset` and `step`. Verified to fail against the previous implementation.

source/isaaclab_rl/isaaclab_rl/rsl_rl/vecenv_wrapper.py
hujc7: Covered by the same file in `41b7ed5` — it asserts `DirectRLEnv` publishes `obs_buf` in both `reset` and `step`, which is the contract the wrapper depends on.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
hujc7: Removed. Both were in `__init__` rather than a hot path, but no-ops on correctly-placed data.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
hujc7: Trimmed — two were stating the same host-sync rationale.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Renamed, both sites.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Removed.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Restored in `4144bf8`. `develop` had `render_interval=decimation` and this PR had hard-coded it to `2` — same value today, but the coupling was lost. It was the only such site.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_common.py
hujc7: `ruff` splits it back under our isort config (`combine-as-imports` isn't enabled), so the current shape is what the formatter produces.

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/__init__.pyi
hujc7: Kept relative to match `isaaclab_contrib/mdp/__init__.pyi` and `isaaclab_experimental/envs/mdp/__init__.pyi`. Changing it would be a repo-wide edit to those too.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py
hujc7: It's the base for #6421, which adds `commands.py`, `events.py`, `observations.py` here and extends `rewards.py`. Introducing it now avoids inlining and moving it a PR later.

source/isaaclab_assets/isaaclab_assets/robots/shadow_hand.py
hujc7: `SHADOW_HAND_CFG` (PhysX, same file) and `ALLEGRO_HAND_CFG` already use `"fingers"` with the same `robot0_WR.*` inclusion, so renaming only this group breaks parity — a rename belongs across the asset module.

One real consequence though: handover's `.replace(stiffness=20.0, damping=2.0)` swaps the per-joint dict for a scalar, so wrists get the raised gain too. Pre-existing, but the comments said "only the finger gains" — corrected.

source/isaaclab_experimental/isaaclab_experimental/envs/direct_rl_env_warp.py
hujc7: Not new — `develop` already clones into the returned dict in both `reset` and `step`. This only hoists that expression into `obs_buf`, so the copy count is unchanged.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
hujc7: Taking the follow-up option — the ID is still registered unchanged; moving it needs a deprecated alias since the quickstart docs reference it.

source/isaaclab_tasks/test/core/test_shadow_hand_camera_presets.py
hujc7: No longer applies — the `rgb_depth` preset was dropped and the camera config matches `develop` again.

source/isaaclab_experimental/isaaclab_experimental/envs/direct_rl_env_warp.py
AntoineRichard: I think this is coming from som rl libraries messing with the data from the simulator inplace

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -258,10 +275,14 @@ class PhysicsCfg(PresetCfg):
             update_data_interval=2,
```
AntoineRichard: Why not the update data interval to 4?

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -258,10 +275,14 @@ class PhysicsCfg(PresetCfg):
...
             ccd_iterations=50,  # bumped from default 35 for multi-finger contact geometry
         ),
-        num_substeps=2,
+        # 4 substeps (vs the reorient task's 2): sustained ball-palm contact against the
+        # distal joints explodes ~0.7% of 8192 envs to NaN at 2 substeps (zero-action
+        # probe, 300 steps); 4 substeps shows none.
```
AntoineRichard: I think this comment could be trimmed

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -124,61 +131,70 @@ class EventCfg:
...
         init_state=ArticulationCfg.InitialStateCfg(pos=init_pos, rot=init_rot, joint_pos={".*": 0.0})
     )
-    newton_cfg = _SHADOW_HAND_NEWTON_CFG.replace(
+    # Newton's importer bakes the asset's native root orientation into the
+    # root joint (see the note on SHADOW_HAND_NEWTON_CFG.init_state), so the
+    # task rotation must compose with that base rotation rather than replace
+    # it — replacing left both palms heading 90 degrees off and the object
+    # never rested in the right hand.
+    # Composed in float64 via the shared (x, y, z, w) quaternion product,
+    # matching the previously used wp.quatd math bit-for-bit.
```
AntoineRichard: I think this comment could be trimmed

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -124,61 +131,70 @@ class EventCfg:
...
+    newton_rot = tuple(
+        math_utils.quat_mul(
+            torch.tensor(init_rot, dtype=torch.float64),
+            torch.tensor(SHADOW_HAND_NEWTON_CFG.init_state.rot, dtype=torch.float64),
+        ).tolist()
+    )
+    # The Newton hand inherits the shared actuator from ``SHADOW_HAND_NEWTON_CFG``, raising its
+    # gains: the catch needs a higher nominal gain (20.0/2.0) than reorient's base. Per-hand prim
+    # path and init pose also differ here.
```
AntoineRichard: comment can be trimmed

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -124,61 +131,70 @@ class EventCfg:
...
-      ``stiffness=10.0`` / ``damping=0.1``. The Newton USD bakes
-      ``stiffness=286 / damping=57`` on these joints from the MJCF→USD
-      translation, which fights the ``MjcTendon`` coupling and bounces the
-      ball. ``stiffness=10`` (~1/3 of PhysX's ``limit_stiffness=30``) keeps
-      the joints near-passive while the tendon constraint dominates.
+    The Newton variant reuses the shared actuator defined on ``SHADOW_HAND_NEWTON_CFG``,
+    raising its gains to ``20.0`` / ``2.0`` -- the smallest tested setting at which the
+    handover policy learns the catch (mean reward at iter 200 / 2048 envs ~27 -> ~777 vs
+    reorient's base gains). The scalar replaces the per-joint gain mapping, so this applies
+    to every joint in the group, wrists included.
```
AntoineRichard: This here are traces that could be trimmed / removed.

AntoineRichard (approved):
Still some comment issues, but overall it looks fine

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Trimmed in `4fec9dd` — dropped the tuning trace (`mean reward at iter 200 / 2048 envs ~27 -> ~777`) and kept only why the gains are raised.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Cut from 6 lines to 3 in `4fec9dd`, dropping the migration note (`matching the previously used wp.quatd math bit-for-bit`). The reason stays: Newton's importer bakes the root orientation, so the task rotation composes rather than replaces it — otherwise both palms end up rotated 90 degrees.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Removed entirely in `4fec9dd` — it repeated the docstring 20 lines above.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Cut to two lines in `4fec9dd`. Dropped the probe details (`~0.7% of 8192 envs`, `zero-action probe, 300 steps`), kept the reason for 4 substeps.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Raised to 4 in `4fec9dd`.

Every other task pairs `update_data_interval` with `num_substeps` 1:1 — all eight sites are 2/2 — and this PR raised handover's substeps to 4 without updating the interval, so 4 restores that invariant.

Validated at 1500 iterations on both 2048 and 8192 envs: reward 1341 and 1483, no NaN. The 8192 run is the scale that motivated the substep bump.

PR #6415, opened 2026-07-08
[Task Clean-up][Benchmark] Dexterous Part 9/10: Add success-rate support to the benchmark utilities

hujc7:
## Review Map

- **Exact changes**: a stacked PR's page shows the cumulative diff of its dependency chain; the link pins the commit range that is the PR's own contribution.
- Links pin specific SHAs and can go stale after a branch update — this table on #6324 is refreshed first.

| PR | Status | Depends on | Exact changes |
|---|---|---|---|
| #6411 Part 1/10: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/10: OVPhysX articulation + manager runtime |  | — | merged |
| #6413 Part 3/10: Reorient Direct, torch |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6413/changes/f4895f0f9ee..d29afc75e71) |
| #6414 Part 4/10: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/d29afc75e71..b10a84948f8) |
| #6418 Part 5/10: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/3ea6f7bbf6c..b31b04eaf71) |
| #6421 Part 6/10: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/b31b04eaf71..0cae3d87be4) |
| #6410 Part 7/10: Environment overview docs |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6410/changes/b50ac8906fc..af259c0778d) |
| #6954 Part 8/10: OpenAI Shadow variants → contrib (draft) |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6954/changes/0cae3d87be4..b0d2d8448af) |
| 📌 #6415 Part 9/10: Benchmark success-rate utilities |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6415/changes/b50ac8906fc..c7f2f019d8b) |
| #6582 Part 10/10: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |
[...]

PR #6418, opened 2026-07-08
[Task Clean-up][Manager] Dexterous Part 5/10: Add the reorientation manager counterparts

hujc7:
## Review Map

|---|---|---|---|
| #6411 Part 1/10: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/10: OVPhysX articulation + manager runtime |  | — | merged |
| #6413 Part 3/10: Reorient Direct, torch |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6413/changes/f4895f0f9ee..d29afc75e71) |
| #6414 Part 4/10: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/d29afc75e71..b10a84948f8) |
| 📌 #6418 Part 5/10: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/3ea6f7bbf6c..0271783ea17) |
| #6421 Part 6/10: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/b31b04eaf71..0cae3d87be4) |
| #6954 Part 8/10: OpenAI Shadow variants → contrib (draft) |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6954/changes/0cae3d87be4..b0d2d8448af) |
| #6582 Part 10/10: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |

## Summary

Adds manager-based counterparts for the Shadow cube reorientation task and aligns the existing
Allegro manager task with the Direct contracts.

- **3 manager tasks** mirror their Direct counterparts: `Isaac-Reorient-Cube-Shadow`,
  `-Shadow-Camera` and `-Allegro`.
- The OpenAI variants move to the contributed tasks **unchanged**. They encode a paper-specific
  training regime, so they get no manager counterpart.
- `Metrics/success_rate` now means the same thing in both workflows.
- The in-hand cube is one object per hand instead of one per physics backend.

## 1. Manager tasks match the Direct contracts

Observations, actions, rewards, terminations, reset distributions and timing were brought to the
Direct values. A value-parity test asserts the decimation / episode length / simulation step
triple, the orientation success tolerance and the fall distance, so drift on either side fails CI.

The Allegro observation space changes size, so existing manager checkpoints must be retrained.
Domain randomization becomes opt-in through `presets=randomized`, replacing a configuration-file
flag that the command line could not override.

## 2. OpenAI variants move to the contributed tasks

`Isaac-Reorient-Cube-Shadow-OpenAI-{FF,LSTM}-Direct` become
`IsaacContrib-Reorient-Cube-Shadow-OpenAI-{FF,LSTM}-Direct`. Both identifiers are kept: they share
one environment configuration and differ only in their agent configuration, so collapsing them
would leave the recurrent policy unreachable — the training entry point selects an agent
configuration by workflow, not by name.

## 3. One success rate, defined once

`Metrics/success_rate` named two quantities. The manager tasks reported goals reached over goals
presented; the Direct tasks reported a per-episode success bit drawn at `success_count_threshold`.
The curves were not comparable despite sharing a name and measuring the same task.

Both now report the per-attempt rate, and `success_count_threshold` is removed as unused. The goal
accounting moves into a `SuccessTracker` beside `EpisodeErrorRecorder`, which also suppresses
success detection for the single step after an automatic reset — where the goal has been drawn but
no physics has run against it, so a reset pose inside the success threshold would otherwise bank a
goal the policy never earned.

## 4. One in-hand cube per hand, not per backend

The cube was a `PresetCfg` with a variant per physics backend, so each hand manipulated a different
object depending on the backend. On Shadow the branches disagreed on size and spawn height: PhysX
spawned a 60 mm rigid body at `z=0.6`, Newton a 54 mm **articulation with no joints** at `z=0.535`.
A policy trained on one backend was not solving the same task as one trained on the other.

Each hand now declares a single `CUBE_CFG`. Shadow keeps the PhysX values; Allegro keeps its 1.2
scale. The `mass_props` density overrides are dropped: the asset authors an absolute `physics:mass`
of 0.216 kg, which takes precedence over density, so they never applied on either backend.

## 5. Configuration hierarchy flattened, classes named by role

Every manager task overrode all seven of the shared base's sub-configurations, so the base carried
pre-alignment defaults that each task then undid. The environment configurations now derive from
`ManagerBasedRLEnvCfg` directly. Removed: `ReorientObjectEnvCfg`, the shared sub-configurations no
task constructed, and `reorient_common`. `isaaclab_tasks.core.utils` moves to
`isaaclab_tasks.core.reorient.utils`, which the hand-over task already imports from.

One naming convention now applies throughout, since `SceneCfg`, `EventCfg` and
`ObservationsCfg` are each defined in dozens of files and gave no clue which base a task
was specializing. Classes shared across files carry the task prefix; the `Base` suffix
marks only those a subclass must complete by supplying fields the parent leaves unset, so
the observation halves are named as parts rather than templates. Classes nested inside
another configuration drop the prefix. Registered and preset-selectable configurations
always name their task and hold fields only, with their helpers declared beside them.

## Validation

Eight arms, 1500 iterations, 8192 environments, `physics=newton_mjwarp`, one L40 each. All exited
cleanly with no regression in reward.

| Arm | reward before | reward after |
|---|---|---|
| `Isaac-Reorient-Cube-Shadow` | 67.08 | 66.24 |
| `-Shadow` `presets=asymmetric` | 42.08 | 52.95 |
| `-Shadow-Direct` | 3532.95 | 3967.58 |
| `-Allegro` | 25.22 | 34.03 |
| `-Allegro` `presets=randomized` | 16.03 | 39.59 |
| `-Allegro-Direct` | 656.78 | 1003.11 |
| `IsaacContrib-…-OpenAI-FF-Direct` | 2757.56 | 4533.21 |
| `IsaacContrib-…-OpenAI-LSTM-Direct` | — | 2585.54 |

The LSTM arm resolves `ShadowHandAsymLSTMPPORunnerCfg`, confirming the recurrent policy is reachable
through its own identifier.

### Known failure

`test_rendering_shadow_hand[newton-isaacsim_rtx-rgb]` fails; it passes on `develop`. Only the `rgb`
modality is affected — depth, normals and all three segmentations pass on the same backend, so the
cube move is not an obvious explanation. Root cause is not yet established.


source/isaaclab_tasks/changelog.d/task-cleanup-dex-part08.major.rst
```diff
@@ -0,0 +1,31 @@
...
+  common MDP terms.
+* Added opt-in domain randomization to the manager-based Allegro environment
+  (``enable_domain_randomization``, disabled by default; enabling requires
+  retraining).
+* Added a Newton physics preset to the manager-based Allegro environment.
+
+Removed
+^^^^^^^
+
+* Removed the legacy manager-based reorientation configuration
```
AntoineRichard: **[AI Review] — Important**

These symbols are public today (`success_bonus`, `track_*`, and the termination terms are exported from `mdp.__all__`), and `ReorientObjectEnvCfg` is an importable configuration. Removing them in this PR violates the repository rule that breaking changes require a prior deprecation release. The existing Allegro task ID also changes contract in place, making released checkpoints incompatible.

Please retain the old configuration/task behavior and exported terms with deprecation warnings for at least one release, introduce the replacement under new names/IDs, and record this under `Deprecated` rather than `Removed`.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/__init__.pyi
```diff
@@ -4,19 +4,55 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 __all__ = [
+    "NoisyEMAJointPositionToLimitsAction",
+    "NoisyEMAJointPositionToLimitsActionCfg",
+    "ShadowHandCameraFeatures",
```
AntoineRichard: **[AI Review] — Important**

This export is broken: `ShadowHandCameraFeatures`, `shadow_hand_camera_cached_features`, and `shadow_hand_goal_keypoints` are imported from `observations.py`, but that module defines none of them. A runtime access to `mdp.ShadowHandCameraFeatures` raises `AttributeError`, and `from ...mdp import *` consequently fails.

Please remove these three stale exports (and the orphaned camera-section header in `observations.py`) until the implementation lands, or add the missing implementation. A small test that resolves every explicit symbol in this stub would catch this regression.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_direct_env_cfg.py
```diff
@@ -113,59 +32,30 @@ class AllegroHandEnvCfg(DirectRLEnvCfg):
     state_space = 0
     asymmetric_obs = False
     obs_type = "full"
-    # simulation
+
+    # simulation — values mirrored by the manager cfg (guarded by the value-parity test)
```
AntoineRichard: **[AI Review] — Moderate**

There is no value-parity test in this PR or the existing Reorient tests, so this comment makes a false maintenance guarantee. The simulation, reset, reward, and observation contracts are duplicated across the Direct and manager configurations; future edits can silently break the parity promised by this PR.

Please centralize these tunables in a private shared parameter bundle and add an explicit Direct/manager contract test. Once enforced, keep only a concise functional comment.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_env_cfg.py
```diff
@@ -121,6 +121,18 @@ class ShadowHandCameraEnvCfg(ShadowHandEnvCfg):
     observation_space = 164 + 27  # state observation + vision CNN embedding
     state_space = 187 + 27  # asymmetric states + vision CNN embedding
 
+    def __post_init__(self):
+        # The vision env renders through the Isaac RTX tiled camera, whose render
```
AntoineRichard: **[AI Review] — Moderate**

This comment narrates the backend investigation, failure mode, CLI overrides, and benchmark history merely to justify one default. That is review/validation context rather than a durable source-code contract.

Please reduce it to the functional constraint, for example: `# Isaac RTX requires Fabric cloning, so the camera task defaults to PhysX.` Keep the benchmark and alternative-command details in the PR description or user documentation.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/events.py
```diff
@@ -0,0 +1,131 @@
...
+        device=default_position.device,
+    )
+    position_fraction = 0.5 * (position_sample + 1.0)
+    position_delta = limits[..., 0] - default_position
+    position_delta = position_delta + (limits[..., 1] - limits[..., 0]) * position_fraction
+    joint_position = default_position + noise_scale * position_delta
+    return torch.clamp(joint_position, min=limits[..., 0], max=limits[..., 1])
+
+
+def random_xy_rotation(count: int, device: str | torch.device) -> torch.Tensor:
```
AntoineRichard: **[AI Review] — Moderate**

`random_xy_rotation` and `randomize_rotation` below implement the same X/Y quaternion composition. The Direct path uses one and the manager reset path uses the other, creating two sources of truth for quaternion order; `randomize_rotation` also lacks the specific type annotations required for new interfaces.

Please extract one typed composition helper and reuse it from both sampling paths, with a known-input regression test.

AntoineRichard (changes requested):
**[AI Review]**

Requesting changes for two merge blockers: this removes public APIs without the required prior deprecation, and the lazy-export stub advertises camera symbols that do not exist. Inline comments also cover the unenforced Direct/manager parity, source-comment hygiene, and duplicated quaternion construction.

Verification: `./isaaclab.sh -f` passed; the added math tests passed (8 tests); Reorient config-loading tests passed (18 tests); a runtime probe of `mdp.ShadowHandCameraFeatures` failed with the expected `AttributeError`.

hujc7:
@greptileai review

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,26 +3,295 @@
...
-from isaaclab_assets import ALLEGRO_HAND_CFG  # isort: skip
+from isaaclab_assets.robots.allegro import ALLEGRO_ACTUATED_JOINT_NAMES, ALLEGRO_FINGERTIP_BODY_NAMES
 
 
 @configclass
-class AllegroCubeEnvCfg(ReorientObjectEnvCfg):
-    def __post_init__(self):
-        # post init of parent
-        super().__post_init__()
+class AllegroCubeSceneCfg(InteractiveSceneCfg):
```
StafaH: Why didi we remove the manager base class?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
```diff
@@ -17,6 +17,40 @@
...
+    kwargs={
+        "env_cfg_entry_point": f"{__name__}.shadow_hand_manager_env_cfg:ShadowHandManagerEnvCfg",
+        "rl_games_cfg_entry_point": f"{agents.__name__}:rl_games_ppo_cfg.yaml",
+        "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:ShadowHandPPORunnerCfg",
+        "skrl_cfg_entry_point": f"{agents.__name__}:skrl_ppo_cfg.yaml",
+    },
+)
+
+gym.register(
+    id="Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct",
```
StafaH: Move to contrib

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,241 @@
...
+    joint_pos = mdp.EMAJointPositionToLimitsActionCfg(
+        asset_name="robot",
+        joint_names=SHADOW_ACTUATED_JOINT_NAMES,
+        alpha=1.0,
+        rescale_to_limits=True,
+    )
+
+
+@configclass
+class FullStateWithoutActionCfg(ObsGroup):
```
StafaH: Odd naming, diverges from rest of isaaclab

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,241 @@
...
+_RESET_PARAMS = {
+    "position_noise": 0.01,
+    "joint_position_noise": 0.2,
+    "joint_velocity_noise": 0.0,
+    "action_name": "joint_pos",
+}
+
+
+@configclass
+class ShadowPhysxEventCfg(PhysxEventCfg):
```
StafaH: I don't see a difference between this and the newton one, why do we need preset cfg?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
```diff
@@ -0,0 +1,231 @@
...
+    FullStateWithoutActionCfg,
+    ShadowHandManagerSceneCfg,
+)
+from isaaclab_tasks.utils import PresetCfg
+
+from isaaclab_assets.robots.shadow_hand import SHADOW_ACTUATED_JOINT_NAMES, SHADOW_FINGERTIP_BODY_NAMES
+
+
+@configclass
+class OpenAICommandsCfg:
```
StafaH: Im not a fan of the openai naming scheme. The MDP configs should keep very generic names like the rest of isaaclab.

For openai, keep a comment at the top, link to the paper and give them a reference and acknowledgement. We don't need to use the name everywhere.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
```diff
@@ -0,0 +1,231 @@
...
+            },
+        )
+
+        def __post_init__(self):
+            self.enable_corruption = False
+            self.concatenate_terms = True
+
+    @configclass
+    class CriticCfg(FullStateWithoutActionCfg):
+        # -- contact sensing
```
StafaH: Strip away these stray comments. The code itself is readable and doesnt need a commen title.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/commands.py
```diff
@@ -97,48 +105,39 @@ def __str__(self) -> str:
...
-        self.metrics["success_rate"][env_ids] = completed / (completed + 1.0)
         extras = super().reset(env_ids)
-        # super().reset() invoked _resample_command for the new initial goal, which
-        # incremented _completed_attempts; zero it back out so the new episode starts clean.
-        self._completed_attempts[env_ids] = 0.0
-        # Route success_rate to the unified ``Metrics/success_rate`` path (shared TensorBoard
-        # card across tasks); pop it from the returned dict so CommandManager does not
-        # additionally log it under ``Metrics/<term_name>/success_rate``.
-        self._env.extras.setdefault("log", {})["Metrics/success_rate"] = extras.pop("success_rate")
+        reset_buf = getattr(self._env, "reset_buf", None)
```
StafaH: We shouldn't use getattr inside the MDP like this. Either this needs to exist (so we can write a test), or it doesn't and we need to modify code elsewhere to ensure there is a reset_buf? Is it intended for it to be optional like this? When would it be None?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/commands.py
```diff
@@ -208,8 +215,17 @@ class ReorientCommandCfg(CommandTermCfg):
...
+    """Minimum number of goals reached in an episode to count it as a successful episode."""
+
+    fixed_marker_pos: tuple[float, float, float] | None = None
+    """Fixed goal-marker position [m] in each environment, or ``None`` to follow the goal."""
+
     orientation_success_threshold: float = MISSING
-    """Threshold for the orientation error to consider the goal orientation to be reached."""
+    """Threshold [rad] for the orientation error to consider the goal orientation to be reached.
+
+    Set per family at the declaration site, matching the Direct configuration's value.
```
StafaH: Stray coment

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/terminations.py
```diff
@@ -85,3 +88,58 @@ def object_away_from_robot(
     dist = torch.linalg.norm(robot.data.root_pos_w.torch - object.data.root_pos_w.torch, dim=1)
 
     return dist > threshold
+
+
+class reorient_timeout(ManagerTermBase):
```
StafaH: Some of these MDP functions/terms look very generic. Is there nothing in the global MDP for this? Maybe we need to promote this to the global MDP so multiple tasks can take advantage?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/reorient_manager_env_cfg.py
```diff
@@ -1,348 +0,0 @@
-# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: I think this base class was good (and matched the convenetion across all isaaclab)

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
hujc7: The base class was mostly used for allegro task and inheriting into the shadow task causes a lot of overrides, so I decided to make the base class for allegro and shadow but not shared.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
hujc7: updated

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
hujc7: renamed.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
hujc7: removed

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/commands.py
hujc7: it was not available at init and only get assigned during step. Created a reset buffer at init time to avoid it.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/reorient_manager_env_cfg.py
hujc7: It's kind of combined into the allegro one.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/terminations.py
hujc7: With the SR setup, I think some mdps are quite specific. I don't have a clear picture yet how to make it cleaner.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,290 @@
...
+    """Shared reorientation scene with the Allegro hand and a ground plane."""
+
+    # ``clone_in_fabric`` is the only backend-varying field: PhysX/OvPhysX use Fabric
+    # cloning for speed, Newton does not support it.
+    clone_in_fabric = preset(default=False, physx=True, ovphysx=True, newton_mjwarp=False)
```
ooctipus: do you need this

source/isaaclab_tasks/isaaclab_tasks/contrib/reorient/config/shadow_hand/__init__.py
```diff
@@ -0,0 +1,41 @@
...
+    disable_env_checker=True,
+    kwargs={
+        "env_cfg_entry_point": f"{__name__}.shadow_hand_camera_benchmark_env_cfg:ShadowHandCameraBenchmarkEnvCfg",
+        "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:ShadowHandCameraFFPPORunnerCfg",
+        "rl_games_cfg_entry_point": f"{agents.__name__}:rl_games_ppo_camera_cfg.yaml",
+    },
+)
+
+# Retain the released ID as a deprecated alias by inserting its spec directly.
+registry.update(
```
ooctipus: just remove this

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_common.py
```diff
@@ -24,9 +24,11 @@
 
 from isaaclab_assets.robots.allegro import ALLEGRO_HAND_CFG
 
+ALLEGRO_HAND_ROBOT_CFG = ALLEGRO_HAND_CFG.replace(prim_path="/World/envs/env_.*/Robot")
+
 
 @configclass
-class ObjectCfg(PresetCfg):
+class CubeCfg(PresetCfg):
```
ooctipus: collapse this

docs/source/setup/quickstart_details.rst
```diff
@@ -57,7 +57,7 @@ options (observation modes, camera configs, etc.). They fold into Hydra override
 
          # OVRTX rendering (kit-less, no Kit visualizer)
          uv run isaaclab train --rl_library rsl_rl \
-           --task=Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct \
+           --task=IsaacContrib-Reorient-Cube-Shadow-Camera-Benchmark-Direct \
```
kellyguo11: let's make sure this change is also reflected in the docs.

@fatimaanes @yts-nv @r-schmitt for viz on renaming of the task

docs/source/setup/quickstart_details.rst
```diff
@@ -62,7 +62,7 @@ options (observation modes, camera configs, etc.). They fold into Hydra override
```
StafaH: Quickstart no longer exists

source/isaaclab/isaaclab/envs/manager_based_rl_env.py
```diff
@@ -73,8 +73,11 @@ def __init__(self, cfg: ManagerBasedRLEnvCfg, render_mode: str | None = None, **
         # -- counter for curriculum
         self.common_step_counter = 0
 
-        # initialize the episode length buffer BEFORE loading the managers to use it in mdp functions.
+        # initialize the episode length and reset buffers BEFORE loading the managers to use them in
```
StafaH: Here and elsewhere, let's strip comments down to to be simple

source/isaaclab_visualizers/test/visualizer_integration_utils.py
```diff
@@ -1524,12 +1525,17 @@ def _make_shadow_hand_env(
...
-    env_cfg.sim.physics = getattr(env_cfg.sim.physics, preset_key)
-    env_cfg.scene = getattr(env_cfg.scene, preset_key)
-    env_cfg.robot_cfg = getattr(env_cfg.robot_cfg, preset_key)
-    env_cfg.object_cfg = getattr(env_cfg.object_cfg, preset_key)
+
+    def resolve(cfg):
+        """Pick the backend variant, leaving fields that do not vary by backend alone."""
+        return getattr(cfg, preset_key) if isinstance(cfg, PresetCfg) else cfg
+
+    env_cfg.sim.physics = resolve(env_cfg.sim.physics)
```
StafaH: Don't we have a utility for this? Seems fishy to do it by hand

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
```diff
@@ -97,22 +101,139 @@ def cube_keypoints_from_quat(
...
+    reset_action = getattr(env, "_reorient_reset_action", None)
+    reset_step = getattr(env, "_reorient_reset_step", None)
+    common_step_counter = getattr(env, "common_step_counter", None)
+    if reset_action is None or reset_step is None or common_step_counter is None:
+        return raw_action
+    return torch.where((reset_step == common_step_counter).unsqueeze(-1), reset_action, raw_action)
+
+
+# -- composed observation groups
+class openai_policy_observation(ManagerTermBase):
```
StafaH: openai naming still leaking through

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,219 @@
...
+    SHADOW_HAND_ROBOT_CFG,
+    CubeCfg,
+    PhysicsCfg,
+    ShadowHandManagerEventCfg,
+)
+from isaaclab_tasks.utils import PresetCfg
+
+from isaaclab_assets.robots.shadow_hand import SHADOW_ACTUATED_JOINT_NAMES, SHADOW_FINGERTIP_BODY_NAMES
+
+# ---------------------------------- state task ----------------------------------
```
StafaH: Lets strip the comments

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
```diff
@@ -0,0 +1,216 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: Keep only openai?

StafaH (approved):

docs/source/setup/quickstart_details.rst
hujc7: rebased

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
hujc7: Will defer to a followup PR to remove the base variant.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
hujc7: this is open ai specific observation

source/isaaclab_visualizers/test/visualizer_integration_utils.py
hujc7: Yes. updated

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/terminations.py
ooctipus: if it trains without it, remove it, time out could potentially work

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,286 @@
...
+    env_spacing = 0.75
+
+    robot: ArticulationCfg = ALLEGRO_HAND_ROBOT_CFG
+    object: RigidObjectCfg = CUBE_CFG
+    ground = AssetBaseCfg(prim_path="/World/ground", spawn=sim_utils.GroundPlaneCfg())
+    light = AssetBaseCfg(
+        prim_path="/World/Light",
+        spawn=sim_utils.DomeLightCfg(intensity=2000.0, color=(0.75, 0.75, 0.75)),
+    )
+    dome_light = None
```
ooctipus: why dome_light = None? maybe just remove?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,286 @@
...
+            func=mdp.joint_pos_limit_normalized,
+            params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
+        )
+        joint_vel = ObsTerm(
+            func=mdp.joint_vel,
+            scale=0.2,
+            params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
+        )
+        # -- object
+        object_pos = ObsTerm(func=mdp.root_pos_w, params={"asset_cfg": SceneEntityCfg("object")})
```
ooctipus: isn't this pos quat lin vel and ang vel just root state?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,286 @@
...
+    class PolicyCfg(ObsGroup):
+        # -- robot
+        joint_pos = ObsTerm(
+            func=mdp.joint_pos_limit_normalized,
+            params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
+        )
+        joint_vel = ObsTerm(
+            func=mdp.joint_vel,
+            scale=0.2,
+            params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
```
ooctipus: , joint_names=".*", preserve_order=False

this is default you don't need it

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,286 @@
...
+            params={"asset_cfg": SceneEntityCfg("object")},
+        )
+        # -- command
+        goal_pose = ObsTerm(func=mdp.generated_commands, params={"command_name": "object_pose"})
+        goal_quat_diff = ObsTerm(
+            func=mdp.goal_quat_diff,
+            params={"asset_cfg": SceneEntityCfg("object"), "command_name": "object_pose", "make_quat_unique": False},
+        )
+        # -- robot fingertips
+        fingertip_pos = ObsTerm(
```
ooctipus: isn't this just finger tip state?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
```diff
@@ -3,30 +3,286 @@
...
+    Off by default so the task matches its Direct counterpart, which randomizes nothing beyond
+    the reset distributions. ``__post_init__`` reads it while building the configuration, before
+    Hydra applies command-line overrides, so ``env.enable_domain_randomization=true`` has no
+    effect -- set it on the configuration. Changing it requires retraining.
+    """
+
+    def __post_init__(self):
+        # visualizer camera settings
+        self.sim.default_visualizer_cfg = VisualizerCfg(eye=(2.0, 2.0, 2.0))
+        if not self.enable_domain_randomization:
```
ooctipus: lets remove this and always enable randomziation, because this won't be able to be hydro overridable

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,217 @@
...
+        func=mdp.joint_pos_limit_normalized,
+        params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
+    )
+    joint_vel = ObsTerm(
+        func=mdp.joint_vel,
+        scale=0.2,
+        params={"asset_cfg": SceneEntityCfg("robot", joint_names=".*", preserve_order=False)},
+    )
+    # -- object
+    object_pos = ObsTerm(func=mdp.root_pos_w, params={"asset_cfg": SceneEntityCfg("object")})
```
ooctipus: again this is just object state

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,217 @@
...
+        params={"asset_cfg": SceneEntityCfg("object")},
+    )
+    # -- command
+    goal_pose = ObsTerm(func=mdp.generated_commands, params={"command_name": "object_pose"})
+    goal_quat_diff = ObsTerm(
+        func=mdp.goal_quat_diff,
+        params={"asset_cfg": SceneEntityCfg("object"), "command_name": "object_pose", "make_quat_unique": False},
+    )
+    # -- robot fingertips
+    fingertip_pos = ObsTerm(
```
ooctipus: this is just finger tip state

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
```diff
@@ -0,0 +1,217 @@
...
+        physics=PhysicsCfg(),
+    )
+    observations: ObservationsCfg = ObservationsCfg()
+    actions: ActionsCfg = ActionsCfg()
+    commands: CommandsCfg = CommandsCfg()
+    rewards: RewardsCfg = RewardsCfg()
+    terminations: TerminationsCfg = TerminationsCfg()
+    events: ShadowHandManagerEventCfg = ShadowHandManagerEventCfg()
+
+    enable_domain_randomization: bool = False
```
ooctipus: remove this always enable randomizaiton

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
```diff
@@ -0,0 +1,217 @@
...
+    )
+
+
+@configclass
+class ObservationsCfg:
+    """OpenAI 42-dimensional actor and 187-dimensional critic observations."""
+
+    @configclass
+    class PolicyCfg(ObsGroup):
+        openai = ObsTerm(
```
ooctipus: what is opain policy observation? this name is not descriptive?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
```diff
@@ -0,0 +1,217 @@
...
+        func=mdp.object_away_from_goal,
+        params={
+            "threshold": 0.24,
+            "command_name": "object_pose",
+            "object_cfg": SceneEntityCfg("object"),
+        },
+    )
+    max_consecutive_success = DoneTerm(
+        func=mdp.max_consecutive_success,
+        time_out=True,
```
ooctipus: this is not time out

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
```diff
@@ -97,22 +101,136 @@ def cube_keypoints_from_quat(
...
+    """Flattened fingertip spatial velocities [m/s, rad/s], shape ``(num_envs, num_fingertips * 6)``."""
+    asset = env.scene[asset_cfg.name]
+    return asset.data.body_vel_w.torch[:, asset_cfg.body_ids].reshape(env.num_envs, -1)
 
-    # compute quaternion difference
-    quat = math_utils.quat_mul(asset_quat_w, math_utils.quat_conjugate(goal_quat_w))
-    # make sure the quaternion real-part is always positive
-    return math_utils.quat_unique(quat) if make_quat_unique else quat
+
+class fingertip_wrench(ManagerTermBase):
```
ooctipus: are you sure there is not shared wrench mdp?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
```diff
@@ -97,22 +101,136 @@ def cube_keypoints_from_quat(
...
+    reset_action = getattr(env, "_reorient_reset_action", None)
+    reset_step = getattr(env, "_reorient_reset_step", None)
+    common_step_counter = getattr(env, "common_step_counter", None)
+    if reset_action is None or reset_step is None or common_step_counter is None:
+        return raw_action
+    return torch.where((reset_step == common_step_counter).unsqueeze(-1), reset_action, raw_action)
+
+
+# -- composed observation groups
+class openai_policy_observation(ManagerTermBase):
```
ooctipus: this is not descriptive abotu what this is.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
```diff
@@ -97,22 +101,136 @@ def cube_keypoints_from_quat(
...
+        torque_data = sensor.data.torque
+        if force_data is None or torque_data is None:
+            return self._zeros
+        force = force_data.torch[:, sensor_cfg.body_ids]
+        torque = torque_data.torch[:, sensor_cfg.body_ids]
+        return torch.cat((force, torque), dim=-1).reshape(env.num_envs, -1)
+
+
+# -- action terms
+def reorient_last_action(env: ManagerBasedRLEnv, action_name: str) -> torch.Tensor:
```
ooctipus: why not just last action?

hujc7: Agreed — this is now the framework `mdp.last_action`. The task-local `reorient_last_action` term and the hidden `env._reorient_*` buffer it read from are both gone.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
hujc7: There is — replaced with the framework `mdp.body_incoming_wrench`. The task-local wrench term is deleted.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
hujc7: Removed. The composite term is gone; the asymmetric actor/critic split is now selected with `presets=asymmetric` on the core task instead of being baked into an observation term.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
hujc7: Deleted. The reduced actor is now four plain framework `ObsTerm`s, and the OpenAI task itself moved to `contrib` — per Kelly's direction, the OpenAI variants are paper-specific and no longer need a manager counterpart in core.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_openai_manager_env_cfg.py
hujc7: Right, it wasn't. `reorient_timeout` is removed; the task uses the framework `mdp.time_out`.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/terminations.py
hujc7: Removed — it does train without it. `max_consecutive_success` is no longer in `mdp/` at all.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
hujc7: Yes. These are now grouped into a dedicated `ObjectObsCfg` group. They stay as four separate terms because the framework exposes `root_pos_w` / `root_quat_w` / `root_lin_vel_w` / `root_ang_vel_w` individually — there is no combined root-state observation term in core `mdp` to collapse them into.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
hujc7: Same as the Allegro side: grouped into `ObjectObsCfg`. Kept as four framework `root_*` terms because core `mdp` has no combined root-state term.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
hujc7: It is, and it's now its own `RobotObsCfg` group. `fingertip_pose` uses the framework `mdp.body_pose_w`. `fingertip_vel` has to stay task-local: core `mdp` has no `body_vel_w` and no combined body-state term, so there is nothing shared to promote it to.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
hujc7: Same as the Allegro side: now in `RobotObsCfg`, with `fingertip_pose` on the framework `mdp.body_pose_w`. `fingertip_vel` stays task-local because core `mdp` has no `body_vel_w` equivalent.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
hujc7: The flag is gone. Randomization is now a preset — `presets=randomized` — which *is* selectable from the CLI, so the non-overridable bool you flagged was exactly the reason for the change.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_manager_env_cfg.py
hujc7: Same on the Shadow side: the bool is removed and randomization is selected with `presets=randomized`, which is CLI-overridable.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/allegro_hand/allegro_hand_manager_env_cfg.py
hujc7: Removed. The shared `SceneCfg` in `reorient_manager_env_cfg.py` owns the ground plane and the dome light now, so a hand config no longer nulls it out.

source/isaaclab_tasks/test/core/test_reorient_value_parity.py
AntoineRichard, line 1: Do we really need this test?

AntoineRichard (approved):

source/isaaclab_tasks/test/core/test_reorient_value_parity.py
hujc7: No. deleted.

PR #6421, opened 2026-07-08
[Task Clean-up][Manager] Dexterous Part 6/10: Add the handover and camera manager counterparts

hujc7:
## Review Map

|---|---|---|---|
| 📌 #6421 Part 6/10: Handover + camera manager counterparts |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/b31b04eaf71..0cae3d87be4) |
| #6582 Part 10/10: Warp variants → experimental (draft; merges last) |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |

## Summary

Completes manager coverage of the dexterous task families by adding manager-based
counterparts for the Shadow handover and Shadow camera reorientation tasks.

- **2 manager tasks** added: `Isaac-Shadow-Handover` and `Isaac-Reorient-Cube-Shadow-Camera`.
- **The last 2 `-Play` registrations** in `isaaclab_tasks` are removed.

## 1. Handover manager counterpart

The fused handover reward became a plain reward term, with success and goal-distance
bookkeeping moved to `HandoverCommand`, which owns the goal. It reports
`Metrics/success_rate` as a per-episode bit, matching the handover Direct environment.

`HandoverCommand.command` builds its pose per call rather than returning a persistent
buffer, so consumers that store it across steps are not aliased to live data.

The task keeps no observation, reset or event code of its own. Every observation is a
framework term or a reorientation one: fingertip pose is `body_pose_w`, the object block
is the root pose, root velocities, the generated command and the reorientation
goal-quaternion difference, and only the fingertip velocity stays task-local because the
framework has no `body_vel_w`. Both hands reset through the reorientation hand reset, one
term per hand. The object and goal block was previously observed once per agent with
identical values, so a single copy now carries the same information.

`presets=randomized` adds the domain-randomization terms to the reset, matching how the
reorientation manager tasks expose theirs.

The reset and action observations use the framework terms rather than task-local ones,
following Part 5: the object resets through `reset_root_state_with_random_orientation`
and both hands observe `last_action`, leaving only the two-hand joint and PD-target
reset task-local. This samples the object's reset orientation uniformly over SO(3)
where the Direct task composes X/Y rotations, and reports zero actions on the reset
step where the Direct task reports the pre-reset action.

## 2. Camera manager counterpart

The camera task runs on PhysX by default: the RTX render modalities require Fabric
cloning, which Newton does not support, so the inherited default could not render.
Newton remains selectable with `physics=newton_mjwarp` for the state-only observation
groups.

The feature-extractor observation term ignored its declared `feature_extractor_cfg`
parameter and read the environment configuration instead; it now honours the parameter.

## 3. Camera playback moves to `play_mode`

`Isaac-Reorient-Cube-Shadow-Camera-Play` and `-Camera-Direct-Play` were the last `-Play`
registrations left in `isaaclab_tasks`; every other task moved to `play_mode` in
https://github.com/isaac-sim/IsaacLab/pull/6676 — Remove the _PLAY configs. Their
playback environment count and inference feature extractor now live in `play_mode`
overrides on the manager and Direct camera configurations.

The override mutates the feature extractor rather than replacing it. Replacing the
object resets every field the caller does not name, which silently re-enabled the CNN
in `ShadowHandCameraBenchmarkEnvCfg` — a configuration whose entire purpose is to
disable it.

## 4. Reorientation action configuration

The action term is named through a module path, so loading a task configuration no
longer imports the USD bindings.

## Validation

Trained at 4096 environments, 1500 iterations, seed 42, on one L40
(`jichuanh-autorun-8gpu`). Success rate is logged per reset batch, so the figures below
are the mean and range over the final eight batches rather than a single value.

| Task | Backend | Success rate | Range |
|---|---|---|---|
| `Isaac-Shadow-Handover` (manager) | `newton_mjwarp` | 0.756 | 0.745 - 0.779 |
| `Isaac-Shadow-Handover-Direct` (reference) | `newton_mjwarp` | 0.697 | 0.563 - 0.875 |
| `Isaac-Shadow-Handover` (manager) | `isaacsim_physx` | 0.999 | 0.992 - 1.000 |

The manager task matches the Direct reference it replaces on the same backend. The
spread between backends is a pre-existing property of the Newton Shadow Hand actuation,
not something this pull request introduces; it is tracked separately.

`Isaac-Reorient-Cube-Shadow-Camera` was smoke-tested rather than trained to convergence:
300 iterations at 256 environments through the Kit renderer completes cleanly. Its
rendering matrix and golden images are covered separately.

`presets=randomized` converges more slowly than the default and does not plateau within
1500 iterations. A longer run is in progress to set an appropriate iteration budget.


source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_manager_env_cfg.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+@configclass
+class ShadowHandCameraManagerSceneCfg(PresetCfg):
+    """Backend-specific camera scene alternatives for training and benchmarking."""
+
+    physx = _ShadowHandCameraManagerSceneCfg(clone_in_fabric=True)
+    newton_mjwarp = _ShadowHandCameraManagerSceneCfg(clone_in_fabric=False)
+    ovphysx = physx
+    default = newton_mjwarp
```
AntoineRichard: [AI Review][Important] **The Manager camera task still defaults to an unsupported backend.** Resolving `ShadowHandCameraManagerEnvCfg()` selects `NewtonCfg` with `clone_in_fabric=False`. The latest commit documents that the default RTX modalities require Fabric cloning, but its PhysX-default override was applied only to the Direct config. The new Manager and Play registrations therefore remain broken by default. Please default the Manager scene plus its sim-physics, robot, and object presets to PhysX, keep explicit `newton_mjwarp` selection available, and add the Manager task to the registered-camera rendering matrix so this configuration is exercised.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_manager_env_cfg.py
```diff
@@ -0,0 +1,200 @@
...
+
+@configclass
+class TerminationsCfg:
+    """Termination conditions matching the Direct single-agent adapter."""
+
+    object_out_of_reach = DoneTerm(
+        func=mdp.root_height_below_minimum,
+        params={"minimum_height": 0.24, "asset_cfg": SceneEntityCfg("object")},
+    )
+    time_out = DoneTerm(func=mdp.time_out, time_out=True)
```
AntoineRichard, line 268: [AI Review][Important] **This timeout is one control step later than the Direct task.** The generic Manager `time_out` returns `episode_length_buf >= max_episode_length`, while `HandoverEnv._get_dones()` uses `>= max_episode_length - 1`. For `max_episode_length == 450`, buffer value 449 terminates Direct but not Manager, changing rollout boundaries and reset observations despite this config promising Direct parity. Please add a Direct-compatible handover timeout term (or share the boundary implementation) and cover 448/449/450 explicitly.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
```diff
@@ -46,43 +81,66 @@
...
+        "env_cfg_entry_point": f"{__name__}.shadow_hand_camera_manager_env_cfg:ShadowHandCameraManagerPlayEnvCfg",
+        "rsl_rl_cfg_entry_point": f"{agents.__name__}.rsl_rl_ppo_cfg:ShadowHandCameraFFPPORunnerCfg",
+        "rl_games_cfg_entry_point": f"{agents.__name__}:rl_games_ppo_camera_cfg.yaml",
+    },
+)
+
 gym.register(
     id="Isaac-Reorient-Cube-Shadow-Camera-Direct",
-    entry_point=f"{__name__}.shadow_hand_camera_env:ShadowHandCameraEnv",
+    entry_point=f"{__name__}.shadow_hand_direct_camera_env:ShadowHandCameraEnv",
```
AntoineRichard, line 52: [AI Review][Important] **The Direct camera module rename needs a deprecation shim.** This PR changes the registered entry point to `shadow_hand_direct_camera_env`, but the previously documented `shadow_hand_camera_env` and `shadow_hand_camera_env_cfg` modules are deleted and no longer importable. That is a breaking public import-path change without the deprecation required by the repository guidelines; `docs/source/overview/environments.rst` also still links the old path. Please retain compatibility modules that warn and re-export the renamed symbols, and add migration guidance to the changelog before removing them in a later release.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_manager_env_cfg.py
```diff
@@ -0,0 +1,200 @@
...
+    actions: ActionsCfg = ActionsCfg()
+    commands: CommandsCfg = CommandsCfg()
+    rewards: RewardsCfg = RewardsCfg()
+    terminations: TerminationsCfg = TerminationsCfg()
+    events: EventCfg = EventCfg()
+
+    def __post_init__(self):
+        self.decimation = 2
+        self.episode_length_s = 7.5
+        # simulation — mirrors the Direct cfg (guarded by the value-parity test)
```
AntoineRichard: [AI Review][Moderate] **This comment claims a safeguard that is not present on this branch.** There is no checked-in Direct-vs-Manager value-parity test for handover; the available dexterous test covers math helpers only. The duplicated task values have already drifted at the timeout boundary. Please either centralize the shared task constants/configuration or land the claimed parity tests in this PR. The source comment itself should describe the functional setting, not the review/test process.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
```diff
@@ -7,36 +7,342 @@
...
+
+    def __init__(self, cfg: ObservationTermCfg, env: ManagerBasedRLEnv):
+        super().__init__(cfg, env)
+        sensor_cfg: SceneEntityCfg = cfg.params["sensor_cfg"]
+        camera: Camera = env.scene.sensors[sensor_cfg.name]
+        # Runtime-only import: the mdp layer must not import the task-config layer
+        # at module load (config modules import mdp; see the layering note above).
+        from isaaclab_tasks.core.reorient.config.shadow_hand.feature_extractor import FeatureExtractor
+
+        feature_extractor_cfg: FeatureExtractorCfg = env.cfg.feature_extractor
```
AntoineRichard: [AI Review][Moderate] **The declared `feature_extractor_cfg` term parameter is silently ignored.** `CameraPolicyCfg` supplies and rewires this parameter, but this constructor reads `env.cfg.feature_extractor` instead and `__call__` deletes the declared argument. Consequently, configuring the observation term directly has no effect, and the later docstring claim that the term owns its copy is inaccurate. Please construct from `cfg.params["feature_extractor_cfg"]`, or remove the parameter and all associated rewiring so there is only one explicit source of truth.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -136,9 +145,9 @@ def _shadow_hand_cfg(
     init_pos: tuple[float, float, float],
     init_rot: tuple[float, float, float, float],
 ) -> PresetCfg:
-    """Per-hand Shadow Hand preset (PhysX and Newton MJWarp variants).
+    """Per-hand Shadow Hand preset (PhysX, Newton MJWarp, and OVPhysX variants).
```
AntoineRichard: [AI Review][Moderate] **Please reduce this to durable functional documentation.** This docstring is largely a tuning diary: comparisons with another backend, specific MAPPO rewards, the search history for gains, and historical asset names. Those details will age quickly and obscure the actual contract. Keep a concise description of what each actuator override controls, its units, and the backend constraint it addresses; move experiment results and rejected values to the PR or a design note. The same standard should be applied to the `num_substeps` benchmark anecdote below.

AntoineRichard (commented):
[AI Review] **Review summary — changes requested before showroom-quality merge.**

I left six inline findings: three Important correctness/compatibility issues and three Moderate duplication/configuration/comment-quality issues. The principal blockers are the unsupported Newton default on the new Manager camera task, the one-step Direct/Manager handover timeout mismatch, and removal of documented camera module paths without a deprecation shim.

Verification performed on head `ba3e66c`: `./isaaclab.sh -f` passed all hooks; focused dexterous-math and MARL-adapter tests passed (24 tests). I also reproduced the Manager camera default as `NewtonCfg` with `clone_in_fabric=False`, the 449/450 timeout boundary mismatch, and failure to resolve both old camera module paths. Full Isaac Sim/GPU training was not run, and the PR currently has no substantive CI result beyond the labeler.

The existing Greptile suggestion to call `.item()` on per-step log tensors should not be applied: zero-dimensional device tensors follow the Manager logging convention, while `.item()` would force a host synchronization on every environment step.

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/rewards.py
```diff
@@ -0,0 +1,87 @@
...
+        object_asset: RigidObject = env.scene[object_cfg.name]
+        object_pos = object_asset.data.root_pos_w.torch - env.scene.env_origins
+        goal_pos = env.command_manager.get_command(command_name)[:, :3]
+        succeeded, goal_distance = evaluate_handover_success(object_pos, goal_pos, success_distance_threshold)
+        self._goal_distance.update(goal_distance)
+        per_agent_reward = handover_reward(goal_distance, distance_scale)
+
+        env.extras.setdefault("log", {})["dist_reward"] = per_agent_reward.mean()
+        env.extras["log"]["dist_goal"] = goal_distance.mean()
+        env.extras["log"]["Metrics/goal_distance"] = goal_distance.mean().item()
```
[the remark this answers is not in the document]
hujc7: Resolved. The reward term no longer writes to the extras log dict at all; the success metric is routed through the command term's metrics and namespaced as `Metrics/success_rate` by `CommandManager.reset()`. No mixed tensor/scalar values remain.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_manager_env_cfg.py
hujc7: Resolved. `ShadowHandCameraManagerEnvCfg.__post_init__` now pins both the physics and robot presets to their PhysX variants, so resolving the cfg selects `PhysxCfg` rather than `NewtonCfg`. Newton stays selectable via `physics=newton_mjwarp`.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_manager_env_cfg.py
hujc7: Keeping the generic term. The difference is one control step out of 450 (~0.2% of an episode), which is below the run-to-run spread of the success-rate metric: the manager task measures 0.756 (0.745-0.779 over the final eight reset batches) against the Direct reference at 0.697 (0.563-0.875), on the same seed and iteration budget. Matching Direct's `max_episode_length - 1` boundary would mean re-implementing a framework term to reproduce an off-by-one that carries no measurable behavioural cost. The docstring documents the difference so it is not a surprise.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
hujc7: Declining, with a note. The registered gym IDs are unchanged -- `Isaac-Reorient-Cube-Shadow-Camera-Direct` and `Isaac-Reorient-Cube-Shadow-Camera` both still resolve -- and those IDs are the public surface for a task. Module paths under `config/<robot>/` are internal layout, so an import shim would create a compatibility promise where none was documented. Happy to add one if you consider the module path public.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_manager_env_cfg.py
hujc7: Resolved. The claim was removed rather than backed by a new test: the values it referred to are now shared between the Direct and manager configurations instead of duplicated, so there is nothing left for a parity test to guard.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/mdp/observations.py
hujc7: Resolved. `ShadowHandCameraFeatures.__init__` now reads `cfg.params["feature_extractor_cfg"]`, so configuring the term through the observation cfg takes effect. `__call__` deletes the argument only after `__init__` has consumed it.

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
hujc7: Resolved. The docstring is reduced to the durable contract; the cross-backend comparisons, MAPPO reward figures, gain-search history, and historical asset names are removed.

StafaH (approved):

AntoineRichard (approved):

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/commands.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def _debug_vis_callback(self, event) -> None:
+        self._goal_visualizer.visualize(
+            translations=self.pos_command_e + self._env.scene.env_origins,
+            orientations=self.quat_command_w,
+        )
+
+
+@configclass
+class HandoverCommandCfg(CommandTermCfg):
```
ooctipus, line 104: Do not put cfg together in implementation

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -33,120 +30,21 @@
...
-        interval_range_s=(36.0, 36.0),  # time_s = num_steps * (decimation * dt)
-        params={
-            "gravity_distribution_params": ([0.0, 0.0, 0.0], [0.0, 0.0, 0.4]),
-            "operation": "add",
-            "distribution": "gaussian",
-        },
-    )
-
-
 def _shadow_hand_cfg(
```
ooctipus, line 33: something is wrong in this cfg, we shouldn't work around the asset and behavior different like this.....

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/commands.py
```diff
@@ -0,0 +1,114 @@
...
+        # card across tasks); pop it from the returned dict so CommandManager does not
+        # additionally log it under ``Metrics/<term_name>/success_rate``.
+        log["Metrics/success_rate"] = extras.pop("success_rate")
+        for statistic, value in self._minimum_goal_distance.reset(env_ids).items():
+            log[f"Diagnostics/episode_min_goal_distance_{statistic}"] = value
+        return extras
+
+    def _resample_command(self, env_ids: Sequence[int]) -> None:
+        random_values = 2.0 * torch.rand((len(env_ids), 2), device=self.device) - 1.0
+        self.quat_command_w[env_ids] = math_utils.quat_mul(
```
ooctipus, line 80: this looks like a very inefficient way to sample in quaternion space, you can just randomly sample r p y, then convert to quaternion instead...

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
ooctipus, line 55: .............. can we fix the issue from root not patch like that

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/commands.py
hujc7: Will be addressed in a followup PR for asset consolidation.

PR #6426, opened 2026-07-08
Honor replicate_physics=False in the environment cloner

mmichelis:
# Description

The queued environment cloner ignored `InteractiveSceneCfg.replicate_physics`: the flag was only consumed by the scene summary print and the prestartup event-manager guards, and never gated replication. As a result the physics-engine replicator always ran, even for presets that set `replicate_physics=False` (for example the `Isaac-Lift-Soft-Franka` task under the PhysX backend), producing spurious `omni.physx.plugin` "Replication of this type is not supported" errors for deformable prims and their materials. This restores the pre-refactor behavior in which `replicate_physics=False` disables physics-engine replication while USD geometry is still cloned to every environment.

Backend context classes now carry an `is_physics_backend` class attribute. `UsdReplicateContext` sets it to `False` so the authoritative geometry clone always runs, and the physics-engine backends (PhysX, OvPhysX, Newton) are treated as physics via the `getattr(..., True)` default, so no changes are needed in those extension packages. `replicate()` and `ReplicateSession` gain a `replicate_physics` parameter (default `True`, so existing callers are unaffected) that skips physics-engine backends when set to `False`; the physics engine then parses each environment individually from the USD-copied schemas. The fix is confined to the core `isaaclab` package.

Fixes # (no linked issue)

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/honor-replicate-physics-flag.rst
+++ b/source/isaaclab/changelog.d/honor-replicate-physics-flag.rst
@@ -1,0 +1,9 @@
+Fixed
+^^^^^
+
+* Fixed :func:`~isaaclab.cloner.replicate` and :class:`~isaaclab.cloner.ReplicateSession`
+  to honor :attr:`~isaaclab.scene.InteractiveSceneCfg.replicate_physics`. When set to
+  ``False``, physics-engine backends are now skipped so only USD geometry is replicated
+  and the physics engine parses each environment individually. This removes spurious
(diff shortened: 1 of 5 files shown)


mmichelis:
Better solution for now is to directly set the cloning context of the deformable: `DeformableCfg.cloning_context=isaaclab.cloner:USDReplicateContext`

Disabling physics cloning for the whole scene can be inefficient.

PR #6432, opened 2026-07-08
Fix ground-plane import for assets without legacy prim layout

### LINKED ISSUE #6326: TerrainImporter plane import assumes legacy collision and shader prim paths
## Environment

- Isaac Lab: `3.0.0b2`
- Isaac Sim assets: `6.0`

## Problem

`TerrainImporter` cannot import `terrain_type="plane"` when a caller configures its documented physics or visual material overrides. The current default ground-plane USD does not contain the legacy prims that the importer assumes are present.

This makes a valid `TerrainImporterCfg` fail during environment creation.

## Minimal reproducer

```python
from isaaclab.terrains import TerrainImporter, TerrainImporterCfg

cfg = TerrainImporterCfg(
    prim_path="/World/ground",
    terrain_type="plane",
    terrain_generator=None,
    physics_material=sim_utils.RigidBodyMaterialCfg(
        static_friction=1.0,
        dynamic_friction=1.0,
    ),
)
TerrainImporter(cfg)
```

## Actual behavior

With `physics_material` configured, `spawn_ground_plane()` searches below the spawned asset for a child prim whose type is `Plane`, then fails with:

```text
ValueError: No collision prim found at path: '/World/ground/terrain'.
```

If the physics override is removed, `TerrainImporter.import_ground_plane()` still always passes a non-null `color` to `GroundPlaneCfg`. `spawn_ground_plane()` then tries to write the legacy shader property below `Looks/theGrid/Shader`, which does not exist in the current asset:

```text
Prim does not exist at path '/World/ground/terrain/Looks/theGrid/Shader'.
```

## Expected behavior

A `TerrainImporterCfg` with `terrain_type="plane"` should support the documented material overrides, or gracefully skip an override that cannot be applied to the spawned ground-plane asset. It should not depend on fixed internal USD paths that are absent from the asset it spawns.

## Suggested direction

Please make `TerrainImporter.import_ground_plane()` / `spawn_ground_plane()` discover the actual collision and material prims in the spawned USD, or update the shipped ground-plane asset and importer together so their contract is consistent. A no-override plane should also pass `color=None` rather than unconditionally targeting the legacy grid shader.

ooctipus:
# Description

`TerrainImporter.import_ground_plane()` / `spawn_ground_plane()` assumed the internal layout of the default grid asset from Isaac Sim: a `Plane`-typed collision prim and the `Looks/theGrid/Shader` material. Any ground-plane USD file with a different layout failed with a `ValueError` during environment creation — even for a default `TerrainImporterCfg`, since `physics_material` defaults to a non-None material.

This PR makes the plane import robust to the asset's layout:

- **Physics material**: bound to all collision-enabled prims (prims with `UsdPhysics.CollisionAPI` applied) instead of requiring a `Plane`-typed child prim. If the asset has no collision prim, the binding is skipped with a warning instead of raising. Note that binding by collision API matches what `bind_physics_material()` accepts anyway — binding to a `Plane` prim without a collision API was already a silent no-op.
- **Color override**: `spawn_ground_plane()` only writes the legacy `Looks/theGrid/Shader.inputs:diffuse_tint` property when that shader exists in the spawned asset, and skips it with a warning otherwise.
- **Color contract**: `TerrainImporter.import_ground_plane()` now passes `color=None` when no visual material is configured (or the material provides no diffuse color), matching the documented semantics of `GroundPlaneCfg.color` ("If None, then the color remains unchanged"). Previously it always forced a black tint.

Behavior notes for reviewers:

- The default grid asset (verified identical on the production and staging asset servers) has `PhysicsCollisionAPI` authored on its `CollisionPlane`, so binding, tint, and scale behavior with the shipped asset is unchanged. Verified end-to-end against the cloud asset.
- The only intentional behavior change: an explicit `visual_material=None` on `TerrainImporterCfg` now keeps the asset's authored color instead of tinting it black. The default config (dark-gray preview surface) is unaffected.

The new unit tests run without Kit and without network access (they build minimal legacy-replica and mesh-collider assets locally). All 5 regression tests fail without the fix; the 3 backward-compatibility tests pass on both sides.

Fixes #6326

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/octi-fix-6326-terrain-plane.rst
+++ b/source/isaaclab/changelog.d/octi-fix-6326-terrain-plane.rst
@@ -1,0 +1,13 @@
+Fixed
+^^^^^
+
+* Fixed :func:`~isaaclab.sim.spawners.from_files.spawn_ground_plane` raising ``ValueError`` for
+  ground-plane USD files that do not follow the layout of the default grid asset from Isaac Sim
+  (issue `#6326 <https://github.com/isaac-sim/IsaacLab/issues/6326>`__). The physics material is now
+  bound to all collision-enabled prims in the spawned asset instead of requiring a ``Plane``-typed
+  prim, and the color override is skipped with a warning when the asset does not contain the grid
+  shader.
+* Fixed :meth:`~isaaclab.terrains.TerrainImporter.import_ground_plane` overriding the ground plane's
+  color with black when :attr:`~isaaclab.terrains.TerrainImporterCfg.visual_material` is set to None.
+  The spawned asset's authored color is now kept unchanged in this case, matching the documented
+  semantics of :attr:`~isaaclab.sim.spawners.from_files.GroundPlaneCfg.color`.
--- a/source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
+++ b/source/isaaclab/isaaclab/sim/spawners/from_files/from_files.py
@@ -24,8 +24,8 @@ from isaaclab.sim.utils import (
     change_prim_property,
     clone,
     create_prim,
+    get_all_matching_child_prims,
     get_current_stage,
-    get_first_matching_child_prim,
     select_usd_variants,
     set_prim_visibility,
 )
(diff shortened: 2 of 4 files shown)


kellyguo11:
fixed in https://github.com/isaac-sim/IsaacLab/pull/7399

PR #6439, opened 2026-07-09
Fix lazy PhysX IMU and PVA acceleration

AntoineRichard:
## Summary

- compute PhysX IMU and PVA finite-difference acceleration from the elapsed time between sensor samples
- keep timestamp inputs stable inside the recorded Warp launches from #6390
- remove mutable scalar inverse-dt launch parameters
- add lazy-read and nonzero-update-period regressions for IMU linear acceleration and PVA linear/angular acceleration

This PR is stacked on #6390. Until #6390 merges, its commits are included in this comparison against develop.

## Root cause

IMU and PVA cached velocity only when their buffers were recomputed, but divided the next velocity delta by the most recent physics-step dt. With four lazy physics updates or an update period of four physics steps, both sensors reported 4x acceleration.

The kernels now derive a per-environment interval from timestamp minus timestamp_last_update.

## Validation

- Unmodified #6390 base: 4 regressions failed with the expected 4x acceleration error
- Fixed targeted regressions: 4 passed
- Complete PhysX IMU suite: 11 passed
- Complete PhysX PVA suite: 11 passed
- Final pre-commit hooks: passed
- Existing recorded-launch assertions remain enabled and pass

## Performance

Benchmarking is deferred until the test machine is connected to AC power. The change removes scalar launch-parameter updates and retains the recorded-launch path; this PR makes no performance claims.


AntoineRichard:
Newton cross-backend validation: using this PR worktree source explicitly on PYTHONPATH, the Newton IMU and PVA ten-step lazy free-fall acceleration tests both pass (2 passed). Newton is not affected by this bug because it consumes native accelerometer/body-acceleration state rather than numerically differentiating sensor samples. No Newton code changes are needed.

ooctipus (approved):

kellyguo11 (approved):

kellyguo11:
there's a failing test case in test_imu_pva_recorded_launch.py

PR #6444, opened 2026-07-09
Add perf-smoke infra probe workflow

Neil4561:
## Summary
- Add a temporary perf-smoke infra probe workflow targeting the same-repo `perf-smoke/infra-probe` branch.
- Validate the self-hosted GPU runner, Docker GPU passthrough, NGC image pull, source overlay + `_isaac_sim` import flow, baseline write auth, and omni-github artifact upload.
- Keep the workflow non-destructive by default; baseline branch writes are dry-run unless manually dispatched with `baseline_write_mode=real`, which only creates and deletes a throwaway `_tmp/*` ref.

## Test plan
- Ran `pre-commit run --files .github/workflows/perf-smoke-infra-probe.yaml`.
- Locally parsed the workflow YAML and simulated the baseline write probe outcomes for accepted, non-fast-forward, denied, and unknown push responses.
- Verified the PR contains a single new workflow file against `perf-smoke/infra-probe`.

### DIFF
--- a/.github/workflows/perf-smoke-infra-probe.yaml
+++ b/.github/workflows/perf-smoke-infra-probe.yaml
@@ -1,0 +1,356 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Perf Smoke - Infra Probe
+#
+# Throwaway workflow that validates the infrastructure the perf-smoke gate depends
+# on, in the real isaac-sim/IsaacLab repo context, BEFORE we open the full gate PR.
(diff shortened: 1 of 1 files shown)


fatimaanes (approved):
Reviewed — verified the external contracts against upstream CI (NGC_API_KEY secret name, [self-hosted, gpu] labels, isaaclab_image_name + latest-develop tag, artifact-name contract incl. job.check_run_id matching upload-omni-github-test-results, action majors). Two notes for follow-up, neither blocking: (1) the probe-5 comment overstates dry-run — it proves repo-level write auth but pre-receive/branch-protection rules on perf-baselines are not exercised by --dry-run (nor by the _tmp ref); worth a one-time branches/perf-baselines/protection API check. (2) workflow_dispatch may 404 while the file exists only on this branch (GitHub indexes dispatchables from the default branch) — if so, retrigger via a push. Merging; the merge itself fires the first probe run.

PR #6445, opened 2026-07-09
Fixes app file selection and visualizer support for load robot benchmark

kellyguo11:
# Description

The load robot benchmark script was previously not launching with the correct app files, thus producing undesired hydra errors. We now use the same app launching mechanism as other standalone scripts, which also enables the visualizer options. 

The script can now run both with viewer and headless

## Type of change

## Checklist


scripts/benchmarks/benchmark_load_robot.py
```diff
@@ -74,7 +73,11 @@
 # Stop the timer for imports
 imports_time_end = time.perf_counter_ns()
 
-# Create the benchmark
+# Create the benchmark. This is constructed at module scope (before the simulation app is launched),
+# matching the other benchmark scripts. The Kit-based frametime recorders provided by
+# ``isaacsim.benchmark.services`` are only available once the app is running, so they are gracefully
+# skipped here; the remaining recorders (CPU/GPU/memory/version) and timing measurements do not
+# require Kit.
 backend_type = args_cli.benchmark_backend
```
hujc7: So we intentionally don't want those? This seems changed previous behavior. Otherwise it needs to be moved after kit startup.

scripts/benchmarks/benchmark_load_robot.py
```diff
@@ -218,15 +224,25 @@ def main():
...
 
 if __name__ == "__main__":
-    # run the main function
-    main()
-    # close sim app
-    simulation_app.close()
+    # Build the scene configuration up front so that launch_simulation() can scan it and
+    # select the appropriate experience file (e.g. auto-enable cameras when camera sensors
+    # are present in the scene).
+    scene_cfg = RobotSceneCfg(num_envs=args_cli.num_envs, env_spacing=2.0)
```
hujc7, line 248: This only aim to benchmark physx right?

scripts/benchmarks/benchmark_load_robot.py
AntoineRichard:
```suggestion
formatter_type = formatter_type
```

scripts/benchmarks/benchmark_load_robot.py
AntoineRichard:
```suggestion
formatter_type = args_cli.formatter_type
```

scripts/benchmarks/benchmark_load_robot.py
AntoineRichard, line 1: This should use formatter_type and not backend_type which is confusing now since we have backends for rendering, physics, etc...

kellyguo11: fixed

scripts/benchmarks/benchmark_load_robot.py
kellyguo11: fixed

scripts/benchmarks/benchmark_load_robot.py
kellyguo11: added newton as well

PR #6452, opened 2026-07-10
Register Newton frame sites with the scene and remove post-build resolution

ooctipus:
# Description

Follow-up to #6423. That PR speeds up `NewtonSiteFrameView` startup by replacing the per-environment
regex scan over `model.body_label` with an exact lookup map — a resolution-side fix. This PR removes
the post-build resolution path entirely by fixing the root cause: camera frame views were the only
site consumer constructed *after* the Newton model build, which forced a second frame-resolution
mechanism that no other sensor needs.

## Frame views construct with the scene (commit 1)

- Every frame-view backend now follows one two-phase lifecycle: construct alongside the scene,
  complete initialization through the backend's own readiness signal (`_initialize_impl` hook on
  `BaseFrameView`). Newton views register frame sites at construction so they clone with the scene
  like IMU/FrameTransformer/PVA sites; PhysX and OVPhysX views resolve prims once physics is ready.
- `Camera` constructs its frame view in `__init__` on every backend.
- `ReplicateSession` publishes the clone plan on `__enter__` (instead of `__exit__`), so entities
  constructed inside the session resolve prims through the plan exactly like post-replication code.
  A failed session unpublishes the plan.
- The Newton post-build resolution path (`_initialize_from_specs`, per-env pattern expansion, and
  the label lookup map from #6423) is deleted — with views constructed before the build there is
  nothing left for it to serve.
- Also fixes: weak-ref physics callbacks leaking registry entries after owner GC; the Newton
  multi-mesh ray caster feeding glob-style patterns into regex site matching.

## Sites must register before the scene builds (commit 2)

- `NewtonManager.cl_register_site` now raises once the scene builder exists. Sites are injected
  into the scene builders and cloned with the scene; a site registered after replication silently
  skipped cloning (`count == 1` instead of `num_envs`) and could recycle another registrant's
  counter-based `ft_N` label, silently corrupting the earlier registrant's resolved bodies. The
  flat injection at simulation start remains only for never-replicated stages; the dead STOP-time
  re-registration is removed from the IMU, frame transformer, and PVA sensors (STOP is only
  dispatched from `close()` on this backend, and `clear()` discards the re-registered state
  immediately after).
- With registration guaranteed to happen before replication, all matching happens in source space:
  frame views register the source body path of each matched clone-plan row and merge the resulting
  per-world sites per environment. This fixes heterogeneously spawned assets
  (`MultiAssetSpawnerCfg`/`MultiUsdFileCfg`), where every variant row previously collapsed onto one
  destination-wide `.*` pattern and produced duplicated or cross-variant frames.
- Sites on bodies outside the cloned environments (a shared table, fixture, ...) now inject into
  the main builder once, like global sites, instead of failing replication.

## Breaking changes

- Constructing a Newton frame view — or registering any Newton site — after the scene replicates
  (or, without replication, after the first reset) raises `RuntimeError`. Construct sensors and
  frame views inside the scene cfg, or before `sim.reset()` on non-replicated stages.

## Review-driven hardening (commit 3)

An adversarial review pass over the branch surfaced and fixed three more holes:

- Per-world sites now carry world membership (`cl_register_site(..., worlds=...)`), so a
  heterogeneous asset's world-anchored frame sites land only in their variant's environments
  instead of every world (previously `count == variants * num_envs` for body-less frame prims
  with variant-differing offsets).
- Non-cloned main-builder bodies are matched only when a pattern matches no source builder — the
  site map entry cannot represent a label owning sites in both spaces, and the main-builder site
  was silently discarded on overlap.
- The direct-workflow camera environments (cartpole camera, shadow hand camera) constructed their
  tiled camera against the empty scene plan, yielding one global site instead of per-env frames;
  they now publish the `from_env_0` plan before constructing the camera.

## Known open items

- Runtime render-only cameras over an already-built scene hit the new contract error on Newton:
  visualizer-owned cameras (`create_visualizer_camera`, called from `initialize_visualizers`
  *after* the physics reset, when tiled camera views are enabled with `tiled_cam_prim_path=None`)
  and the two `ppisp_camera` demos (cameras constructed after `InteractiveScene`). These flows
  spawn or bind cameras post-build by design; they need their own decision (USD-backed views for
  runtime render cameras, or pre-declared cameras) and are tracked as a follow-up. They fail with
  a clear contract error rather than silently misbehaving.
- Pre-existing (unchanged by this PR): the Newton multi-mesh ray caster's tracked-target
  registration cannot rendezvous with `BaseMultiMeshRayCaster` for articulation-root targets or
  multi-owner sources (KeyError at initialization); PhysX handles these. Follow-up.

## Type of change

## Checklist

### DIFF
--- a/scripts/benchmarks/benchmark_view_comparison.py
+++ b/scripts/benchmarks/benchmark_view_comparison.py
@@ -173,13 +173,15 @@ def benchmark_newton(num_iterations: int) -> dict[str, float]:
         prim.GetAttribute("xformOp:translate").Set(Gf.Vec3d(0.1, 0.0, 0.05))
         prim.GetAttribute("xformOp:orient").Set(Gf.Quatd(1.0, 0.0, 0.0, 0.0))
 
+    # Newton frame views are constructed before reset; their sites resolve during the model build.
+    init_start = time.perf_counter()
+    view = NewtonSiteFrameView("/World/envs/env_.*/Cube/Sensor", device=args_cli.device)
+    timing_results["init"] = time.perf_counter() - init_start
+
     sim.reset()
     print(f"  Newton scene setup: {time.perf_counter() - start_time:.4f}s")
 
-    start_time = time.perf_counter()
-    view = NewtonSiteFrameView("/World/envs/env_.*/Cube/Sensor", device=args_cli.device)
     num_prims = view.count
-    timing_results["init"] = time.perf_counter() - start_time
 
     print(f"  Newton FrameView managing {num_prims} prims")
 
--- a/scripts/benchmarks/benchmark_view_comparison.py
+++ b/scripts/benchmarks/benchmark_view_comparison.py
@@ -138,25 +138,23 @@ def benchmark_usd_or_fabric(view_type: str, num_iterations: int) -> dict[str, fl
 @torch.no_grad()
 def benchmark_newton(num_iterations: int) -> dict[str, float]:
     """Benchmark Newton FrameView."""
+    from isaaclab import cloner
     from isaaclab.assets import RigidObjectCfg
-    from isaaclab.scene import InteractiveScene, InteractiveSceneCfg
     from isaaclab.sim import SimulationCfg, build_simulation_context
-    from isaaclab.utils.configclass import configclass
 
     timing_results = {}
+    num_envs = args_cli.num_envs
 
-    @configclass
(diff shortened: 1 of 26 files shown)


ooctipus:
Closing in favor of a fresh PR with a clean base history (this one was originally opened against main and retargeted).

PR #6455, opened 2026-07-10
Fixes pretrained checkpoint downloads broken by the task renames

myurasov-nv:
# Description

The task renames that dropped the `-v0` suffix broke published pre-trained checkpoint downloads: the asset server still stores the old names (`.../Isaac-Velocity-Rough-H1-v0/checkpoint.pt` exists, the new name returns 404), so retrieval returns None and the h1_locomotion demo crashes with `AttributeError: 'NoneType' object has no attribute 'seek'`. Retrieval now retries the legacy `<task>-v0` name (transitional until the checkpoints are re-published), and the demo exits with a clear message when nothing is found. Verified: the demo downloads via the legacy name and runs.

| Before | After |
| ------ | ----- |
| `AttributeError: 'NoneType' object has no attribute 'seek'` at demo startup | checkpoint downloads via the legacy name; demo runs and the H1 walks |

## Reproduction

1. On develop, run:

   ```bash
   ./isaaclab.sh -p scripts/demos/h1_locomotion.py
   ```

2. The demo crashes at startup with `AttributeError: 'NoneType' object has no attribute 'seek'`.
3. Asset-server check: `PretrainedCheckpoints/rsl_rl/Isaac-Velocity-Rough-H1/checkpoint.pt` returns 404, while `.../Isaac-Velocity-Rough-H1-v0/checkpoint.pt` returns 200.

## Type of change

## Checklist


myurasov-nv:
The ARM CI failures are a runner network outage (PyPI unreachable during pip install), unrelated to this change.

source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
```diff
@@ -117,8 +117,15 @@ def get_published_pretrained_checkpoint(workflow: str, task_name: str) -> str |
...
-            return None
+            # published checkpoints may still be stored under the pre-rename task name,
+            # which carried a -v0 suffix; retry with the legacy name before giving up
+            legacy_path = get_published_pretrained_checkpoint_path(workflow, f"{task_name}-v0")
+            print(f"Not found, retrying with the legacy task name : {legacy_path}")
+            try:
+                resume_path = retrieve_file_path(legacy_path, download_dir)
+            except Exception:
+                print("A pre-trained checkpoint is currently unavailable for this task.")
+                return None
```
ooctipus, line 131: Two things on the retry:

1. It suffixes `-v0` unconditionally, so a task whose name already ends in `-v0` produces an `X-v0-v0` request — a guaranteed second failing round-trip on every genuinely missing checkpoint. A guard keeps the retry meaningful.
2. Both `except Exception` arms classify every failure — network down, auth, server error — as "unavailable for this task". If `retrieve_file_path` can't distinguish not-found from transport errors that's pre-existing, but the message can at least stop asserting the checkpoint doesn't exist when the request may simply have failed.

```suggestion
            # published checkpoints may still be stored under the pre-rename task name,
            # which carried a -v0 suffix; retry with the legacy name before giving up
            if task_name.endswith("-v0"):
                print("The pre-trained checkpoint could not be retrieved for this task.")
                return None
            legacy_path = get_published_pretrained_checkpoint_path(workflow, f"{task_name}-v0")
            print(f"Not found, retrying with the legacy task name: {legacy_path}")
            try:
                resume_path = retrieve_file_path(legacy_path, download_dir)
            except Exception:
                print("The pre-trained checkpoint could not be retrieved for this task.")
                return None
```

Bigger-picture (not blocking): the repo's pattern for deprecated names is an explicit alias table — see `preset_target.py`'s `legacy_aliases` (old → canonical, `FutureWarning` on use, stated removal window) — rather than reconstructing old names by suffix arithmetic. If task renames recur, a small `_LEGACY_CHECKPOINT_NAMES` map next to where checkpoint paths are derived would localize the rename knowledge instead of encoding "append `-v0`" as a universal rule.

ooctipus (approved):

source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
myurasov-nv: Applied your suggestion in 675235f: guarded so a name already ending in -v0 returns immediately instead of requesting -v0-v0, and both arms now say the checkpoint could not be retrieved rather than asserting it does not exist.

On the bigger picture, agreed the suffix rule is a heuristic, not a general alias. Kelly noted on the side that some renames changed more than dropping -v0, so this retry only covers the -v0 cases and the durable fix is re-publishing the checkpoints under the current task names. Happy to swap this for an explicit legacy_aliases-style map (old -> canonical) if you'd prefer that land here rather than as a server-side re-publish; wasn't sure which side owns it.

myurasov-nv:
Parking this per discussion: for the GA release the checkpoints will be re-uploaded to Nucleus under the current task names, which is the durable fix. Converting to draft; the client-side legacy-name retry here can be revisited if a stopgap is still wanted before that lands.

StafaH:
Closing this as this should now be resolved. We have new checkpoints and have refined our checkpoint loading with updated semantics for loading presets.

PR #6456, opened 2026-07-10
Drain stale CUDA errors before simulation start

AntoineRichard:
## Description

A latent Warp defect can abort Newton simulation (re)initialization with a spurious `Warp CUDA error 1: invalid argument` at the first buffer copy (`wp.clone` during `PHYSICS_READY`), in processes that run several simulation lifecycles (test suites, notebooks).

Root cause (fully diagnosed; a ~50-line pure-Warp repro exists and a Warp-side fix is in review):
- `wp.capture_while`/`wp.capture_if` pause the parent capture and resume capturing into the conditional body graph on the same stream. If a Python garbage-collection pass runs inside a body callback and drops the last reference to an array allocated in the parent graph, Warp inserts the memory free node into the body graph with parent-graph dependency nodes — which fails with `cudaErrorInvalidValue`, leaks the allocation, and leaves the CUDA per-thread error latched.
- `wp_array_copy_device` ends with `check_cuda(cudaGetLastError())`, so the next innocent Warp copy surfaces and mis-attributes the stale error — typically one test/lifecycle later, which made this look like flakiness.

This PR adds two independent layers in `NewtonManager` until the upstream Warp fix propagates:
1. **Prevention** — garbage collection is paused for the duration of graph capture (both the standard `ScopedCapture` path and the RTX relaxed-capture path), with a collection run immediately after the capture window, where freeing graph-scoped allocations is handled correctly. Reference-count-driven frees are unaffected.
2. **Defense in depth** — `start_simulation` drains (and logs) any latched CUDA error via Warp's own `wp_cuda_context_check` before dispatching initialization callbacks, so no prior lifecycle can poison the next one through any other latched-error path.

Validation: the deterministic 4-test repro (from the articulation-ordering test additions in #6336) goes red→green with the GC guard alone, with zero `failed to add a memory free node` warnings (the corrupting frees no longer happen); the full Newton articulation suite is green with ordering tests (257 passed) and on develop (236 passed). A develop-only regression test is not included because develop's current suite never arranges the triggering sequence; #6336's tests cover it organically once merged.

## Type of change

## Checklist

- [x] I have read and understood the [contribution guidelines]
- [x] I have run the [pre-commit checks] with `./isaaclab.sh --format`

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -1144,6 +1144,30 @@ class NewtonManager(PhysicsManager):
             NewtonManager._fk_reset_mask.fill_(True)
 
     @classmethod
+    def _drain_stale_cuda_error(cls) -> None:
+        """Clear a stale CUDA error latched on the device before (re)initialization.
+
+        Warp 1.15 leaves the per-thread CUDA error uncleared when
+        ``wp_free_device_async`` fails to add a graph memory free node while a
+        capture is still registered (its "capture ended" sibling branch clears
+        the identical error as benign), and the next Warp array copy then
+        surfaces that stale error as its own failure, aborting simulation
+        initialization. Draining here keeps a prior simulation lifecycle's
+        latched error from poisoning this one. Remove once the upstream Warp
+        fix lands.
+        """
+        device = wp.get_device(str(PhysicsManager._device))
+        if not device.is_cuda:
+            return
+        # Private Warp API: the drain primitive is not exposed publicly; getting the
+        # device above guarantees the runtime is initialized.
+        from warp._src.context import runtime as _wp_runtime
+
+        stale = _wp_runtime.core.wp_cuda_context_check(device.context)
+        if stale != 0:
+            logger.warning("Cleared stale CUDA error %d latched by a prior Warp graph-memory free failure.", stale)
+
+    @classmethod
     def start_simulation(cls) -> None:
         """Start simulation by finalizing model and initializing state.
 
@@ -1152,6 +1176,8 @@ class NewtonManager(PhysicsManager):
         we determine whether the solver needs external collision detection.
         """
         logger.debug(f"Builder: {cls._builder}")
+
+        cls._drain_stale_cuda_error()
 
         # Create builder from USD stage if not provided
         if cls._builder is None:
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -1186,13 +1186,42 @@ class NewtonManager(PhysicsManager):
         device = wp.get_device(str(PhysicsManager._device))
         if not device.is_cuda:
             return
-        # Private Warp API: the drain primitive is not exposed publicly; getting the
-        # device above guarantees the runtime is initialized.
-        from warp._src.context import runtime as _wp_runtime
+        # Private Warp API: the drain primitives are not exposed publicly; getting the
+        # device above guarantees the runtime is initialized. Guard the whole private
+        # interaction so a future Warp internals reshuffle degrades to a skipped drain
+        # rather than hard-failing simulation start.
+        try:
+            from warp._src.context import runtime as _wp_runtime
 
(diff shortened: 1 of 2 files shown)


source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1143,6 +1170,30 @@ def invalidate_fk(
...
+        device = wp.get_device(str(PhysicsManager._device))
+        if not device.is_cuda:
+            return
+        # Private Warp API: the drain primitive is not exposed publicly; getting the
+        # device above guarantees the runtime is initialized.
+        from warp._src.context import runtime as _wp_runtime
+
+        stale = _wp_runtime.core.wp_cuda_context_check(device.context)
+        if stale != 0:
+            logger.warning("Cleared stale CUDA error %d latched by a prior Warp graph-memory free failure.", stale)
```
adenzler-nvidia: This warning can't fire in the scenario it targets: `wp_cuda_context_check` drains the latched error via its first `cudaGetLastError()`, but then (nothing is capturing at this point, so the sync branch always runs) executes `cudaDeviceSynchronize()` and re-reads the now-clean error state, returning that. So it returns 0 precisely when a non-sticky stale error was successfully drained, and nonzero only for sticky errors that can't be cleared — where "Cleared" would be the opposite of what happened. Meanwhile the drained error still reaches stderr as a raw `Warp CUDA error N: ... (in function wp_cuda_context_check, ...)` line via Warp's internal `check_cuda`, with nothing marking it as benign/handled.

Verified on `warp 1.15.0.dev20260612` by latching a real error (failed `cudaMallocAsync` inside Warp's cudart) and calling the drain: it cleared the error and returned 0.

Warp already binds what's needed to report the drain properly: `wp_get_error_string` (the buffer `check_cuda` writes into) and `wp_set_error_output_enabled`. Suppressing the raw print and diffing the buffer reports the drained error once, with framing — and `before` tends to hold the original failing call's message (in the bug scenario, the `wp_free_device_async` free-node failure), which gives attribution the return value can't:

```suggestion
        core = _wp_runtime.core
        # wp_cuda_context_check drains via cudaGetLastError() but returns the
        # post-sync error state (0 once a non-sticky error was drained), and its
        # internal check_cuda() prints the drained error verbatim to stderr;
        # suppress the print and diff Warp's error buffer to report the drain.
        before = core.wp_get_error_string()
        was_enabled = bool(core.wp_is_error_output_enabled())
        core.wp_set_error_output_enabled(0)
        try:
            persistent = core.wp_cuda_context_check(device.context)
        finally:
            core.wp_set_error_output_enabled(1 if was_enabled else 0)
        after = core.wp_get_error_string()

        if persistent != 0:
            logger.error(
                "CUDA error %d persists after drain; the device context is likely unrecoverable: %s",
                persistent,
                after.decode(errors="replace"),
            )
        elif after != before:
            logger.warning(
                "Drained stale CUDA error latched by a prior lifecycle: %s "
                "(last Warp error recorded before drain: %s)",
                after.decode(errors="replace"),
                before.decode(errors="replace") or "<none>",
            )
```

Two known limitations, acceptable for a temporary layer: the buffer diff misses byte-identical consecutive drains (the drain itself still works, only the log is skipped), and the buffer/output flag are process-global, so a concurrent Warp error on another thread during this call could interleave.

AntoineRichard: Applied in `24ad0925c4f` — thanks, the inverted-return-value catch is exactly right. The drain now diffs `wp_get_error_string()` across `wp_cuda_context_check` with `check_cuda`'s stderr print suppressed via `wp_set_error_output_enabled`, so:

- an actual drain (`after != before`, `persistent == 0`) is logged once with the pre-drain buffer for attribution (usually the originating free-node failure), and
- a nonzero return is logged as a likely-unrecoverable persistent error.

I also wrapped the whole `warp._src.context` interaction in `try/except (ImportError, AttributeError)` (folding in greptile's P1) — since the suggestion widens the private-symbol surface to four calls, this degrades to a skipped drain instead of raising out of `start_simulation` if Warp reshuffles internals. Kept your two documented limitations (byte-identical consecutive drains skip only the log; process-global buffer/flag) as acceptable for this temporary layer. Pre-commit clean.

adenzler-nvidia (approved):

ooctipus (approved):

PR #6458, opened 2026-07-10
Fixes first CLI runs failing on aarch64 without a manual LD_PRELOAD

myurasov-nv:
# Description

On Linux aarch64 (e.g. DGX Spark), the very first `./isaaclab.sh` run fails: Isaac Sim refuses to start until the system libgomp is preloaded, and the CLI only prints a warning banner and exits 1. The CLI now preloads `/lib/aarch64-linux-gnu/libgomp.so.1` automatically for its python subprocesses, mirroring what the install CI utilities already do. Also documents `OMNI_KIT_ACCEPT_EULA=yes` for non-interactive first launches and the aarch64 build path in the source installation guide.

| Before | After |
| ------ | ----- |
| warning banner asking to export `LD_PRELOAD` by hand; run exits 1 | same command trains and exits 0 with no manual environment setup |

## Reproduction

1. Fresh install per the pip installation guide on Linux aarch64; do not export `LD_PRELOAD`.
2. Run:

   ```bash
   ./isaaclab.sh train --rl_library rsl_rl --task Isaac-Cartpole-Direct --num_envs 16 \
     presets=newton_mjwarp --max_iterations 5
   ```

3. The run prints `WARNING: For the application to run, some shared libraries must be loaded` and exits 1.

## Type of change

## Checklist


myurasov-nv:
Update: the earlier failures were pre-existing (pillow prebundle issue, fixed upstream in #6437; cartpole tile sizes, fixed in #6463); this branch is rebased onto both fixes.

myurasov-nv:
The ARM CI failures are a runner network outage (PyPI unreachable during pip install), unrelated to this change.

docs/source/setup/installation/source_installation.rst
```diff
@@ -77,7 +77,7 @@ variables to your terminal for the remaining of the installation instructions:
 
       .. code:: bash
 
-         # Isaac Sim root directory
+         # Isaac Sim root directory (on aarch64, replace x86_64 with aarch64)
```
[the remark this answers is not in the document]
myurasov-nv: Dear Greptile! Applied: the export comment now uses your suggested wording, and the source-build symlink example in src_symlink_isaacsim.rst gained the same aarch64 note (commit 491db450a54).

source/isaaclab/isaaclab/cli/utils.py
```diff
@@ -529,6 +529,28 @@ def determine_python_version() -> str:
...
+        return env
+    libgomp = "/lib/aarch64-linux-gnu/libgomp.so.1"
+    if not os.path.exists(libgomp):
+        return env
+    merged = dict(os.environ if env is None else env)
+    preload = merged.get("LD_PRELOAD", "")
+    if "libgomp" in preload:
+        return env
+    merged["LD_PRELOAD"] = f"{libgomp}:{preload}" if preload else libgomp
+    return merged
```
ooctipus: There's an in-repo precedent worth aligning with: the conda activation hook this repo generates (`cli/commands/envs.py`, `torch_gomp.sh`) solves the same problem class and does two things differently — it resolves the library programmatically instead of hardcoding a distro path, and it checks idempotence by exact `LD_PRELOAD` entry rather than substring.

For the system libgomp here it gets even simpler: `LD_PRELOAD` accepts a bare soname, and `ld.so` resolves it through the ldconfig cache on any distro. That deletes both the Debian-multiarch path and the `os.path.exists` check — and removes the silent no-op on RHEL/SLES-layout arm systems, where this helper currently falls back to the broken first-run banner without a hint. The substring check also over-matches (a user preloading a different gomp build would be skipped); comparing path basenames per entry keeps the idempotence exact, including for users who already exported the full path per the docs.

```suggestion
    here makes first runs through the CLI work out of the box. The bare soname is used so
    ``ld.so`` resolves the library through the ldconfig cache on any distro. Returns the
    env unchanged on other platforms or when a libgomp is already preloaded.
    """
    if platform.system() != "Linux" or platform.machine().lower() not in ("aarch64", "arm64"):
        return env
    libgomp = "libgomp.so.1"
    merged = dict(os.environ if env is None else env)
    preload = merged.get("LD_PRELOAD", "")
    if any(os.path.basename(entry) == libgomp for entry in preload.split(":") if entry):
        return env
    merged["LD_PRELOAD"] = f"{libgomp}:{preload}" if preload else libgomp
    return merged
```

Longer-term it may be worth this and the conda hook sharing one implementation (or cross-referencing comments) so the two preload WARs don't drift.

ooctipus (approved):

source/isaaclab/isaaclab/cli/utils.py
myurasov-nv: Applied in the latest commit: uses the bare soname libgomp.so.1 so ld.so resolves it through the ldconfig cache (dropping the hardcoded Debian-multiarch path and the os.path.exists no-op that silently skipped RHEL/SLES-layout arm systems), and the idempotence check now compares per-entry basenames instead of a substring match. Verified on an aarch64 host that LD_PRELOAD=libgomp.so.1 resolves. Agree the conda hook and this WAR should share one implementation eventually; left that as a follow-up rather than widening this PR.

PR #6462, opened 2026-07-10
Reduce CI Log Verbosity and Improve Failure Visibility

mataylor-nvidia:
# Description

CI output is noticeably less verbose for normal successful setup and initialization paths
Failure logs preserve the important debugging details for failed tests
The end of the CI log includes a clear summary of failed tests
Logs from tasks/core are easier to navigate and no longer dominated by scene/task init noise
Engineers can identify the failing test and its main error from the log without excessive scrolling

Fixes # (OMPE-100763)

## Type of change

## Screenshots

## Checklist


tools/conftest.py
mataylor-nvidia, line 1: `--show-capture=all` → added

This tells pytest what to do with the buffered output at the end of each test. failed means: only emit the captured stdout/stderr for tests that actually failed. For passing tests the buffer is silently discarded. For failing tests it's printed as part of the failure report, which then flows through the subprocess stdout to CI.

Net result: a test file with 20 passing tests and 1 failure produces output only for the failure, not for all 21 tests.

source/isaaclab/isaaclab/envs/manager_based_env.py
```diff
@@ -359,17 +365,17 @@ def load_managers(self):
 
         """
         # prepare the managers
-        # -- event manager (we print it here to make the logging consistent)
```
mataylor-nvidia, line 362: print vs logger usage could effect how verbose the logs are

comment says `print it here to make the logging consistent`, but I do not see print used elsewhere besides scripts

tools/conftest.py
```diff
@@ -491,14 +490,15 @@ def _run_one_pass(
         sys.executable,
         "-m",
         "pytest",
-        "-s",
         "--no-header",
+        "--show-capture=failed",
+        "--log-level=WARNING",
```
mataylor-nvidia: @r-schmitt do you think this is okay

I was going to add `"--log-level=WARNING",` but having info messages on test errors may be useful

mataylor-nvidia: We would lose context if these are omitted

tools/conftest.py
mataylor-nvidia: `-s` disables output capture, so print() output and logs show up live, which can make the run look more verbose

pyproject.toml
```diff
@@ -293,6 +293,9 @@ ignore-words-list = "aheared,collet,followng,haa,publically,rouines,slq,collapsa
 
 [tool.pytest.ini_options]
 
+log_level = "WARNING"
+log_file_level = "INFO"
```
mataylor-nvidia: log file should have full traceback for download

tools/conftest.py
```diff
@@ -459,7 +463,7 @@ def _retry_failed_test_in_fresh_process(
 
     while has_test_failures and process_failure_attempts < max_process_failure_retries:
         process_failure_attempts += 1
-        print(
+        logger.info(
             f"⚠️  {test_file}: failed in subprocess"
```
nvsekkin, line 467: nit (non-blocking): this looks like a warning but logged as an info (same question to all old prints that had the triangle emoji)

tools/conftest.py
```diff
@@ -1228,29 +1231,29 @@ def pytest_sessionstart(session):
     if os.environ.get("ISAACLAB_TEST_QUEUE"):
         test_files = list(test_status)
 
-    print("failed tests:", failed_tests)
```
nvsekkin, line 1231: todo: I think failed tests should be an error?

mataylor-nvidia: yes, I was thinking this however is it a test error or an error of the manager

mataylor-nvidia: I think the errors of the manager world be `unable to run tests`, ....

tools/conftest.py
```diff
@@ -601,9 +605,9 @@ def _run_one_pass(
         sys.executable,
         "-m",
         "pytest",
-        "-s",
```
nvsekkin, line 604: question (non-blocking): if there is no capture how does the new flag work `--show-capture=all` if the main change is to remove this flag which is `--capture=no`?

StafaH (approved):

tools/conftest.py
mataylor-nvidia: I am going back to `print` so this should be a concern

tools/conftest.py
mataylor-nvidia: `--capture=no (or -s)` disables capturing entirely. stdout/stderr stream directly to the terminal in real-time, nothing is buffered.
`--show-capture=all` keeps capturing on (default fd mode buffers everything), but when a test fails, pytest dumps all captured stdout, stderr, and logs into the failure report

i.e. only dump the capture on failure

nvsekkin (approved):

PR #6355, opened 2026-07-04
[Newton] Add a kitless training container

hujc7:
## 1. Summary

- Add a standalone Ubuntu 24.04 image for kitless Newton training without Isaac Sim.
- Install the supported `newton,rl[rsl-rl]` dependency set into a Python 3.12 virtual environment.
- Match the existing Isaac Lab container conventions for the workspace path, non-root user, ownership, and GPU environment.
- Document local build and Cartpole validation commands.

## 2. Design

- Use a multi-stage build so compilers and development headers stay out of the runtime stage.
- Assert during the build that the `isaacsim` distribution and `_isaac_sim` tree are absent.
- Include only the runtime packages needed by the installed stack, including the shared Python library required by OpenUSD.
- Keep this image independent of the existing Isaac Sim Compose setup. No GitHub workflow builds, publishes, or consumes it in this PR.

## 3. Test plan

- [x] Built `docker/Dockerfile.newton` locally for `linux/amd64` and loaded the final image.
- [x] Verified the final image runs as uid/gid 1000, has no Isaac Sim installation, imports Isaac Lab/Newton, and sees CUDA.
- [x] Ran the canonical 16-environment Newton MJWarp Cartpole training probe for five RSL-RL iterations; it completed without a traceback and emitted `Training time`.
- [x] Ran `source/isaaclab_newton/test/test_mock_interfaces`: 46 passed.
- [x] Ran `docker/test/test_dockerfile_nonroot.py`: 16 passed.
- [x] Ran `./isaaclab.sh -f` on all files.

## 4. Validation notes

- The locally built final image was 9.26 GB before cleanup.
- The current dependency graph has a pre-existing Pillow constraint conflict between `isaaclab` and `moviepy`; this PR does not alter package constraints.

PR #6465, opened 2026-07-10
[CI] Share one path-gating engine across build, install, and Newton workflows

hujc7:
## Summary

- Extracts the changed-paths gating engine — duplicated as near-identical ~80-line blocks in three workflows — into a single **reusable workflow** `.github/workflows/detect-changes.yml` (net **-98 lines**, and the next gated workflow gets it for free).
- The gate job keeps its original shape exactly: one ubuntu-latest job with a single API-only step and **no checkout** — `workflow_call` workflows are fetched server-side from the caller's own commit, unlike local composite actions which need the repo on disk.
- Behavior-preserving per caller: identical pattern lists, identical fail-open semantics (non-PR events and changed-file listing failures default to run), and the skip-reports-green contract for branch protection is untouched.

> **Stacked on https://github.com/isaac-sim/IsaacLab/pull/6355** — review the last two commits; the rest is the base PR. Commit 1 is the composite-action prototype, commit 2 reworks it as a reusable workflow to keep the checkout-free gate job; kept separate for comparison.

### 1.1 Reusable workflow over composite action
A composite action would force each caller to add a checkout step (local `./` actions resolve against the runner workspace). `workflow_call` keeps the gate byte-for-byte equivalent to the previous inline jobs. This is the repo's first reusable workflow; it is GitHub's intended mechanism for a callable job.

### 1.2 Inputs / output
- `patterns`: one `regex<whitespace>description` per line (regexes are whitespace-free; the first whitespace run is the column separator — avoids the tab-quoting fragility of the previous bash arrays).
- `summary-title` / `summary-note`: absorb the per-workflow summary wording, including build.yaml's push-vs-PR "Triggered jobs" line via a caller-side expression.
- Single `run` output; consumers reference `needs.changes.outputs.run` (five mechanical renames from the old per-workflow output names).

### 1.3 Self-triggering
Each workflow's pattern list includes `^\.github/workflows/detect-changes\.yml$`, so changes to the gate re-run the gated jobs.

## 2. Validation

- All four YAML files parse; caller wiring and the absence of stale output references verified programmatically.
- The engine script was executed directly against all three pattern inputs with 12 simulated cases (pipe-containing regexes, push events, failing `gh api`) — decisions match the previous per-workflow implementations exactly.
- Pre-commit passes on all files.

## 3. Notes

- The gate check is now reported as "Detect Changes / Detect Changes" (caller job / called job). It is not in the required-checks set, so branch protection is unaffected.
- install-ci's step summary previously printed only the match table; it now also gets the standard heading and decision line (cosmetic unification).


StafaH:
Resolved by #6699

PR #6470, opened 2026-07-12
Fix body velocities reading zeros for fixed-base articulations on Newton

ooctipus:
## Description

For fixed-base articulations, the Newton backend's `ArticulationData._create_buffers` replaced the body COM velocity sim binding with zeros as part of the "no root velocity" fallback. Root and body velocities are independent: `ArticulationView.get_root_velocities` legitimately returns `None` for a non-floating base, but `get_link_velocities` is well-defined and already bound by the normal binding path — the fallback then silently overwrote it. Every downstream body-velocity read (`body_com_vel_w`, `body_lin_vel_w`, `body_ang_vel_w`, and observations built on them) returned zeros for every fixed-base robot.

In practice this silently disables all body-velocity observations for fixed-base manipulators: we found it as a dead fingertip-velocity observation channel that a policy trained against (zeros on Newton, live values on PhysX), producing an artificial sim-to-sim transfer gap.

Fix: zero the body-velocity binding only when the view genuinely cannot provide link velocities.

## Type of change

## Testing

Added `test_fixed_base_reports_body_velocities`: fixed-base Panda, commanded joint-space motion, asserts the reported body velocities track the finite difference of body positions. Verified to fail without the fix (reported max 0.0 vs 3.7 m/s finite-difference ground truth) and pass with it, on the mjwarp solver.

## Checklist

- [x] I have updated the changelog with a fragment under `changelog.d`
- [ ] I have updated the documentation (not applicable)

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
+++ b/source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
@@ -1655,6 +1655,12 @@ class ArticulationData(BaseArticulationData):
             self._sim_bind_root_com_vel_w = wp.zeros(
                 (self._num_instances), dtype=wp.spatial_vectorf, device=self.device
             )
+        # Body velocities are well-defined regardless of the base type (fixed-base articulations
+        # still report link velocities); fall back to zeros only when the view genuinely cannot
+        # provide them. Zeroing this binding together with the root velocity silently zeroes
+        # every body-velocity read for fixed-base robots.
+        if self._root_view.get_link_velocities(SimulationManager.get_state_0()) is None:
+            logger.warning("Failed to get body com velocities. Setting body com velocities to zeros.")
             self._sim_bind_body_com_vel_w = wp.zeros(
                 (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
             )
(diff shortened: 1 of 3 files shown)


source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py
```diff
@@ -1655,6 +1655,12 @@ def _create_buffers(self) -> None:
...
             )
+        # Body velocities are well-defined regardless of the base type (fixed-base articulations
+        # still report link velocities); fall back to zeros only when the view genuinely cannot
+        # provide them. Zeroing this binding together with the root velocity silently zeroes
+        # every body-velocity read for fixed-base robots.
+        if self._root_view.get_link_velocities(SimulationManager.get_state_0()) is None:
+            logger.warning("Failed to get body com velocities. Setting body com velocities to zeros.")
             self._sim_bind_body_com_vel_w = wp.zeros(
                 (self._num_instances, self._num_bodies), dtype=wp.spatial_vectorf, device=self.device
             )
```
AntoineRichard, line 1666: Thanks @ooctipus do we even need that case?

AntoineRichard (approved):
Thanks for catching this!

PR #6472, opened 2026-07-12
Remove per-step CUDA stream syncs from mdp term hot paths

ooctipus:
## Description

Several manager terms issue synchronizing host-device transfers on every call:

- `reset_root_state_uniform` rebuilds its range tensors from Python dicts per call (2 pageable H2D copies with stream sync).
- `joint_vel_out_of_limit` indexes CUDA tensors with a Python id list per step (H2D upload + sync).
- `TerminationManager.reset` calls `.item()` once per termination term.
- `randomize_physics_scene_gravity` rewrites its baked distribution tensors with 6 scalar H2D copies per reset batch even when the ranges are unchanged.

This converts the two mdp terms to class-based manager terms that bake their tensors at construction (the manager resolves `SceneEntityCfg` params before instantiating class terms, so `__init__` sees resolved indices), batches the termination-stats transfer, and caches the gravity params.

## Benchmark

Host-blocking cost per call, measured against a CUDA stream with ~2.4 ms of queued work (RTX 5090) — the training-loop situation where a synchronizing call stalls the host on everything queued ahead of it, killing CPU/GPU overlap:

| pattern | before | after |
|---|---|---|
| range dict → device tensor (`reset_root_state_uniform`) | 2396 µs | 3.5 µs (baked read) |
| Python-list joint gather (`joint_vel_out_of_limit`) | 2648 µs | 397 µs (async tensor index) |
| 6× scalar setitem (`randomize_physics_scene_gravity`) | 4154 µs | 0.4 µs (params cache, unchanged ranges) |
| 3× `.item()` term stats (`TerminationManager.reset`) | 2315 µs | 2479 µs (1× `.cpu()`) |

The last row is honest about being the smallest win: both variants drain the stream once; the change reduces three sync points to one, which matters as term counts grow but is near-neutral at three terms.

Equivalence checks (all pass): baked range tensor equals the per-call construction; list-index and tensor-index gathers produce identical results; per-element `.item()` equals batched `.cpu()` reads; the gravity cache writes identical tensor contents.

End-to-end, this was part of a sweep that removed ~56 of ~133 `cudaStreamSynchronize` calls per env step in a dexsuite training loop at 8192 envs (Newton/mjwarp), worth ~+20% collection FPS overall.

**Breaking note:** configurations referencing the terms via `func=mdp.<term>` are unaffected; direct function calls must construct the term first. Flagged as **Breaking:** in the changelog fragment.

## Type of change

- Performance; breaking only for direct function callers of the two converted terms

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/envs/mdp/events.py
+++ b/source/isaaclab/isaaclab/envs/mdp/events.py
@@ -1117,12 +1117,16 @@ class randomize_physics_scene_gravity(ManagerTermBase):
             operation: The operation to apply ('add', 'scale', or 'abs').
             distribution: The distribution type (cached at init, param ignored at runtime).
         """
-        self._dist_param_0[0] = gravity_distribution_params[0][0]
-        self._dist_param_1[0] = gravity_distribution_params[1][0]
-        self._dist_param_0[1] = gravity_distribution_params[0][1]
-        self._dist_param_1[1] = gravity_distribution_params[1][1]
-        self._dist_param_0[2] = gravity_distribution_params[0][2]
-        self._dist_param_1[2] = gravity_distribution_params[1][2]
+        # rewrite the baked device tensors only when the curriculum-driven ranges change
+        params = (tuple(gravity_distribution_params[0]), tuple(gravity_distribution_params[1]))
+        if params != getattr(self, "_last_gravity_params", None):
+            self._last_gravity_params = params
+            self._dist_param_0[0] = gravity_distribution_params[0][0]
+            self._dist_param_1[0] = gravity_distribution_params[1][0]
+            self._dist_param_0[1] = gravity_distribution_params[0][1]
+            self._dist_param_1[1] = gravity_distribution_params[1][1]
+            self._dist_param_0[2] = gravity_distribution_params[0][2]
+            self._dist_param_1[2] = gravity_distribution_params[1][2]
 
         if self._backend == "newton":
             self._call_newton(env, env_ids, operation)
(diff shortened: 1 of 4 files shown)


source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -1838,50 +1843,63 @@ def push_by_setting_velocity(
...
+        env_ids: torch.Tensor,
+        pose_range: dict[str, tuple[float, float]],
+        velocity_range: dict[str, tuple[float, float]],
+        asset_cfg: SceneEntityCfg = SceneEntityCfg("robot"),
+    ):
+        # extract the used quantities (to enable type-hinting)
+        asset: RigidObject | Articulation = env.scene[asset_cfg.name]
+        # get default root state
+        default_root_pose = asset.data.default_root_pose.torch[env_ids].clone()
+        default_root_vel = asset.data.default_root_vel.torch[env_ids].clone()
```
AntoineRichard: Do these really need to be cloned?
If so should they be cached? i.e. clone once?

ooctipus: Good catch — they don't need cloning at all, dropped in 79d67444f4b. Two reasons: `tensor[env_ids]` with a tensor index is fancy indexing, which already returns a copy (never a view), so the `.clone()` was a second allocation of already-private data; and even so, the values are only read — everything downstream (`positions`, `orientations`, `velocities`) is built out-of-place. The clones were inherited from the old function-term.

On caching a clone once: that would actually be wrong here — `default_root_pose`/`default_root_vel` can be legitimately updated at runtime by other events, so a construction-time snapshot would go stale. Since no copy is needed, the question dissolves.

AntoineRichard (approved):

PR #6473, opened 2026-07-12
Gate Newton reset-mask consumption on a host-side dirty flag

ooctipus:
## Description

**This is not a bug fix — behavior is identical before and after.** It removes unnecessary per-step CPU work.

Every physics step, `NewtonManager.step()` and `forward()` run a cleanup block meant for environments that were just reset: it tells the solver to clear its carried-over state for those environments, recomputes their body poses, and clears the bookkeeping masks. The block is driven by per-environment masks, so when nothing was reset it does nothing — but the CPU still dispatches all of it every step, and most steps nothing was reset (with decimation 4, at least 3 of every 4 steps).

This change keeps a simple boolean on the CPU — set whenever any asset write marks an environment as reset (`invalidate_fk`, the only place the masks are written), cleared when the block consumes the masks — and skips the whole block when the boolean says nothing happened.

## Measured effect — honest framing

What it saves: a few GPU-call dispatches per quiet step (a solver reset call, one kernel launch, two memsets).

What it does **not** do: make training measurably faster. A/B at 8192 environments (block gated vs deliberately run every step, two rounds each): 202.1k / 202.9k samples-per-second vs 203.7k / 201.9k — identical within noise. The saved work was not on the critical path in my setup.

So the value is: less pointless per-step dispatch, and the mask lifecycle becomes explicit (one writer, one consumer, a flag that says whether there is anything to consume). If that is not judged worth an extra state flag, closing this PR is a completely reasonable call.

No interaction with #6470 — different file, different mechanism.

## Type of change

- Performance / cleanliness (non-breaking, no behavior change)

## Checklist

### DIFF
--- a/source/isaaclab_newton/changelog.d/zhengyuz-perf-newton-reset-mask-gating.rst
+++ b/source/isaaclab_newton/changelog.d/zhengyuz-perf-newton-reset-mask-gating.rst
@@ -1,0 +1,8 @@
+Changed
+^^^^^^^
+
+* Changed :meth:`~isaaclab_newton.physics.NewtonManager.step` and
+  :meth:`~isaaclab_newton.physics.NewtonManager.forward` to skip the reset-mask consumption
+  (solver-internals reset, masked FK, mask zeroing) when no asset write dirtied the masks
+  since the last boundary. A clean mask already made every launch a GPU no-op, but the host
+  dispatch ran on every physics step.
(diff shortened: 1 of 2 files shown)


ooctipus:
Closing per the measurements in the description: the dispatch it saves is real but produced no measurable end-to-end speedup (A/B at 8192 envs was identical within noise), and the extra state flag isn't worth carrying without a demonstrated win. The two PRs from this series with measured benefits are #6471 and #6472.

PR #6474, opened 2026-07-12
Improve benchmark sampling and synchronized timing diagnostics

AntoineRichard:
# Description

Improves benchmark sampling so short runs better represent steady-state environment throughput. Runtime benchmarks support explicit warm-up frames, always exclude the dedicated startup step from throughput, and compute effective FPS over the complete measured interval. Training and play benchmarks now exclude the first `env.step()` call by default through `--warmup_steps 1`; pass `--warmup_steps 0` to include every step.

Adds environment-step timing to runtime, training, and play with two explicit modes:

- `host_return` records host wall time until `env.step()` returns without forcing queued device work to complete.
- `serialized_synchronized`, enabled by `--measure_sync_step`, synchronizes every measured environment and simulation boundary and partitions `env.step()` time into nested `SimulationContext.step()` calls and the remainder outside them.

The synchronized mode is diagnostic because its synchronization changes the observed schedule. The outside-simulation remainder includes action and actuator processing, state updates, observations, rewards, terminations, resets, manager execution, wrappers, and measurement synchronization; it is not automatically classified as removable Isaac Lab overhead.

Benchmark workload and warm-up counts are validated directly by argparse. Play retains the cross-argument requirement that `warmup_steps < num_frames` so every run produces at least one measured sample.

The schema remains version 1.2 and validates the additive timing partition.

## Dependencies

- Direct dependencies: none.

This is the root PR for the benchmark refactor stack.

## Type of change

- Bug fix
- New feature
- Documentation update

## Validation

- Adapter parser tests across RSL-RL, RL-Games, SKRL, and SB3: 36 passed
- `./isaaclab.sh -p -m pytest source/isaaclab/test/benchmark -q`: 149 passed
- Runtime benchmark smoke tests for default and synchronized modes: 2 passed
- Regression coverage was confirmed to fail before the timing partition fix and pass afterward.
- `./isaaclab.sh -f`
- `./isaaclab.sh -d`
- Earlier branch validation covered PhysX and Newton/MJWarp runtime probes and training/play smoke tests.

## Checklist


AntoineRichard:
Addressed both timing-diagnostic findings in c292852d776:

- **Zero-warmup startup timing:** the runtime benchmark now always executes the first `env.step()` as an excluded startup probe before installing the optional timing recorder. This keeps `startup_time_s.first_step` under the same instrumentation whether `--measure_isaaclab_overhead` is enabled or disabled. `--warmup_frames 0` therefore still excludes the mandatory startup step from throughput. Runtime smoke coverage now exercises zero warmup in both modes.
- **Negative derived overhead:** per-step overhead is now floored at zero. Independently sampled nested wall timers can cross slightly because of clock resolution and observer noise, but the derived metric can no longer report a physically impossible negative value.

Added direct regressions for both cases and verified that each test fails when its corresponding fix is removed. The full benchmark unit suite passes (`133 passed`), both zero-warmup runtime smoke modes pass, and all pre-commit checks pass.

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
```diff
@@ -50,6 +50,17 @@ def _parse_args(argv: list[str]):
     add_isaaclab_launcher_args(parser)
 
     parser.add_argument("--output_path", type=str, default=".", help="Directory to write the output JSON.")
+    parser.add_argument(
+        "--measure_synchronized_step_breakdown",
```
StafaH: This is too long for a command line arg. I suggest finding a short user-friendly term.

"measure_sync_step"?

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
```diff
@@ -50,6 +50,17 @@ def _parse_args(argv: list[str]):
...
+        action="store_true",
+        help="Measure a serialized synchronized simulation and outside-simulation step breakdown.",
+    )
+    parser.add_argument(
+        "--warmup_steps",
+        type=int,
+        default=0,
```
StafaH: I wonder if we should just default this to 1 to ignore the cold start for the end-user.

Similarly, we already have first_step in the startup benchmark, so I wonder what the correct way here is, should it just be warmup >= 1 for runtime FPS

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
```diff
@@ -81,6 +92,10 @@ def _parse_args(argv: list[str]):
         parser.error("Distributed training benchmarks are not supported.")
 
     args_cli, remaining_args = setup_preset_cli(parser, argv)
+    if args_cli.max_iterations is not None and args_cli.max_iterations <= 0:
+        parser.error("--max_iterations must be greater than zero")
+    if args_cli.warmup_steps < 0:
```
StafaH: Nit: A nifty feature of parser.add_argument is the type feature, can add type checks:

parser.add_argument(
   "--my-arg",
   type=lambda x: int(x) if int(x) >= 0 else (_ for _ in ()).throw(argparse.ArgumentTypeError("must be non-negative")),
   help="Unsigned integer")

StafaH (approved):

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: That would work!

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: Yeah we could ignore the first step by default, I didn't want to change the default behavior.

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: Oh I did not know that was an option x). interesting

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: Addressed in 5509354bb24. The public CLI flag is now `--measure_sync_step` across runtime, training, and play, with the docs and smoke coverage updated. I kept the internal `EnvironmentStepTimingRecorder` keyword descriptive because it is not part of the user-facing CLI.

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: Addressed in 5509354bb24. Training and play now default to `--warmup_steps 1`, which removes the first `env.step()` cold start; users can pass `0` to retain every step. Runtime already treats its first step as a dedicated startup diagnostic outside throughput and then defaults to 50 additional warm-up frames, so its runtime behavior did not need another implicit step.

scripts/benchmarks/rl_games/benchmark_rl_games_train.py
AntoineRichard: Addressed in 5509354bb24. Shared argparse converters now enforce positive workloads and non-negative warm-up counts at parse time, including `max_iterations` through the common training parser hook. The play-only relationship `warmup_steps < num_frames` remains a post-parse check because it depends on two arguments.

scripts/benchmarks/test/test_benchmark_smoke.py
```diff
@@ -127,14 +216,59 @@ def test_training_and_play_write_bundles(
     play_data = _load_play_bundle(play_output)
     assert play_data["run"]["framework"] == library
     assert play_data["runtime"]["total_fps"]["mean"] > 0
+    play_timing = play_data["runtime"]["environment_step_timing"]
+    assert play_timing["environment_step_calls"] == 250
```
mmichelis, line 221: [P1] This assertion is incompatible with the new default `warmup_steps=1`. Each play adapter runs exactly `num_frames` calls, and the recorder excludes the first, so this value is 249. The current core CI job confirms the same `249 == 250` failure for all four libraries. The synchronized 10-frame assertion below will likewise receive 9. Please either derive the expected count as `num_frames - warmup_steps` or pass `--warmup_steps 0` when this test intends to record every call.

source/isaaclab/isaaclab/test/benchmark/stepping.py
```diff
@@ -87,6 +273,19 @@ def run_runtime_loop(env, num_frames: int) -> list[float]:
...
+    """Run excluded warmup steps, including at least the first startup step.
+
+    Args:
+        env: A Gym-compatible environment.
+        num_frames: Requested number of warmup environment steps.
+
+    Returns:
+        Per-step wall times [s] for ``max(num_frames, 1)`` excluded steps.
+    """
+    return run_runtime_loop(env, max(num_frames, 1))
```
mmichelis: [P2] This executes `N` total excluded steps for every positive `warmup_frames=N`, but the public docs and runtime call site define `N` as additional warm-up steps after the mandatory startup probe. A focused check gives 1 total step for `N=1` instead of 2, and 50 instead of 51 for the default. Please run `num_frames + 1` here, or change the public contract consistently, and add positive-count cases for 1 and 50.

source/isaaclab/isaaclab/test/benchmark/stepping.py
```diff
@@ -59,21 +60,206 @@ def sample_random_actions(env) -> torch.Tensor | dict[str, torch.Tensor]:
...
+                recording = self._environment_step_index >= self._warmup_steps
+                self._environment_step_index += 1
+                start_time_ns = time.perf_counter_ns()
+                try:
+                    return self._original_env_step(*args, **kwargs)
+                finally:
+                    if recording:
+                        self.step_times_s.append((time.perf_counter_ns() - start_time_ns) / 1e9)
+
+        self._env.step = timed_environment_step
```
mmichelis, line 227: [P3] Please make wrapper installation exception-safe. In synchronized mode `sim.step` is replaced before this assignment. If assigning `env.step` raises, `__enter__` never completes, so `__exit__` is not called: the simulation wrapper remains installed and the recorder stays marked active. This requires an environment that rejects method replacement, so it is a robustness edge case, but a `try`/`except` rollback around installation would preserve context-manager guarantees.

source/isaaclab/isaaclab/test/benchmark/builders.py
```diff
@@ -116,13 +119,30 @@ def build_run_identity(
...
 
 
+def _with_effective_mean(stats: MeanStd, samples: Sequence[float], effective_mean: float) -> MeanStd:
+    """Return throughput statistics centered on an effective aggregate mean."""
+    values = list(samples)
+    if not values:
+        return stats
+    return MeanStd(
+        mean=effective_mean,
+        std=statistics.stdev(values, xbar=effective_mean) if len(values) > 1 else 0.0,
```
mmichelis: [P3] This deliberately reports dispersion around the effective aggregate mean, but the flat formatter labels the result `Std`. For rates `[8, 8/3]`, this emits `4.216370` while the ordinary sample standard deviation is `3.771236`; `statistics.stdev` documents `xbar` as the sample mean. Since the schema and test intentionally define the current calculation, please either give this nonstandard dispersion an explicit name or use a conventional or duration-weighted variance so consumers do not interpret it as sample standard deviation.

scripts/benchmarks/rsl_rl/benchmark_rsl_rl_train.py
```diff
@@ -46,11 +47,23 @@ def _parse_args(argv: list[str]):
...
+    parser.add_argument(
+        "--measure_sync_step",
+        action="store_true",
+        help="Measure a serialized synchronized simulation and outside-simulation step breakdown.",
+    )
+    parser.add_argument(
+        "--warmup_steps",
+        type=parse_non_negative_int,
+        default=1,
+        help="Exclude the first N env.step() calls from environment-step timing. Default 1 removes cold start.",
```
mmichelis, line 65: [P2] Training accepts any nonnegative warm-up count, unlike play's cross-argument validation. If this value consumes every `env.step`, the full training workload completes before `build_runtime` rejects the empty timing series. A one-iteration RSL-RL run completed 16 steps and wrote a checkpoint, then failed with the warm-up exhaustion `ValueError` and emitted no benchmark bundle. The same ordering exists in all four adapters, and `env.close()` is after the failing assembly path rather than in `finally`. Please validate against the resolved workload where possible, define the early-stop empty-sample behavior, and make environment cleanup unconditional.

mmichelis (changes requested):
Requesting changes for three confirmed benchmark regressions and two lower-severity robustness and schema concerns. The current core CI failure is caused by the play warm-up count mismatch. Runtime warm-up semantics and training warm-up exhaustion also need correction before merge. The remaining two comments are P3 robustness and design issues.

source/isaaclab/isaaclab/test/benchmark/stepping.py
mmichelis: Thanks, the concrete read-only `env.step` case now restores `sim.step`. One exception-safety gap remains: `_original_env_step` is set before the protected block, so an exception during synchronized setup, such as the `env.unwrapped.device` lookup, leaves the recorder marked active and a retry raises `EnvironmentStepTimingRecorder is already active`. Also, both installation flags are set after their assignments, so a setter that mutates and then raises leaves that wrapper installed. Please cover all setup after the active check with rollback, mark each installation as pending before assignment, and assert that entry can be retried after a setup failure.

scripts/benchmarks/runtime.py
```diff
@@ -134,15 +153,25 @@ def run(argv: list[str]) -> None:
...
+            )
             with BenchmarkMonitor(benchmark, interval=1.0):
-                step_times_s = stepping.run_runtime_loop(env, args.num_frames)
+                timer_context = environment_step_timer
+                with timer_context:
+                    step_times_s = stepping.run_runtime_loop(env, args.num_frames, reset=False)
+
+            # Keep the actual first step as a startup diagnostic. With no warmup,
+            # the first measured step is also the first post-reset step.
+            first_step_s = warmup_step_times_s[0] if warmup_step_times_s else step_times_s[0]
```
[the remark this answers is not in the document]
mmichelis: This remains reproducible after the exact-count update. With `--measure_sync_step`, `warmup_frames=0` derives `first_step` from the recorder-instrumented measured loop, while `warmup_frames=1` derives it from the raw pre-recorder warm-up loop. A focused boundary counter observed four synchronization calls in the zero-warm-up first step and none in the positive-warm-up first step. Please collect this startup diagnostic at the same timing boundary for every warm-up value so the field remains comparable.

source/isaaclab/isaaclab/test/benchmark/builders.py
```diff
@@ -133,22 +147,120 @@ def build_runtime(
...
+            if simulation_step_times_s is not None and len(simulation_step_times_s) > 0:
+                raise ValueError("simulation timing cannot be present without environment-step timing samples")
+            if simulation_step_calls not in (None, 0):
+                raise ValueError("simulation_step_calls must be zero when environment-step timing is omitted")
+            warnings.warn(
+                "environment-step timing omitted because no samples remained after warm-up",
+                RuntimeWarning,
+                stacklevel=2,
+            )
+            environment_step_times_s = None
```
mmichelis: [P2] When early stopping consumes every post-warm-up sample in synchronized mode, this conversion erases the measurement mode. `_runtime_measurements` infers serialization only from `Runtime.environment_step_timing`, which is now `None`, so the flat output contains `Mean Total FPS` and omits `Mean Serialized Diagnostic Total FPS` even though the recorder synchronized every observed step. A focused empty-sample build reproduces `timing_present=False, normal_label=True, serialized_label=False`. Please preserve the run measurement mode independently of optional timing samples and cover the early-stop plus synchronized case.

source/isaaclab/changelog.d/antoiner-runtime-benchmark-warmup.minor.rst
```diff
@@ -0,0 +1,27 @@
...
+^^^^^^^
+
+* Changed the runtime benchmark to exclude configurable warmup frames and
+  report effective throughput over the complete measured interval, and
+  report environment-step host-return rates across runtime, play, and training
+  benchmarks. Added a ``--warmup_steps`` flag to the play and training
+  benchmarks that excludes the first N ``env.step()`` calls from
+  environment-step timing and defaults to ``1`` to remove cold start. Runtime
+  and play now execute the requested warmup steps before the requested number
+  of measured steps; pass ``--warmup_steps 0`` to measure the first step.
```
mmichelis: [P3] Runtime does not accept `--warmup_steps`; its flag is `--warmup_frames`. As written, this sentence tells runtime users to pass an unknown option. Please name both flags explicitly or split the runtime and play instructions.

mmichelis:
The PR description is stale after the follow-up fixes. It still says runtime always excludes a dedicated startup step, but the current contract makes `warmup_frames` the exact excluded count and lets zero measure the first step. It also says play retains `warmup_steps < num_frames`, while the current adapters deliberately allow larger warm-up values by adding them before the requested measured frames. Please update those acceptance statements and the benchmark test count so the description matches the current head.

scripts/benchmarks/rl_games/benchmark_rl_games_play.py
mmichelis: why have the benchmark_rl_games_train.py script run `with contextlib.closing(env):` and here we manually run env.close()? Is there a reason it is not used consistently?

scripts/benchmarks/runtime.py
```diff
@@ -134,15 +153,21 @@ def run(argv: list[str]) -> None:
...
             num_envs = env.unwrapped.num_envs
 
-            with BenchmarkMonitor(benchmark, interval=1.0):
-                step_times_s = stepping.run_runtime_loop(env, args.num_frames)
+            warmup_step_times_s = stepping.run_runtime_warmup(env, args.warmup_frames)
+            environment_step_timer = stepping.EnvironmentStepTimingRecorder(
+                env, measure_synchronized_step_breakdown=args.measure_sync_step
+            )
+            with environment_step_timer, BenchmarkMonitor(benchmark, interval=1.0):
+                step_times_s = stepping.run_runtime_loop(env, args.num_frames, reset=False)
```
mmichelis, line 161: [P2] With synchronized measurement and positive warm-up, the first headline throughput sample includes work that should have been excluded. The warm-up runs outside the recorder, then `run_runtime_loop` starts its outer timer before the wrapper performs its entry synchronization. Pending work from the final warm-up step is therefore drained inside `step_times_s[0]`. A deterministic boundary reproduction produced outer samples `[0.12, 0.06]` while the recorder produced `[0.06, 0.06]`; lines 175-183 use the contaminated outer series for effective FPS. Please drain the warm-up boundary before outer timing, use the synchronized recorder series for synchronized throughput, or run warm-up through equivalent boundaries, and add a positive-warm-up synchronized regression.

source/isaaclab/isaaclab/test/benchmark/_cli.py
```diff
@@ -0,0 +1,41 @@
...
+def validate_warmup_steps(warmup_steps: int, available_steps: int) -> None:
+    """Validate that training warm-up leaves at least one measured environment step."""
+    if warmup_steps >= available_steps:
+        raise ValueError(
+            f"warmup_steps ({warmup_steps}) must be less than resolved training environment steps "
+            f"({available_steps})"
+        )
+
+
+def _parse_int(value: str) -> int:
```
mmichelis: is this necessary? couldn't int(value) throw the same error

mmichelis (approved):
Left some minor comments, but looks like a great benchmark addition, thanks Antoine!

source/isaaclab/isaaclab/test/benchmark/stepping.py
AntoineRichard: After discussing the lifecycle here, we removed the setup rollback machinery in ddb2521aaf. A failure while installing the recorder aborts the benchmark, so the same partially mutated live environment is not retried and cannot produce a valid result. The recorder still restores wrappers when an exception occurs after successful context entry; we are deliberately not adding recovery state for partial setter side effects during a failing `__enter__`.

scripts/benchmarks/runtime.py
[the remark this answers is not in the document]
AntoineRichard: Keeping this boundary intentionally. Warm-up frames must stay outside the synchronized recorder because they are excluded from timing and throughput. Therefore, with positive warm-up, `first_step` is the raw first excluded warm-up frame; with zero warm-up, it is the first measured frame under the requested measurement mode. Adding an unconditional raw startup call would violate the exact `warmup_frames + num_frames` call contract. Commit 196c03af06 documents this distinction explicitly.

source/isaaclab/isaaclab/test/benchmark/builders.py
AntoineRichard: Changed the empty-sample behavior in 954ee28303. A run with no post-warm-up timing samples is not useful as a benchmark bundle, so `build_runtime` now raises an explicit error explaining that the workload may have stopped before warm-up completed and advising the user to reduce warm-up or execute more steps. The allow-empty branches and trainer opt-ins were removed rather than preserving a mode for an empty aggregate.

source/isaaclab/changelog.d/antoiner-runtime-benchmark-warmup.minor.rst
AntoineRichard: Fixed in 196c03af06. Runtime and play now consistently use `--warmup_frames`, while training uses `--warmup_steps` because its workload is expressed as iterations and environment steps. The play adapters, tests, documentation, metadata, and changelog were updated together.

scripts/benchmarks/rl_games/benchmark_rl_games_play.py
AntoineRichard: Fixed in 8fa155b0c4. All four play adapters now wrap the fully constructed environment in `contextlib.closing`, matching training and guaranteeing cleanup if policy loading, execution, result assembly, or formatting raises. The trailing manual `env.close()` calls were removed.

source/isaaclab/isaaclab/test/benchmark/_cli.py
AntoineRichard: Fixed in 6e3b5a983c. Both range validators now call `int(value)` directly and let argparse handle the resulting `ValueError`; the private `_parse_int` helper was removed. A focused test covers the native argparse conversion error for both validators.

PR #6480, opened 2026-07-13
Add OVRTX instance segmentation idToLabels/idToSemantics

rilei-nvidia:
# Description

The main changes involved updating ovrtx renderer integration's `instance_segmentation_fast` annotator to output `idToLabels` and `idToSemantics` matching Isaac RTX.  And making sure there is test coverage in our rendering correctness.

The the midst of these changes, I noticed that we were stringifying the keys of `semantic_segmentation`'s idToLabels which seem un-desirable, so I've changed it to a tuple.  IsaacRTX is still producing the stringified keys for this annotator, something I will follow up with the synthetic data team.

When I tried to adding an additional ovrtx integration test validating the non-colorized code paths with a procedurally generated USDA scene - I noticed a potential bug with the StableIdMap render var, this has been flagged in internal slack, will see if it can be addressed of which I will follow up with adding this additional integration test. 

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
@@ -88,6 +88,8 @@ def _decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tupl
     if data.size < 4:
         return []
 
+    # ``<u4`` (little-endian): matches the ovannotators reference decoder. kit writes native ``uint32_t``
+    # struct fields with no byte-swap, so this is correct on the (little-endian) platforms it targets.
     entry_dtype = np.dtype([("id", "<u4", (4,)), ("label_length", "<u4"), ("label_offset", "<u4")])
     num_entries = int.from_bytes(data[-4:].tobytes(), byteorder="little")
     entries_size = num_entries * entry_dtype.itemsize
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
@@ -64,7 +64,7 @@ def parse_semantic_label(raw_label: str) -> dict[str, str]:
     return labels
 
 
-def decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tuple[int, int, int, int], str]]:
+def _decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tuple[int, int, int, int], str]]:
     """Decode an OVRTX ``IdentifierMap``-layout render var buffer into ``(id, raw_label)`` pairs.
 
     Shared layout for the ``SemanticIdMap`` and ``StableIdMap`` render vars: the buffer packs, in order,
@@ -118,7 +118,7 @@ def decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tuple
 def decode_semantic_id_map(id_map: np.ndarray) -> dict[int, dict[str, str]]:
     """Decode the OVRTX ``SemanticIdMap`` render var buffer into ``{semantic_id: {type: label}}``.
 
-    The ``id[0]`` field of each :func:`decode_identifier_map` entry is the semantic ID, and the label is
+    The ``id[0]`` field of each :func:`_decode_identifier_map` entry is the semantic ID, and the label is
     parsed with :func:`parse_semantic_label`.
 
     Args:
@@ -130,14 +130,14 @@ def decode_semantic_id_map(id_map: np.ndarray) -> dict[int, dict[str, str]]:
     """
     return {
         id_words[0]: parse_semantic_label(raw_label)
-        for id_words, raw_label in decode_identifier_map(id_map, "SemanticIdMap")
+        for id_words, raw_label in _decode_identifier_map(id_map, "SemanticIdMap")
     }
 
 
 def decode_stable_id_map(id_map: np.ndarray) -> dict[tuple[int, int, int, int], str]:
     """Decode the OVRTX ``StableIdMap`` render var buffer into ``{stable_id: prim_path}``.
 
-    Same :func:`decode_identifier_map` layout as the SemanticIdMap, but the four ``id`` words form the
+    Same :func:`_decode_identifier_map` layout as the SemanticIdMap, but the four ``id`` words form the
     128-bit stable ID key and the label is a USD prim-path string (not a semantic label, so it is not parsed).
 
     Args:
@@ -146,7 +146,7 @@ def decode_stable_id_map(id_map: np.ndarray) -> dict[tuple[int, int, int, int], 
     Returns:
         Mapping from the four-word stable ID to its USD prim-path string.
     """
-    return {id_words: raw_label for id_words, raw_label in decode_identifier_map(id_map, "StableIdMap")}
+    return {id_words: raw_label for id_words, raw_label in _decode_identifier_map(id_map, "StableIdMap")}
 
 
 def decode_stable_id_semantic_id_map(id_map: np.ndarray) -> list[tuple[tuple[int, int, int, int], int]]:
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
@@ -3,10 +3,16 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-"""Semantic annotator utilities for the OVRTX renderer.
+"""Host-side stand-in for the not-yet-released ``ovannotators`` segmentation annotators.
 
-Decodes the OVRTX ``SemanticIdMap`` render var into an ``idToLabels`` mapping compatible with the
-Isaac RTX / Replicator contract exposed through ``camera.data.info["semantic_segmentation"]``.
+Until ``ovannotators`` ships, these helpers decode the OVRTX segmentation "map" render vars and rebuild the
+``idToLabels`` / ``idToSemantics`` info dicts that Replicator / Isaac RTX expose through ``camera.data.info``:
+
+- ``semantic_segmentation``: from ``SemanticIdMap`` -> ``idToLabels`` (id -> ``{type: label}``).
+- ``instance_segmentation_fast``: from ``StableIdSemanticIdMap`` + ``StableIdMap`` + ``SemanticIdMap`` ->
+  ``idToLabels`` (instance -> USD prim path) and ``idToSemantics`` (instance -> ``{type: label}``).
+
+For colorized outputs the keys are raw ``(r, g, b, a)`` tuples, matching Replicator's fast segmentation nodes.
 """
 
 from __future__ import annotations
@@ -21,9 +27,17 @@ from .ovrtx_renderer_kernels import generate_random_colors_from_ids_kernel
 # with ID >= 2 are decoded from the SemanticIdMap render var.
 SEMANTIC_ID_BACKGROUND = 0
 SEMANTIC_ID_UNLABELLED = 1
+# Single source of truth for the reserved ID 0/1 label names, shared across every segmentation map and
+# matching the ``"BACKGROUND"`` / ``"UNLABELLED"`` strings Replicator writes for pixel IDs 0 and 1.
+_RESERVED_LABEL_NAMES: dict[int, str] = {
+    SEMANTIC_ID_BACKGROUND: "BACKGROUND",
+    SEMANTIC_ID_UNLABELLED: "UNLABELLED",
+}
+# Reserved prim-path labels (plain strings) for the instance-segmentation ``idToLabels`` mapping.
+RESERVED_INSTANCE_LABELS: dict[int, str] = dict(_RESERVED_LABEL_NAMES)
+# Reserved ``{type: label}`` dicts for the semantic ``idToLabels`` and instance ``idToSemantics`` mappings.
 RESERVED_SEMANTIC_LABELS: dict[int, dict[str, str]] = {
-    SEMANTIC_ID_BACKGROUND: {"class": "BACKGROUND"},
-    SEMANTIC_ID_UNLABELLED: {"class": "UNLABELLED"},
+    reserved_id: {"class": name} for reserved_id, name in _RESERVED_LABEL_NAMES.items()
 }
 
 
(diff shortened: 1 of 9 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -112,6 +115,13 @@ def get_render_var_configs(data_types: list[str]) -> list[tuple[str, str, str]]:
     # renderer can decode the semantic-ID-to-label mapping into camera.data.info["semantic_segmentation"].
     if render_vars[0][2] == "SemanticSegmentation":
         render_vars.append(("/Render/Vars/SemanticIdMap", "SemanticIdMap", "SemanticIdMap"))
+    # When instance segmentation is the resolved AOV, author the three map render vars needed to resolve each
+    # NonStableInstanceSegmentation pixel ID to a prim path (StableIdSemanticIdMap -> StableIdMap) and a
+    # semantic label (StableIdSemanticIdMap -> SemanticIdMap) for camera.data.info["instance_segmentation_fast"].
+    if render_vars[0][2] == "NonStableInstanceSegmentation":
+        render_vars.append(("/Render/Vars/StableIdSemanticIdMap", "StableIdSemanticIdMap", "StableIdSemanticIdMap"))
+        render_vars.append(("/Render/Vars/SemanticIdMap", "SemanticIdMap", "SemanticIdMap"))
+        render_vars.append(("/Render/Vars/StableIdMap", "StableIdMap", "StableIdMap"))
```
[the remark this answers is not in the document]
rilei-nvidia: Fixed

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
```diff
@@ -50,101 +64,219 @@ def parse_semantic_label(raw_label: str) -> dict[str, str]:
     return labels
 
 
-def decode_semantic_id_map(id_map: np.ndarray) -> dict[int, dict[str, str]]:
-    """Decode the OVRTX ``SemanticIdMap`` render var buffer into ``{semantic_id: {type: label}}``.
+def decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tuple[int, int, int, int], str]]:
```
pbarejko: What would you say about introducing some type aliases here? The return type hint seems very cryptic:

```python
def decode_identifier_map(id_map: np.ndarray, map_name: str) -> list[tuple[tuple[int, int, int, int], str]]:
```

something as follows:

``` python
from typing import TypeAlias
import numpy as np

EncodedId: TypeAlias = tuple[int, int, int, int]
RawLabel: TypeAlias = str
EncodedIdLabelPair: TypeAlias = tuple[EncodedId, RawLabel]
EncodedIdList: TypeAlias = list[EncodedIdLabelPair]

def decode_identifier_map(id_map: np.ndarray, map_name: str) -> IdentifierList:
```

I see reading this becoming difficult in the future, entire code is scattered with dicts/list etc. Type hints would be very helpful for readability.

pbarejko: Another question, is there a reason why we return a list instead of a map?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
```diff
@@ -50,101 +64,219 @@ def parse_semantic_label(raw_label: str) -> dict[str, str]:
...
+    """Decode an OVRTX ``IdentifierMap``-layout render var buffer into ``(id, raw_label)`` pairs.
 
-    The buffer packs, in order: an array of ``SemanticIdentifierMap`` entries, the label string blob, and a
-    trailing ``uint32`` entry count. Each entry is ``{uint32 id[4]; uint32 labelLength; uint32 labelOffset}``;
-    the ``id[0]`` field is the semantic ID and the label bytes live at ``[labelOffset, labelOffset + labelLength)``.
-    Mirrors the reference decode shipped with the ovrtx runtime.
+    Shared layout for the ``SemanticIdMap`` and ``StableIdMap`` render vars: the buffer packs, in order,
+    an array of ``IdentifierMap`` entries, the label string blob, and a trailing ``uint32`` entry count.
+    Each entry is ``{uint32 id[4]; uint32 labelLength; uint32 labelOffset}``; the label bytes live at
```
pbarejko, line 89: Isn't the `labelLength` and `labelOffset` an implementation detail that we don't want to expose? To my understanding you use `uint32 id[4]`  only.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
```diff
@@ -50,101 +64,219 @@ def parse_semantic_label(raw_label: str) -> dict[str, str]:
...
+    The ``id[0]`` field of each :func:`decode_identifier_map` entry is the semantic ID, and the label is
+    parsed with :func:`parse_semantic_label`.
 
-def semantic_color_keys(semantic_ids: list[int], device: str) -> list[str]:
-    """Return the ``"(r, g, b, a)"`` color-tuple key for each semantic ID.
+    Args:
+        id_map: Raw SemanticIdMap buffer mapped to host memory (any dtype; viewed as bytes).
+
+    Returns:
+        Mapping from semantic ID (>= 2) to its parsed ``{type: label}`` dict. Reserved IDs (0/1) are not
```
pbarejko, line 147: I wouldn't hard code `2` value in multiple places, with information that first two ids are reserved. Perhaps we could introduce a global variable that is used to offset ids.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
```diff
@@ -50,101 +64,219 @@ def parse_semantic_label(raw_label: str) -> dict[str, str]:
...
+
+    Args:
+        id_map: Raw StableIdSemanticIdMap buffer mapped to host memory (any dtype; viewed as bytes).
+
+    Returns:
+        One ``(stable_id, semantic_id)`` per entry, indexed by ``pixel_id - 2``. ``stable_id`` is the four
+        ``uint32`` stable-ID words (the :func:`decode_stable_id_map` key); ``semantic_id`` is ``semanticId[0]``.
+    """
+    data = np.ascontiguousarray(id_map).view(np.uint8).reshape(-1)
+    entry_dtype = np.dtype([("stable_id", "<u4", (4,)), ("semantic_id", "<u4", (4,))])
```
pbarejko, line 188: This decoder currently interprets all uint32 fields as little-endian via "<u4". Is the OVRTX render-var buffer format explicitly little-endian, or should this be native-endian? Is little-endian part of the contract?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
rilei-nvidia: Good call on the type aliasing - that would certainly make it more readible.  

> Another question, is there a reason why we return a list instead of a map?

decode_identifier_map is an internal utility - the *Map render vars are an array of structs and this function returns that data structure as is before being converted into a dict by the public facing functions.  I will add a _ prefix to mark as internal!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
rilei-nvidia: Yes good call

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
rilei-nvidia: Per other comment I will mark decode_identifier_map as internal as it is only used within the ovrtx_annotator_utils module

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_annotator_utils.py
rilei-nvidia: According to our reference implementation that is the case:
https://github.com/NVIDIA-Omniverse/ovannotators-internal/blob/main/source/ovannotators/writers/_helpers.py#L81

However I looked into kit and it is not explicitly little-endian - it happens to be because that kit currently only targets little-endian platforms.  So good to call that out.  I'll add a comment here describing this, and we should flag this with the synthetic data team around this little-endian assumption that could result in future issues should ovrtx target new platforms.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -766,6 +773,44 @@ def _process_semantic_id_map(self, render_data: OVRTXRenderData, frame) -> None:
...
+        tuples the segmentation buffer uses (``colorize_instance_segmentation=True``); the reserved BACKGROUND
+        (ID 0) and UNLABELLED (ID 1) entries are always included.
+
+        Args:
+            render_data: OVRTX render data for the current frame.
+            frame: OVRTX frame holding the mapped render vars.
+        """
+        required_vars = ("StableIdSemanticIdMap", "StableIdMap", "SemanticIdMap")
+        if any(var_name not in frame.render_vars for var_name in required_vars):
+            return
```
huidongc: `_process_instance_segmentation_maps()` silently returns when any of the three map render vars is missing. Consider a debug log or explicit empty-metadata contract so misconfiguration isn’t silent.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -925,6 +970,10 @@ def _process_render_frame(self, render_data: OVRTXRenderData, frame, output_buff
             "instance_segmentation_fast",
             self.cfg.colorize_instance_segmentation,
         )
+        # Decode the StableIdSemanticIdMap/StableIdMap/SemanticIdMap trio into
+        # camera.data.info["instance_segmentation_fast"]["idToLabels" / "idToSemantics"].
```
huidongc: ` / ` is vague in that it could mean A or B, not A and B. :)

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -766,6 +773,44 @@ def _process_semantic_id_map(self, render_data: OVRTXRenderData, frame) -> None:
             )
         }
 
+    def _process_instance_segmentation_maps(self, render_data: OVRTXRenderData, frame) -> None:
+        """Decode the instance-segmentation map render vars into ``renderer_info["instance_segmentation_fast"]``.
+
+        Populates ``"idToLabels"`` (instance pixel ID -> USD prim path) and ``"idToSemantics"`` (instance pixel
+        ID -> ``{semantic_type: label}``) compatible with Isaac RTX / Replicator. Resolving both requires all
+        three map render vars — ``StableIdSemanticIdMap`` (pixel ID -> stable ID + semantic ID), ``StableIdMap``
```
huidongc, line 969: > ``StableIdSemanticIdMap`` (pixel ID -> stable ID + semantic ID)

Isn't it stable id -> semantic id?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: maybe a warning even an error if you go with logging.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
huidongc, line 95: The method doc should state what each of the str in the tuple represents. Can copy from get_render_var_config().

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
```diff
@@ -100,17 +100,27 @@ def get_render_var_configs(data_types: list[str]) -> list[tuple[str, str, str]]:
     in ``data_types`` so PPISP can consume the HDR AOV alongside the LDR
     destination on the same render product, plus ``SemanticIdMap`` when
     ``"semantic_segmentation"`` is requested so the semantic-ID-to-label
-    mapping can be decoded for ``camera.data.info``. Other multi-AOV
-    combinations are not supported.
+    mapping can be decoded for ``camera.data.info``, plus the
+    ``StableIdSemanticIdMap`` / ``SemanticIdMap`` / ``StableIdMap`` trio when
+    ``"instance_segmentation_fast"`` is requested so the instance-ID-to-prim-path
+    (``idToLabels``) and instance-ID-to-semantic (``idToSemantics``) mappings can be
+    decoded. Other multi-AOV combinations are not supported.
```
huidongc: This paragraph has become hard to read. Maybe make a bullet list.

huidongc (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Great catch.  In this instance we would expect all 3 to be available, so will raise an exception!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Will improve clarity

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Nope, each entry StableIdSemanticIdMap contains a Stable ID, Semantic ID pair!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
rilei-nvidia: I'll update the documentation.  I think in this instance we can benefit from it being a namedtuple so it is more self-documenting.  I've created OMPE-101469 for refactor

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py
rilei-nvidia: Will do

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Hard fail would be good too.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: I am confused about the description "pixel ID -> stable ID + semantic ID". where is pixel ID from?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -948,6 +955,51 @@ def _process_semantic_id_map(self, render_data: OVRTXRenderData, frame) -> None:
...
+                f"instance_segmentation_fast was requested but the following render vars are missing from the "
+                f"OVRTX frame: {missing}. Available vars: {list(frame.render_vars.keys())}"
+            )
+
+        with frame.render_vars["StableIdSemanticIdMap"].map(device=Device.CPU) as mapping:
+            stable_id_semantic_id_map = decode_stable_id_semantic_id_map(np.from_dlpack(mapping.tensor))
+        with frame.render_vars["StableIdMap"].map(device=Device.CPU) as mapping:
+            stable_id_to_path = decode_stable_id_map(np.from_dlpack(mapping.tensor))
+        with frame.render_vars["SemanticIdMap"].map(device=Device.CPU) as mapping:
+            semantic_id_to_labels = decode_semantic_id_map(np.from_dlpack(mapping.tensor))
```
huidongc: Replace ``mapping.tensor`` with direct DLPack access on render-var mappings. The ovrtx runtime now deprecates ``.tensor`` for single-tensor render variables in favor of passing the mapping directly to ``wp.from_dlpack``/``np.from_dlpack``.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: The instance pixel ID are from the NonStableInstanceSegmentation render var and is a 2D image of per-pixel IDs that index into StableIdSemanticIdMap

huidongc: thanks for explaining. could you add more docs here? don't be short ;)

I worked on something similar to these render vars in AVSim, but my brain gets rusty and can't remember the details.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Fixed

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Added documentation

PR #6481, opened 2026-07-13
Decouple motor velocity limit from solver clamp for implicit actuators

ooctipus:
## Description

Actuator configs have two velocity-limit fields that mean different things, and implicit actuators currently refuse to use them together: `velocity_limit` is silently ignored (with a deprecation warning), and setting both raises a `ValueError`.

This PR gives each field one clear job and lets them coexist:

- **`velocity_limit` — what the motor can actually do** (the datasheet number). It fills the articulation data buffers (`soft_joint_vel_limits`) that velocity-limit terminations, rewards, and observations read. It is **not** sent to the physics solver.
- **`velocity_limit_sim` — an artificial hard clamp inside the solver** (PhysX `maxJointVelocity`). Useful as a numerical safety net, but it has no physical counterpart: a real motor limits speed through its torque curve, not by kinematically clamping the joint.

Why this matters: a task that wants to terminate or penalize when a joint exceeds the motor's rated speed needs the datasheet number in the data buffers — but pushing that same number into the solver as a hard clamp changes the physics (the joint can never exceed it, so the termination becomes unreachable and sim-to-sim transfer across engines with different clamp behavior breaks). Keeping the two independent lets configs express "the motor is rated at 2.6 rad/s, terminate above it, but leave the solver clamp at a loose numerical guard."

Backwards compatibility: configurations that set only one of the two fields behave exactly as before (only `velocity_limit_sim` set → it also fills the data buffers, unchanged). The breaking part is narrow and flagged in the changelog: setting both is now valid instead of an error, and `velocity_limit` alone is now honored in the data buffers instead of ignored with a warning.

## Type of change

- Breaking change (narrow: removes an error/ignore path on implicit actuators)

## Checklist


AntoineRichard (approved):
In general i'm a bit concerned with these changes. This could actively change the way an actuator responds and break things for users. This should me marked as a breaking change. LGTM otherwise

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -74,15 +74,18 @@ class ActuatorBase(ABC):
     """
 
     velocity_limit: torch.Tensor
-    """The velocity limit for the actuator group. Shape is (num_envs, num_joints).
+    """The motor velocity limit for the actuator group [rad/s or m/s]. Shape is (num_envs, num_joints).
```
Mayankm96: Suggest renaming it to "joint" velocity. Motor velocity limit typically refers to the actual motor before the gear box.

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -78,28 +78,16 @@ def __init__(self, cfg: ImplicitActuatorCfg, *args, **kwargs):
...
-            #   We should do this once all parameters have an "_sim" suffix.
+        # 'velocity_limit' is the motor's rated speed: it feeds the data buffers
+        # (:attr:`ArticulationData.soft_joint_vel_limits`, read by e.g. the
+        # ``joint_vel_out_of_limit`` termination) but is NOT pushed to the physics
+        # solver. 'velocity_limit_sim' is a solver-level hard clamp (PhysX
+        # ``maxJointVelocity``) with no physical counterpart -- a real motor limits
+        # speed through its torque curve, not a kinematic clamp. The two are
+        # therefore resolved independently; when only the sim clamp is given, it
+        # doubles as the motor limit so the data buffers stay meaningful.
+        if cfg.velocity_limit_sim is not None and cfg.velocity_limit is None:
```
Mayankm96, line 92: What about the case when `velocity_limit` is not None but `velocity_limit_sim` is None?

source/isaaclab_newton/test/assets/test_articulation.py
```diff
@@ -1660,9 +1660,10 @@ def test_setting_velocity_limit_implicit(
     """Test setting of velocity limit for implicit actuators.
 
     This test verifies that:
-    1. Velocity limits can be set correctly for implicit actuators
-    2. The limits are applied correctly to the simulation
-    3. The limits are handled correctly when both sim and non-sim limits are set
+    1. The solver clamp ``velocity_limit_sim`` is applied to the simulation; when unset, the
+       USD-authored value is kept
+    2. The motor limit ``velocity_limit`` is never pushed to the solver and keeps its
```
Mayankm96: Rename motor to joint.

source/isaaclab_newton/test/assets/test_articulation.py
```diff
@@ -1697,28 +1694,21 @@ def test_setting_velocity_limit_implicit(
...
-        #   In this case, the velocity limit is set to the USD value
-        limit = vel_limit_sim
-
-    # check max velocity is what we set
-    expected_velocity_limit = torch.full_like(newton_vel_limit, limit)
+        sim_limit = vel_limit_sim
+    expected_velocity_limit = torch.full_like(newton_vel_limit, sim_limit)
     torch.testing.assert_close(newton_vel_limit, expected_velocity_limit)
 
+    # the motor limit keeps its configured value and is not pushed to the solver;
```
Mayankm96: Rename motor to joint.

source/isaaclab/isaaclab/actuators/actuator_pd.py
ooctipus: in this case there is no solver velocity clamping, and velocity_limit is recorded for as default (soft_limit=1.0 means soft limit equals to the default, same behavior as explicit actuator), the only difference is that implicit actuator does not provide motor level clamping, now I think about it maybe we can ask newton actuator to provide the clamping.

Mayankm96 (dismissed):
> velocity_limit — what the motor can actually do (the datasheet number). It fills the articulation data buffers (soft_joint_vel_limits) that velocity-limit terminations, rewards, and observations read. It is not sent to the physics solver.

This seems consistent with how we handle explicit actuators (where velocity limit is the peak velocity limit). I can see in the future, `soft_joint_vel_limits` shouldn't be a simple factor but could be defined more independently (where velocity limit is the true limits, but soft joint limit is the limit we'd like for safe operation).

Otherwise, looks okay. It may break existing configuration so a behavior change should be notified to the user with a warning.

source/isaaclab/isaaclab/actuators/actuator_base.py
ooctipus: Renamed to joint velocity terminology throughout (docstrings, cfg attention block, comments, and test wording) in 55d3d5c4bec.

source/isaaclab/isaaclab/actuators/actuator_pd.py
ooctipus: Documented this case explicitly in the comment and added the construction-time warning you asked for in the review: only-velocity_limit-set now warns that the value feeds the data buffers (previously ignored) and that velocity_limit_sim is the solver clamp. The solver keeps the USD-authored clamp in this case. Done in 55d3d5c4bec.

source/isaaclab_newton/test/assets/test_articulation.py
ooctipus: Renamed in 55d3d5c4bec (both newton and physx test copies, plus the unit test).

source/isaaclab_newton/test/assets/test_articulation.py
ooctipus: Renamed in 55d3d5c4bec.

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -78,28 +78,16 @@ def __init__(self, cfg: ImplicitActuatorCfg, *args, **kwargs):
...
             cfg.velocity_limit = cfg.velocity_limit_sim
```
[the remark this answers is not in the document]
ooctipus: Confirmed real: the dict branch of _parse_joint_parameter zero-fills unmatched joints. Fixed at the velocity-limit call site (unmatched joints fall back to the resolved solver clamp) with a regression test that fails without the fix. Done in 55d3d5c4bec.

source/isaaclab/changelog.d/zhengyuz-actuators-velocity-limit-semantics.minor.rst
```diff
@@ -0,0 +1,12 @@
...
+
+* **Breaking:** Changed the velocity-limit semantics of implicit actuators.
+  :attr:`~isaaclab.actuators.ActuatorBaseCfg.velocity_limit` now describes the motor's rated
+  speed: it populates the articulation data buffers (e.g.
+  :attr:`~isaaclab.assets.ArticulationData.soft_joint_vel_limits`, read by velocity-limit
+  terminations and rewards) and is not pushed to the physics solver, while
+  :attr:`~isaaclab.actuators.ActuatorBaseCfg.velocity_limit_sim` remains the solver-level
+  clamp. Setting both on an implicit actuator is now valid (previously a ``ValueError``), and
+  ``velocity_limit`` alone is no longer silently ignored with a deprecation warning. Existing
+  configurations that set only one of the two attributes behave exactly as before.
```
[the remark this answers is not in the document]
ooctipus: Corrected in 55d3d5c4bec: the changelog now states that velocity_limit-only implicit configs change behavior (previously ignored, now feeds the data buffers), that a warning is emitted at construction, and how to keep the old behavior.

ooctipus:
Addressed the review in 55d3d5c4bec:
- Renamed *motor* → *joint* velocity terminology in docstrings, cfg docs, comments, and tests (@Mayankm96)
- Added a construction-time warning when an implicit actuator sets only `velocity_limit`, since that value was previously ignored and now feeds the velocity-limit data buffers (@Mayankm96, @AntoineRichard)
- The changelog entry is marked **Breaking:** and now spells out exactly which configurations change behavior and how to migrate

Locally verified: 164 actuator unit tests + 16 newton + 16 physx velocity-limit tests pass.

source/isaaclab/isaaclab/actuators/actuator_pd.py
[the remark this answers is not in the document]
ooctipus: Correction after a second look: the zero-fill is real but it is the uniform, long-standing dict convention of _parse_joint_parameter for ALL actuator parameters (effort_limit, stiffness, damping, and velocity_limit on explicit actuators alike) — dict keys are regexes expected to cover the actuator group's joints. Special-casing only implicit-actuator velocity_limit made dict semantics inconsistent across actuator types, so the carve-out is reverted in d3914119631. If per-joint fallback for partial dicts is desirable, it should be changed for all parameters in _parse_joint_parameter in a separate PR.

PR #6482, opened 2026-07-13
Support multi-wildcard articulation root expressions in PhysX views

ooctipus:
## Description

An articulation config whose prim path matches several articulations per environment (for example a rig with independent articulated parts: `env_.*/Rig/parts/part_.*`) currently cannot create a physics view — the expression has two wildcards, and `create_articulation_view` only understands one.

This handles that case at the view-creation site in `Articulation._initialize_impl`, using the existing `find_matching_prim_paths` utility: the extra wildcard dimensions are expanded into one single-wildcard pattern per distinct sub-asset path (the view API accepts a pattern list), keyed on the full per-environment relative path so the wildcard does not have to be in the last segment, validated so every environment contains the same set of articulations (clear error otherwise — a shared view needs a homogeneous layout), and sorted so the view layout is deterministic. Single-wildcard expressions take the exact same path as before.

Motivation: assets composed of multiple independent articulations per environment (e.g. a device with per-part articulations). Related in spirit to #6256's explicit prim-path APIs, but independent of it.

**Draft** until the multi-articulation scene is validated end-to-end in simulation; single-wildcard behavior is unchanged by construction and covered by every existing articulation test.

## Type of change

- New feature (non-breaking)

## Checklist


ooctipus:
Closing: on review, nothing in the dexsuite/franka PR series this was split from actually needs multi-wildcard articulation roots — it only serves a side project's multi-part asset that isn't upstream-bound yet. The branch stays on the fork; when that asset work goes upstream this can be reopened alongside a real in-tree use case and a sim-level test.

PR #6484, opened 2026-07-13
Add opt-in async OVRTX rendering

pv-nvidia:
# Description

Adds an opt-in asynchronous render path to the OVRTX renderer, so rendering can overlap with simulation and Python work. It is off by default (`async_rendering=False`), leaving existing behavior unchanged.

## What this adds

A single on/off switch, `OVRTXRendererCfg.async_rendering` (default `False`):

- `False`: synchronous rendering.
- `True`: asynchronous rendering. Each `render()` submits the render and returns immediately, so rendering overlaps simulation and Python work; camera outputs arrive one frame later. The first frame after (re)initialization is consumed synchronously so the first camera read returns a valid frame instead of the zero-initialized buffer.

Internally the renderer delegates frame execution to a `_RenderStrategy` pair: `_SyncRenderStrategy` writes transforms straight into OVRTX and consumes each step inline, while `_AsyncRenderStrategy` pipelines steps and double-buffers transform staging. The async strategy is created only when async is enabled, so sync call sites carry no branching.

Async works on both scene-ownership paths. Under ovstage the renderer borrows the stage's storage, so each ovstage scene write first settles any render still in flight.

Two environment variables support testing and benchmarking:

- `OVRTX_ASYNC_RENDERING` overrides `async_rendering` (`0`/`false`/`no`/`off` disable, any other non-empty value enables). This is the only task-agnostic switch; a Hydra override such as `env.scene.base_camera.renderer_cfg.async_rendering=True` names a camera that a given task may not define.
- `OVRTX_NUM_BUFFERS` sets the render queue depth: renders kept in flight before the oldest is drained (default `2`, values below `2` are clamped). Larger values overlap more simulation at the cost of extra frames of camera latency.

## Camera-output read ordering

Also improves camera-output throughput for synchronous rendering as well as asynchronous. OVRTX waits for render completion with a GPU-side wait on the CUDA default stream, which on Linux hits a driver GPU-context scheduling issue that runs most of the graphics work while the render-output copy waits. Camera outputs are now read without a GPU-side wait, with the calling thread waiting instead — roughly 30% higher throughput on a camera task on Linux, and slightly faster on Windows too.

Two variables select how the read is ordered against the render: `ISAAC_LAB_OVRTX_RENDER_VAR_SYNC_STREAM` picks the GPU-side barrier (`0` default/none, `default`, or `warp`), and `ISAAC_LAB_OVRTX_RENDER_VAR_MAP_WAIT` (default `1`) toggles the host-side wait. Selecting neither would race the render and is rejected.

## Correctness

The async path is covered by golden-image tests that reuse the synchronous goldens, since async output must match within the existing SSIM / pixel-difference tolerances: `test_rendering_cartpole_kitless.py::test_rendering_cartpole_kitless_async` for both kitless physics backends (`ovphysx`, `newton`). Per-env tolerances are declared as `[sync, async]` and resolved through `max_different_pixels_percentage_for()`, guarded by `test_rendering_tolerance_lookup.py`.

Unit tests cover the pieces that are hard to observe end to end:

- `test_ovrtx_scene_write_barrier.py` — an ovstage scene write settles in-flight renders, and delivery targets the correct frame's buffers.
- `test_ovrtx_scene_backend_binding.py` — every scene operation resolves to the selected backend, with no unpaired implementation.
- `test_ovrtx_strategy_selection.py`, `test_ovrtx_render_ordinal.py`, `test_ovrtx_strategy_drain_state.py`, `test_ovrtx_async_teardown.py`.

The `isaaclab_ov` suite passes (164 tests), including the clone-plan tests after the async-state refactor.

## Benchmarks

Measured with the RSL-RL training benchmark (`scripts/benchmarks/training.py`) with **training in the loop** (real PPO updates, not just environment stepping) on the KukaAllegro lift camera task — a manipulation task with per-env RGB cameras, so representative of a real training workload rather than a rendering microbenchmark.

FPS is RSL-RL `Perf/total_fps` (whole training step, including the policy update); collection FPS is the rollout phase where rendering runs. Single runs with schema output — treat as throughput signals.

Benchmark provenance:

- GPU: NVIDIA L40
- OVRTX: `0.3.0.312915`
- Task: `Isaac-Lift-KukaAllegro-Camera`, 4096 envs, seed 42, 200 iterations
- Selectors: `physics=newton_mjwarp renderer=ovrtx presets=rgb64`

### KukaAllegro lift camera (RSL-RL, training in the loop)

| metric | sync | async | speedup |
|:-------|-----:|------:|--------:|
| total FPS (mean)      | 31,864 | 45,545 | 1.43x |
| collection FPS (mean) | 33,663 | 50,199 | 1.49x |

End-to-end training throughput improves ~1.43x; the rollout phase, where rendering overlaps simulation, improves ~1.49x. A per-phase profile of this task shows rendering is ~63% of the iteration, so overlapping it with the remaining ~37% of non-render work is where the gain comes from.

Command (prefix with `OVRTX_ASYNC_RENDERING=1` for async, `OVRTX_ASYNC_RENDERING=0` for sync):

```bash
OVRTX_ASYNC_RENDERING=1 python scripts/benchmarks/training.py \
    --rl_library rsl_rl \
    --task Isaac-Lift-KukaAllegro-Camera \
    --seed 42 \
    --max_iterations 200 \
    --benchmark_formatter schema \
    --output_path ./results/kuka_<sync_or_async> \
    --headless \
    physics=newton_mjwarp renderer=ovrtx presets=rgb64
```

Enabling async adds one frame of camera latency.

## Type of change

## Screenshots

Not applicable — performance/API change with no visual UI. Async RGB output is validated against the sync goldens by the correctness tests above.

## Checklist


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -295,6 +306,29 @@ def __init__(self, cfg: OVRTXRendererCfg):
         self._camera_rel_path: str | None = None
         self._output_id_color_buffers: dict[str, wp.array] = {}
         self._clone_plan: ClonePlan | None = None
+        self._num_envs = 0
+        async_num_buffers = int(getattr(self.cfg, "async_num_buffers", 0))
+        # Test override: OVRTX_ASYNC_NUM_BUFFERS sets the render depth (and enables async when
+        # non-zero). Lets the existing golden-image tests validate the async path without editing cfgs.
+        _force_bufs = os.environ.get("OVRTX_ASYNC_NUM_BUFFERS")
+        if _force_bufs:
```
pv-nvidia: This is wrong, since setting OVRTX_ASYNC_NUM_BUFFERS=0 will have no effect

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -295,6 +306,29 @@ def __init__(self, cfg: OVRTXRendererCfg):
...
+            async_num_buffers = int(_force_bufs)
+        if async_num_buffers < 0:
+            raise ValueError(f"OVRTXRendererCfg.async_num_buffers must be >= 0, got {async_num_buffers}")
+        self._async_render = async_num_buffers >= 1
```
pv-nvidia: This should be a property, and be called `is_async_rendering_enabled`

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -295,6 +306,29 @@ def __init__(self, cfg: OVRTXRendererCfg):
...
+        self._async_num_buffers = async_num_buffers
+        # Transform writes always use two buffers, which is enough regardless of render depth.
+        # A DataAccess.ASYNC write copies the data into OVRTX's own storage, so OVRTX reads a buffer
+        # only until its write op completes (not until the render completes). A buffer is safe to
+        # recycle once its write has drained. Two buffers overlap "write this frame" with "last
+        # frame's write op still finishing"; there is only ever one write per frame, so two is also
+        # the most we could ever need, independent of how many renders (async_num_buffers) are in flight.
+        self._async_num_slots = 2
```
pv-nvidia: Doesn't make any sense to have an instance field for this

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -311,6 +345,72 @@ def __init__(self, cfg: OVRTXRendererCfg):
...
+        # OVRTX async binding writes with DataAccess.ASYNC copy the transform data into OVRTX's own
+        # storage; OVRTX reads the caller-owned buffer only until the returned write op completes.
+        # We wait on a buffer's previous write before reusing it (see _begin_async_slot_update ->
+        # _wait_for_slot_writes), so once that wait returns OVRTX has its own copy and overwriting the
+        # buffer cannot disturb an in-flight render. With one write per frame and two alternating
+        # buffers, a buffer's only outstanding work on reuse is the write from two frames ago, so two
+        # buffers are always enough regardless of render depth. OVRTX operations are stream-ordered,
+        # so a step_async() observes earlier write_async() calls issued for the same stage.
+        for _ in range(self._async_num_slots):
+            self._async_slots.append(
```
pv-nvidia: _async_slots is not cleared, so this method should only be called once? If so, it should have guard against this

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -311,6 +345,72 @@ def __init__(self, cfg: OVRTXRendererCfg):
...
+            return None
+        self._ensure_async_slots(num_envs)
+        slot = self._async_slots[self._async_slot_index]
+        self._async_slot_index = (self._async_slot_index + 1) % len(self._async_slots)
+        self._wait_for_slot_writes(slot)
+        self._async_current_slot = slot
+        return slot
+
+    def _get_async_update_slot(self, num_envs: int) -> _AsyncRenderSlot | None:
+        if not self._async_render:
```
pv-nvidia: If prefer ternary operator here

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -587,8 +689,9 @@ def update_transforms(self) -> None:
         if self._object_binding is None or self._object_newton_indices is None:
             return
 
-        # If self._object_newton_indices is not None, then Newton's the current physics backend
+        slot = self._begin_async_slot_update(self._num_envs) if self._async_render else None
 
+        # If self._object_newton_indices is not None, then Newton's the current physics backend
```
pv-nvidia, line 916: This comment should be moved up

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -599,14 +702,27 @@ def update_transforms(self) -> None:
         if body_q is None:
             return
 
-        with self._object_binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
-            ovrtx_transforms = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)
+        if slot is not None:
+            object_transforms = slot.object_transforms
```
pv-nvidia: why not also ovrtx_transforms = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)?

Investigate

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -616,26 +732,35 @@ def update_camera(
         intrinsics: ProxyArray,
     ) -> None:
         """Update camera transforms in OVRTX binding."""
+        self._async_last_render_data = render_data
         num_envs = positions.shape[0]
-        converted_wp = wp.empty(num_envs, dtype=wp.quatf, device=self._device)
+        slot = self._get_async_update_slot(num_envs) if self._async_render else None
+        if slot is not None:
```
pv-nvidia: If prefer `if slot is None` and swap branches.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -616,26 +732,35 @@ def update_camera(
...
             dim=num_envs,
             inputs=[positions, converted_wp, camera_transforms],
             device=self._device,
         )
         if self._camera_binding is not None:
-            with self._camera_binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
-                wp_transforms_view = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)
-                wp.copy(wp_transforms_view, camera_transforms)
+            if slot is not None:
+                self._write_binding_async(slot, self._camera_binding, camera_transforms)
```
pv-nvidia: If prefer `if slot is None` and swap branches.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -954,10 +1079,49 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
             return
+        self._async_last_render_data = render_data
+
+        try:
+            if self._async_render:
+                self._render_async(render_data)
+            else:
+                self._render_sync(render_data)
+        except Exception as e:
+            logger.warning("OVRTX rendering failed: %s", e, exc_info=True)
```
pv-nvidia: What a warning on exception? This exception swallowing wasn't present before

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -974,9 +1138,18 @@ def render(self, render_data: OVRTXRenderData) -> None:
                 render_data.warp_buffers[str(RenderBufferKind.RGBA)],
             )
 
+    def _drain_async_ring(self, render_data: OVRTXRenderData | None = None) -> None:
+        while self._async_ring:
+            try:
+                self._drain_one_async(render_data)
+            except Exception as e:
+                logger.warning("Error draining OVRTX async render op: %s", e, exc_info=True)
```
pv-nvidia: What swallow the exception?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -954,10 +1079,49 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
         products = self._renderer.step(
             render_products=set(self._render_product_paths),
             delta_time=1.0 / 60.0,
         )
+        self._consume_products(render_data, products)
+
+    def _render_async(self, render_data: OVRTXRenderData) -> None:
+        op = self._renderer.step_async(
+            render_products=set(self._render_product_paths),
+            delta_time=1.0 / 60.0,
```
pv-nvidia: Why is delta_time fixed to 1/60? It seems it was already like this, but feels weird

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -954,10 +1079,49 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+        )
+        self._async_ring.append({"op": op, "pending": None})
```
pv-nvidia: Why do we add a dictionary to the _async_ring instead of a property class or @dataclass? I prefer to typed _async_ring

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -295,6 +306,29 @@ def __init__(self, cfg: OVRTXRendererCfg):
...
+        # only until its write op completes (not until the render completes). A buffer is safe to
+        # recycle once its write has drained. Two buffers overlap "write this frame" with "last
+        # frame's write op still finishing"; there is only ever one write per frame, so two is also
+        # the most we could ever need, independent of how many renders (async_num_buffers) are in flight.
+        self._async_num_slots = 2
+        self._async_ring = deque()
+        self._async_slots: list[_AsyncRenderSlot] = []
+        self._async_slot_index = 0
+        self._async_current_slot: _AsyncRenderSlot | None = None
+        self._async_last_render_data: OVRTXRenderData | None = None
```
pv-nvidia: Suggestion: would it make sense to have a seperate AsyncRenderState class and move all self._async_XXX fields into it? And maybe this class could also have async-only related methods? Only when the code becomes easier to read and maintain of course

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Good catch. Fixed: the override now checks `_force_bufs is not None and _force_bufs != ""`, so `OVRTX_ASYNC_NUM_BUFFERS=0` is honored (forces synchronous) and only an unset/empty value is ignored.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done. Removed the `_async_render` field and added a `is_async_rendering_enabled` property that delegates to the async-state object.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Agreed. Folded into the new `_AsyncRenderState` class (`num_buffers` lives there now), so there is no longer a loose `_async_num_buffers` instance field on the renderer.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: The method already guards: it returns early when slots exist (`if self._slots: ... return`), and `reset_slots()` is what clears them (called on init and cleanup) so it can legitimately be re-created after a scene reset. It is now `_AsyncRenderState._ensure_slots` with the same guard.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: This one dissolved in the refactor: the `if not self._async_render: return None` early-out now lives inside `_AsyncRenderState.begin_slot_update` (guarding on `self.enabled`), and the renderer no longer branches on `_async_render` at the call site.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Moved. The "Newton is the current physics backend" comment now sits directly above the `NewtonManager` block it describes, and the slot acquisition happens after the Newton-state early-returns.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done \u2014 `update_transforms` now does `if slot is None:` (synchronous branch first), then the async branch.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done \u2014 `update_camera` now uses `if slot is None:` with the synchronous branch first.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: It does now \u2014 in the synchronous branch the object transforms are written straight into the mapped OVRTX buffer via `object_transforms = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)`, then the kernel writes into that view (no extra copy). The async branch is different by necessity: it writes into a reusable staging slot buffer and then issues `write_async`, because OVRTX needs its own copy while a previous frames render is still in flight. So sync = write in place into the mapped buffer; async = write into a staging buffer + async binding write.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Correct \u2014 that try/except was not in the original `render()` and shouldnt have been added. Removed it so exceptions propagate as before.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: This one is only on the teardown/reset drain (`_drain_async_ring`, called from `cleanup`/`reset_slots`). During teardown we want best-effort: a failure draining one in-flight op must not stop us draining the rest or tearing the renderer down, so we log and continue. The per-frame drain in `_render_async` does *not* swallow \u2014 it propagates. I added a comment making that split explicit. Happy to make it stricter if you prefer.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: This is pre-existing behavior (the sync path already passed `delta_time=1.0/60.0`), and OVRTX only uses `delta_time` for motion-blur / temporal effects, not for stepping physics (physics is stepped by Newton/sim separately). A fixed 1/60 is effectively a placeholder. I left it as-is to keep this PR scoped to async rendering, but I agree it is worth a follow-up to feed the real sim dt through \u2014 want me to open an issue for that?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done \u2014 the ring now holds a typed `_AsyncRenderOp` dataclass (`op`, `pending`) instead of a dict, and `_async_ring` is typed `deque[_AsyncRenderOp]`.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Did this \u2014 introduced an `_AsyncRenderState` class that owns the ring and the staging slots plus all the async-only methods (`begin_slot_update`, `get_update_slot`, `write_binding_async`, `reset_slots`, etc.). The renderer now just holds a single `self._async` and delegates, which also cleaned up several of the other comments above. Let me know if the split feels right.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -101,6 +95,137 @@
...
+            for op in write_ops:
+                op.wait()
+        except Exception as e:
+            raise RuntimeError("Failed to complete OVRTX async binding write before slot reuse") from e
+        finally:
+            slot.write_ops = []
+
+    def write_binding_async(self, slot: _AsyncRenderSlot, binding, data: wp.array) -> None:
+        """Issue an async binding write and record its op so the slot waits on it before reuse."""
+        if binding is None:
```
pv-nvidia: Can binding be None

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -389,6 +531,7 @@ def _initialize_from_spec(self, spec: CameraRenderSpec):
         width = spec.cfg.width
         height = spec.cfg.height
         num_envs = spec.num_instances
+        self._num_envs = num_envs
```
pv-nvidia: _num_envs  should be part of AsyncRenderState and should be passed to reset_slots.

When num_envs == 0, we are unbinding

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -805,26 +965,35 @@ def update_camera(
         intrinsics: ProxyArray,
     ) -> None:
         """Update camera transforms in OVRTX binding."""
+        self._async_last_render_data = render_data
```
pv-nvidia: _async_last_render_data  is not read, only written. Feels redundant?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1251,4 +1467,7 @@ def _safe_unbind(binding, name: str) -> None:
 
         self._render_product_paths.clear()
         self._output_id_color_buffers.clear()
+        self._async_last_render_data = None
```
pv-nvidia: _async_last_render_data can be removed

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1251,4 +1467,7 @@ def _safe_unbind(binding, name: str) -> None:
...
+        self._async.reset_slots()
+        self._num_envs = 0
```
pv-nvidia: _num_envs  can be moved into AsyncRenderState (or maybe removed completely)

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: No \u2014 both call sites already guard: update_transforms returns early when _object_xform_binding is None, and the camera write is inside if self._camera_xform_binding is not None. So binding is never None here. Removed the dead if binding is None guard.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done. num_envs now lives in AsyncRenderState: reset_slots(num_envs) records it (and num_envs == 0 = unbinding, slots just cleared), and begin_slot_update()/get_update_slot() take no args and build slots from the stored count. Dropped the renderer-side _num_envs field.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Right, it was write-only dead state. Removed it entirely (here, in render(), cleanup, and __init__).

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Removed \u2014 the field is gone everywhere.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Moved into AsyncRenderState. cleanup now just calls self._async.reset_slots(0); no more renderer _num_envs.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -101,6 +95,138 @@
...
+    def enabled(self) -> bool:
+        return self.num_buffers >= 1
+
+    def set_device(self, device: str) -> None:
+        """Update the CUDA device used for staging buffers. Must be called before slots are built."""
+        self._device = device
+
+    @property
+    def ring(self) -> deque[_AsyncRenderOp]:
+        return self._ring
```
pv-nvidia: This feels wrong, exposing the ring violates the encapsulation. We need higher level methods and keep the ring private.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1198,10 +1364,44 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+        self._consume_products(render_data, products)
+
+    def _render_async(self, render_data: OVRTXRenderData) -> None:
+        op = self._renderer.step_async(
+            render_products=set(self._render_product_paths),
+            delta_time=1.0 / 60.0,
+        )
+        ring = self._async.ring
+        ring.append(_AsyncRenderOp(op=op))
+        self._async.clear_current_slot()
```
pv-nvidia: These 3 lines should be a single higher level method in the AsyncRenderState, called enqueue_render_op or something, which could return the length of the ring.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1198,10 +1364,44 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+
+        if len(ring) >= self._async.num_buffers:
+            self._drain_one_async(render_data)
+
+    def _drain_one_async(self, render_data: OVRTXRenderData | None = None) -> None:
+        entry = self._async.ring.popleft()
```
pv-nvidia: Same here, should be _async.dequeue_render_op or something. Encapsulate the ring!. Should return None if the ring is empty.

Also the method should be called _try_drain_one_async, and should be safe to called when the ring is empty.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1198,10 +1364,44 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+        if entry.pending is None:
+            entry.pending = entry.op.wait()
+
+        products = entry.pending.fetch()
+
+        if render_data is not None:
+            self._consume_products(render_data, products)
+        entry.op = None
```
pv-nvidia: Modifying the entry should be done inside the AsyncRenderState, encapulsate!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1218,9 +1418,21 @@ def render(self, render_data: OVRTXRenderData) -> None:
                 render_data.warp_buffers[str(RenderBufferKind.RGBA)],
             )
 
+    def _drain_async_ring(self, render_data: OVRTXRenderData | None = None) -> None:
+        # Teardown/reset path: drain every queued render best-effort. A failure draining one op
+        # must not prevent us from draining the rest or from tearing the renderer down, so we log
+        # and continue rather than propagate. (The per-frame drain in _render_async does propagate.)
+        while self._async.ring:
```
pv-nvidia: This should be like `while self._try_drain_one_async` but exception safe. Ring should be encapsulated

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1218,9 +1418,21 @@ def render(self, render_data: OVRTXRenderData) -> None:
                 render_data.warp_buffers[str(RenderBufferKind.RGBA)],
             )
 
+    def _drain_async_ring(self, render_data: OVRTXRenderData | None = None) -> None:
```
pv-nvidia: _drain_async_ring is not a good name, ring is an internal detail

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Agreed. Removed the `ring` property; the deque is now fully private to AsyncRenderState. The renderer only calls higher-level methods (enqueue_render_op / try_dequeue_render_op / has_pending_ops).

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done \u2014 those three lines are now `depth = self._async.enqueue_render_op(op)`, which appends the op, ends the current frame slot, and returns the ring depth.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done. AsyncRenderState.try_dequeue_render_op() waits/fetches/pops the oldest op and returns its products, or None if the ring is empty. The renderer method is now _try_drain_one_async and is safe to call on an empty ring (returns False).

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Moved \u2014 the entry lifecycle (op.wait(), pending.fetch(), clearing op/pending) all happens inside try_dequeue_render_op now. The renderer just consumes the returned products.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Done \u2014 _drain_all_async now loops `while self._async.has_pending_ops(): try: self._try_drain_one_async(...) except: log+continue`. Exception-safe, ring not touched.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Renamed to _drain_all_async (ring is no longer part of the name or the interface).

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Opened the follow-up: https://github.com/isaac-sim/IsaacLab/issues/6545. Leaving the 1/60 as-is in this PR to keep it scoped to async rendering.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -43,15 +45,7 @@
 
 
 try:
-    from ovrtx import (
```
pv-nvidia, line 58: Why this change? Keep the original

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -101,6 +95,151 @@
...
+
+
+class _AsyncRenderState:
+    """Owns all state for OVRTX asynchronous rendering: the render ring and the transform
+    staging slots.
+
+    When async rendering is enabled (``num_buffers >= 1``), the renderer keeps ``num_buffers``
+    ``step_async`` calls in flight (the ring) and stages transform writes into a small set of
+    reusable buffers (the slots) so a new frame can be written while OVRTX is still reading the
+    previous one. When disabled, all methods are no-ops and callers fall back to the synchronous
```
pv-nvidia: This is an anti-pattern. We have an _AsyncRenderState that can do synchronous operations.

_AsyncRenderState should only ever be created when async rendering is enabled, otherwise it should be None

The exist code will be much cleaner when we do if self._async:

And _AsyncRenderState can be simplified

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -324,6 +473,11 @@ def __init__(self, cfg: OVRTXRendererCfg):
             )
         logger.info("OVRTX renderer created successfully")
 
+    @property
+    def is_async_rendering_enabled(self) -> bool:
+        """Whether asynchronous rendering is enabled (``async_num_buffers >= 1``)."""
+        return self._async.enabled
```
pv-nvidia: So this becomes self._async is not None

source/isaaclab/isaaclab/test/benchmark/stepping.py
```diff
@@ -59,22 +59,25 @@ def sample_random_actions(env) -> torch.Tensor | dict[str, torch.Tensor]:
         return 2.0 * torch.rand(u.num_envs, u.single_action_space.shape[0], device=u.device) - 1.0
 
 
-def run_runtime_loop(env, num_frames: int) -> list[float]:
-    """Step the environment ``num_frames`` times and record per-step wall times [s].
+def run_runtime_loop(env, num_frames: int, warmup_frames: int = 0) -> list[float]:
+    """Step the environment and record per-step wall times [s].
 
-    Calls ``env.reset()`` once before the loop, then on each frame samples
```
pv-nvidia: The original comment was erased, restore it, just include warmup frames

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1183,10 +1356,43 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
             delta_time=1.0 / 60.0,
         )
+        self._consume_products(render_data, products)
+
+    def _render_async(self, render_data: OVRTXRenderData) -> None:
+        op = self._renderer.step_async(
+            render_products=set(self._render_product_paths),
+            delta_time=1.0 / 60.0,
+        )
+        async_state = self._async
```
pv-nvidia: why cache into async_state?

and why raise RuntimeError? assert should be fine, as this is a private method

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1183,10 +1356,43 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+        if async_state is None:
+            raise RuntimeError("OVRTX async render state is not initialized")
+        depth = async_state.enqueue_render_op(op)
+        if depth >= async_state.num_buffers:
+            self._try_drain_one_async(render_data)
+
+    def _try_drain_one_async(self, render_data: OVRTXRenderData | None = None) -> bool:
+        """Complete the oldest queued render and consume it. Returns False when nothing was queued."""
+        if self._async is None:
```
pv-nvidia: Simplify with oneline assert

scripts/benchmarks/runtime.py
```diff
@@ -41,7 +41,17 @@ def _parse_args(argv: list[str]) -> tuple[argparse.Namespace, list[str]]:
     parser = argparse.ArgumentParser(description="Benchmark environment runtime (random actions, no policy).")
     parser.add_argument("--task", type=str, required=True, help="Gym task id to benchmark.")
     parser.add_argument("--num_envs", type=int, default=None, help="Number of parallel environments.")
-    parser.add_argument("--num_frames", type=int, default=100, help="Number of environment steps to benchmark.")
+    parser.add_argument("--num_frames", type=int, default=100, help="Number of timed environment steps to benchmark.")
+    parser.add_argument(
+        "--warmup_frames", type=int, default=0, help="Number of untimed environment steps to run first."
+    )
+    parser.add_argument(
+        "--ovrtx_render_latency_frames",
```
r-schmitt: can we rename this to clearly describe what it does?  `--ovrtx_enable_asyc_render` or something like that? Also lets make this a bool instead of an int and set a solid default (my guess would be True as default for this feature)

scripts/benchmarks/runtime.py
```diff
@@ -61,6 +71,29 @@ def _parse_args(argv: list[str]) -> tuple[argparse.Namespace, list[str]]:
...
+def _apply_ovrtx_render_latency_frames(env_cfg: object, latency_frames: int | None) -> int | None:
+    """Apply the OVRTX render-latency benchmark option and return the async buffer count."""
+    if latency_frames is None:
+        return None
+
+    try:
+        renderer_cfg = env_cfg.tiled_camera.renderer_cfg
+    except AttributeError as exc:
+        raise ValueError(
+            "--ovrtx_render_latency_frames requires a task with env_cfg.tiled_camera.renderer_cfg"
```
r-schmitt: do we want to raise a ValueError here? it seems like we could instead log a warning about the argument's compatibility instead and ignore the arg when its not applicable

scripts/benchmarks/runtime.py
```diff
@@ -61,6 +71,29 @@ def _parse_args(argv: list[str]) -> tuple[argparse.Namespace, list[str]]:
...
+        ) from exc
+
+    renderer_type = getattr(renderer_cfg, "renderer_type", None)
+    if renderer_type != "ovrtx":
+        raise ValueError(
```
r-schmitt: same as above, should we just log a warning and then ignore the argument instead of raising a ValueError?

r-schmitt (dismissed):

scripts/benchmarks/runtime.py
pv-nvidia: This isn't a bool but the number of frames of latency to introduce that allows async rendering, like in Omniverse MEGA, it should mimic the real sensor on the robot (and actually shoudn't be in frames but nanoseconds but that is much harder to implement I guess)

I see the choices are 0 or 1, that is not correct in that case, could be any positive number or zero.

But we could make this a `bool` if we just want to start with a single frame of latency yes.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_strategies.py
```diff
@@ -0,0 +1,365 @@
...
+        """Stage camera transforms into the current frame slot; publish them on exit.
+
+        See :meth:`_RenderStrategy.stage_camera_transforms`. Uses :meth:`_current_or_begin_slot` to join
+        the slot opened by the object update (or open one when there was no object update this frame).
+        The slot's camera buffers are pre-sized to ``num_envs`` in :meth:`_ensure_slots`; ``num_rows``
+        is expected to match.
+        """
+        slot = self._current_or_begin_slot()
+        yield slot.camera_quats, slot.camera_transforms
+        self._write_binding_async(slot, binding, slot.camera_transforms)
```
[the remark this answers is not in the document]
pv-nvidia: Fixed in 83da64bcf4 — added the size guard so `stage_camera_transforms` reallocates its slot buffers when `num_rows` diverges from the pre-sized `num_envs`, mirroring `stage_object_transforms`.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_strategies.py
```diff
@@ -0,0 +1,365 @@
...
+        """
+        op = renderer.step_async(render_products=render_products, delta_time=delta_time)
+        self._enqueue_render_op(op, render_data, consume_products)
+
+    def _try_drain_one(
+        self,
+        render_data: OVRTXRenderData | None,
+        consume_products: _RenderProductConsumer,
+    ) -> bool:
+        """Complete the oldest queued render and consume it. Returns False when no render products where queued."""
```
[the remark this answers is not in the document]
pv-nvidia: Fixed in 83da64bcf4 (where → were).

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_cfg.py
```diff
@@ -67,3 +67,18 @@ class OVRTXRendererCfg(RendererCfg):
...
+    When False (default), rendering is synchronous, preserving the existing blocking behavior. When
+    True, rendering is asynchronous: the renderer submits a render and returns immediately, overlapping
+    rendering with simulation and Python work to improve throughput, at the cost of camera outputs
+    arriving one frame later.
+
+    The render latency is fixed at one frame. Exposing user control over the number of frames to delay
+    is deferred to a future Isaac Lab-side configuration rather than a renderer-level knob.
+
+    For tests this can be overridden with the ``OVRTX_ASYNC_RENDERING`` environment variable
+    (``0``/``false``/``off`` disable, any other non-empty value enables).
```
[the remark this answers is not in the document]
pv-nvidia: Fixed in 83da64bcf4 — added `no` to the disable list in both this docstring and the matching one in `try_create`.

docs/source/testing/benchmarks.rst
```diff
@@ -126,6 +126,22 @@ Measure environment stepping performance without any RL library:
...
+.. code-block:: bash
+
+   uv run isaaclab benchmark runtime \
+       --task Isaac-Cartpole-Camera-Direct \
+       --num_envs 256 \
+       --warmup_frames 30 \
+       --num_frames 200 \
+       --benchmark_formatter schema,omniperf \
+       --output_path ./results/ovrtx_async \
+       --ovrtx_async_rendering \
```
pbarejko: I think setting up this through CLI could be considered an anti-pattern. This setting could be part of renderer cfg?

source/isaaclab/isaaclab/test/benchmark/stepping.py
```diff
@@ -59,22 +59,26 @@ def sample_random_actions(env) -> torch.Tensor | dict[str, torch.Tensor]:
...
-    random actions via :func:`sample_random_actions`, steps the environment,
-    and records the elapsed wall-clock time for that step.
+    Calls ``env.reset()`` once before the loop, runs ``warmup_frames`` untimed steps, then on each
```
pbarejko: Can you explain why do we have warmup frames?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -736,12 +749,14 @@ def update_transforms(self) -> None:
         if body_q is None:
             return
 
-        with self._object_xform_binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
-            ovrtx_transforms = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)
+        # The state stages the transforms (mapped OVRTX buffer when sync, a double-buffered slot when
+        # async) and publishes them to the binding when the context exits.
+        num_objects = len(self._object_newton_indices)
+        with self._strategy.stage_object_transforms(self._object_xform_binding, num_objects) as object_transforms:
```
pbarejko, line 931: How important for this work is to have object update kernels and rendering kernels on the same stream and have them properly enqueued? Can, in some situation, we have async render before we update the transforms?

scripts/benchmarks/runtime.py
```diff
@@ -61,6 +76,39 @@ def _parse_args(argv: list[str]) -> tuple[argparse.Namespace, list[str]]:
...
+
+    renderer_type = getattr(renderer_cfg, "renderer_type", None)
+    if renderer_type != "ovrtx":
+        logger.warning(
+            "--ovrtx_async_rendering ignored: requires renderer=ovrtx, resolved renderer_type=%r.",
+            renderer_type,
+        )
+        return None
+
+    renderer_cfg.async_rendering = async_rendering
```
pbarejko: We shouldn't be doing this. If we did we would end up with 1 million of CLI args.

docs/source/testing/benchmarks.rst
huidongc: Maybe try hydra config override like this in the CLI (the config path might not be valid):

```bash
    env.tiled_camera.renderer_cfg.async_rendering=True
```

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Nice catch, yes, we should pass the warp stream so that ovrtx doesn't read from the xform buffer while the kernel is still filling it. I hope that ovrtx will then use a cuda event to wait for the warp stream to complete, but that is not documented, I would need to read the source code (but it feels standard CUDA practice?)

docs/source/testing/benchmarks.rst
pv-nvidia: Yes, but I found out the hard way that this doesn't work for the Kuka task for some reason. So this is not a general solution.

huidongc: I see why you think it is not general, but actually it is quite general, you just need to adapt the hydra config path for different env. For Kuka task, it will be this path:

```bash
    env.scene.base_camera.renderer_cfg.async_rendering=True
```

Without the Hydra config solution, think about how many CLI args we would need for so many env cfgs.

huidongc: agents will figure out the config path for you.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_strategies.py
```diff
@@ -0,0 +1,437 @@
...
+
+    camera_transforms: wp.array
+    camera_quats: wp.array
+    object_transforms: wp.array | None
+    write_ops: list[Operation]
+
+    def record_write(self, binding: Any, data: wp.array, cuda_stream: int) -> None:
+        """Issue an async binding write and record its op so it can be drained before reuse.
+
+        ``cuda_stream`` is the Warp stream the staging kernel that filled ``data`` ran on. Handing it to
```
pv-nvidia, line 213: Although the ovrtx docs state:

>             cuda_stream: Optional CUDA stream handle (``int``) on which the input
                tensor's data is (or will be) ready. ovrtx synchronizes its read of
                the tensor against this stream and forwards it to the DLPack producer
                (e.g. Warp) so any pending work on a different stream is bridged
                automatically. If omitted, the caller must ensure the tensor's state
                is fully settled before calling.
                
We should double check if the implementation actually does this, as this API is deprecated in favor of ovstage

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -889,8 +926,7 @@ def _update_transforms_legacy(self) -> None:
         if self._object_xform_binding is None or self._object_newton_indices is None:
             return
 
-        # If self._object_newton_indices is not None, then Newton's the current physics backend
-
+        # A non-None self._object_newton_indices means Newton is the current physics backend.
```
pv-nvidia: This change was not needed

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -2140,27 +2188,15 @@ def _render_ovstage(self, render_data: OVRTXRenderData) -> None:
...
-                render_data.warp_buffers,
-            )
-
-        # Post-render PPISP: HDR scene-linear → LDR RGBA. Source/destination
-        # buffers are the same warp buffer map used by extraction.
-        if render_data.ppisp_pipeline is not None:
-            render_data.ppisp_pipeline.apply(
-                render_data.warp_buffers[str(RenderBufferKind.RGB_HDR)],
-                render_data.warp_buffers[str(RenderBufferKind.RGBA)],
-            )
```
pv-nvidia, line 2174: This code was deleted, where did it end up?

PR #6466, opened 2026-07-10
Add perf-smoke performance regression gate

Neil4561:
## Summary

Adds the **perf-smoke performance regression gate** on top of the infra probe (#6444), targeting `perf-smoke/infra-probe` so it runs in the same-repo context (self-hosted L40S, NGC secrets, write-capable token) needed to validate end-to-end before integrating into `develop`.

- Benchmarks a matrix of tasks/backends on the L40S pool and compares FPS against per-GPU **rolling baselines**, emitting verdicts as an omni-github artifact.
- Baselines live on an orphan `perf-baselines` branch, keyed by a **`runtime_contract_hash`** so samples are only pooled across identical environments.
- The pinned CI image is resolved via an **image-era manifest** (immutable `sha-<digest>`) instead of the floating `latest-develop` tag, so a baseline stays valid until the container environment actually changes.
- **Automated baseline upkeep:**
  - `perf-smoke-era-roll` republishes an immutable image and seeds it when the container era changes;
  - a **post-gate under-filled-bucket detector** reseeds exactly the tasks whose `runtime_contract_hash` window is short of `MIN_BASELINE_SAMPLES` after a protected `develop` push;
  - the seeder runs an **ancestry preflight** and `verify_baselines` confirms seeded samples are gate-usable.
- Credential hygiene: nvcr login temp config is wiped via `trap ... EXIT` so a failed pull/push never leaves the NGC key on the runner.

All gate tooling is self-contained under `tools/perf_smoke_test/` (no `isaaclab` import on the host side); the benchmark runs inside the era-pinned image.

## Test plan

- [x] `tools/perf_smoke_test` unit tests: 46 passed, 1 skipped (GPU-free)
- [x] pre-commit (ruff, ruff-format, license, codespell, yaml, shebang checks) clean
- [x] Reseed detector + seed→verify chain validated end-to-end on the RTX 6000 fork fleet (physx PASS consuming a reseeded baseline)
- [ ] Same-repo run on L40S after merge: confirm image pull, GPU passthrough, artifact upload, and `perf-baselines` write
- [ ] Seed L40S baselines and confirm a subsequent gate run PASSes using them

> Note: integrate `perf-smoke/infra-probe` into `develop` via a **merge commit** (not squash) so seeded baseline commit SHAs remain ancestors of `develop`; otherwise the post-gate detector will simply reseed on the first `develop` push.

PR #6490, opened 2026-07-13
Add reviewed CUDA 13 pip license exceptions

Neil4561:
## Summary

The `license-check` job pip-installs and audits the resolved dependency set at CI runtime. The environment now resolves the **CUDA 13** wheel stack, whose package names differ from the `cu12` variants currently whitelisted in `license-exceptions.json`. As a result, 18 NVIDIA-proprietary packages (plus a `typing_extensions` metadata-string change) are flagged and `license-check` fails.

This is **environment-wide dependency drift, not introduced by any feature branch** — any PR against this base hits the same failure. Splitting the fix out so it's cleanly attributed and reviewable on its own.

## Changes

`.github/workflows/license-exceptions.json` (additive):

- Added the newly surfaced CUDA 13 NVIDIA wheels as reviewed exceptions (bare `license: null`, matching the existing `*-cu12` convention):
  - unsuffixed: `nvidia-cublas`, `nvidia-cuda-cupti`, `nvidia-cuda-nvrtc`, `nvidia-cuda-runtime`, `nvidia-cufft`, `nvidia-cufile`, `nvidia-curand`, `nvidia-cusolver`, `nvidia-cusparse`, `nvidia-nvjitlink`, `nvidia-nvtx`
  - cu13-suffixed: `nvidia-cudnn-cu13`, `nvidia-cusparselt-cu13`, `nvidia-nccl-cu13`, `nvidia-nvshmem-cu13`
- Added `cuda-bindings` and `cuda-toolkit` (copied verbatim from upstream `develop`).
- Relaxed `typing_extensions` to `license: null`: newer wheels report `PSF-2.0` rather than `Python Software Foundation License`, which the inline exact-string exception no longer matches.

No change to the `license-check` workflow logic — data-only, additive. Does not weaken the check for any non-listed package.

## Test plan

- [x] JSON validated, no duplicate keys
- [x] Simulated the inline exact-match check against all 18 flagged packages with their reported licenses → 0 failures
- [ ] `license-check` CI passes on this PR

## Context

Unblocks `license-check` for the perf-smoke gate PR (#6466), which is otherwise red purely due to this drift.

### DIFF
--- a/.github/workflows/license-exceptions.json
+++ b/.github/workflows/license-exceptions.json
@@ -180,6 +180,76 @@
     "license": null
   },
   {
+    "package": "nvidia-cublas",
+    "license": null
+  },
+  {
+    "package": "nvidia-cuda-cupti",
+    "license": null
+  },
+  {
+    "package": "nvidia-cuda-nvrtc",
+    "license": null
+  },
+  {
+    "package": "nvidia-cuda-runtime",
+    "license": null
+  },
+  {
+    "package": "nvidia-cudnn-cu13",
+    "license": null
+  },
+  {
+    "package": "nvidia-cufft",
+    "license": null
+  },
+  {
+    "package": "nvidia-cufile",
+    "license": null
+  },
+  {
+    "package": "nvidia-curand",
+    "license": null
+  },
+  {
+    "package": "nvidia-cusolver",
+    "license": null
+  },
+  {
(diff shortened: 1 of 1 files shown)


Neil4561:
Folding this into the perf-smoke gate PR #6466 instead — same base branch and reviewer, so a single approval unblocks it. Superseded there.

PR #6499, opened 2026-07-13
Removes Isaac Sim core extensions from app file to avoid warp conflict

kellyguo11:
# Description

Isaac Sim requires different Warp versions as Isaac Lab. To avoid conflicts, we are removing Isaac Sim extensions from our .kit files to avoid loading in Isaac Sim extensions that would pull in Isaac Sim's warp version. This avoids conflicts between the warp versions and ensures that Isaac Lab's warp build gets used.

## Type of change

## Checklist


source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -551,18 +556,26 @@ def _resolve_indices_wp(self, indices: wp.array | None) -> wp.array:
     def _initialize_fabric(self) -> None:
         """One-time Fabric setup: hierarchy handle, attribute population, selections, indexed arrays."""
         import usdrt  # noqa: PLC0415
+
+        # The hierarchy bindings are a separate submodule and are not loaded by ``import usdrt``.
         from usdrt import Rt  # noqa: PLC0415
 
+        try:
```
ooctipus, line 563: why try catch? what is unknown?

source/isaaclab_physx/isaaclab_physx/__init__.py
```diff
@@ -90,4 +98,42 @@ def _patch_isaacsim_simulation_manager():
     original_module.IsaacEvents = IsaacEvents
 
 
+def _subscribe_to_simulation_manager_enable():
```
ooctipus: this function implementation reads a bit susy, is there a cleaner way to patch than using global, try catch, and handling all the fallback behavior for all None cases?

StafaH (approved):

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
kellyguo11: usdrt.hierarchy doesn't exist when we run headless, I think it only becomes available when we have hydra related extensions enabled for rendering. so when it's not available, we skip it.

tools/conftest.py
```diff
@@ -1318,33 +1317,6 @@ def pytest_sessionstart(session):
...
-                dev_match = re.search(r"cuda:\d+|\bcpu\b", bracket.group(1))
-                if dev_match:
-                    device = dev_match.group(0)
-            elapsed = float(case.time) if case.time is not None else 0.0
-            test_times.append((full_name, device, elapsed))
-    for full_name, device, elapsed in sorted(test_times, key=lambda row: row[2], reverse=True):
-        per_test_time_table.add_row([full_name, device, f"{elapsed:0.3f}"])
-
-    summary_str += per_test_time_table.get_string()
-
```
hujc7, line 1355: Is the concern about deleting those verbosity?

kellyguo11: ya it was kind of difficult to find the summary table. was this per test table important?

hujc7:
**Agent suggestion:** Follow-up context, nothing this PR needs to change — the subscription in `_simulation_manager_patch.py` is the right mechanism today, but one upstream change would let us delete most of it.

The machinery exists because the only off-switch, `enable_all_default_callbacks(False)`, is imperative: it can only run *after* the extension loads and registers its callbacks, and with lazy enabling that can happen at any time and repeatedly (reloads re-register). So Isaac Lab has to chase every enable event.

A carb setting inverts this: settings are populated from the `.kit` file before any extension loads, so they configure extensions regardless of when they load. The extension already reads `/exts/isaacsim.core.simulation_manager/default_engine`, so the ask is a sibling key checked at startup before registering the default callbacks:

```toml
# .kit [settings]
exts."isaacsim.core.simulation_manager".default_callbacks_enabled = false
```

With that, there is nothing to undo and no code that must run at the right moment — the enable-event subscription, the reload-recovery branch, and the `enable_all_default_callbacks(False)` call all become deletable. Only the symbol swap remains, and it drops from safety-critical to cosmetic (an un-swapped consumer meets an inert manager rather than a competing lifecycle owner).

**Jichuan:** Is that something we can ask?

tools/conftest.py
hujc7: The value was mostly in the head of the list — spotting slow or newly-regressed tests; printing every test is what buried the rest of the summary. Suggest trimming instead of deleting, so the signal survives:

```python
# slowest tests only — full timings remain in the JUnit XML
for full_name, device, elapsed in sorted(test_times, key=lambda row: row[2], reverse=True)[:30]:
    per_test_time_table.add_row([full_name, device, f"{elapsed:0.3f}"])
```

(plus a "Slowest 30 Tests" header). Alternatively `--durations=30` on the inner pytest command gives roughly the same for free, though per-file rather than aggregated across the merged report and without the device column.

apps/isaaclab.python.kit
```diff
@@ -12,42 +12,11 @@ keywords = ["experience", "app", "usd"]
...
-"isaacsim.robot.schema" = {}
-"isaacsim.robot.experimental.wheeled_robots" = { optional = true }
-"isaacsim.robot.wheeled_robots.nodes" = { optional = true }
-"isaacsim.sensors.experimental.physics" = {}
-"isaacsim.sensors.experimental.rtx" = {}
-"isaacsim.simulation_app" = {}
-"isaacsim.storage.native" = {}
-"isaacsim.util.debug_draw" = {}
+# USD schema plugins must load before the schema registry is first accessed.
+"omni.usd.schema.mujoco" = {}
```
hujc7, line 19: **Agent review:** `isaaclab.python.rendering.kit` layers on `"isaaclab.python"` and only adds `omni.kit.material.library`, so removing `omni.replicator.core` / `omni.syntheticdata` here also removes them from the GUI camera app. Unlike the headless case, the rendering kit doesn't re-declare them, and `isaac_rtx_renderer.py` imports both unconditionally. CI runs headless, so a break in `--enable_cameras` with the GUI wouldn't show up here.

Could you either confirm these still arrive transitively in the GUI camera path, or mirror the headless-rendering dependencies into `isaaclab.python.rendering.kit`? Same question for `isaaclab.python.xr.openxr.kit`, which layers on the same base.

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -2709,6 +2705,15 @@ def __call__(
 """
 
 
+def _enable_kit_extension(extension_name: str) -> None:
```
hujc7: Nice — this helper is exactly the right replacement. Suggest promoting it from module-local to `isaaclab.sim.utils` and finishing the sweep with it, since the same snippet now exists in six places (`urdf_converter.py`, `mjcf_converter.py`, `mesh_converter.py`, `controllers/utils.py`, `surface_gripper.py`, and here), and a repo grep still shows unguarded `isaacsim.core.experimental.utils` imports that would raise `ModuleNotFoundError` under the trimmed apps: `resolve_rmpflow_path` in `controllers/utils.py` (uses `get_extension_path`) and `isaaclab_tasks/contrib/stack/mdp/stack_events.py:324`.

```python
# isaaclab/sim/utils/... (public, next to has_kit())
def enable_extension(ext_name: str) -> None:
    """Enable a Kit extension immediately if it is not already enabled."""
    import omni.kit.app  # noqa: PLC0415

    manager = omni.kit.app.get_app().get_extension_manager()
    if not manager.is_extension_enabled(ext_name):
        manager.set_extension_enabled_immediate(ext_name, True)


def get_extension_path(ext_name: str) -> str:
    """Return the install path of an enabled Kit extension."""
    import omni.kit.app  # noqa: PLC0415

    manager = omni.kit.app.get_app().get_extension_manager()
    return manager.get_extension_path(manager.get_enabled_extension_id(ext_name))
```

Call-site migration is then mechanical, e.g. in `stack_events.py`:

```python
# before
from isaacsim.core.experimental.utils.app import enable_extension
enable_extension("omni.replicator.core")

# after
from isaaclab.sim.utils import enable_extension
enable_extension("omni.replicator.core")
```

Done-criterion: a grep for `isaacsim.core.experimental.utils` over non-deprecated source returns only guarded hits.

.github/workflows/wheel.yml
```diff
@@ -192,6 +197,7 @@ jobs:
 
           uv pip install \
             --dry-run \
+            --overrides "$overrides" \
```
hujc7, line 199: **Agent review:** Passing `--overrides` makes the dry-run pass, but it also means this check no longer validates what end users run: the documented `uv pip install isaaclab[isaacsim]` has no overrides file, and with `isaacsim==6.0.0.1` pinning `typing-extensions==4.12.2` against the new `>=4.15.0` floor, that resolution should now fail. `gen_uv_overrides.py` gives the file a single source of truth, but nothing ships or documents it for users.

Could the install docs cover the overrides file (or the wheel ship it), e.g.:

```bash
curl -LO https://raw.githubusercontent.com/isaac-sim/IsaacLab/main/tools/wheel_builder/overrides.txt
uv pip install --overrides overrides.txt "isaaclab[isaacsim,all]"
```

and install-CI gain a case that resolves the `isaacsim` extra the way a user would? If users don't actually hit the conflict, a note in the PR description explaining why the dry-run needs the override but users don't would settle it.

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1241,6 +1241,10 @@ def _create_app(self):
                 indexes_to_remove.append(idx)
             if arg == "--capture=no":
                 indexes_to_remove.append(idx)
+            # Kit also interprets pytest's -v/-vv flags, enabling verbose startup
+            # logs that dump configuration tokens and every registered extension.
+            if re.fullmatch(r"-v+", arg):
```
hujc7: **Agent review:** The enclosing loop iterates `sys.argv[:-1]` (needed for the `-m` value lookahead), so a trailing `-v` — the most common manual form, `./isaaclab.sh -p -m pytest path/to/test.py -v` — escapes removal and Kit still goes verbose. Suggest keeping the lookahead bound only for the `-m` pair and scanning the full argv for single-token flags:

```python
for idx, arg in enumerate(sys.argv):
    if arg == "-m" and idx + 1 < len(sys.argv):
        ...  # existing pair handling
    if arg.startswith("--config-file=") and "pyproject.toml" in arg:
        indexes_to_remove.append(idx)
    if arg == "--capture=no" or re.fullmatch(r"-v+", arg):
        indexes_to_remove.append(idx)
```

Might also be worth gating the `-v` removal on pytest being detected (e.g. `"pytest" in sys.modules`), since `-v` is a common flag a standalone script could legitimately parse after app launch.

source/isaaclab_physx/isaaclab_physx/physics/physx_manager.py
```diff
@@ -286,6 +286,13 @@ def initialize(cls, sim_context: SimulationContext) -> None:
         """Initialize the physics manager."""
         from isaaclab.sim.utils.stage import get_current_stage_id
 
+        from isaaclab_physx import _simulation_manager_patch
+
+        # Optional Isaac Sim extensions can load their SimulationManager after
+        # isaaclab_physx. Subscribe once Kit is available so its callbacks are
+        # disabled before they can compete with this manager for tensor views.
+        _simulation_manager_patch.subscribe()
```
hujc7: Nit on API shape, two parts.

Naming: `_simulation_manager_patch.subscribe()` names the mechanism (which Isaac Sim class gets patched, via which Kit facility) rather than the guarantee it provides. A reader of `initialize()` can't tell what this line ensures without opening the other module. Since the package already scopes it to PhysX, suggest naming the entry point for the intent — e.g. `claim_physics_lifecycle()` (module could stay as-is or become `_lifecycle_ownership.py`).

Structure: the module exposes two entry points whose pairing is implicit — `__init__.py` calls patch-then-subscribe at import, this site calls subscribe-only, and the "extension enabled between a pre-Kit import and `initialize()`" case relies on Kit firing the enable-hook retroactively for already-enabled extensions. Suggest collapsing both behind one idempotent entry point, with `__init__.py` updated to use it too:

```python
# _simulation_manager_patch.py
def claim_physics_lifecycle() -> None:
    """Make PhysxManager the sole owner of the PhysX simulation lifecycle.

    Idempotent; safe pre-Kit (no-op), post-Kit (arms the watcher), or after
    Isaac Sim's simulation manager has already loaded (patches immediately).
    """
    _neutralize_if_loaded()   # existing patch logic; no-op when module absent
    _arm_enable_hook()        # existing subscription; no-op when Kit missing/armed
```

```python
# __init__.py                              # covers: Kit already running at import
claim_physics_lifecycle()

# physics/physx_manager.py, initialize()   # covers: imported before Kit launched
_simulation_manager_patch.claim_physics_lifecycle()
```

Same behavior, but callers stop needing to know the timing story, and the retroactive-hook assumption becomes explicit, testable code. (See the general comment on the PR for a longer-term declarative option.)

apps/isaaclab.python.kit
```diff
@@ -91,6 +61,8 @@ keywords = ["experience", "app", "usd"]
 "omni.kit.window.status_bar" = {}
 "omni.kit.window.toolbar" = {}
 "omni.physics.stageupdate" = {}
+"omni.physx" = {}
+"omni.physx.tensors" = {}
```
hujc7, line 66: The headless kit lists `omni.physx.fabric` alongside these two, but it's absent here after dropping `omni.physx.bundle`. Is it pulled transitively for the GUI app, or intentionally omitted? If PhysX-to-Fabric transform sync matters for the GUI render path, worth adding explicitly for symmetry with headless.

Also flagging: dropping `isaacsim.core.deprecation_manager` removes the `omni.isaac.*` → `isaacsim.*` import aliasing. Isaac Lab itself doesn't need it, but downstream user tasks with legacy imports may — if intentional, a changelog note with migration guidance would help.

tools/conftest.py
hujc7: I think the agent suggested just heavy ones could also work. It's mostly for information in case CI regress and we need to find what's consuming all the time.

hujc7 (approved):
Thanks!!

PR #6504, opened 2026-07-14
Fix Newton cfgs cancelling inherited PhysX field routing

vidurv-nvidia:
# Description

Seven Newton/MuJoCo schema cfg classes redeclare `_usd_field_exceptions: ClassVar[dict] = {}`. Python attribute lookup stops at the first class in the MRO that defines the attribute — even when the value is empty — so the redeclaration **shadows the base classes' routing table** for PhysX-consumed fields those classes inherit (`disable_gravity`, `contact_offset`, `rest_offset`, `max_joint_velocity`).

Observed behavior on develop (verified by executing the writer per class/field):

| Cfg | Field set | Actual | Intended |
|---|---|---|---|
| `NewtonRigidBodyPropertiesCfg`, `MujocoRigidBodyPropertiesCfg` | `disable_gravity` | bare `physics:disableGravity` (no consumer) | `physxRigidBody:disableGravity` |
| `NewtonCollisionPropertiesCfg` + Mesh/SDF subclasses | `contact_offset`/`rest_offset` | bare `physics:*` (no consumer) | `physxCollision:*` |
| `NewtonJointDrivePropertiesCfg`, `MujocoJointDrivePropertiesCfg` | `max_joint_velocity` | **raises `ValueError`** | `physxJoint:maxJointVelocity` |

That the shadowing is accidental is shown by `NewtonArticulationRootPropertiesCfg`, which does **not** redeclare the attribute and routes its inherited `articulation_enabled` to `physxArticulation:*` correctly — one sibling inherited the pattern right, six copied boilerplate that cancels it.

## Fix

Remove the empty redeclarations from the seven classes so the base routing applies, with a comment at each site explaining why the attribute must stay inherited. No writer changes.

## Tests

Six regression tests drive each affected field through the public writers (`define_rigid_body_properties`, `define_collision_properties`, `modify_joint_drive_properties`) and assert the PhysX-namespaced attribute is authored (with the rad/s→deg conversion for the joint velocity) and no bare `physics:*` attribute appears. All fail before the fix (two with the `ValueError`), pass after. `test_newton_schemas.py` 25/26 and core `test_schemas.py` 40/40 locally — the one unrelated failure (`test_newton_material_fragment_authors_all_six_newton_attrs`) also fails on clean develop in this environment and appears to be resolver-key drift from the recent Newton bump; flagged separately.

## Type of change

## Checklist

- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/`

### DIFF
--- a/source/isaaclab_newton/changelog.d/vidurv-exception-shadowing.rst
+++ b/source/isaaclab_newton/changelog.d/vidurv-exception-shadowing.rst
@@ -1,0 +1,13 @@
+Fixed
+^^^^^
+
+* Fixed the Newton and MuJoCo schema cfg classes cancelling the PhysX routing of fields they
+  inherit from the solver-common base classes. Redeclaring ``_usd_field_exceptions`` as an empty
+  dict shadowed the base class's routing table, so setting ``disable_gravity``,
+  ``contact_offset``, or ``rest_offset`` on a Newton or MuJoCo cfg authored a bare ``physics:*``
+  attribute that no backend reads, and setting ``max_joint_velocity`` on
+  :class:`~isaaclab_newton.sim.schemas.NewtonJointDrivePropertiesCfg` or
+  :class:`~isaaclab_newton.sim.schemas.MujocoJointDrivePropertiesCfg` raised ``ValueError``.
+  These fields now route to their PhysX namespaces on every subclass, matching the base classes
+  and :class:`~isaaclab_newton.sim.schemas.NewtonArticulationRootPropertiesCfg`, which already
+  inherited the routing correctly.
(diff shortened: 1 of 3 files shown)


vidurv-nvidia:
Superseded by #7832, which re-derives this fix against current `develop` — this branch predates substantial changes to `schemas_cfg.py` and no longer applies cleanly.

Confirmed the bug still reproduces before closing: the empty `_usd_field_exceptions` redeclarations shadow the base routing table, so `disable_gravity` authors `physics:disableGravity` (which no backend reads) instead of `physxRigidBody:disableGravity`. #7832 fixes the 7 classes that shadow a non-empty base, with six regression tests verified to fail without the fix.

PR #6506, opened 2026-07-14
Add segmentation annotator support to Newton Warp renderer integration

rilei-nvidia:
# Description

Bring the segmentation annotators (`semantic_segmentation`,  `instance_segmentation_fast`) in the Newton Warp renderer integration to full parity with what the Isaac RTX renderer provides.  This includes producing the idToLabels and/or idToSemantics dicts for mapping segmented IDs/colors to their respective prim path or semantic label.

The main difference between supporting segmentation annotators for Newton warp vs ovrtx is that the Newton model only tracks the prim paths for each shape but not their semantic labels.   So via the NewtonModel.shape_label array and its USD stage, we build the mappings from shape_index -> semantics upfront.  Then there are warp kernels that process the shape index image into either an ID map or colorized map (see new golden images).  All this logic is encapsulated in the new `isaaclab_newton/renderers/segmentation.py` module.

Light refactoring was done to share common variables / logic for around segmentation and colorization between the ovrtx and newton renderer integrations.

## Type of change

- Documentation update

## Screenshots

See golden images

## Checklist

### DIFF
--- a/source/isaaclab/test/renderers/test_segmentation_colors.py
+++ b/source/isaaclab/test/renderers/test_segmentation_colors.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Tests for the shared segmentation colorization (host vs Warp device parity)."""
+
+import pytest
+
+pytest.importorskip("warp")
+
+import warp as wp
+
+from isaaclab.renderers.segmentation_colors import (
+    BACKGROUND_ID,
+    UNLABELLED_ID,
+    color_hash,
+    pack_rgba,
+    random_color_from_id,
+    random_color_from_id_wp,
+    unpack_rgba,
+)
+
+
+@wp.kernel
+def _colorize_ids_kernel(ids: wp.array(dtype=wp.uint32), out: wp.array(dtype=wp.uint32)):
+    i = wp.tid()
+    out[i] = random_color_from_id_wp(ids[i])
+
+
+def test_reserved_colors():
+    """BACKGROUND is transparent black, UNLABELLED is opaque black."""
+    assert random_color_from_id(BACKGROUND_ID) == (0, 0, 0, 0)
+    assert random_color_from_id(UNLABELLED_ID) == (0, 0, 0, 255)
+    assert pack_rgba(random_color_from_id(BACKGROUND_ID)) == 0
+    assert pack_rgba(random_color_from_id(UNLABELLED_ID)) == 0xFF000000
+
+
+def test_pack_rgba_is_little_endian():
+    assert pack_rgba((1, 2, 3, 4)) == 1 | (2 << 8) | (3 << 16) | (4 << 24)
+
+
+def test_unpack_rgba_is_inverse_of_pack():
+    for color in [(0, 0, 0, 0), (255, 255, 255, 255), (1, 2, 3, 4), (0, 0, 0, 255)]:
+        assert unpack_rgba(pack_rgba(color)) == color
+
+
+def test_color_hash_is_uint32():
+    for seed in (0, 1, 2, 12345, 0xFFFFFFFF):
+        h = color_hash(seed)
+        assert 0 <= h <= 0xFFFFFFFF
+
+
+@pytest.mark.parametrize("device", ["cpu"])
+def test_host_and_device_random_color_agree(device):
+    """The host and Warp implementations must produce byte-identical colors for every id."""
+    ids = list(range(256)) + [1000, 65535, 123456]
+    ids_wp = wp.array(ids, dtype=wp.uint32, device=device)
+    out_wp = wp.zeros(len(ids), dtype=wp.uint32, device=device)
+    wp.launch(_colorize_ids_kernel, dim=len(ids), inputs=[ids_wp], outputs=[out_wp], device=device)
+
+    device_colors = out_wp.numpy().tolist()
+    host_colors = [pack_rgba(random_color_from_id(i)) for i in ids]
+    assert device_colors == host_colors
--- a/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
+++ b/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
@@ -22,6 +22,7 @@ from isaaclab.utils.warp.warp_math import convert_camera_frame_orientation_conve
 
 from ..physics.newton_manager import NewtonManager
 from .newton_warp_renderer_cfg import NewtonWarpRendererCfg
+from .segmentation import NewtonSegmentationMapper, NewtonSegmentationMapping
 
 if TYPE_CHECKING:
     from isaaclab_ppisp import PpispPipeline
@@ -57,12 +58,16 @@ class RenderData:
     # The depth family (``distance_to_camera`` / ``distance_to_image_plane`` / ``depth``) is handled
     # separately in :meth:`set_outputs` rather than through this map, because Newton emits a single
     # ray-hit-distance buffer that must be reused as the source for the planar-depth conversion.
+    #
+    # The segmentation family (``semantic_segmentation`` / ``instance_segmentation_fast``) is likewise
+    # handled separately: Newton emits a single per-shape index buffer that is remapped into each
+    # requested segmentation output by
+    # :class:`~isaaclab_newton.renderers.segmentation.NewtonSegmentationMapper`.
     _OUTPUT_MAP: dict[str, tuple[str, type]] = {
         str(RenderBufferKind.RGBA): ("color_image", wp.uint32),
         str(RenderBufferKind.RGB_HDR): ("hdr_color_image", wp.vec3f),
         str(RenderBufferKind.ALBEDO): ("albedo_image", wp.uint32),
         str(RenderBufferKind.NORMALS): ("normals_image", wp.vec3f),
-        str(RenderBufferKind.INSTANCE_SEGMENTATION_FAST): ("instance_segmentation_image", wp.uint32),
     }
 
     # Newton's native ``depth_image`` is the ray-hit (euclidean) distance from the camera optical
@@ -87,10 +92,20 @@ class RenderData:
         # ``distance_to_camera`` output or to an internal scratch buffer (see :meth:`set_outputs`).
         depth_image: wp.array(dtype=wp.float32, ndim=4) = None
         normals_image: wp.array(dtype=wp.vec3f, ndim=4) = None
-        instance_segmentation_image: wp.array(dtype=wp.uint32, ndim=4) = None
+        # Buffer Newton fills with the per-pixel shape index; the source for all segmentation outputs.
+        shape_index_image: wp.array(dtype=wp.uint32, ndim=4) = None
 
-    def __init__(self, newton_sensor: newton.sensors.SensorTiledCamera, spec: CameraRenderSpec):
+    def __init__(
+        self,
+        newton_sensor: newton.sensors.SensorTiledCamera,
+        spec: CameraRenderSpec,
+        seg_mapper: NewtonSegmentationMapper | None = None,
+        renderer_cfg: NewtonWarpRendererCfg | None = None,
+    ):
         self.newton_sensor = newton_sensor
+        # Shared, scene-static segmentation lookup builder (``None`` until segmentation is requested).
+        self._seg_mapper = seg_mapper
+        self._renderer_cfg = renderer_cfg
 
         self.num_cameras = 1
 
@@ -107,6 +122,9 @@ class RenderData:
         # Internal ray-depth buffer allocated only when a planar-depth output is requested without
         # ``distance_to_camera``; gives ``convert_ray_depth_to_forward_depth`` a source to read from.
         self._ray_depth_scratch: wp.array | None = None
+        # Requested segmentation outputs keyed by data-type name -> (destination view, mapping). Each view
+        # aliases the caller's output buffer as ``(world_count, 1, H, W)`` uint32.
+        self._seg_dests: dict[str, tuple[wp.array, NewtonSegmentationMapping]] = {}
         self.width = getattr(spec.cfg, "width", 100)
         self.height = getattr(spec.cfg, "height", 100)
         # Camera clipping planes [m] from ``spawn.clipping_range`` (``[0]`` near, ``[1]`` far).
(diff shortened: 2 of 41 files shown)


source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -383,7 +456,11 @@ def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
...
         owns the sentinel-resolution + cfg-normalization step. Newton has no
         USD-side overrides to author beyond this.
+
+        Also captures the USD ``stage`` so the segmentation mapper can read the scene's
+        :class:`UsdSemantics.LabelsAPI` labels when a segmentation output is requested.
         """
+        self._stage = stage
         if spec.cfg.isp_cfg is None:
             return
         try:
```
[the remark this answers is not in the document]
rilei-nvidia: It's initialized in NewtonWarpRenderer.initialize

source/isaaclab_tasks/test/golden_images/franka_cloth/physx-newton_renderer-instance_segmentation_fast.png
huidongc, line 1: physx sim backend is not really supported by the franka cloth env. It just falllbacks to the default - newton.

In another [PR](https://github.com/isaac-sim/IsaacLab/pull/6563), I am removing all these physx-prefixed images for franka cloth, to reduce confusion.

Same to the next image `source/isaaclab_tasks/test/golden_images/franka_cloth/physx-newton_renderer-semantic_segmentation.png`

rilei-nvidia: good call out.  Will remove these thanks!

rilei-nvidia: rebased ontop of develop which includes your PR and deleted the franka_cloth physx-newton_renderer golden images

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -402,7 +479,18 @@ def prepare_stage(self, stage: Any, num_envs: int) -> None:
...
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.create_render_data`."""
-        return RenderData(self.newton_sensor, spec)
+
+        # Build the shared segmentation mapper and its per-kind lookup tables up-front for all
+        # requested segmentation outputs.
+        for dt in spec.cfg.data_types:
+            if dt in RenderData._SEG_COLORIZE_ATTR:
+                if self._seg_mapper is None:
+                    self._seg_mapper = NewtonSegmentationMapper(self._newton_model, self._stage, self.cfg)
+                colorize = bool(getattr(self.cfg, RenderData._SEG_COLORIZE_ATTR[dt]))
```
daniela-hase: I find this block of code fairly hard to read and wonder if it would be simpler to do something like:

```python
if RenderBufferKind.SEMANTIC_SEGMENTATION in spec.cfg.data_types:
    self._seq_mapper.build_mapping(RenderBufferKind.SEMANTIC_SEGMENTATION, self.cfg.colorize_semantic_segmentation)

if RenderBufferKind.INSTANCE_SEGMENTATION_FAST in spec.cfg.data_types:
    self._seq_mapper.build_mapping(RenderBufferKind.INSTANCE_SEGMENTATION_FAST, self.cfg.colorize_instance_segmentation)
```

And the same further above, basically get rid of the `RenderData._SEG_COLORIZE_ATTR` altogether. What do you think?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -174,6 +218,11 @@ def set_outputs(self, output_data: dict[str, ProxyArray]):
             self.outputs.depth_image = self._ray_depth_scratch
         else:
             self.outputs.depth_image = None
+        # Allocate the shape-index buffer Newton fills when any segmentation output is requested; all
+        # requested segmentation outputs are remapped from this single buffer in :meth:`convert_segmentation`.
+        if self._seg_dests:
+            self._shape_index_scratch = wp.zeros(shape, dtype=wp.uint32, device=self.newton_sensor.model.device)
+        self.outputs.shape_index_image = self._shape_index_scratch
```
daniela-hase: What's the point of creating `self._shape_index_scratch`? Why not use `self.outputs.shape_index_image` directly?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -488,6 +576,9 @@ def render(self, render_data: RenderData):
             render_data.convert_plane_depth()
             render_data.apply_depth_clipping(self.cfg.depth_clipping_behavior)
 
+        # Remap the shape-index buffer into the requested segmentation outputs.
+        render_data.convert_segmentation()
```
daniela-hase: I realize we did the same with the `convert_plane_depth` and `apply_depth_clipping`, but I personally am not a fan of having public functions that *need* to be called in a specific order for things to work.

So I would either turn them into private functions prefix them with `_` or maybe even move it to be a private method of the `NewtonWarpRenderer` class, not the `RenderData`.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -57,12 +58,23 @@ class RenderData:
...
         str(RenderBufferKind.ALBEDO): ("albedo_image", wp.uint32),
         str(RenderBufferKind.NORMALS): ("normals_image", wp.vec3f),
-        str(RenderBufferKind.INSTANCE_SEGMENTATION_FAST): ("instance_segmentation_image", wp.uint32),
+    }
+
+    # Segmentation outputs, and the renderer cfg flag selecting colorized (uint8 RGBA) vs raw (uint32)
+    # layout for each. Both are derived from Newton's single ``shape_index_image`` buffer.
+    _SEG_COLORIZE_ATTR: dict[str, str] = {
+        str(RenderBufferKind.SEMANTIC_SEGMENTATION): "colorize_semantic_segmentation",
+        str(RenderBufferKind.INSTANCE_SEGMENTATION_FAST): "colorize_instance_segmentation",
```
daniela-hase: Unrelated to this PR I guess, but why do we call it `_FAST`? Do we have a slow version of this as well? 🤨

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: I've made these private (including the depth ones)!  Did not move to NewtonWarpRenderer class as those methods themselves were further accessing `RenderData` internal members

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: Good catch.   It was redundant.  Removed!

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: Good call out, simplified and removed the `RenderData._SEG_COLORIZE_ATTR` dict, it was confusing to read.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: Per discussion in rendering sync today we will name this semantic_instance_segmentation :-)

daniela-hase (approved):

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
```diff
@@ -0,0 +1,505 @@
...
+    if isinstance(semantic_filter, list):
+        text = "; ".join(f"{t}:*" for t in semantic_filter)
+    else:
+        text = semantic_filter
+    clauses: list[tuple[str, str]] = []
+    for clause in text.split(";"):
+        clause = clause.strip()
+        if not clause:
+            continue
+        sem_type, _, label_expr = clause.partition(":")
```
StafaH: **Major:** Parse comma-separated semantic-type clauses correctly

The documented RTX predicate grammar uses commas to separate additional `(semantic type, label expression)` pairs inside a semicolon group (for example, `typeA:labelA, typeB:labelB`). This parser only splits on `;` and partitions once, while `_eval_label_expr` later treats the comma as a label-level OR. As a result, the second type is never evaluated and its Newton pixels become UNLABELLED for a valid filter. Implement the same DNF parser as the canonical RTX predicate grammar (or share a parser) and add a test covering the documented multi-type comma form.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -362,9 +426,11 @@ def initialize(self) -> None:
     def supported_output_types(self) -> dict[RenderBufferKind, RenderBufferSpec]:
         """Publish the per-output layout this Newton Warp backend writes.
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.supported_output_types`."""
-        seg_spec = (
-            RenderBufferSpec(4, wp.uint8) if self.cfg.colorize_instance_segmentation else RenderBufferSpec(1, wp.int32)
-        )
+
+        def seg_spec(colorize: bool) -> RenderBufferSpec:
+            # Colorized segmentation is RGBA uint8; raw segmentation is a single uint32 id channel.
+            return RenderBufferSpec(4, wp.uint8) if colorize else RenderBufferSpec(1, wp.uint32)
```
StafaH: **Major:** Preserve the int32 raw-segmentation contract

Every existing camera backend advertises non-colorized segmentation as one-channel `wp.int32`, and the new Newton config docstrings also promise `int32`. Returning `wp.uint32` makes backend-independent consumers observe a different dtype when switching to Newton. Keep the public `RenderBufferSpec` and destination output as `wp.int32` (the private Newton shape-index scratch can remain `uint32`) and update the new contract test, whose docstring currently says int32 while asserting uint32.

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
```diff
@@ -0,0 +1,505 @@
...
+                BACKGROUND_ID: self._reserved_semantics_label("BACKGROUND"),
+                UNLABELLED_ID: self._reserved_semantics_label("UNLABELLED"),
+            }
+
+        all_labels = {**reserved_labels, **id_labels}
+
+        def key_for(seg_id: SegId) -> SegId | tuple:
+            if colorize:
+                return self._id_to_color(seg_id, id_labels.get(seg_id, all_labels[seg_id]))
+            return seg_id
```
StafaH, line 508: **Major:** Match raw idToLabels keys across renderers

RTX/OVRTX expose raw segmentation metadata with decimal string keys (`"0"`, `"1"`, ...), because the mapping originates from the annotator's JSON-compatible contract; colorized outputs use RGBA tuple keys. Newton returns integer keys here, so backend-independent code that indexes `idToLabels[str(pixel_id)]` or `idToSemantics[str(pixel_id)]` fails only on Newton despite the claimed parity. Return `str(seg_id)` when `colorize` is false and cover both semantic and instance metadata in the parity tests.

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
```diff
@@ -0,0 +1,505 @@
...
+        if prim_path in self._matched_cache:
+            return self._matched_cache[prim_path]
+
+        result: tuple[dict[SemanticType, SemanticLabels], SemanticPrimPath] | None = None
+        if self._stage is not None and prim_path.startswith("/"):
+            from isaaclab.sim.utils.semantics import get_labels  # noqa: PLC0415
+
+            prim = self._stage.GetPrimAtPath(prim_path)
+            while prim is not None and prim.IsValid() and prim.GetPath().pathString != "/":
+                labels = get_labels(prim)
```
StafaH, line 332: **Minor:** Consult the ancestor cache while walking the hierarchy

`_matched_cache` is checked only for the original leaf path. Although a matched ancestor is stored, the loop never checks that cached ancestor, so siblings still repeat the USD label query and hierarchy traversal that the cache is meant to avoid. A 100-shape scene under one labelled root performed 300 `get_labels` calls, including querying the same root 100 times. Check cached resolutions for each visited ancestor (before calling `get_labels`) or memoize the traversed chain so large cloned scenes do not pay redundant USD work during camera creation.

source/isaaclab/test/renderers/test_segmentation_colors.py
```diff
@@ -0,0 +1,111 @@
...
+    assert device_colors == host_colors
+
+
+@pytest.mark.gpu
+class TestGenerateRandomColorsFromIdsKernel:
+    """generate_random_colors_from_ids_kernel agrees with the shared segmentation_colors reference."""
+
+    @pytest.fixture(autouse=True)
+    def _import_kernel(self):
+        from isaaclab_ov.renderers.ovrtx_renderer_kernels import generate_random_colors_from_ids_kernel
```
StafaH: **Minor:** Keep OVRTX CUDA tests in the OVRTX package

This core-package test now imports the downstream optional `isaaclab_ov` package and hardcodes `cuda:0`. The class-level `gpu` marker is not registered by the root pytest configuration (the focused run emits `PytestUnknownMarkWarning`) and does not skip the test on a core-only or CPU-only setup. Keep host/device parity for the shared core functions here on CPU, but move the OVRTX wrapper-kernel assertions back to `source/isaaclab_ov/test/test_ovrtx_renderer_kernels.py` (or explicitly gate both the optional import and CUDA availability there).

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
rilei-nvidia: Hey good catch!

- "the annotator's JSON-compatible contract" - where is this reference from? 
- RTX/Replicator may have been stringifying the id keys in the past, but we don't necessarily need to adopt that approach 
- I realize we have done this for ovrtx which is an oversight

So what do you think of keeping the keys as int going forward, and I can follow up with a PR that makes ovrtx output int keys?  As for RTX/Replicator that is on its way out and can be considered legacy behavior?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
rilei-nvidia: Good catch, will fix

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
rilei-nvidia: Will address

source/isaaclab/test/renderers/test_segmentation_colors.py
rilei-nvidia: Good catch will move to isaaclab_ov

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
rilei-nvidia: Good catch, addressed!

source/isaaclab_newton/isaaclab_newton/renderers/segmentation.py
StafaH: Yea lets fix in follow-up. Ignore for now

StafaH (approved):

pbarejko (approved):

ooctipus (approved):

PR #6507, opened 2026-07-14
Encapsulate the articulation element kind into a type

AntoineRichard:
## Description

Follow-up to #6335 addressing @hujc7's review note that the element `kind` in `ordering_resolvers.py` should be encapsulated rather than threaded as a `Literal["joint", "body"]` string with per-kind data scattered across inline ternaries and module dicts.

Introduces a module-private frozen `_ArticulationElementKind` carrying the per-kind data as fields:
- `backend_names_attr` — `backend_joint_names` / `backend_body_names`
- `relationship_name` — `isaac:physics:robotJoints` / `isaac:physics:robotLinks`
- `name_override_attrs` — the schema name-override attributes (joint canonical `isaac:NameOverride` + lowercase fallback; body inverse)
- `config_field` — `joint_ordering` / `body_ordering`
- `matches_backend_spelling` — whether backend DoF-spelling matching applies (joints only)

Two interned singletons (`_JOINT_KIND` / `_BODY_KIND`) plus a label registry. `_validate_ordering_kind` is the single boundary conversion point; the public entry functions keep their `Literal["joint", "body"]` signatures, so callers and tests are unchanged. The two per-kind module dicts and the scattered `kind == "joint"` ternaries are gone.

Pure internal refactor, no behavior change and no public API change.

## Type of change

- Refactor (non-breaking, no functional change)

## Checklist

- [x] Ran `./isaaclab.sh --format`
- [x] Existing tests pass unchanged (`test_articulation_ordering.py`: 62 passed; `test_articulation_ordering_kernels.py`: 27 passed)
- [x] Added a `.skip` changelog fragment (internal refactor, no user-facing entry)

### DIFF
--- a/source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
+++ b/source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Sequence
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal
 
 from .ordering import (
@@ -50,6 +51,63 @@ if TYPE_CHECKING:
     from .base_articulation import BaseArticulation
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class _ArticulationElementKind:
+    """Per-kind data for one articulation element axis (joints or bodies).
+
+    Instances are module-private, frozen, and interned as the :data:`_JOINT_KIND`
+    and :data:`_BODY_KIND` singletons, so the kind threaded through ordering
+    resolution carries its own backend attribute, USD relationship, name-override
+    spellings, config field, and spelling-match policy instead of scattering
+    those as per-call ternaries and per-kind lookup tables.
+
+    Attributes:
+        label: Public element-kind alias, either ``"joint"`` or ``"body"``, used
+            for diagnostics and as the per-articulation convention cache key.
+        backend_names_attr: Name of the articulation attribute exposing active
+            backend names for this kind.
+        relationship_name: Isaac Sim robot-schema relationship authored on the
+            source asset for this kind.
+        name_override_attrs: Candidate name-override attribute spellings, tried in
+            order when reading a robot-schema target's authored name.
+        config_field: Configuration field naming this kind's public ordering.
+        matches_backend_spelling: Whether convention names for this kind are
+            rewritten to active-backend spellings; only joints carry the per-DoF
+            separator differences that need matching.
+    """
+
+    label: Literal["joint", "body"]
+    backend_names_attr: str
+    relationship_name: str
+    name_override_attrs: tuple[str, ...]
+    config_field: str
+    matches_backend_spelling: bool
+
+
+_JOINT_KIND = _ArticulationElementKind(
+    label="joint",
+    backend_names_attr="backend_joint_names",
+    relationship_name="isaac:physics:robotJoints",
+    name_override_attrs=("isaac:NameOverride", "isaac:nameOverride"),
+    config_field="joint_ordering",
+    matches_backend_spelling=True,
+)
+
+_BODY_KIND = _ArticulationElementKind(
+    label="body",
+    backend_names_attr="backend_body_names",
+    relationship_name="isaac:physics:robotLinks",
+    name_override_attrs=("isaac:nameOverride", "isaac:NameOverride"),
+    config_field="body_ordering",
+    matches_backend_spelling=False,
+)
+
+_ELEMENT_KINDS_BY_LABEL: dict[str, _ArticulationElementKind] = {
+    _JOINT_KIND.label: _JOINT_KIND,
+    _BODY_KIND.label: _BODY_KIND,
+}
 
 
 def _backend_matches_ordering_convention(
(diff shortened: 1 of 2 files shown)


source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -71,59 +129,63 @@ def _backend_matches_ordering_convention(
...
+    if resolved is not None:
+        return resolved
     raise ValueError(f"kind must be 'joint' or 'body'; got {kind!r}.")
 
 
-def _get_backend_names(articulation: BaseArticulation, kind: Literal["joint", "body"]) -> tuple[str, ...]:
+def _get_backend_names(articulation: BaseArticulation, kind: _ArticulationElementKind) -> tuple[str, ...]:
     """Return active backend names from an articulation."""
-    attr_name = "backend_joint_names" if kind == "joint" else "backend_body_names"
+    attr_name = kind.backend_names_attr
```
hujc7: Agent suggestion: `backend_names_attr` leaks its usage protocol — every consumer must know to `getattr` an articulation by it. A method on the kind keeps that knowledge with the field and this free function disappears:

```python
def get_backend_names(self, articulation) -> tuple[str, ...]:
    return _validate_articulation_names(
        getattr(articulation, self.backend_names_attr),
        parameter_name=f"Articulation {self.backend_names_attr}",
    )
```

Call sites: `kind.get_backend_names(articulation)`.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -153,9 +215,9 @@ def _get_prim_name(prim: Usd.Prim) -> str:
     return prim.GetName()
 
 
-def _get_robot_schema_target_name(prim: Usd.Prim, kind: Literal["joint", "body"]) -> str:
+def _get_robot_schema_target_name(prim: Usd.Prim, kind: _ArticulationElementKind) -> str:
     """Return the articulation name represented by a robot schema target prim."""
-    name_override = _get_prim_authored_string(prim, _ROBOT_SCHEMA_NAME_OVERRIDE_ATTRS[kind])
+    name_override = _get_prim_authored_string(prim, kind.name_override_attrs)
```
hujc7: Agent suggestion: same pattern — the try-in-order protocol for `name_override_attrs` belongs with the spellings:

```python
def resolve_target_name(self, prim) -> str:
    override = _get_prim_authored_string(prim, self.name_override_attrs)
    return override if override is not None else _get_prim_name(prim)
```

Call sites: `kind.resolve_target_name(target_prim)` — the `_robot_schema_` name qualifier becomes unnecessary once the receiver carries the context.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -298,19 +360,17 @@ def _match_backend_joint_name_spellings(
...
-    kind: Literal["joint", "body"],
+    kind: _ArticulationElementKind,
     names: Sequence[str],
     backend_names: Sequence[str],
 ) -> tuple[str, ...]:
     """Return convention names rewritten with active-backend spellings when needed."""
-    if kind == "joint":
-        return _match_backend_joint_name_spellings(names, backend_names)
-    return tuple(names)
+    return _match_backend_joint_name_spellings(names, backend_names) if kind.matches_backend_spelling else tuple(names)
```
hujc7: Agent suggestion: third of the trio — the policy flag currently branches at the caller; behind a verb it can't be branched on inconsistently:

```python
def match_backend_spellings(self, names, backend_names) -> tuple[str, ...]:
    return _match_backend_joint_name_spellings(names, backend_names) if self.matches_backend_spelling else tuple(names)
```

Call sites: `kind.match_backend_spellings(convention_names, backend_names)` (the `kind=` kwarg ceremony drops). With these three the kind stays six fields + three small methods; the walkers, orchestrators, and pure name utilities rightly stay module-level.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
AntoineRichard: Moved backend-name retrieval to _ArticulationElementKind.get_backend_names() and updated every caller in 6fedeab1ac.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
AntoineRichard: Moved robot-schema target-name resolution to _ArticulationElementKind.resolve_target_name() and removed the qualified free helper in 6fedeab1ac.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
AntoineRichard: Moved joint-only backend spelling policy to _ArticulationElementKind.match_backend_spellings() and updated both call sites in 6fedeab1ac.

hujc7 (approved):

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -71,59 +151,57 @@ def _backend_matches_ordering_convention(
     return convention.value in getattr(articulation, "__backend_native_orderings__", ())
 
 
-def _validate_ordering_kind(kind: object) -> Literal["joint", "body"]:
-    """Return a supported articulation element kind."""
-    if kind == "joint" or kind == "body":
+def _validate_ordering_kind(kind: _ArticulationElementKind | str) -> _ArticulationElementKind:
```
hujc7: Agent suggestion (non-blocking, fine as a follow-up): this boundary converter, the `_ELEMENT_KINDS_BY_LABEL` registry it reads, and the hardcoded `(_JOINT_KIND, _BODY_KIND)` iteration in `_get_complete_convention_names_by_kind` are three places that must stay in sync as the kind set evolves. Interning is also only conventional: a structurally-equal duplicate kind would compare and hash equal (frozen dataclass), silently shadowing a singleton anywhere kinds are dict keys. A self-registering `ClassVar` registry closes both — membership becomes a consequence of construction, and the duplicate guard turns interning into an enforced invariant:

```python
_registry: ClassVar[dict[str, _ArticulationElementKind]] = {}

def __post_init__(self) -> None:
    if self.label in self._registry:
        raise ValueError(f"duplicate articulation element kind {self.label!r}")
    self._registry[self.label] = self

@classmethod
def all(cls) -> tuple[_ArticulationElementKind, ...]:
    return tuple(cls._registry.values())

@classmethod
def resolve(cls, kind: _ArticulationElementKind | str) -> _ArticulationElementKind:
    if isinstance(kind, _ArticulationElementKind):
        return kind
    resolved = cls._registry.get(kind) if isinstance(kind, str) else None
    if resolved is None:
        raise ValueError(f"kind must be 'joint' or 'body'; got {kind!r}.")
    return resolved
```

`resolve` absorbs this function and `_ELEMENT_KINDS_BY_LABEL` — the name also repairs a drift this PR introduced: `_validate_ordering_kind` stopped being a validator when it started returning the converted kind rather than its input. `resolve` is honest about the idempotent `str | kind → kind` contract, where `from_label` would imply strings only (error message unchanged, so no test churn). The `_get_complete_convention_names_by_kind` loop becomes `for candidate_kind in _ArticulationElementKind.all():`, dropping the second hand-listed kind set. `__post_init__` mutates the class-level dict, not the frozen instance, so `frozen=True` doesn't object.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
```diff
@@ -337,14 +391,19 @@ def _get_complete_convention_names(
...
-        backend_names = _get_backend_names(articulation, candidate_kind)
+) -> dict[_ArticulationElementKind, tuple[str, ...]]:
+    """Return only complete convention-name candidates from a multi-kind provider.
+
+    :paramref:`names_by_kind` is keyed by the public ``"joint"``/``"body"`` label
+    (both kinds are produced together by the builder); the returned dict is keyed
+    by the resolved :class:`_ArticulationElementKind`.
+    """
+    complete_names: dict[_ArticulationElementKind, tuple[str, ...]] = {}
+    for candidate_kind in (_JOINT_KIND, _BODY_KIND):
```
hujc7: probably wont be a new type, but having a singleton of all the types and more encapsulation should help maintainability in general.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
AntoineRichard: Implemented in e3ba853d54. _ArticulationElementKind now self-registers by label, rejects duplicate labels in __post_init__(), and owns all()/resolve(); the manual registry and _validate_ordering_kind() are removed. Added a focused invariant test covering lookup, ordered enumeration, idempotence, and duplicate rejection.

source/isaaclab/isaaclab/assets/articulation/ordering_resolvers.py
AntoineRichard: Replaced the hand-listed (_JOINT_KIND, _BODY_KIND) iteration with _ArticulationElementKind.all() in e3ba853d54, so enumeration now derives from the same registry as construction and resolution.

AntoineRichard:
Included with preserved ancestry in aggregate PR #6784. Please keep this source PR unchanged; after #6784 merges, this PR can be closed as superseded if it is still open.

AntoineRichard:
Superseded by #6784, which preserves this PR's commits and ancestry in the aggregate integration branch.

PR #6521, opened 2026-07-14
Bump usd-core dependency to 26.05 to fix USD physics crash

mataylor-nvidia:
Resolves a fatal crash in libusd_ms triggered by LoadUsdPhysicsFromRange, caused by ABI incompatibility in the 25.11 USD library (nvbugspro #6441062).

Note this fix will work for kitless, but kit brings it's own USD

# Description

Fixes # (OMPE-100515)

## Type of change

## Screenshots

## Checklist


mataylor-nvidia:
The ovphysx_manager.py comment referencing `25.11` was left unchanged — it describes the version mismatch between IsaacSim's bundled pxr and ovphysx's internal check, which is a separate dependency from the usd-core PyPI pin.


```
@classmethod
    def _construct_physx(cls, ovphysx_device: str, gpu_index: int) -> None:
        """Bootstrap the ``ovphysx`` wheel and create the :class:`ovphysx.PhysX` instance.

        Runs once per process.  Configures worker threads, registers the
        process-exit ``os._exit(0)`` handler, and stores the result on
        ``cls._physx``.  See HACK on :meth:`_release_physx` for why the
        instance must outlive every individual :class:`SimulationContext`.
        """
        # HACK (temporary): hide pxr from sys.modules during ovphysx bootstrap.
        # IsaacSim's pxr reports version 0.25.5 (pip convention) while ovphysx
        # expects 25.11 (OpenUSD release convention).  Hiding pxr causes
        # ovphysx.check_usd_compatibility() to skip the Python-side version
        # check.  This should go away once ovphysx ships a namespaced USD
        # copy with isolated symbols (same "import pxr" API, no collision).
```

isaaclab.sh
```diff
@@ -61,5 +61,28 @@ if [ -d "$ISAACLAB_PATH/_isaac_sim" ]; then
...
+    _ov_usd_libs_dir=$(find "$ISAACLAB_PATH/_isaac_sim/extscache" -maxdepth 1 -name "omni.usd.libs-*" -type d 2>/dev/null | sort | tail -1)
+    if [ -n "$_ov_usd_libs_dir" ]; then
+        if [ -d "$_ov_usd_libs_dir/pxr" ] && [ ! -f "$_ov_usd_libs_dir/pxr/__init__.py" ]; then
+            printf '' > "$_ov_usd_libs_dir/pxr/__init__.py" 2>/dev/null || true
+        fi
+        export PYTHONPATH="$_ov_usd_libs_dir:$PYTHONPATH"
+        export LD_LIBRARY_PATH="$_ov_usd_libs_dir/bin:$LD_LIBRARY_PATH"
+    fi
+    unset _ov_usd_libs_dir
+fi
```
mataylor-nvidia, line 72: nit: these could be unified with a small helper script, however this is a small snippet so leaving in both entry points may be okay

isaaclab.sh
```diff
@@ -61,5 +61,28 @@ if [ -d "$ISAACLAB_PATH/_isaac_sim" ]; then
...
+
```
mataylor-nvidia, line 74: - 1. Find  the most recently versioned omni.usd.libs-<version> directory in the extension cache. sort | tail -1 picks the highest version string lexicographically.

- 2. Create a pxr package marker. Python linker skips otherwise

```
if [ -d "$_ov_usd_libs_dir/pxr" ] && [ ! -f "$_ov_usd_libs_dir/pxr/__init__.py" ]; then
    printf '' > "$_ov_usd_libs_dir/pxr/__init__.py" 2>/dev/null || true
fi
```

- 3. Expose the libraries to Python and the linker `export`

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

source/isaaclab/test/cli/test_source_package_metadata.py
```diff
@@ -23,16 +23,16 @@ def _repo_root() -> Path:
...
+    """The kit-less USD package must stay on the Isaac Sim compatible USD 26 ABI."""
     with (_repo_root() / "pyproject.toml").open("rb") as f:
         pyproject = tomllib.load(f)
 
     usd_core_dependencies = [
         dependency for dependency in pyproject["project"]["dependencies"] if dependency.startswith("usd-core")
     ]
 
-    assert usd_core_dependencies == ["usd-core>=25.11,<26.0 ; platform_machine in 'x86_64 AMD64'"]
+    assert usd_core_dependencies == ["usd-core>=26.05,<27.0 ; platform_machine in 'x86_64 AMD64'"]
```
StafaH: Maybe we can just get rid of these tests, since they're just checking that a hardcoded value is the correct hardcoded value? I don't see anything in this test file that's actually a useful test.

StafaH (approved):

source/isaaclab/test/cli/test_source_package_metadata.py
mataylor-nvidia: sure sounds good to me

kellyguo11 (approved):

mataylor-nvidia:
Possible issues:


Issue 1 — USD schema conflict (without omni.usd.libs in PYTHONPATH):
usd-core 26.8 registers UsdVolParticleField → alias ParticleField. Then ovrtx's old usd_particle_field plugin also tries to register ParticleField → alias UsdParticleFieldParticleField. Conflict. This is what setup_usd_libs.py was designed to fix — by making Python prefer omni.usd.libs/pxr (Isaac Sim's USD 25.11) over usd-core 26.8/pxr.

Issue 2 — OmniClient version mismatch (independent, present even after the USD fix):
ovphysx 0.5.9 bundles OmniClient 2.72.3 and hardcodes that exact version in ovstage-omniclient.version. Its Carbonite loader checks which omniclient.dll is already in the process and refuses if it's not 2.72.3. But omniverseclient==2.72.1 (pinned in pyproject.toml line 55) ships omni/client/omniclient.dll version 2.72.1 — which lands in memory first, causing the rejection.

AntoineRichard:

Closing as superseded: develop has since removed `usd-core` entirely in favor of a single `usd-exchange` provider (#6935/#6938), with guard tests that this branch would now fail (`test_isaaclab_uses_one_standalone_usd_provider`). The underlying OpenUSD crash is still real, and the upgrade path is now carried by #7192 (usd-exchange 3.0.0 → OpenUSD 26.08), which also retires the thread-limit workarounds.

Feel free to reopen if this was closed in error.

PR #6523, opened 2026-07-14
Updates PhysX external forces default to True

kellyguo11:
# Description

Default enable_external_forces_every_iteration to true for PhysX and warn when users opt out with the deprecated flag.

Add a deprecation warning and a breaking changelog fragment.

## Type of change

## Checklist


source/isaaclab_physx/test/assets/test_rigid_object.py
```diff
@@ -701,7 +701,7 @@ def test_rigid_body_no_friction(num_cubes, device):
                 tolerance = 1e-5
 
             torch.testing.assert_close(
-                cube_object.data.root_lin_vel_w.torch, initial_velocity[:, :3], rtol=1e-5, atol=tolerance
+                cube_object.data.root_lin_vel_w.torch[:, :2], initial_velocity[:, :2], rtol=1e-5, atol=tolerance
```
AntoineRichard, line 704: That change is odd, why not increase the tolerance?

AntoineRichard (approved):
Lots of behavioral changes which is a bit concerning, let's hope it doesn't show in training.

source/isaaclab_physx/test/assets/test_rigid_object.py
kellyguo11: I think this was a friction test with object sliding on a surface, so in theory, the check is only necessary on the x and y axes

PR #6529, opened 2026-07-15
Add heterogeneous clone scene composition

ooctipus:
> [!IMPORTANT]
> **Depends on #6534** (task scene-binding normalization). Merge #6534 first — this PR's task-side diffs assume the canonical `{ENV_REGEX_NS}` bindings introduced there.

## Summary
- Added heterogeneous clone-plan support so one `InteractiveScene` can instantiate different task asset sets across different environment rows.
- Added `scene_add` composition utilities for folding registered task scene configs into a single declarative scene with clone combinations.
- Added `scripts/demos/multitask_clone_scene.py`, which composes supported task scenes, skips global floor/light assets, and runs the resulting scene directly without task envs or managers.
- Converted env-local static task props such as tables, stands, and static scene props from plain `AssetBaseCfg` to kinematic `RigidObjectCfg`, so they participate in normal physics/USD replication instead of needing scene-level special 
Screenshot from 2026-07-14 19-56-19
handling.

## Motivation
This PR is meant to support multitask scene demos where each cloned environment may contain a different robot/task asset bundle. The previous clone flow assumed every env received the same source assets, which made it hard to build one scene from multiple existing task configs without manually rewriting each task into a shared schema.

The composition path keeps the existing task scene configs as the source of truth. It reads their declarative scene fields, deduplicates equivalent asset slots, and emits clone-combination rows that the cloner can use to populate only the assets needed by each task group.

## Design Notes
- `scene_add` only composes env-scoped, spawned, physics-backed assets rooted at literal `{ENV_REGEX_NS}/Leaf` paths. Global floors/lights should be skipped during composition and added once to the composed scene.
- Plain env-scoped `AssetBaseCfg` is rejected in `scene_add`; env-local props that need per-env instances should be modeled as `RigidObjectCfg` or `ArticulationCfg` instead.
- Static props that should not move are represented as kinematic rigid objects, which keeps them fixed while letting the existing backend replication paths handle them.
- The demo filters out scenes outside its current scope, including Newton-only scenes and scenes without a level-0 declarative flat floor.

## Test Plan
- `./isaaclab.sh -p -m py_compile source/isaaclab/isaaclab/scene/interactive_scene.py source/isaaclab/isaaclab/scene/scene_composition.py source/isaaclab_tasks/isaaclab_tasks/core/lift/lift_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_soft_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/core/reach/reach_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/stack/stack_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/stack/stack_instance_randomize_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/reach/reach_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/deploy/gear_assembly/gear_assembly_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/lift/config/openarm/lift_openarm_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/reach/config/openarm/unimanual/reach_openarm_uni_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_cloth_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/assemble_trocar/g129_dex3_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/place/config/agibot/place_toy2box_rmp_rel_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/place/config/agibot/place_upright_mug_rmp_rel_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/nutpour_gr1t2_base_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/exhaustpipe_gr1t2_base_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/locomanip_pick_place/locomanipulation_g1_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/locomanip_pick_place/fixed_base_upper_body_ik_g1_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/pickplace_gr1t2_env_cfg.py source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/pickplace_unitree_g1_inspire_hand_env_cfg.py`
- Searched for remaining env-scoped plain table/stand/scene/cube `AssetBaseCfg` assignments after converting static props to `RigidObjectCfg`.


vidurv-nvidia:
This PR is missing scripts/demos/multitask_clone_scene.py which is reported in the PR description!

scripts/demos/heterogeneous_scene.py
```diff
@@ -0,0 +1,181 @@
...
+        if task_spec.id.endswith("-Play"):
+            continue
+        entry_point = task_spec.kwargs.get("env_cfg_entry_point")
+        if isinstance(entry_point, str):
+            module_name = entry_point.split(":", maxsplit=1)[0]
+        else:
+            module_name = getattr(entry_point, "__module__", type(entry_point).__module__)
+        if module_name.startswith("isaaclab_tasks."):
+            task_ids.append(task_spec.id)
+    return sorted(task_ids)
```
AntoineRichard: Why do we need a private method here?

scripts/demos/heterogeneous_scene.py
```diff
@@ -0,0 +1,181 @@
...
+        value.init_state.pos[2]
+        for value in fields
+        if isinstance(value, AssetBaseCfg) and isinstance(value.spawn, sim_utils.GroundPlaneCfg)
+    ]
+    if not floor_levels:
+        return "no declarative flat floor"
+    mismatched = [level for level in floor_levels if abs(level) > 1e-6]
+    if mismatched:
+        return f"floor level {mismatched[0]} is not 0"
+    return None
```
AntoineRichard: Why would this one be public?

scripts/demos/heterogeneous_scene.py
```diff
@@ -0,0 +1,181 @@
...
+    """Gather every registered task scene and filter unsupported ones."""
+    accepted_ids, accepted_scenes, skipped = [], [], []
+    for task_id in _registered_task_ids():
+        env_cfg, _ = resolve_task_config(task_id, "")
+        reason = reject_scene(env_cfg)
+        if reason is not None:
+            skipped.append((task_id, reason))
+            continue
+        accepted_ids.append(task_id)
+        accepted_scenes.append(env_cfg.scene)
```
AntoineRichard: This is a bit ugly. Why do not return a bool and a string? 
Something like if not rejected then add it. etc...

scripts/demos/heterogeneous_scene.py
AntoineRichard: I would return a bool, and a string.

scripts/demos/heterogeneous_scene.py
```diff
@@ -0,0 +1,181 @@
...
+        if args_cli.num_task is not None and len(accepted_ids) == args_cli.num_task:
+            break
+
+    if skipped:
+        print("\n[INFO] Skipped task scenes outside the composition scope:")
+        for task_id, reason in skipped:
+            print(f"  {task_id}: {reason}")
+    if len(accepted_ids) < 2:
+        raise ValueError("Select at least two supported task scenes.")
+    return accepted_ids, accepted_scenes
```
AntoineRichard: Do we really need such a fancy system to skip tasks?

scripts/demos/heterogeneous_scene.py
AntoineRichard, line 1: I feel like the task selection is over engineered. I love over engineered stuff but this is making the readability of the code way worse. Why not pick a set of tasks here, it will be fixed yes, but at least we know what we load? I feel like the current code is opening the door to breaking sutff.

AntoineRichard (changes requested):
Some bits are overly complicated for no reasons that I can see. Please simplify a bit :D

scripts/demos/heterogeneous_scene.py
```diff
@@ -0,0 +1,163 @@
...
+    "IsaacContrib-Velocity-Flat-AnymalB",
+    "IsaacContrib-Velocity-Flat-AnymalC",
+    "IsaacContrib-Velocity-Flat-UnitreeA1",
+    "IsaacContrib-Velocity-Flat-UnitreeGo1",
+    "Isaac-Velocity-Flat-UnitreeGo2",
+    "Isaac-Velocity-Flat-Cassie",
+    "Isaac-Velocity-Flat-Digit",
+    "Isaac-Velocity-Flat-G1",
+    "Isaac-Velocity-Flat-H1",
+    "IsaacContrib-Navigation-Flat-AnymalC",
```
AntoineRichard, line 90: does it make sense to have 2 anymal C?

source/isaaclab/isaaclab/scene/interactive_scene_cfg.py
```diff
@@ -129,3 +136,99 @@ class MySceneCfg(InteractiveSceneCfg):
...
     ``useFabricForReplication`` is always ``False``.
     """
+
+    clone_cfg: CloneCfg = CloneCfg()
+    """Clone execution and legal scene-combination configuration."""
+
+
+def add(
+    this: InteractiveSceneCfg,
+    other: InteractiveSceneCfg,
```
AntoineRichard: would it make sense to rename `this` into `target` and `other` into `source`? `this` makes me think of C++'s this, which is odd in this scenario

source/isaaclab/isaaclab/scene/interactive_scene_cfg.py
```diff
@@ -129,3 +136,99 @@ class MySceneCfg(InteractiveSceneCfg):
...
+        ``this``, extended with the added scene's assets and one new clone combination.
+
+    Raises:
+        ValueError: If a scene has no environment asset or an asset root is
+            global or malformed, or if the added scene declares clone combinations.
+    """
+    if other.clone_cfg.clone_combinations:
+        raise ValueError("the added scene must not declare clone combinations; fold it as the base scene instead.")
+
+    left, right = _scene_assets(this, asset_skip), _scene_assets(other, asset_skip)
```
AntoineRichard: This kind of assigment is a bit odd

source/isaaclab/isaaclab/scene/interactive_scene_cfg.py
```diff
@@ -129,3 +136,99 @@ class MySceneCfg(InteractiveSceneCfg):
...
+    if not left or not right:
+        raise ValueError("both scenes must contain at least one spawned environment asset.")
+
+    # ``this`` accumulates only the participating assets: clear everything else
+    base_fields, left_names = InteractiveSceneCfg.__dataclass_fields__, {name for name, _ in left}
```
AntoineRichard: here it's even weirder imo because they don't really relate?

AntoineRichard (approved):

scripts/demos/heterogeneous_scene.py
ooctipus: I shouldn't matter :D

ooctipus:
Addressed the review comments in 9b985034f35: renamed `add()`'s operands to `target`/`source` (they now match the pre-existing `target_name`/`source_name` locals), renamed `left`/`right` to `target_assets`/`source_assets`, and split the bundled tuple assignments so each name gets its own line. Also 98575595295 restores the deferred `isaaclab.assets` import in the scene config that the develop merge accidentally hoisted to module level — that was the one failing core test (`test_factories_are_kitless_in_fresh_process`).

kellyguo11 (approved):

PR #6541, opened 2026-07-15
Honor replicate_physics in replication sessions

nblauch:
## Summary

- propagate `InteractiveSceneCfg.replicate_physics` into `ReplicateSession`
- skip physics-backend replication when `replicate_physics=False`
- continue USD replication so physics can discover the completed stage normally
- add coverage for policy forwarding, queue draining, and backend dispatch

## Problem

The replication-session refactor in `c7cbe661d07` stopped honoring `InteractiveSceneCfg.replicate_physics`.

Before that refactor, `replicate_physics=False` performed USD cloning without invoking backend physics replication. After the refactor, `InteractiveScene` entered a `ReplicateSession` that unconditionally dispatched every queued replication context.

This caused scenes explicitly configured with `replicate_physics=False` to invoke native PhysX replication unexpectedly.

## Implementation

`replicate()` and `ReplicateSession` now accept a `replicate_physics` argument, defaulting to `True` for compatibility with existing callers.

Physics replication contexts declare `replicates_physics = True`. When physics replication is disabled, those contexts are drained from the queue but not dispatched. USD replication still runs, preserving the complete stage needed for normal physics discovery.

PhysX, OVPhysX, and Newton replication contexts are marked consistently.

## Validation

- focused tests cover both enabled and disabled physics replication
- Ruff passes on all changed Python files
- downstream CPU suite: `347 passed, 41 skipped`
- 4/4 independent four-GPU PhysX + RTX runs using heterogeneous object mappings completed 10/10 training epochs, saved final checkpoints, and uploaded outputs with `replicate_physics=False`

The runtime-tested patch was applied to the IsaacLab revision pinned by the downstream image. The commit submitted here is an equivalent clean port onto the current `develop` branch.

## Important follow-up

This change restores the intended `replicate_physics=False` behavior, but it does not resolve a second issue observed during debugging:

With `replicate_physics=True`, native PhysX replication of a heterogeneous `ClonePlan`, where different environment rows receive different rigid-object prototypes/configurations, intermittently corrupts native memory under a multi-GPU PhysX + RTX workload. Observed failures include:

- `malloc(): unaligned tcache chunk detected`
- `double free or corruption`
- worker termination by `SIGABRT`

Disabling only object physics replication eliminated the failure in 4/4 runs, and restoring the scene-level `replicate_physics=False` policy passed another 4/4 runs.

This suggests a separate issue in the heterogeneous mGPU physics-replication path. It should be investigated independently; this PR restores the supported opt-out path but does not claim to fix that underlying failure.

### DIFF
--- a/source/isaaclab/changelog.d/restore-replicate-physics-policy.minor.rst
+++ b/source/isaaclab/changelog.d/restore-replicate-physics-policy.minor.rst
@@ -1,0 +1,6 @@
+Fixed
+^^^^^
+
+* Restored :attr:`isaaclab.scene.InteractiveSceneCfg.replicate_physics` behavior after the replication-session
+  refactor. Replication sessions can now skip physics-backend replication while retaining complete USD clones
+  for normal stage discovery.
--- a/source/isaaclab/isaaclab/cloner/replicate_session.py
+++ b/source/isaaclab/isaaclab/cloner/replicate_session.py
@@ -24,12 +24,14 @@ REPLICATION_QUEUE: list[tuple[Any, type]] = []
 """``(cfg, BackendCtxCls)`` pairs appended by ``queue_<backend>_replication`` and drained by :func:`replicate`."""
 
 
-def replicate(plan: ClonePlan, *, stage: Usd.Stage) -> None:
+def replicate(plan: ClonePlan, *, stage: Usd.Stage, replicate_physics: bool = True) -> None:
     """Drain :data:`REPLICATION_QUEUE` against ``plan``, dispatch each backend, publish the plan.
 
-    Cfgs absent from ``plan.cfg_rows`` are silently skipped. Backend contexts run in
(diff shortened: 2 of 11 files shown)


ooctipus:
Thanks again for identifying this regression and the repro, @nblauch! I've opened #6550 as an alternative mechanism for the same fix, following the design discussion on our side: assets register their replication contexts as a per-cfg list (capability as data), and the scene — the owner of the `replicate_physics` argument — filters physics contexts out of the queue, keeping `replicate()`/`ReplicateSession` policy-free. It also guards the two backends where skipping physics replication cannot work (OvPhysX assets have no USD-only cloning path; Newton's stage parser cannot rebuild `/World/envs` layouts), which would otherwise break silently under either mechanism.

The `replicate_physics=True` + heterogeneous mGPU double-free is a separate bug in the PhysX replication path (self-replication of single-variant rows creating duplicate PhysX objects at the same prim path) — we'll follow up on that one separately as you suggested.

PR #6549, opened 2026-07-15
Add golden image correctness tests for Kit and Newton visualizers

matthewtrepte:
## Description

Adds golden image correctness tests to Kit and Newton visualizers. 

Checks both interactive and tilted cameras for cartpole, shadow hand, anymal, and franka cloth for both physX & newton mjwarp

Also tries to reduce flakiness in existing visualizer integration tests and reduce overall visualizer test time.

Along the way, several flakiness or visual bugs were fixed to make the tests more reliable

1. Consolidate tiled visualizer tests into per-backend integration files (`test_visualizer_integration_physx.py`, `test_visualizer_integration_newton.py`) — removes the combined cross-backend test that was masking contamination
2. Fix frozen-body flakiness: pre-test RNG seeding now sets a minimum initial pole displacement so the cartpole always produces visible motion within the motion-check step window
3, Fix PhysX tiled camera warmup: `NewtonVisualizer` skipped `_log_camera_sensor_image()` on PhysX backend, leaving owned tiled cameras at zero renderer updates; the capture helper now pumps `camera_sensor.update()` before sampling
4. Fix `KitVisualizer` silently skipping tiled camera sensor creation in headless mode even when `--enable_cameras` is active
5. **Stale `NewtonManager._model`/`_state_0` on PhysX backend** — `PhysXManager.close()` does not call `NewtonManager.clear()`, so the shadow model built for a 4-env tiled stage persisted into the 1-env viewport test, causing the Newton viewer to auto-fit its camera to the wrong bounding box. Fix: explicitly clear `_model` and `_state_0` in the between-test prepare step.
6. **Stale `_scene_data_mapping` size mismatch** — after rebuilding `_model` from the new 1-env stage, the old 4-env body-index mapping was still used in `update_visualization_state()`, writing wrong transforms into `state_0.body_q` (1.38% consistent pixel diff). Fix: size check in `NewtonManager.update_visualization_state()` that invalidates the mapping when body count changes.
7. **CUDA RNG drift** — Newton tiled tests consume CUDA random samples, so `configure_seed(42)` at fixture start arrived at `env.reset()` at a different offset depending on prior test history, producing a different initial pole angle (6.04% consistent diff). Fix: reseed immediately before `env.reset()` in `run_visualizer_golden_cartpole()`.

## Type of change

## Screenshots

| physx-kit-tiled | physx-newton-tiled |
|---|---|
| physx-kit-tiled | physx-newton-tiled |

| physx-kit-viewport | physx-newton-viewport |
|---|---|
| physx-kit-viewport | physx-newton-viewport |

## Checklist


source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -385,9 +385,12 @@ def _setup_camera_sensor_view(self, num_envs: int) -> None:
         """Resolve or create the Camera sensor backing non-interactive image views."""
         if not self._uses_camera_sensor_view():
             return
-        if self._runtime_headless:
-            return
-        if not get_settings_manager().get("/isaaclab/cameras_enabled", False):
+        cameras_enabled = get_settings_manager().get("/isaaclab/cameras_enabled", False)
+        if not cameras_enabled:
+            if self._runtime_headless:
```
mataylor-nvidia, line 441: question: is there a reason the order of these checks changed?

source/isaaclab_visualizers/test/golden_images/cartpole/newton-newton-tiled.png
mataylor-nvidia, line 1: question: why is the pole on the right being culled?

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
matthewtrepte: Yes — this is an intentional fix. The original order was:

```python
if self._runtime_headless:
    return
if not cameras_enabled:
    ...
```

That early return caused tiled camera sensor creation to be silently skipped in headless mode even when `--enable_cameras` was passed (i.e., `cameras_enabled=True` + `_runtime_headless=True`). The test suite exposed this: Kit visualizer golden tests run headless with `enable_cameras=True`, so the tiled camera sensor was never created and all Kit-tiled captures produced a black frame.

The new order makes the intent explicit: the `_runtime_headless` guard is only meaningful when cameras are NOT enabled (headless-without-cameras → silent skip, non-headless-without-cameras → error). When cameras ARE enabled, we proceed with sensor creation regardless of headless mode, and only suppress the interactive UI image window below.

mataylor-nvidia: Is there any documentation that needs to be updated in regards to these settings?

I see `--headless` (sets sim_app.config["headless"]) does feed into `_runtime_headless`

source/isaaclab_visualizers/test/golden_images/cartpole/newton-newton-tiled.png
matthewtrepte: Expected behavior of the Newton GL tiled camera's viewport geometry. The Newton viewer renders via OpenGL with a fixed FOV; at the default `tiled_cam_eye` offset the cartpole pole (1 m tall) extends above the top of the viewport frame in certain tiles. This is the same behavior you'd see if you ran the Newton tiled camera demo interactively — the viewport crops the top of the pole.

The golden image captures this specific cropping as the reference. The SSIM comparison validates structural consistency across runs (same crop, same lighting, same pose) — not that it matches Kit RTX's rendering.

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
matthewtrepte: The changelog fragment documents the fix. For user-facing docs, `docs/source/how-to/visualizer_tiled_camera.rst` already shows `--enable_cameras` as the required flag for tiled camera use — that remains the correct guidance.

The behavior change is: **before this fix**, `--headless --enable_cameras` silently skipped the tiled camera sensor (producing black frames); **after**, it correctly creates the sensor and suppresses only the interactive UI window. The docs don't need to describe the old broken behavior, but if you think a note like "this also works headlessly — pass `--headless --enable_cameras`" would be useful in the how-to guide, I'm happy to add it.

matthewtrepte: haha ok next time i will make it clear when my bot is responding vs me (if its not already)

i think no documentation change since this is the correct user behavior described in the docs

source/isaaclab_visualizers/test/golden_images/cartpole/newton-newton-tiled.png
matthewtrepte: ah its just cropped off by the initial movement of the cartpole toward the left in those 2 envs. i think we adequately capture the pole in the other 2 envs.

huidongc (approved):

matthewtrepte:
@isaaclab-review-bot review

kellyguo11 (approved):

PR #6550, opened 2026-07-15
Honor replicate_physics via cfg-registered replication lists

ooctipus:
## Description

Since the replication-session refactor (#5770), `InteractiveSceneCfg.replicate_physics` has been ignored: scenes configured with `replicate_physics=False` invoked native physics replication anyway, silently discarding per-environment USD differences (e.g. prestartup scale randomization). Identified by @nblauch in #6541 with a multi-GPU heterogeneous-cloning repro.

This PR restores the flag as **cloner-side policy directed by asset cfgs**:

- **`CloneCfg.replicate_physics` is the policy's home.** `InteractiveScene` pipes its cfg flag into it and forwards it through `ReplicateSession` to `replicate()` — the scene contains no policy code.
- **`REPLICATION_QUEUE` holds bare asset cfgs.** Construction only registers *which* cfgs participate (`queue_replication(cfg)`); this stays at construction because `ClonePlan.from_env_0` scans the queue after constructors run in direct workflows, and because construction distinguishes clonable assets from lights/terrain.
- **`replicate()` resolves *how* each cfg clones at dispatch**: the cfg's `cloning_contexts` field when set (`None` → backend default, `()` → not cloned), otherwise the active backend's default stack, exported under the conventional name `isaaclab_<backend>.cloner.REPLICATION`:
  - PhysX: `(UsdReplicateContext, PhysxReplicateContext)` — USD clones are required beyond rendering (collision groups are authored on the per-env prims), and PhysX has no kitless mode.
  - Newton: `(Usd, Newton)` under Kit, `(Newton,)` kitless — headless runs skip the USD authoring cost, with the Kit check made at import under the documented app-first timing contract.
  - OvPhysX: `(OvPhysxReplicateContext,)` — its clone replay authors USD itself.
- **With `replicate_physics=False`, cloning is USD-only**: every context except `UsdReplicateContext` is dropped and the physics engine parses the per-env USD prims directly. An asset whose contexts are all physics-based is simply not cloned; if that matters, it surfaces at asset initialization. The Newton limitation is documented on the cfg.
- **Asset implementations know nothing about cloning** beyond registering their cfg. Per-asset control is a cfg assignment (e.g. `MPMObjectCfg` pins Newton-only contexts because particle rendering syncs through Fabric).
- The per-backend `queue_*_replication` helpers are removed in favor of the single path (changelogs carry the migration; fragments are `.major` for the affected packages).

Relation to #6541: same regression, alternative mechanism, per the design discussion — policy lives in the cloner cfg rather than the dispatcher signature, and asset cfgs are the single source of truth for how an asset clones.

## Validation

- `test_cloner.py`: 52 passed — queue/drain/dedup/priority under the bare-cfg shape, a drain-level USD-only filter test, and plan-publish forwarding of the flag
- `test_interactive_scene.py`: spy-based test asserts the PhysX replicator is registered with `replicate_physics=True` and NOT registered with `False` while envs still simulate; verified to fail with the drain filter removed (regression rule); cfg-override test covers `cloning_contexts`
- Newton default stack verified in both modes: `(Usd, Newton)` under Kit, `(Newton,)` kitless
- Pre-commit passes on all files

## Type of change

- Bug fix (restores documented behavior)
- Breaking change (REPLICATION_QUEUE shape; removed queue helpers — changelog fragments carry migration guidance)

## Checklist

- [x] I have added changelog fragments for all touched packages

### DIFF
--- a/docs/source/how-to/cloning.rst
+++ b/docs/source/how-to/cloning.rst
@@ -303,14 +303,17 @@ cfgs participate:
         ...
 
 *How* each cfg is cloned is resolved by :func:`~isaaclab.cloner.replicate` at
-dispatch: the cfg's :attr:`~isaaclab.assets.AssetBaseCfg.cloning_contexts`
-when set, otherwise the active backend's default stack. Each backend cloner
-exports that stack under the conventional name ``REPLICATION`` — PhysX pairs
-its native replication with USD clones, Newton includes USD only under Kit,
-and OvPhysX replicates alone because its clone replay authors USD itself.
-With :attr:`~isaaclab.cloner.CloneCfg.replicate_physics` disabled, cloning is
-USD-only: every context except ``UsdReplicateContext`` is dropped and the
-physics engine parses the per-env USD prims directly.
+dispatch. The physics side comes from the cfg's
+:attr:`~isaaclab.assets.AssetBaseCfg.cloning_contexts` when set, otherwise the
+active backend's default physics context, which each backend cloner exports as
+``PHYSICS_CONTEXT`` (PhysX and Newton replicate natively; OvPhysX replays its
+own clones). :class:`~isaaclab.cloner.UsdReplicateContext` is never authored by
+assets — :func:`~isaaclab.cloner.replicate` adds it automatically whenever a cfg
+has a spawner and Kit is available, so USD clones accompany physics replication
+under Kit and are skipped in headless runs. With
+:attr:`~isaaclab.cloner.CloneCfg.replicate_physics` disabled, cloning is
+USD-only: every physics context is dropped and the physics engine parses the
+per-env USD prims directly.
 
 Deferring the work like this buys three things at once:
 
--- a/docs/source/how-to/cloning.rst
+++ b/docs/source/how-to/cloning.rst
@@ -283,26 +283,34 @@ Every backend supplies its own :class:`~isaaclab.cloner.UsdReplicateContext` /
 ``PhysxReplicateContext`` / ``NewtonReplicateContext``, a class that hides the
 timing and granularity differences above behind a single uniform interface. A
 shared :data:`~isaaclab.cloner.REPLICATION_QUEUE` then remembers which asset
-belongs to which backend's context until it is time to run. The three
+cfgs participate until it is time to run. The three
 subsections below explain the queue, the contexts, and the function that joins
 them against a plan.
 
 The registration queue
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Asset constructors do not replicate inline. They register their intent with
-:data:`~isaaclab.cloner.REPLICATION_QUEUE` and the framework defers the actual
-work to the drain. The queue ends up holding one entry per ``(asset, backend)``
-pair:
+Assets do not replicate inline. Construction only registers the asset's cfg
+through :func:`~isaaclab.cloner.queue_replication`; the queue records *which*
+cfgs participate:
 
 .. code-block:: text
 
     REPLICATION_QUEUE
-        (cartpole_cfg, PhysxReplicateContext)
-        (cartpole_cfg, UsdReplicateContext)
-        (cube_cfg,     UsdReplicateContext)
-        (light_cfg,    UsdReplicateContext)
+        cartpole_cfg
+        cube_cfg
+        camera_cfg
         ...
+
+*How* each cfg is cloned is resolved by :func:`~isaaclab.cloner.replicate` at
+dispatch: the cfg's :attr:`~isaaclab.assets.AssetBaseCfg.cloning_contexts`
+when set, otherwise the active backend's default stack. Each backend cloner
+exports that stack under the conventional name ``REPLICATION`` — PhysX pairs
+its native replication with USD clones, Newton includes USD only under Kit,
+and OvPhysX replicates alone because its clone replay authors USD itself.
+With :attr:`~isaaclab.cloner.CloneCfg.replicate_physics` disabled, cloning is
+USD-only: every context except ``UsdReplicateContext`` is dropped and the
+physics engine parses the per-env USD prims directly.
 
 Deferring the work like this buys three things at once:
 
@@ -312,6 +320,11 @@ Deferring the work like this buys three things at once:
   call per asset.
 * Asset code stays free of any branching on which backend is active — it just
   registers and lets the framework take it from there.
+
+:attr:`~isaaclab.scene.InteractiveSceneCfg.replicate_physics` is piped into
+:attr:`~isaaclab.cloner.CloneCfg.replicate_physics` and applied at dispatch;
+an asset whose only cloning mechanism is physics replication is then simply
+not cloned.
 
 Backend contexts
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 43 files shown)


ooctipus:
@greptile-apps review again

ooctipus:
@greptile-apps review again

kellyguo11 (commented):
Two inline findings from the review.

source/isaaclab/isaaclab/cloner/replicate_session.py
```diff
@@ -20,32 +26,71 @@
...
+            physics_ctx = getattr(
+                importlib.import_module(f"isaaclab_{FactoryBase._get_backend()}.cloner"), "PHYSICS_CONTEXT", None
+            )
+            contexts = [physics_ctx] if physics_ctx else []
+        else:
+            contexts = [string_to_callable(c) if isinstance(c, str) else c for c in cfg.cloning_contexts]
+        if not replicate_physics:
+            contexts = [c for c in contexts if c is UsdReplicateContext]
+        ctx_set = dict.fromkeys(contexts)
+        if getattr(cfg, "spawn", None) is not None and has_kit():
```
kellyguo11, line 91: **[P1] Preserve explicit `cloning_contexts` overrides under Kit**

This unconditional USD insertion means a spawned cfg with an explicit override is not authoritative: `cloning_contexts=()` still clones through `UsdReplicateContext`, and the Newton-only tuple on `MPMObjectCfg` also receives USD clones under Kit despite its documented no-USD contract. I reproduced the empty-tuple case and observed the USD context being instantiated, queued, and run. Please only apply the automatic USD fallback when `cfg.cloning_contexts is None` (or otherwise define an explicit opt-out), and add dispatch-level coverage for both `()` and the MPM override.

docs/source/how-to/cloning.rst
```diff
@@ -283,27 +283,35 @@ Every backend supplies its own :class:`~isaaclab.cloner.UsdReplicateContext` /
...
-        (light_cfg,    UsdReplicateContext)
+        cartpole_cfg
+        cube_cfg
+        camera_cfg
         ...
 
+*How* each cfg is cloned is resolved by :func:`~isaaclab.cloner.replicate` at
+dispatch: the cfg's :attr:`~isaaclab.assets.AssetBaseCfg.cloning_contexts`
+when set, otherwise the active backend's default stack. Each backend cloner
+exports that stack under the conventional name ``REPLICATION`` — PhysX pairs
```
kellyguo11: **[P2] Document the backend symbol that actually exists**

No backend in this head exports `REPLICATION`; PhysX, Newton, and OvPhysX export only `PHYSICS_CONTEXT`. The new package changelog fragments also advertise `REPLICATION`, so users will be directed to a nonexistent API and the compiled changelog will carry invalid symbol references. Please either implement the documented stack API or update this guide and all affected fragments to describe `PHYSICS_CONTEXT` plus the USD auto-add behavior.

hujc7:
Went through with agent and it was mostly the same comment from Kelly.

source/isaaclab/isaaclab/cloner/replicate_session.py
ooctipus: the usd cloning is purely gated by have kit or not and this is intentional

ooctipus:
> Went through with agent and it was mostly the same comment from Kelly.

I can fix the doc but P1 is expected behavior, the requirement for usd cloning is now clear: has_kit + has_spawner_cfg

kellyguo11 (approved):

source/isaaclab/isaaclab/cloner/replicate_session.py
```diff
@@ -20,32 +26,71 @@
...
+
+    Args:
+        plan: Replication layout to dispatch.
+        stage: USD stage to author replicated prim specs into.
+        replicate_physics: Whether physics replication clones each environment. If False,
+            cloning is USD-only; an asset whose contexts are all physics-based is not cloned.
     """
     from isaaclab.sim import SimulationContext  # noqa: PLC0415
 
     queued = list(REPLICATION_QUEUE)
```
StafaH: Isn't replication queue already a list?

source/isaaclab/isaaclab/cloner/replicate_session.py
```diff
@@ -20,32 +26,71 @@
...
     backend_rows: dict[type, set[int]] = {}
-    for cfg, BackendCtxCls in queued:
+    for cfg in queued:
         rows = plan.cfg_rows.get(id(cfg))
         if rows is None:
             continue
-        backend_rows.setdefault(BackendCtxCls, set()).update(rows)
+        if cfg.cloning_contexts is None:
+            physics_ctx = getattr(
+                importlib.import_module(f"isaaclab_{FactoryBase._get_backend()}.cloner"), "PHYSICS_CONTEXT", None
```
StafaH: import lib inside a for loop is weird, we can just pull it out above the for loop?

source/isaaclab/test/scene/test_interactive_scene.py
```diff
@@ -153,11 +153,87 @@ def fake_replicate(plan, *, stage):
...
-    plan, stage = captured[0]
+    plan, stage, replicate_physics = captured[0]
     assert plan.sources == ("/World/envs/env_0",)
     assert plan.destinations == ("/World/envs/env_{}",)
     assert plan.clone_mask.shape == (1, 4)
     assert stage is scene.stage
+    assert replicate_physics is True
+
+
+@pytest.mark.parametrize("device", ["cuda:0"])
```
StafaH, line 164: Can't we just do it on CPU? Is there something different about GPU replication?

source/isaaclab_contrib/test/deformable/test_deformable_builder_hooks.py
```diff
@@ -137,30 +133,16 @@ def test_builder_hook_resets_entry_offsets_on_first_environment():
...
-        self._initialize_handle = None
-        self._invalidate_initialize_handle = None
-        self._prim_deletion_handle = None
-        self._debug_vis_handle = None
-        self._physics_ready_handle = None
-
-    monkeypatch.setattr(BaseDeformableObject, "__init__", fake_base_init)
-    monkeypatch.setattr(DeformableObject, "_register_deformable", lambda self: object())
-    REPLICATION_QUEUE.clear()
+def test_newton_physics_context_is_replicate_context():
```
StafaH, line 136: This is a weird test, why would it be something else?

source/isaaclab_physx/isaaclab_physx/cloner/replicate.py
```diff
@@ -95,6 +93,25 @@ def replicate(self) -> None:
...
+        # heap corruption (double-free / SIGABRT) under mGPU, likely due to per-call
+        # PhysX-internal allocations summing to a problematic total across processes.
+        # For these layouts the source prims are already in their correct env positions
+        # and PhysX can parse them from the stage without any replicator registration.
+        def _is_self_only(src: str, destination: str, target_envs: tuple[int, ...]) -> bool:
+            if len(target_envs) != 1:
+                return False
+            pre, suf = split_clone_template(destination)
+            env_id = src.removeprefix(pre).removesuffix(suf)
+            return env_id.isdigit() and int(env_id) == target_envs[0]
```
StafaH: This and the line above could just be:

`return src == f"{prefix}{target_envs[0]}{suffix}"`?

That would also leave no doubt about the match

StafaH (approved):
LGTM

source/isaaclab/isaaclab/scene/interactive_scene_cfg.py
```diff
@@ -114,6 +114,17 @@ class MySceneCfg(InteractiveSceneCfg):
     .. note::
         Optimized parsing of certain prim types (such as deformable objects) is not currently supported
         by the physics engine. In these cases, this flag needs to be set to False.
+
+    .. attention::
+        Setting this flag to False is currently not supported on the Newton physics backend:
+        Newton discovers the scene through its replication path, which stage parsing cannot
+        replace for cloned environments.
```
hujc7, line 121: Is this enforced?

PR #6552, opened 2026-07-15
Fix pytet arch issue

StafaH:
# Description

Fix for pytetwild pinning for uv resolution parsing.


ooctipus (approved):

myurasov-nv:
Added the install-CI regression test: `uv pip compile --universal` on the root pyproject reproduces this failure in seconds and passes with your fix (verified both directions locally).

PR #6556, opened 2026-07-16
Fix contact sensor filter matching geometry children with same name

ooctipus:
## Summary

- Wrap the leaf segment of each `filter_prim_paths_expr` in PhysX alternation syntax `(leaf)` before the `.*` → `*` conversion.

**Root cause:** URDF-to-USD conversion nests a rigid body and its geometry child under the same leaf name (e.g. `.../Object` body, `.../Object/Object` geometry). PhysX's glob engine prefix-matches a bare leaf pattern `env_*/Object` against both prims and raises a count-consistency error (`expected 1, got 2`). The same issue does not affect the sensor body resolution because `body_names_glob` already wraps leaf names in alternation `(name1|name2)` — anchoring the match to the correct path depth.

**Fix:**

```python
# before
filter_prim_paths_glob = [expr.replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr]

# after
filter_prim_paths_glob = [
    re.sub(r"([^/]+)$", r"(\1)", expr).replace(".*", "*")
    for expr in self.cfg.filter_prim_paths_expr
]
```

`re.sub(r"([^/]+)$", r"(\1)", expr)` captures the last path segment and wraps it in `()`. Example:

| expression | before | after |
|---|---|---|
| `env_.*/Object` | `env_*/Object` | `env_*/(Object)` |
| `env_.*/Robot/.*_FOOT` | `env_*/Robot/*_FOOT` | `env_*/Robot/(*_FOOT)` |

**Not affected:** Newton backend — it uses Python `fnmatch` against physics-model body labels, which never crosses `/`.

## Test plan

- [ ] Reproduce with a URDF asset where a link body and its geometry share a leaf name (standard URDF-to-USD output), configure a `ContactSensorCfg` pointing at another such asset in `filter_prim_paths_expr`, verify initialization no longer raises a count error.
- [ ] Existing contact sensor tests pass.

### DIFF
--- a/source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
+++ b/source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
@@ -340,7 +340,14 @@ class ContactSensor(BaseContactSensor):
         # note: with a list of patterns, the views order bodies pattern-major:
         #   view_id = body_id * num_envs + env_id
         body_path_globs = [expr.replace(".*", "*") for _, expr in body_matches]
-        filter_prim_paths_glob = [expr.replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr]
+        # Wrap the leaf segment of each filter expression in PhysX alternation syntax before
+        # the standard .* -> * conversion. PhysX's glob engine prefix-matches bare leaf
+        # patterns, so env_*/Object also matches env_0/Object/Object (a geometry child
+        # sharing the same name -- a common URDF-to-USD artifact). Wrapping in (leaf)
+        # anchors the match to that exact path depth.
+        filter_prim_paths_glob = [
+            re.sub(r"([^/]+)$", r"(\1)", expr).replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr
+        ]
 
         # create a rigid prim view for the sensor
         self._body_physx_view = self._physics_sim_view.create_rigid_body_view(body_path_globs)
--- a/source/isaaclab_physx/changelog.d/octi-fix-contact-filter-rigid-body-only.rst
+++ b/source/isaaclab_physx/changelog.d/octi-fix-contact-filter-rigid-body-only.rst
@@ -1,0 +1,10 @@
+Fixed
+^^^^^
+
+* Fixed :class:`~isaaclab_physx.sensors.ContactSensor` incorrectly matching geometry children
+  when resolving ``filter_prim_paths_expr``. URDF-to-USD conversion creates prim hierarchies
+  where a rigid body and its geometry child share the same leaf name (e.g. ``Object`` and
+  ``Object/Object``). PhysX's glob engine prefix-matches bare leaf patterns, finding both
+  prims and raising a count-consistency error. The fix wraps the leaf segment in PhysX
+  alternation syntax ``(leaf)`` before the ``.*`` → ``*`` conversion, anchoring the match
+  to the correct path depth.
--- a/source/isaaclab_physx/test/sensors/test_contact_sensor.py
+++ b/source/isaaclab_physx/test/sensors/test_contact_sensor.py
@@ -519,19 +519,23 @@ def test_contact_sensor_filter_same_name_geometry_child(setup_simulation, device
             env_xform = UsdGeom.Xform.Define(stage, env_path)
             env_xform.AddTranslateOp().Set(Gf.Vec3d(3.0 * env_id, 0.0, 0.0))
 
-            # Sensor subject: a simple kinematic rigid body cube.
+            # Sensor subject: an Xform rigid body with a separate geometry child.
+            # RigidBodyAPI must be on the Xform parent, not the geometry prim itself,
+            # so PhysX can resolve the body for the contact view.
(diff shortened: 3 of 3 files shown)


source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
```diff
@@ -340,7 +340,14 @@ def has_contact_report(prim) -> bool:
...
         body_path_globs = [expr.replace(".*", "*") for _, expr in body_matches]
-        filter_prim_paths_glob = [expr.replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr]
+        # Wrap the leaf segment of each filter expression in PhysX alternation syntax before
+        # the standard .* -> * conversion. PhysX's glob engine prefix-matches bare leaf
+        # patterns, so env_*/Object also matches env_0/Object/Object (a geometry child
+        # sharing the same name -- a common URDF-to-USD artifact). Wrapping in (leaf)
+        # anchors the match to that exact path depth.
+        filter_prim_paths_glob = [
+            re.sub(r"([^/]+)$", r"(\1)", expr).replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr
+        ]
```
hujc7, line 350: Just need a positive and negative test for the change if it's failing before

source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
```diff
@@ -340,7 +340,14 @@ def has_contact_report(prim) -> bool:
...
         #   view_id = body_id * num_envs + env_id
         body_path_globs = [expr.replace(".*", "*") for _, expr in body_matches]
-        filter_prim_paths_glob = [expr.replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr]
+        # Wrap the leaf segment of each filter expression in PhysX alternation syntax before
+        # the standard .* -> * conversion. PhysX's glob engine prefix-matches bare leaf
+        # patterns, so env_*/Object also matches env_0/Object/Object (a geometry child
+        # sharing the same name -- a common URDF-to-USD artifact). Wrapping in (leaf)
+        # anchors the match to that exact path depth.
+        filter_prim_paths_glob = [
+            re.sub(r"([^/]+)$", r"(\1)", expr).replace(".*", "*") for expr in self.cfg.filter_prim_paths_expr
```
hujc7, line 349: Agent review: this change doesn't resolve the count-consistency error — the regression test added in this PR fails in the `isaaclab_physx` CI job on the current head with the same error the fix targets:

```
[Error] [omni.physx.tensors.plugin] Filter pattern '/World/envs/env_*/(Table)' did not match the correct number of entries (expected 1, found 2)
```

https://github.com/isaac-sim/IsaacLab/actions/runs/29629359322/job/88040380661

To find out why, I called `create_rigid_contact_view` directly with the test's stage layout (`Table` rigid body Xform + `Table/Table` collider child, kinematic cube sensor body) across six filter-pattern forms:

| filter pattern | collider child name | result |
|---|---|---|
| `env_*/Table` (pre-fix) | `Table` | error: expected 1, found 2 |
| `env_*/(Table)` (this PR) | `Table` | error: expected 1, found 2 |
| `env_*/Table/Table` | `Table` | OK, `filter_count=1` |
| `/World/envs/env_0/Table` (exact, no wildcard) | `Table` | OK, `filter_count=1` |
| `env_*/Table` | `TableGeom` | OK, `filter_count=1` |
| `env_*/Table/*` | `Table` | OK, `filter_count=1` |

All six results fit one mechanism: the matcher tests patterns against the full paths of both rigid bodies and their shapes, and the glob `*` matches across `/`. So `env_*/Table` matches the shape `/World/envs/env_0/Table/Table` via `* = "0/Table"` — the ambiguity comes from the mid-path `env_*` wildcard, and wrapping the leaf as `(Table)` cannot anchor the depth (row 2: identical failure). The two shape-anchored forms also hold at 4 envs with the per-env data layout intact (`sensor_count=4`, `filter_count=1`).

Possible directions:

- Resolve the filter expressions USD-side — the same `resolve_matching_prims_from_source` infrastructure already used for the sensor bodies a few lines above — and emit shape-anchored globs (e.g. `<expr>/<collider-leaf>`) when a matched body has a descendant collider whose path also matches the glob; or detect the ambiguity and raise an actionable error. Note the config-side form works today with no code change: `filter_prim_paths_expr=["{ENV_REGEX_NS}/Object/Object"]`.
- The `*`-crossing-`/` glob semantics arguably belong upstream in `omni.physx.tensors` as the root-cause fix.

Also, the PR description's rationale ("`body_names_glob` already wraps leaf names in alternation, anchoring the match to the correct path depth") refers to code removed in #6378 — the current per-body patterns use a plain replace — and per row 2 alternation doesn't anchor depth regardless. The description and changelog fragment would need updating alongside the rework.

Done criterion: `test_contact_sensor_filter_same_name_geometry_child` passes in the `isaaclab_physx` CI job.

kellyguo11:
I think a similar fix went into physx upstream. perhaps this is no longer necessary?

PR #6559, opened 2026-07-16
Skip PhysX replicator for fully-heterogeneous 1:1 layouts (fixes mGPU double-free)

ooctipus:
## Summary

Fixes intermittent `double free or corruption` (SIGABRT) in `PhysxReplicateContext.replicate()` when running fully-heterogeneous scenes (one variant per env) across multiple GPUs. Stacks onto #6550.

**Root cause:** For a fully-heterogeneous 1:1 layout (N variants → N envs, each source mapped only to its own env), `replicate()` calls `rep.replicate()` N times — once per source with a single self-target each. This causes intermittent native heap corruption under mGPU. The source prims are already in their correct env positions; no cross-env replication is needed. Each call has PhysX-internal per-call allocations that apparently sum to a problematic total across concurrent processes.

**Fix:** Check whether all queued rows are self-only (`target_envs == (own_env_id,)`). When true, skip `register_replicator` entirely — PhysX parses the source prims directly from the stage, which is correct and avoids the crash.

**Controls (from `mgpu_repro.zip` in #6550):**
| Condition | Before | After |
|---|---|---|
| Heterogeneous mGPU (4 GPU, 192 variants) | 3/4 pass, 1/4 SIGABRT | expected 4/4 |
| Heterogeneous 1-GPU | 2/2 pass | unaffected |
| Homogeneous mGPU | 2/2 pass | unaffected |

## Failure signature (before fix)

```
Replicate physics     : True
Starting the simulation. This may take a few seconds. Please wait...
double free or corruption (!prev)
torch.distributed.elastic.multiprocessing.errors.ChildFailedError
traceback : Signal 6 (SIGABRT)
```

### DIFF
--- a/docs/source/how-to/cloning.rst
+++ b/docs/source/how-to/cloning.rst
@@ -283,26 +283,34 @@ Every backend supplies its own :class:`~isaaclab.cloner.UsdReplicateContext` /
 ``PhysxReplicateContext`` / ``NewtonReplicateContext``, a class that hides the
 timing and granularity differences above behind a single uniform interface. A
 shared :data:`~isaaclab.cloner.REPLICATION_QUEUE` then remembers which asset
-belongs to which backend's context until it is time to run. The three
+cfgs participate until it is time to run. The three
 subsections below explain the queue, the contexts, and the function that joins
 them against a plan.
 
 The registration queue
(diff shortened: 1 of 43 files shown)


ooctipus:
Folding this into #6550 directly.

PR #6553, opened 2026-07-15
Refactor train and play into rl subpackage

StafaH:
# Description

This PR moves the reinforcement-learning training and playback logic out of the
`scripts/reinforcement_learning/` launcher scripts and into a reusable
`isaaclab_rl.entrypoints` subpackage. The per-library launcher scripts
(`rl_games/`, `rlinf/`, `rsl_rl/`, `sb3/`, `skrl/`) were thin wrappers that
couldn't be imported or reused by downstream applications; the actual backend
logic now lives in the `isaaclab_rl` package and is exercised through a single
dispatcher.

**What changed**

- Added an `isaaclab_rl.entrypoints` subpackage that owns the backend
  implementations (`entrypoints/backends/train_*.py`, `play_*.py`), the shared
  `common.py` helpers, and a `dispatch.py` that selects a backend from
  `--rl_library` and forwards the remaining arguments.
- Added a typed **programmatic API** so downstream code can launch training and
  playback without building an `argparse.Namespace`:

  ```python
  from isaaclab_rl import TrainingRequest, train

  train(TrainingRequest(backend="rsl_rl", task="Isaac-Cartpole", max_iterations=100))
  ```

  `PlaybackRequest` / `play` provide the same for playback. Backend-specific
  options go through `backend_args`, and Hydra / preset selectors through
  `hydra_args`.
- Reduced the top-level `scripts/reinforcement_learning/train.py` and `play.py`
  to thin executables that call `run_train_cli` / `run_play_cli`.
- Removed the deprecated per-library scripts under
  `scripts/reinforcement_learning/<library>/` (`train.py`, `play.py`,
  `cli_args.py`). Use the unified `train.py`/`play.py` with `--rl_library`, or
  the programmatic API instead.
- Updated all internal callers — benchmarks, demos, Ray runners, AutoMate
  contrib launchers, the project template generator, and CI workflows — to use
  the unified entrypoints.
- Fixed `--deterministic` not configuring PyTorch deterministic operations in
  the RL-Games, RSL-RL, SB3, and skrl backends.
- Removed the `--use_last_checkpoint` flag from the RL-Games `play` entrypoint;
  use `--checkpoint latest` instead.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

**Breaking:** The per-library scripts under
`scripts/reinforcement_learning/<library>/` and the RL-Games
[...]

PR #6564, opened 2026-07-16
Promote benchmarks to library API

AntoineRichard:
# Description

Promotes the benchmark framework from `isaaclab.test.benchmark` to the public `isaaclab.benchmark` package. Runtime, startup, training, play, and RL backend workflows become library entrypoints with typed in-process dispatch.

The transitional import and script shims were removed from this PR to keep the move reviewable and leave one implementation for each workflow. This is an intentional breaking change: callers must replace `isaaclab.test.benchmark` imports with `isaaclab.benchmark` and use `isaaclab benchmark` instead of the former runtime, startup, training, and play scripts.

## Highlights

- Adds typed runtime, startup, training, and play request objects and `run_benchmark` convenience APIs.
- Returns workflow-specific result types such as `BenchmarkResult[RuntimeBundle]`.
- Adds in-process `isaaclab benchmark` CLI dispatch.
- Moves the supported benchmark workflows and RL backend adapters into the package.
- Removes the duplicate `isaaclab.test.benchmark` namespace and standalone workflow forwarding scripts.
- Restores process-global Torch, SKRL, and RL Games state and closes environments safely across repeated calls.
- Rejects schema finalization without an attached typed bundle instead of reporting a nonexistent output.
- Carries forward #6474 validation: positive workloads, non-negative warm-up counts, one default train/play warm-up step, and `--measure_sync_step`.
- Adds public API documentation, migration guidance, tests, and changelog fragments.

## Dependencies

- Direct dependency: #6474.
- Satisfied prerequisite: #6553, already merged, provides reusable RL train/play entrypoints.

## Breaking migration

- `from isaaclab.test.benchmark ...` becomes `from isaaclab.benchmark ...`.
- `scripts/benchmarks/{runtime,startup,training,play}.py` becomes `isaaclab benchmark {runtime,startup,training,play}`.
- Per-library train/play scripts become the unified training/play commands with `--rl_library`.
- Independent camera, loading, Hydra, and other standalone benchmarks remain unchanged.

## Type of change

- New feature
- Breaking API migration
- Documentation update

## Validation

- Focused benchmark API, CLI, parser, adapter, and early-stop suite: 265 passed, 4 simulator smoke cases deselected.
- Focused public API and CLI entrypoint suite: 26 passed.
- `./isaaclab.sh -f`: all applicable hooks pass. The conflict-marker hook was skipped because current `develop` contains literal `=======` headings in three third-party license files.
- `tools/changelog/cli.py check develop`: passed.
- Runtime/startup smoke tests now invoke `uv run isaaclab benchmark`; the local clean-`uv` bootstrap exceeded the validation window while downloading the CUDA environment, so CI remains the end-to-end CLI confirmation.

## Checklist

- [x] I have added tests for the public API and canonical CLI paths
- [x] I have documented the intentional breaking migration


AntoineRichard:
Integrated the latest #6474 tip and exposed its RL warm-up control through the typed Python API. `BenchmarkTrainingRequest` and `BenchmarkPlayRequest` now accept `warmup_steps`, and dispatch translates it to `--warmup_steps`. Existing training CLI coverage was extended and a focused play-request test was added. Validation: 18 API tests passed; 12 focused warm-up/timing tests passed; full `./isaaclab.sh -f` passed.

AntoineRichard:
Advanced-guide end-to-end validation found that typed requests were passed to each workflow explicitly, but nested `AppLauncher` parsing still read the caller process's `sys.argv`. Commit `ae91002357d` scopes `sys.argv` to the typed request during dispatch and restores it afterward.

Validation:
- regression test fails before the fix and checks both explicit and nested parser arguments
- all 18 API tests pass
- real PhysX typed runtime workflow reaches environment creation and stepping and writes schema + summary outputs
- full pre-commit passes

AntoineRichard:
Intentional API cleanup: this PR removes `MethodBenchmarkDefinition.dependencies` and the `dependencies` argument from `MethodBenchmarkRunner.run_benchmarks()`. Both public surfaces were unused and had no effect on benchmark execution. The maintainer has explicitly accepted this removal for the benchmark API promotion, so no compatibility shim is being added.

StafaH (commented):
Two major comments:

- The scripts under scripts/benchmark/** can I think all be dropped. We only need one entrypoint script now, all the backend specific stuff can be dropped. Also play/runtime/startup might be replaced with benchmark.py which the main CLI can call with the command play/runtime/startup and the script can just re-direct to the correct entrypoint in isaaclab.benchmark


- All these new play train scripts in isaaclab.benchmark.entrypoints should be named with specific file names (benchmark_play_rsl_rl.py) as we discussed in a previous pR

AntoineRichard:
Addressed the two review points:

- The compatibility play, training, runtime, startup, and backend-specific scripts under scripts/benchmarks were removed. The isaaclab benchmark CLI now dispatches directly into isaaclab.benchmark. The remaining files in scripts/benchmarks are independent benchmark workloads rather than forwarding entrypoints.
- The backend adapters remain grouped by backend and now use explicit filenames consistently, for example entrypoints/backends/rsl_rl/benchmark_play_rsl_rl.py and benchmark_train_rsl_rl.py. Dispatch paths and tests were updated for all four backends in 2114b111f0.

source/isaaclab/isaaclab/benchmark/entrypoints/backends/_state.py
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: This file is not good. Can we get rid of it and move it to the backend that needs it

StafaH (approved):

source/isaaclab/isaaclab/benchmark/entrypoints/backends/_state.py
AntoineRichard: Removed `benchmark/entrypoints/backends/_state.py` and moved the reusable guards to `isaaclab_rl.entrypoints.common`, matching the pattern in #6631. The Torch values are now explicit at each backend call site, and the rollback regression test exercises the common helpers. Implemented in 3afcb4c5d9 and 42667d591f.

kellyguo11 (approved):

PR #6579, opened 2026-07-16
Add uv lock and initial smoke tests

StafaH:
# Description

Upload a uv.lock and create uv testing workflow.

## Checklist


kellyguo11 (approved):

myurasov-nv:
This changes `[tool.isaaclab.versions] ovphysx` from an exact pin to a range, but `test_version_single_source_matches_literal_pins` (test_uv_run_pyproject.py:93) asserts `f"ovphysx=={versions['ovphysx']}"`, which now renders as `ovphysx==>=0.4.13` and fails. The test needs updating in this PR to accept a range spec for ovphysx, like it already does for ovrtx.

myurasov-nv:
The first step runs `uv lock --upgrade`, which regenerates the lock before anything uses it, so the committed uv.lock is never actually exercised and nothing fails when pyproject changes without a lock update. Suggest adding `uv lock --check` as a gate, then `uv sync --frozen` + train from the committed lock, and keeping `--upgrade` as a separate (or scheduled) job.

myurasov-nv:
Develop pinned `ovphysx==0.5.2+head` deliberately - is isaaclab_ovphysx actually tested against 0.4.13? Also `>=0.4.13` has no upper bound, so resolution will silently jump to 0.5.x the moment it publishes. Suggest `>=0.4.13,<0.5` for now and a re-pin once the 0.5.2 wheel is published.

myurasov-nv:
The header comment says this runs on every change to the dependency graph, but `paths:` includes `source/**`, so every Python change triggers a long double-train GPU job that overlaps install-ci. Suggest narrowing to `pyproject.toml`, `uv.lock`, `source/**/pyproject.toml`, and this workflow file.

myurasov-nv:
Minor: the job only runs on x86 - since the uv path has been aarch64-sensitive lately, it would be worth also running it on the ARM runner. And the success check greps for `Training time:` but not for the absence of `Traceback` (install-ci's helper checks both).

ooctipus (approved):

myurasov-nv:
CI triage: arm-ci and rendering-correctness-kitless fail identically on develop (unpinned ovrtx now resolves to 0.4.0.x, outside the supported <0.4.0 range) - not this PR. The ARM uv_pip failure was a CUDA OOM from runner contention - retry. The lock-check and version-test failures were real: fixed by merging develop + regenerating uv.lock, and by making the single-source test accept range specs.

myurasov-nv:
Existing CI failures:

| Failure | Status | Likely Reason |
|---|---|---|
| Installation Tests (x86, ARM) | fixed in 48f3484 (develop merged, lock regenerated) | committed uv.lock went stale when #6596 landed on develop |
| arm-ci | not from this PR | fails on develop too: unpinned ovrtx resolves to 0.4.0.x, outside the supported <0.4.0 range |
| rendering-correctness, rendering-correctness-kitless | not from this PR | same ovrtx cause as arm-ci, also red on develop |

The lock check will go stale again whenever develop's pyproject changes before this merges. The fix is always the same: run uv lock on the branch and commit.

myurasov-nv (approved):

PR #6582, opened 2026-07-17
[Task Clean-up] Dexterous Part 10/10: Move the warp Direct task variants to isaaclab_tasks_experimental

hujc7:
## Review Map

|---|---|---|---|
| #6411 Part 1/10: Newton cloner/cubric/visualizer fixes |  | — | merged |
| #6412 Part 2/10: OVPhysX articulation + manager runtime |  | — | merged |
| #6413 Part 3/10: Reorient Direct, torch |  | — | [changes](https://github.com/isaac-sim/IsaacLab/pull/6413/changes/f4895f0f9ee..d29afc75e71) |
| #6414 Part 4/10: MARL-to-single-agent fix + handover/camera Direct |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6414/changes/d29afc75e71..b10a84948f8) |
| #6418 Part 5/10: Reorient manager counterparts |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6418/changes/3ea6f7bbf6c..b31b04eaf71) |
| #6421 Part 6/10: Handover + camera manager counterparts |  | #6413, #6414, #6418 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6421/changes/b31b04eaf71..0cae3d87be4) |
| #6954 Part 8/10: OpenAI Shadow variants → contrib (draft) |  | #6421 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6954/changes/0cae3d87be4..b0d2d8448af) |
| 📌 #6582 Part 10/10: Warp variants → experimental (draft; merges last) |  | #6413 | [changes](https://github.com/isaac-sim/IsaacLab/pull/6582/changes/d29afc75e71..167c28578b3) |

## Summary

- Moves the warp-first implementation of the Shadow reorientation and handover Direct environments (kernels, environments, kernel behavior tests) to `isaaclab_tasks_experimental`, following the torch-first mainline convention: torch task logic in `isaaclab_tasks`, warp implementations in the experimental track.
- The variants register under `-Direct-Warp-v0` IDs and reuse the stable configurations and agents from `isaaclab_tasks.core`. The Allegro warp variant keeps its existing legacy registration in the experimental package.
- The diff is largely the review-polished warp implementation from the lumped validation branch #6324, relocated with import-path updates.

## Stacking

- Based on #6413's branch; this page shows the cumulative diff until #6413 merges. Merges last in the series.
- Deeper consolidation with the experimental warp track (`DirectRLEnvWarp`) and the performance focus for the experimental variants are follow-on scope, out of this series.

## Validation

- 44/44 warp kernel behavior tests; pre-commit green. Full integrated training/validation evidence: #6324.


source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/__init__.py
```diff
@@ -0,0 +1,58 @@
...
+reorient_warp_entry = "isaaclab_tasks_experimental.direct.reorient.reorient_warp_env:ReorientWarpEnv"
+stable_shadow = "isaaclab_tasks.core.reorient.config.shadow_hand"
+stable_shadow_agents = f"{stable_shadow}.agents"
+
+gym.register(
+    id="Isaac-Reorient-Cube-Shadow-Direct-Warp-v0",
+    entry_point=reorient_warp_entry,
+    disable_env_checker=True,
+    kwargs={
+        "env_cfg_entry_point": f"{stable_shadow}.shadow_hand_env_cfg:ShadowHandEnvCfg",
```
AntoineRichard, line 30: [AI Review] **Blocking:** `shadow_hand_env_cfg.py` no longer exists. All three Shadow Warp registrations in this file use this stale module name, and the current `isaaclab_tasks [3/3]` CI shard fails exactly these three config loads. Point them to `shadow_hand_direct_env_cfg`, then add a registry/config-load test covering every new ID so a successful kernel-only suite cannot mask broken entry points.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+import isaaclab.sim as sim_utils
+from isaaclab import cloner
+from isaaclab.assets import Articulation, RigidObject
+from isaaclab.envs import DirectRLEnv
+from isaaclab.markers import VisualizationMarkers
+from isaaclab.sensors import JointWrenchSensor, JointWrenchSensorCfg
+from isaaclab.sim.spawners.from_files import GroundPlaneCfg, spawn_ground_plane
+from isaaclab.utils.math import sample_uniform
+
+from isaaclab_tasks.core.reorient.reorient_task_base import GOAL_MARKER_POSITION, IN_HAND_POS_OFFSET
```
AntoineRichard, line 30: [AI Review] **Blocking:** `reorient_task_base` was removed; these constants now live in `reorient_common`. The `TYPE_CHECKING` import below also still names the removed `shadow_hand_env_cfg` instead of `shadow_hand_direct_env_cfg`. Once the registration strings are fixed, importing this environment will fail here, so please update both stale references and add an environment-entry import smoke test.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+import isaaclab.sim as sim_utils
+from isaaclab import cloner
+from isaaclab.assets import Articulation, RigidObject
+from isaaclab.envs import DirectMARLEnv
+from isaaclab.markers import VisualizationMarkers
+from isaaclab.sim.spawners.from_files import GroundPlaneCfg, spawn_ground_plane
+from isaaclab.utils.math import sample_uniform
+
+from isaaclab_tasks.core.handover.handover_env_cfg import HandoverEnvCfg
+from isaaclab_tasks.core.handover.handover_task_base import GOAL_POSITION_OFFSET
```
AntoineRichard, line 28: [AI Review] **Blocking:** `handover_task_base.py` does not exist in this head, and `GOAL_POSITION_OFFSET` has no current definition. This makes the new handover entry point fail on import. Please put the goal definition at a genuine shared boundary (configuration or a small common module used by both torch and Warp implementations) rather than restoring a single-consumer “task base,” and cover the registered entry point with a smoke test.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_env_cfg.py
```diff
@@ -121,6 +121,18 @@ class ShadowHandCameraEnvCfg(ShadowHandEnvCfg):
...
+        # The vision env renders through the Isaac RTX tiled camera, whose render
+        # products require the Fabric cloning path. The Newton backend disables Fabric
+        # cloning (see the base env's ``clone_in_fabric`` scene preset), so under Newton
+        # the ``rgb`` annotator has no render products for ``num_envs > 1`` and the
+        # default RGB/depth/semantic render fails. Default the vision env to PhysX so it
+        # renders out of the box; Newton stays selectable via ``physics=newton_mjwarp``
+        # for the depth-only Newton-warp-renderer benchmark path (``presets=newton_renderer``).
+        super().__post_init__()
+        for backend_cfg in (self.sim.physics, self.robot_cfg, self.object_cfg):
+            backend_cfg.default = backend_cfg.physx
```
AntoineRichard, line 134: [AI Review] **Blocking:** assigning `default` on these `PresetCfg` instances does not change preset resolution. `_preset_fields()` explicitly prefers the class attribute, so the inherited class-level Newton defaults still win and the camera remains Newton-default. Define camera-specific preset types or declarative `preset(...)` fields whose class-level `default` is PhysX, and add a `collect_presets(ShadowHandCameraEnvCfg())` assertion for physics, robot, and object defaults. The preceding seven-line rationale should also become one functional sentence, e.g. “Isaac RTX tiled cameras require the PhysX Fabric cloning path”; benchmark CLI history belongs in tests/docs, not this method.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+            observations["critic"] = self.compute_full_state().clone()
+        return observations
+
+    def _get_rewards(self) -> torch.Tensor:
+        # success flags, distances, and the episode-minimum error tracking were
+        # already computed this step by the kernel launches in :meth:`_get_dones`;
+        # the goal-reset flags, success counts, and moving average update in place.
+        # passing the time-out flags alone is exact here: the kernel ORs in the
+        # fallen-object condition itself, and that is this env's only other reset
+        reorient_reward(
```
AntoineRichard, line 267: [AI Review] The comment immediately above says reward reuses the success flags and orientation distances computed in `_get_dones`, but this call passes object/goal quaternions and `reorient_reward_kernel` recomputes `_rotation_distance` and the tolerance comparison. This is both misleading documentation and duplicate GPU hot-loop work. Pass `success_flags` and `orientation_error_buf` into the reward path, matching the torch implementation, and remove the redundant quaternion evaluation.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+
+        The raw ``[-1, 1]`` action is rescaled to the joint limits, blended with the previous
+        target via the exponential moving average, clamped to the limits, and set on the hand.
+        """
+        prev_wp, curr_wp = self._hand_targets_wp[agent]
+        wp.launch(
+            ema_actuation_kernel,
+            dim=(self.num_envs, self._actuated_dof_ids_wp.shape[0]),
+            inputs=[
+                wp.from_torch(self.actions[agent]),
```
AntoineRichard, line 171: [AI Review] The initialization comment claims the hot loop launches kernels without conversions, but `wp.from_torch(self.actions[agent])` is created here and again in `_hand_proprio_obs` for both agents every step. Use persistent per-agent action tensors, copy incoming actions into them in `_pre_physics_step`, and cache their Warp views once, as the reorientation variant already does.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+        dof_vel = self.right_hand.data.default_joint_vel.torch[env_ids] + self.cfg.reset_dof_vel_noise * dof_vel_noise
+
+        self.right_hand_prev_targets[env_ids] = dof_pos
+        self.right_hand_curr_targets[env_ids] = dof_pos
+
+        self.right_hand.set_joint_position_target_index(target=dof_pos, env_ids=env_ids)
+        self.right_hand.write_joint_position_to_sim_index(position=dof_pos, env_ids=env_ids)
+        self.right_hand.write_joint_velocity_to_sim_index(velocity=dof_vel, env_ids=env_ids)
+
+        # reset left hand
```
AntoineRichard, line 356: [AI Review] The left-hand reset block duplicates the right-hand block above almost statement-for-statement. Please extract a `_reset_hand(hand, prev_targets, curr_targets, env_ids)` helper (or iterate a per-agent state map). This is exactly the kind of copied lifecycle code that drifts during follow-up fixes and is avoidable within this PR.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+        device = str(device)
+        self.reward = wp.zeros(num_envs, dtype=wp.float32, device=device)
+        self.resets = wp.empty(num_envs, dtype=wp.bool, device=device)
+        # scratch for the consecutive-success reduction: [0] resetting-env
+        # count, [1] summed episode successes of the resetting environments
+        self.stats = wp.zeros(2, dtype=wp.float32, device=device)
+
+
+def reorient_reward(
+    reset_buf: wp.array,
```
AntoineRichard, line 369: [AI Review] This public launcher uses generic `wp.array` annotations for every array argument, contrary to the repository requirement to annotate Warp arrays with concrete dimensionality and dtype. Please specify the actual contracts here (`wp.bool`, `wp.float32`, `wp.vec3f`, `wp.quatf`, and 2D float actions) and on the return type. `compute_cube_keypoints()` has the same issue.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Warp kernels and device math for the reorientation task family.
+
+This is the family's math root: the experimental Direct environments and the
```
AntoineRichard, line 8: [AI Review] Please reduce this module docstring to the module’s functional purpose. The long “math root”/step-order narrative is maintenance-heavy, and it claims manager terms consume these kernels even though the current tree only imports them from the experimental environments/tests. Keep execution contracts beside the relevant launcher; remove historical validation and aspirational dependency narration.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_kernels.py
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Warp kernels for the handover task family.
+
+This is the family's math root: the experimental Direct environment launches
```
AntoineRichard, line 8: [AI Review] This “math root” and “reverse dependency never exists” prose is architectural narration rather than a functional kernel contract. A concise module description plus precise function-level frame/unit comments is enough. In particular, retain useful invariants such as world-frame versus environment-frame positions, but remove commentary about how the codebase is intended to be organized.

AntoineRichard (changes requested):
[AI Review] Requesting changes. The current Part 9 head is not merge-ready: the new Gym registrations and environment modules contain stale imports, and the camera PhysX-default override is ineffective under the current PresetCfg resolution rules. The kernel tests are useful and pass on CPU/CUDA in CI, but they do not instantiate the registered environments; the current `isaaclab_tasks [3/3]` CI shard is red on the three Shadow Warp config loads.

Before re-review, please also tighten the source comments to durable functional constraints, remove hot-loop duplication/conversions called out inline, use concrete Warp array annotations on public launchers, and add an `isaaclab_tasks` changelog fragment for the camera-default change.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_env_cfg.py
```diff
@@ -121,6 +121,18 @@ class ShadowHandCameraEnvCfg(ShadowHandEnvCfg):
...
     state_space = 187 + 27  # asymmetric states + vision CNN embedding
 
+    def __post_init__(self):
+        # The vision env renders through the Isaac RTX tiled camera, whose render
+        # products require the Fabric cloning path. The Newton backend disables Fabric
+        # cloning (see the base env's ``clone_in_fabric`` scene preset), so under Newton
+        # the ``rgb`` annotator has no render products for ``num_envs > 1`` and the
+        # default RGB/depth/semantic render fails. Default the vision env to PhysX so it
+        # renders out of the box; Newton stays selectable via ``physics=newton_mjwarp``
+        # for the depth-only Newton-warp-renderer benchmark path (``presets=newton_renderer``).
```
AntoineRichard, line 131: [AI Review][Comment cleanup] Replace this seven-line narrative with one durable functional constraint: `# Isaac RTX tiled cameras require the PhysX Fabric cloning path.` The failure history, `num_envs` scenario, CLI spelling, and benchmark path belong in the regression test or user documentation, not in `__post_init__`.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Warp-first Direct environment for the Shadow handover task.
+
+Experimental counterpart of the torch-first mainline implementation in
+:mod:`isaaclab_tasks.core.handover`; see the reorient variant for the conventions.
```
AntoineRichard, line 9: [AI Review][Comment cleanup] Make this module docstring self-contained. “See the reorient variant for the conventions” forces readers to inspect an unrelated module and does not describe a functional contract. A single sentence such as `Warp Direct environment for the Shadow handover task.` is sufficient.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # default goal positions
+        self.goal_rot = torch.zeros((self.num_envs, 4), dtype=torch.float, device=self.device)
+        self.goal_rot[:, 3] = 1.0  # identity quaternion in (x, y, z, w) layout
+        self.goal_pos = torch.zeros((self.num_envs, 3), dtype=torch.float, device=self.device)
+        # goal = object default position + shared offset (mirrors HandoverCommand.__init__)
+        self.goal_pos[:, :] = self.object.data.default_root_pose.torch[:, :3].to(self.device) + torch.tensor(
+            GOAL_POSITION_OFFSET, dtype=torch.float, device=self.device
+        )
+        # initialize goal marker
```
AntoineRichard, line 91: [AI Review][Comment cleanup] Remove the labels on lines 49, 63, 70, 78, 83, and 91; each merely repeats the immediately following variable lookup/allocation. Also drop the historical `mirrors HandoverCommand.__init__` clause on line 87. Once the goal definition is moved to a shared config/common boundary, its name should carry that relationship; if a comment remains, document only the coordinate frame or invariant.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+        self.goal_markers = VisualizationMarkers(self.cfg.goal_object_cfg)
+
+        # Sticky per-env flag: True once the object reached the goal within threshold.
+        self._episode_succeeded = torch.zeros(self.num_envs, dtype=torch.bool, device=self.device)
+        self._goal_distance = EpisodeErrorRecorder(self.num_envs, self.device)
+        self._success_flags = torch.empty(self.num_envs, dtype=torch.bool, device=self.device)
+        self._goal_distance_buf = torch.empty(self.num_envs, dtype=torch.float32, device=self.device)
+        self._reward_buf = torch.empty(self.num_envs, dtype=torch.float32, device=self.device)
+        # cached Warp views of the buffers above; the hot loop launches kernels without conversions
```
AntoineRichard, line 100: [AI Review][Comment cleanup] This comment is factually incorrect: action views are still created with `wp.from_torch(...)` in `_apply_hand_action` and `_hand_proprio_obs` every step. After introducing persistent per-agent action buffers, rewrite this narrowly as `# Cache Warp views of persistent Torch buffers.` Until then, remove it rather than claiming a property the implementation does not have.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+        spawn_ground_plane(prim_path="/World/ground", cfg=GroundPlaneCfg())
+        src, dest = "/World/envs/env_0", "/World/envs/env_{}"
+        pos = cloner.grid_transforms(self.scene.num_envs, self.scene.cfg.env_spacing, device=self.device)[0]
+        plan = cloner.ClonePlan.from_env_0(src, dest, self.scene.num_envs, self.device, pos)
+        cloner.replicate(plan, stage=self.scene.stage)
+        # add articulation to scene - we must register to scene to randomize with EventManager
+        self.scene.articulations["right_robot"] = self.right_hand
+        self.scene.articulations["left_robot"] = self.left_hand
+        self.scene.rigid_objects["object"] = self.object
+        # add lights
```
AntoineRichard, line 149: [AI Review][Comment cleanup] Remove `# add hand...`, `# add ground plane`, and `# add lights`; they only narrate the next constructor/call. Keep the EventManager constraint on line 145, but tighten it to the functional reason: `# Register assets with the scene for EventManager randomization.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+            device=self._compact_targets_wp.device,
+        )
+        hand.set_joint_position_target_index(target=self._compact_targets, joint_ids=self.actuated_dof_indices)
+
+    def _hand_proprio_obs(self, agent: str) -> torch.Tensor:
+        """Per-hand proprioceptive observation block for ``agent`` (133 dims).
+
+        Layout: normalized DOF positions (24), scaled DOF velocities (24), fingertip positions
+        (5*3), rotations (5*4), linear+angular velocities (5*6), and the applied actions (20).
+        """
```
AntoineRichard, line 189: [AI Review][Comment cleanup] Avoid the hard-coded `133 dims` and Shadow-specific `24/5/20` expansion because this method computes its width from `num_joints`, `num_fingertips`, and the configured action space. Document the stable formula instead: normalized positions (`J`), velocities (`J`), fingertip pose/velocity (`13F`), and actions (`A`).

AntoineRichard (commented):
[AI Review][Comment cleanup] Follow-up pass covering every comment/docstring block that violates the requested standard. Each inline note identifies the exact narration, restatement, stale claim, or hard-coded dimensional prose to remove or rewrite. Comments that encode coordinate frames, zero-copy lifetime, unavailable-sensor behavior, sticky state, or synchronization constraints are intentionally left untouched.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+        # reset when object has fallen
+        wp.launch(
+            fall_kernel,
+            dim=self.num_envs,
+            inputs=[self.object.data.root_pos_w.warp, self._env_origins_wp, self.cfg.fall_dist],
+            outputs=[self._fell_flags_wp],
+            device=self._fell_flags_wp.device,
+        )
+        out_of_reach = self._fell_flags
+        # reset when episode ends
```
AntoineRichard, line 300: [AI Review][Comment cleanup] Remove the comments on lines 253 (`compute reward`), 277 (`log reward components`), 291 (`reset when object has fallen`), and 300 (`reset when episode ends`); the calls and variable names already state those operations. Keep the host-synchronization note on line 281 and sticky-success invariant on line 285 because they explain non-obvious behavior.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
```diff
@@ -0,0 +1,381 @@
...
+        self.left_hand.write_joint_velocity_to_sim_index(velocity=dof_vel, env_ids=env_ids)
+
+    def _reset_target_pose(self, env_ids):
+        # reset goal rotation
+        rand_floats = sample_uniform(-1.0, 1.0, (len(env_ids), 2), device=self.device)
+        new_rot = randomize_rotation(
+            rand_floats[:, 0], rand_floats[:, 1], self.x_unit_tensor[env_ids], self.y_unit_tensor[env_ids]
+        )
+
+        # update goal pose and markers
```
AntoineRichard, line 378: [AI Review][Comment cleanup] Remove the procedural labels on lines 317, 320, 323, 341, 356, 372, and 378. They restate the next call/assignment and make the reset read like generated narration. Extracting the duplicated hand reset into `_reset_hand(...)` will also make the right/left labels unnecessary. Keep the episode-success flush comment above this range because it describes lifecycle semantics.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+            dtype=wp.float32,
+            device=str(self.device),
+        )
+        self.state_obs_buf = wp.empty((self.num_envs, state_dim), dtype=wp.float32, device=str(self.device))
+        # placeholder for kernel launches while the wrench sensor has no data
+        self.dummy_wrench = wp.zeros((1, 1), dtype=wp.vec3f, device=str(self.device))
+
+        # -- Warp buffers: reset randomization constants --
+        # batched unit axes consumed (through their torch aliases) by the torch.jit
+        # reset helper; a Warp reset path would replace them with wp.vec3f constants
```
AntoineRichard, line 137: [AI Review][Comment cleanup] Remove the hypothetical “a Warp reset path would replace...” narration. State only the current purpose if it is not obvious from naming: `# Unit axes used by the Torch reset rotation helper.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+        self.x_unit_vec_torch = wp.to_torch(self.x_unit_vec)
+        self.y_unit_vec_torch = wp.to_torch(self.y_unit_vec)
+        self.z_unit_vec_torch = wp.to_torch(self.z_unit_vec)
+
+        # -- Cached warp views of externally-owned torch tensors --
+        self.episode_length_buf_warp = wp.from_torch(self.episode_length_buf)
+        self.env_origins_warp = wp.from_torch(self.scene.env_origins, dtype=wp.vec3f)
+
+        # -- Initial buffer contents (written through the torch aliases) --
+        # in-hand target = object default position + shared offset (mirrors ReorientCommand)
```
AntoineRichard, line 184: [AI Review][Comment cleanup] Drop the historical `mirrors ReorientCommand` clause. The shared constant import should establish the common source; the only durable content here is the actual invariant, `# In-hand target is the default object position plus the task offset.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+        src, dest = "/World/envs/env_0", "/World/envs/env_{}"
+        pos = cloner.grid_transforms(self.scene.num_envs, self.scene.cfg.env_spacing, device=self.device)[0]
+        plan = cloner.ClonePlan.from_env_0(src, dest, self.scene.num_envs, self.device, pos)
+        cloner.replicate(plan, stage=self.scene.stage)
+        # add articulation to scene - we must register to scene to randomize with EventManager
+        self.scene.articulations["robot"] = self.hand
+        self.scene.rigid_objects["object"] = self.object
+        if self._joint_wrench_sensor is not None:
+            self.scene.sensors["joint_wrench"] = self._joint_wrench_sensor
+        # add lights
```
AntoineRichard, line 213: [AI Review][Comment cleanup] Remove `# add hand...`, `# add ground plane`, and `# add lights`; each restates the following operation. Preserve the EventManager requirement on line 208, shortened to `# Register assets with the scene for EventManager randomization.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+        if self.cfg.asymmetric_obs:
+            observations["critic"] = self.compute_full_state().clone()
+        return observations
+
+    def _get_rewards(self) -> torch.Tensor:
+        # success flags, distances, and the episode-minimum error tracking were
+        # already computed this step by the kernel launches in :meth:`_get_dones`;
+        # the goal-reset flags, success counts, and moving average update in place.
+        # passing the time-out flags alone is exact here: the kernel ORs in the
+        # fallen-object condition itself, and that is this env's only other reset
```
AntoineRichard, line 266: [AI Review][Comment cleanup] This execution narrative is both too long and inaccurate: `reorient_reward()` recomputes orientation distance/success instead of reusing the `_get_dones` buffers. After fixing that duplication, replace the block with at most one contract comment about the reset input, e.g. `# The reward kernel adds the fallen-object reset to these timeout flags.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+
+        # reset goals if the goal has been reached
+        goal_env_ids = self.reset_goal_buf_torch.nonzero(as_tuple=False).squeeze(-1)
+        if len(goal_env_ids) > 0:
+            self._reset_target_pose(goal_env_ids)
+
+        return self.reward_torch
+
+    def _get_dones(self) -> tuple[torch.Tensor, torch.Tensor]:
+        # reset when cube has fallen
```
AntoineRichard, line 302: [AI Review][Comment cleanup] Remove `# reset goals if the goal has been reached` and `# reset when cube has fallen`; the `goal_env_ids` branch and `out_of_reach_kernel` invocation already express both operations.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+                self.env_origins_warp,
+                self.in_hand_pos,
+                self.cfg.fall_dist,
+            ],
+            outputs=[self.out_of_reach_flags],
+            device=self.out_of_reach_flags.device,
+        )
+        # single per-step success evaluation (rewards and metrics reuse the buffers),
+        # followed by the progress bookkeeping (episode-counter reset in place,
+        # time-out by length or consecutive-success cap, episode-minimum error tracking)
```
AntoineRichard, line 317: [AI Review][Comment cleanup] Do not narrate the two launches and all of their outputs here; the kernel names and function-level contracts should carry that information. This block also repeats the false “rewards reuse the buffers” claim. After the dataflow fix, retain only a non-obvious ordering constraint if one actually exists.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+        self._compute_intermediate_values()
+
+    def _reset_target_pose(self, env_ids):
+        # reset goal rotation
+        rand_floats = sample_uniform(-1.0, 1.0, (len(env_ids), 2), device=self.device)
+        new_rot = randomize_rotation(
+            rand_floats[:, 0], rand_floats[:, 1], self.x_unit_vec_torch[env_ids], self.y_unit_vec_torch[env_ids]
+        )
+
+        # update goal pose and markers
```
AntoineRichard, line 403: [AI Review][Comment cleanup] Remove the procedural labels on lines 357, 360, 378, 397, and 403; each restates the following helper/assignment. Rewrite line 364 as the actual frame conversion constraint—`# Convert reset positions to the world frame.`—instead of the vague `global object positions`.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+        self.fingertip_rot = self.hand.data.body_quat_w.torch[:, self.finger_bodies]
+        self.fingertip_pos -= self.scene.env_origins.repeat((1, self.num_fingertips)).reshape(
+            self.num_envs, self.num_fingertips, 3
+        )
+        self.fingertip_velocities = self.hand.data.body_vel_w.torch[:, self.finger_bodies]
+
+        self.hand_dof_pos = self.hand.data.joint_pos.torch
+        self.hand_dof_vel = self.hand.data.joint_vel.torch
+
+        # data for object
```
AntoineRichard, line 423: [AI Review][Comment cleanup] The docstring claims a camera-environment consumer that does not exist in the current experimental package, while `# data for hand` and `# data for object` merely label obvious assignments. Either remove this unused compatibility method, identify and test its real consumer, or rewrite the docstring to that concrete contract; remove the two generic labels in either case.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+    new value), and clamped back to the limits. The result lands three ways:
+    ``cur_targets``/``prev_targets`` at the full-joint index ``dof_ids[j]`` (so
+    the smoothing state persists across steps), and ``compact_targets`` at the
+    actuated-joint index ``j`` for the articulation's indexed target write.
+    """
+    i, j = wp.tid()
+    dj = dof_ids[j]
+    lo = lower[i, dj]
+    hi = upper[i, dj]
+    # unscale from [-1, 1] to the joint range, blend, and clamp
```
AntoineRichard, line 68: [AI Review][Comment cleanup] Remove this line. The function docstring already describes unscaling, EMA blending, and clamping, and the next three assignments directly implement those operations.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Success and reward math (Warp).
+#
+# Kernels shared by the Direct environments and the manager terms in
+# :mod:`~isaaclab_tasks.core.reorient.mdp`; state arrays update in place. The
+# quaternion arithmetic mirrors :func:`isaaclab.utils.math.quat_mul`
+# (``(x, y, z, w)`` layout) term for term so results match the previously
+# validated reference implementation.
```
AntoineRichard, line 184: [AI Review][Comment cleanup] Remove this architectural/history block. Manager terms do not import these experimental kernels in the current tree, and “previously validated reference implementation” is not a maintainable contract. Keep the short section heading if useful; quaternion layout and mutation behavior belong in the relevant function signatures/docstrings.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+    success: wp.array(dtype=wp.bool),
+    distance: wp.array(dtype=wp.float32),
+):
+    """Evaluate goal-orientation success while exposing the physical error.
+
+    The single per-step success evaluation: writes the orientation distance
+    [rad] and whether it is within ``success_tolerance`` [rad]. Downstream
+    consumers — :func:`reorient_progress_kernel`, the manager command/reward/
+    termination terms, and the environments' metrics — reuse these buffers
+    instead of recomputing the quaternion math.
```
AntoineRichard, line 234: [AI Review][Comment cleanup] Trim this to the function’s direct output contract: it writes orientation error [rad] and the threshold comparison. The downstream-consumer list is stale—the manager terms do not consume this experimental kernel—and will keep drifting as callers change.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+    goal_reset = rotation_distance <= success_tolerance or goal_resets[i]
+    action_penalty = float(0.0)
+    for j in range(actions.shape[1]):
+        action_penalty += actions[i, j] * actions[i, j]
+    value = (
+        goal_distance * distance_scale
+        + rotation_scale / (rotation_distance + rotation_epsilon)
+        + action_penalty * action_penalty_scale
+    )
+    # one-time bonus on reaching the goal; penalty and episode reset when the object falls
```
AntoineRichard, line 297: [AI Review][Comment cleanup] “One-time bonus” is not guaranteed by this kernel’s inputs: an already-set `goal_resets[i]` also enters this branch and increments `successes`. Rewrite the comment to the exact condition, or change the dataflow so the kernel receives a newly-reached flag and the one-time claim becomes true.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+        inputs=[buffers.stats, averaging_factor],
+        outputs=[consecutive_successes],
+        device=buffers.reward.device,
+    )
+    return buffers.reward
+
+
+# ---------------------------------------------------------------------------
+# Cube-keypoint math shared by the Direct camera environment and the manager
+# camera observation terms.
```
AntoineRichard, line 473: [AI Review][Comment cleanup] Remove the nonexistent manager-camera consumer from this section description. The current tree uses this code from the experimental Direct path and tests; name only current consumers, or simply use `# Cube keypoint math.`

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
```diff
@@ -0,0 +1,765 @@
...
+    4. fingertips: positions (``3F``), rotations (``4F``), spatial
+       velocities (``6F``)
+    5. fingertip wrenches (``6F``) — full state only; the segment exists when
+       ``with_forces != 0`` and zero-fills while the sensor has no data
+       (``with_forces == -1``)
+    6. last actions (``A``)
+
+    Launched over ``(num_envs, obs_dim)``: each thread walks a branch ladder over the
+    segment boundaries and writes one output column, so warps (32 consecutive columns)
+    stay branch-uniform except at segment boundaries.
```
AntoineRichard, line 616: [AI Review][Comment cleanup] Keep the launch shape, but remove the speculative warp-uniformity explanation. The branch behavior is an implementation/performance claim without a benchmark or enforced invariant and makes this already-long observation docstring harder to scan.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_kernels.py
```diff
@@ -0,0 +1,154 @@
...
+    finger_ids: wp.array(dtype=wp.int32),
+    env_origins: wp.array(dtype=wp.vec3f),
+    actions: wp.array2d(dtype=wp.float32),
+    out: wp.array2d(dtype=wp.float32),
+):
+    """Per-hand proprioceptive block, matching the torch concatenation order.
+
+    Launched over ``(num_envs, obs_dim)``: each thread walks a branch ladder over the
+    segment boundaries and writes one output column, so warps (32 consecutive columns)
+    stay branch-uniform except at segment boundaries.
```
AntoineRichard, line 57: [AI Review][Comment cleanup] Keep the launch shape and output layout, but remove the warp-uniformity narrative. It is an implementation/performance claim rather than the kernel’s functional contract and duplicates the same prose in `full_obs_kernel`.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_kernels.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@wp.kernel
+def handover_reward_kernel(
+    goal_distance: wp.array(dtype=wp.float32),
+    distance_scale: float,
+    reward: wp.array(dtype=wp.float32),
+):
+    i = wp.tid()
+    # Direct handover reward: exponential falloff with object-goal distance
```
AntoineRichard, line 91: [AI Review][Comment cleanup] Remove this line; `handover_reward_kernel`, `goal_distance`, and the immediately following exponential expression already state exactly what it says.

source/isaaclab_tasks_experimental/test/direct/test_reorient_kernels.py
```diff
@@ -0,0 +1,495 @@
...
+    dist = torch.empty(N_ENVS, dtype=torch.float32, device=device)
+    wp.launch(
+        rotation_distance_kernel,
+        dim=N_ENVS,
+        inputs=[_wp2d(q1, wp.quatf), _wp2d(q2, wp.quatf)],
+        outputs=[wp.from_torch(dist)],
+        device=device,
+    )
+    torch.testing.assert_close(dist, _rotation_distance_ref(q1, q2), atol=1e-4, rtol=1e-4)
+    # identical quaternions have zero distance
```
AntoineRichard, line 103: [AI Review][Comment cleanup] Remove this comment. The test inputs (`q1`, `q1.clone()`), assertion, and test name make the zero-distance case self-explanatory.

source/isaaclab_tasks_experimental/test/direct/test_reorient_kernels.py
```diff
@@ -0,0 +1,495 @@
...
+        _wp2d(obj, wp.vec3f),
+        _wp2d(origins, wp.vec3f),
+        _wp2d(q_obj, wp.quatf),
+        _wp2d(target_pos, wp.vec3f),
+        _wp2d(q_goal, wp.quatf),
+        _wp2d(actions, wp.float32),
+        buffers=buffers,
+        **params,
+    )
+    # torch reference
```
AntoineRichard, line 292: [AI Review][Comment cleanup] Remove this generic label. The `_ref` variable names and subsequent assertions already distinguish the Torch reference computation from the Warp result.

source/isaaclab_tasks_experimental/changelog.d/task-cleanup-dex-warp-exp.minor.rst
```diff
@@ -0,0 +1,10 @@
+Added
+^^^^^
+
+* Added warp-first Direct variants of the Shadow cube reorientation and handover
```
AntoineRichard, line 4: [AI Review][Pointed follow-up] Part 9 also changes `source/isaaclab_tasks/.../shadow_hand_camera_env_cfg.py`, but this is the only fragment added by the Part 9 commit range. The inherited `source/isaaclab_tasks/changelog.d/task-cleanup-dex-part03.minor.rst` belongs to Part 3 and does not mention the camera default; after the stack is rebased, Part 9 will touch `isaaclab_tasks` with no fragment. Either move commit `386272f2` to its owning stable-task PR or add a separate `source/isaaclab_tasks/changelog.d/<slug>.rst` `Fixed` entry for the camera backend default.

source/isaaclab_tasks_experimental/test/direct/test_reorient_kernels.py
```diff
@@ -0,0 +1,495 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Parity unit tests for the reorient family Warp kernels.
+
+Kit-free: imports only torch/warp and the kernel module. Every kernel is
+checked against an independent torch reference implementation on randomized
+inputs, on CPU and CUDA.
```
AntoineRichard, line 10: [AI Review][Pointed follow-up] Keep these kernel tests Kit-free, but add a separate registration smoke test because these 44 cases never load an `env_cfg_entry_point` or import either environment class. The current CI failure proves the gap: parameterize the four new Gym IDs, resolve each config entry point, and import its environment entry point. That test would fail on all five stale references currently present (`shadow_hand_env_cfg` ×3, `reorient_task_base`, and `handover_task_base`) while preserving this file’s focused kernel contract.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
```diff
@@ -0,0 +1,520 @@
...
+                self.orientation_error_buf,
+                self.successes,
+                float(self.cfg.max_consecutive_success),
+                self.max_episode_length,
+            ],
+            outputs=[
+                self.episode_length_buf_warp,
+                self._orientation_error.minimum_error,
+                self.time_out_flags,
+                self._orientation_error._has_sample,
```
AntoineRichard, line 339: [AI Review][Pointed follow-up] This reaches into `EpisodeErrorRecorder._has_sample`, a private field owned by `isaaclab_tasks.core.utils`, and makes the experimental kernel depend on that class’s internal storage layout. `EpisodeErrorRecorder.update()` is the public boundary and also owns shape/finite-value semantics. Expose an explicit Warp-compatible recorder API/buffer contract, or keep Warp-owned minimum/sample buffers in this environment and pass summarized values to the recorder at reset; do not bind a kernel output directly to another class’s private attribute.

AntoineRichard (commented):
[AI Review][Pointed follow-up] Three additional places where the review should name the exact missing coverage or ownership boundary, rather than leaving the fixer to infer it.

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/shadow_hand_camera_env_cfg.py
AntoineRichard: [AI Review][Pointed follow-up] The exact precedence rule is in [`_preset_fields()`](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks/isaaclab_tasks/utils/hydra.py#L192-L204): `getattr(type(preset_obj), field_name)` wins over the instance field. The existing regression [`test_collect_fields_prefers_class_attr_over_instance`](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks/test/core/test_hydra.py#L1091-L1110) codifies that behavior. Add the new camera-specific assertion beside the [`shadow_hand_camera_presets` fixture](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks/test/core/test_shadow_hand_camera_presets.py#L177-L180), checking that `sim.physics.default`, `robot_cfg.default`, and `object_cfg.default` resolve to their PhysX alternatives.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_warp_env.py
AntoineRichard: [AI Review][Pointed follow-up] The stable torch path shows the intended API: [`ReorientDirectEnv._get_rewards()`](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks/isaaclab_tasks/core/reorient/reorient_direct_env.py#L184-L206) passes `_success_flags` and `_orientation_error_buf` into `reorient_reward`. By contrast, the Warp kernel recomputes both at [`reorient_kernels.py:283-288`](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py#L283-L288). Concretely, replace the object/target quaternion plus tolerance inputs with `goal_reached: wp.array(dtype=wp.bool)` and `rotation_distance: wp.array(dtype=wp.float32)`, then use those values for the reset, bonus, and inverse-error reward.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/reorient/reorient_kernels.py
AntoineRichard: [AI Review][Pointed follow-up] The concrete annotation map is: `reset_buf`/`goal_resets` → `wp.array(dtype=wp.bool)`, `successes`/`consecutive_successes` → `wp.array(dtype=wp.float32)`, positions/origins → `wp.array(dtype=wp.vec3f)`, quaternions → `wp.array(dtype=wp.quatf)`, actions → `wp.array2d(dtype=wp.float32)`, and the return → `wp.array(dtype=wp.float32)`. For `compute_cube_keypoints`, use `wp.array2d(dtype=wp.float32)` for poses and `wp.array2d(dtype=wp.vec3f)` for keypoints. This makes the public launcher contract match the typed kernel signatures instead of discarding their information.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/direct/handover/handover_warp_env.py
AntoineRichard: [AI Review][Pointed follow-up] The stable implementation currently hard-codes the goal at [`handover_env.py:73-77`](https://github.com/isaac-sim/IsaacLab/blob/386272f275f0492831b30d5082ce02000adf6361/source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env.py#L73-L77). That is the concrete duplication to remove: define one environment-frame `GOAL_POSITION` in `handover_env_cfg.py` or a `handover_common.py`, then make both `HandoverEnv` and `HandoverWarpEnv` consume it. Do not recreate the deleted `handover_task_base.py` solely to satisfy this import.

PR #6584, opened 2026-07-17
Pin Newton to v1.4.0 and override the Isaac Sim MuJoCo pins

hujc7:
## Summary

- Advances the Newton pin from a pre-1.4 development commit (`c7ae7c7648c`) to the **v1.4.0 release**, which contains the `ignore_paths` custom-frequency USD traversal fix ([newton#3406](https://github.com/newton-physics/newton/pull/3406), membership in the release verified) plus the 1.4 stabilization series — 50 commits newer than the previously targeted dev SHA.
- Newton 1.4 requires the MuJoCo 3.10 stack while `isaacsim-core 6.0.0.1` pins `mujoco-warp==3.8.0.3` exactly, so `mujoco-warp>=3.10.0.2,<3.11` and `mujoco~=3.10.0` join the `[tool.uv]` override-dependencies (the same mechanism already used to out-rank isaacsim's torch and newton pins; the 3.10.0.2 floor is what v1.4.0 declares). The version table and its single-source test carry the new entries.
- Extracted from #6411 so the approved runtime fixes there stay decoupled from the dependency advance: #6411's Multi-GPU CI failure bisects to this pin, not to its code.

## Problem: mesh-mesh contact regression in the MuJoCo pipeline (upstream)

The Newton range this pin crosses includes [newton#2980](https://github.com/newton-physics/newton/pull/2980) (MuJoCo 3.10 margin/gap semantics), which introduces a regression: **mesh-mesh shape pairs in the MuJoCo contacts pipeline lose collision response entirely** — bodies tunnel through each other with zero contact forces.

Bisect (via `test_horizontal_collision_detects_contact[mesh_capsule-mujoco_contacts]`):

| Newton ref | Result |
|---|---|
| `c7ae7c7648c` (current develop pin) | pass |
| `346121aaaca` = newton#2980 | **fail (first bad)** |
| `9af5a9f4181` (1.5.0.dev0) | fail |
| `v1.4.0` | fail |

- No older pin avoids it: the `ignore_paths` fix landed 31 commits **after** the regression, so every Newton commit containing the fix also contains the regression.
- `legacy_margin_gap=True` at USD import does not restore the behavior (the failure is contact-pair loss, not margin-value translation), and neither does raising the builder default gap.
- Scope: primitive pairs and mesh-vs-primitive pairs are unaffected (the dexterous task families train to full success rates under this pin).
- Tracked upstream as [newton#3559](https://github.com/newton-physics/newton/issues/3559); the affected test case is marked xfail referencing that issue and should be re-enabled when the upstream fix lands.

## Stacking

- Independent of the dexterous task-cleanup series (based on develop). If #6411 merges first, this PR picks up the removal of its now-redundant ignore_paths cloner workaround on rebase.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -168,8 +168,11 @@ torchaudio = "2.11.0"
 ovphysx = "0.5.2+head.f62c22207c"
 ovrtx = ">=0.3.0,<0.4.0"
 # Newton git commit + matching warp-lang prerelease (upgraded together).
-newton = "c7ae7c7648cd0717df39e5c94b95d5a02c997320"
+newton = "v1.4.0"
 warp = "1.15.0.dev20260626"
+# MuJoCo stack required by the Newton pin; overrides isaacsim-core's exact 3.8 pin.
+mujoco_warp = ">=3.10.0.2,<3.11"
+mujoco = "~=3.10.0"
 
 [tool.ruff]
 line-length = 120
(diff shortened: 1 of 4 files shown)


AntoineRichard (approved):

AntoineRichard:
Closing this as we're going to merge this one instead: https://github.com/isaac-sim/IsaacLab/pull/6586

PR #6585, opened 2026-07-17
Make PyTetWild optional for tetrahedralization

mmichelis:
# Description

Move PyTetWild out of the base dependencies and expose it through an explicit `tetrahedralization` extra. This keeps PyVista and VTK, which add roughly 670 MiB in a fresh environment, out of workflows that do not automatically generate volume tetrahedral meshes.

The extra installs `pytetwild[all]>=0.3.0,<0.4`, providing a validated 0.3 floor and allowing patch updates while retaining Linux x86_64, Linux aarch64, and Windows amd64 wheel support. It is intentionally excluded from both `isaaclab[all]` and the default `-i all` installation.

Users requiring automatic volume tetrahedralization must install it explicitly:

- uv (recommended): `uv sync --inexact --extra tetrahedralization`, or add `--extra tetrahedralization` to a `uv run` invocation
- Wheel: `pip install "isaaclab[tetrahedralization]"`
- Legacy installer: `./isaaclab.sh -i tetrahedralization`

The documentation and the runtime error raised when PyTetWild is missing both lead with the uv commands and keep the `isaaclab.sh` equivalent as a fallback, steering users toward the recommended uv workflow.

Pre-tetrahedralized `UsdGeom.TetMesh` assets and deformable workflows that do not generate volume tetrahedral meshes are unaffected. Missing optional dependencies now produce installation guidance at the point of use.

## Type of change

- Breaking change
- Documentation update

## Screenshots

Not applicable.

## Validation

- Merged current `develop` at `a050ceade7`, resolving the recurring PyTetWild dependency conflicts in favour of the optional extra
- Regenerated the committed `uv.lock` and verified `uv lock --check` passes and is idempotent
- Verified `uv sync --inexact --extra tetrahedralization` resolves PyTetWild 0.3.0, PyVista, and VTK
- Ran 66 focused install CLI, wheel-metadata, uv-project, and lockfile tests
- Ran the full `uv run isaaclab -f` suite; all hooks passed
- Ran `./isaaclab.sh -p tools/changelog/cli.py check develop`
- Verified base and `all` metadata exclude PyTetWild
- Tetrahedralized a unit cube into 503 tetrahedra with total volume `1.0`
- Verified Linux x86_64, Linux aarch64, and Windows amd64 dependency resolution
- Added install-CI coverage confirming source and wheel `all` installations omit PyTetWild
- Opted only CI lanes that exercise procedural volume deformables into the bounded PyTetWild extra, including the standalone demo job that runs `scripts/demos/deformables.py`
- Verified remote x86 and ARM installation tests pass
- Verified all three task shards, contrib, PhysX, and the Franka-soft rendering suites use PyTetWild 0.3.0
- Confirmed the remaining CI failures are unrelated to tetrahedralization: infrastructure flakes and the newly added visualizer golden-image tests, which also fail on other pull requests

## Checklist

- [x] I have added a major changelog fragment under `source/isaaclab/changelog.d/`


pyproject.toml
```diff
@@ -38,11 +38,8 @@ dependencies = [
     "gymnasium>=1.2.0",
     "trimesh",
     "pyglet>=2.1.6,<3",
-    # pytetwild: keep the behavior-stable 0.2.3 pin on x86_64. 0.3.0 added manylinux
-    # aarch64 wheels; it imports pyvista at import time but only declares it in its
-    # [all] extra, so depend on that extra there.
-    "pytetwild==0.2.3 ; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
-    "pytetwild[all]>=0.3.0,<0.4 ; platform_machine == 'aarch64' or platform_machine == 'arm64'",
+    # pytetwild 0.3.0 imports pyvista at import time but declares it only in the [all] extra.
+    "pytetwild[all]==0.3.0",
```
StafaH: Should we do >= 0.3.0, or would that be dangerous?

StafaH (approved):

pyproject.toml
mmichelis: I used an exact pin intentionally. >=0.3.0 resolves identically today, but it would allow future minor, major, and prerelease versions without review. PyTetWild is still Alpha, upstream is already developing 0.4, and its releases can change native behavior and wheel availability. Since only 0.3.0 was validated here, I recommend retaining ==0.3.0. If patch uptake is preferred, >=0.3.0,<0.4 would be the safer upper-bounded alternative.

StafaH: Upper bound seems fine to me, to allow for patches to be pulled

StafaH (approved):

pyproject.toml
mmichelis: Done in #[8478256](https://github.com/isaac-sim/IsaacLab/pull/6585/commits/8478256badd8ee4471117fcd126fb6288e3028d8)

AntoineRichard (approved):

kellyguo11 (approved):

PR #6590, opened 2026-07-17
Harden perf-smoke seeder mount permissions

Neil4561:
## Summary
- write benchmark output inside the container and copy normalized artifacts back to the runner
- isolate writable JIT and Kit caches per run/task/backend while preserving reuse across samples
- use disposable source clones and clean all run-scoped directories to avoid stale ownership on long-lived runners

## Test plan
- [x] Run the complete perf-smoke unit suite (`52 passed, 1 skipped`)
- [x] Run pre-commit hooks on both changed files
- [x] Verify cache reuse, artifact extraction, source reuse, and cleanup with UID-1000 Docker containers

### DIFF
--- a/tools/perf_smoke_test/seed_baselines.py
+++ b/tools/perf_smoke_test/seed_baselines.py
@@ -350,6 +350,11 @@ def _prepare_seed_source(workdir: Path, seed_src_dir: Path, sha: str) -> None:
     subprocess.run(["chmod", "-R", "a+rwX", str(seed_src_dir)], check=False)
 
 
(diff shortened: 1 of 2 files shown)


Neil4561:
@fatimaanes

fatimaanes (approved):
LGTM — solid fix for the L40S seeder output/cache permission blocker. Verified the docker cp + host chmod path is sound (CI image runs as USER isaaclab uid 1000 = runner uid) and per-bucket Kit cache isolation is the right call for the camera-task crash. Approving.

PR #6596, opened 2026-07-18
Fixes Newton cloner ignoring collision approximation

myurasov-nv:
# Description

The Newton cloner throws away whatever `physics:approximation` an asset authors on its collision meshes: it imports USD with `skip_mesh_approximation=True`, then convex-hulls everything. So in cloned envs a concave part always collides as one convex blob, with no warning. Plain scene loading honors the attribute, so the two paths didn't even agree with each other.

Two changes:

- Cloner now lets Newton process authored approximations, and only runs the default convex-hull pass on meshes that don't author one.
- `coacd` is now a core dependency. Without it `convexDecomposition` just warns and falls back to a single hull. Kit-based installs already get it via isaacsim (which pins 1.0.7, hence the floor); this covers kit-less installs.

| Before | After |
|---|---|
| Cloned envs: every authored mode collapses to one convex hull, silently | `convexDecomposition` -> multiple hulls, `boundingSphere` -> sphere, `boundingCube` -> box, `meshSimplification`/`none` -> mesh |
| Plain scenes: `convexDecomposition` falls back to one hull (no `coacd`) | Decomposes out of the box |
| Concave L-prism (volume 3.0) collides as its filled hull (volume 3.5) | Decomposed hulls preserve the exact shape |

Tests: cloner unit tests, plus an install-CI probe that runs after `./isaaclab.sh -i core` and fails on either regression (verified failing before the fix, passing after).

## Reproduction

1. Author a concave collision mesh with `physics:approximation = "convexDecomposition"` (`UsdPhysics.MeshCollisionAPI`).
2. Load it in a cloned env with the Newton backend, or run `python source/isaaclab/test/install_ci/misc/convex_decomposition_probe.py` in an installed environment.
3. Before: one convex hull, concavity filled in. After: multi-hull decomposition on both import paths.

## Type of change

## Checklist


source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
```diff
@@ -10,14 +10,38 @@
...
+        if attr and attr.HasAuthoredValue():
+            authored.add(prim.GetPath().pathString)
+
+    return [
+        index
+        for index, shape_type in enumerate(builder.shape_type)
+        if shape_type == GeoType.MESH
+        and (builder.shape_flags[index] & ShapeFlags.COLLIDE_SHAPES)
+        and builder.shape_label[index] not in authored
+    ]
```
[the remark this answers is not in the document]
myurasov-nv: Hoisted: the authored-prim set is now computed once per stage, outside the source loop (0e10f8e).

source/isaaclab_newton/test/cloner/test_collision_approximation.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_unauthored_mesh_still_simplifies_to_single_hull(self):
+        """Meshes with no authored approximation keep the default convex-hull treatment."""
+        shapes = _collision_shapes(_build(_make_stage(None)))
+        assert len(shapes) == 1
+        assert shapes[0] in (GeoType.MESH, GeoType.CONVEX_MESH)
+
+    def test_simplify_meshes_false_keeps_raw_mesh(self):
+        """simplify_meshes=False leaves an unauthored mesh untouched."""
+        shapes = _collision_shapes(_build(_make_stage(None), simplify_meshes=False))
+        assert shapes == [GeoType.MESH]
```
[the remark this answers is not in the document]
myurasov-nv: Added tests for boundingCube, meshSimplification, and none (0e10f8e).

source/isaaclab/test/install_ci/misc/convex_decomposition_probe.py
```diff
@@ -0,0 +1,124 @@
...
+
+def _collision_shape_count(builder) -> int:
+    """Number of colliding mesh/convex-mesh shapes in the builder."""
+    return sum(
+        1
+        for i, shape_type in enumerate(builder.shape_type)
+        if shape_type in (GeoType.MESH, GeoType.CONVEX_MESH)
+        and builder.shape_flags[i] & ShapeFlags.COLLIDE_SHAPES
+        and builder.shape_source[i] is not None
+    )
```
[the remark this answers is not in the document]
myurasov-nv: Aligned: dropped the shape_source guard from the probe so both count identically (0e10f8e).

source/isaaclab/test/install_ci/cli/test_cli_install_core_in_uvenv_correctness.py
```diff
@@ -10,11 +10,15 @@
     - ./isaaclab.sh -i core -> verify all core packages importable
```
StafaH, line 10: This install test is not required. The install test should at most just check for the coacd dependency is there, the rest is a duplicate of the test in isaaclab_newton made below.

pyproject.toml
```diff
@@ -43,6 +43,11 @@ dependencies = [
     # [all] extra, so depend on that extra there.
     "pytetwild==0.2.3 ; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
     "pytetwild[all]>=0.3.0,<0.4 ; platform_machine == 'aarch64' or platform_machine == 'arm64'",
+    # coacd: Newton's USD importer maps ``physics:approximation = "convexDecomposition"``
+    # to CoACD and silently falls back to a single convex hull when it is missing, which
+    # breaks collision geometry for concave assets. Small MIT wheel, numpy-only deps.
+    # Floor at 1.0.7: isaacsim-core pins coacd==1.0.7, and ``--extra isaacsim`` must co-resolve.
+    "coacd>=1.0.7,<2",
```
StafaH: The list of dependencies for approximate meshes seems longer than coacd, would it not be better to just do newton[sim,importers] below to grab all the optional dependencies required?

https://github.com/newton-physics/newton/blob/main/pyproject.toml#L49

source/isaaclab_newton/test/cloner/test_collision_approximation.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    return builders[_SOURCE]
+
+
+class TestClonerCollisionApproximation:
+    """build_source_builders must honor authored approximations and hull the rest."""
+
+    def test_authored_convex_decomposition_produces_multiple_hulls(self):
+        """A concave mesh authored with convexDecomposition decomposes into 2+ hulls."""
+        pytest.importorskip("coacd", reason="convexDecomposition requires the coacd dependency")
```
StafaH: We don't need to add skip test since the dependency is in our required list, this test should always run

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
```diff
@@ -37,12 +66,19 @@ def build_source_builders(
...
+            # simplify pass below only touches meshes with no authored approximation.
+            skip_mesh_approximation=False,
             schema_resolvers=schema_resolvers,
             ignore_paths=ignore_paths,
         )
         if simplify_meshes:
-            builder.approximate_meshes("convex_hull", keep_visual_shapes=True)
+            builder.approximate_meshes(
+                "convex_hull",
+                shape_indices=_unauthored_collision_mesh_shapes(builder, authored_prim_paths),
```
StafaH: Every test currently builds a stage containing only one collision mesh. The new behavior, however, is selective:

```
shape_indices=_unauthored_collision_mesh_shapes(
    builder,
    authored_prim_paths,
)
```

Add one stage containing at least:

An explicitly authored "none" or "meshSimplification" mesh.
An unauthored mesh.

Then verify that only the unauthored mesh receives the default hull approximation. This tests the actual new algorithm rather than testing each importer mode independently.

source/isaaclab_newton/test/cloner/test_collision_approximation.py
```diff
@@ -0,0 +1,111 @@
...
+        shapes = _collision_shapes(_build(_make_stage("convexDecomposition")))
+        assert len(shapes) >= 2, f"expected a multi-hull decomposition, got {shapes}"
+        assert all(s == GeoType.CONVEX_MESH for s in shapes)
+
+    def test_authored_bounding_sphere_produces_sphere(self):
```
StafaH: nit: These tests could be parameterized
```

@pytest.mark.parametrize(
    ("approximation", "expected"),
    [
        ("boundingSphere", GeoType.SPHERE),
        ("boundingCube", GeoType.BOX),
        ("meshSimplification", GeoType.MESH),
        ("none", GeoType.MESH),
    ],
)
```

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
```diff
@@ -37,12 +66,19 @@ def build_source_builders(
             stage,
             root_path=source,
             load_visual_shapes=True,
-            skip_mesh_approximation=True,
+            # Honor USD-authored ``physics:approximation`` (convexDecomposition,
```
StafaH: Here and elsewhere, let's trim the AI comments:

e.g: # Preserve explicitly authored USD collision approximations.

source/isaaclab_newton/test/cloner/test_collision_approximation.py
```diff
@@ -0,0 +1,111 @@
...
+    def test_authored_none_keeps_raw_mesh(self):
+        """An explicit approximation of "none" keeps the raw trimesh collision."""
+        shapes = _collision_shapes(_build(_make_stage("none")))
+        assert shapes == [GeoType.MESH]
+
+    def test_unauthored_mesh_still_simplifies_to_single_hull(self):
+        """Meshes with no authored approximation keep the default convex-hull treatment."""
+        shapes = _collision_shapes(_build(_make_stage(None)))
+        assert len(shapes) == 1
+        assert shapes[0] in (GeoType.MESH, GeoType.CONVEX_MESH)
```
StafaH: The original input is already one GeoType.MESH, so this passes even when approximate_meshes() does nothing or receives no indices. Combined with:

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
```diff
@@ -10,14 +10,42 @@
...
+    ``meshSimplification`` or ``none`` keeps its source prim's label). The convex-hull
+    simplify pass must leave those alone, otherwise it would override the authored
+    intent; only meshes with no authored approximation get the default hull treatment.
+    """
+    return [
+        index
+        for index, shape_type in enumerate(builder.shape_type)
+        if shape_type == GeoType.MESH
+        and (builder.shape_flags[index] & ShapeFlags.COLLIDE_SHAPES)
+        and builder.shape_label[index] not in authored_prim_paths
```
StafaH: This makes correctness depend on an undocumented importer detail: that every relevant shape_label exactly equals the absolute USD prim path. Labels can be transformed, generated, duplicated, or repurposed without this code failing loudly.

myurasov-nv: Reworked in add8b77: the exclusion now keys on add_usd's returned path_shape_map, no label matching.

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
myurasov-nv: Added in add8b77: mixed stage with an authored none mesh plus an unauthored sibling; only the sibling gets hulled.

source/isaaclab_newton/test/cloner/test_collision_approximation.py
myurasov-nv: Covered by the new mixed-stage test in add8b77: it fails if approximate_meshes receives no indices.

pyproject.toml
myurasov-nv: Done in add8b77: newton[sim,importers] replaces the bare coacd pin; the git override carries the same extras.

source/isaaclab/test/install_ci/cli/test_cli_install_core_in_uvenv_correctness.py
myurasov-nv: Slimmed in add8b77: install CI now only checks the importer deps (import coacd, fast_simplification) after -i core; probe removed.

source/isaaclab_newton/test/cloner/test_collision_approximation.py
myurasov-nv: Dropped the skip in add8b77; the test always runs.

source/isaaclab_newton/test/cloner/test_collision_approximation.py
myurasov-nv: Parameterized in add8b77.

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
myurasov-nv: Trimmed in add8b77.

myurasov-nv:
The rendering-correctness failure was real: the dexsuite kuka hetero test maps different asset variants to different worlds, and honoring their authored approximations left the worlds with differing shape-type sequences, which SolverMuJoCo's homogeneous-worlds validation rejects (this is presumably why the cloner skipped approximations in the first place). 35194c9 detects that case after building the source builders and falls back to the previous uniform convex-hull treatment with a warning; identically authored sources keep their authored modes. Covered by two new unit tests.

myurasov-nv:
d5f2109 fixes the remaining real failures: authored modes are now applied after import (visual shapes preserved), so kit-less ovrtx golden renders are unchanged (verified locally against the goldens on aarch64) and the visualization-builder unit test passes again. Stages with no authored approximation take exactly the pre-existing code path.

myurasov-nv:
CI triage — remaining failures are not from this diff:

- **franka_soft timeout**: known intermittent; also hit #6552 before this branch existed.
- **rendering-correctness-kitless**: re-running the same commit changed the failure set (cartpole passed first run, failed on retry with the same numbers seen on mh/uv_lock_tests); fails on other open PRs too.
- **Local check**: cartpole and shadow_hand kitless pass against goldens with this branch.

The one real rendering failure (dexsuite hetero) was fixed in 35194c9.

StafaH (approved):

PR #6598, opened 2026-07-18
Refactor viz and video-recording related classes. Expand video-recording feature.

### LINKED ISSUE #6605: [Bug Report] Slow offline rendering during training
### Questions

1. Is continuous rendering with `--video` the intended behavior, or is there an existing way to restrict
   rendering (and camera pipeline updates) to the active recording windows only?
2. Is there a recommended alternative for periodic offline visual monitoring under the Newton backend that
   avoids the RTX pipeline entirely (e.g. periodic Newton viewer / Rerun captures)?

### Environment

- Isaac Lab: develop (Newton physics backend, MJWarp solver)
- Isaac Sim: 6.0.x
- Setup: headless training in Docker, RSL-RL
- Task: manager-based humanoid locomotion, `dt=0.005`, `decimation=4`, `render_interval=4`
- Repro: `train.py --task <task> --headless --viz none --video --video_length 200 --video_interval 2000`
  → ~10 s/iteration vs ~1 s/iteration without `--video`

matthewtrepte:
## Summary

This PR refactors visualizer and video-recording related classes and expands & cleans-up the video-recording feature.

ViewerCfg is deprecated. instead we can use Visualizer eye and look_at fields.
ViewportControllerCfg is deprecated and combined with KitVisualizer
Two VideoRecording Config Classes are removed

Video recording is redesigned
- replaced gym RecordVIdeo wrapper with an internal method from VideoRecorder
- eliminated eye and lookat fields from VideoRecorderCfg. the idea now is the video recorder either records a visualizer or sensor (renderer), and doesn't have any camera based fields - those fields should just live with visualizer or renderer configs
- now multiple video recordings can be launched
- can now record tiled camera views in visualziers
- can now specify which sensor to record
- also added a few more VideoRecorderCfg fields, like to keep the last n clips and start recording only after n steps

Added video recorder tutorial script and separated out the video recorder docs from visualizer into its own docs page

Fixed an live plots issue in Rerun

Fixes #6605 

## Screenshots

Example captures from 1 cmd with 4 VideoRecordingCfgs. These examples are from the new docs page.

Kit Visualizer
example_kit_viewport

Newton Visualizer
example_newton_viewport

Kit Visualizer Tiled View
example_tiled_kit_viewport

Kit Sensor (Renderer)
example_sensor

## Test plan

- [ ] `source/isaaclab/test/envs/test_video_recorder.py` — 17/17 unit tests pass (covers all `_select_video_backend` dispatch paths including new Newton fallback)
- [ ] `source/isaaclab/test/sim/test_simulation_context.py` — 29 passed before pre-existing hang on `test_timeline_callbacks_with_weakref` (unrelated)
- [ ] `source/isaaclab/test/envs/test_env_rendering_logic.py` — 17/17 pass
- [ ] `source/isaaclab/test/sim/test_simulation_context_visualizers.py` — 28/28 pass
- [ ] `source/isaaclab_tasks/test/core/test_recording_backends.py` — **5/5 integration tests pass** (Kit/PhysX, Kit/Newton fallback, Newton GL, PhysX renderer, Newton renderer)


AntoineRichard (commented):
AI review: Basic camera positioning remains available through `VisualizerCfg` and `SimulationContext.set_camera_view()`, but this refactor currently removes or regresses several related capabilities. I left inline comments for the runtime UI/tracking regressions, independent recorder camera positioning, and the undeprecated public controller removal. Please also preserve or document a replacement for `ViewerCfg.cam_prim_path`, since the new renderer synchronization hardcodes `/OmniverseKit_Persp`.

source/isaaclab/isaaclab/envs/ui/base_env_window.py
```diff
@@ -378,49 +379,51 @@ def _toggle_recording_animation_fn(self, value: bool):
...
+            viz.cfg.origin_type = "env"
         else:
-            # find which index the asset is
             fancy_names = [name.replace("_", " ").title() for name in self._viewer_assets_options]
-            # store the desired env index
             viewer_asset_name = self._viewer_assets_options[fancy_names.index(value)]
-            # update the camera view
-            vcc.update_view_to_asset_root(viewer_asset_name)
+            viz.cfg.origin_type = "asset_root"
+            viz.cfg.asset_name = viewer_asset_name
```
AntoineRichard: AI review: Changing the origin type now only mutates `viz.cfg.origin_type`; it does not recompute `_viewer_origin` or apply a new camera pose. `KitVisualizer.step()` only refreshes `asset_root` and `asset_body`, so switching to World or Env leaves the viewport at its previous pose. The env-index callback has the same issue in Env mode. Could `KitVisualizer` expose a public origin/tracking setter that validates the selection, recomputes the origin, and updates the camera immediately, with these UI callbacks calling that method?

source/isaaclab/isaaclab/envs/ui/base_env_window.py
```diff
@@ -378,49 +379,51 @@ def _toggle_recording_animation_fn(self, value: bool):
...
-            raise ValueError("Viewport camera controller is not initialized! Please check the rendering mode.")
-        # obtain the camera locations and set them in the viewpoint camera controller
+        viz = self._get_kit_visualizer()
+        if viz is None:
+            return
         eye = [self.ui_window_elements["viewer_eye"][i].get_value_as_float() for i in range(3)]
         lookat = [self.ui_window_elements["viewer_lookat"][i].get_value_as_float() for i in range(3)]
-        # update the camera view
-        vcc.update_view_location(eye, lookat)
+        viz.set_camera_view(eye, lookat)
```
AntoineRichard, line 446: AI review: `set_camera_view()` applies these values once but does not persist them to `viz.cfg.eye` and `viz.cfg.lookat`. During asset tracking, the next `KitVisualizer.step()` calls `_apply_viewer_origin_to_camera()` with the old configured offsets and immediately undoes the UI edit. The previous controller stored new eye/look-at offsets before applying them. Please persist the relative offsets here, ideally through a public visualizer API so UI and programmatic changes share the same semantics.

source/isaaclab/isaaclab/envs/utils/video_recorder_cfg.py
```diff
@@ -3,65 +3,51 @@
...
-    class_type: type = VideoRecorder
-    """Recorder class to instantiate; must accept ``(cfg, scene)``."""
-
-    env_render_mode: str | None = None
-    """Gym render mode forwarded from the environment constructor (``"rgb_array"`` when ``--video`` is active).
-
-    Set automatically by the environment base classes; do not set manually.
-    """
-
-    eye: tuple[float, float, float] = (7.5, 7.5, 7.5)
```
AntoineRichard, line 35: AI review: Removing `eye` and `lookat` makes the recorder passive, but it also removes the ability to configure a recording camera independently from the interactive viewer. A named sensor is a possible workaround, although that requires modifying the scene and is not equivalent to a recorder-owned perspective camera. Is this feature loss intentional? If so, the migration guidance should show how to create and select an independently positioned camera source; otherwise these fields need an equivalent on a recording-source configuration.

source/isaaclab/isaaclab/envs/ui/viewport_camera_controller.py
```diff
@@ -1,243 +0,0 @@
...
-import numpy as np
-import torch
-
-from isaaclab.assets.articulation.articulation import Articulation
-
-if TYPE_CHECKING:
-    from isaaclab.envs import DirectRLEnv, ManagerBasedEnv, ViewerCfg
-
-
-class ViewportCameraController:
```
AntoineRichard, line 22: AI review: `ViewportCameraController` is an exported public class with runtime methods for selecting an environment and switching among world, env, asset-root, and asset-body tracking. I could not find a prior deprecation for this class, and the replacement tracking operations in `KitVisualizer` are private. Under the repository policy that public API removals require deprecation in a prior release, please retain a deprecated compatibility shim and add a public replacement API before removing this class.

AntoineRichard:
@matthewtrepte please look into: https://github.com/isaac-sim/IsaacLab/issues/6605. Would this PR help with this?

matthewtrepte:
> @matthewtrepte please look into: #6605. Would this PR help with this?

@AntoineRichard yep this should be resolved in the redesign, thanks for sharing

source/isaaclab/isaaclab/envs/ui/viewport_camera_controller.py
matthewtrepte: added deprecation

source/isaaclab/isaaclab/envs/utils/video_recorder_cfg.py
matthewtrepte: this is intentional. we can keep the eye and lookat fields to just the visualizer and sensor cameras. then the videorecorder can just capture what they are pointed to.

matthewtrepte:
@isaaclab-review-bot review

docs/source/_static/how-to/record_video/example_kit_viewport.gif
kellyguo11, line 1: could we upload gifs to s3 instead?

matthewtrepte:
@isaaclab-review-bot review

source/isaaclab_tasks/isaaclab_tasks/contrib/assemble_trocar/g129_dex3_env_cfg.py
```diff
@@ -397,6 +392,7 @@ def __post_init__(self):
         self.sim.dt = 1 / 200
         self.sim.render_interval = self.decimation
         self.sim.physics = PhysxCfg(bounce_threshold_velocity=0.01)
+        self.sim.default_visualizer_cfg = VisualizerCfg(eye=(-0.5, 2.4, 1.6), lookat=(-5.4, 0.2, -1.2))
```
kellyguo11, line 395: let's document this in the migration to 3.0 doc page and skills

kellyguo11 (approved):
let's make sure we capture all the breaking changes in the migration guide

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,90 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
+            path.  Use ``"train"`` for training runs and ``"play"`` for evaluation.
+    """
+    if not getattr(args_cli, "video", False):
+        return
+
+    from isaaclab.envs.utils.video_recorder_cfg import VideoRecorderCfg
+
+    existing: list = getattr(env_cfg, "video_recorders", []) or []
+    if not existing:
+        env_cfg.video_recorders = [VideoRecorderCfg()]
```
kellyguo11: **[P1] Default `--video` records no frames for normal task configs.** `VideoRecorderCfg()` defaults to `source="visualizer"`, but the existing task configs do not activate a visualizer in `sim.visualizer_cfgs`. `_frame_from_visualizer()` therefore raises on the first step, `_get_frame()` catches it and permanently suppresses further capture, and no clip is written even though the CLI prints "Video recording enabled." Please either activate a capture-capable visualizer for this fallback recorder, retain a renderer fallback, or select a configured sensor/backend.

source/isaaclab_experimental/isaaclab_experimental/envs/manager_based_env_warp.py
```diff
@@ -76,6 +74,14 @@ def __init__(self, cfg: ManagerBasedEnvCfg):
         cfg.validate()
         # store inputs to class
         self.cfg = cfg
+        # Video recording is not supported on Warp environments.
```
kellyguo11, line 77: **[P1] `--video --frontend warp` no longer produces videos.** `apply_video_recording()` installs `cfg.video_recorders` for every frontend, while `wrap_record_video()` is now a no-op. Manager-based Warp only warns that no clips will be written, and direct Warp silently ignores the recorder, so the documented CLI flag stops working for the restored Warp frontend. Please retain a working Warp capture path or reject/disable the combination before reporting that recording is enabled.

source/isaaclab/isaaclab/envs/ui/viewport_camera_controller.py
```diff
@@ -3,241 +3,40 @@
...
-from isaaclab.assets.articulation.articulation import Articulation
+import warnings
+from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from isaaclab.envs import DirectRLEnv, ManagerBasedEnv, ViewerCfg
+    from isaaclab.envs import ViewerCfg
 
 
 class ViewportCameraController:
```
kellyguo11, line 36: **[P1] This is not a behavior-preserving compatibility shim.** The class now only emits a warning; public methods such as `set_view_env_index()`, `update_view_to_world()`, and the asset tracking methods are gone, so existing callers fail during the purported deprecation period. Isaac Lab's API policy requires a deprecation before removing public behavior. Please keep these methods functional by forwarding to the replacement for at least the deprecation window, then remove the class in a later release.

source/isaaclab/isaaclab/envs/common.py
```diff
@@ -46,71 +48,133 @@ class ViewerCfg:
...
+    viewer = getattr(env_cfg, "viewer", None)
+    if viewer is None:
+        return
+
+    _defaults = ViewerCfg()
+    eye_changed = not _viewer_cfg_field_matches_default(viewer.eye, _defaults.eye)
+    lookat_changed = not _viewer_cfg_field_matches_default(viewer.lookat, _defaults.lookat)
+    origin_changed = viewer.origin_type != _defaults.origin_type
+
+    if not (eye_changed or lookat_changed or origin_changed):
```
kellyguo11: **[P2] Resolution-only legacy overrides are silently discarded here.** A caller that changes only `ViewerCfg.resolution` receives the deprecation warning from `ViewerCfg.__post_init__()`, but this condition returns before `window_width`/`window_height` are forwarded. `cam_prim_path` is also never forwarded. Please include every supported non-default legacy field in the detection and translation, or explicitly reject fields that cannot be preserved.

source/isaaclab/isaaclab/envs/utils/video_recorder.py
```diff
@@ -3,136 +3,273 @@
...
-        if self._use_newton_visualizer and self._capture is None:
-            visualizer = next(
-                (
-                    visualizer
-                    for visualizer in self._scene.sim.visualizers
-                    if visualizer.cfg.visualizer_type == "newton"
-                ),
-                None,
+        viz = candidates[0]
+        if sub == "tiled":
```
kellyguo11, line 190: **[P2] The documented `visualizer:newton:tiled` source cannot work.** This branch requires `render_tiled_rgb_array()`, but the Newton visualizer only implements `render_rgb_array()`. The resulting exception is caught by `_get_frame()` and permanently disables this stream, so no clip is produced. Please treat `newton:tiled` as the normal Newton tiled framebuffer, implement the tiled method, or remove this source from the configuration docs.

source/isaaclab/isaaclab/envs/utils/video_recorder.py
```diff
@@ -3,136 +3,273 @@
...
-            self._capture = create_isaacsim_kit_perspective_video(
-                IsaacsimKitPerspectiveVideoCfg(
-                    eye=eye,
-                    lookat=lookat,
-                    window_width=cfg.window_width,
-                    window_height=cfg.window_height,
+        # Kit Replicator requires cubric to propagate Newton Fabric transforms to RTX's
+        # scene delegate. Without cubric, frames will be black. Log a warning but allow
+        # the capture to proceed — users with cubric available will get correct frames.
+        if viz_type == "kit":
```
kellyguo11, line 180: **[P2] Kit+Newton behavior contradicts the new documentation and changelog.** Those state that the combination logs an error and produces no clip, while this implementation only warns and continues capturing (potentially writing black frames). Please make the implementation and user-facing documentation agree on whether capture is supported.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,114 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
-        "video_folder": os.path.join(log_dir, "videos", "train"),
-        "step_trigger": lambda step: step % args_cli.video_interval == 0,
-        "video_length": args_cli.video_length,
-        "disable_logger": True,
-    }
-    print("[INFO] Recording videos during training.")
-    print_dict(video_kwargs, nesting=4)
-    return gym.wrappers.RecordVideo(env, **video_kwargs)
+        _logging.getLogger(__name__).warning(
+            "wrap_record_video() is no longer functional — recording is now driven inside env.step(). "
```
kellyguo11, line 670: [P1] Keep benchmark `--video` functional

All four benchmark trainers still call `_common.wrap_record_video()` after creating the environment (`benchmark_train_rsl_rl.py`, `benchmark_train_rl_games.py`, `benchmark_train_skrl.py`, and `benchmark_train_sb3.py`). This stub now returns the environment unchanged, so their existing `--video` option silently produces no clips while the benchmark result bundle still advertises a videos directory. Please migrate those entrypoints to `apply_video_recording()` before launch/environment creation, or keep this wrapper functional during the transition.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,114 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
+        env_cfg.video_recorders = [VideoRecorderCfg()]
+        # Auto-inject a default visualizer so source="visualizer" can resolve a capture target.
+        sim_cfg = getattr(env_cfg, "sim", None)
+        if sim_cfg is not None:
+            existing_viz = getattr(sim_cfg, "visualizer_cfgs", None) or []
+            if not (existing_viz if isinstance(existing_viz, list) else [existing_viz]):
+                try:
+                    from isaaclab_visualizers.kit import KitVisualizerCfg as _KitCfg
+
+                    sim_cfg.visualizer_cfgs = [_KitCfg()]
```
kellyguo11: [P1] Inject the capture visualizer before the runtime scan

The RL backends call `apply_video_recording()` from inside `with launch_simulation(env_cfg, args_cli)`, after the launcher has already scanned the config, decided whether Kit is required, and resolved the automatic backends. Injecting `KitVisualizerCfg` here is therefore invisible to launch selection: a nominally kitless/default run can remain kitless and produce no recording, `auto_rtx` can remain on `ovrtx` instead of autoswitching to `isaacsim_rtx`, and `PhysxAutoCfg` can remain on `ovphysx` instead of autoswitching to `isaacsim_physx` despite the newly introduced Kit requirement. Please configure recording before `launch_simulation()` (and cover both autoswitch pairs in tests).

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -568,12 +554,12 @@ def render(self, recompute: bool = False) -> np.ndarray | None:
         if not self.has_rtx_sensors and not recompute:
             self.sim.render()
         # decide the rendering mode
+        if self.render_mode == "rgb_array":
```
kellyguo11, line 557: [P1] Deprecate `rgb_array` before removing it

This turns a previously supported public Gym API into `NotImplementedError`. Existing `gym.make(..., render_mode="rgb_array")`, `env.render()`, Gym `RecordVideo`, and custom frame-consuming wrappers will fail immediately. Isaac Lab's compatibility policy requires a deprecation release before removing public API behavior; please retain `rgb_array` in metadata and forward it through a deprecated adapter to the new capture path for that window. The same break exists in `DirectMARLEnv` and `ManagerBasedRLEnv`.

source/isaaclab/isaaclab/envs/ui/viewport_camera_controller.py
```diff
@@ -3,241 +3,158 @@
...
-
-            set_kit_renderer_camera_view(eye=cam_eye, target=cam_target, camera_prim_path=self.cfg.cam_prim_path)
-        except (ImportError, ModuleNotFoundError):
-            pass
-
-    """
-    Private Functions
-    """
+        for viz in self._get_kit_visualizers():
+            viz.set_camera_view(
```
kellyguo11: [P1] Preserve origin-relative behavior in the compatibility shim

`update_view_location()` previously added the current `viewer_origin` to `eye` and `lookat`; this now forwards those offsets as absolute world coordinates. Existing code that selects an environment or asset origin and then updates the camera location will jump back toward the world origin. The constructor also no longer applies the configured initial origin/tracking behavior. Since this class is being retained as the deprecation shim, please preserve its observable behavior by forwarding through the Kit visualizer's origin-aware path (or keep the old implementation for the deprecation window).

docs/source/how-to/record_video.rst
```diff
@@ -1,211 +1,255 @@
...
-* Direct RL, Direct MARL and manager-based RL environments copy the task's
-  :class:`~isaaclab.envs.common.ViewerCfg` ``eye`` and ``lookat`` into those fields before the
-  recorder is constructed, so training clips align with the task's intended viewport when
-  ``origin_type`` is ``"world"``.
+No visualizer window opens.  Frames are read directly from the ``tiled_camera`` sensor,
+writing one clip to ``videos/recording_tutorial/example_2/sensor_0000.mp4``.
 
+``source="sensor:tiled_camera"`` refers to the key under which the camera is registered
+in ``env.scene.sensors``.  The sensor must have ``"rgb"`` in its ``data_types``; other
+channels can be recorded by appending the channel name (e.g. ``"sensor:tiled_camera:depth"``).
```
kellyguo11: [P2] Do not document unsupported sensor channels

The recorder's sensor path currently always reads `sensor.data.output["rgb"]`; it does not parse or use the third `source` component. Consequently, `source="sensor:tiled_camera:depth"` either records RGB or disables the stream when RGB is unavailable; it cannot record depth as documented. Please remove this claim or implement channel selection plus conversion of depth data to an encodable frame, with a test.

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -140,6 +146,7 @@ def initialize(self, scene_data_provider: SceneDataProvider) -> None:
             ],
         )
         self._setup_camera_sensor_view(num_envs)
+        self._setup_initial_camera_view()
```
kellyguo11: [P1] Apply the initial camera pose after initialization

`_setup_initial_camera_view()` calls `_apply_viewer_origin_to_camera()`, which calls `set_camera_view()`, but `set_camera_view()` returns while `_is_initialized` is false; the flag is only set on the next line. The optional `isaaclab_physx` synchronization helper masks this in some installations, but without that package the configured world/env `eye` and `lookat` are silently dropped. Please set the initialized state before applying the pose, or use an unguarded private setter during initialization.

kellyguo11 (commented):
Inline findings from the latest review.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,114 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
+        subdir: Sub-directory name appended to ``<log_dir>/videos/`` for the fallback output
+            path.  Use ``"train"`` for training runs and ``"play"`` for evaluation.
+    """
+    if not getattr(args_cli, "video", False):
+        return
 
-    Returns:
-        The original or video-wrapped environment.
+    frontend = getattr(args_cli, "frontend", "torch") or "torch"
+    if frontend != "torch":
```
kellyguo11, line 604: [P1] Preserve Warp video recording and configured RGB output

This unconditional rejection removes the previously working `--video --frontend warp` path from the unified CLI without a deprecation period. It also leaves the experimental Warp rendering path in a regressed state: both `DirectRLEnvWarp` and `ManagerBasedRLEnvWarp` still expose `rgb_array`, but this PR replaces `cfg.viewer.cam_prim_path` and `cfg.viewer.resolution` with `/OmniverseKit_Persp` and `(1280, 720)`, so existing Warp clients receive a different camera and image shape even outside the CLI recorder. Please retain a working Warp recording path (or deprecate it first) and preserve the configured camera/resolution, potentially through the new visualizer configuration compatibility layer.

source/isaaclab/isaaclab/envs/utils/video_recorder_cfg.py
```diff
@@ -3,65 +3,101 @@
...
+    output_dir: str | None = None
+    """Directory for output mp4 files (created on demand).
+
+    ``None`` (default): when recording is enabled via ``--video``, the RL entrypoint sets
+    this to ``<log_dir>/videos/<subdir>`` automatically.  Set an explicit path to override.
     """
 
-    lookat: tuple[float, float, float] = (0.0, 0.0, 0.0)
-    """Perspective camera look-at target in world space (metres).
+    fps: int = 30
```
kellyguo11: [P1] Preserve automatic FPS from the environment step rate

The old Gym `RecordVideo` path used `env.metadata["render_fps"]`, which these environments set to `1 / step_dt`, so default recordings played back at the simulation's effective step rate. Making the recorder default unconditionally to 30 FPS changes playback speed for every task whose `step_dt` is not `1 / 30`, including the default `--video` path. Please keep explicit `fps` overrides, but make the default automatic (for example, `fps=None` resolved from `env.metadata["render_fps"]` or `1 / env.step_dt` when the recorder is constructed).

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,114 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
+            f"--video is not supported with --frontend {frontend!r}. "
+            "Video recording requires the standard torch frontend. "
+            "Remove --video or switch to --frontend torch."
+        )
+
+    from isaaclab.envs.utils.video_recorder_cfg import VideoRecorderCfg
+
+    existing: list = getattr(env_cfg, "video_recorders", []) or []
+    if not existing:
+        env_cfg.video_recorders = [VideoRecorderCfg()]
```
kellyguo11: [P1] Preserve the default cadence of bare `--video`

Previously, the unified RL CLI defaulted `--video_interval` to 2000, so an unchanged `--video` command recorded a clip immediately and then recurring clips every 2,000 environment steps. The parser now defaults that option to `None`, and this fallback constructs `VideoRecorderCfg()` whose `video_interval=0`, silently changing bare `--video` to a single initial clip. Please preserve the historical cadence for CLI-created fallback recorders (for example, construct this one as `VideoRecorderCfg(video_interval=2000)`) while continuing to leave task-defined recorder configs untouched and honoring explicit CLI overrides.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -563,34 +570,107 @@ def wrap_sensor_capture(env: gym.Env, log_dir: str, args_cli: argparse.Namespace
...
+        env_cfg.video_recorders = [VideoRecorderCfg(video_interval=2000)]
+        # Set a default visualizer hint so source="visualizer" can resolve a capture target.
+        # Use default_visualizer_cfg rather than visualizer_cfgs so the launcher's backend
+        # selection (auto_rtx, PhysxAutoCfg) remains visible to launch_simulation.
+        sim_cfg = getattr(env_cfg, "sim", None)
+        if sim_cfg is not None and getattr(sim_cfg, "default_visualizer_cfg", None) is None:
+            try:
+                from isaaclab_visualizers.kit import KitVisualizerCfg as _KitCfg
+
+                sim_cfg.default_visualizer_cfg = _KitCfg()
```
kellyguo11, line 625: `default_visualizer_cfg` is only a hint applied to a visualizer that was already selected; it does not add one to `sim.visualizers`. The RL entrypoints call this after entering `launch_simulation(...)`, so launch-time visualizer intent has already been resolved and plain `--video` still reaches `source=visualizer` with no capture target. Please establish the appropriate capture visualizer before launch while preserving the existing `ovrtx`/`isaacsim_rtx` and `ovphysx`/`isaacsim_physx` autoswitch selection rather than hard-coding Kit.

source/isaaclab_experimental/isaaclab_experimental/envs/manager_based_env_warp.py
```diff
@@ -76,6 +74,14 @@ def __init__(self, cfg: ManagerBasedEnvCfg):
...
+        if getattr(cfg, "video_recorders", None):
+            import logging as _logging
+
+            _logging.getLogger(__name__).warning(
+                "cfg.video_recorders is set but ManagerBasedEnvWarp does not support VideoRecorder. "
```
kellyguo11, line 82: The unified CLI now accepts `--frontend warp --video`, but this path only warns and writes no clip; direct Warp environments ignore the recorder configuration as well. That turns an explicit unsupported invocation into a successful-looking run with missing output. Please wire `VideoRecorderCfg` through both Warp environment families, or keep a clear frontend-level rejection until recording is actually supported. The Warp `rgb_array` fallback also still hard-codes `/OmniverseKit_Persp` at 1280×720, so it is not a configuration-preserving compatibility path.

source/isaaclab/isaaclab/envs/direct_rl_env.py
```diff
@@ -568,12 +554,18 @@ def render(self, recompute: bool = False) -> np.ndarray | None:
...
+            import warnings
+
+            warnings.warn(
+                "render_mode='rgb_array' is deprecated and will be removed in a future release. "
+                "Use VideoRecorderCfg on env_cfg.video_recorders to capture frames instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            return None
```
kellyguo11, line 566: `rgb_array` is still advertised in `metadata[render_modes]`, but this branch now returns `None` for it. Emitting a deprecation warning does not preserve the API contract: existing Gym clients expect an RGB ndarray and will fail immediately. Please continue returning frames during the deprecation period, and remove the mode from metadata only when the behavior is actually removed in a later release. The same issue exists in the manager-based and MARL base environments.

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_cloth_env_cfg.py
```diff
@@ -271,6 +271,14 @@ def __post_init__(self) -> None:
         self.sim.dt = 1 / 60.0
         self.sim.render_interval = self.decimation
 
+        # Hint for the viewport camera when running interactively with --viz kit.
+        # Using default_visualizer_cfg rather than visualizer_cfgs avoids forcing
+        # Kit viewport creation in kitless/headless contexts.
+        from isaaclab_visualizers.kit import KitVisualizerCfg
+
+        self.sim.default_visualizer_cfg = KitVisualizerCfg(
```
kellyguo11, line 279: `SimulationCfg` does not declare `default_visualizer_cfg` as a config/dataclass field. Dynamic assignment works on this instance, but `SimulationCfg(default_visualizer_cfg=...)` raises `TypeError`, and `to_dict()`/Hydra serialization omit the value. Since this is now documented and used as public configuration throughout the branch, please add a typed field such as `default_visualizer_cfg: VisualizerCfg | None = None` to `SimulationCfg`.

source/isaaclab/isaaclab/envs/utils/video_recorder.py
```diff
@@ -3,136 +3,284 @@
...
+            return
+        try:
+            os.makedirs(self._effective_output_dir(), exist_ok=True)
+            path = self._clip_path(self._clip_index)
+            fps = self.cfg.fps
+            if fps is None:
+                fps = self._env.metadata.get("render_fps") if hasattr(self._env, "metadata") else None
+                if fps is None:
+                    step_dt = getattr(self._env, "step_dt", None)
+                    fps = round(1.0 / step_dt) if step_dt else 30
```
kellyguo11: Automatic FPS should account for `frame_stride`. Frames are captured every `frame_stride` environment steps, so using `render_fps` or `1 / step_dt` unchanged makes `frame_stride=2` play at twice real simulation speed. Unless time-lapse behavior is explicitly intended, resolve automatic FPS as `render_fps / frame_stride` (or `1 / (step_dt * frame_stride)`) and add coverage for this case.

source/isaaclab/isaaclab/envs/ui/viewport_camera_controller.py
```diff
@@ -3,241 +3,165 @@
...
-    """
+        for viz in self._get_kit_visualizers():
+            # Preserve the historical origin-relative behaviour: eye and lookat are
+            # offsets from the current viewer origin, not absolute world coordinates.
+            origin = getattr(viz, "_viewer_origin", None)
+            if origin is not None:
+                origin_np = origin.detach().cpu().numpy()
+            else:
+                origin_np = np.zeros(3)
+            viz.set_camera_view(
```
kellyguo11: This updates the absolute viewport pose once but does not update `viz.cfg.eye` and `viz.cfg.lookat`. When `origin_type=asset`, the next visualizer step recomputes the camera from those stale config offsets and overwrites this call; later `reapply_origin()` calls do the same. Please persist the relative offsets into the visualizer config and reapply the origin so the deprecated method preserves its historical behavior.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1790,6 +1790,8 @@ def rendering_test_franka_cloth(
     if renderer == "ovrtx_renderer" and data_type == "instance_segmentation":
         pytest.skip("instance_segmentation crashes with the OVRTX renderer on franka_cloth (NVBUG#6463802).")
 
+    _skip_if_newton_motion_vectors(physics_backend, data_type)
```
kellyguo11: This removes the Franka/Newton motion-vector lane that was passing before the PR-specific visualizer change. The observed regression appeared only on this branch, and the subsequent switch from a forced Kit visualizer to `default_visualizer_cfg` is intended to correct that root cause. Please retain the coverage and verify the fix, or provide evidence and a tracked backend issue showing that this exact lane is independently nondeterministic before skipping it.

source/isaaclab_tasks/test/core/test_video_recording.py
```diff
@@ -0,0 +1,255 @@
...
+import tempfile
+
+import numpy as np
+import pytest
+import torch
+
+try:
+    from moviepy.editor import VideoFileClip
+except ImportError:
+    pytest.skip("moviepy is not installed; install with: pip install 'moviepy<2'", allow_module_level=True)
```
kellyguo11: The optional-dependency check happens after `AppLauncher` has already started SimulationApp at module-import time. When MoviePy is absent, collection skips the module only after launching Kit, with no cleanup path, which can interfere with later test-module collection and defeats the purpose of a graceful skip. Please check/import MoviePy before constructing `AppLauncher`.

source/isaaclab_rl/test/test_apply_video_recording.py
```diff
@@ -0,0 +1,104 @@
...
+def test_apply_video_recording_uses_cfg_defaults_when_cli_not_passed():
+    """video=True without --video_length/--video_interval keeps VideoRecorderCfg defaults."""
+    from isaaclab.envs.utils.video_recorder_cfg import VideoRecorderCfg
+
+    defaults = VideoRecorderCfg()
+    env_cfg = _env_cfg()
+    apply_video_recording(env_cfg, "/my/log", _args())  # no video_length/interval
+    rec = env_cfg.video_recorders[0]
+    assert rec.video_length == defaults.video_length  # default kept
+    assert rec.video_interval == defaults.video_interval  # default kept
```
kellyguo11: This assertion currently fails on the branch: the fallback recorder is explicitly created with `video_interval=2000` to preserve historical no-override CLI behavior, while `VideoRecorderCfg()` defaults to `0`. Please update this test and its docstring to assert the intended historical CLI interval, or change the implementation if `0` is the desired behavior. As written, the targeted suite has one deterministic failure.

PR #6602, opened 2026-07-18
Opt-in Ovstage integration into OVRTX Renderer

rilei-nvidia:
# Description

A opt-in ovstage implementation to ovrtx that can be enabled via `ISAAC_LAB_OVRTX_USE_OVSTAGE=1`.  

Ovrtx-only methods have been suffixed with `_legacy`, and ovrtx+ovstage methods suffixed with `_ovstage`

Notable differences in the ovstage code path:
1. Lack of zero-copy writes to xform and points 
    - Will be addressed in ovstage 0.1.1
2. Stage.clone expects clone targets to not exist at all
    - Changed order of operations and how we export the stage to string pre-clone
3. Lack of PrimMode.EXISTING_ONLY and MUST_EXIST.  Stage.write_attributes default to PrimMode.UPSERT which will write if present, otherwise create.

Render correctness tests now include ovstage variant.  So they run both the ovrtx-only codepath, and ovrtx+ovstage code path.

## Type of change

- Documentation update

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -42,14 +42,6 @@ import isaaclab.utils.warp  # noqa: F401  # initializes Warp runtime
 _OVSTAGE_AVAILABLE = False
 try:
     import ovstage
-    from ovstage import (
-        AttributeSemantic,
-        DLDataType,
-        DLDataTypeCode,
-        PopulationDomain,
-        make_dltensor,
-        population,
-    )
 
     _OVSTAGE_AVAILABLE = True
 except ModuleNotFoundError:
@@ -136,7 +128,7 @@ _USE_OVSTAGE_ENV = "ISAAC_LAB_OVRTX_USE_OVSTAGE"
 if _OVSTAGE_AVAILABLE:
     # DLDataType for a 4×4 double matrix (omni:xform column). ovstage stores omni:xform
     # as one 16-lane float64 element per prim; wp.mat44d maps to the same layout via __dlpack__.
-    _OVSTAGE_XFORM_DTYPE = DLDataType(code=DLDataTypeCode.kDLFloat, bits=64, lanes=16)
+    _OVSTAGE_XFORM_DTYPE = ovstage.DLDataType(code=ovstage.DLDataTypeCode.kDLFloat, bits=64, lanes=16)
 
     def _xform_tensor_from_numpy(xforms: np.ndarray) -> Any:
         """Wrap a ``(N, 4, 4)`` float64 array as a 16-lane DLTensor for ``omni:xform`` writes.
@@ -148,12 +140,12 @@ if _OVSTAGE_AVAILABLE:
             A :class:`ovstage.DLTensor` with shape ``[N]`` and ``lanes=16``.
         """
         flat = np.ascontiguousarray(xforms, dtype=np.float64).reshape(-1)
-        return make_dltensor(flat, dtype=_OVSTAGE_XFORM_DTYPE, shape=[xforms.shape[0]])
+        return ovstage.make_dltensor(flat, dtype=_OVSTAGE_XFORM_DTYPE, shape=[xforms.shape[0]])
 
     # DLDataType for a float32 3-vector (``points`` column). ovstage stores ``point3f[] points``
     # as one 3-lane float32 element per vertex; a warp ``vec3f`` array exports as ``(N, 3)`` lanes=1
     # via DLPack, so a lanes=3 override on a host numpy array is required to match the column.
-    _OVSTAGE_POINT_DTYPE = DLDataType(code=DLDataTypeCode.kDLFloat, bits=32, lanes=3)
+    _OVSTAGE_POINT_DTYPE = ovstage.DLDataType(code=ovstage.DLDataTypeCode.kDLFloat, bits=32, lanes=3)
 
     def _points_tensor_from_numpy(points: np.ndarray) -> Any:
         """Wrap an ``(N, 3)`` float32 array as a 3-lane DLTensor for ``points`` writes.
@@ -165,7 +157,7 @@ if _OVSTAGE_AVAILABLE:
             A :class:`ovstage.DLTensor` with shape ``[N]`` and ``lanes=3``.
         """
         flat = np.ascontiguousarray(points, dtype=np.float32).reshape(-1)
-        return make_dltensor(flat, dtype=_OVSTAGE_POINT_DTYPE, shape=[points.shape[0]])
+        return ovstage.make_dltensor(flat, dtype=_OVSTAGE_POINT_DTYPE, shape=[points.shape[0]])
 
 
 def ovrtx_use_ovstage_enabled() -> bool:
@@ -1610,11 +1602,11 @@ class OVRTXRenderer(BaseRenderer):
         self._stage_paths = self._ovstage_exit_stack.enter_context(ovstage.PathDictionary(self._stage))
         # Ordinal 0 is the empty/unwritten state in ovstage; the first write must use >= 1.
         self._current_ordinal += 1
-        population.open_usd_from_string(
+        ovstage.population.open_usd_from_string(
             self._stage,
             combined_usd_string,
             ordinal=self._current_ordinal,
-            domains=PopulationDomain.RENDERING,
+            domains=ovstage.PopulationDomain.RENDERING,
         )
 
         if num_envs > 1:
@@ -1640,7 +1632,7 @@ class OVRTXRenderer(BaseRenderer):
                 ordinal=self._current_ordinal,
                 tensors=camera_target_ids,
                 is_array=True,
-                semantic=AttributeSemantic.RELATIONSHIP_PATH_ID,
+                semantic=ovstage.AttributeSemantic.RELATIONSHIP_PATH_ID,
             ).wait()
         self._stage_paths.destroy_path_list(render_product_paths)
 
@@ -1753,7 +1745,7 @@ class OVRTXRenderer(BaseRenderer):
             ordinal=self._current_ordinal,
             tensors=_xform_tensor_from_numpy(env_root_xforms),
             is_array=False,
-            semantic=AttributeSemantic.MATRIX,
+            semantic=ovstage.AttributeSemantic.MATRIX,
         ).wait()
         self._env_root_xforms = None
         logger.info("Restored per-env root transforms after cloning")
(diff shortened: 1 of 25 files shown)


source/isaaclab_ov/pyproject.toml
```diff
@@ -19,6 +19,7 @@ requires-python = ">=3.12"
 # Only interpackage dependencies should go here.
 dependencies = [
     "isaaclab",
+    "ovstage>=0.1.0,<0.2.0",
```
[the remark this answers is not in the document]
rilei-nvidia: Yeah good call.  Before we actually merge this PR we may want to back out that this change and also remove it from the "all" selector, so that this can go in and folks can install ovstage separately

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -127,6 +167,19 @@ def _read_gpu_transforms_enabled() -> bool:
...
+def ovrtx_use_ovstage_enabled() -> bool:
+    """Return whether the ovstage scene-ownership path should be used.
+
+    Requires both ``ISAAC_LAB_OVRTX_USE_OVSTAGE=1`` and ovstage being importable
+    (``_OVSTAGE_AVAILABLE``). Defaults to ``0`` so existing deployments are unaffected
+    until ovstage is explicitly opted into.
+    """
+    value = os.environ.get(_USE_OVSTAGE_ENV, "0").strip()
+    if value not in {"0", "1"}:
+        raise ValueError(f"Invalid value for environment variable `{_USE_OVSTAGE_ENV}`: {value}. Expected 0 or 1.")
```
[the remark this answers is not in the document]
rilei-nvidia: Great call.  Will push up a change to throw a RuntimeError when a user sets that environment variable to 1 but ovstage is not importable.

rilei-nvidia:
@greptile-apps please review!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -35,6 +36,25 @@
...
+try:
+    import ovstage
+    from ovstage import (
+        AttributeSemantic,
+        DLDataType,
+        DLDataTypeCode,
+        PopulationDomain,
+        make_dltensor,
+        population,
+    )
```
huidongc: Why import entire ovstage then do another selective import from ovstage?

source/isaaclab_ov/pyproject.toml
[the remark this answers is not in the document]
huidongc: Yeah, we should remove this hard dep.

huidongc:
Marking the default code path "legacy" sounds a little contradictory. I think we'd better do the "legacy" renaming when we make ovstage code path the opt-out. We are not convinced yet the ovstage code path will be the default for GA.

Re the dispatch approach, instead of still mixing scene data management inside the OVRTX renderer class, I wonder maybe we could create a cleaner design:

1. Defne a `OVStageScene` class to which we will delegate all the scene data loading/updating calls.
2. When opting to use the `ovstage` code path, every existing method of OVRTXRenderer just redirects the scene management to the new class:

```py
class OVRTXRenderer:

    def do_something(self, foo, bar):
        if self._ovstage_scene is not None:
            self._ovstage_scene.do_something(foo, bar)
            return

        # update scene data using deprecated ovrtx APIs.
       ...
```

It seems to me this approach has a clean separation of scene data management vs. rendering and is also achievable as a short term goal.

For long term, maybe we can explore to have the simulation side to create/update the ovstage scene, then the renderer class just takes the ovstage scene for rendering, this will then enable single shared source of truth between sim and render. But it is likely a question we will try to answer after GA, probably together with the scene data provider refactoring to replace newton manager usage in the renderer class.

huidongc:
image

rilei-nvidia:
Hey @huidongc - thanks for taking a look 

> Marking the default code path "legacy" sounds a little contradictory. I think we'd better do the "legacy" renaming when we make ovstage code path the opt-out. We are not convinced yet the ovstage code path will be the default for GA.

I went through a few different names and chose `legacy` because those ovrtx APIs as of 0.4 are now deprecated.   `legacy` was a clear and short suffix for internally identifying our intention to replace it.  

`_ovrtx` was another option I considered but it didn't make sense because even in the ovstage codepath we'd be using ovrtx.   Would you prefer something like `_current` or `_default`?  If you have stronger naming options please shout out

> Re the dispatch approach, instead of still mixing scene data management inside the OVRTX renderer class, I wonder maybe we could create a cleaner design:

That could be stronger design longer term and should be a follow up PR.

The goal of this PR is to get equivalent ovstage functionality to match what we had with ovrtx and I felt the dispatch method was the simplest to follow due to the parallel nature (for both an agent and human).  Form-wise it is not too different from having a OvstageSceneData class, it is just relocating the code.  I am aware of a few changes going into ovrtx over the next week so I also don't want to rock the boat too much. 
 
There are shared members and methods on the OVRTXRenderer class between the ovstage and ovrtx code paths.  For some of them, which class (Renderer vs SceneData) they should live on will require more thought, and design around the OvstageSceneData interface for it to work cleanly.  I think this deserves a proper design session and better understanding of requirements, such as the current access pattern of the Newton Manager

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: What do you think of only having `import ovstage` then using the classes/methods (e.g. ovstage.DLDatatype, ovstage.make_dltensor` as is throughout the code?

source/isaaclab_ov/pyproject.toml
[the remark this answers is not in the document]
rilei-nvidia: Hey i thought more about this.  Given that we have a runtime env var, it is not too much risk to also install ovstage as part of "all".  Should make our benchmarking / CI tests easier without adding additional params - thoughts?

ooctipus (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Yeah, that could be easier to distinguish which methods are from which module, given the similarity of the methods in OVRTX vs. OVStage.

huidongc (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -109,6 +129,68 @@
...
+    Enabled by ``ISAAC_LAB_OVRTX_USE_OVSTAGE=1``. Defaults to ``0`` so existing deployments are
+    unaffected until ovstage is explicitly opted into.
+
+    Raises:
+        ValueError: If the environment variable is set to anything other than ``0`` or ``1``.
+        RuntimeError: If the environment variable is ``1`` but ovstage is not importable. Falling
+            back to the legacy path here would silently ignore an explicit request and make the
+            renderer look like it had honoured it.
+    """
+    value = os.environ.get(_USE_OVSTAGE_ENV, "0").strip()
```
kellyguo11, line 175: I think the environment variable is fine for our CI and testing, but if we eventually want this to be a user-specific setting, we should look at making this a render cfg parameter instead

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1350,3 +1459,777 @@ def _safe_unbind(binding, name: str) -> None:
...
+            minimal_mode=_resolve_rtx_minimal_mode(data_types),
+            camera_rel_path=self._camera_rel_path,
+        )
+        self._render_product_paths.append(render_product_path)
+
+        combined_usd_string = self._exported_usd_string + "\n\n" + render_product_string
+        self._exported_usd_string = None  # Free memory
+
+        # If temp_usd_dir is set, write the combined USD stage to a temporary file.
+        if self.cfg.temp_usd_dir is not None:
```
kellyguo11, line 1596: just curious, is this mostly for the legacy path now?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1350,3 +1459,777 @@ def _safe_unbind(binding, name: str) -> None:
...
+        num_envs = positions.shape[0]
+        converted_wp = wp.empty(num_envs, dtype=wp.quatf, device=self._device)
+        convert_camera_frame_orientation_convention_wp(
+            src=orientations.warp,
+            dst=converted_wp,
+            origin="world",
+            target="opengl",
+            device=self._device,
+        )
+        camera_transforms = wp.zeros(num_envs, dtype=wp.mat44d, device=self._device)
```
kellyguo11, line 2115: would this cause additional overhead if it's called every frame? is it possible to cache a buffer to avoid allocating it on the GPU every time?

source/isaaclab_ov/pyproject.toml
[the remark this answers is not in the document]
kellyguo11: I think all is fine, since ovphysx will also require it. perhaps we should remove ovstage as an optional selective install then? we can just ensure it gets installed whenever ovphysx or ovrtx gets installed.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: That sounds good.  I will ticket this up as a follow up!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: This is currently a debugging feature for both paths to get a copy of the usda stage before we load it in ovrtx / ovstage

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Great spot.  I see this allocation in the legacy path as well.  If the positions is a constant size (always == total number of environments) after initialization then we can certainly preallocate this to reduce dynamic allocs!  I will ticket this up

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Done!

source/isaaclab_ov/pyproject.toml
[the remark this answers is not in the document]
rilei-nvidia: I've removed ovstage as optional selective install, and install whenever ovrtx or ovphysx gets installed

kellyguo11 (approved):

AntoineRichard (approved):

PR #6608, opened 2026-07-19
Add native OpenCV lens-distortion cameras for OVRTX

liorbenhorin:
## Description

Isaac Lab's stock camera cfgs (PinholeCameraCfg, FisheyeCameraCfg) can only express a focal-length + aperture projection. They cannot represent an OpenCV fx/fy/cx/cy + distortion-coefficient intrinsic model — the format real-camera and LeRobot calibrations come in — so a calibrated real camera (non-square pixels, off-center principal point, lens distortion) cannot be reproduced in simulation.

## This PR adds first-class support for OpenCV lens-distortion cameras, rendered natively through the RTX/OVRTX renderer:

- Config: new OpenCvDistortionCfg base with OpenCvPinholeDistortionCfg (k1..k6, p1, p2, s1..s4) and OpenCvFisheyeDistortionCfg (k1..k4) sub-configs, plus a distortion field on PinholeCameraCfg. The model is renderer-agnostic — it lives on the camera cfg, and each renderer decides how to consume it.
- Authoring: spawn_camera authors the model on the camera prim as the omni:lensdistortion:* USD API (schema + model token + fx/fy/cx/cy + imageSize + coefficients). OVRTX honors this natively — no renderer-package change is required.
- Intrinsics readback: Camera._update_intrinsic_matrices now reads the authored fx/fy/cx/cy when a distortion model is present, so camera.data.intrinsic_matrices reflects a real calibration instead of assuming fx == fy and a centered principal point.
- Scope: the RTX/OVRTX path is implemented. The Newton distortion path is intentionally out of scope; it is left as a documented no-op that logs a warning and renders undistorted, and the cfg is designed so Newton can consume the same model later.

This upstreams a capability that has been proven out-of-tree in the SO-101 sim-to-real workshop; once merged, that workshop can delete its private OpenCV-pinhole spawner and use these APIs directly.

## Type of change

## Rendering the grid-textured ground plane - Coefficients exaggerated (--k_scale 15) for visibility.

Screenshot 2026-07-19 135849

## Checklist

- [x] I have read and understood the contr
(https://isaac-sim.github.io/IsaacLab/mainl)
- [x] I have run the pre-commit checks (ht/isaaclab.sh --format
- [x] I have made corresponding changes to
- [x] I have added tests that prove my fixure works
- [x] I have updated the changelog and the extension's config/extension.toml file
- [x] I have added my name to the CONTRIBUists there


liorbenhorin:
greptile issues rexsolved

AntoineRichard:
Nice work — the RTX/OVRTX authoring path here is a reasonable way to get calibrated OpenCV cameras rendering natively, and the renderer-agnostic cfg is the right shape.

One thought for the roadmap: we'd likely want lens distortion supported in **Newton** directly too, not only on the RTX side, so a calibrated camera behaves the same across backends. To see what that looks like, I prototyped the Newton consumption path as a stacked follow-up — Newton's tiled camera traces an explicit per-pixel ray field, so the distortion can be applied at ray-generation time (inverting the OpenCV forward model per pixel) with no Newton-package change:

- Stacked draft PR (Newton path, based on this branch): https://github.com/liorbenhorin/IsaacLab/pull/1
- Upstream issue asking whether the Newton team would take it natively: https://github.com/newton-physics/newton/issues/3578

The Isaac Lab-side implementation works and is GPU-validated today; longer term a native Newton API (per-camera intrinsics + distortion, or a distortion-aware ray util) would be the cleaner home for it.

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -384,6 +384,18 @@ def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
         owns the sentinel-resolution + cfg-normalization step. Newton has no
         USD-side overrides to author beyond this.
         """
+        # NOTE: OpenCV lens distortion (``spawn.distortion``) is not yet applied by the Newton
+        # renderer. The distortion cfg is renderer-agnostic and could be piped through Newton's warp
+        # ray-tracing utilities here in the future; for now the camera renders undistorted. This is
+        # the intended extension point.
+        spawn = getattr(spec.cfg, "spawn", None)
```
StafaH, line 482: Following the other comment and the newton implementation which already exists: https://github.com/newton-physics/newton/blob/main/newton/_src/sensors/warp_raytrace/camera_utils.py#L444

Can we get this implementation into this PR so there is parity across all renderers?

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -633,10 +658,67 @@ def _create_buffers(self):
...
+        """
+        distortion_model = prim.GetAttribute("omni:lensdistortion:model").Get()
+        if not distortion_model:
+            return None
+        prefix = f"omni:lensdistortion:{distortion_model}"
+        # a prim may carry the model token without fx/fy/cx/cy
+        intrinsics = tuple(prim.GetAttribute(f"{prefix}:{name}").Get() for name in ("fx", "fy", "cx", "cy"))
+        if None in intrinsics:
+            # a model token without intrinsics falls back to the focal-length/aperture projection
+            if not self._warned_distortion_missing_intrinsics:
```
StafaH: Why do we need to supress repeated warning? Are there truly going to be that many?

Even if there are, let's not create new variables like this, it could easily be inlined:
```

require_warning=False

do some loop:
   return warning=True

if require_warning:
   warn
```

But even that might not be necessary if there are not that many warnings, so either or

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -633,10 +658,67 @@ def _create_buffers(self):
...
+                )
+                self._warned_distortion_missing_intrinsics = True
+            return None
+        # the intrinsics are reported in the calibrated pixel space; warn once per authored size if the
+        # render resolution differs, since the matrix will not match the rendered pixels.
+        image_size = prim.GetAttribute(f"{prefix}:imageSize").Get()
+        if image_size is not None:
+            authored_size = (int(image_size[0]), int(image_size[1]))
+            if authored_size != (width, height) and authored_size not in self._warned_distortion_image_sizes:
+                logger.warning(
```
StafaH, line 699: Similar comment about the warn guard

StafaH (commented):
LGTM, let's implement the newton path as well and get parity?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
liorbenhorin: Will get done on a follow up MR

source/isaaclab/isaaclab/sensors/camera/camera.py
liorbenhorin: For now removed the suppression

StafaH (approved):

PR #6613, opened 2026-07-20
Fix root fixing on stages without omni.physx

ooctipus:
# Fix root fixing on stages without omni.physx

## What

- Adds `create_world_fixed_joint(articulation_prim, stage)` to
  `source/isaaclab/isaaclab/sim/schemas/schemas.py`: a pure-USD helper that authors the
  world-attached `UsdPhysics.FixedJoint` used to fix an articulation root link. It mirrors
  the behavior of `omni.physx.scripts.utils.createJoint(joint_type="Fixed", from_prim=None,
  to_prim=...)`: unique joint naming, walking up to the first writable ancestor for
  instanceable/prototype/instance-proxy roots, anchoring `localPos0`/`localRot0` at the root
  link's world pose, and setting break force/torque to the same effectively-unbreakable
  `MAX_FLOAT` value.
- `modify_articulation_root_properties` (the `fix_root_link` spawn path) now calls this
  helper instead of importing `omni.physx.scripts.utils`.
- `PhysicsManager.fix_articulation_root` delegates to the same helper instead of carrying
  its own inline joint-authoring copy, so there is a single authoring path for every
  backend.
- Adds a kitless regression test,
  `source/isaaclab/test/sim/test_schemas_kitless_fixed_joint.py`, which runs on an
  in-memory USD stage without launching Isaac Sim / Kit.
- Adds a changelog fragment: `source/isaaclab/changelog.d/schemas-kitless-fixed-joint.rst`.

## Why

Spawning a fixed-base articulation (`fix_root_link=True`) previously imported
`omni.physx.scripts.utils` inside `modify_articulation_root_properties`. On stages where
`omni.physx` is not available (kitless / usd-core-only setups, e.g. Newton backends), that
import raises `ModuleNotFoundError`, breaking the spawn path. Authoring the joint directly
with plain USD removes the `omni.physx` dependency while preserving the authored result.
Sharing the helper with `PhysicsManager.fix_articulation_root` also removes a duplicated
copy of the joint-authoring logic.

This is a standalone correctness fix plus a regression test; no performance claims.

## Evidence (existing upstream workflow)

Measured with the repository's own benchmark, identical command on the PR branch and its base (`upstream/develop` @ 5c42d3904):

```
./isaaclab.sh -p scripts/benchmarks/runtime.py --task Isaac-Velocity-Flat-AnymalD --num_envs 1024 --num_frames 300 --seed 7
```
Platform: RTX 5090, Newton backend, CUDA graphs on.

| arm | wall (300 iters × 1024 envs) | ms/iter |
|---|---|---|
| develop base | 3.811 s | 12.70 |
| this PR | n/a — CPU-side USD authoring fix, no runtime path touched | — |

Runtime evidence is not applicable; the change fixes `fix_root_link` on kit-less stages. Correctness evidence: the new regression test (`test_schemas_kitless_fixed_joint.py`, 2 passed) exercises the plain-USD path end-to-end.

### DIFF
--- a/source/isaaclab/changelog.d/schemas-kitless-fixed-joint.rst
+++ b/source/isaaclab/changelog.d/schemas-kitless-fixed-joint.rst
@@ -1,0 +1,8 @@
+Changed
+^^^^^^^
+
+* Changed :meth:`~isaaclab.sim.schemas.modify_articulation_root_properties` to author
+  the fixed-root-link world joint directly with USD on every backend instead of calling
+  ``omni.physx.scripts.utils.createJoint``. This removes the ``omni.physx`` dependency
+  from the spawn path, which previously raised ``ModuleNotFoundError`` when spawning
+  fixed-base articulations on kitless backends (e.g. Newton).
--- a/source/isaaclab/isaaclab/physics/physics_manager.py
+++ b/source/isaaclab/isaaclab/physics/physics_manager.py
@@ -119,13 +119,14 @@ class PhysicsManager(ABC):
         Raises:
             NotImplementedError: If a new joint is needed and the root is not a rigid body.
         """
-        # Keep this import local to avoid the SimulationContext -> PhysicsManager ->
-        # sim.utils.queries -> SimulationContext import cycle.
-        # Keep pxr local as well: this module is imported while environment configs load (via the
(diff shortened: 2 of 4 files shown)


ooctipus:
Closing — this is already covered by #6536.

PR #6616, opened 2026-07-20
Run all Newton actuators through one flat-layout adapter path

ooctipus:
# Description

Newton actuation previously ran as two parallel code paths selected by
`SimulationCfg.use_newton_actuators`: the default path computed Lab actuator models
eagerly on the host once per articulation per step, and the opt-in fast path ran Newton
actuators inside the physics step but assumed every environment repeats env-0's DOF
rectangle, so it broke on any scene whose articulations do not tile uniformly.

This PR makes the fast path the only path and removes the layout assumption:

- **Flat DOF layout.** The adapter is built from the model's own `joint_world` /
  `joint_qd_start` tables instead of an `env_stride × num_envs` rectangle. Per-articulation
  state is assembled by pure gathers through each articulation's flat DOF index map
  (including the view layout's global `offset` — previously only articulations placed
  first in their world resolved correctly), so homogeneous and heterogeneous scenes
  construct identically.
- **One path.** `SimulationCfg.use_newton_actuators` is deleted. Newton articulations
  always bind their explicit actuators (ideal PD, DC motor, delayed PD, remotized PD,
  actuator nets) to Newton actuators that run inside the physics step — inside the CUDA
  graph when every actuator is graph-safe, stepped eagerly per iteration otherwise.
  `_is_all_graphable` is derived from the adapter alone, and the manager always owns the
  decimation loop; the activation flag state is gone.
- **Deletions.** The legacy Lab-actuator tail in the Newton articulation (including its
  `_apply_actuator_model` override), the flag plumbing in `schemas_actuators`, and the
  PhysX-vs-Newton A/B flag test are removed. `BaseArticulation._apply_actuator_model` is a
  concrete no-op hook: backends that compute actuator models on the host (PhysX) override
  it; backends whose actuators run inside the physics step (Newton) do not.

## Dependency

This branch is **rebased over #6614** and must merge after it. The dependency is
functional, not just textual: without #6614's invalidate-then-capture policy, removing the
flag makes every explicit-actuator scene take the old "skip initial capture, recapture on
`set_decimation`" arm — and plain `SimulationContext` loops (demos, standalone scripts,
benchmarks) never call `set_decimation`, so they run eager forever (measured 7.0 ms/step
vs 2.9 ms baseline on the benchmark below). On #6614's manager, decision sites only
invalidate and the first `step()` captures, on every path.

## Performance

Actuator dispatch cost scales with the number of articulation groups, not environments —
single-robot benchmarks cannot show it. This PR adds
`scripts/benchmarks/benchmark_actuator_dispatch.py`: N ANYmal-D per environment cycling DC
motor, ideal PD, and solver-implicit PD actuator configs, stepped steady-state on Newton
MJWarp. Every environment is identical, so the same command runs on any branch without
multi-world grouping.

```bash
./isaaclab.sh -p scripts/benchmarks/benchmark_actuator_dispatch.py --headless \
    --num_envs 1024 --num_articulations 8
```

Baseline = `develop` + the benchmark commit cherry-picked; identical commands, single
RTX 5090, per-step time over 500 steps after 100 warmup steps, median of 3 interleaved
runs per configuration:

| scene | this PR (on #6614) | #6614 only | develop | vs develop |
|---|---|---|---|---|
| 1024 envs × 2 robots (DC + ideal PD) | **0.622 ms** | — | 0.816 ms | −24% |
| 64 envs × 8 robots | **1.556 ms** | 2.421 ms | 2.408 ms | −35% |
| 256 envs × 8 robots | **1.822 ms** | — | 2.408 ms | −24% |
| 1024 envs × 8 robots | **2.780 ms** | 3.051 ms | 2.980 ms | −7% |
| 4096 envs × 8 robots | **7.544 ms** | — | 7.936 ms | −5% |
| 1024 envs × 16 robots | **11.485 ms** | — | 12.443 ms | −8% |

The #6614-only column sits at develop parity, so the gain is this PR's in-graph actuators
rather than the capture-policy change. The win tracks the ratio of per-articulation host
dispatch to GPU solver time: largest at low and mid environment counts, smaller once
per-world solver cost dominates (in this benchmark, adding robots per environment also
grows per-world solver work, which is why 16 robots/world gains less than 8).

## Tests

`test_newton_actuators_newton.py` is rewritten as behavior checks: 73 passed, 3 xfailed.

- Each actuator type is asserted against its analytic law — PD torque, max-effort clamp,
  DC-motor speed-torque clamp, delay semantics, implicit drive telemetry — plus
  joint-ordering permutation cases and a heterogeneous ANYmal + cartpole scene whose
  per-robot laws catch cross-articulation DOF-map clobbering (this suite caught the view
  `offset` bug fixed in this PR).
- Golden captures (`test_actuator_adapter_golden.py`): single-tick bitwise captures from
  explicitly written states; machine-generated on first run, not committed.
- Three implicit-drive motion tests are `expectedFailure`, documenting an upstream MJWarp
  defect this PR does not change: implicit joint drives are baked into the MuJoCo actuator
  set at model finalize from USD-authored gains, so assets authored with zero drive gains
  (e.g. ANYmal) get no MuJoCo actuator, and gains written at initialization land in
  `model.joint_target_ke` where the solver never reads them back. Reproduced identically
  on `develop`.

## Breaking change & migration

`SimulationCfg.use_newton_actuators` is removed. Newton scenes always use the Newton
actuator path; delete the field from configs. PhysX behavior is unchanged.


ooctipus:
Rework in progress before review: the legacy Lab actuator compute path on the Newton backend will be removed in this PR (the adapter becomes the only actuation path on Newton; PhysX keeps its existing default path unchanged), and the test suite will be rewritten to assert current behavior directly (tracking, clamping, decimation, reset isolation, golden captures) instead of A/B-comparing against the removed path. Keeping as draft until then.

ooctipus:
Rebased onto #6614 — the dependency is functional: without its invalidate-then-capture policy, removing the actuator flag would leave plain `SimulationContext` loops eager forever (7.0 ms vs 2.9 ms per step on the new benchmark). Merge order: #6614 first; this branch then shows as its 7 own commits.

PR #6622, opened 2026-07-20
Draft: Per tile render history reset

rilei-nvidia:
# Description

```bash
./isaaclab.sh -p -m pytest -s -rA -k "newton" source/isaaclab_tasks/test/core/test_rendering_history_reset_cartpole_kitless.py
```

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/renderers/base_renderer.py
+++ b/source/isaaclab/isaaclab/renderers/base_renderer.py
@@ -8,6 +8,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+from collections.abc import Sequence
 from typing import TYPE_CHECKING, Any
 
 from .camera_render_spec import CameraRenderSpec
@@ -166,6 +167,20 @@ class BaseRenderer(ABC):
                 instance to populate.
         """
         pass
+
+    def reset_tile_history(self, render_data: Any, env_ids: Sequence[int] | None) -> None:
+        """Signal that the specified environments have been reset and temporal render history should be cleared.
+
+        Renderers that accumulate history across frames (e.g. DLSS) should override this to flush
+        per-tile state for the affected environments, preventing ghosting after sudden camera jumps.
+
+        The base implementation is a no-op; only renderers with temporal state need to override it.
+
+        Args:
+            render_data: The render data object from :meth:`create_render_data`.
+            env_ids: Indices of environments that have been reset. ``None`` means all environments.
+        """
+        return
 
     @abstractmethod
     def cleanup(self, render_data: Any) -> None:
--- a/source/isaaclab/isaaclab/sensors/camera/camera.py
+++ b/source/isaaclab/isaaclab/sensors/camera/camera.py
@@ -459,6 +459,8 @@ class Camera(SensorBase):
         else:
             env_ids_wp = self._resolve_env_ids_wp(env_ids)
             self._update_poses(env_ids_wp, frame_op=2)
+        # clear temporal denoiser state (e.g. DLSS) for environments that teleported
+        self._renderer.reset_tile_history(self._render_data, env_ids)
 
     """
     Implementation.
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -23,6 +23,7 @@ import logging
 import math
 import os
 import re
+from collections.abc import Sequence
 from itertools import compress
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NoReturn, cast
@@ -312,6 +313,7 @@ class OVRTXRenderer(BaseRenderer):
         self._camera_rel_path: str | None = None
         self._output_id_color_buffers: dict[str, wp.array] = {}
         self._clone_plan: ClonePlan | None = None
+        self._render_history_reset_request_id: int = 0
 
         logger.info("Creating OVRTX renderer...")
         OVRTX_CONFIG = RendererConfig(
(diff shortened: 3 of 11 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -688,6 +690,46 @@ def create_render_data(self, spec: CameraRenderSpec) -> OVRTXRenderData:
...
+        self._renderer.write_array_attribute(
+            prim_paths=[rp],
+            attribute_name="omni:rtx:viewTile:renderHistoryReset:tileIndices",
+            tensors=[tile_indices],
+        )
+        self._renderer.write_attribute(
+            prim_paths=[rp],
+            attribute_name="omni:rtx:viewTile:renderHistoryReset:requestId",
+            tensor=np.array([self._render_history_reset_request_id], dtype=np.int32),
+        )
```
huidongc, line 796: Wondering if it makes sensor to create atrribute bindings? Env reset is not a rare op during training.

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -167,6 +168,20 @@ def read_output(self, render_data: Any, camera_data: CameraData) -> None:
         """
         pass
 
+    def reset_tile_history(self, render_data: Any, env_ids: Sequence[int] | None) -> None:
```
pbarejko, line 171: I would just call it `reset`. `reset_tile_history` seems very specific. If it's called more generic as `reset` it could be reused for other reasons.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -688,6 +690,46 @@ def create_render_data(self, spec: CameraRenderSpec) -> OVRTXRenderData:
...
+        ``omni:rtx:viewTile:renderHistoryReset:requestId`` on the render product so that
+        OVRTX discards accumulated denoiser state for environments that just reset.
+
+        See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.reset_tile_history`.
+        """
+        if not self._render_product_paths or self._renderer is None:
+            return
+
+        if env_ids is None:
+            tile_indices = np.arange(render_data.num_envs, dtype=np.float32)
```
pbarejko, line 771: nit: 

```
`/rtx/viewTile/renderHistoryReset/tileIndices`

* `float[]`
* `-1.0` resets all valid tiles
* `-2.0 `is no-op/default padding
* `0..N-1` reset selected flat tile indices
```

then this becomes:

```python
if env_ids is None:
      tile_indices = np.array([-1.0], dtype=np.float32)
```

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -688,6 +690,46 @@ def create_render_data(self, spec: CameraRenderSpec) -> OVRTXRenderData:
...
+        else:
+            # np.asarray requires a CPU array; move off device if needed.
+            # write_array_attribute only accepts CPU/numpy data — the GPU path
+            # requires a pre-allocated binding, which doesn't fit variable-length
+            # reset indices.  The copy is negligible: this runs once per reset,
+            # not per frame, on an array of at most num_envs floats.
+            ids = env_ids.cpu() if isinstance(env_ids, torch.Tensor) else env_ids
+            tile_indices = np.asarray(ids, dtype=np.float32)
```
pbarejko, line 779: This could be fragile because you made an assumption:

 ```
  rel camera = [
      </World/envs/env_0/Camera>,  # tile 0
      </World/envs/env_1/Camera>,  # tile 1
      </World/envs/env_2/Camera>,  # tile 2
  ]
```

could the mapping of the indices be different? i.e.

 ```
  rel camera = [
      </World/envs/env_0/Camera>,  # tile 0
      </World/envs/env_2/Camera>,  # tile 2
      </World/envs/env_1/Camera>,  # tile 1
  ]

File: source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py [hunk 5]
```diff
@@ -688,6 +690,46 @@ def create_render_data(self, spec: CameraRenderSpec) -> OVRTXRenderData:
...
+
+        if len(tile_indices) == 0:
+            return
+
+        self._render_history_reset_request_id += 1
+        rp = self._render_product_paths[0]
+
+        self._renderer.write_array_attribute(
```
pbarejko, line 787: Since you experience problem with the feature, do we know if this write op succeeded? Can you read it back for debugging?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Yeah good call out, since we author the camera rel in https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_usd.py#L125 in env ascending order I will add a comment here about this assumption.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Yup, have validated!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Will do this

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Ah nice!

source/isaaclab/isaaclab/renderers/base_renderer.py
rilei-nvidia: Since the call chain will be `SimulationEnv.reset -> Camera.reset -> Renderer.[our new method]` it does feel like we want to name it specific for resetting the tiled camera state.   A generic `Renderer.reset` would suggest the method could be resetting other internal renderer state which seems undesirable given the context and call site.

huidongc: But this is not a helper and internal util, it is a method inherited from base renderer class. We can use it to do specific task in the subclass, but as an interface we should keep it generic if possible, maybe the next day we will need to use it for some other resetting work.

rilei-nvidia: Say we add some new behavior to reset other renderer internal state, unrelated to tiled camera.  

Are we sure we want Camera.reset to trigger resetting of these un-related state?

pbarejko: I am going to leave it up to you guys to decide, but to me you are resetting some internal state of a renderer.  I would push the details to the implementation and generalize it as reset on interface. The reset tile is very specific to one single implementation and it's not going to be used in any other implementation. Otherwise you will keep bloating base interface with implementation details.

```
SimulationEnv.reset -> Camera.reset -> Renderer.reset
```

```
┌────────────────────┐    ┌──────────────┐    ┌────────────────┐           
│SimulationEnv.reset ┼────► Camera.reset ┼────► Renderer.reset │           
└────────────────────┘    └──────────────┘    └▲────────────▲──┘           
                                               │            │              
                                               │            │              
                                               │            │              
                                               │            │              
                                               │            │              
                                     ┌─────────┴─────┐    ┌─┴─────────────┐
                                     │ImplementationA│    │ImplementationB│
                                     │  (whatever)   │    │ (tile_reset)  │
                                     └───────────────┘    └───────────────┘                                                                                               
```

source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -459,6 +459,8 @@ def reset(self, env_ids: Sequence[int] | None = None, env_mask: wp.array | None
         else:
             env_ids_wp = self._resolve_env_ids_wp(env_ids)
             self._update_poses(env_ids_wp, frame_op=2)
+        # clear temporal denoiser state (e.g. DLSS) for environments that teleported
```
pbarejko, line 462: This comment is incorrect for Newton Warp.

source/isaaclab/isaaclab/renderers/base_renderer.py
rilei-nvidia: @pbarejko I get your point of wanting to make this more generic.  I'm not tied to reset_tile_history that is only the current placeholder.  But a Renderer may have a lot of different kinds of internal state that a caller may want to reset to achieve some kind of purpose.  

Say the newton warp renderer has caches that accelerate segmentation annotator.  That can be considered internal state and be implemented under .reset().  Now Camera.reset() will trigger the clearing of that cache which is not desirable.  What camera.reset ultimately wants to do with the call is to combat ghosting - I feel the base class method needs to be named and documented with that purpose in mind or it is both confusing to call sites and subclass implementations exactly what needs to be "reset"

PR #6628, opened 2026-07-20
Fix ovphysx reset-velocity cache staleness and slow ASCII warmup export

ClemensSchwarke:
# Description

Fixes two issues I encountered in the velocity env cleanup.

1. The velocity cache was stale after reset, which caused huge velocities to enter the MDP terms. ArticulationData._reset_velocity() invalidated the world-frame velocity buffers but not the body-frame ones.
2. OvPhysxManager exported the warmup stage as ASCII USD (.usda). Serializing large robot/terrain geometry to text is extremely slow (It took 38 min to load the rough Digit environment).

## Type of change


source/isaaclab_ovphysx/test/assets/test_articulation.py
AntoineRichard, line 1: If we do this tests here we should probably add it to the other backends.

AntoineRichard:
Working on a PR that will fix it on all backend.

PR #6631, opened 2026-07-20
Restore RL entrypoint backend state

AntoineRichard:
# Description

The reusable reinforcement learning entrypoints introduced in #6553 restore `sys.argv` after in-process calls, but backend-specific global settings can still leak into the caller.

This PR scopes those mutations to each invocation:

- RSL-RL training parses arguments before temporarily enabling its Torch backend flags and restores all four flags on success or failure.
- SKRL training restores `skrl.config.jax.backend`, including deleting it when the attribute did not exist before the call.
- SKRL playback scopes the same state inside `main()`, covering both unified dispatch and direct imported calls.
- Adds failure-path regressions for rejected RSL-RL requests and exceptional SKRL train/play calls.

## Relationship

- Focused follow-up to #6553.
- Independent of the benchmark PR stack.

## Type of change

- Bug fix

## Validation

- `./isaaclab.sh -p -m pytest source/isaaclab_rl/test/test_entrypoints.py -q` (11 passed)
- `./isaaclab.sh -f`

The full `source/isaaclab_rl/test` directory was also attempted locally but did not complete or emit results within the five-minute command timeout.

## Checklist


source/isaaclab_rl/isaaclab_rl/entrypoints/_state.py
```diff
@@ -0,0 +1,46 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: This type of file is not good practice. Let's just use common.py

source/isaaclab_rl/isaaclab_rl/entrypoints/_state.py
```diff
@@ -0,0 +1,46 @@
...
+            setattr(target, name, previous)
+
+
+@contextmanager
+def scoped_torch_backend_flags(torch_module: ModuleType) -> Iterator[None]:
+    """Temporarily configure the Torch backend flags used by RSL-RL training."""
+    cuda_matmul = torch_module.backends.cuda.matmul
+    cudnn = torch_module.backends.cudnn
+    settings = (
+        (cuda_matmul, "allow_tf32", True),
```
StafaH: Why not just allow all these to be args, then this function can set them, and the function can be used by rsl_rl training.

That way no hardcoded values?

StafaH (approved):

ooctipus (approved):

PR #6634, opened 2026-07-20
Fix AppLauncher exception exit status

### LINKED ISSUE #6573: [Bug Report] AppLauncher atexit handler masks uncaught exceptions as exit code 0
### Description

Commit [`0090622`](https://github.com/isaac-sim/IsaacLab/commit/0090622b0ddba0bf11da289cc784afa5dcd7e009) added an `atexit` handler that unconditionally calls `SimulationApp.close()`:

```python
def _atexit_close(app=self._app):
    with contextlib.suppress(Exception):
        app.close()
```

When an uncaught exception occurs after `AppLauncher` initialization, this handler runs during interpreter shutdown. Kit fast shutdown exits with the default `SimulationApp.close(exit_code=0)`, replacing the pending nonzero Python exit status.

As a result, launchers such as `torchrun`, CI systems, and workflow schedulers report failed training processes as successful. We observed an unhandled exception and traceback from every `torchrun` worker, followed by `torchrun` and the workflow task exiting with code 0.

### Expected behavior

Uncaught exceptions preserve a nonzero process exit code.

### Actual behavior

The traceback is printed, but the process exits with code 0.

### Regression

The behavior was introduced by commit `0090622` and is present in `v3.0.0-beta2`. `v2.3.2` and `v3.0.0-beta` predate the change.

The current `SimulationApp.close(exit_code=...)` documentation explicitly notes that a nonzero value is required to prevent fast shutdown from replacing the status with 0.

nblauch:
# Description

`AppLauncher` registers an `atexit` callback that closes `SimulationApp`. With
Kit fast shutdown enabled, calling `close()` with its default zero exit code can
replace Python's pending nonzero status after an unhandled exception.

This change detects the unhandled exception state exposed by the interpreter
and passes exit code 1 to `SimulationApp.close()`. Normal interpreter shutdown
continues to pass exit code 0.

The new integration test launches a real headless CPU `AppLauncher` in a child
process, raises an unhandled `RuntimeError`, and verifies both the traceback and
exit status 1.

Fixes #6573

## Type of change

## Testing

- Before the fix, the regression test failed because the child returned 0.
- After the fix, running the trigger script directly reports exit status 1.
- `python -m pytest source/isaaclab/test/app/test_app_launcher_exit_code.py source/isaaclab/test/app/test_kwarg_launch.py::test_launch_simulation_preserves_failure_exit_code -q` (2 passed)
- `./isaaclab.sh -f` (all hooks passed)

## Screenshots

Not applicable.

## Checklist

- [x] I have read and understood the contribution guidelines.
- [x] I have run the pre-commit checks with `./isaaclab.sh --format`.
- [x] Documentation is not required for this internal shutdown correction.
- [x] I have added tests that prove my fix is effective.
- [x] I have added a changelog fragment under `source/isaaclab/changelog.d/`.
- [x] My name is already present in `CONTRIBUTORS.md`.

### DIFF
--- a/source/isaaclab/test/app/test_app_launcher_exit_code.py
+++ b/source/isaaclab/test/app/test_app_launcher_exit_code.py
@@ -1,0 +1,38 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Verify that AppLauncher preserves the exit status of unhandled exceptions."""
+
+import subprocess
+import sys
+
+import pytest
+
+_TRIGGER_UNHANDLED_EXCEPTION = "--trigger-unhandled-exception"
+
+
+def _raise_after_app_launcher_initialization() -> None:
+    from isaaclab.app import AppLauncher
+
+    AppLauncher(headless=True, device="cpu")
+    raise RuntimeError("intentional AppLauncher failure")
+
+
+@pytest.mark.integration
+def test_unhandled_exception_exits_with_failure():
+    """Verify that AppLauncher shutdown does not replace an exception's exit status with zero."""
+    result = subprocess.run(
+        [sys.executable, __file__, _TRIGGER_UNHANDLED_EXCEPTION],
+        capture_output=True,
+        text=True,
+        timeout=120,
+    )
+
+    assert result.returncode == 1, result.stdout + result.stderr
+    assert "RuntimeError: intentional AppLauncher failure" in result.stderr
+
+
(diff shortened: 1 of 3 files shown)


source/isaaclab/test/app/test_app_launcher_exit_code.py
AntoineRichard, line 1: Do we need this test?

AntoineRichard (commented):
Should we enable this for debug only?

source/isaaclab/test/app/test_app_launcher_exit_code.py
nblauch: thanks for review @AntoineRichard 

re: needing for debug only - why? shouldn't users be correctly informed when their code fails by the exit status of the script? 

re: the test, i defer judgment on this to code owners

kellyguo11:
superseded by https://github.com/isaac-sim/IsaacLab/pull/6636

PR #6635, opened 2026-07-20
Fix fragment writers to reach nested rigid-body and mass prims

vidurv-nvidia:
# Description

Follow-up reconciling the fragment writers with #6377, which merged the same day as #6501.

#6377 established that nested rigid-body hierarchies are legitimate asset topology: the URDF importer in Isaac Sim 6.0 and later authors child links under their parent link prims. It fixed the legacy `modify_rigid_body_properties` / `modify_mass_properties` path via `apply_nested(stop_on_success=False)` so every link is reached. The fragment target resolver introduced in #6501 still stops descending at the first schema-bearing prim on each branch, so on such assets the fragment path modifies only the outermost link — silently reproducing the exact 1-of-N-links bug #6377 fixed on the legacy path.

This PR restores legacy parity:

- `_resolve_fragment_targets` gains a `stop_at_carrier` flag mirroring `apply_nested`'s `stop_on_success`. The rigid-body and mass fragment writers now collect every carrier in the subtree; the collision writer keeps stop-at-carrier traversal, matching the legacy collision writer, which #6377 left on the default.
- Docstrings no longer claim nested applications of the same physics schema are not allowed.
- The nested-target regression tests author a nested child link in the test asset, so the spawn-path tests, the fragment-vs-legacy parity test, and a new direct-writer test all cover the nested topology. Without the fix, four of them fail (nested link untouched by the fragment path while legacy reaches it); with it, all pass, and the legacy schema tests still pass (41/41).

Fixes the fragment-path counterpart of #6377.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/vidurv-descend-nested-fragments.rst
+++ b/source/isaaclab/changelog.d/vidurv-descend-nested-fragments.rst
@@ -1,0 +1,9 @@
+Fixed
+^^^^^
+
+* Fixed :func:`~isaaclab.sim.schemas.apply_rigid_body_properties` and
+  :func:`~isaaclab.sim.schemas.apply_mass_properties` stopping at the outermost schema-bearing
+  prim on each branch. On assets with nested rigid-body hierarchies (child links authored under
+  their parent link prims, as produced by the URDF importer in Isaac Sim 6.0 and later), only
+  the outermost link received the fragment properties. The writers now modify every carrier in
+  the subtree, matching the legacy writers' full-subtree traversal.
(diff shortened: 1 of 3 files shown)


vidurv-nvidia:
Superseded by #6640: the fragment writers are moving from subtree traversal to explicit prim-path-expression targeting, which removes the traversal policy this PR patched. The nested-hierarchy regression tests and the fix itself are carried inside #6640 (this branch is part of its history), so nested URDF-importer assets are covered there. Closing to keep a single review surface — the fragment path is still experimental, so there is no released behavior needing an interim fix.

Issue #6530, opened 2026-07-15
[Bug Report] AppLauncher SIGTERM handler does not terminate Python workers


RX-02333:
### Describe the bug

`AppLauncher` currently registers `_abort_signal_handle_callback` as the
SIGTERM handler. The callback calls `SimulationApp.close()` and then returns.

Installing a Python signal handler replaces the operating system's default
SIGTERM termination behavior. Returning from the callback therefore allows
Python execution to continue instead of terminating the process.

During distributed training shutdown, this can leave a worker alive after the
`torchrun` parent and TCPStore have exited. The surviving worker's
ProcessGroupNCCL heartbeat thread continues accessing the closed TCPStore and
repeatedly prints:

```text
[rank1] TCPStore.cpp:106 sendBytes failed ... Broken pipe
[rank1] ProcessGroupNCCL.cpp:1802 Failed to check the "should dump" flag
on TCPStore ... Broken pipe
```

### Steps to reproduce

From the Isaac Lab repository root, start a two-rank RSL-RL training job:

```bash
CUDA_VISIBLE_DEVICES=1,2 ./isaaclab.sh -p \
  -m torch.distributed.run \
  --standalone \
  --nnodes=1 \
  --nproc_per_node=2 \
  scripts/reinforcement_learning/rsl_rl/train.py \
  --task Isaac-Cartpole \
  --num_envs 2 \
  --max_iterations 5000 \
  --distributed \
  --viz none
```

After both ranks enter training, send SIGTERM to the
`torch.distributed.run` parent process:

```bash
kill -TERM <torchrun-pid>
```

Expected behavior:

- Both workers terminate.
- Normal Python cleanup paths run.
- No training processes remain.
- TCPStore `Broken pipe` messages are not repeatedly emitted.

Actual behavior:

- The SIGTERM callback returns after closing `SimulationApp`.
- A worker may continue executing.
- Its ProcessGroupNCCL heartbeat thread continues accessing the closed
  TCPStore.
- The terminal repeatedly prints `Broken pipe` errors.

### System Info

- Commit: `6f991d4becf764b151e0a1c775561ddd9c406f72`
- Isaac Sim Version: `6.0.1.0`
- OS: Ubuntu 24.04.4 LTS
- GPU: NVIDIA RTX PRO 6000 Blackwell Workstation Edition
- CUDA: 12.8
- GPU Driver: 610.43.02

### Additional context

[...]

PR #6636, opened 2026-07-20
Make AppLauncher teardown truthful and re-entrancy-safe; document mGPU NCCL workaround

hujc7:
# Summary

- Consolidates `AppLauncher` process-lifecycle handling into one nested `_SimulationAppLifecycle` class: startup announcements (CI marker, Kit version diagnostics) plus the entire exit-path policy, with the policy table as the class docstring.
- Fixes every exit path that misreported failure as success or destroyed its own diagnostics (full failure-case table below). Absorbs the atexit exit-code fix from https://github.com/isaac-sim/IsaacLab/pull/6634.
- Documents the NCCL cuMem workaround for multi-GPU RTX training on NUMA-spanning GPU allocations (`NCCL_CUMEM_HOST_ENABLE=0` first, `NCCL_CUMEM_ENABLE=0` as fallback).
- The upstream `SimulationApp` exit-status fix (public reference: https://github.com/isaac-sim/IsaacSim/pull/717) has been **merged** (`isaacsim.simulation_app` >= 2.18.5): the signal handler now simply passes `close(exit_code=128 + signum)` — full teardown + truthful status on fixed builds, truthful status on older builds. The one remaining `WORKAROUND(isaac-sim)` is the SIGINT re-registration (upstream handler still exits 0 before user code unwinds); it fails loudly if drift disables it.
- Fixes https://github.com/isaac-sim/IsaacLab/issues/6573 (absorbed atexit exit-code fix, originally https://github.com/isaac-sim/IsaacLab/pull/6634 by @nblauch).
- Fixes https://github.com/isaac-sim/IsaacLab/issues/6530: the SIGTERM handler no longer returns to the interrupted execution path — the worker exits through `close(exit_code=128 + signum)` (full Kit teardown on `isaacsim.simulation_app` >= 2.18.5) or dies by the re-raised signal, so distributed workers terminate with a truthful status instead of surviving and spamming TCPStore `Broken pipe` errors.

# Failure cases and how this PR addresses them

Root mechanism: Kit fast shutdown terminated the process with exit code 0 from inside `SimulationApp.close()`, so any death funneled through an unqualified `close()` was reported as success; additionally, the abort-signal handler was unguarded against re-entrancy.

| How the process ends | Before this PR | After this PR |
|---|---|---|
| Unhandled Python exception | atexit close overwrote the pending failure with **exit 0** (CI false-green) | exits 1 (`sys.last_exc` detected; absorbed from #6634; `SystemExit` documented as not yet covered) |
| Single SIGTERM (torchrun teardown, SLURM preemption, `kill`) | graceful close → **exit 0**; launcher marks the killed rank SUCCEEDED; surviving ranks hang until the NCCL watchdog | `close(exit_code=128 + signum)`: full teardown + truthful status on `isaacsim.simulation_app` >= 2.18.5; truthful status on older builds; dies by the signal if `close()` returns |
| Second signal while a close is running (repeated SIGTERM; fault inside the replicator stop/wait) | handler re-entered `close()` → **infinite recursion** → SIGKILL-only shutdown, spurious SIGSEGV, logs flooded (~975 recursion frames/job observed on OSMO pods) | guard: re-entrant signal falls back to `SIG_DFL` |
| Signal racing the normal atexit close | nested full second teardown of a half-closed app | atexit arms the same guard |
| `kill -ABRT` | graceful close → **exit 0** | same truthful-close path as SIGTERM |
| Real SIGSEGV, main thread | Python handler can never run → process **spins forever** at 100% CPU, crash reporter clobbered (no minidump) | SIGSEGV no longer intercepted → default action, minidumps restored |
| Real SIGSEGV, worker thread | handler ran on the main thread → **exit 0** for a crashed process | same: default action, truthful signal death |
| Ctrl-C | SimulationApp's handler exits **0** before user `finally`/`KeyboardInterrupt` code runs | Python default handler restored: `KeyboardInterrupt` unwinds user code, nonzero exit |

Not addressed here (tracked elsewhere): `sys.exit(N)`/`SystemExit` still exits 0 (gap inside the #6634 mechanism, documented at the detection site).

# Implementation notes

1. All exit-path logic lives in `AppLauncher._SimulationAppLifecycle`; the class docstring is the policy table, and each decision carries its rationale in place.
2. The signal handler passes the killed-by-signal status through `close(exit_code=128 + signum)`. With the merged upstream fix (`isaacsim.simulation_app` >= 2.18.5) the app performs its full teardown and exits with that status; on older builds the status is preserved without the teardown; if `close()` returns (fast shutdown disabled), the handler re-raises with the default action. A `TypeError` fallback warns loudly if a future `SimulationApp` drops the parameter.
3. Docs: distributed camera training fails deterministically when the allocated GPUs span NUMA nodes and passes on a single-switch set; disabling NCCL cuMem host allocations rescues the failing shape in paired same-node experiments. Added to the multi-GPU NCCL troubleshooting section.

**Testing.** Six kitless unit tests (`test_simulation_app_lifecycle.py`: killed-by-signal status, re-entrancy both directions, exit-code selection, drift fallbacks on both paths) plus two real-Kit integration tests (`test_app_launcher_exit_status.py`: SIGTERM → truthful termination status with no handler recursion; unhandled exception → exits 1). The integration tests pass against both the pre-fix and the fixed (>= 2.18.5) `SimulationApp` builds; on develop's original behavior the SIGTERM test observes exit code 0 (the bug).

## Type of change

- Documentation update

## Checklist


source/isaaclab/test/app/test_app_launcher_exit_status.py
```diff
@@ -0,0 +1,101 @@
...
+    proc = subprocess.Popen(
+        [sys.executable, __file__, _TRIGGER_WAIT_FOR_SIGTERM],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    try:
+        # wait for the app to finish starting up
+        deadline = time.time() + 300
+        for line in proc.stdout:
```
AntoineRichard: [AI review] [P2] Make the startup timeout interrupt the pipe read

The loop over proc.stdout blocks until a line or EOF arrives, so the 300-second deadline is never evaluated if Kit hangs silently during startup. In that failure mode this test can stall the entire shard until the GitHub Actions job timeout.

Please use a bounded readiness mechanism such as select/selectors, a reader thread with a timeout, or equivalent, and preserve the child output in the timeout failure.

hujc7: Fixed in ddc38bde463: readiness now waits on an `Event` fed by a daemon reader thread, so a silent startup hang or an early child exit fails within the 300 s deadline instead of stalling the shard, and the failure report preserves the child's stdout and stderr.

docs/source/features/multi_gpu.rst
```diff
@@ -146,6 +146,25 @@ If the issue persists, additional NCCL fallbacks that may help are:
     export NCCL_IB_DISABLE=1
     export NCCL_ALGO=Ring
 
+On some multi-GPU systems, distributed training with RTX rendering enabled (for example,
+camera-based tasks launched with ``--enable_cameras``) fails when the participating GPUs span
```
kellyguo11: let's avoid mentioning --enable_cameras, it shouldn't be necessary anymore

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1417,23 +1393,99 @@ def _hide_play_button(self, flag):
...
+            # close() only returns when fast shutdown is disabled
+            signal.signal(signum, signal.SIG_DFL)
+            signal.raise_signal(signum)
+
+        def _close_app(self, exit_code):
+            """Close the app with ``exit_code``, warning loudly if the parameter is unsupported."""
+            try:
+                self._app.close(exit_code=exit_code)
+            except TypeError:
+                print(
```
kellyguo11, line 1444: should we use logging instead of print here?

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1417,23 +1393,99 @@ def _hide_play_button(self, flag):
...
+                # builds exit with the status but skip the teardown
+                self._close_app(128 + signum)
+            # close() only returns when fast shutdown is disabled
+            signal.signal(signum, signal.SIG_DFL)
+            signal.raise_signal(signum)
+
+        def _close_app(self, exit_code):
+            """Close the app with ``exit_code``, warning loudly if the parameter is unsupported."""
+            try:
+                self._app.close(exit_code=exit_code)
```
kellyguo11, line 1442: what happens if app.close() hits a different TypeError? could that ever become the case?

kellyguo11 (approved):

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1417,23 +1393,99 @@ def _hide_play_button(self, flag):
...
-        tokens = carb.tokens.get_tokens_interface()
-
-        kit_version = app.get_kit_version()
-        kernel_version = app.get_kernel_version()
-        kit_git_hash = tokens.resolve("${kit_git_hash}") or "unknown"
-
-        print(f"[ISAACLAB] Kit version: {kit_version}", file=sys.__stderr__, flush=True)
-        print(f"[ISAACLAB] Kit kernel:  {kernel_version}", file=sys.__stderr__, flush=True)
-        print(f"[ISAACLAB] Kit hash:    {kit_git_hash}", file=sys.__stderr__, flush=True)
+    class _SimulationAppLifecycle:
```
ooctipus, line 1345: do you want to make this in isaaclab_physx package?

hujc7: might be a bit aggressive for this pr?

docs/source/features/multi_gpu.rst
hujc7: dropped

source/isaaclab/isaaclab/app/app_launcher.py
hujc7: Good catch — yes, it could. A `TypeError` raised *inside* `close()`'s own teardown (rather than from the missing kwarg) would be swallowed by the same `except`, mislabeled "does not accept exit_code," and then trigger a second `close()` on a half-torn-down app — the exact re-entrancy this class exists to prevent. Fixed in `7dfdbd3e93c`: `_close_app` now probes `inspect.signature(self._app.close)` for the `exit_code` parameter up front, so only a genuinely missing parameter takes the fallback; any `TypeError` from within `close()` propagates normally.

source/isaaclab/isaaclab/app/app_launcher.py
hujc7: This fires during teardown, where `carb.logging` may already be disabled (`close()` calls `set_log_enabled(False)`), so I kept it on raw `sys.__stderr__` — the same channel this class uses for its other teardown diagnostics (the startup marker, Kit-version prints), chosen because `__stderr__` survives stdout redirection and log-subsystem shutdown. Happy to switch to `logger.warning` for consistency if you'd prefer, but it'd be less reliable at this point in shutdown.

ooctipus (approved):

AntoineRichard (approved):

PR #6637, opened 2026-07-20
Add controller haptic feedback to G1 teleoperation environments

rwiltz:
# Description

Adds controller haptic feedback to the IsaacTeleop stack: the XR motion
controller vibrates when a teleoperated G1 hand applies force to an object,
giving operators grasp feedback during teleoperation and demo recording.

Introduces a small device-agnostic feedback seam in `isaaclab_teleop` — a
`HapticFeedbackReceiver` protocol (kept off the input-only `DeviceBase`), a
`HapticFeedbackCfg`, and a `HapticFeedbackDriver` that reads per-hand contact
forces from the scene. `IsaacTeleopDevice` implements the protocol by building
an `isaacteleop` `HapticSink` (`ControllerHapticDevice` + `TactileVectorToControllerPulse`)
that reuses the existing controller tracker, and feeds per-step force through the
same external-input path already used for the anchor transform.

Both G1 locomanipulation teleop examples (`IsaacContrib-PickPlace-Locomanipulation-G1-Abs`
and `IsaacContrib-PickPlace-FixedBaseUpperBodyIK-G1-Abs`) gain per-hand finger
`ContactSensor`s and a `haptic_feedback` config; `teleop_se3_agent.py` and
`record_demos.py` drive the shared driver via the existing `hasattr`-based
capability discovery.

Requires `isaacteleop` with the controller haptic device (`ControllerHapticDevice`,
`TactileVectorToControllerPulse`, `HapticSink`).

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_teleop/isaaclab_teleop/haptic_feedback.py
+++ b/source/isaaclab_teleop/isaaclab_teleop/haptic_feedback.py
@@ -1,0 +1,191 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Haptic feedback (controller vibration) for IsaacTeleop teleoperation.
+
+This module adds an *output* path to the otherwise input-only teleop stack:
+a sim-side contact force (e.g. a robot gripper pressing on an object) is
+turned into an XR motion-controller vibration on the corresponding hand.
+
+The design mirrors the existing input pipeline in reverse and keeps the four
+concerns separated:
+
+* **Sensing** lives in the environment -- a :class:`~isaaclab.sensors.ContactSensor`
+  per hand, referenced by name from :class:`HapticFeedbackCfg`.
+* **Output port** is the narrow :class:`HapticFeedbackReceiver` protocol.  Only
+  devices that can vibrate implement it; input-only devices (keyboard, gamepad)
+  are unaffected, so the base ``DeviceBase`` contract stays input-only.
+* **Mapping + transport** lives inside :class:`~isaaclab_teleop.IsaacTeleopDevice`,
+  which builds an ``isaacteleop`` ``HapticSink`` fed by a
+  ``TactileVectorToControllerPulse`` -- Isaac Lab supplies the force, ``isaacteleop``
+  owns the force->amplitude curve and the OpenXR ``xrApplyHapticFeedback`` write.
+* **Orchestration** is :class:`HapticFeedbackDriver`, used identically by every
+  teleop script: it reads the contact sensors after ``env.step`` and forwards a
+  per-hand force scalar to the device.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol, runtime_checkable
+
+import torch
+
+from isaaclab.utils.configclass import configclass
+
+if TYPE_CHECKING:
+    from isaaclab.envs import ManagerBasedRLEnv
+
+# Well-known endpoint identifiers for the two motion controllers. These match
+# ``isaacteleop``'s ``Endpoint`` strings for hand-mounted haptic devices.
+ENDPOINT_LEFT: str = "left"
+ENDPOINT_RIGHT: str = "right"
+
+
+@runtime_checkable
+class HapticFeedbackReceiver(Protocol):
+    """Protocol for a teleop device that can render haptic feedback.
+
+    This is deliberately a single-method protocol rather than an addition to
+    ``DeviceBase``: input-only devices (keyboard, space-mouse, gamepad) cannot
+    honor a feedback call, so forcing the method onto every device would break
+    Liskov substitutability.  Scripts guard with ``isinstance(device,
+    HapticFeedbackReceiver)`` and only drive feedback when the active device
+    actually supports it.
+    """
+
+    def send_haptic(self, endpoint: str, force: float) -> None:
+        """Render a haptic pulse on one controller from a contact force.
+
+        Args:
+            endpoint: Which controller to vibrate -- :data:`ENDPOINT_LEFT` or
+                :data:`ENDPOINT_RIGHT`.
+            force: Contact-force magnitude [N] on that hand.  ``0`` stops any
+                active pulse.  The force-to-amplitude mapping (gain, deadband,
+                saturation) is applied downstream by the device.
+        """
+        ...
+
+
+@configclass
+class HapticFeedbackCfg:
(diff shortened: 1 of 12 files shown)


rwiltz:
@greptile

rwiltz:
@greptile

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
```diff
@@ -443,6 +477,72 @@ def _build_control_pipeline(channel_uuid: bytes) -> tuple[Any, TeleopMessageProc
...
+
+        Args:
+            controllers_source: The ``ControllersSource`` whose tracker the
+                haptic device should drive.
+
+        Returns:
+            The connected ``HapticSink`` node, ready to pass to
+            ``TeleopSessionConfig(sinks=[...])``.
+        """
+        from isaacteleop.haptic_devices.controller import ControllerHapticDevice
```
hougantc-nvda, line 613: Could we also raise the declared `isaacteleop` lower bound in the root `pyproject.toml`? The working environment has `1.3.132rc1` and the lockfile's `1.3.129rc1` is also compatible, but the current `~=1.3.0` constraint permits older releases such as `1.3.75rc1`, which I verified lacks `ControllerHapticDevice`, `TactileVectorToControllerPulse`, and `HapticSink`. An existing environment at that version can satisfy dependency resolution and then fail here at runtime. The lower bound should be the first release containing these haptic APIs.

source/isaaclab_teleop/isaaclab_teleop/haptic_feedback.py
```diff
@@ -0,0 +1,191 @@
...
+        per-body contact-force magnitudes for the first environment.  This
+        favors a stable "am I pressing on something" signal over directional
+        detail, which is all a single rumble motor can render.
+        """
+        for endpoint, sensor in self._endpoint_sensors.items():
+            # net_forces_w: [num_envs, num_bodies, 3]; reduce env 0 to a scalar [N].
+            net_forces = sensor.data.net_forces_w
+            if net_forces is None:
+                continue
+            force = torch.linalg.vector_norm(net_forces[0], dim=-1).sum().item()
```
hougantc-nvda: `net_forces_w` is a `ProxyArray` in the current sensor API, so indexing it implicitly emits the `ProxyArray` deprecation warning. Please use the explicit Torch view (`net_forces.torch`). Also, each per-hand `.item()` forces a CUDA-to-host synchronization; consider reducing both hands and transferring the two scalars together to avoid two serialized sync points per teleop frame.

source/isaaclab_teleop/isaaclab_teleop/haptic_feedback.py
```diff
@@ -0,0 +1,191 @@
...
+        The force is reduced to a single scalar per hand: the sum of the
+        per-body contact-force magnitudes for the first environment.  This
+        favors a stable "am I pressing on something" signal over directional
+        detail, which is all a single rumble motor can render.
+        """
+        for endpoint, sensor in self._endpoint_sensors.items():
+            # net_forces_w: [num_envs, num_bodies, 3]; reduce env 0 to a scalar [N].
+            net_forces = sensor.data.net_forces_w
+            if net_forces is None:
+                continue
```
hougantc-nvda: If sensor data is unavailable, `continue` leaves the previously cached force active. Because the sink re-emits that cache, the controller can keep vibrating with a stale value. This output path should fail safe by sending `0.0` for the endpoint when `net_forces_w` is `None`.

hougantc-nvda (commented):
Three findings from a local review of the haptic-feedback implementation:

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
hougantc-nvda: In practice, I don't think this is an issue. This is only an issue if user pulls latest and doesn't run ./isaaclab.sh --install.

hougantc-nvda (approved):
I tested after pulling and building this branch and haptics were working.

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
rwiltz: Yes, I did that here https://github.com/isaac-sim/IsaacLab/pull/6655

source/isaaclab_teleop/isaaclab_teleop/isaac_teleop_device.py
```diff
@@ -282,6 +292,22 @@ def advance(self, target_T_world: np.ndarray | torch.Tensor | SupportsDLPack | N
...
         return action
 
+    def send_haptic(self, endpoint: str, force: float) -> None:
+        """Render a haptic pulse on one controller from a contact force.
+
+        Implements the :class:`~isaaclab_teleop.HapticFeedbackReceiver` protocol.
+        The force is cached and injected into the haptic sink on the next
+        :meth:`advance`; the force -> amplitude mapping and the OpenXR write are
+        handled by the ``isaacteleop`` haptic sink built at session start.  This
+        is a no-op unless the device was constructed with a ``haptic_cfg``.
```
AntoineRichard: Could this be shortened a bit? I think this `; the force -> amplitude mapping and the OpenXR write are
        handled by the ``isaacteleop`` haptic sink built at session start.` Is not really required?

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
```diff
@@ -140,6 +151,11 @@ def __init__(
                 OpenXR connection and feeds the recorded tracker stream
                 through the pipeline.  Mutually exclusive with
                 *mcap_record_path*.
+            haptic_cfg: Optional haptic-feedback configuration.  When provided,
+                :meth:`start` builds an ``isaacteleop`` ``HapticSink`` subgraph
+                (fed by ``TactileVectorToControllerPulse``) so that per-hand
+                contact forces pushed via :meth:`push_haptic` are rendered as
+                controller vibration.  ``None`` disables haptics entirely.
```
AntoineRichard: Same here, i think this could be reduced by removing: `:meth:`start` builds an ``isaacteleop`` ``HapticSink`` subgraph
                (fed by ``TactileVectorToControllerPulse``) so that`

source/isaaclab_teleop/isaaclab_teleop/isaac_teleop_device.py
rwiltz: Done

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
rwiltz: Done

kellyguo11 (approved):

PR #6640, opened 2026-07-21
Drive fragment schema writers by prim path expressions

vidurv-nvidia:
# Description

Moves the physics schema **fragment** surface onto prim path expressions, so a writer targets prims
by pattern instead of traversing the subtree of a single input prim.

Since this PR was opened, #6841 landed the plain-regex prim path matcher on `develop`. This PR has
been rebuilt on top of it: it no longer touches `sim/utils/queries.py` at all, and the bespoke `**`
recursive token it originally introduced is gone. Targeting is now expressed entirely in standard
Python regular expressions matched against whole prim paths.

## What changes

**Family writers take an expression.** `apply_rigid_body_properties`,
`apply_collision_properties`, `apply_mass_properties`, `apply_articulation_root_properties`,
`apply_joint_drive_properties`, and the two tendon writers now take `prim_path_expr` and resolve
targets through `find_matching_prims`. A bare prim path matches only that prim; `f"{prim_path}/.*"`
reaches its descendants and `f"{prim_path}(/.*)?"` reaches the prim together with its subtree.

**Spawner fields take a mapping.** The fragment fields (`rigid_props`, `collision_props`,
`mass_props`, `articulation_props`, `joint_drive_props`, `fixed_tendons_props`,
`spatial_tendons_props`) accept `dict[str, list[<Family>Fragment]]`. A key is a regular-expression
**suffix appended to the prim the spawner anchors that family on**, so it carries its own leading
`/` when it targets descendants:

| Key | Selects |
|---|---|
| `""` | the anchor prim itself |
| `"/[^/]+"` | its direct children |
| `"/.*"` | all of its descendants |
| `"(/.*)?"` | the anchor together with its descendants |

Entries apply in insertion order, so on overlapping targets later entries override earlier ones per
attribute. Legacy single-cfg values are unaffected and still route to the legacy writers.

```python
spawn = sim_utils.UsdFileCfg(
    usd_path=...,
    rigid_props={
        "(/.*)?": [UsdPhysicsRigidBodyCfg(rigid_body_enabled=True), PhysxRigidBodyCfg(...)],
        "/.*_hand(/.*)?": [PhysxRigidBodyCfg(max_depenetration_velocity=1.0)],
    },
)
```

**Explicit API creation.** The writers no longer apply their defining USD API implicitly on a bare
prim. `create_if_missing=True` (and the spawner flags `mass_props_create_if_missing`,
`articulation_props_create_if_missing`, `joint_drive_props_create_if_missing`) applies the defining
API to matched prims that lack it.

**Articulation roots.** `apply_articulation_root_properties` authors on every matched root and warns
when roots nest, rather than silently pruning nested ones. Asset validity is the author's
responsibility.

**Tendon backends de-traverse.** The PhysX and Newton tendon fragment functions now author on the
given prim only; target selection is owned by the core family writers.

**Deformable meshes accept the mapping too.** On the mesh spawner's deformable branch,
`collision_props` given as a mapping previously failed the fragment type check and raised. It also
passed the bare body prim to the collision writer, which under expression matching no longer
reaches the simulation mesh authored beneath it. Mapping entries now anchor at the body prim (e.g.
`{"/sim_mesh": [...]}`) and a bare fragment list uses a subtree pattern to reach the collider.

## Fixes

Rigid-body and mass fragments previously reached only the outermost schema-bearing prim on assets
with nested rigid-body hierarchies (child links authored under their parent link prims, as produced
by the URDF importer in Isaac Sim 6.0 and later). A whole-subtree pattern now reaches every carrier.

## Type of change

**Breaking within the fragment surface only.** The fragment path is experimental and has no
in-tree users yet; legacy `define_*` / `modify_*` writers and legacy cfg classes are untouched.
Migration for anyone already on fragments: wrap lists as `{"(/.*)?": [...]}`, and pass
`create_if_missing=True` where an API was previously created implicitly.

## Screenshots

N/A

## Checklist

## Testing

Per-file, via the wrapped Python:

| Suite | Result |
|---|---|
| `test_utils_queries` | 12 passed |
| `test_schema_fragments` | 12 passed |
| `test_schema_writer_nested_targets` | 13 passed |
| `test_mass_fragments` | 9 passed |
| `test_collision_fragments` | 19 passed |
| `test_articulation_fragments` | 28 passed, 1 xfailed |
| `test_joint_drive_fragments` | 22 passed |
| `test_tendon_fragments` | 23 passed |
| `test_mesh_converter` | 15 passed |
| `test_spawn_meshes` (physx) | 3 passed |

### Note on the xfail

`test_physx_fix_root_link_migrates_preauthored_newton_root_api` is marked `xfail`. It is not caused
by this PR. `develop`'s experience files were restructured so the schema plugin that registered
`NewtonArticulationRootAPI` no longer loads, making it an unregistered token schema.
`PhysicsManager._relocate_articulation_root` only migrates applied schemas it can resolve through
`Usd.SchemaRegistry.FindAppliedAPIPrimDefinition`, so on root relocation the token and its authored
`newton:selfCollisionEnabled` value are left stranded on the former root link. Migrating it needs a
backend-registered schema-to-namespace mapping, which is out of scope here — flagging it for a
follow-up.


source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -1313,55 +1390,47 @@ def apply_joint_drive_properties(
...
-    if not root_prim.IsValid():
-        raise ValueError(f"Prim path '{prim_path}' is not valid.")
+    # match revolute/prismatic joints not excluded by a backend-registered skip predicate;
+    # non-joint matches (creation_candidates) are ignored silently since a ``**`` expression matches
+    # whole subtrees
+    targets, _, _ = _match_fragment_targets(
+        prim_path_expr,
+        lambda p: (p.IsA(UsdPhysics.RevoluteJoint) or p.IsA(UsdPhysics.PrismaticJoint)) and not _skip_joint_drive(p),
+        stage,
+    )
```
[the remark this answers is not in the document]
vidurv-nvidia: Confirmed and fixed in f69a3d4285d: the writer now captures the instanced-skip signal, gates the zero-success warning on it (the matcher already reported the skips), and folds it into the return value like the sibling writers. Locked by a regression test that asserts a single warning and a False return for an instanced joint.

source/isaaclab/isaaclab/sim/utils/queries.py
```diff
@@ -385,6 +403,15 @@ def find_matching_prims(prim_path_regex: str, stage: Usd.Stage | None = None) ->
...
             output_prims = []
+    if not tokens:
+        output_prims = all_prims
+    if recursive:
+        # expand each anchor to itself plus all descendants, traversing into instances so
+        # callers can detect read-only proxy prims
+        expanded = []
+        for prim in output_prims:
+            expanded.extend(Usd.PrimRange(prim, Usd.TraverseInstanceProxies()))
+        output_prims = expanded
```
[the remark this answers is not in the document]
vidurv-nvidia: Confirmed and fixed in f69a3d4285d: the recursive expansion now uses Usd.TraverseInstanceProxies(Usd.PrimAllPrimsPredicate), matching the per-token traversal's inclusion of inactive prims. Regression test compares the two traversals on a deactivated prim.

docs/source/overview/core-concepts/schema_fragments.rst
```diff
@@ -0,0 +1,178 @@
...
+.. code-block:: python
+
+   import isaaclab.sim as sim_utils
+   from isaaclab.sim.schemas import MassCfg, UsdPhysicsCollisionCfg, UsdPhysicsRigidBodyCfg
+   from isaaclab_newton.sim.schemas import NewtonCollisionCfg
+
+   cuboid = sim_utils.CuboidCfg(
+       size=(0.1, 0.1, 0.1),
+       rigid_props={"": [UsdPhysicsRigidBodyCfg()]},
+       mass_props={"": [MassCfg(mass=0.5)]},
```
ooctipus:
```
       rigid_props={"": [UsdPhysicsRigidBodyCfg()]},
       mass_props={"": [MassCfg(mass=0.5)]},for this pattern

```

I remember you told me there were reason why you dont want


```
       rigid_props=UsdPhysicsRigidBodyCfg(),
       mass_props=MassCfg(mass=0.5),

```
can you remind me again?

docs/source/overview/core-concepts/schema_fragments.rst
```diff
@@ -0,0 +1,178 @@
...
+   from isaaclab.sim.schemas import UsdPhysicsDriveCfg, UsdPhysicsRigidBodyCfg
+   from isaaclab_newton.sim.schemas import MujocoRigidBodyCfg
+   from isaaclab_physx.sim.schemas import PhysxRigidBodyCfg
+   from isaaclab.utils.assets import ISAAC_NUCLEUS_DIR
+
+   spawn = sim_utils.UsdFileCfg(
+       usd_path=f"{ISAAC_NUCLEUS_DIR}/Robots/Franka/franka_instanceable.usd",
+       rigid_props={
+           # every rigid body: universal + PhysX + MuJoCo attributes side by side
+           "(/.*)?": [
```
ooctipus: why not `.*`

source/isaaclab/isaaclab/sim/converters/mesh_converter_cfg.py
```diff
@@ -12,29 +12,43 @@
...
+    per attribute.
+
     Note:
         If None, then no rigid body properties will be added.
     """
 
-    collision_props: (
-        schemas_cfg.CollisionPropertiesCfg | schemas_cfg.CollisionFragment | list[schemas_cfg.CollisionFragment]
-    ) = None
+    collision_props: dict[str, list[schemas_cfg.CollisionFragment]] | schemas_cfg.CollisionPropertiesCfg | None = None
```
ooctipus, line 43: looks like it can accept both dict or direct Cfg

so mass_props=MassCfg(mass=0.5), should work ?

ooctipus (approved):

kellyguo11:
run-ci

PR #6645, opened 2026-07-21
Fix Warp import ordering for Newton renderer

yts-nv:
# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/__init__.py
+++ b/source/isaaclab/isaaclab/__init__.py
@@ -51,32 +51,47 @@ def _deprioritize_prebundle_paths():
                 return True
         return False
 
-    # Partition: keep non-conflicting in place, collect conflicting.
-    clean = []
-    demoted = []
-    for p in sys.path:
-        if _should_demote(p):
-            demoted.append(p)
-        else:
-            clean.append(p)
-
-    if not demoted:
-        return
+    def _demote_paths(paths):
+        # Partition: keep non-conflicting in place, collect conflicting.
+        clean = []
+        demoted = []
+        for p in paths:
+            if _should_demote(p):
+                demoted.append(p)
+            else:
+                clean.append(p)
+        return clean + demoted, bool(demoted)
 
     # Rebuild sys.path: originals first, then demoted at the very end.
-    sys.path[:] = clean + demoted
+    sys.path[:], _ = _demote_paths(sys.path)
 
     # Rewrite PYTHONPATH with the same ordering for subprocesses.
     if "PYTHONPATH" in os.environ:
         parts = os.environ["PYTHONPATH"].split(os.pathsep)
-        env_clean = []
-        env_demoted = []
-        for p in parts:
-            if _should_demote(p):
-                env_demoted.append(p)
-            else:
-                env_clean.append(p)
-        os.environ["PYTHONPATH"] = os.pathsep.join(env_clean + env_demoted)
+        os.environ["PYTHONPATH"] = os.pathsep.join(_demote_paths(parts)[0])
+
+    # Kit can mutate a loaded package's search path directly.  If ``warp`` is
+    # already imported and an older ``omni.warp.core`` path is appended to
+    # ``warp.__path__``, a later ``import warp.fem`` can still mix the extension
+    # package with Isaac Lab's pip-installed Warp even when ``sys.path`` is
+    # clean.  Reorder loaded package paths with the same policy.
+    for module in tuple(sys.modules.values()):
+        package_path = getattr(module, "__path__", None)
+        if package_path is None:
+            continue
+        try:
+            reordered, changed = _demote_paths(list(package_path))
+        except TypeError:
+            continue
+        if not changed:
+            continue
+        try:
+            package_path[:] = reordered
+        except TypeError:
+            try:
+                module.__path__ = reordered
+            except Exception:
+                pass
 
 
 _deprioritize_prebundle_paths()
--- a/source/isaaclab/isaaclab/utils/backend_utils.py
+++ b/source/isaaclab/isaaclab/utils/backend_utils.py
@@ -15,6 +15,15 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _sanitize_import_paths() -> None:
+    """Re-apply Isaac Lab import path ordering before dynamic backend imports."""
+    try:
+        from isaaclab import _deprioritize_prebundle_paths
+    except (AttributeError, ImportError):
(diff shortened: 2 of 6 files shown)


source/isaaclab/isaaclab/__init__.py
```diff
@@ -51,32 +51,47 @@ def _should_demote(path: str) -> bool:
...
-                env_clean.append(p)
-        os.environ["PYTHONPATH"] = os.pathsep.join(env_clean + env_demoted)
+        os.environ["PYTHONPATH"] = os.pathsep.join(_demote_paths(parts)[0])
+
+    # Kit can mutate a loaded package's search path directly.  If ``warp`` is
+    # already imported and an older ``omni.warp.core`` path is appended to
+    # ``warp.__path__``, a later ``import warp.fem`` can still mix the extension
+    # package with Isaac Lab's pip-installed Warp even when ``sys.path`` is
+    # clean.  Reorder loaded package paths with the same policy.
+    for module in tuple(sys.modules.values()):
```
AntoineRichard, line 78: This is much too broad for a Warp-specific problem. Iterating over every entry in `sys.modules` and using `getattr(module, "__path__", None)` executes module-level `__getattr__` hooks. I reproduced an unrelated loaded module raising `RuntimeError` from its `__getattr__`, which aborts this sanitizer and therefore backend resolution. Please do not scan and mutate every loaded package; first identify which extension and app configuration are introducing the conflicting path and fix that boundary.

source/isaaclab/isaaclab/__init__.py
```diff
@@ -51,32 +51,47 @@ def _should_demote(path: str) -> bool:
...
+        if package_path is None:
+            continue
+        try:
+            reordered, changed = _demote_paths(list(package_path))
+        except TypeError:
+            continue
+        if not changed:
+            continue
+        try:
+            package_path[:] = reordered
```
AntoineRichard, line 89: Reordering `__path__` cannot safely repair a package whose root was already loaded from the wrong distribution. `warp.__file__` and already-imported `warp.*` modules remain from bundled Warp, while future submodules can resolve from pip Warp. I reproduced exactly that state against this PR; the subsequent `import warp.fem` can therefore create the mixed package this code claims to prevent. We need to guarantee the desired Warp distribution before its first import, or detect the wrong origin and fail clearly rather than attempting in-place repair.

source/isaaclab/isaaclab/utils/backend_utils.py
```diff
@@ -114,8 +123,10 @@ def resolve_class(cls, *args, **kwargs) -> type:
             # Construct the module name from the backend and the determined subpath.
             module_name = cls._get_module_name(backend)
             try:
+                _sanitize_import_paths()
```
AntoineRichard, line 126: Please do not make every backend factory resolution rerun global import-state repair for a Newton-renderer issue. Our app configurations already exclude `omni.warp.core`, `isaacsim.pip.newton`, `omni.kit.pip_archive`, and the other conflicting extensions. The first question should be why those exclusions are not effective in the failing run: is the launcher using the wrong app configuration, or is an unexpected dependency/configuration bypassing the exclusion? Fix that root cause and cover the actual launch path with an integration regression test.

source/isaaclab_newton/isaaclab_newton/__init__.py
```diff
@@ -7,6 +7,31 @@
...
+    except (AttributeError, ImportError):
+        def _deprioritize_prebundle_paths() -> None:
+            return
+
+    _deprioritize_prebundle_paths()
+    try:
+        import warp  # noqa: F401
+
+        _deprioritize_prebundle_paths()
+        import warp.fem  # noqa: F401
```
AntoineRichard, line 24: This defeats `isaaclab_newton`'s lazy import boundary: every package import now eagerly imports `warp.fem`, although FEM is used by the MPM manager rather than the Newton renderer. The broad `except ImportError` also hides genuine import-time failures inside `warp.fem`. Please keep FEM at its concrete use site and solve Warp ownership/configuration before package import instead.

source/isaaclab/test/test_import_path_sanitization.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def test_factory_sanitizes_paths_before_dynamic_backend_import(monkeypatch):
+    calls = []
+    backend_module = types.ModuleType("_fake_backend_module")
+
+    class DummyFactoryImpl:
+        pass
+
+    backend_module.DummyFactory = DummyFactoryImpl
```
AntoineRichard, line 50: This assignment means the test never exercises `backend_module.__getattr__`: `getattr(module, class_name)` finds `DummyFactory` directly in the module dictionary. Consequently the assertion in `lazy_getattr` is dead code and the second sanitizer call is not verified. More importantly, the PR needs a reproducer for the real app/renderer failure that asserts the origins of `warp`, `warp.fem`, and relevant loaded submodules, rather than only testing calls between new mocks.

AntoineRichard (changes requested):
# Review summary

Requesting changes. This patch tries to repair a partially loaded Python environment by repeatedly mutating global import state. That is not a safe foundation for resolving two incompatible Warp distributions, and it can create the mixed-package state it is intended to prevent.

The first task should be root-cause analysis of the extension exclusions. All Isaac Lab app configurations already exclude `omni.warp.core`, `isaacsim.pip.newton`, `omni.kit.pip_archive`, and the other known conflicting extensions. Please establish:

1. The exact failing command and app configuration/experience file.
2. Whether the failing path is using the intended Isaac Lab app configuration at all.
3. If it is, why Kit still loads or exposes the excluded extension path: an unexpected dependency, an exclusion that is not honored, or another configuration being layered in.
4. The origins of `warp`, already-loaded `warp.*` modules, and `warp.fem` before and after Kit startup.

A good solution should fix ownership at that configuration/dependency boundary. If the wrong app config is selected, fix app selection. If an extension dependency bypasses the exclusions, fix the dependency/exclusion chain. We should ensure one Warp distribution wins before the first import, or fail with a clear diagnostic when that invariant is violated.

I do not think we should solve this by scanning and mutating every loaded package, rerunning the mutation from every backend factory, or eagerly importing `warp.fem` from `isaaclab_newton.__init__`. Please add a real integration regression test for the failing launch/renderer path, fill in the PR description with the reproducer and root cause, and add the required changelog fragments for both touched packages.

kellyguo11:
will try to address this in https://github.com/isaac-sim/IsaacLab/pull/6499

PR #6651, opened 2026-07-21
Sim parameter validation

rubengrandia:
# Description

Work in progress on setting up unit tests that validate that parameters can be authored via either USD, python configs, or runtime writes to the articulation. Through simple simulation examples we verify that the resulting physics is as expected.


rubengrandia:
No need for detailed review at the moment. I can't unassign reviewers even if it is a draft.. 
The point was that @AntoineRichard  can follow along.

rubengrandia:
Closing this work in progress to avoid triggering CI

PR #6656, opened 2026-07-21
Installation docs re-write

StafaH:
# Description

This PR focuses on 3 main things:

- Re-write of the installation docs, with a focus on uv run
- Repo-wide cleanup of --headless and --enable-cameras
- Repo-wide cleanup to make uv run the recommended method in all docstrings, comments, skills, etc.

## Checklist


frlai (approved):

AntoineRichard (changes requested):
Calling this PR `docs/source/deployment/index.rst` is widely misleading, it provides a new doc app for tasks, as well as moves section like docker, changes things around headless behavior. I understand there is some urgency, but let's split PRs like this. For instance I don't agree with the doc formatting changes but some of the rest looks fine to me, do I block it do I approve it? Can you move the doc formating change in another PR?

source/isaaclab/test/app/test_argparser_launch.py
```diff
@@ -46,14 +46,14 @@ def test_visualizer_alias_parsing():
     assert args.visualizer_explicit is True
 
 
-def test_headless_deprecated_arg_parsing():
-    """Test that deprecated --headless is still accepted by the parser."""
+@pytest.mark.parametrize("deprecated_arg", ["--head" + "less", "--enable_" + "cameras"])
```
AntoineRichard: 🤖 **AI-assisted review:** These concatenations appear intended to keep the exact deprecated spellings out of source-level searches while still testing parser rejection. That is too opaque without a comment or an enforced literal-ban check, and the audit currently gives a false sense of completeness: this branch still contains six active literal `--headless` usages in `.vscode/tools/launch.template.json`, `scripts/benchmarks/benchmark_newton_raycast.py`, and `tools/template/templates/external/.vscode/tools/launch.template.json`. The VS Code configurations generated from those templates now fail because this parser rejects the flag. Please remove those remaining usages and either spell the test inputs normally or explain and enforce the source-search convention.

docs/index.rst
```diff
@@ -127,6 +126,7 @@ Table of Contents
    :caption: Features
 
    source/features/isaac_teleop
+   source/features/docker_cloud
```
AntoineRichard, line 129: 🤖 **AI-assisted review:** Consolidating the material into `features/docker_cloud` preserves most content and the included Docker/cluster anchors, but deleting the original pages without compatibility stubs breaks all published URLs under `source/deployment/` plus `source/setup/installation/cloud_installation.html`. The configured documentation tooling has no per-page redirect mechanism. Please keep orphaned compatibility pages that point readers to the new sections, or add explicit redirects, so existing bookmarks and inbound links do not become 404s.

source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -945,19 +902,9 @@ def _resolve_visualizer_settings(self, launcher_args: dict) -> None:
...
-            raise ValueError(
-                f"Invalid value for environment variable `ENABLE_CAMERAS`: {enable_cameras_env} ."
-                f"Expected: {enable_cameras_valid_vals} ."
-            )
-        # We allow enable_cameras kwarg to supersede ENABLE_CAMERAS envvar
-        if enable_cameras_arg is True:
-            self._enable_cameras = enable_cameras_arg
-        else:
-            self._enable_cameras = bool(enable_cameras_env)
+        self._enable_cameras = bool(
```
AntoineRichard, line 926: 🤖 **AI-assisted review:** Removing the `ENABLE_CAMERAS` fallback exposes a camera-launch gap. `launch_simulation(cfg, launcher_args=None)` explicitly skips its camera auto-enable block because it requires `launcher_args is not None`; it then reaches `AppLauncher(None)`, where this code resolves `enable_cameras` to `False`. Thus the public no-arguments path does not auto-enable Kit camera rendering even though `launch_simulation` documents that behavior, and the former environment workaround is gone. Please normalize `None` to a mutable empty argument mapping before the scan/auto-enable flow and add coverage for a Kit-camera config with omitted launcher arguments.

docs/source/_static/css/environment-browser.js
```diff
@@ -0,0 +1,302 @@
...
+            [/UnitreeGo1/, "tasks/locomotion/go1_flat.jpg"],
+            [/UnitreeGo2/, "tasks/locomotion/go2_flat.jpg"],
+        ];
+        return imageRules.find(([pattern]) => pattern.test(taskName))?.[1] || "tasks/classic/cartpole.jpg";
+    };
+
+    const updateTaskControls = () => {
+        const task = selectedTask();
+        populateSelect(fields.rl, task.rl, [fields.rl.value, "rsl_rl", "rl_games", "skrl", "sb3"]);
+        populateSelect(fields.physics, task.physics, [fields.physics.value, "newton_mjwarp", "physx", "newton_kamino"]);
```
AntoineRichard, line 168: 🤖 **AI-assisted review:** The browser models physics and renderer choices as independent lists, so it offers combinations that the launcher rejects. For example, `Isaac-Cartpole-Camera-Direct` exposes both `physx` and `ovrtx`; selecting those produces `physics=physx renderer=ovrtx`, which `_validate_runtime()` rejects. Changing physics also leaves the current renderer untouched. Please represent valid combinations rather than taking a Cartesian product, or filter and reset dependent selectors whenever physics changes. The hard-coded task matrix should ideally be generated and checked against the actual environment metadata.

docs/conf.py
```diff
@@ -324,7 +324,8 @@ def _read_pinned_versions() -> dict:
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = ["source/_static/css"]
-html_css_files = ["custom.css"]
+html_css_files = ["custom.css", "environment-browser.css"]
+html_js_files = ["environment-browser.js"]
```
AntoineRichard, line 328: 🤖 **AI-assisted review (non-blocking structural note):** This adds a 481-line custom stylesheet and a 302-line JavaScript mini-app, places the JavaScript under `_static/css/` because that directory is the entire static root, and loads both assets on every documentation page even though only one page uses them. The browser data is also manually duplicated in JavaScript despite the `Generated` comment, with no generator or synchronization test. Please consider moving this browser to a focused follow-up PR and implementing it through generated Sphinx data/page-scoped assets; at minimum, move the JavaScript to an appropriate static directory and document how the task matrix is regenerated.

source/isaaclab/test/app/test_argparser_launch.py
AntoineRichard: Let's remove the concat that was used to avoid returning hits in the search and replace phase

AntoineRichard (approved):
CI is congested, let's get this in the way it is.

kellyguo11 (approved):

PR #6658, opened 2026-07-22
 OVRTX visualization: add NewtonRTX visualizer and Streaming View feature to existing visualizers

matthewtrepte:
# Description

To support OVRTX in visualization, change adds NewtonRTX visualizer and adds a streaming view feature to all visualizers (except NewtonRTX) which can show arbitrary pixel streams, like ovrtx renders. Also tweaks the VideoRecording feature.

NewtonRTX
- Add initial NewtonRTX visualizer which renders with OVRTX. Rename existing Newton Visualizer to NewtonGL Visualizer.
- Both Newton Visualizers inherit from the base parent class but use different NewtonViewer classes.
- Currently the NewtonRTX Visualizer has a number of limitations due no-ops in the NewtonRTXViewer class (no visualizer markers, live plots, streaming tiled view).

Visualizer Streaming View
- All visualizers except NewtonRTX can show pixel streams in additional panels
- The original Tiled Camera View in Newton and Kit visualizers is expanded and renamed to Streaming View
- Streaming view supports tiled cameras, existing sensors or custom visualizer cameras which can follow specified robot bodies, non RGB groundtruth, like depth, segmentation, and normals. and for RGB, specified renderers like OVRTX

Tweaks to Video Recording
- We now support headless Kit and Newton visualizers, and users can record these headless visualizers. Also when headless visualizers are not being recorded, they don't render.
- To preserve original behavior, if the user just provides --video, and doesn't provide a visualizer or sensor source, we automatically launch a headless kit visualizer and set the recording_src to "visualizer:kit" to capture the default kit visualizer viewport

## Type of change

## Screenshots

Some examples from these features

https://github.com/user-attachments/assets/aa37fdac-d9c3-4506-8a82-9ff19ecb6b6b

https://github.com/user-attachments/assets/18d9a48e-3a26-4ceb-802b-d9fde32c8097

https://github.com/user-attachments/assets/712acac3-376b-4232-9e85-dd3c70c6faff

## Checklist


pbarejko:
Hey @mataylor-nvidia you have ran into some issues when using ovrtx with visualizer? Can you share the details?

rilei-nvidia:
Hey @matthewtrepte we had been talking about needing a visualizer for the OVRTX renderer so this looks very useful.  Would you be able to attach some screenshots of what this looks like?

matthewtrepte:
@isaaclab-review-bot review

source/isaaclab/isaaclab/visualizers/visualizer_cfg.py
```diff
@@ -35,38 +35,61 @@ class VisualizerCfg:
...
-    tiled_cam_view: bool = False
-    """Enable a non-interactive tiled camera image view."""
+    # ── Streaming view ────────────────────────────────────────────────────────
+    # Captures pixels from a camera sensor (existing or auto-created), tiles them
+    # across envs and GT types, and shows the result as an image panel in interactive
+    # visualizers (Newton GL, Kit) or pushes it per-step to sink-based ones (Rerun, Viser).
 
-    tiled_cam_num: int = 16
-    """Number of camera tiles to show when tiled_cam_env_indices is None, capped at 100."""
+    streaming_view: bool = True
```
kellyguo11: 🔴 **Streaming view is unintentionally enabled by default.** This makes ordinary visualizer configurations create and update a camera unless they explicitly opt out. It also changes Viser's default behavior to suppress `log_state()` whenever streaming is active, so a normal Viser session loses its 3D scene. The submitted configuration and initialization tests expect streaming to be opt-in. Please restore the default to `False`.

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer_cfg.py
```diff
@@ -3,18 +3,22 @@
...
 @configclass
 class NewtonVisualizerCfg(VisualizerCfg):
-    """Configuration for Newton OpenGL visualizer."""
+    """Shared configuration base for Newton visualizer backends.
 
-    visualizer_type: str = "newton"
-    """Type identifier for Newton visualizer."""
+    Use :class:`NewtonGLVisualizerCfg` for the OpenGL rasterizer or
+    :class:`NewtonRTXVisualizerCfg` for the OVRTX path tracer. This class is
+    not intended to be instantiated directly.
```
kellyguo11: 🔴 **`NewtonVisualizerCfg` is now non-instantiable.** This public config previously selected the Newton GL backend; without a `visualizer_type`, passing it through `create_visualizer()` now raises. Current documentation, tutorials, and even the factory error message still instantiate or recommend this class. Since this is a breaking API change, please update those callers and the migration guidance consistently, or retain it as a functional alias for `NewtonGLVisualizerCfg`.

docs/source/overview/core-concepts/visualization.rst
```diff
@@ -228,50 +237,82 @@ selection and visual offsets. This includes point-cloud and task-geometry marker
...
+ground-truth frames into a single image panel updated every step.
 
-For runnable Kit and Newton examples that use generated and existing tiled cameras,
-see :doc:`/source/how-to/visualizer_tiled_camera`.
+.. note::
 
-The default visualizer camera mode is interactive, with ``eye`` and ``lookat`` specifying the initial pose.
-Kit and Newton visualizers can also run additional tiled camera image panels.
+   The legacy ``tiled_cam_*`` fields (``tiled_cam_view``, ``tiled_cam_prim_path``, etc.) have been
+   replaced by the ``streaming_*`` fields described in the :ref:`streaming-camera-view` section below.
```
kellyguo11, line 247: 🟡 **This is a breaking configuration API change.** Removing the `tiled_cam_*` fields causes existing config construction to fail with `TypeError`; it is more than a field rename. Please identify it as breaking in the changelog/PR migration notes and update every in-tree caller. `run_video_recording.py` example 3 still passes the removed fields, as noted in the existing inline thread.

source/isaaclab/isaaclab/envs/utils/camera_view.py
```diff
@@ -114,15 +120,132 @@ def ensure_camera_initialized(camera: Camera) -> None:
...
 ) -> tuple[Camera, list[str]]:
-    """Create an internal RGB Camera sensor for visualizer image views."""
+    """Create an internal Camera sensor for visualizer image views.
+
+    When the renderer is a singleton (e.g. OVRTX) a shared instance is returned
+    on subsequent calls to avoid the ``/Render`` prim duplication error.
+    """
+    renderer_key = type(renderer_cfg).__name__
+    if renderer_key in _shared_streaming_cameras:
+        return _shared_streaming_cameras[renderer_key]
```
kellyguo11: 🔴 **The shared-camera lookup ignores the requested camera contract and lifecycle.** Keying only by renderer class means a later visualizer silently reuses the first camera's environment count, resolution, GT channels, and stage. The callers also all mark the returned camera as owned, so their pose updates and cleanup can interfere with each other. Please scope the cache to the current stage and full compatible configuration with reference-counted ownership, or share the singleton renderer without sharing the `Camera` object.

source/isaaclab_visualizers/isaaclab_visualizers/rerun/rerun_visualizer.py
```diff
@@ -392,6 +428,104 @@ def is_running(self) -> bool:
...
+
+    def _push_streaming_frame(self) -> None:
+        """Colorize and push the composited streaming frame to Rerun."""
+        from isaaclab.envs.utils.camera_colorizer import CameraFrameColorizer, sensor_key_for_gt_type
+        from isaaclab.envs.utils.camera_view import camera_gt_batch, compose_streaming_grid
+
+        if self._camera_sensor is None:
+            return
+        if self._camera_is_owned:
+            self._camera_sensor.update(dt=0.0, force_recompute=True)
```
kellyguo11, line 629: 🟡 **The auto-created camera targets the prim only once.** `_apply_streaming_camera_pose()` runs during setup, while each subsequent frame only updates the sensor. If `streaming_cam_target_prim_path` moves, the Rerun camera stops following it. Viser has the same flow. Please refresh the target-relative pose before the per-frame sensor update, or attach a tracked transform.

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
```diff
@@ -868,293 +1270,269 @@ def _filter_visible_env_tensor(self, tensor: torch.Tensor, num_envs: int) -> tor
...
+        # which means the image converges/clears while paused. This is correct
+        # path-tracer behavior: the longer you pause, the cleaner the image.
+        self._viewer.begin_frame(self._sim_time)
+        self._viewer.end_frame()
+
+    def _uses_streaming_view(self) -> bool:
+        # Streaming view uses a TiledCamera sensor independent of ViewerRTX, so it
+        # is fully supported even though render_rgb_array() (interactive frame capture)
+        # is still stubbed pending ViewerRTX.get_frame().
+        return bool(self.cfg.streaming_view)
```
kellyguo11: 🟡 **Newton RTX enables streaming work but has no visible sink.** `NewtonViewerRTX` has no `_image_logger`, so the inherited `log_image()` path is effectively a no-op: the camera is updated, colorized, composited, and transferred every frame without displaying a panel. This also conflicts with the documentation stating that the RTX streaming panel is unsupported. Please return `False`/reject this mode until a sink exists, or implement the missing output path.

docs/source/overview/core-concepts/visualization.rst
```diff
@@ -228,50 +237,82 @@ selection and visual offsets. This includes point-cloud and task-geometry marker
...
 
-If ``tiled_cam_view=True`` is set, another window is launched in the visualizer which shows
-a non-interactive tiled camera image view. Number of tiles is capped at 100.
 
-Kit tiled camera views work without an additional camera option.
+Streaming Camera View
```
kellyguo11, line 252: 🟡 **The missing RST target breaks the documentation build.** This section is referenced as `:ref:`streaming-camera-view`` from two pages, but no `.. _streaming-camera-view:` label is defined. Please add that target immediately above this heading.

kellyguo11 (commented):
Additional review of the streaming/OVRTX delta. The PR is currently based on an older #6598 head and GitHub reports conflicts. Please rebase before merge; the current stack also edits an immutable parent changelog fragment. Focused visualizer suites: 113 passed, 10 failed.

matthewtrepte:
@isaaclab-review-bot review

matthewtrepte:
@isaaclab-review-bot review

huidongc:
I am curious how does the model data flow through to the newton rtx visualizer if we use ovphysx backend? Does it go through this data pipe?

```
USD --> shadow newton model --> ovrtx internal representation (which is fabric)
```

matthewtrepte:
> I am curious how does the model data flow through to the newton rtx visualizer if we use ovphysx backend? Does it go through this data pipe?
> 
> ```
> USD --> shadow newton model --> ovrtx internal representation (which is fabric)
> ```

for Geometry, during initialization

Kit USD stage --> shadow Newton model --> Newton's own internal USD stage --> OVRTX private Fabric

then for Transforms, every frame

OVPhysX Kit Fabric --> shadow Newton state --> OVRTX private Fabric

matthewtrepte:
@isaaclab-review-bot review

kellyguo11 (changes requested):
I’m requesting changes based on a review of the current head `893ac30` against base `7c60d5f`. The PR was rebased after my initial pass, so I rechecked the affected paths on the new head.

1. **[P1] Resolve video’s runtime requirements before the launcher scan.**

   The RL entry points call `launch_simulation()` before `apply_video_recording()` (for example, `source/isaaclab_rl/isaaclab_rl/entrypoints/backends/train_rsl_rl.py:127-163`), while `apply_video_recording()` may inject a Kit visualizer afterward (`source/isaaclab_rl/isaaclab_rl/entrypoints/common.py:573-681`). In a kitless Newton launch, the launcher therefore never sees the Kit requirement. The import-based fallback does not address this: `KitVisualizerCfg` can be importable even though no Kit app was launched.

   This also breaks auto-switching with OvPhysX and OVRTX. With `--video` and no explicit Kit visualizer, the launcher scan sees no Kit runtime source, so auto-selection can choose OvPhysX plus OVRTX and `_validate_runtime()` passes. Injecting Kit later bypasses that compatibility check, which can then fail because no Kit app is running or because the native/Carbonite runtimes conflict. An explicit `--viz kit` avoids this because Kit is visible during the scan.

   Please determine the recorder source and inject any required visualizer before `launch_simulation()`. This should remain source-aware: sensor/Newton GL recording can be kitless, so `--video` should not unconditionally force Kit.

2. **[P1] Validate explicit visualizer capture capability before selecting a recorder source.**

   `apply_video_recording()` uses the first active CLI visualizer as the source without checking its capture API (`source/isaaclab_rl/isaaclab_rl/entrypoints/common.py:628-642`). `VideoRecorder` later calls `render_rgb_array()` (`source/isaaclab/isaaclab/envs/utils/video_recorder.py:154-205`). Rerun and Viser do not implement that method, so explicit `--viz rerun --video` / `--viz viser --video` reaches an uncaught `AttributeError`. Newton RTX implements it but returns `None` (`source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py:1618`), silently producing no usable clip.

   Please select only capture-capable sources, prefer tiled capture where that is the supported API, and reject unsupported combinations with a clear error instead of failing later or silently producing no recording.

3. **[P1] Do not share a streaming camera across consumers with incompatible sampling or poses.**

   The shared-camera key in `source/isaaclab/isaaclab/envs/utils/camera_view.py:create_visualizer_camera` covers renderer/stage/count/resolution/data types, but not the resolved `streaming_envs`, target prim path, or eye offset. Rerun and Viser commonly use the same renderer and resolution, so they can share one camera while independently sampling different environments. Only the owner applies its requested camera pose; a non-owner returns without applying its configuration (for example, `source/isaaclab_visualizers/isaaclab_visualizers/rerun/rerun_visualizer.py:511-611`). The second consumer can therefore read different environment indices from a camera posed using the owner’s selection and target.

   Please make sampling and pose part of a canonical shared-camera contract (or share one resolved sample), or reject incompatible consumers.

4. **[P1] Preserve a stable tiled frame while rendering is paused.**

   The pre-rebase visualizer integration job repeatedly failed the Kit/Newton tiled pause assertion, with roughly 826–933 changed pixels where at most 100 are allowed: https://github.com/isaac-sim/IsaacLab/actions/runs/30957733522/job/92158410468. The current head still unconditionally updates the Kit app and force-recomputes the streaming camera from `KitVisualizer.step()` / `_update_camera_image_panel()`, including during paused rendering (`source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py:190-221,694-744`). That is inconsistent with the stable-frame contract and is the most direct suspect. The post-rebase visualizer CI has not run yet.

   Please cache/reuse the paused composite, or otherwise enforce stable paused-frame semantics, and get these integration cases passing.

5. **[P2] Honor an explicit `--viz none` when video is requested.**

   Parsing `--viz none` records an explicit disable-all choice, but `apply_video_recording()` filters out `none` and injects Kit as though no visualizer was selected (`source/isaaclab_rl/isaaclab_rl/entrypoints/common.py:628-681`). By then, `launch_simulation()` has already persisted the disable-all decision, so `SimulationContext` discards the injected visualizer configuration. The requested Kit recorder source is absent and recording disables after a runtime error instead of honoring the user’s selection.

   Please preserve the explicit `none` intent: require a compatible sensor source or reject `--viz none --video` with a clear error.

6. **[P2] Keep the `visualizer:newton` recorder alias limited to Newton GL.**

   The alias in `source/isaaclab/isaaclab/envs/utils/video_recorder.py:154-163` matches both `newton_gl` and `newton_rtx`, while the public recorder documentation describes `visualizer:newton` as Newton GL and the RTX implementation returns `None` for RGB capture. If RTX is the first or only match, recording silently yields no frames even when GL may be available later.

   Please resolve `visualizer:newton` only to `newton_gl`, or implement and validate equivalent capture for RTX before including it.

7. **[P2] Either resample `streaming_envs` on reset or correct its public documentation.**

   `source/isaaclab/isaaclab/visualizers/visualizer_cfg.py:87-92` promises that an integer `streaming_envs` randomly samples environments “each reset,” but each backend calls `resolve_streaming_envs()` only from its one-time `_setup_streaming_view()` initialization path. No reset hook resamples or reconfigures it, so the actual behavior is one sample for the visualizer’s lifetime.

   Please resample/reconfigure on reset as documented, or change the documentation to state that selection occurs once at initialization.

Validation note: the current head’s pre-commit CI passes. The focused suite and diff checks passed on the pre-rebase head; the post-rebase visualizer CI is still pending. I intentionally did not attribute the rendering-correctness failures that matched the base branch.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -612,24 +612,68 @@ def apply_video_recording(env_cfg: Any, log_dir: str, args_cli: argparse.Namespa
...
+                # field hints to existing configs; it does not create a new visualizer).
+                if not isinstance(getattr(sim_cfg, "visualizer_cfgs", None), list):
+                    sim_cfg.visualizer_cfgs = []
                 try:
-                    from isaaclab_visualizers.newton import NewtonVisualizerCfg as _NwtCfg
-
-                    sim_cfg.default_visualizer_cfg = _NwtCfg(headless=True)
+                    from isaaclab_visualizers.kit import KitVisualizerCfg as _KitCfg
+
+                    sim_cfg.visualizer_cfgs.append(_KitCfg())
```
kellyguo11: Resolve the recorder's runtime requirements before launching the simulation. Every RL entry point enters `launch_simulation()` before this function runs, so the launcher has already scanned the config, resolved automatic backends, and validated runtime compatibility when this Kit visualizer is appended. With `--video` and automatic physics/RTX selection, that scan can select OvPhysX plus OVRTX and validate a kitless process; this late injection then bypasses the OvPhysX/OVRTX-versus-Kit guard and can also request Kit when no AppLauncher exists. Determine the recorder source and inject any required visualizer before `launch_simulation()`, while preserving kitless sensor and Newton GL recording.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -612,24 +612,68 @@ def apply_video_recording(env_cfg: Any, log_dir: str, args_cli: argparse.Namespa
...
+            cli_visualizers = [cli_visualizers]
+
+        # Filter out "none" entries — "none" disables all visualizers, so treat it the
+        # same as "no visualizer specified" for the purpose of picking a recorder source.
+        active_cli_visualizers = [v for v in cli_visualizers if v != "none"]
+
+        if active_cli_visualizers:
+            # User passed --viz <type>: record from whichever visualizer they requested.
+            # Use the first requested type as the source so the recorder is specific.
+            visualizer_source = f"visualizer:{active_cli_visualizers[0]}"
```
kellyguo11: Validate the visualizer's capture capability before selecting it as the recorder source. `--viz rerun --video` and `--viz viser --video` select visualizers that do not implement `render_rgb_array()`, causing an uncaught `AttributeError`; `newton_rtx` implements the method but returns `None`, producing no usable clip. Select or rank only capture-capable sources, use tiled capture where that is the supported API, and reject unsupported combinations with a clear error.

source/isaaclab/isaaclab/envs/utils/camera_view.py
```diff
@@ -114,15 +128,161 @@ def ensure_camera_initialized(camera: Camera) -> None:
...
+        RuntimeError: If the same renderer class is requested on the same stage
+            with a different camera configuration (e.g. different ``num_envs`` or
+            resolution), which would indicate a /Render prim singleton conflict.
+    """
+    stage = sim_utils.get_current_stage()
+    # Use identifier (unique per anonymous stage) rather than realPath (always "" for in-memory stages).
+    stage_id = stage.GetRootLayer().identifier if stage is not None else ""
+    renderer_class = type(renderer_cfg).__name__
+    dt_key = tuple(sorted(data_types or ["rgb"]))
+    full_key = (renderer_class, stage_id, int(num_envs), int(width), int(height), dt_key)
```
kellyguo11: Do not reuse a streaming camera based on this key when consumers can request different samples and poses. The key omits the resolved `streaming_envs`, target prim path, and eye offset. Rerun and Viser can therefore share one camera while independently selecting environments, and only the owner applies its requested pose; the other consumer reads from a camera configured for the owner's selection. Make sampling and pose part of a canonical shared-camera contract, share one resolved sample, or reject incompatible consumers.

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -656,29 +718,45 @@ def _update_owned_camera_poses(self) -> None:
...
             if self._generated_camera_poses_dirty:
                 self._sync_camera_pose_updates_to_kit()
                 self._generated_camera_poses_dirty = False
-        if self._camera_is_owned:
             self._camera_sensor.update(dt=dt, force_recompute=True)
-        rgb = camera_rgb_batch(self._camera_sensor, self._camera_sensor_indices)
-        image = compose_rgb_grid_tensor(rgb) if self.cfg.tiled_cam_view else rgb[0].contiguous()
-        self._upload_camera_image_to_panel(image)
+
+        gt_types = list(self.cfg.streaming_gt_types)
```
kellyguo11, line 798: Preserve a stable composite while rendering is paused. This refresh path follows an unconditional owner-side `camera.update(..., force_recompute=True)` on every Kit visualizer update. The tiled pause integration repeatedly observed hundreds of changed pixels while the expected limit was 100. Cache and reuse the paused composite, or otherwise avoid recomputing and uploading the streaming frame while paused.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -612,24 +612,68 @@ def apply_video_recording(env_cfg: Any, log_dir: str, args_cli: argparse.Namespa
...
+        # Check whether the user explicitly requested a visualizer via --viz.  If so,
+        # that visualizer will be created by the launcher and we should record from it
+        # without injecting a second one.
+        cli_visualizers: list[str] = getattr(args_cli, "visualizer", None) or []
+        if isinstance(cli_visualizers, str):
+            cli_visualizers = [cli_visualizers]
+
+        # Filter out "none" entries — "none" disables all visualizers, so treat it the
+        # same as "no visualizer specified" for the purpose of picking a recorder source.
+        active_cli_visualizers = [v for v in cli_visualizers if v != "none"]
```
kellyguo11, line 710: `none` is an explicit disable-all choice, not the absence of a visualizer selection. Filtering it here lets the later branch inject Kit after the launcher has already persisted `visualizer_disable_all`; `SimulationContext` then discards the injected visualizer and the recorder disables after failing to find its source. Honor the explicit `none` intent by requiring a compatible sensor source or rejecting `--viz none --video` clearly.

source/isaaclab/isaaclab/envs/utils/video_recorder.py
```diff
@@ -155,7 +155,14 @@ def _frame_from_visualizer(self, viz_type: str, sub: str) -> np.ndarray | None:
         visualizers = getattr(sim, "visualizers", [])
 
         if viz_type:
-            candidates = [v for v in visualizers if getattr(v.cfg, "visualizer_type", None) == viz_type]
+            # "newton" is a backward-compatible alias for the newton_gl / newton_rtx visualizer
+            # types.  The canonical visualizer_type on NewtonGLVisualizerCfg is "newton_gl", but
+            # source strings in tutorials and docs use the shorter "newton" form.
+            _newton_aliases = ("newton_gl", "newton_rtx")
```
kellyguo11: Keep the `visualizer:newton` recorder alias limited to `newton_gl`. The public recorder documentation describes this alias as Newton GL, while `NewtonRTXVisualizer.render_rgb_array()` currently returns `None`. If RTX is the first or only candidate, recording silently yields no frames even when a GL visualizer may be available later.

source/isaaclab/isaaclab/visualizers/visualizer_cfg.py
```diff
@@ -50,38 +50,63 @@ class VisualizerCfg:
...
     """
 
-    tiled_cam_eye: tuple[float, float, float] = (4.0, -4.0, 3.0)
-    """Offset of the camera eye from tiled_cam_target_prim_path for generated tiled cameras.
+    # Shared settings
+    streaming_envs: int | list[int] = 32
+    """Environments to capture.
 
-    The camera follows the target prim and always maintains this fixed offset relative to it.
+    * ``int`` — randomly sample this many envs each reset (from visible envs).
```
kellyguo11: The implementation does not resample at reset. Each backend calls `resolve_streaming_envs()` only from its one-time `_setup_streaming_view()` initialization path, and the reset hooks do not reconfigure the selection. Either resample on reset as documented or state that an integer is sampled once at visualizer initialization.

source/isaaclab_visualizers/isaaclab_visualizers/kit/kit_visualizer.py
```diff
@@ -656,29 +718,45 @@ def _update_owned_camera_poses(self) -> None:
...
+        available = frozenset(self._camera_sensor.data.output.keys())
+        frames = []
+        for env_idx in self._camera_sensor_indices:
+            for gt in gt_types:
+                key = sensor_key_for_gt_type(gt, available)
+                raw = camera_gt_batch(self._camera_sensor, [env_idx], key)[0]
```
kellyguo11: Preserve batched GPU processing here. This singleton fetch occurs inside the environment-by-GT loop, so every tile creates a GPU index tensor, performs a singleton `index_select`, and then `CameraFrameColorizer` calls `.cpu().numpy()`. The previous RGB path selected and composed one Torch batch on the GPU. With 32 environments this introduces 32 GPU-to-CPU synchronization points for RGB, or 128 for four GT types, followed by another CPU canvas copy. Fetch all selected environments once per GT type, colorize in batches, transfer at most once per GT type when required by the sink, and retain Kit's GPU-resident RGB upload fast path.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -612,24 +612,68 @@ def apply_video_recording(env_cfg: Any, log_dir: str, args_cli: argparse.Namespa
...
+
+                        sim_cfg.visualizer_cfgs.append(_NwtCfg(headless=True))
+                        visualizer_source = "visualizer:newton_gl"
+                        print(
+                            "[INFO] --video specified without --viz: auto-creating a headless Newton GL "
+                            "visualizer for video recording (Kit not available)."
+                        )
+                    except (ImportError, ModuleNotFoundError):
+                        pass
+        env_cfg.video_recorders = [VideoRecorderCfg(source=visualizer_source, video_interval=2000)]
```
kellyguo11, line 789: Avoid pumping Kit twice for every captured frame in this new internal recording path. A normal rendered environment step reaches `KitVisualizer.step()`, which calls `app.update()`. The recorder then runs after rendering and `KitVisualizer.render_rgb_array()` calls `omni.kit.app.get_app().update()` again. With the default `frame_stride=1`, each recorded step can execute two Kit/renderer updates. Cache the Replicator frame during the existing visualizer update, or use `SimulationContext.render_generation` so capture pumps Kit only when no render has occurred since the last state change.

kellyguo11 (commented):
Inline follow-up review covering the previously reported findings and the additional performance concerns.

huidongc:
> > I am curious how does the model data flow through to the newton rtx visualizer if we use ovphysx backend? Does it go through this data pipe?
> > ```
> > USD --> shadow newton model --> ovrtx internal representation (which is fabric)
> > ```
> 
> for Geometry, during initialization
> 
> Kit USD stage --> shadow Newton model --> Newton's own internal USD stage --> OVRTX private Fabric
> 
> then for Transforms, every frame
> 
> OVPhysX Kit Fabric --> shadow Newton state --> OVRTX private Fabric

We should flag this for future performance improvement. For ovphysx + newton_rtx the long trip for the data is unnecesarily bad. Maybe we can find a way to shorten the pipe.

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
```diff
@@ -14,46 +14,88 @@
...
 from isaaclab_visualizers.newton.newton_visualization_markers import render_newton_visualization_markers
 from isaaclab_visualizers.newton_adapter import resolve_visible_env_indices
 
-from .newton_visualizer_cfg import NewtonVisualizerCfg
+from .newton_visualizer_cfg import NewtonGLVisualizerCfg, NewtonRTXVisualizerCfg, NewtonVisualizerCfg
 
 logger = logging.getLogger(__name__)
 
 
+def _ovrtx_rtx_soname_fixed() -> bool:
```
rilei-nvidia: As discussed on the call, please take a look and see if this check is still needed in a fresh environment launching only the `newton_rtx` visualizer.

rilei-nvidia (changes requested):
Jumped on a call with @matthewtrepte to test & discuss the changes.

We tested the following visualizers: `newton_gl`, `viser`, `rerun` and `newton_rtx` with the Isaac-Lift-KukaAllegro-Camera task:
```
uv run --extra ovrtx --extra viser --extra rerun isaaclab train --rl_library rsl_rl --task Isaac-Lift-KukaAllegro-Camera presets=newton_mjwarp,ovrtx,rgb64,single_camera --seed 42 --num_envs 4 --max_iterations 1000 --visualizer <visualizer>
```

For `newton_gl`: 
1. Can you make it so that the Streaming View accordion panel is always present so the user can toggle open the view?
2. We discussed having a dropdown to configure the camera sensors in the streaming view window.  @matthewtrepte this would be in a follow-up PR?

For `viser` and `rerun` we observed that the robot arm is not present in the streaming view.  @matthewtrepte to investigate and address in a follow-up PR

For `newton_rtx`
1. Please assess if the `_ovrtx_rtx_soname_fixed()` check is still necessary.  We should also hard-fail instead of only printing a warning if there is a real conflict.
2. @matthewtrepte to file a ticket for the Newton team to allow support for using the lighting directly from the USD scene instead of "studio", "default", "none".  I would advocate that using the lights from the USD scene should be the "default" behavior rather than a stand-in dome & directional light

Thanks for this work @matthewtrepte!  Happy to test it again tomorrow

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
```diff
@@ -868,293 +1368,265 @@ def _filter_visible_env_tensor(self, tensor: torch.Tensor, num_envs: int) -> tor
...
+            )
+            return None
 
-    def _live_plots_panel_imgui(self, imgui) -> None:
-        """Render a Live Plots collapsing section at the bottom of the Newton panel.
+        return NewtonViewerRTX(
+            width=self.cfg.window_width,
+            height=self.cfg.window_height,
+            headless=runtime_headless,
+            up_axis="Z",
```
rilei-nvidia, line 1895: I think we should respect the up axis as defined by the USD scene rather than hardcoding to Z?

rilei-nvidia:
@matthewtrepte additionally - I noticed that after launching the newton_rtx visualizer, it creates an `imgui.ini` file in the current working directory.  Is that needed?

source/isaaclab_visualizers/isaaclab_visualizers/viser/viser_visualizer.py
```diff
@@ -396,20 +469,173 @@ def step(self, dt: float) -> None:
 
         if not has_clients:
             self._render_live_plots()  # still throttled internally; no-ops when no clients
+            # Still push streaming even with no clients so render_tiled_rgb_array() has data
+            # for VideoRecorder capture (headless recording without a browser connection).
+            self._push_streaming_frame()
```
kellyguo11: Make streaming-frame production demand-driven instead of rendering on every visualizer update. With no Viser clients, this still force-updates the camera, transfers and composites its tiles, and attempts to publish a frame that nobody sees; Rerun performs the same work outside its rendering-pause gate. This can materially reduce headless training throughput. The VideoRecorder justification does not require speculative work on every step because recording is intermittent through video_interval/frame_stride. Let render_tiled_rgb_array() lazily build or reuse the current-step composite, return here when no client is connected, and apply the equivalent change to Rerun's paused path. When the CLI selects the Rerun/Viser tiled recording source, it should also explicitly enable streaming_view so the requested recording can actually produce frames.

source/isaaclab/isaaclab/envs/utils/camera_view.py
```diff
@@ -114,15 +128,181 @@ def ensure_camera_initialized(camera: Camera) -> None:
...
+            with a different camera configuration (e.g. different ``num_envs`` or
+            resolution), which would indicate a /Render prim singleton conflict.
+    """
+    stage = sim_utils.get_current_stage()
+    # Use identifier (unique per anonymous stage) rather than realPath (always "" for in-memory stages).
+    stage_id = stage.GetRootLayer().identifier if stage is not None else ""
+    renderer_class = type(renderer_cfg).__name__
+    dt_key = tuple(sorted(data_types or ["rgb"]))
+    eye_key = tuple(float(x) for x in eye) if eye is not None else None
+    full_key = (renderer_class, stage_id, int(num_envs), int(width), int(height), dt_key, target_prim_path, eye_key)
```
kellyguo11: Include the resolved environment sample in the shared-camera contract and manage shared lifetime explicitly. A second visualizer can currently request different streaming_envs but receive this same Camera with is_owner=False, so it reads cameras that only the first visualizer positions; users running Rerun and Viser together can therefore see stale, empty, or wrong environment views. Although this key has target_prim_path and eye slots, the Newton, Rerun, and Viser call sites do not pass those arguments, so they are currently None as well. Pass the resolved env indices, target path, and eye into a canonical key, reject incompatible singleton consumers, and use a reference-counted lease with ownership transfer so closing the first visualizer does not remove camera prims still used by another.

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
```diff
@@ -868,293 +1348,457 @@ def _filter_visible_env_tensor(self, tensor: torch.Tensor, num_envs: int) -> tor
...
-        # Promote episode metrics (mean_reward, episode_length) to the top.
-        episode_keys = [k for k in groups if k.startswith("episode/")]
-        other_keys = [k for k in groups if not k.startswith("episode/")]
-        groups = {k: groups[k] for k in episode_keys + other_keys}
-
-        imgui.set_next_item_open(False, imgui.Cond_.appearing)
-        if not imgui.collapsing_header("Live Plots"):
+        # When the panel is hidden (selected=None), skip all camera rendering to keep
+        # per-step overhead zero.  Work resumes the next step after the user opens it.
+        if image_logger is not None and image_logger._selected is None:
```
kellyguo11: Keep the hidden-panel optimization without disabling explicit tiled capture. This return prevents _last_streaming_composite from ever being populated while the panel is closed; in headless mode the panel remains unselected permanently. A VideoRecorder using source="visualizer:newton_gl:tiled" therefore receives None on every capture and silently produces no clip, while an interactive recording starts only after the user opens the panel. Move frame production into a shared on-demand helper: call it here only when the panel is visible, and call it from render_tiled_rgb_array() when a recorder explicitly requests a frame. Cache by visualizer step so a visible panel and recorder reuse the same composite.

rilei-nvidia (approved):
Tested newton_gl and newton_rtx again.  

It was great to see the streaming view available to be toggled on + camera dropdown for newton_gl.  

LGTM

kellyguo11 (approved):

PR #6660, opened 2026-07-22
Update OVPhysX to 0.5.9

AntoineRichard:
## Summary

- pin OVPhysX exactly to `ovphysx==0.5.9` and inherit the bounded OVStage dependency from current `develop`
- rename the unreleased root uv extras from `ov` / `rtx` to `ovphysx` / `ovrtx`, including installer dispatch and direct-uv documentation
- inherit the Warp 1.16 and Newton pins from current `develop`
- pin `omniverseclient==2.72.3` and route public OV packages through public PyPI
- migrate OVPhysX initial USD ingestion to OVStage, including in-memory stage attachment
- keep OVPhysX and OVRTX coverage in separate ARM processes and exercise both resolved packages in x86 CI

## Testing

- [x] `./isaaclab.sh -f`
- [x] `uv lock --check` (446 packages)
- [x] `./isaaclab.sh -p -m pytest source/isaaclab/test/cli/test_uv_run_pyproject.py source/isaaclab/test/cli/test_install_command_parsing.py source/isaaclab_ovphysx/test/physics/test_ovphysx_scene_data_backend.py -q` (80 passed)
- [x] `./isaaclab.sh -p -m pytest "source/isaaclab_tasks/test/core/test_rendering_cartpole_kitless.py::test_rendering_cartpole_kitless[ovstage-ovphysx-ovrtx-rgb]" -q` (1 passed)


pyproject.toml
```diff
@@ -137,7 +137,7 @@ rerun = [
 # Isaac Sim (PhysX backend); co-resolves with the base install via the [tool.uv] conflicts table.
 isaacsim = ["isaacsim[all,extscache]==6.0.0.1"]
 # Omniverse renderer / physics backends (``ov[ovphysx|ovrtx|all]`` selectors).
-ov = ["ovphysx>=0.4.13"]
+ov = ["ovphysx>=0.5,<0.6"]
```
marcodiiga: This PR specifically targets the public OVPhysX 0.5.9 ingestion contract. The range can resolve other 0.5.x builds, and IsaacLab imports `ovstage` directly, so both should be explicit.

```suggestion
ov = ["ovphysx==0.5.9", "ovstage"]
```

pyproject.toml
```diff
@@ -178,11 +178,11 @@ isaacsim = "6.0.0.1"
 torch = "2.11.0"
 torchvision = "0.26.0"
 torchaudio = "2.11.0"
-ovphysx = ">=0.4.13"
+ovphysx = ">=0.5,<0.6"
```
marcodiiga: Please keep the CI source of truth exact too; `resolve-ov-pins` will then emit `ovphysx==0.5.9` rather than a floating range.

```suggestion
ovphysx = "0.5.9"
```

uv.lock
```diff
@@ -5381,15 +5381,16 @@ wheels = [
 
 [[package]]
 name = "ovphysx"
-version = "0.4.13"
+version = "0.5.9"
```
marcodiiga, line 5462: This lock still resolves `ovphysx`, `ovstage`, and `omniverseclient` from `pypi.nvidia.com`, so it does not validate the public PyPI.org path this PR is meant to exercise. Please route those three packages through a named uv index for `https://pypi.org/simple` with `explicit = true` in `pyproject.toml`, then regenerate the lock rather than editing it manually.

.github/workflows/build.yaml
```diff
@@ -312,13 +312,7 @@ jobs:
...
         container-name: isaac-lab-arm-ci-${{ github.run_id }}-${{ github.run_attempt }}
         image-tag: ${{ needs.config.outputs.ci_image_tag }}-arm64
-        # ovphysx stays pinned to 0.4.13 here (see the aarch64 note below).
-        extra-pip-packages: "${{ steps.ov_pins.outputs.ovrtx }} ovphysx==0.4.13"
-        # ovphysx-backed test params require ovphysx >= 0.5.1, which has no aarch64
-        # wheel yet; keep the newton/ovrtx rendering coverage, mirroring the public
-        # pip-index fallback of rendering-correctness-kitless. Running them anyway
-        # exhausts ovrtx SyncScopeIds (>15 renderer creations in one process).
-        test-k-expr: not ovphysx
+        extra-pip-packages: "${{ steps.ov_pins.outputs.ovrtx }} ${{ steps.ov_pins.outputs.ovphysx }}"
```
marcodiiga: This puts every ARM renderer/physics parameter in one pytest process, which can exhaust OVRTX SyncScopeIds. Keep the existing invocation non-OVPhysX and add a second invocation with a distinct report/container for the OVPhysX parameters.

```suggestion
        extra-pip-packages: "${{ steps.ov_pins.outputs.ovrtx }} ${{ steps.ov_pins.outputs.ovphysx }}"
        test-k-expr: not ovphysx
```

source/isaaclab/test/cli/test_uv_run_pyproject.py
```diff
@@ -103,6 +107,7 @@ def spec(package: str) -> str:
     # either by carrying the literal range, or by referencing the ``resolve-ov-pins``
     # action output, which reads the pin from this same table. Never a bare ``ovrtx``.
     build_workflow = (_repo_root() / ".github/workflows/build.yaml").read_text(encoding="utf-8")
+    assert "ovphysx==0.4.13" not in build_workflow
```
marcodiiga, line 113: This only catches the stale 0.4.13 literal. The fork/public paths in `test-isaaclab-ov` and `rendering-correctness-kitless` still exclude `isaaclab_ovphysx` or filter `not ovphysx`, so x86 fork CI runs no OVPhysX backend tests. Please remove only those public-path exclusions and preserve the container-wheelhouse path.

source/isaaclab_ovphysx/changelog.d/antoiner-bump-ovphysx-warp.rst
```diff
@@ -0,0 +1,5 @@
+Changed
+^^^^^^^
+
+* Changed the OVPhysX dependency to ``ovphysx>=0.5,<0.6`` and migrated USD
+  loading to the in-memory OVStage API introduced in OVPhysX 0.5.
```
marcodiiga: The API belongs to OVStage, and only initial USD ingestion changes; TensorBindings remains the runtime path afterward.

```suggestion
* Changed the OVPhysX dependency to ``ovphysx==0.5.9`` and migrated initial
  USD ingestion to OVStage.
```

marcodiiga (commented):
The ingestion implementation itself looks right, but I would not merge this yet. In a clean environment using the exact wheels published on PyPI.org (`ovphysx==0.5.9`, `ovstage==0.1.0.346039`, `omniverseclient==2.72.3`, `warp-lang==1.15.0`), IsaacLab's existing `Tests/spatial_tendons.usd` asset populates successfully and then segfaults in `PhysX.attach_ovstage()`. The same test passes on the pre-0.5 base checkout. `ovstage.population.open_usd()` succeeds for both the original asset and a minimized scene; the subsequent attach call crashes, so this reproduces without this PR's `Flatten()` / `open_usd_from_string()` serialization. The minimized trigger is a spatial tendon whose root and intermediate attachment are on the same rigid link, followed by a leaf attachment; removing the tendon schemas makes attach succeed. This localizes the regression to OVPhysX 0.5.9's OVStage ingestion/attach path, not this PR. In the same public-wheel environment, Cartpole, rigid-body loading, cloning, stepping, and TensorBindings smoke tests pass. The existing spatial-tendon regression nevertheless blocks the current public 0.5.9 wheel.

The inline suggestions below are the IsaacLab-side improvements I would carry over while the upstream wheel is fixed.

source/isaaclab/test/cli/test_uv_run_pyproject.py
AntoineRichard: Applied in 7e3a722edb: removed only the public-path OVPhysX exclusions from test-isaaclab-ov and rendering-correctness-kitless, and changed public installs to the exact resolve-ov-pins output. The wheelhouse resource/package branches remain unchanged. Added workflow regression coverage.

pyproject.toml
AntoineRichard: Applied in 7e3a722edb: the ov extra now declares ovphysx==0.5.9 and ovstage explicitly. The metadata regression test enforces both requirements.

source/isaaclab_ovphysx/changelog.d/antoiner-bump-ovphysx-warp.rst
AntoineRichard: Applied the suggested wording exactly in 7e3a722edb.

.github/workflows/build.yaml
AntoineRichard: Applied in 7e3a722edb: the existing ARM invocation keeps both packages installed but filters not ovphysx; a second invocation runs only ovphysx parameters with its own report and container. Added a workflow regression assertion for the split.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -457,92 +476,40 @@ def get_scene_data_backend(cls) -> SceneDataBackend:
...
-            layer.Export(target_file)
-            logger.info(
-                "OvPhysxManager: stripped %d env_<i!=0> subtrees from exported USD (kept env_0 + globals)",
-                len(names_to_remove),
-            )
+        for child_name in list(envs_spec.nameChildren.keys()):
+            match = env_name_re.match(child_name)
+            if match and match.group(1) != "0":
+                del envs_spec.nameChildren[child_name]
+        return layer.ExportToString()
```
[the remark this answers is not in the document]
AntoineRichard: Restored both diagnostics in 7e3a722edb: a debug message when /World/envs is absent and an info message with the exact stripped subtree count. Added caplog coverage for both paths.

uv.lock
AntoineRichard: Applied in 7e3a722edb: added an explicit pypi-public index for https://pypi.org/simple, routed omniverseclient, ovphysx, and ovstage through it, and regenerated uv.lock. All three lock entries now resolve from pypi.org/files.pythonhosted.org.

pyproject.toml
AntoineRichard: Applied in 7e3a722edb: tool.isaaclab.versions now uses the exact value 0.5.9, so resolve-ov-pins emits ovphysx==0.5.9. The source-of-truth test enforces the exact value.

AntoineRichard:
Review feedback is addressed in 7e3a722edb:

- exact ovphysx==0.5.9 plus explicit ovstage
- public-PyPI source routing and regenerated lock for ovphysx, ovstage, and omniverseclient
- isolated ARM OVPhysX pytest process and restored public x86 OVPhysX coverage
- restored in-memory ingestion diagnostics and precise changelog wording

Verification on the committed tree:
- 26 focused metadata/manager tests passed
- all pre-commit hooks passed
- uv lock --check resolved 442 packages

Known upstream blocker — do not merge yet:

./isaaclab.sh -p -m pytest "source/isaaclab_ovphysx/test/assets/test_articulation.py::test_spatial_tendons[cpu-1]" -q -s

With omniverseclient 2.72.3, ovphysx 0.5.9, ovstage 0.1.0.346039, and warp-lang 1.15.0, this exits via SIGSEGV (-11) in ovphysx.api.PhysX.attach_ovstage. The traceback confirms the failure occurs at the public wheel attach boundary reported in review.

pyproject.toml
```diff
@@ -137,7 +137,7 @@ rerun = [
 # Isaac Sim (PhysX backend); co-resolves with the base install via the [tool.uv] conflicts table.
 isaacsim = ["isaacsim[all,extscache]==6.0.0.1"]
 # Omniverse renderer / physics backends (``ov[ovphysx|ovrtx|all]`` selectors).
-ov = ["ovphysx>=0.4.13"]
+ov = ["ovphysx==0.5.9", "ovstage"]
```
StafaH: Let's do ovphysx here, and below just change it to ovrtx

pyproject.toml
```diff
@@ -57,14 +57,14 @@ dependencies = [
     "pytetwild[all]>=0.3.0,<0.4 ; platform_machine == 'aarch64' or platform_machine == 'arm64'",
     "transformers==4.57.6",
     "einops",
-    "warp-lang==1.15.0.dev20260626",
+    "warp-lang==1.15.0",
```
StafaH: >=1.15.0 (newton is already doing some pins with 1.16). There is another PR already doing this though so let's see which one goes in first

source/isaaclab/test/cli/test_uv_run_pyproject.py
```diff
@@ -120,6 +126,49 @@ def spec(package: str) -> str:
     assert f"warp-lang=={versions['warp']}" in dependencies
 
 
+def test_public_ov_packages_use_public_pypi_index():
```
StafaH, line 135: This is a weird test to have

source/isaaclab/test/cli/test_uv_run_pyproject.py
```diff
@@ -120,6 +126,49 @@ def spec(package: str) -> str:
...
+    assert indexes["pypi-public"] == {
+        "name": "pypi-public",
+        "url": "https://pypi.org/simple",
+        "explicit": True,
+    }
+    for package in ("omniverseclient", "ovphysx", "ovstage"):
+        assert sources[package] == {"index": "pypi-public"}
+
+
+def test_ci_runs_public_ovphysx_coverage_in_isolated_processes():
```
StafaH: This is also a weird test

StafaH (approved):

marcodiiga (approved):

pyproject.toml
AntoineRichard: Kept the existing `ov` / `rtx` extra names in this PR because they are documented user-facing `uv --extra` selectors; replacing them would be a breaking rename without a deprecation path. A follow-up can add `ovphysx` / `ovrtx` aliases while preserving compatibility.

pyproject.toml
AntoineRichard: Rebased onto current `develop` and removed this PR’s Warp 1.15 override, assertion, and changelog entry. The branch now inherits `warp-lang==1.16.0.dev20260723` and the matching Newton pin from `develop`.

source/isaaclab/test/cli/test_uv_run_pyproject.py
AntoineRichard: Kept this focused metadata regression because public-PyPI routing for OVPhysX, OVStage, and OmniClient is an explicit contract of this PR and previously regressed to `pypi.nvidia.com`. It parses TOML rather than workflow text and lives in the CLI metadata test location.

source/isaaclab/test/cli/test_uv_run_pyproject.py
AntoineRichard: Removed during the `develop` rebase. The YAML string-splitting assertions were brittle; the actual ARM process split and public x86 coverage changes remain in the workflows and pass YAML/pre-commit validation.

pyproject.toml
AntoineRichard: I'll inherit from newton's 1.16

pyproject.toml
AntoineRichard: Implemented in `530f4419b0`. The root uv extras are now `ovphysx` and `ovrtx`; `ov[ovphysx]` / `ov[ovrtx]` dispatch to the matching names, the direct-uv docs use the new extras, and the unreleased `ov` / `rtx` root names were removed.

PR #6665, opened 2026-07-22
Add OVPhysX deformable object support

AntoineRichard:
# Description

Adds PhysX-style volume and surface deformable-object support to the OVPhysX backend while preserving OVPhysX native device placement and full-buffer indexed write semantics.

- derives Warp allocation and validation dtypes from each OVPhysX binding DLPack metadata
- keeps `OvPhysxDeformableBodyView` thin, adding only nodal and element count properties over the regular `OvPhysxView`
- accesses deformable body and material tensors directly through the generic view with `TensorType` keys
- mirrors PhysX volume/surface discovery, optional-material behavior, data access, and contract tests while retaining additive OVPhysX topology, cache, and full-stage coverage
- supports nodal state, volume kinematic targets, mesh connectivity, optional materials, and derived root state
- preserves heterogeneous rigid/deformable clones across full-stage loads, re-warmups, ancestor shells, and nested clone targets
- adds authored CUDA coverage for volume, surface, multi-environment replication, mixed rigid/deformable scenes, and the Franka soft-lift task
- documents the public API and adds the `isaaclab_ovphysx` changelog fragment

Dependency status: #6660 has landed on develop. It owns the OVPhysX dependency floor and in-memory OVStage serializer. This PR intentionally does not modify dependency pins or lock files.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Validation

- 141 focused OVPhysX deformable, view, manager, replication, and task tests passed
- Python byte-compilation passed
- documentation build succeeded without warnings
- all repository pre-commit hooks passed twice at the final branch tip
- independent whole-branch review findings were resolved

## Screenshots

Not applicable.

## Checklist

- [x] I have added tests that prove the feature works
- [x] I have added a changelog fragment under `source/isaaclab_ovphysx/changelog.d/`


source/isaaclab_ovphysx/isaaclab_ovphysx/assets/deformable_object/deformable_object.py
```diff
@@ -0,0 +1,588 @@
...
+                f"OVPhysX deformable tensors require a CUDA simulation device; received {self._device!r}."
+            )
+
+        def has_deformable_body_api(prim) -> bool:
+            return "OmniPhysicsDeformableBodyAPI" in prim.GetPrimTypeInfo().GetAppliedAPISchemas()
+
+        asset_prim, root_expr = sim_utils.resolve_matching_prims_from_source(self.cfg.prim_path)[0]
+        walk_root = asset_prim.GetPath().pathString
+        resolve_kwargs = {"predicate": has_deformable_body_api, "expected_num_matches": 1}
+        root_prim, root_path_expr = sim_utils.resolve_matching_prims_from_source(self.cfg.prim_path, **resolve_kwargs)[
```
mmichelis: odd way of splitting the line

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/deformable_object/deformable_object.py
```diff
@@ -0,0 +1,582 @@
...
+            nodal_pos = wp.from_torch(nodal_pos.contiguous(), dtype=wp.vec3f)
+        if (
+            full_data
+            and env_ids.shape[0] < self.num_instances
+            and self._data._nodal_pos_w.timestamp < self._data._sim_timestamp
+            and nodal_pos.ptr == self._data._nodal_pos_w.data.ptr
+        ):
+            nodal_pos = wp.clone(nodal_pos)
+        if env_ids.shape[0] < self.num_instances:
+            _ = self._data.nodal_pos_w
```
marcodiiga, line 223: [P1] Preserve aliased partial edits before refreshing the cache.

When a caller retains `data.nodal_pos_w`, advances the simulation, edits `retained.torch[1:2]`, and then calls the default `write_nodal_pos_to_sim_index(..., env_ids=[1])`, the input slice aliases the stale stable cache. This line refreshes that allocation before the write kernel and overwrites the command. With `ovphysx==0.5.9` on GPU, I reproduced both the slice and the simulator readback reverting to the old simulator values. The velocity path has the same issue at line 270.

The current clone guard covers only `full_data=True` and exact base pointers. Please snapshot any overlapping source range before hydration and add retained-slice regressions for both position and velocity.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/deformable_object/deformable_object_data.py
```diff
@@ -0,0 +1,166 @@
...
+    def root_pos_w(self) -> ProxyArray:
+        """Mean simulation-node position in simulation world frame [m].
+
+        Shape is ``(num_instances,)``, dtype ``wp.vec3f``.
+        """
+        if self._root_pos_w.timestamp < self._sim_timestamp:
+            wp.launch(
+                compute_mean_vec3f_over_vertices,
+                dim=self._num_instances,
+                inputs=[self.nodal_pos_w.warp, self._max_sim_vertices],
```
marcodiiga, line 140: [P1] Account for each body's valid node count here, or reject nonuniform topology.

This binding dimension is the maximum node count across all matches, not the valid count for every body. In a real `ovphysx==0.5.9` binding with a five-node and a four-node volume, the shape is `(2, 5, 3)` and the native read leaves the short body's fifth slot untouched. This kernel therefore averages padding: with the normal zero-initialized buffer, the valid mean `[2.25, 0.25, 0.25]` becomes `[1.8, 0.2, 0.2]`; after a full-buffer write, arbitrary stale padding can persist across refreshes.

The same issue affects `root_vel_w` and padded `nodal_state_w`, and the soft-lift task consumes the derived COM in observations, rewards, and termination. Please carry a per-row valid-node count or mask into these reductions, or fail initialization for nonuniform matched topology until the runtime exposes that metadata.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -565,16 +708,16 @@ def _warmup_and_load(cls) -> None:
...
-        # the full N-env stage; only the wheel-side physics ingestion is
-        # scoped to env_0, and ``physx.clone()`` re-populates env_1..N in
-        # the physics runtime with proper clone lineage (which is what the
-        # binding fast path expects).
-        stage_usda = cls._serialize_env0_only_stage(sim.stage)
+        # By default the serialized stage keeps only env_0 and runtime cloning
+        # re-populates env_1..N. Features that need distinct authored physics in
+        # every environment request the full stage. Missing heterogeneous clone
+        # targets are copied into the serialized stage before it is attached.
+        cls._rearm_pending_clones()
```
marcodiiga, line 720: [P2] Invalidate asset bindings before replacing the attached stage on a forced re-warmup.

This reuse path resets and destroys the attached stage, but it never emits the
lifecycle event that clears initialized asset views. I reproduced the documented
forced path (`_warmup_done = False; sim.reset()`) with a real deformable: the view
and TensorBinding identities were unchanged, and the next read failed because
the binding's stage ID referred to the prior attachment. `PHYSICS_READY` does not
rebuild it because the asset remains initialized.

Please dispatch the stop/invalidation event before detaching (or explicitly
recreate dependent views), then add a real-binding regression that verifies a
post-rewarm read/step and changed binding identity.

marcodiiga (changes requested):
Reviewed exact head `b6b479b5d16834547d77b922d2b6419d3ae62534` after its rebase onto `develop`.

I am requesting changes for the two reproduced silent-correctness failures and
the stale-binding lifecycle failure in the inline threads:

- an aliased selected position/velocity command can be overwritten by stale-cache hydration before it reaches OVPhysX;
- mixed deformable topologies average invalid padded nodes into `root_pos_w` and `root_vel_w`;
- a forced manager re-warm replaces the attached stage without invalidating existing asset bindings.

Validation was otherwise encouraging: the live GPU job reports 18/18 deformable asset tests and the task smoke passing, and focused view/manager/data suites passed locally. The in-memory OVStage ingestion is directionally consistent with the current backend architecture.

Two additional merge gates remain separate from those findings:

- Current CI is red in Docs because the daily `warp-lang==1.17.0.dev20260728` Linux wheel is unavailable, and in `isaaclab_ov` on an unchanged joint-wrench tolerance assertion. Other required jobs are still running.
- The support matrix still says OvPhysX deformables are unsupported, and the OvPhysX backend page does not document the new support or its current full-stage scaling limit. In a minimal five-node volume scene I measured reset at 1.59 s for 128 environments, 10.54 s for 512, 36.66 s for 1024, and no completion within 120 s for 4096. Please either bound the supported scale or address this before advertising normal thousand-environment use.

AntoineRichard:
Implemented the submitted review feedback in `a4c0f89ce6`. I am replying here because GitHub will not create inline replies while this account has an existing pending review, and I am leaving that draft untouched.

- [Aliased partial writes](https://github.com/isaac-sim/IsaacLab/pull/6665#discussion_r3664749448): position and velocity writes now snapshot any source storage range that overlaps a stale stable cache before hydrating unselected rows. This covers both full-buffer aliases and retained slices. Added CPU red/green regressions and real OVPhysX GPU readback coverage for both paths.
- [Mixed topology](https://github.com/isaac-sim/IsaacLab/pull/6665#discussion_r3664749454): took the conservative rejection option rather than adding padded-state masks in this PR. The thin deformable view derives each body's node count from simulation connectivity, reduces it on the binding's native device, and rejects counts below the binding maximum during initialization. Uniform volume and surface views remain unchanged.
- [Forced re-warm](https://github.com/isaac-sim/IsaacLab/pull/6665#discussion_r3664749457): a forced re-warm now dispatches `STOP` before replacing an attached stage. The real-binding regression verifies that the view and binding identities change and that post-rewarm reads and stepping succeed.
- [Line split](https://github.com/isaac-sim/IsaacLab/pull/6665#discussion_r3659008408): fixed earlier in `b6b479b5d1` by assigning the resolved prim list before unpacking the first match; Ruff preserves the resulting split.

The support matrix and OvPhysX backend page now mark deformables as experimental and CUDA-only, require uniform simulation-node counts, and explicitly scope the current full-stage path to small validation scenes; training-scale workloads with thousands of environments are unsupported. I intentionally kept full-stage ingestion optimization outside this PR.

Fresh validation on the PR head:

- 52 CPU-safe deformable view/helper and manager tests passed.
- 21 real OVPhysX GPU deformable tests passed.
- Full pre-commit passed.
- Sphinx documentation build passed.

Fresh CI is running for the new head. I did not add unrelated changes for the previously reported daily Warp-wheel availability or unchanged `isaaclab_ov` joint-wrench tolerance failures.

marcodiiga (approved):

mmichelis (approved):
thanks for adding deformable support in ovphysx!

ooctipus (approved):

PR #6671, opened 2026-07-22
Emit error annotations for failing tests in CI log view

mataylor-nvidia:
## Summary

- Adds `::error::` annotations to the "Write job summary" step in `.github/actions/run-tests/action.yml`
- Parses the JUnit XML report (already produced by every test run) and emits one annotation per failing/erroring test case
- No new dependencies; uses stdlib `xml.etree.ElementTree`

## Motivation

Currently, when a test fails, users must scroll through thousands of lines of Docker log output to find which test failed. The `$GITHUB_STEP_SUMMARY` markdown table already exists but requires navigating to the Summary tab.

With this change, `::error::` annotations surface failures in three places:
1. **Log view** — red highlighted lines in the "Write job summary" step, immediately visible
2. **PR Checks tab** — annotations appear inline on the PR
3. **Notification emails** — failures named explicitly

## Test plan

- [ ] Verify a failing CI run shows `::error::` annotations in the "Write job summary" step log
- [ ] Verify passing runs emit no annotations
- [ ] Verify corrupt/missing XML does not break the step (falls through to existing summary logic)


AntoineRichard:

Verified still relevant: current `.github/actions/run-tests/action.yml` has no `::error::` annotations (#6462 only added log folding), and although this branch predates the #6591 restructure, the anchor text it patches is unchanged on develop, so it should apply cleanly. The implementation is solid — stdlib XML parse wrapped in try/except (corrupt reports emit `::warning::` instead of breaking the step), 200-char truncation, and `%`/CR escaping for workflow-command syntax.

Asks before merge:
- Rebase to get a current diff, and trigger at least one real CI run that exercises the annotation path (only the labeler has ever run here, so the behavior on an actual failing test has never been observed live).
- Optional: the parse duplicates the one in `junit_summary.py` — harmless in CI cost, but worth a comment or consolidation later.

(The asset/isaac-sim/isaac-mimic labels are auto-labeler noise — the PR only touches the one action file.)

PR #6681, opened 2026-07-23
Fix ISAAC_LAB_SAVE_STAGES writing bare Nucleus-relative asset paths

mataylor-nvidia:
maybe_save_stage was reading the raw USDA text from the temp file and copying it verbatim to the output directory. Because resolve_paths silently no-ops when the live stage root layer is anonymous/in-memory (common in Kit), asset references like Assets/Isaac/6.0/... were preserved as bare Nucleus content-root paths rather than being resolved relative to the output file.

Use Usd.Stage.Flatten() + flat_layer.Export() for the ISAAC_LAB_SAVE_STAGES write path, matching what the compare_golden branch already does. This also removes the duplicate Stage.Open/Flatten in compare_golden by hoisting both operations before the two branches.

# Description

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


StafaH (approved):

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -316,34 +315,31 @@ def maybe_save_stage(
     try:
         if not sim_utils.save_stage(stage_path, save_and_reload_in_place=False):
             pytest.fail(f"save_stage reported failure while writing the USD stage to {stage_path}.")
-        with open(stage_path, encoding="utf-8") as file:
-            stage_text = file.read()
+
+        from pxr import Usd  # noqa: PLC0415
+
+        # Open and flatten the saved stage once. Flattening inlines sublayer references and
```
StafaH: This comment seems a bit too verbose

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -316,34 +315,31 @@ def maybe_save_stage(
     try:
```
StafaH, line 315: Maybe we should add a targeted test that can catch this behaviour we are trying to avoid

StafaH (approved):

source/isaaclab_tasks/test/rendering_test_utils.py
mataylor-nvidia: added tests

nvsekkin (approved):

nvsekkin:
this change is straightforward. one non-blocking question: `sim_utils.save_stage()` stays the same, that's why I was thinking maybe other `save_stage()` callers have the same issue? if the scope is testing only or this is not a concern, we are good to merge

PR #6688, opened 2026-07-23
Add Newton cable object support

mmichelis:
## Description

Adds first-class cable authoring and runtime support for Newton.

- Adds `CableObjectCfg`, the `CableCfg` spawner, and `CableMaterialCfg` for authoring open, linear, nonperiodic `UsdGeom.BasisCurves` cables, with validated thickness, density, and the four stiffness moduli (finite/positive geometry, finite/nonnegative stiffness).
- Adds a backend-dispatched `CableObject` and `CableObjectData` API exposing per-segment world pose and velocity, with indexed and masked pose/velocity writes, default-state capture for restoration, interactive-scene integration, and model-rebuild rebinding on `PHYSICS_READY`.
- Implements the Newton VBD backend for cables. Cables run under a standalone `NewtonVBDManager` and under `CouplerProxyCfg` when a named VBD entry owns the cable segments. Other backends are not registered, so selecting PhysX or OpenUSD PhysX raises the generic factory import error; PhysX non-support is documented in the supported-features tables.
- Adds opt-in cable collision (`collision_props`) with adjacent-segment-only collision filtering, so connected segments are filtered while the cable still collides with the ground and with other cables, matching Newton's cable-pile behavior.
- Adds a runnable Newton VBD cable-pile demo (`scripts/demos/cables.py`) with configurable cable and segment counts and periodic state resets, plus a showroom gallery entry.
- Supports imported and replicated cables, colors VBD-integrated cable bodies at simulation start, and skips particle-BVH rebuilds in cable-only (particle-less) scenes.
- Allows physics materials to bind to deformable curves (`PhysicsCurvesDeformableSimAPI`) while preserving existing deformable-body query semantics.
- Adds the `Using Cables` guide covering authoring, materials, collision, runtime state, USD import, and limitations.
- Adds authoring, importer, runtime, state, proxy-coupler, render-sync, and callback-lifetime test coverage, plus public API documentation.

All four stiffness moduli are exposed. `shear_stiffness` and `twist_stiffness` default to `None` and are then left unauthored, so Newton keeps applying its stretch and bend fallbacks and existing cables are unchanged. An authored zero is kept as zero rather than treated as unset.

Cable curve points are synchronized to the render delegate through CPU Fabric because the RTX Hydra delegate does not read GPU-backed Fabric arrays for `BasisCurves` (NVBug 6502662). The on-device sync path can be restored once that bug is fixed.

VBD builder coloring is unified into `NewtonVBDManager.start_simulation`, which colors any VBD builder (cables and deformables) on the clone path while the USD-import path is colored during builder instantiation. This supersedes the `post_replicate_hooks` / `color_registered_deformables` post-replication coloring, which is removed. Deformable coloring is preserved through the unified path.

## Known limitations

These are documented in the guide's Limitations section rather than worked around.

- **Uniform point spacing is assumed.** Newton derives one stiffness pair from the mean segment length, so uneven control-point spacing mistunes the outlier segments. Tracked upstream as newton-physics/newton#3722. This cannot be corrected on the Isaac Lab side: `joint_penalty_k` is snapshotted once during initialization, so post-finalize writes to `joint_target_ke` have no effect.
- **Contact properties are not per-cable.** The Newton cable importer derives each cable's `ShapeConfig` from the import-wide default and carries over only density and the collision flags, so friction, contact stiffness/damping, margin, and gap fall back to solver-global values. Tracked upstream as newton-physics/newton#3734. `collision_props` is therefore honored for enabling collision, not for tuning it.

## Type of change

- Documentation update

## Screenshots

N/A. This change adds backend API and simulation behavior without a user-interface change.

## Validation

- The cable authoring, USD import (including adjacent collision filtering and the shear/twist fallback semantics), runtime/state, proxy-coupler, and render-sync test suites pass.
- The cable-pile demo runs under standalone Newton VBD with periodic resets on CPU and CUDA, and renders under the Kit/RTX visualizer (`--visualizer kit`, 250 steps) with the reset firing and no missing-curve-point warnings.
- Masked segment writes are verified CUDA-graph capturable.
- Full pre-commit checks passed.
- Changelog validation passed for every touched package.
- Documentation built successfully without warnings.

## Checklist

- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package; CI handles changelog compilation and version bumps


source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -589,6 +648,46 @@ def sync_transforms_to_usd(cls) -> None:
...
+                require_attrs=[
+                    (usdrt.Sdf.ValueTypeNames.Point3fArray, "points", usdrt.Usd.Access.ReadWrite),
+                    (usdrt.Sdf.ValueTypeNames.UInt, cls._newton_cable_offset_attr, usdrt.Usd.Access.Read),
+                    (usdrt.Sdf.ValueTypeNames.UInt, cls._newton_cable_count_attr, usdrt.Usd.Access.Read),
+                    (usdrt.Sdf.ValueTypeNames.Matrix4d, "omni:fabric:worldMatrix", usdrt.Usd.Access.Read),
+                ],
+                device="cpu",
+            )
+            if selection.GetCount() == 0:
+                return
```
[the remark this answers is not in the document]
mmichelis: Good catch, fixed in f7a7971 by clearing `_cables_dirty` on the empty-selection early return.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
```diff
@@ -112,6 +112,8 @@ def start_simulation(cls) -> None:
         having Newton bind a surface mesh to volume deformable tetrahedral mesh
         in addition to removing the deformable_registry data structure.
         """
+        if cls._builder is not None:
+            cls._builder.color()
         super().start_simulation()
```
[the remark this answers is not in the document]
mmichelis: This is guarded and cannot double-color. `NewtonVBDManager.start_simulation` colors only when `_builder is not None`, while the parent `start_simulation` runs `instantiate_builder_from_stage` (which colors) only when `_builder is None`. Nothing mutates `_builder` between those two checks, so the paths are mutually exclusive and exactly one `color()` runs per start. On the USD-import path `_builder` is `None` at entry, so the override call is skipped and only `instantiate_builder_from_stage` colors. Removing that call would instead leave the USD-import path uncolored. Added a comment documenting this in f7a7971.

maxkra15:
From what I understand this filters adjacent segments but leaves non-aadjacent segments of the same cable colliding? Is this behavior desired?

maxkra15:
There is no per-cable friction, contact stiffness, damping, margin or gap etc cfg; the cable inherits the scene-global newton shape configs? i think a public self-collision policy and per-cable contact etc configuration would make this more usable

maxkra15:
In practice we see that we often need rigid plug/end fitting with calibrated mass and inertia and/or fixed tail attachments. See [the waterhose demo](github.com/maxkra15/IsaacLab/tree/IsaacLab-rby1-waterhose-demo-experiemntal )
Also we might then need connector-specific collision parameters and corresponding body-frame conventions

maxkra15:
the docs say shear and twist cannot be expressed and are ignored. The pinned newton importer actually reads those properties and when absent, shear defaults to stretch and twist defaults to bend, therefore the current two-parameter API introduces implicit shear/twist stiffness if i see correctly in my experiments

Also, IsaacLab accepts nonuniform point spacing, while Newton converts the moduli using one mean segment length. that matters for eg again the experiemntal example, whose first segment is much longer than the remaining segments.

The relevant conversion is in the pinned Newton [importer](https://github.com/newton-physics/newton/blob/d7581b732d83843577411d36f6a58cccee909bce/newton/_src/utils/import_usd_deformable_cable.py)

source/isaaclab/isaaclab/assets/cable_object/base_cable_object.py
AntoineRichard, line 1: Should the abstract object provide handles for basic shared cable properties like stifness, etc..?

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,296 @@
...
+        ``0.001``.
+    * - ``density``
+      - Cable density [kg/m^3]. Must be finite and positive. Newton derives
+        per-segment mass from the density and the capsule volume. Default
+        ``1000.0``.
+    * - ``stretch_stiffness``
+      - Axial (stretch) elastic modulus ``E`` [Pa], i.e. force per area. Must be
+        finite and nonnegative. Newton converts it to the rod's per-joint axial
+        stiffness ``E * A / L``, where ``A`` is the circular cross-section area
+        and ``L`` is the segment rest length. Higher values reduce elongation
```
AntoineRichard: AI-Generated Review: **Important — this describes a per-segment conversion that the pinned importer does not perform.**

Newton computes one `seg_len = sum(segment_lengths) / num_segments` and uses that same value for every joint ([pinned importer, lines 686–703](https://github.com/newton-physics/newton/blob/d7581b732d83843577411d36f6a58cccee909bce/newton/_src/utils/import_usd_deformable_cable.py#L686-L703)). With nonuniform control-point spacing, `L` is therefore the cable-wide mean rest length, not the current segment's rest length.

Please choose and encode one contract:

- reject nonuniform segment lengths during spawning/import;
- implement per-joint conversion; or
- explicitly document that `L` is the mean segment/rest-centerline length.

Add an unequal-spacing regression case and inspect `builder.joint_target_ke`; otherwise this mismatch remains invisible.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,296 @@
...
+      - Bending elastic modulus ``E`` [Pa]. Must be finite and nonnegative.
+        Newton converts it to the per-joint bend stiffness ``E * I / L``, where
+        ``I`` is the second moment of area of the circular cross-section.
+        ``0.0`` gives a limp rope; increase for a stiff hose or wire. Default
+        ``1.0e6``.
+
+.. note::
+    Shear and twist stiffness are intentionally not exposed, and there are no
+    separate damping parameters. The imported rod cannot express shear or twist
+    stiffness, and Newton warns and ignores those attributes if they are present
```
AntoineRichard: AI-Generated Review: **Important — this statement is factually opposite to the pinned Newton behavior.**

The importer reads `shearStiffness` and `twistStiffness` when authored ([conversion helper](https://github.com/newton-physics/newton/blob/d7581b732d83843577411d36f6a58cccee909bce/newton/_src/utils/import_usd_deformable_cable.py#L83-L98)); it does not warn and ignore them. More importantly, when they are absent, `add_joint_cable()` defaults shear stiffness to stretch and twist stiffness to bend ([builder defaults](https://github.com/newton-physics/newton/blob/d7581b732d83843577411d36f6a58cccee909bce/newton/_src/sim/builder.py#L5241-L5264)).

Thus the current four-field Isaac Lab material silently creates nonzero shear and twist. Please expose all four moduli, or explicitly author zero shear/twist if that is the intended cable model, then rewrite this note to describe the resulting behavior exactly.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -57,6 +58,39 @@ class PhysicsMaterialCfg:
...
+    thickness: float = 0.001
+    """The finite, positive full cable thickness (diameter) [m]. Defaults to 0.001 m."""
+
+    density: float = 1000.0
+    """The finite, positive cable density [kg/m^3]. Defaults to 1000 kg/m^3."""
+
+    stretch_stiffness: float = 1.0e9
+    """The finite, nonnegative cable stretch elastic modulus [Pa]. Defaults to 1e9 Pa."""
+
+    bend_stiffness: float = 1.0e6
```
AntoineRichard, line 79: AI-Generated Review: **Important — omitting shear/twist here is not a neutral API simplification.**

The Newton joint has four driven DOFs. Because the importer passes missing values as `None`, the builder substitutes `shear = stretch` and `twist = bend`. A user setting only these two advertised moduli therefore gets two additional, hidden stiffnesses.

Please either add validated `shear_stiffness` and `twist_stiffness` fields with SI-unit docstrings and author their USD attributes, or make the spawner author explicit zeros. The selected semantics should be asserted against the effective `builder.joint_target_ke`, not only the USD material dictionary.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
```diff
@@ -0,0 +1,63 @@
...
+
+    assert builder.body_count == 2
+    assert builder.joint_count == 1
+    assert builder.shape_count == 2
+    assert import_result["path_cable_map"][cable_path] == ([0, 1], [0])
+    assert cable_attrs["closed"] is False
+    assert cable_attrs["material"]["thickness"] == pytest.approx(material.thickness)
+    assert cable_attrs["material"]["density"] == pytest.approx(material.density)
+    assert cable_attrs["material"]["stretchStiffness"] == pytest.approx(material.stretch_stiffness)
+    assert cable_attrs["material"]["bendStiffness"] == pytest.approx(material.bend_stiffness)
```
AntoineRichard, line 97: AI-Generated Review: **Important coverage gap — these assertions only prove that authored values were copied into the import-result dictionary.**

They pass even when the effective Newton joint has unintended shear/twist stiffness and when every joint uses the wrong length convention. After `builder.add_usd`, assert the four per-joint entries in `builder.joint_target_ke` (stretch, shear, bend, twist) against the intended `E*A/L` and `E*I/L` values. Include unequal segment lengths so the test distinguishes per-segment from mean-length conversion. This regression test should fail before the contract fix and pass afterward.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
```diff
@@ -129,3 +132,32 @@ class ConeCfg(ShapeCfg):
...
+    Requires at least three finite points with consecutive points separated by more than 1e-8 m.
+    """
+
+    physics_material_path: str = "physics_material"
+    """Path to the physics material, relative to the cable geometry prim."""
+
+    physics_material: materials.CableMaterialCfg = MISSING
+    """Cable physics material."""
+
+    collision_props: (
```
AntoineRichard, line 160: AI-Generated Review: **Important — this public type promises more collision configuration than cable import honors.**

The spawner accepts and authors arbitrary `CollisionFragment`s, including `NewtonCollisionCfg(contact_margin=..., contact_gap=...)`. The pinned cable importer then constructs `cable_cfg` from `builder.default_shape_cfg` and carries over only density plus `has_shape_collision`/`has_particle_collision`; margin, gap, friction, stiffness, damping, and restitution remain global defaults.

Please either plumb supported fragments/material values into each cable's Newton `ShapeConfig`, or narrow this field to the actually supported enabled/disabled contract and document that all remaining contact properties are solver-global. Silently accepting ignored fields is especially risky for a new public API.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def test_newton_imports_cable_with_adjacent_collision_filtering():
+    """Test that only connected cable segments are collision-filtered."""
+    stage = sim_utils.create_new_stage()
+    cfg = CableCfg(
+        positions=[(0.0, 0.0, 0.0), (0.2, 0.0, 0.0), (0.4, 0.0, 0.0), (0.6, 0.0, 0.0)],
+        physics_material=CableMaterialCfg(),
+        collision_props=[sim_utils.UsdPhysicsCollisionCfg(collision_enabled=True)],
```
AntoineRichard, line 109: AI-Generated Review: **Add a behavioral collision-configuration test, not only an enabled/filtering test.**

Author distinctive values, for example `NewtonCollisionCfg(contact_margin=0.123, contact_gap=0.045)`, then assert the imported segment entries in `builder.shape_margin` and `builder.shape_gap`. If the public contract is narrowed instead, assert that unsupported fragments are rejected with a clear error. Also cover the chosen per-cable friction/material path via `builder.shape_material_mu` (or explicitly document and test that it is unavailable).

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -600,6 +659,47 @@ def sync_transforms_to_usd(cls) -> None:
...
+                    (usdrt.Sdf.ValueTypeNames.UInt, cls._newton_cable_count_attr, usdrt.Usd.Access.Read),
+                    (usdrt.Sdf.ValueTypeNames.Matrix4d, "omni:fabric:worldMatrix", usdrt.Usd.Access.Read),
+                ],
+                device="cpu",
+            )
+            if selection.GetCount() == 0:
+                NewtonManager._cables_dirty = False
+                return
+            _, _, body_q, _, _ = cls._cable_sync_cpu_buffers
+            wp.copy(body_q, cls._state_0.body_q)
```
AntoineRichard, line 744: AI-Generated Review: **Important scalability issue — this copies every body transform to CPU for cable rendering.**

`body_q` was allocated with `empty_like(state_0.body_q)`, so one cable causes an O(total model bodies) device-to-host copy each dirty render frame, although `_sync_cable_points` reads only cable segment bodies. This will penalize the mixed robot+cable scenes advertised in the guide.

Please build a compact cable-body mapping during `_initialize_fabric_cable_prims`, gather only those transforms on-device into a compact GPU array, and copy that compact array to its CPU mirror. The Fabric offsets can then index the compact transform array directly. Add a unit-level assertion that the CPU buffer length equals the number of cable segment bodies, not `model.body_count`.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
```diff
@@ -0,0 +1,296 @@
...
+material to the imported prim before construction.
+
+
+Limitations
+-----------
+
+* **Newton + VBD only.** Other backends and other Newton solvers are not
+  supported.
+* **One curve per object.** :class:`~isaaclab.assets.CableObject` expects exactly
+  one qualifying ``BasisCurves`` under its ``prim_path``.
```
AntoineRichard: AI-Generated Review: **Document the stronger runtime topology restriction here.**

The implementation raises unless there is exactly one **standalone, unwelded cable articulation per simulation world**. “One curve per object” does not tell users that rigid plugs, end fittings, welded cable graphs, and cable-to-rigid attachments cannot be managed through `CableObject`.

Please state that restriction verbatim and identify the supported lower-level path, if one exists. If attached cables are intended to be part of this PR's public task story, the runtime object must support them instead of failing during `_initialize_impl`.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
```diff
@@ -0,0 +1,322 @@
...
+    ) -> None:
+        """Set segment poses for selected environments.
+
+        Args:
+            segment_pose: Actor-frame poses in simulation world frame. The Torch shape is
+                (len(env_ids), num_segments, 7), with position ``(x, y, z)`` [m] followed by quaternion
+                ``(x, y, z, w)``. The Warp shape is (len(env_ids), num_segments), dtype ``wp.transformf``.
+            env_ids: Environment indices. If None, all instances are used.
+        """
+        env_ids = self._resolve_env_ids(env_ids)
```
AntoineRichard: AI-Generated Review: **Moderate duplication — centralize the segment-state write lifecycle.**

All four public writers repeat: selector resolution, `ProxyArray` unwrapping, Torch contiguity conversion, shape/dtype validation, launching into both active states, invalidation, and `update(0.0)`. Any future change to state coherence or capture behavior must be kept manually synchronized across four methods.

Please introduce a private helper such as `_write_segment_state_to_sim(value, selector, dtype, index_kernel, mask_kernel, state_attribute)` and leave the public methods responsible only for their selector-specific contract. This reduces duplication without merging the public indexed/masked APIs.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/kernels.py
```diff
@@ -0,0 +1,74 @@
...
+    segment_pose: wp.array2d(dtype=wp.transformf),
+    env_ids: wp.array(dtype=wp.int32),
+    root_body_ids: wp.array(dtype=wp.int32),
+    link_body_ids: wp.array2d(dtype=wp.int32),
+    body_q: wp.array(dtype=wp.transformf),
+):
+    """Write selected cable segment poses to Newton body state."""
+    index, segment = wp.tid()
+    env_id = env_ids[index]
+    body_id = root_body_ids[env_id]
```
AntoineRichard: AI-Generated Review: **Moderate duplication — factor segment-to-body resolution once.**

The root/link selection at lines 20–22 is repeated in all four kernels. Define a small typed `@wp.func`, for example `_segment_body_id(env_id, segment, root_body_ids, link_body_ids) -> int`, and call it from each specialized pose/velocity kernel. The separate kernels can remain for Warp dtype and selector specialization while the topology rule has one implementation.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
```diff
@@ -112,6 +112,11 @@ def start_simulation(cls) -> None:
         having Newton bind a surface mesh to volume deformable tetrahedral mesh
         in addition to removing the deformable_registry data structure.
         """
+        # Color the VBD graph once. The replicate path supplies an uncolored builder,
+        # colored here. The USD-import path leaves _builder None here, so this is skipped
+        # and super() colors it in instantiate_builder_from_stage. The guards are exclusive.
```
AntoineRichard: AI-Generated Review: **Comment hygiene — this is review-thread defense rather than durable functional documentation.**

“The guards are exclusive” and the narration of both review paths explain why a reported double call is not a bug, rather than what this block must accomplish. The guard and method names already expose the mechanism.

Please remove the comment or reduce it to one functional statement, for example:

```python
# Color replicated builders before model finalization.
```

The nearby NVBug TODO is a good example of a useful comment because it records a concrete external constraint and removal condition.

source/isaaclab/isaaclab/assets/cable_object/base_cable_object.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @property
+    @abstractmethod
+    def num_segments(self) -> int:
+        """Number of rigid segments per cable."""
+        raise NotImplementedError()
+
+    @property
+    @abstractmethod
+    def root_view(self):
```
AntoineRichard, line 63: AI-Generated Review: **Public interface is missing a specific return type.**

`root_view` is part of the new public abstract contract, but callers receive no static type while the Newton override returns `ArticulationView`. Please make `BaseCableObject` generic in a root-view type, or define a small backend-neutral protocol and annotate this property with it. Avoid `object`/`Any`; that would preserve the current unusable contract rather than satisfy the repository's specific-public-types rule.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
```diff
@@ -0,0 +1,322 @@
...
+class CableObject(BaseCableObject):
+    """A Newton cable that requires the VBD solver."""
+
+    cfg: CableObjectCfg
+    """Configuration instance for the cable object."""
+
+    __backend_name__: str = "newton"
+    """The name of the backend for the cable object."""
+
+    def __init__(self, cfg: CableObjectCfg):
```
AntoineRichard: AI-Generated Review: **Complete constructor return annotations.**

This new public constructor should be `def __init__(self, cfg: CableObjectCfg) -> None:`. Apply the same correction to the other new constructors at `source/isaaclab/isaaclab/assets/cable_object/base_cable_object_data.py:14` and `source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object_data.py:28` so the new API consistently follows the repository's modern type-hint rule.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -57,6 +58,39 @@ class PhysicsMaterialCfg:
...
+    """The finite, nonnegative cable bend elastic modulus [Pa]. Defaults to 1e6 Pa."""
+
+    def validate_config(self):
```
AntoineRichard: AI-Generated Review: **Add the missing public return annotation.**

This mutates no state and returns no value, so the signature should be:

```suggestion
    def validate_config(self) -> None:
```

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -1927,6 +1927,33 @@ def _fix_tet_winding_kernel(
         tet_indices[i, 3] = v2
 
 
+def define_deformable_curve_properties(prim_path: str, stage: Usd.Stage | None = None):
```
AntoineRichard: AI-Generated Review: **Add the missing public return annotation.**

This schema writer either returns normally or raises and should be declared explicitly:

```suggestion
def define_deformable_curve_properties(prim_path: str, stage: Usd.Stage | None = None) -> None:
```

scripts/demos/cables.py
```diff
@@ -0,0 +1,140 @@
...
+    ground_cfg.func("/World/defaultGroundPlane", ground_cfg)
+    light_cfg = sim_utils.DomeLightCfg(intensity=3000.0, color=(0.75, 0.75, 0.75))
+    light_cfg.func("/World/light", light_cfg)
+
+    cable_length = 0.5
+    segment_length = cable_length / num_segments
+    thickness = 0.01
+    radius = 0.5 * thickness
+    stretch_modulus = 5.0e5 * segment_length / (math.pi * radius**2)
+    bend_modulus = 20.0 * segment_length / (0.25 * math.pi * radius**4)
```
AntoineRichard: AI-Generated Review: **Showroom hygiene — name the physical targets instead of embedding unexplained constants in the conversion.**

`5.0e5` and `20.0` are the target per-joint axial and bend stiffnesses, while the resulting variables are elastic moduli. Introduce names such as `target_stretch_stiffness` [N/m] and `target_bend_stiffness` [N·m/rad], then derive the two moduli. This makes the calculation self-documenting without adding a long rationale about why those demonstration values were chosen.

scripts/demos/cables.py
```diff
@@ -0,0 +1,140 @@
...
+    for index in range(num_cables):
+        angle = random.uniform(0.0, 2.0 * math.pi)
+        position = (
+            random.uniform(-xy_jitter, xy_jitter) - 0.5 * cable_length * math.cos(angle),
+            random.uniform(-xy_jitter, xy_jitter) - 0.5 * cable_length * math.sin(angle),
+            z_base + index * z_spacing,
+        )
+        orientation = (0.0, 0.0, math.sin(0.5 * angle), math.cos(0.5 * angle))
+        visual_material = None
+        if args_cli.visualizer and "kit" in args_cli.visualizer:
```
AntoineRichard: AI-Generated Review: **Showroom hygiene — avoid a hidden dependency on global CLI state.**

`design_scene(num_cables, num_segments)` appears reusable but reads `args_cli.visualizer` internally. Pass the visualizer selection (or a `create_visual_materials: bool`) explicitly from `main()`. The function's behavior should be determined by its arguments, which also makes this showroom path straightforward to test.

AntoineRichard (changes requested):
AI-Generated Review: **Request changes**

I reviewed the complete PR diff, the existing review discussion, the pinned Newton implementation, and the relevant tests. The inline comments contain precise fix locations and regression-test targets.

The merge-blocking issues are:

1. The documented stiffness contract is not the model Newton actually creates: absent shear/twist values inherit stretch/bend, and a single mean segment length is used for every joint.
2. `CableCfg.collision_props` accepts and authors collision fragments that the Newton cable importer silently drops except for the enabled flag; per-cable friction/contact behavior is consequently not configurable.
3. Kit cable rendering performs a full-model GPU-to-CPU `body_q` copy at render cadence when any cable exists.

Before merge, please also make the standalone/unwelded topology restriction explicit, consolidate the repeated write/kernel paths, replace review-history comments with functional comments, complete the new public type annotations, and remove hidden/magic state from the showroom demo.

Verification performed on this exact head:

- all-files pre-commit hooks passed twice;
- 20 focused cable spawner/import/object tests passed;
- the PR's `isaaclab_newton`, core, and rendering CI jobs pass.

The prior `_cables_dirty` issue is fixed. The reported double `color()` call is not a functional double-call because the builder paths are mutually exclusive, but the newly added defensive narrative comment should still be simplified.

AntoineRichard (commented):
At a high level my main issue with this PR is the lack of domain randomization options for runtime changes. But I may have missed this API.

mmichelis:
> From what I understand this filters adjacent segments but leaves non-aadjacent segments of the same cable colliding? Is this behavior desired?

Indeed, this is desired behavior that we have self collision enabled by default (except immediate neighbors since the rods run into self-collision the entire time). We could have the cables not self collide at all as another option, but since there is a newton selfCollisionEnabled USD flag but newton is not yet reading that for cables, we could push this as an additional implementation in newton. I've opened a draft PR for this: https://github.com/newton-physics/newton/pull/3709

mmichelis:
> There is no per-cable friction, contact stiffness, damping, margin or gap etc cfg; the cable inherits the scene-global newton shape configs? i think a public self-collision policy and per-cable contact etc configuration would make this more usable

Good point. Right now the cable segments inherit the scene-global Newton shape config (NewtonShapeCfg: mu/ke/kd), same as any shape without an authored contact material. Per-cable friction/stiffness/damping would need Newton's cable importer to read a per-shape contact material bound to the curve prim, which it doesn't do yet: unlike the rigid parse_usd path, the capsule segments are created inside add_rod with cfg=replace(builder.default_shape_cfg, ...) and never resolve a bound material.

I'd prefer to keep this PR as-is and add per-cable contact once Newton supports reading those parameters on the cable import path. At that point we can expose them declaratively through the existing material machinery (CableMaterialCfg) with no post-hoc plumbing on the IsaacLab side. Or do you think this is needed before we should merge in this cable PR?

AntoineRichard:
> > There is no per-cable friction, contact stiffness, damping, margin or gap etc cfg; the cable inherits the scene-global newton shape configs? i think a public self-collision policy and per-cable contact etc configuration would make this more usable
> 
> Good point. Right now the cable segments inherit the scene-global Newton shape config (NewtonShapeCfg: mu/ke/kd), same as any shape without an authored contact material. Per-cable friction/stiffness/damping would need Newton's cable importer to read a per-shape contact material bound to the curve prim, which it doesn't do yet: unlike the rigid parse_usd path, the capsule segments are created inside add_rod with cfg=replace(builder.default_shape_cfg, ...) and never resolve a bound material.
> 
> I'd prefer to keep this PR as-is and add per-cable contact once Newton supports reading those parameters on the cable import path. At that point we can expose them declaratively through the existing material machinery (CableMaterialCfg) with no post-hoc plumbing on the IsaacLab side. Or do you think this is needed before we should merge in this cable PR?

Can we capture the follow-ups somewhere, maybe in an issue so we can add this later on?

mmichelis:
> In practice we see that we often need rigid plug/end fitting with calibrated mass and inertia and/or fixed tail attachments. See [the waterhose demo](github.com/maxkra15/IsaacLab/tree/IsaacLab-rby1-waterhose-demo-experiemntal) Also we might then need connector-specific collision parameters and corresponding body-frame conventions

I will keep this PR as purely cables, with the attachments being the followup! This should be a straightforward addition, but I would prefer to keep it separate. But thanks for the request, this will be the next important thing to add!

mmichelis:
> the docs say shear and twist cannot be expressed and are ignored. The pinned newton importer actually reads those properties and when absent, shear defaults to stretch and twist defaults to bend, therefore the current two-parameter API introduces implicit shear/twist stiffness if i see correctly in my experiments
> 
> Also, IsaacLab accepts nonuniform point spacing, while Newton converts the moduli using one mean segment length. that matters for eg again the experiemntal example, whose first segment is much longer than the remaining segments.
> 
> The relevant conversion is in the pinned Newton [importer](https://github.com/newton-physics/newton/blob/d7581b732d83843577411d36f6a58cccee909bce/newton/_src/utils/import_usd_deformable_cable.py)

Thanks, you were right on both counts and both are now addressed.

On the shear/twist topic, fixed in f383430: CableMaterialCfg now exposes shear_stiffness and twist_stiffness. Both are optional and default to None, in which case the attribute is not authored and the solver keeps applying its stretch/bend fallback, but you can now set them to decouple torsion from bending. An authored 0.0 is deliberately distinct from unset: it removes that resistance rather than falling back.

The docs were wrong here and are corrected: four DOFs, unset means "reuses stretch/bend" rather than "no shear/twist resistance". There are regression tests pinning the fallback behaviour and the override in both directions.

For non-uniform spacing, I agree it's the more consequential of the two. Since this is a conversion bug in Newton rather than something Isaac Lab should patch, it's being fixed upstream in an upcoming Newton PR. Until that lands and we bump the pin, unevenly spaced cables are mistuned; the Isaac Lab docs note the limitation explicitly rather than just advising uniform spacing, since resampling would defeat the point of an intentionally graded cable.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
```diff
@@ -112,6 +112,10 @@ def start_simulation(cls) -> None:
         having Newton bind a surface mesh to volume deformable tetrahedral mesh
         in addition to removing the deformable_registry data structure.
         """
+        # Color the replicated builder before finalization. The USD-import path leaves
+        # _builder None here and colors in instantiate_builder_from_stage instead.
```
AntoineRichard: Keep this comment functional. `# Color the replicated builder before finalization.` states the required action; the second sentence only restates the guard and the alternate method's control flow. Please remove that sentence.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
```diff
@@ -0,0 +1,301 @@
...
+                kernel,
+                dim=(env_ids.shape[0], self.num_segments),
+                inputs=[value, env_ids, self.data._sim_bind_root_body_ids, self.data._sim_bind_link_body_ids],
+                outputs=[getattr(state, state_attribute)],
+                device=self.device,
+            )
+        SimulationManager.invalidate_body_state(env_ids)
+        self.update(0.0)
+
+    def _write_segment_state_to_sim_mask(
```
AntoineRichard: The refactor still duplicates the complete write lifecycle between the index and mask helpers: value unwrapping/contiguity, state iteration, kernel launch, invalidation, and refresh. Normalize and validate the selector first, then share that lifecycle in one helper. A Python selector branch is outside the Warp hot path and is preferable to maintaining two copies.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
```diff
@@ -0,0 +1,143 @@
...
+    assert cable_attrs["material"]["thickness"] == pytest.approx(material.thickness)
+    assert cable_attrs["material"]["density"] == pytest.approx(material.density)
+    assert cable_attrs["material"]["stretchStiffness"] == pytest.approx(material.stretch_stiffness)
+    assert cable_attrs["material"]["bendStiffness"] == pytest.approx(material.bend_stiffness)
+    # Shear/twist are left unauthored when unset.
+    assert "shearStiffness" not in cable_attrs["material"]
+    assert "twistStiffness" not in cable_attrs["material"]
+
+
+def test_newton_imports_cable_shear_and_twist_stiffness():
```
AntoineRichard: This test only rechecks the importer's intermediate result dictionary. `test_spawn_cable_authors_optional_shear_and_twist_stiffness` already verifies the authored USD attributes, while `test_newton_cable_authored_shear_and_twist_override_fallbacks` verifies the effective `joint_target_ke` behavior through the importer. Remove this duplicate; it adds setup and runtime without unique regression coverage.

skills/user/prepare-assets-for-newton/reference.md
```diff
@@ -98,3 +111,12 @@ actions. Classify it before tuning:
...
 Use `debug_mode` for iteration-cap evidence. Raise overflowing capacity first and change
 convergence work only after the model, reset, controller, contact path, and capacities are valid.
+
+## Cable Assets
+
+Cables are authored fresh as Newton deformables, not converted from a PhysX rigid asset. They are Newton + VBD only.
+
+- Author with `CableCfg` + `CableMaterialCfg` (`spawn_cable`), or load an external USD via `UsdFileCfg`.
+- For Newton to import a curve as a cable, its `UsdGeom.BasisCurves` prim must be open, linear, nonperiodic, carry the `PhysicsCurvesDeformableSimAPI` applied schema, and bind a deformable-curve material (`PhysicsCurvesDeformableMaterialAPI`) with `thickness`, `density`, `stretchStiffness`, and `bendStiffness` in the `physics:` namespace. A bare exported curve without these imports as static geometry.
+- `thickness` is the full diameter; `stretch_stiffness` / `bend_stiffness` are elastic moduli in Pa. Shear and twist stiffness are not expressed by the rod. Collision is opt-in via `collision_props`, with adjacent-segment-only filtering.
```
AntoineRichard: This duplicates the cable contract from `using-cables.rst` and has already drifted: the current PR exposes `shear_stiffness` and `twist_stiffness`, so “not expressed by the rod” is false. Keep this MJWarp migration skill as a short link to the canonical cable guide (or move cable guidance to a dedicated cable skill) instead of maintaining a second field list here.

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -882,11 +882,14 @@ def bind_physics_material(
     has_physics_scene_api = "PhysxSceneAPI" in applied
     has_collider = prim.HasAPI(UsdPhysics.CollisionAPI)
     has_deformable_body = has_deformable_body_api(prim)
+    has_deformable_curve = "PhysicsCurvesDeformableSimAPI" in prim.GetPrimTypeInfo().GetAppliedAPISchemas()
```
AntoineRichard: The raw `PhysicsCurvesDeformableSimAPI` membership test is now repeated here, in `CableObject._initialize_impl`, and in `NewtonManager._initialize_fabric_cable_prims`. Add `has_deformable_curve_api` next to `has_deformable_body_api` and use it consistently so schema detection has one definition.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
```diff
@@ -57,6 +58,55 @@ class PhysicsMaterialCfg:
     """Function to use for creating the material."""
 
 
+@configclass
+class CableMaterialCfg(PhysicsMaterialCfg):
```
AntoineRichard, line 62: Adding this AOUSD curve-material subclass makes `PhysicsMaterialCfg`'s statement that physics materials are PhysX schemas inaccurate. Update the base docstring to describe backend-neutral USD physics material configurations. The helper's `Returns:` entry should likewise say “deformable material prim,” not “deformable body material prim.”

AntoineRichard (changes requested):
Requesting changes. The latest head passes the targeted authoring/import, runtime, and GPU/Kit rendering tests, full pre-commit, and the complete documentation build. The remaining issues are design and maintainability gaps that those checks do not cover.

The existing unresolved inline threads remain applicable: `collision_props` accepts properties that Newton silently ignores; cable rendering copies the full model `body_q` to CPU; the new public `BaseCableObject.root_view` contract is untyped; and the showroom demo helper reads global CLI state. The new inline comments cover a non-functional implementation comment, duplicated state-write and schema-detection logic, duplicated/stale documentation, an unnecessary importer test, and a now-inaccurate base-class docstring.

Please also update the PR description: it still says shear and twist stiffness are intentionally not exposed, which contradicts the current API. Close the runtime-domain-randomization decision as well by either exposing the supported material/property handles or explicitly documenting that these values are author-time only.

The branch currently conflicts with `develop`, and the only reported check on the current head is the labeler, so it needs a rebase and a full CI run before merge.

Local verification on `108ddd37d5cb3f9dda72846a5087358617db7516`: 23 authoring/import tests passed; 6 cable-object tests passed; 2 GPU/Kit render-sync tests passed; `./isaaclab.sh -f` passed; `./isaaclab.sh -d` passed.

mmichelis:
> AI-Generated Review: **Request changes**
> 
> I reviewed the complete PR diff, the existing review discussion, the pinned Newton implementation, and the relevant tests. The inline comments contain precise fix locations and regression-test targets.
> 
> The merge-blocking issues are:
> 
> 1. The documented stiffness contract is not the model Newton actually creates: absent shear/twist values inherit stretch/bend, and a single mean segment length is used for every joint.
> 2. `CableCfg.collision_props` accepts and authors collision fragments that the Newton cable importer silently drops except for the enabled flag; per-cable friction/contact behavior is consequently not configurable.
> 3. Kit cable rendering performs a full-model GPU-to-CPU `body_q` copy at render cadence when any cable exists.
> 
> Before merge, please also make the standalone/unwelded topology restriction explicit, consolidate the repeated write/kernel paths, replace review-history comments with functional comments, complete the new public type annotations, and remove hidden/magic state from the showroom demo.
> 
> Verification performed on this exact head:
> 
> * all-files pre-commit hooks passed twice;
> * 20 focused cable spawner/import/object tests passed;
> * the PR's `isaaclab_newton`, core, and rendering CI jobs pass.
> 
> The prior `_cables_dirty` issue is fixed. The reported double `color()` call is not a functional double-call because the builder paths are mutually exclusive, but the newly added defensive narrative comment should still be simplified.

Thanks, this was a genuinely useful review. I'll answer each point separately below. 

1. Stiffness contract. Split in two, and you were right on both halves.

The shear/twist half is fixed in f38343080c. CableMaterialCfg now exposes shear_stiffness and twist_stiffness; both default to None and are left unauthored, so Newton keeps applying its stretch/bend fallbacks and existing cables are bit-identical.

The mean-segment-length half is a Newton bug, not something Isaac Lab can paper over, so I filed it upstream with a repro: newton-physics/newton#3722.

2. collision_props. Confirmed, and also upstream: newton-physics/newton#3734.

3. Kit render copy. Agreed, though this CPU transfer will be entirely fixed once basiscurves are supported properly on the kit rendering side. We keep the workaround for now.

Follow-ups. All addressed except one, in 75c288b688, fc3c288671, bf12018c7d and 136677346b.

The write and kernel paths are consolidated, though split by selector rather than into the single helper you sketched: a combined helper needs a use_mask flag plus three internal branches and always passes a dead kernel, whereas splitting on the selector axis leaves both helpers branch-free. 

On type annotations, the constructors and define_deformable_curve_properties now declare -> None, since validate_config already did and the new API was inconsistent with itself. I did not type root_view. Every backend returns an unrelated view type (newton.selection.ArticulationView, physx.ArticulationView, physx.DeformableBodyView, OvPhysxView) with no common base, and the property is documented as requiring backend-specific tensor handling, so a neutral protocol has an essentially empty intersection. 

One correction to the topology premise. Rigid plugs, end fittings and cable-to-rigid attachments are not rejected by _initialize_impl. Newton lowers a PhysicsAttachment to a ball joint without add_articulation, so those joints are loop-closing and sit outside the articulation window, invisible to the view. What actually fails is cable-to-cable welds, multi-curve BasisCurves prims and periodic curves. e6ea8dd724 documents that.

On the color() comment. Simplified.

source/isaaclab/isaaclab/assets/cable_object/base_cable_object.py
mmichelis: Not today, and deliberately so. Stiffness is authored at spawn through `CableMaterialCfg` and baked into the model at import time. Newton snapshots `joint_penalty_k` once in `_init_joint_penalty_k` and never refreshes it, so a runtime setter on the abstract object would write `joint_target_ke` and then silently have no effect. Exposing handles that quietly do nothing seemed worse than not exposing them yet.

This is also the root cause behind your domain-randomization point on the review itself: runtime cable material changes need upstream support before the API is worth adding. Happy to add the shared handles to the abstract object as soon as Newton can consume post-finalize edits.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: Correct, and this one is not fixable on our side. Filed upstream with a repro: newton-physics/newton#3722. On a cable with segment lengths `[0.5, 0.1, 0.1, 0.1]` the per-joint stiffness is off by roughly ±50%, and nothing downstream compensates.

Patching it after import does not work either: `joint_penalty_k` is snapshotted once in `_init_joint_penalty_k`, so writes to `joint_target_ke` after finalize silently no-op. That rules out an Isaac Lab side fix rather than merely making one awkward.

Of your three options I took the third for now. The conversion table states that `L` is the mean segment rest length, and Limitations carries an explicit "uniform point spacing assumed" entry noting that uneven spacing mistunes the outlier segments. I would rather land the per-joint Voronoi length upstream than encode a workaround here.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: You are right, that note was backwards, and it is fixed in f38343080c.

`CableMaterialCfg` now exposes all four moduli. `shear_stiffness` and `twist_stiffness` default to `None` and are left unauthored, so Newton keeps applying its stretch and bend fallbacks and existing cables are unchanged. The note now states that a cable joint has four degrees of freedom and that leaving shear or twist unset reuses the stretch and bend moduli rather than removing that resistance.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
mmichelis: Added rather than zeroed, in f38343080c.

`shear_stiffness` and `twist_stiffness` are validated `float | None` fields in Pa with the same SI docstrings as the other two, and the spawner authors `physics:shearStiffness` / `physics:twistStiffness` only when they are set. Defaulting to unauthored preserves the Newton fallbacks and keeps existing cables bit-identical; authoring explicit zeros would have silently changed the behaviour of every cable already out there. An authored zero is kept as zero rather than being treated as unset.

Asserted against the effective `builder.joint_target_ke` after `add_usd`, not only the USD material dictionary.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
mmichelis: Half done in f38343080c.

The four per-joint `builder.joint_target_ke` entries (stretch, shear, bend, twist) are now asserted after `add_usd`, covering fallback when unset, explicit override, partial authoring, and the authored-zero case that distinguishes zero from a fallback.

The unequal-spacing case is deliberately not added, because right now it would be a test asserting the buggy mean-length behaviour. That is newton-physics/newton#3722. I would rather add it alongside the upstream fix so it lands as a genuine regression test instead of pinning the current conversion in place.

source/isaaclab/isaaclab/sim/spawners/shapes/shapes_cfg.py
mmichelis: Confirmed, and filed upstream with a repro: newton-physics/newton#3734.

The importer builds one `ShapeConfig` per curve and overrides only `density` plus `has_shape_collision` / `has_particle_collision`, so `ke`, `kd`, `kf`, `ka`, `mu`, `margin` and `gap` all fall back to `default_shape_cfg`. In the repro a cable and a rigid capsule carry identical authored contact properties; the capsule picks up all seven values and the cable picks up none. The rigid path already resolves these from `NewtonMaterialAPI` and `NewtonCollisionAPI`, so the request upstream is to do the same per cable prim.

Your call on this PR, and I would like your preference: narrow the field now to the enabled/disabled contract that is actually honored and document the remaining contact properties as solver-global, or keep it as-is and document the gap until the importer lands. I lean toward narrowing, since silently accepting ignored fields on a brand-new public API is the worse failure mode, but it does mean a second signature change once upstream ships.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
mmichelis: Waiting on the decision in the `collision_props` thread, since the test follows from it.

If we narrow the field, the test asserts that unsupported fragments are rejected with a clear error. If we keep it, the test authors distinctive values and asserts that `builder.shape_margin` / `builder.shape_gap` remain at the import defaults, documenting that as the current contract rather than leaving it implicit. Either way it becomes a real behavioural assertion instead of the enabled/filtering one.

Per-cable friction via `shape_material_mu` is part of newton-physics/newton#3734 and is not reachable today.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
mmichelis: Agreed, and this is the one merge-blocker still open in-tree.

The compact gather is the right shape: build the cable-body mapping during `_initialize_fabric_cable_prims`, gather only those transforms on device into a compact array, and copy that to its CPU mirror so the Fabric offsets index it directly. Plus the assertion you suggested, that the CPU buffer length equals the number of cable segment bodies rather than `model.body_count`.

Not done yet. It is next, and I will push it before asking for another look.

docs/source/overview/core-concepts/physical-backends/newton/using-cables.rst
mmichelis: Done in fc3c288671, with one correction to the premise.

Rigid plugs, end fittings and cable-to-rigid attachments are **not** rejected by `_initialize_impl`. Newton lowers a `PhysicsAttachment` to a ball joint without calling `add_articulation`, so the joint is loop-closing and sits outside the articulation window, invisible to the view (`include_loop_closing_joints` defaults to `False`). What actually fails is cable-to-cable welds, multi-curve `BasisCurves` prims, and periodic curves.

That said, "not rejected" is not "supported", and my first version of this fix said plugs and end fittings work, which was wrong. e6ea8dd724 corrects it: a ball joint constrains position only, so a cable end can be pinned but never clamped, and an end fitting that must transfer orientation is not representable. The joint is also only created when the attachment stiffness is unauthored or infinite; a finite stiffness is preserved as metadata with no joint at all. Both are import warnings rather than errors, so the cable initializes normally with the attachment silently missing. That is also the honest answer to @maxkra15's plug and end-fitting question: it is a real gap, and it currently fails quietly.

The lower-level path is documented too. Cable import is driven by the whole stage, so the rejected topologies are still built into the model and still simulate; they can be driven through `NewtonManager` `get_model()` / `get_state_0()` plus your own `newton.selection.ArticulationView`. There is no Isaac Lab asset wrapper for them.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
mmichelis: Done in 75c288b688, though split by selector rather than folded into the single helper you sketched.

A combined `_write_segment_state_to_sim(...)` needs a `use_mask` flag on top of the six parameters, because `selector=None` is legal on both APIs and `wp.int32` vs `wp.bool` would otherwise have to be sniffed. It then branches three ways internally (resolve, validate, invalidate) and always passes one dead kernel. Splitting on the selector axis leaves both helpers branch-free, which is the axis that actually varies, and matches `MPMObject._scatter_to_sim_index` / `_mask` in the same package. The four public signatures and docstrings are unchanged.

One thing worth recording: the `ProxyArray` unwrap has to stay ordered before the shape assertion. `assert_shape_and_dtype` only branches on `int|float`, `wp.array` and `torch.Tensor`, so it silently no-ops on a `ProxyArray`. Both helpers preserve that ordering, but it is a trap for anyone who reorders them later.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/kernels.py
mmichelis: Done in 75c288b688. `_segment_body_id(env_id, segment, root_body_ids, link_body_ids) -> int` is now a single `wp.func` and the four kernels stay specialized for dtype and selector, as you suggested.

Kept the assignment form rather than an early return, to avoid Warp conditional-return edge cases inside the masked guard. Verified numerically identical for both the index and mask variants, and still CUDA-graph capturable.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
mmichelis: Simplified in bf12018c7d and refined in 108ddd37d5.

I kept one clause rather than reducing it to a single line, because the non-obvious part is why the guard exists at all: the replicate path sets `_builder` before `start_simulation` and colors it here, while the USD-import path leaves it `None`, so `super().start_simulation()` dispatches back into `instantiate_builder_from_stage`, which colors there instead. Dropping that leaves a bare guard with no explanation of the second path.

Worth noting the original comment had the mechanism right but the attribution wrong: it credited the base class with coloring, and the base class never calls `color()` at all.

source/isaaclab/isaaclab/assets/cable_object/base_cable_object.py
mmichelis: This is the one I would rather not do here, and I would like to push back.

Every backend returns an unrelated view type: `newton.selection.ArticulationView`, `physx.ArticulationView`, `physx.DeformableBodyView`, `OvPhysxView`. They share no common base, so a backend-neutral protocol would either have an essentially empty intersection, conveying nothing beyond `object`, or would exclude a backend the moment one is added. The property is also documented as requiring backend-specific tensor handling, which is the actual contract: `root_view` is the escape hatch out of the neutral API, and the absence of a neutral type is what makes it honest.

The pattern is also repo-wide, not cable-specific. `base_rigid_object.py:91`, `base_articulation.py:350` and `base_rigid_object_collection.py:89` all declare it exactly this bare way, each with a typed per-backend override, so cable is following the existing convention rather than deviating from it.

A `Generic[ViewT]` base would work technically, but there is no `Generic` anywhere in `source/` today and none in the asset hierarchy at all, so doing it for cable alone makes cable the outlier in the opposite direction, and generics are viral in a way the `-> None` annotations are not. If you want it, I would rather do one follow-up applying it uniformly to all the asset bases.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
mmichelis: Done in 136677346b, plus `base_cable_object.py:35`, which was not in your list. Leaving that one bare would have reproduced exactly the inconsistency being fixed, with the abstract base unannotated while its Newton override and both data classes are annotated. Five signatures in total.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
mmichelis: Already applied in f38343080c.

source/isaaclab/isaaclab/sim/schemas/schemas.py
mmichelis: Done in 136677346b.

scripts/demos/cables.py
mmichelis: Done in bf12018c7d. `target_stretch_stiffness` [N/m] and `target_bend_stiffness` [N·m/rad] are now named, and the two moduli are derived from them, so the conversion reads for itself without a rationale comment.

Verified bit-identical rather than only algebraically equal, since the operand order is unchanged.

scripts/demos/cables.py
mmichelis: Left as-is for now, but I am happy to change it if you feel strongly.

Reading `args_cli` inside `design_scene` is what most of the demos already do: `deformables.py`, `hands.py` and `procedural_terrain.py` all do it, against `arms.py`, `bipeds.py` and `quadrupeds.py`, which take no arguments at all. So threading the visualizer through explicitly makes `cables.py` the demo that differs from the house pattern rather than the one that conforms to it.

The narrower version of your point does stand though: `cables.py` is currently the only `design_scene` that mixes explicit parameters with a global read, which is arguably the worst of both. Say the word and I will pass the visualizer selection in from `main()`.

AntoineRichard (approved):

mmichelis:
> At a high level my main issue with this PR is the lack of domain randomization options for runtime changes. But I may have missed this API.

A very fair point, this is still lacking for deformables as well, which will need further investigations in followup PRs. Some of it is Newton limitations, so we need to file some requests there, but some we could potentially already accommodate in lab directly. But I think this is better in a separate PR.

scripts/demos/cables.py
mmichelis: Changed my mind on this after a closer look, and you were right: 19bef7b390.

The ratio I cited was measuring the wrong thing. The other demos that read `args_cli` inside `design_scene` take no configuration parameters at all, so the global is their only channel. `cables.py` already had a parameter contract and then reached past it for one field, which made it the only hybrid rather than the one conforming.

It is now `design_scene(num_cables: int, num_segments: int, colorize: bool)`, with the visualizer selection resolved in `main()`. The condition itself is unchanged, and `design_scene` no longer references `args_cli` at all.

source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py
mmichelis: Done in c8c5e7a669. Reduced to the single line.

source/isaaclab_newton/isaaclab_newton/assets/cable_object/cable_object.py
mmichelis: Done in c2c652961c, and your framing is what made it work. My objection was to a helper that resolved and validated internally, which does need the extra flag and the branches. Hoisting both into the public writers removes that: the launch dimension is the selector length either way, so the only thing left to branch on is the invalidation form.

There is now one `_write_segment_state_to_sim(value, selector, kernel, state_attribute, *, use_mask)` holding contiguity, state iteration, launch, invalidation and refresh. The four public signatures and docstrings are unchanged.

One detail worth recording: the `ProxyArray` unwrap had to move into the public methods rather than stay in the helper, because `assert_shape_and_dtype` silently no-ops on a `ProxyArray` and validation now runs before the helper is called. Leaving the unwrap behind would have quietly disabled validation for those inputs. It is the same inline two-line form used at the sibling deformable-object writers.

Net 20 lines lighter. Test suite passes, including the masked CUDA-graph capture case, and I checked separately that a wrong-shaped or wrong-dtyped `ProxyArray` still raises.

source/isaaclab_newton/test/sim/test_cable_usd_import.py
mmichelis: Removed in c8c5e7a669. I checked both of your claims against the code first, and they hold: `test_newton_cable_authored_shear_and_twist_override_fallbacks` goes through `add_usd` and reads `builder.joint_target_ke`, and `test_spawn_cable_authors_optional_shear_and_twist_stiffness` asserts the authored `physics:shearStiffness` / `physics:twistStiffness` attributes. No imports or module constants were orphaned by the deletion.

skills/user/prepare-assets-for-newton/reference.md
mmichelis: Done in ed37f3e809, and you had firmer ground than the comment claims. `docs/source/overview/developer-guide/agent_skills.rst` says skills should not become parallel documentation and to prefer links over duplicated snippets, and asks reviewers to reject skills that duplicate. So the field list was a standing violation, and the drift was the symptom rather than the problem.

The section is now three short bullets: cables are authored fresh as Newton deformables rather than converted from a PhysX rigid asset, they are Newton + VBD only, and a link to the guide. Everything else lives only in `using-cables.rst`. `SKILL.md` also stopped naming `UsdGeom.BasisCurves`, which was the next thing positioned to drift. There are now no shear or twist mentions anywhere under `skills/`.

One thing I left for a maintainer: `agent_skills.rst` asks for a bidirectional `seealso` back-link on any docs page acting as a skill's source of truth. The migration guide has one, `using-cables.rst` does not, and the skill now links to it from two places. Strictly the rule keys off the pages named in the skill's Maintenance section, which still lists only the migration guide and the MJWarp page, so it is not a violation today. Happy to add the back-link if you would rather have it.

source/isaaclab/isaaclab/sim/utils/prims.py
mmichelis: Done in ec598e019d. `has_deformable_curve_api` now sits directly after `has_deformable_body_api` in `sim/utils/queries.py`, exported the same way, and all three sites call it.

I checked the accessors before consolidating, since `GetPrimTypeInfo().GetAppliedAPISchemas()` and `GetAppliedSchemas()` are not equivalent for runtime-applied schemas and quietly unifying them would have changed detection. All three sites already used the former, so behavior is unchanged at every one. That is also the correct accessor here, because `PhysicsCurvesDeformableSimAPI` is token-authored and unregistered, which is noted in the helper's docstring. Worth flagging that `prims.py` has a separate `GetAppliedSchemas()` local two lines above the curve test, used for the `PhysxSceneAPI` check, so the two are easy to conflate.

I deliberately left two literals alone: the schema writer in `schemas.py`, and the spawner test that asserts the exact token the writer emits. Routing those through the helper would make them tautological, since a typo shared by writer and helper would then pass.

source/isaaclab/isaaclab/sim/spawners/materials/physics_materials_cfg.py
mmichelis: Done in c8c5e7a669. The base docstring no longer claims physics materials are PhysX schemas; it describes backend-neutral USD physics material configuration and notes that subclasses author a specific backend's schema. The PhysX-only `classPxBaseMaterial` link went with it, since it no longer describes the base class. The `Returns:` entry now says "deformable material prim"; it lives in `physics_materials.py` rather than the cfg file.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -202,14 +243,30 @@ def _scatter_reset_masks_from_ids(
     fk_mask[articulation_ids[world, arti]] = True
 
 
-_SCATTER_RESET_MASKS_FROM_IDS_DISPATCHER = IndexKernelDispatcher(_scatter_reset_masks_from_ids, ("env_ids",))
+_SCATTER_RESET_MASKS_FROM_IDS_DISPATCHER = IndexKernelDispatcher(
+    _scatter_reset_masks_from_ids, ("env_ids",)
+)
```
kellyguo11:
```suggestion
_SCATTER_RESET_MASKS_FROM_IDS_DISPATCHER = IndexKernelDispatcher(_scatter_reset_masks_from_ids, ("env_ids",))
```

kellyguo11:
the failure in demos/deformables.py seems to be related to the changes in the PR

mmichelis:
> the failure in demos/deformables.py seems to be related to the changes in the PR

Fixed in [837ccc8](https://github.com/isaac-sim/IsaacLab/pull/6688/commits/837ccc86eee4bb9c3c91a133aded29a0cbfdfc1c)

kellyguo11:
looks like an error in one of the contrib tasks
```
 ____ test_contrib_environments_smoke[IsaacContrib-Lift-Cube-OpenArm-2-cuda] ____
    source/isaaclab/isaaclab/managers/manager_base.py:407: in _resolve_param_value
        value.resolve(self._env.scene)
    source/isaaclab/isaaclab/managers/scene_entity_cfg.py:141: in resolve
        self._resolve_joint_names(scene)
    source/isaaclab/isaaclab/managers/scene_entity_cfg.py:165: in _resolve_joint_names
        raise ValueError(
    E   ValueError: Both 'joint_names' and 'joint_ids' are specified, and are not consistent.
    E   	from joint names: ['openarm_joint.*', 'openarm_finger_joint.*'] [[0, 1, 2, 3, 4, 5, 6, 7, 8]]
    E   	from joint ids: ['openarm_joint1', 'openarm_joint2', 'openarm_joint3', 'openarm_joint4', 'openarm_joint5', 'openarm_joint6', 'openarm_joint7', 'openarm_finger_joint1', 'openarm_finger_joint2'] [[0, 1, 2, 3, 4, 5, 6, 7, 8]]
    E   Hint: Use either 'joint_names' or 'joint_ids' to avoid confusion.
    
    During handling of the above exception, another exception occurred:
    source/isaaclab_tasks/test/contrib/test_contrib_environments_smoke.py:56: in test_contrib_environments_smoke
        _run_environments(task_name, device, num_envs, create_stage_in_memory=False)
    source/isaaclab_tasks/test/env_test_utils.py:328: in _run_environments
        _check_random_actions(
    source/isaaclab_tasks/test/env_test_utils.py:407: in _check_random_actions
        env = gym.make(task_name, cfg=env_cfg)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    _isaac_sim/kit/python/lib/python3.12/site-packages/gymnasium/envs/registration.py:735: in make
        env = env_creator(**env_spec_kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    source/isaaclab/isaaclab/envs/manager_based_rl_env.py:88: in __init__
        super().__init__(cfg=cfg)
    source/isaaclab/isaaclab/envs/manager_based_env.py:124: in __init__
        self._init_sim()
    source/isaaclab/isaaclab/envs/manager_based_env.py:229: in _init_sim
        self.load_managers()
    source/isaaclab/isaaclab/envs/manager_based_rl_env.py:125: in load_managers
        super().load_managers()
    source/isaaclab/isaaclab/envs/manager_based_env.py:373: in load_managers
        self.observation_manager = ObservationManager(self.cfg.observations, self)
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    source/isaaclab/isaaclab/managers/observation_manager.py:83: in __init__
        super().__init__(cfg, env)
    source/isaaclab/isaaclab/managers/manager_base.py:174: in __init__
        self._prepare_terms()
    source/isaaclab/isaaclab/managers/observation_manager.py:545: in _prepare_terms
        self._resolve_common_term_cfg(f"{group_name}/{term_name}", term_cfg, min_argc=1)
    source/isaaclab/isaaclab/managers/manager_base.py:376: in _resolve_common_term_cfg
        self._process_term_cfg_at_play(term_name, term_cfg)
    source/isaaclab/isaaclab/managers/manager_base.py:395: in _process_term_cfg_at_play
        self._resolve_param_value(term_name, key, value)
    source/isaaclab/isaaclab/managers/manager_base.py:409: in _resolve_param_value
        raise ValueError(f"Error while parsing '{term_name}:{key}'. {e}")
    E   ValueError: Error while parsing 'policy/joint_vel:asset_cfg'. Both 'joint_names' and 'joint_ids' are specified, and are not consistent.
    E   	from joint names: ['openarm_joint.*', 'openarm_finger_joint.*'] [[0, 1, 2, 3, 4, 5, 6, 7, 8]]
    E   	from joint ids: ['openarm_joint1', 'openarm_joint2', 'openarm_joint3', 'openarm_joint4', 'openarm_joint5', 'openarm_joint6', 'openarm_joint7', 'openarm_finger_joint1', 'openarm_finger_joint2'] [[0, 1, 2, 3, 4, 5, 6, 7, 8]]
    E   Hint: Use either 'joint_names' or 'joint_ids' to avoid confusion.
```

kellyguo11:
actually maybe that was fixed in a different PR, will rebase the branch and run CI though the latest

PR #6692, opened 2026-07-23
Apply teleop RTX settings only when a renderer actually runs

rwiltz:
# Description

The teleop scripts applied the RTX/DLSS global settings before creating the
environment — `teleop_replay_agent.py` did so unconditionally, while
`teleop_se3_agent.py` and `record_demos.py` gated on `use_isaac_teleop or --xr`.
Applying these settings eagerly imports `omni.replicator` (part of the
SDG/rendering extensions), which is **not** loaded in a pure-headless run. As a
result, a headless replay (the teleop CI path) — and equally a headless
IsaacTeleop record/teleop session — aborted at startup with:

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/scripts/environments/teleoperation/teleop_replay_agent.py
+++ b/scripts/environments/teleoperation/teleop_replay_agent.py
@@ -996,7 +996,25 @@ def _maybe_launch_cloudxr(cloudxr_env_path: str | None, auto_launch: bool):
     return launcher
 
 
-def _prepare_env_cfg(task: str, num_envs: int, device: str) -> tuple[ManagerBasedRLEnvCfg, object | None]:
+def _rtx_rendering_requested(args: argparse.Namespace) -> bool:
+    """Return whether the CLI selects a renderer that actually drives RTX rendering.
+
+    The RTX/DLSS global settings (and the ``omni.replicator`` extension they configure)
+    are only meaningful when something renders through RTX. That happens when the Kit
+    visualizer is enabled (``--viz kit``), when external cameras are rendered
+    (``--enable_cameras``), or in XR mode (``--xr``, which drives the Kit XR pipeline).
+    A pure-headless replay selects none of these and renders nothing.
+
+    This intentionally reads the CLI intent rather than any Kit/carb runtime state so the
+    check keeps working as these scripts grow support for other renderers and kitless runs.
+    """
+    visualizers = getattr(args, "visualizer", None) or []
+    return bool(getattr(args, "enable_cameras", False)) or ("kit" in visualizers) or bool(getattr(args, "xr", False))
+
+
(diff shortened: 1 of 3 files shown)


kellyguo11 (approved):

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_utils.py
```diff
@@ -114,9 +114,19 @@ def _apply_isaac_rtx_global_settings(
 
     antialiasing_mode = getattr(global_settings, "antialiasing_mode", None)
     if antialiasing_mode is not None:
-        import omni.replicator.core as rep
-
-        rep.settings.set_render_rtx_realtime(antialiasing=antialiasing_mode)
+        try:
+            import omni.replicator.core as rep
```
pbarejko: The proposed approach doesn't seem to be the right fix. The real fix is somewhere upstream in the callers to recognise that we don't need to apply any rtx settings when rendering is not required?

@

rwiltz: Updated

rwiltz:
@greptile

pbarejko (approved):

nvsekkin:
nice, the new method makes more sense then discarding the error in the utility

scripts/environments/teleoperation/teleop_replay_agent.py
```diff
@@ -996,7 +996,25 @@ def _maybe_launch_cloudxr(cloudxr_env_path: str | None, auto_launch: bool):
     return launcher
 
 
-def _prepare_env_cfg(task: str, num_envs: int, device: str) -> tuple[ManagerBasedRLEnvCfg, object | None]:
+def _rtx_rendering_requested(args: argparse.Namespace) -> bool:
```
nvsekkin, line 999: suggestion: `_rtx_rendering_requested` is currently copy-pasted in 3 files - it would be good to find a common place for all

nvsekkin: ~~maybe something like `isaac_rtx_rendering_requested` inside `isaac_rtx_renderer_utils.py`? since the upstream script already import the module~~

@pbarejko mentioned this keep the bool rtx focused, not a good place

nvsekkin (approved):

scripts/environments/teleoperation/teleop_replay_agent.py
rwiltz: SG, can we do it in a follow up? Want to unblock CI and cloud usage with this PR.

PR #6699, opened 2026-07-23
[CI] Split arm-ci into a standalone workflow

hujc7:
# Description

## 1. Summary

Moves the `arm-ci` job out of `build.yaml` ("Docker + Tests") into its own workflow, `.github/workflows/arm-ci.yml`, and extracts the change-detection gate into a reusable workflow, `.github/workflows/detect-changes.yml`, shared by both.

**Problem.** `arm-ci` shared the "Docker + Tests" run with every `test-*` job. GitHub only exposes **"Re-run failed jobs" once the entire run completes**, so a fast test job that failed early could not be retried until `arm-ci`'s slow arm64 build + render finished (up to its 60-minute timeout on the scarce DGX Spark runners). `arm-ci` is `continue-on-error` (informational) and never needed to hold the run open.

**Fix.** `arm-ci` now owns its own workflow run — independent status, independent **Re-run failed jobs**, and no coupling to the main workflow's lifecycle.

## 2. What changed

- **New `.github/workflows/arm-ci.yml`** — the `arm-ci` job (verbatim: `continue-on-error`, 60-minute timeout, same build / marker-test / cartpole-smoke steps) plus an inlined `config` step, gated by the shared change detector.
- **New `.github/workflows/detect-changes.yml`** — reusable (`workflow_call`) change-detection gate; each caller passes its own `patterns`.
- **`build.yaml`** — the `arm-ci` block is removed; its `changes` job now calls the reusable workflow (same patterns); `config` and all `test-*` jobs are unchanged.

## 3. Design notes

- **Required-ready.** `arm-ci` triggers on every PR and gates the arm64 build behind the shared `changes` detector + `if:`. An irrelevant PR **skips** the job (which branch protection reads as green) rather than never creating the check. So `arm-ci` can be **promoted to a required status check later with zero workflow change** — a workflow-level `paths:` filter could not (a not-triggered required check stays pending forever). This is the same `changes`+`if:` pattern `build.yaml` already uses for its required test jobs.
- **Reusable detector.** Both workflows share one change-detection implementation; each declares its own `patterns` (`<ERE regex> :: <description>` per line). `build.yaml` keeps its exact patterns; `arm-ci` additionally lists its real inputs — `pyproject.toml`, `environment.yml`, `isaaclab.*` (the root files `Dockerfile.base` COPYs and `resolve-ov-pins` reads) — which the old gating missed.
- **Trigger parity.** Push to `main`/`develop`/`release/**` always runs (post-merge integration); `pull_request` runs only when arm-relevant paths change.
- **Config inlined** in `arm-ci.yml` (only the base-image name/tag + CI image tag it needs) — not shared, since `arm-ci` builds its own independent `-arm64` image.

## 4. Validation

- [x] `./isaaclab.sh -f` clean on all changed files; YAML parses.
- [x] `build.yaml`'s downstream `needs.changes.outputs.run_docker_tests` references preserved (`build`, `build-curobo`); inline detect logic fully removed.
- [x] **ARM CI** runs as its own workflow on this PR, independent of Docker + Tests; its `Load Config` job succeeds and `arm-ci` is confirmed **not** a required check today.

## 5. Notes

`arm-ci` stays `continue-on-error` / non-required for now; the structure above makes promoting it to a required check a branch-protection-only change.


kellyguo11 (approved):

.github/workflows/arm-ci.yml
```diff
@@ -0,0 +1,202 @@
...
+  NGC_API_KEY: ${{ secrets.NGC_API_KEY }}
+
+jobs:
+  # Shared change-detection gate. The trigger is unfiltered (see header) so the
+  # arm-ci check is always creatable; this decides whether the expensive jobs
+  # actually run. Lists arm-ci's real inputs — including the root files
+  # Dockerfile.base copies (pyproject.toml, environment.yml, isaaclab.*) that a
+  # path filter would otherwise miss.
+  changes:
+    name: Detect Changes
```
nvsekkin, line 66: suggestion (non-blocking): Both callers name the job `Detect Changes`, should we add a `Detect Changes (ARM)` and `Detect Changes (Docker)`

.github/workflows/arm-ci.yml
```diff
@@ -0,0 +1,199 @@
...
+        if [ "$EVENT_NAME" = "pull_request" ]; then
+          ref_component="pr-${PR_NUMBER}"
+        else
+          ref_component="$REF_NAME"
+        fi
+
+        # Sanitize the ref name for use as a Docker tag suffix.
+        sanitized_ref=$(echo "$ref_component" | sed 's/[^a-zA-Z0-9._-]/-/g')
+        echo "ci_image_tag=isaac-lab-ci:${sanitized_ref}-${SHA}" >> "$GITHUB_OUTPUT"
+        echo "CI image tag: isaac-lab-ci:${sanitized_ref}-${SHA}"
```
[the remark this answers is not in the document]
nvsekkin: this is valid but I would say non-blocking since this is already commented here and is a nit

nvsekkin:
overall PR looks good, made a non-blocking comment. we can merge

nvsekkin (approved):

.github/workflows/arm-ci.yml
hujc7: The info is available at workflow level, e.g. `ARM CI / Detect Changes / Detect Changes (pull_request)`.

PR #6702, opened 2026-07-24
isaaclab test overhaul

StafaH:
# Make the test suite runnable, enforce rules for it, and cut CI cost

## Description

The repository had no working local test profile. `pytest source/` aborted on 118+ collection
errors, and because pytest stops the whole run on a collection error, several packages executed
**zero** tests — including `isaaclab`, the largest. Nothing enforced how tests were written, so
the patterns that made the suite slow and fragile kept being added.

This PR does three things:

1. **Makes the suite runnable.** The full Kit-less profile now collects **13,422 tests with zero
   errors** and runs **12,673 of them in 2m37s warm**, on a machine with no Isaac Sim installed.
2. **Writes the rules down and enforces them.** `AGENTS.md` gains a complete testing ruleset, with
   a new `isaaclab-writing-tests` skill and a mechanical guard that runs in CI and pre-commit.
3. **Cuts CI cost.** Every package is split along the Kit boundary, and the largest redundant
   sweeps are removed.

Several defects found along the way were silently removing coverage while CI reported green.
Those are called out below.

Fixes / relates to: _(add issue links)_

## Type of change

## Defects fixed that were hiding coverage

**The `isaaclab` fast lane ran zero tests.** Four modules failed to import, aborting the run. Two
entangled causes: `_iface_test_boot.py` stubbed `isaacsim.core` into `sys.modules` at import time
and never restored it, leaving four unrelated modules importing a half-initialised
`isaaclab_physx`; removing those stubs then exposed `test_articulation_ordering.py`, which builds
partially-initialised articulations with `object.__new__` and only passed because the leaked stub
happened to make `AssetBase` concrete. Now 0 → **11,128 tests in 104s**.

**Kit tests were being skipped in CI.** `requires_kit` modules are excluded unless
`ISAACLAB_RUN_CI_TESTS=1`, and **18 workflow steps** across `build.yaml`, `arm-ci.yml` and
`daily-compatibility.yml` never set it. `isaaclab_rl`, `isaaclab_mimic`, `isaaclab_contrib`,
`isaaclab_teleop`, `isaaclab_visualizers` and `isaaclab_assets` would have stopped running their
Kit tests entirely while still reporting green. All 18 now opt in.

**A leaked `omni` stub made an unrelated test run against a mock.**
`test_physx_scene_data_backend.py` guards itself with `pytest.importorskip("omni.physics.tensors")`.
`_iface_test_boot.py` installed a `MagicMock` under that name, so the guard passed and the test ran
against the mock and failed — but only in a full-repo run. The stubs turned out to be dead; removing
them keeps all 4,038 `assets` tests passing and restores the skip.

**`pytestmark` assigned twice, discarding the first.** In `test_operational_space.py` and
`test_device_constructors.py`. The `arm_ci` marker on the former had never been applied.

**`opencv-python` was never declared** although `scripts/tools/hdf5_to_mp4.py` and
`mp4_to_hdf5.py` import `cv2`. The shipped scripts could not run and their 19 tests had been
uncollectable. Added an `opencv` extra (headless, matching what the isaacsim/mimic/teleop stacks
already pull in).

**Two packages had no CI job at all** — `isaaclab_experimental` and `isaaclab_ppisp`.

## CI cost reductions

**Split every package along the Kit boundary.** Previously all files went through the per-file
orchestrator, so Kit-less modules paid a fresh process each and interleaved with modules that start
Kit. `isaaclab` goes from three mixed shards to one kitless job (109 files, ~100s, 30-minute
timeout) plus three Kit shards (80 files). Same for `isaaclab_tasks`, `isaaclab_physx`,
`isaaclab_newton`.

**Halved the environment sweeps.** Eight `isaaclab_tasks` suites built every registered environment
twice, at `num_envs=2` and again at `num_envs=1`. With 41 core environments that is 82 scene
constructions per file, and the same list is already swept three times over (default physics,
Newton, stage-in-memory). These are the files carrying 10,000-second CI timeouts. Now batched only —
the scalar/batched distinction is generic and already covered by the articulation and rigid-object
initialization tests, which retain both deliberately. Two sibling suites had already made the same
trade.

**Moved 9 suites out of the Kit lane entirely.** Each started Kit and then never used it — most
strikingly `test_implicit_actuator.py`, which built a simulation context for each of its 160 cases
while no test body referenced the fixture. Coverage is unchanged:

| suite | before | after |
|---|---|---|
| `isaaclab/.../test_implicit_actuator.py` | 160 cases in the Kit lane | 160 pass in 1.4s |
| `isaaclab/.../test_ray_caster_patterns.py` | Kit lane | 90 pass in 1.8s |
| `isaaclab_contrib/.../test_thruster.py` | Kit lane | 102 pass in 1.3s |
| `isaaclab_contrib/.../test_drone_geometric_controllers.py` | Kit lane | 96 pass in 1.8s |
| `isaaclab_tasks/.../test_shadow_hand_camera_presets.py` | Kit lane | 54 pass in 1.8s |

`isaaclab_contrib`'s fast lane went 88 → 286 tests, `isaaclab_tasks` 332 → 386.

**`isaaclab_tasks`: 285 errors in 257s → 386 passing in 9.5s.** The errors were seven kitless
rendering suites needing the `ov` extra and failing late inside a fixture; they now declare it.

**Trimmed the device axis on five camera tests.** Rendering runs on the GPU whatever the simulation
device is, and none of the five asserted device placement.

## New: declared optional-dependency extras

Several extras cannot co-resolve (`isaacsim` conflicts with `teleop`, `ov`, `mimic`, `viser`, and
even `test`), so no single environment runs everything. Suites now declare what they need:

```python
pytestmark = pytest.mark.requires_extra("ov")
```

A missing extra is a **hard failure naming the exact install command**, not a silent skip, so absent
coverage stays visible. Other tests still run; the session exits non-zero.
`--ignore-missing-extras` is the explicit escape for a repository-wide sweep.

## New: the ruleset and its enforcement

- **`AGENTS.md` → `## Testing Guidelines`** — when a test earns its existence, when it earns its
  runtime, the Kit-less boundary, extras, naming, and correctness.
- **`skills/developer/writing-tests/`** — working procedure plus before/after examples.
- **`docs/source/refs/contributing.rst`** — contributor-facing subset.
- **`source/isaaclab/test/test_repo_test_boundary.py`** — 1,961 mechanical checks in ~3s, wired
  into `uv run isaaclab -f`. Checks Kit dependencies declared at module scope, no redundant Kit
  markers, one `pytestmark` per module, no import-time argument parsing or `sys.modules` mutation,
  a docstring per test, and globally unique module names.
- **`test_repo_test_boundary_allowlist.txt`** — grandfathers pre-existing files with a reason each,
  so new modules cannot land with a known defect.

Module names had to be made unique: test directories are not packages, so same-named files in
different packages collide under pytest's `prepend` import mode and one silently shadows the other.
Ten files were renamed with package suffixes and five empty `__init__.py` markers removed.

## Screenshots

_None — test and CI infrastructure only._

## Checklist

## Verification

```bash
# Fast local profile, no Isaac Sim required
uv run --extra test --locked python -m pytest source/ scripts/ --ignore-missing-extras
# 12,673 passed in 2m37s warm

# Rule enforcement
uv run --extra test --locked python -m pytest source/isaaclab/test/test_repo_test_boundary.py
uv run --no-project python tools/skills/cli.py check
uv run isaaclab -f
```

Kit lanes were verified per file (`OMNI_KIT_ACCEPT_EULA=YES uv run --extra isaacsim --locked
--with pytest python -m pytest <file> --run-ci-tests`). Note that several modules each starting Kit
in one pytest process segfaults with no output, so the Kit lane must stay one file per process.

## Known failures, not introduced here

- `isaaclab_experimental/test/envs/test_frontend_cfg_conversion.py` — 2 failures. The warp frontend
  has no MDP twins for `is_terminated_term` and `pose_command_success`, so `Isaac-Reach-Franka` and
  `Isaac-Reach-UR10` cannot adapt to warp. Surfaced by unblocking collection; pre-existing.
- `isaaclab_teleop` — 40 failures when the directory is run in one process with the `teleop` extra.
  `test_cloudxr_lifecycle` and `test_control_events` both stub the same `sys.modules` names with
  `if name not in sys.modules`, so the second inherits the first's `MagicMock` instances and their
  exhausted `side_effect` iterators. Every file passes alone, which is how CI runs them.


source/isaaclab_newton/test/assets/test_rigid_object_newton.py
```diff
@@ -1273,6 +1287,55 @@ def test_write_state_functions_data_consistency(num_cubes, device, with_offset,
             torch.testing.assert_close(root_com_vel_w[:, 3:], root_link_vel_w[:, 3:])
 
 
+@pytest.mark.isaacsim_ci
```
mataylor-nvidia, line 1290: Do you know why this was added back?

source/isaaclab_newton/test/assets/test_articulation_newton.py
```diff
@@ -1381,8 +1380,8 @@ def test_initialization_fixed_base(sim, num_articulations, device, articulation_
         torch.testing.assert_close(articulation.data.root_com_vel_w.torch, default_root_vel)
 
 
-@pytest.mark.parametrize("num_articulations", [1, 2])
-@pytest.mark.parametrize("device", test_devices())
+@pytest.mark.parametrize("num_articulations", [2])
+@pytest.mark.parametrize("device", PRIMARY_DEVICE)
```
hujc7, line 1478: What's the reason for this change? test_devices() is needed for mgpu test. For local, it should auto detect the devices, e.g. [cpu, cuda:0]. If we don't want to run on GPU, it can be opted out with a enum input.

.github/workflows/build.yaml
```diff
@@ -613,8 +613,32 @@ jobs:
...
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 1
+        lfs: true
+    - uses: ./.github/actions/run-package-tests
+      env:
+        ISAACLAB_RUN_CI_TESTS: "1"
+        PYTEST_ADDOPTS: "--without-kit"
+      with:
+        image-tag: ${{ needs.config.outputs.ci_image_tag }}
```
hujc7, line 620: Can this use kitless docker? https://github.com/isaac-sim/IsaacLab/pull/6355

PR #4732, opened 2026-02-25
Stringfies class-type that prevent Cfg file eager imports implementation files dependencies

ooctipus:
# Description

This PR changes all class-type value from a direct import to string import.
this make sure the implementation files dependencies are not loaded during config reading stage.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6708, opened 2026-07-24
Fix observation modifier callable resolution

### LINKED ISSUE #6067: [Bug Report] V3.0.0-Beta: Observation ModifierCfg func is not resolved from ResolvableString before signature validation
# [Bug Report] Observation ModifierCfg func is not resolved from ResolvableString before signature validation

### Descri

When using a class-based observation modifier through `ModifierCfg`, environment creation fails if the config has gone through Hydra / `from_dict()` and the modifier `func` is represented as a callable string / `ResolvableString`.

The failure happens in `ObservationManager._prepare_terms()` when validating modifier parameters. For regular observation term functions, Isaac Lab resolves string callables via `string_to_callable`, but modifier functions do not appear to receive the same resolution before `inspect.signature(mod_cfg.func)` is called.

As a result, `inspect.signature()` sees `ResolvableString.__call__(*args, **kwargs)` instead of the modifier class or modifier instance signature. This makes the manager believe the modifier expects a mandatory parameter named `kwargs`.

Observed error:

```text
ValueError: Modifier 'ModifierCfg(func='{repo}.mdp.observation:DelayedObsModifier', params={'min_delay': 1, 'max_delay': 3})' of observation term 'joint_pos' expects mandatory parameters: ['kwargs'] and optional parameters: [], but received: ['min_delay', 'max_delay'].
```

### Steps to reproduce

Use a class-based observation modifier in an observation term:

```python

import torch
from isaaclab.utils.buffers import DelayBuffer
from isaaclab.utils.modifiers import ModifierBase, ModifierCfg


class DelayedObsModifier(ModifierBase):
    def __init__(self, cfg: ModifierCfg, data_dim: tuple[int, ...], device: str) -> None:
        super().__init__(cfg, data_dim, device)
        self.max_delay = cfg.params.get("max_delay", 0)
        self.min_delay = cfg.params.get("min_delay", 0)
        self.delay_buffer = DelayBuffer(
            history_length=self.max_delay,
            batch_size=data_dim[0],
            device=device,
        )

    def reset(self, env_ids: Sequence[int] | None = None):
        self.delay_buffer.reset(env_ids)

    def __call__(self, data: torch.Tensor) -> torch.Tensor:
        return self.delay_buffer.compute(data)
```

Configure an observation term with the modifier:

```python
joint_pos = ObsTerm(
    func=mdp.joint_pos,
    modifiers=[
        ModifierCfg(
            func=DelayedObsModifier,
            params={"min_delay": 1, "max_delay": 3},
        )
    ],
)
```

Run the task through the Hydra training flow, so that the environment config is serialized and restored through `env_cfg.from_dict(hydra_cfg["env"])`.

During environment creation, the modifier config becomes similar to:

```python
ModifierCfg(
    func='my_package.my_module:DelayedObsModifier',
    params={'min_delay': 1, 'max_delay': 3},
)
```

Then `ObservationManager` fails during `_prepare_terms()` with:

```text
ValueError: Modifier 'ModifierCfg(func='my_package.my_module:DelayedObsModifier', params={'min_delay': 1, 'max_delay': 3})' of observation term 'joint_pos' expects mandatory parameters: ['kwargs'] and optional parameters: [], but received: ['min_delay', 'max_delay'].
```

A minimal root-cause check:

```python
import inspect
from isaaclab.utils.modifiers import ModifierCfg
from isaaclab.utils.string import ResolvableString

cfg = ModifierCfg(
    func=ResolvableString("my_package.my_module:DelayedObsModifier"),
    params={"min_delay": 1, "max_delay": 3},
)

print(inspect.signature(cfg.func))
# Actual: (*args, **kwargs)
```

Expected behavior: `mod_cfg.func` should be resolved to the actual modifier class before class checks and signature validation. The class-based modifier should then be instantiated, and its call signature should be `(data: torch.Tensor) -> torch.Tensor`.

### System Info

- tag: `v3.0.0-beta`
- Isaac Sim Version: 6.0
- OS: Ubuntu 24.04
- GPU: RTX 5060 mobile
- CUDA: 12.8
- GPU Driver: 595.71.05

### Additional context

The issue seems to be caused by missing callable-string resolution for observation modifiers.

In `isaaclab/managers/manager_base.py`, normal manager term functions are resolved when `term_cfg.func` is a string. However, in `isaaclab/managers/observation_manager.py`, modifier functions are checked with `inspect.isclass(mod_cfg.func)` and later `inspect.signature(mod_cfg.func)` without first resolving string / `ResolvableString` callables.

A potential fix would be to resolve `mod_cfg.func` before the class check in `ObservationManager._prepare_terms()`:

```python
from isaaclab.utils import string_to_callable

...

if isinstance(mod_cfg.func, str):
    mod_cfg.func = string_to_callable(str(mod_cfg.func))

if inspect.isclass(mod_cfg.func):
    ...
```

This would make class-based modifiers behave consistently whether the config is used directly or after Hydra / `from_dict()` processing.

### Checklist

### Acceptance Criteria

- [ ] `ModifierCfg.func` for observation modifiers is resolved from callable strings / `ResolvableString` before class checks and signature validation.
- [ ] Class-based observation modifiers work both when config objects are used directly and when configs are restored through Hydra / `from_dict()`.
- [ ] Existing function-based modifiers continue to pass parameter validation unchanged.

ecstayalive:
# Description

## Summary

Resolve class-based observation modifier callables restored from strings before class instantiation and signature validation.

When an observation configuration containing a class-based modifier is serialized with `to_dict()` and restored with `from_dict()`, the modifier callable is represented as a string-like `ResolvableString`. `ObservationManager` previously inspected this unresolved value before determining whether it represented a modifier class.

This caused signature validation to inspect `ResolvableString.__call__` instead of the actual modifier class and raise an error similar to:

```text
ValueError: Modifier ... expects mandatory parameters: ['kwargs']
```

The modifier callable is now resolved with `string_to_callable()` before the existing class detection and instantiation logic runs. Converting the value with `str()` also normalizes `ResolvableString` instances before resolution.

A regression test covers the complete configuration dictionary round-trip using a class-based `IntegratorCfg` modifier. It verifies that:

- The callable is restored as a string-like value after `to_dict()` and `from_dict()`.
- `ObservationManager` resolves and instantiates the modifier as a `ModifierBase`.
- The resolved modifier produces the expected observation output.

No new runtime dependencies are introduced.

Fixes #6067

## Type of change

## Testing

The regression test was verified to fail before the fix with the reported mandatory `kwargs` signature-validation error and pass after applying the fix.

The complete ObservationManager test file was also executed:

```text
source/isaaclab/test/managers/test_observation_manager.py

16 passed
```

The following additional checks passed:

- Ruff linting
- Ruff formatting
- Trailing-whitespace and end-of-file checks
- Codespell
- RST validation
- License-header validation
- Changelog fragment validation

The Git LFS pre-commit hook could not run locally because `git-lfs` was not installed. The modified Python and RST files were verified with `git check-attr` and are not tracked by Git LFS.

## Checklist

- [x] I have read and understood the [[contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run all pre-commit checks with `./isaaclab.sh --format`
  - All applicable checks passed, except the Git LFS hook because `git-lfs` was unavailable locally.
- [x] Documentation changes are not required for this internal bug fix.
- [x] I have added a regression test that proves the fix is effective.
- [x] I have added a changelog fragment under `source/isaaclab/changelog.d/`.
  - An extension version bump is not required for changes targeting `develop`.
- [ ] I have added my name to `CONTRIBUTORS.md`, or my name already exists there.

### DIFF
--- a/source/isaaclab/changelog.d/ecstayalive-fix-observation-modifier-resolution.rst
+++ b/source/isaaclab/changelog.d/ecstayalive-fix-observation-modifier-resolution.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed class-based observation modifiers restored from dictionaries failing signature validation before their
+  callable was resolved.
--- a/source/isaaclab/isaaclab/managers/observation_manager.py
+++ b/source/isaaclab/isaaclab/managers/observation_manager.py
@@ -15,7 +15,7 @@ import numpy as np
 import torch
 from prettytable import PrettyTable
 
-from isaaclab.utils import class_to_dict, modifiers, noise
+from isaaclab.utils import class_to_dict, modifiers, noise, string_to_callable
 from isaaclab.utils.buffers import CircularBuffer
 
 from .manager_base import ManagerBase, ManagerTermBase
@@ -581,6 +581,9 @@ class ObservationManager(ManagerBase):
                     for mod_cfg in term_cfg.modifiers:
                         # check if class modifier and initialize with observation size when adding
                         if isinstance(mod_cfg, modifiers.ModifierCfg):
+                            # resolve string-based modifiers before checking whether they are classes
+                            if isinstance(mod_cfg.func, str):
+                                mod_cfg.func = string_to_callable(str(mod_cfg.func))
                             # to list of modifiers - instantiate class-based modifiers
                             if inspect.isclass(mod_cfg.func):
                                 mod_cfg.func = mod_cfg.func(cfg=mod_cfg, data_dim=obs_dims, device=self._env.device)
--- a/source/isaaclab/test/managers/test_observation_manager.py
+++ b/source/isaaclab/test/managers/test_observation_manager.py
@@ -668,6 +668,37 @@ def test_modifier_compute(setup_env):
     assert torch.max(obs_critic["term_4"]) <= 0.5
 
 
+def test_modifier_class_from_dict(setup_env):
+    """Test class-based modifiers restored from a dictionary."""
+    env = setup_env
+    modifier = modifiers.IntegratorCfg(dt=env.dt)
+
+    @configclass
+    class MyObservationManagerCfg:
+        """Test config class for observation manager."""
+
+        @configclass
+        class PolicyCfg(ObservationGroupCfg):
+            """Test config class for policy observation group."""
+
+            concatenate_terms = False
+            term_1 = ObservationTermCfg(func=pos_w_data, modifiers=[modifier])
+
+        policy: ObservationGroupCfg = PolicyCfg()
+
+    cfg = MyObservationManagerCfg()
+    cfg.from_dict(cfg.to_dict())
+    modifier_cfg = cfg.policy.term_1.modifiers[0]
+    assert isinstance(modifier_cfg.func, str)
+
+    obs_manager = ObservationManager(cfg, env)
+    modifier_cfg = obs_manager.cfg.policy.term_1.modifiers[0]
+    assert isinstance(modifier_cfg.func, modifiers.ModifierBase)
+
+    observations = obs_manager.compute()
+    torch.testing.assert_close(observations["policy"]["term_1"], env.data.pos_w * env.dt / 2)
+
+
 def test_serialize(setup_env):
     """Test serialize call for ManagerTermBase terms."""
     env = setup_env


ooctipus:
Thanks for the PR, but I think the practice is get ResolvableString function as it should be. rather than manually converting string to callable. I will find a fix for it thanks :D

ecstayalive:
If you have better solutions, thank you very much! You can turn off this PR at any time

j3soon:
Hi, I encountered a similar issue while porting a codebase from 2.3.2 to 3.x.

Below are testcases that passes on Isaac Lab 2.3.2 but failed on both 3.0.0 beta2 and `develop`, with instructions drafted by Claude. I've ran the commands on my machine and can reproduce. May be potentially useful to validate the final fix afterwards.

<hr>

[rs_check.py](https://github.com/user-attachments/files/30748121/rs_check.py)

Save the attached `rs_check.py` and run it twice, changing only `IMAGE`. Same file both times — it starts a headless Kit app automatically on 2.x, where these imports still need `pxr`. No checkout or mounts needed. Each row's site, symptom and rationale is in the comments inside the script.

**Before #4732** — `2.3.2`:

```bash
IMAGE=nvcr.io/nvidia/isaac-lab:2.3.2

docker run --rm --gpus all -e ACCEPT_EULA=Y -e PRIVACY_CONSENT=Y -e OMNI_KIT_ACCEPT_EULA=YES \
  -e PYTHONUNBUFFERED=1 -v $(pwd)/rs_check.py:/rs_check.py \
  --entrypoint /workspace/isaaclab/_isaac_sim/python.sh \
  ${IMAGE} /rs_check.py 2>&1 | sed 's/.*\[py stderr\]: //' | grep -aE '^\[(PASS|FAIL|SKIP)\]|passed,'
```

```
[PASS] terrain repeated_boxes           generated
[PASS] terrain repeated_pyramids        generated
[PASS] terrain repeated_cylinders       generated
[PASS] issubclass(class_type)           ok
[PASS] inspect.signature(class_type)    params=['noise_model_cfg', 'num_envs', 'device']
[PASS] modifier func after from_dict    ABCMeta
6 passed, 0 failed, 0 skipped (of 6)
[PASS] non-callable field still usable  str 'sensor:tiled_camera'   (guard)
```

**After** — `3.0.0-beta2`:

```bash
IMAGE=nvcr.io/nvidia/isaac-lab:3.0.0-beta2-post1

docker run --rm --gpus all -e ACCEPT_EULA=Y -e PRIVACY_CONSENT=Y -e OMNI_KIT_ACCEPT_EULA=YES \
  -e PYTHONUNBUFFERED=1 -v $(pwd)/rs_check.py:/rs_check.py \
  --entrypoint /workspace/isaaclab/_isaac_sim/python.sh \
  ${IMAGE} /rs_check.py 2>&1 | sed 's/.*\[py stderr\]: //' | grep -aE '^\[(PASS|FAIL|SKIP)\]|passed,'
```

```
[FAIL] terrain repeated_boxes           ValueError: The attribute 'object_type' must be a string or
[FAIL] terrain repeated_pyramids        ValueError: The attribute 'object_type' must be a string or
[FAIL] terrain repeated_cylinders       ValueError: The attribute 'object_type' must be a string or
[FAIL] issubclass(class_type)           TypeError: issubclass() arg 1 must be a class
[FAIL] inspect.signature(class_type)    params=['args', 'kwargs']
[FAIL] modifier func after from_dict    ResolvableString
0 passed, 6 failed, 0 skipped (of 6)
[PASS] non-callable field still usable  ResolvableString 'sensor:tiled_camera'   (guard)
```

To check a working tree instead of the image, run from an IsaacLab checkout root and add `-v $(pwd)/source:/workspace/isaaclab/source`. The script exits non-zero if any row fails, so it can be used as a gate.

The last line is a **guard**, not a regression row: it is green on both versions and should stay green. `_CALLABLE_STR_RE` wraps any `ident(.ident)*:ident` string, including fields that were never callable references, so a fix that resolves every wrapped string eagerly breaks them — a prototype of mine did exactly that and failed there with `ValueError: Could not resolve the input string 'sensor:tiled_camera'`, and only went green once resolution was scoped to fields declared as callables.

kellyguo11:
fixed in https://github.com/isaac-sim/IsaacLab/pull/7513

PR #6710, opened 2026-07-24
Fix curriculum live plot collection

huidongc:
# Description

Curriculum states represented as dictionaries caused live plot collection to index a list with a string, producing repeated TypeError messages during Newton visualization for Franka Pour (experimental env for particles).

Append dictionary values to the term data list so curriculum metrics can be collected and plotted correctly.

## Type of change

## Checklist


mmichelis (approved):
Thanks for the fix! A bonus addition would be a small test case for all possible curriculum states that are possible, like dicts, torch.Tensors, float, and anything more? Might also be good to guard the allowed types that we can parse in the first place for the iterable terms.

kellyguo11 (approved):

PR #5867, opened 2026-05-29
[Changelog] Move the nightly lifecycle into cli.py auto-bump and sync uv.lock

hujc7:
# Description

The nightly workflow duplicates the compiler's output paths in shell. When the compiler gains another output, that staging list can miss it; version bumps also leave workspace pins in `uv.lock` behind.

This moves compilation, lock synchronization, staging, bot commits, and push retries into `cli.py auto-bump`. The workflow supplies the target branch and credentials. Files reported by compilation and lock synchronization determine what is staged, including consumed fragments and outputs outside `source/`.

## Behavior

1. Compile managed packages independently, publishing healthy packages while returning a failure for unsuccessful ones. Rollback preserves pre-compile file contents, including local edits and fragments deleted before a later failure.
2. Update only editable workspace members' version pins in `uv.lock`. Validate the complete parsed result and restrict text changes to version lines. Refuse membership changes requiring a reviewed `uv lock`; missing lockfiles remain a no-op.
3. Retry non-fast-forward pushes by fetching the target branch and rebasing the generated commit. Dry runs write nothing and do not commit or push.
4. Provide `sync-lock` and `sync-lock --check` for manual repair and drift detection. Accept dotted fragment slugs under the documented Git branch-name rules; collect Unicode PR paths without Git quoting.

## Code organization

| Module | Responsibility |
|---|---|
| `versions.py` | Version values and bump arithmetic |
| `fragments.py` | Filenames, content validation, aggregation, consumption |
| `packages.py` | Package/root manifests and changelog output |
| `checks.py` | PR diffs and fragment policy |
| `paths.py` | Shared repository layout |
| `lockfile.py` | Workspace pin inspection and repair |
| `gitrepo.py` | Git operations |
| `autobump.py` | Nightly lifecycle |
| `cli.py` | Argument parsing and command dispatch |

Existing class imports from `packages` and Git imports from `autobump` remain available. The tool requires only the Python standard library; no runtime dependencies were added. No package versions or lockfile contents are changed by this PR.

## Validation

1. `uv run --no-project --with pytest python -m pytest tools/changelog/ -q` — **176 passed**. Consolidated overlapping tests from the previous 184-case suite and extended existing cases for distinct regressions.
2. Five self-review regression cases failed before their fixes and passed afterward: Unicode PR paths, partial fragment/skip deletion, preservation of uncommitted edits, unexpected lock metadata changes, and a registry block sharing an editable member's name and version.
[...]

PR #6727, opened 2026-07-26
Add nightly bump for uv lock

StafaH:
# Description

Bump the uv lock when bumping nightly version to avoid mismatch.

## Checklist


hujc7:
Correct me if I am wrong. My understanding is that nightly job bumps version pyproject and left uvlock now, so the only purpose of to match the version in lock to pyproject. Three main comments:
1. It might be preferred to model the utility into a class altering uvlock file, which will make it easier to maintain and expand if ever needed
2. As the version bump of pyproject happens in changelog/cli, it might be preferred to auto call this as part of the changelog workflow instead of a separate step
3. the workflow part is only triggered from the default(main) branch, so we will need to cherry-pick the workflow changes there. And make it account for the situations things might be different on branches

I do have a changelog refactor PR pending which i have not got time to finish. But I think to have default branch only carry the trigger, and have cli part handles all the logic including committing files is more scalable as that could avoid a change to be needed in multiple places. https://github.com/isaac-sim/IsaacLab/pull/5867

PR #6729, opened 2026-07-26
Remove obsolete de-instancing workaround in test and fix rendering test failures due to reuse of `SimpleShadingSD`

huidongc:
# Description

* Fixed :class:`~isaaclab_physx.renderers.IsaacRtxRenderer` failing with  ``Annotator SimpleShadingSD is not attached to any render products`` when  multiple environments sequentially request ``simple_shading_*`` camera  outputs in the same Kit process. Each tiled render product is named from the requested data types, and the owned HydraTexture is destroyed on cleanup to avoid leaking render products across env create/destroy cycles.

* Remove the NVBUG#6418121 workaround from Franka cloth and soft-body rendering tests now that the issue has been fixed by upstream Kit components.

## Type of change

- Test change

## Checklist


source/isaaclab_tasks/test/golden_images/franka_cloth/newton-isaacsim_rtx_renderer-rgb.png
huidongc, line 1: Blessed the new render output which shows diff in lighting and shadow.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -95,7 +95,7 @@ class IsaacRtxRenderData:
     """Render data for Isaac RTX renderer."""
 
     annotators: dict[str, Any]
-    render_product_paths: list[str]
+    render_product: Any
```
pbarejko: Did you get lazy on type safety? ;)

It's `viewport_manager.HydraTexture` defined in Replicator's `utils` sub-module.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -591,9 +597,21 @@ def read_output(self, render_data: IsaacRtxRenderData, camera_data: CameraData)
...
 
     def cleanup(self, render_data: IsaacRtxRenderData | None):
-        """Detach annotators from render product.
+        """Detach annotators, destroy the owned tiled render product, and drop held refs.
         See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.cleanup`."""
-        if render_data:
+        if render_data is None:
+            return
+
+        if render_data.render_product is not None:
```
pbarejko: Is there some situation where render product could be `None`? You follow RAII so I would say no, it can not be `None`.

The  `create.render_product_tiled` returns `viewport_manager.HydraTexture` and never `None`.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -591,9 +597,21 @@ def read_output(self, render_data: IsaacRtxRenderData, camera_data: CameraData)
...
             for annotator in render_data.annotators.values():
-                annotator.detach(render_data.render_product_paths)
-            render_data.spec = None
+                annotator.detach([render_data.render_product.path])
+
+            render_data.render_product.destroy()
```
pbarejko: nit: You could do that, or just assigning `None` and wait for garbage collect to call `destroy` for you.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -293,9 +293,15 @@ def create_render_data(self, spec: CameraRenderSpec) -> IsaacRtxRenderData:
...
-        rp = rep.create.render_product_tiled(cameras=cam_prim_paths, tile_resolution=(spec.cfg.width, spec.cfg.height))
-        render_product_paths = [rp.path]
+        # Create replicator tiled render product, using a unique name from requested outputs so
+        # sequential envs in one Kit process (e.g. simple_shading_* pytest cases) do not reuse
+        # a stale Replicator / SyntheticData activation. Prefer the public data-type names: all
+        # simple_shading_* modes share the same underlying AOV (``SimpleShadingSD``).
+        rp = rep.create.render_product_tiled(
+            cameras=cam_prim_paths,
+            tile_resolution=(spec.cfg.width, spec.cfg.height),
+            name="__".join(spec.cfg.data_types),
```
pbarejko: Why do we use dunder for this? Do we need a test case covering this behavior? 

Also this seems to be fragile. Example: you have `env0/cam0` `env0/cam1` both define same rgb data type, there will be a name clash.

huidongc: > Why do we use dunder for this? 

I need to use dunder to differentiate with the single underscore used in aov names, e.g. "motion_vectors", "distance_to_camera".

> Do we need a test case covering this behavior?

Good point.

> Also this seems to be fragile. Example: you have env0/cam0 env0/cam1 both define same rgb data type, there will be a name clash.

We don't create separate render products for each camera, we create a tiled render product for all the cameras.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
huidongc: I prefer explicit cleanup rather than leaving it to gc. When we hit issue in GC it is more difficult to debug than having this explicit cleanup.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
pbarejko: > We don't create separate render products for each camera, we create a tiled render product for all the cameras.

In the near future there might be two tiled cameras! We could group them as follows:
* all cameras `env_*/cam0` as first render product
* all cameras  `env_*/cam1` as second render product

In this case if both have data type as rgb that would be a conflict! On top of that we could trigger them with different tick rates.

huidongc: Makes sense then. I misread your earlier comment and thought you were talking about cameras in the same camera group.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -591,9 +597,21 @@ def read_output(self, render_data: IsaacRtxRenderData, camera_data: CameraData)
...
+        if render_data.render_product is not None:
             for annotator in render_data.annotators.values():
-                annotator.detach(render_data.render_product_paths)
-            render_data.spec = None
+                annotator.detach([render_data.render_product.path])
+
+            render_data.render_product.destroy()
+            render_data.render_product = None
+
+        render_data.annotators.clear()
```
rilei-nvidia, line 615: should this set of render_data clean-up procedures be implemented as a method on the IsaacRtxRenderData class itself?

rilei-nvidia (approved):
Had a minor comment, Piotr covered most of them.  Preemptive LGTM

pbarejko (approved):

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
huidongc: No. Keep IsaacRtxRenderData as a passive state bag; cleanup belongs on the renderer.

That matches the BaseRenderer contract: Camera calls self._renderer.cleanup(self._render_data), not a method on the data object. Sibling backends (OVRTXRenderData, Newton RenderData) follow the same split — create/teardown live on the renderer.

rilei-nvidia (approved):

kellyguo11 (approved):

PR #6733, opened 2026-07-27
Expose VBD rigid contact controls

ruziniuuuuu:
# Description

Expose the following `SolverVBD` rigid-contact stability controls through `VBDSolverCfg`:

- `rigid_contact_hard`
- `rigid_contact_history`
- `rigid_avbd_contact_alpha`
- `rigid_avbd_beta`
- `rigid_body_contact_buffer_size`

The configuration documentation now explains the interaction between `rigid_avbd_beta` and `rigid_contact_k_start`, including a non-zero scale recommendation and the shared linear/angular units.

This PR intentionally contains only the VBD configuration changes requested in review. The inverse-kinematics generalized-coordinate synchronization and collision contact-matching changes have been removed for separate follow-up work.

## Validation

- `uv run --frozen python -m pytest -q source/isaaclab/test/deps/newton/test_deformable_builder_hooks.py` — 7 passed
- `uv run --frozen isaaclab -f` — passed
- changelog fragment check and dry-run compile — passed


source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -79,6 +91,15 @@ def _build_solver(cls, model: Model, solver_cfg: CouplerCfg) -> None:
 
         cls._validate_config(solver_cfg)
         resolved_entries = [cls._resolve_entry(model, entry) for entry in solver_cfg.entries]
+        vbd_entries = [entry for entry in resolved_entries if isinstance(entry.config.solver_cfg, VBDSolverCfg)]
```
mmichelis: very specific VBD addition in the coupler, ideally we can decouple this and have a manager specific call that handles the IK masking (since it isn't VBD specific, but Kamino might also want/need this)

mmichelis (commented):
Thanks for the additions for the VBD solver for rigid bodies! The config additions would absolutely be helpful to add, but I think the IK mask can be a separate PR, and will likely need a bit more generalization. We would also want more understanding from the Newton side how this will be handled moving forward, perhaps you could simplify this PR down to just the VBD config changes so we have better rigid solver control?

source/isaaclab_contrib/isaaclab_contrib/deformable/newton_manager_cfg.py
```diff
@@ -117,6 +117,30 @@ class VBDSolverCfg(NewtonModelSolverCfg):
...
+    rigid_avbd_contact_alpha: float | None = None
+    """Body-contact C0 stabilization strength in ``[0, 1]``.
+
+    ``None`` uses SolverVBD's shared ``rigid_avbd_alpha`` default. Lower values
+    correct more current penetration and strengthen contacts when history is
+    disabled or the iteration count is modest.
+    """
+
+    rigid_avbd_beta: float = 0.0
+    """Penalty ramp rate per rigid-body AVBD iteration; zero disables ramping."""
```
mmichelis: **`rigid_avbd_beta` docstring omits the `rigid_contact_k_start` interaction, and it is the direction that matters.**

Newton documents the dependency on the other field: "`rigid_contact_k_start`: Body-body and body-particle contact penalty seed for AVBD ramping. Used when ``rigid_avbd_linear_beta`` (or ``rigid_avbd_beta`` fallback) is greater than zero. When the linear beta is 0, k is fixed at the contact stiffness regardless of this value." (`newton/_src/solvers/vbd/solver_vbd.py:358-360` at the pinned commit `d7581b7`).

Since the default added here is `rigid_avbd_beta = 0.0`, the pre-existing `rigid_contact_k_start` field on line 117 is a no-op by default, and its docstring ("Initial stiffness seed for all rigid body contacts [N/m]") is now actively misleading: this PR exposes the field that gates it without cross-referencing in either direction. A user reading line 117 in isolation will tune a value that does nothing.

Two smaller things dropped from Newton's own docs for this field:

- Linear (meters) and angular (radians) constraints have different units, so Newton recommends the `rigid_avbd_linear_beta` / `rigid_avbd_angular_beta` overrides for production tuning. Worth stating even though those overrides are not exposed on this configclass.
- A concrete non-zero magnitude. Newton suggests e.g. `1e5`; `0.0` alone gives no scale hint.

Also, sibling fields in this class consistently carry unit brackets (`[N/m]`, `[N*s/m]`, `[m]`); this one has none.

```suggestion
    rigid_avbd_beta: float = 0.0
    """Penalty ramp rate per rigid-body AVBD iteration; zero disables ramping.

    :attr:`rigid_contact_k_start` only takes effect when this is greater than
    zero; at zero the penalty stiffness is fixed at the contact stiffness.
    Set to e.g. ``1e5`` to enable ramping. Shared by linear [m] and angular
    [rad] constraints, which have different units.
    """
```

PR #6751, opened 2026-07-27
Startup Optimization: Reduce bottlenecks in loop handling for cloning, garbage collection, asset loading

StafaH:
# Speed up environment creation at high env counts

## Description

This PR removes that overhead. There are five independent changes, none of which alter
simulation results:

* **Skip USD replication when Kit is absent** (`cloner/replicate_session.py`).
  `UsdReplicateContext` authors one full copy spec of the source subtree per environment.
  Those prims exist so Kit can compose and render them — kitless backends read the physics
  model instead and never touch them. The context is now dropped when `has_kit()` is false,
  mirroring the spawner condition that was already Kit-gated. This is the single largest win.

* **Vectorize per-world transform composition** (`isaaclab_newton/cloner/newton_clone_utils.py`).
  World transforms were composed with Warp's scalar `wp.transform_multiply` /
  `wp.transform_inverse` builtins, which dispatch through `inspect.signature` binding at tens
  of microseconds per call — once per world per source. The composition is plain quaternion
  algebra, so it is now done in NumPy across all worlds at once. Also replaces per-world
  tensor indexing with a single bulk device-to-host transfer.

* **Cheapen label renaming** (`isaaclab_newton/cloner/newton_clone_utils.py`).
  `rename_builder_labels` called the generic path-prefix helper for every label, re-stripping
  both prefixes each time. It now pre-normalizes destination roots once per source and tests
  the cheapest discriminator first, since most labels belong to a different source.

* **Keep builder shape colors as plain tuples** (`sim/utils/newton_model_utils.py`).
  Writing `wp.vec3` into `builder.shape_color` dropped `wp.array(shape_color, ...)` off
  NumPy's fast conversion path in `finalize`. Since replication copies these entries into
  every world, a single object type using `wp.vec3` slowed down the whole array.

* **Reduce asset I/O on warm caches** (`utils/assets.py`).
  `read_file` re-read from the Nucleus server on every call, re-downloading payloads
  (e.g. actuator networks) that `retrieve_file_path` had already cached; it now reads the local
  mirror. `check_file_path` likewise skips its network status probe when the mirror already
  exists, which also removes a startup dependency on server reachability for fully-cached assets.

* **Scope the post-graph-capture collection to gen 0** (`isaaclab_newton/physics/newton_manager.py`).
  Nothing is promoted while the collector is paused, so all deferred graph-scoped garbage is
  still in generation 0. A full `gc.collect()` instead walked the entire live heap — hundreds of
  milliseconds once the replicated model exists — to reach older objects the periodic collector
  handles anyway.

## Type of change

- Performance improvement (non-breaking change which improves efficiency)

## Measurements

`uv run isaaclab benchmark startup --task <task> --num_envs 4096 --seed 0 presets=newton_mjwarp`,
headless (no Kit), AMD Ryzen Threadripper PRO 7965WX / 1 GPU, CUDA 12.8.

Medians of 2 alternating warm-cache runs per configuration; run-to-run spread was under 0.4 s.
Baseline and change runs were interleaved, and asset caches were warm for both, so the numbers
below exclude first-run asset download.

### `env_creation` phase (`gym.make` + `env.reset()`)

| Task | before | after | saved | speedup |
|---|---|---|---|---|
| `Isaac-Cartpole-Direct` | 7.72 s | 6.73 s | 0.98 s | 1.15× |
| `Isaac-Open-Drawer-Franka` | 14.23 s | 12.46 s | 1.78 s | 1.14× |
| `Isaac-Velocity-Rough-AnymalD` | 17.28 s | 12.32 s | 4.96 s | **1.40×** |

### End-to-end benchmark process wall clock

| Task | before | after | saved | speedup |
|---|---|---|---|---|
| `Isaac-Cartpole-Direct` | 12.5 s | 11.8 s | 0.7 s | 1.06× |
| `Isaac-Open-Drawer-Franka` | 19.4 s | 17.6 s | 1.8 s | 1.10× |
| `Isaac-Velocity-Rough-AnymalD` | 22.6 s | 17.5 s | 5.1 s | **1.29×** |

The other profiled phases (`python_imports`, `task_config`, `app_launch`, `first_step`) are
unchanged within noise, as expected — every change is in the scene-construction path.

### Per-change attribution (`env_creation`, seconds saved)

Measured by reverting one change at a time from the full set:

| Change | `Isaac-Velocity-Rough-AnymalD` | `Isaac-Cartpole-Direct` |
|---|---|---|
| Skip USD replication without Kit | 3.93 s | 0.09 s |
| Warm-cache asset I/O | 0.26 s | — |
| Builder / label / color / gc changes | ~0.77 s | ~0.89 s |

The USD replication skip scales with the prim count of the source environment, so it dominates
for asset-heavy scenes (ANYmal + rough terrain) and is negligible for `Isaac-Cartpole-Direct`,
whose source subtree is small. The remaining changes are roughly scene-independent and are what
Cartpole's gain is made of.

### Caveats

* The startup benchmark runs each phase under `cProfile`, so absolute times are inflated
  relative to a plain run. The relative comparison is what this table is measuring — both
  configurations carry identical profiling overhead.
* `Isaac-Lift-Cube-Franka` was excluded: it registers no Newton preset and requires Isaac Sim,
  which is not installed in this environment.

## Companion Newton change

These measurements were taken against a local Newton checkout carrying an additional
`Improve perf for large batched worlds` commit (builder, terrain generator, and hydroelastic SDF
paths), on top of the currently pinned `d7581b73`. That work ships separately.

**`pyproject.toml` in this branch points `newton` at a local path
(`/home/mhaiderbhai/code/mustafa_newton`) and must be repinned to a real commit before merge.**
The `uv.lock` change is a consequence of that local pin and should be regenerated at the same time.

## Checklist

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `uv run isaaclab -f`

## Risk notes for review

The Kit gate in `replicate_session.py` is the only change with a behavioral condition attached:
it changes what USD prims exist in kitless runs. Anything that reads replicated USD prims
without Kit — rather than reading the physics model — would regress. Worth a careful look from
someone who knows the kitless USD consumers.


source/isaaclab/isaaclab/cloner/replicate_session.py
```diff
@@ -87,8 +89,10 @@ def replicate(plan: ClonePlan, *, stage: Usd.Stage, replicate_physics: bool = Tr
             contexts = [string_to_callable(c) if isinstance(c, str) else c for c in cfg.cloning_contexts]
         if not replicate_physics:
             contexts = [c for c in contexts if c is UsdReplicateContext]
+        if not kit_available:
```
ooctipus: remember to fix this

ooctipus (approved):

kellyguo11 (approved):

PR #6193, opened 2026-06-15
Fix Newton renderer simulation context ownership

ooctipus:
## Summary
- read Newton shadow-model clone plans from the active SimulationContext instead of PhysicsManager._sim
- keep PhysicsManager.close() from clearing shared lifecycle state when a non-active manager class is closing
- add regression coverage for PhysX + Newton renderer initialization with PhysicsManager._sim unset

## Validation
- PYTHONPATH="$PWD/source/isaaclab:$PWD/source/isaaclab_newton:$PWD/source/isaaclab_physx:$PWD/source/isaaclab_assets:$PWD/source/isaaclab_tasks:$PWD/source/isaaclab_mimic:$PWD/source/isaaclab_rl:$PWD/source/isaaclab_visualizers" ../develop/.venv/bin/python -m pytest source/isaaclab/test/sim/test_newton_manager_visualization_state.py
- pre-commit run --files source/isaaclab/isaaclab/physics/physics_manager.py source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py source/isaaclab/test/sim/test_newton_manager_visualization_state.py source/isaaclab/changelog.d/fix-physx-newton-renderer-sim-binding.rst source/isaaclab_newton/changelog.d/fix-physx-newton-renderer-sim-binding.rst
- python3 tools/changelog/cli.py check changelog-check-develop (temporary local origin/changelog-check-develop ref pointed at upstream/develop)

PR #6759, opened 2026-07-28
[BUG] Dispatch PhysicsEvent.STOP when the physics manager is declared lazily

hujc7:
# Description

Fixes a `SIGSEGV` on shutdown after training a camera task with the RTX renderer (nvbug 6492483).

Not a rendering bug: `PhysicsEvent.STOP` is never dispatched, so **no sensor or asset is ever
told to release its resources**.

## Root cause

A physics config may declare its manager lazily (`class_type = "{DIR}.mjwarp_manager:NewtonMJWarpManager"`)
so that reading the config does not import the backend. `configclass` wraps that as a
`ResolvableString` — a `str` subclass that proxies attribute access. `initialize(...)` therefore
works, but the shutdown gate is an identity check:

```python
is_active_manager = sim is not None and sim.physics_manager is cls   # str is class -> always False
if is_active_manager:
    cls.dispatch_event(PhysicsEvent.STOP)                            # never runs
```

The guard exists to stop a *non-active* manager from clearing shared state; written with `is`,
"skip when non-active" became "skip always" for every backend (since #6193, 2026-06-15). The
failure is silent by construction — skipping every listener looks exactly like having none.
PhysX was masked because `physx_manager.py:70` maps `PhysicsEvent.STOP → IsaacEvents.TIMELINE_STOP`,
so its listeners kept receiving `STOP` through Kit's timeline.

`Camera._invalidate_initialize_callback` is registered on that event, so render products stayed
registered at stage teardown: 8 leaked UJITSO processors, `pipelineLayouts: 151`,
`descriptorSetLayouts: 119`, then `-11`. Assets subscribe to the same event, so their
invalidation was skipped too.

## The fix

`PhysicsManager.close` compares against the lazily declared form as well as the class, so the
dispatch no longer depends on how a caller stored the value.

Restoring it exposed teardown code that had not been executing. Also fixed here:

- **`OVRTXRenderer.cleanup(render_data)` ignored its argument** and tore down state shared by
  every camera on that backend. It now releases what `OVRTXRenderData` owns, matching
  `BaseRenderer.cleanup` and both sibling renderers.
- **Renderer-owned state had nowhere to be released from.** A renderer is shared by every camera
  whose config resolves to it, so its lifetime belongs to `RenderContext`, which had no teardown
  counterpart to `get_renderer`. `__del__` is not one — it runs under the collector, where the
  ovrtx objects it touches may already be gone. New `BaseRenderer.close()` (no-op by default, so
  other backends and out-of-tree subclasses are unaffected) and `RenderContext.close()`, called
  from `clear_instance()` after `STOP` and before `close_stage()`, while the stage is still alive.
- **Two Newton RTX goldens captured leaked state.** Render products are now released at teardown
  rather than at collection, so each environment renders from its own. Both recaptured.

## Scope

145 config fields use the lazy `"{DIR}..."` form. Exactly **one** identity comparison exists
against them, the one fixed here; every other is `is None` / `is not None`, which behaves
correctly on a `str`.

Fixes # (not provided -- tracked internally as nvbug 6492483)

## Type of change

## Testing

Reproduced on Isaac Sim `6.1.0rc1+release.44589`, single RTX A6000, driver 570.211.01:

```
./isaaclab.sh train --rl_library skrl --task Isaac-Cartpole-Camera-Direct \
  renderer=rtx physics=newton_mjwarp
```

| | Baseline | With this fix |
|---|---|---|
| exit code | `-11` | `0` |
| leaked UJITSO processors | 8 | 0 |
| `pipelineLayouts` / `descriptorSetLayouts` | 151 / 119 | none |
| `GeometryStreaming` at shutdown | present | none |

The bug requires a **non-headless** run: `--headless` never creates the viewport render path.

Regression tests, each verified to fail without its fix:

- `test_simulation_context.py::test_stop_is_dispatched_for_lazy_class_type` — `2 passed`
  (physx + newton); `2 failed` with the guard reverted.
- `test_ovrtx_renderer_contract.py` — **22 passed**; the cleanup and close release cases fail
  against their previous implementations.
- `test_simulation_render_context.py` — **11 passed**; 2 fail with the `close()` dispatch removed.
- Kitless `dexsuite_kuka_homo` `legacy-ovphysx-ovrtx`: **12/12**; without the fix the 4th test
  fails and the process dies.
- `test_rendering_lift_kuka_hetero_kitless.py` is identical with and without the renderer close,
  and all 54 environment teardowns complete without the process dying.


hujc7:
Ready for review to get bot comment

hujc7:
## Reworked: the fix is now the storage site only

The previous revision changed two files. This one changes `simulation_context.py` alone —
`physics_manager.py` is back to `develop` verbatim.

### Root cause

Two independent defects, and the crash requires **both**:

**① The lazy `class_type` is stored unresolved.** `SimulationContext.__init__` assigns the
physics config's `"module:Class"` value to `self.physics_manager`, annotated
`type[PhysicsManager]`. That value is a `ResolvableString` — a `str` subclass that proxies
`__call__`/`__getattr__`, so it satisfies every *call* the codebase makes on it while failing
every *identity* comparison. `PhysicsManager.close()` gates the repository's only
`dispatch_event(PhysicsEvent.STOP)` on exactly such a comparison, so `STOP` has never been
delivered on any backend since that gate was added in #6193. Sensors and assets both
invalidate on `STOP`, so camera render products stayed registered at stage teardown.

**② Nothing else releases the render product.** `renderer.cleanup(render_data)` has exactly two
call sites, both inside `Camera`. `RenderContext` has no `close()`/`__del__` and
`clear_instance()` never touches it; `InteractiveScene` and `SensorBase` have no `close()`.

Because the crash is the conjunction of ① and ②, negating either suppresses it — which is why
fixes at four different layers each produced a clean exit. ① is what *regressed*; ② is a design
gap that predates it and was masked by the unconditional dispatch #6193 replaced.

This PR fixes ① only.

### Correction to the previous description

The earlier commit message claimed PhysX was unaffected because it declares its manager as a
class. **That is wrong** — `PhysxCfg().class_type` is a `ResolvableString` too, and the gate is
equally dead there. PhysX does not crash because it maps `STOP` onto a Kit timeline
subscription created at registration time (`physx_manager.py:70`), which `sim.stop()` fires
independently of the dispatch. Newton and OvPhysx override neither `stop()` nor
`_subscribe_to_event`, so they have no such channel.

### Why the storage site and not `close()`

`PhysicsManager.close()` has one production caller, `simulation_context.py:836`, and it
dispatches *through* the very attribute the gate compares. Once the stored value is a real
class the predicate is unconditionally true there, so a change in `close()` would be dead code.
Fixing the storage site also repairs the attribute for its ~25 other consumers, which today
work only because they happen to call methods. Resolving costs nothing: the next line
(`self.physics_manager.initialize(self)`) forces the same import anyway.

### Validation

| | Result |
|---|---|
| Repro, 6 baseline runs (no fix) | exit `-11`, 8 leaked processors, 151/119 layouts |
| Repro, 8 runs with fix | exit `0`, zero leaks |
| Repro with fix, PhysX backend | exit `0`, zero leaks |
| Regression test with fix | passes |
| Regression test without fix | fails on `isinstance(sim.physics_manager, type)` |

Direct measurement of the event itself, patching only this file inside the released container
so every other version relationship is unchanged:

| | `sim.physics_manager` | `isinstance(…, type)` | `PhysicsEvent.STOP` delivered |
|---|---|---|---|
| without fix | `ResolvableString` | `False` | **0×** |
| with fix | the class | `True` | **1×** |

Measured on PhysX, which confirms the gate is dead there too — PhysX is spared the crash by its
timeline channel, not by the gate working.

Baselines and fixed runs were interleaved over five hours, so a warm shader cache does not
explain the difference.

### Blocker: re-arming `STOP` hangs Newton environment teardown

`isaaclab_tasks [3/3]` times out on this branch. `test_environments_newton.py` reaches 41/70 and
then spins at 138 % CPU for 73 minutes on
`Isaac-Reorient-Cube-Shadow-OpenAI-LSTM-Direct-1-cuda` before the 5000 s cap. The same test
passes at 60 % on both #6752 and #6756, which branch from the same base and complete 70/70 in
~912 s — and which do **not** cause `STOP` to dispatch.

The mechanism is verified in source, though the specific failing handler is not yet identified.
`PhysicsManager.safe_callback_invoke` swallows `ReferenceError` but re-raises anything else
unless the manager defines `store_callback_exception`, which only `PhysxManager` does. On Newton
an exception therefore propagates out of `dispatch_event` -> `close()` and aborts
`clear_instance()` before visualizer close, `_services.close_all()`, `close_stage()` and
`cls._instance = None` — leaving a live, partially torn-down singleton for the next environment
in the same process.

Ordering compounds it. `env.close()` runs before `clear_instance()`, so on Newton `STOP` now
arrives *after* the scene has been dropped. PhysX never had that exposure: its timeline channel
delivers `STOP` during `sim.stop()`, while the scene is still intact. Every handler on this path
is being invoked in a lifecycle state its author never exercised.

It applies to any fix that re-arms the gate, not just this one.

It does **not** reproduce locally. With the fix applied, the full
`test_environments_newton.py` runs **116 passed / 0 failed in 962 s** on an RTX A6000 against
`b634245` with isaacsim 6.1.0rc1 — including
`Isaac-Reorient-Cube-Shadow-OpenAI-LSTM-Direct-1-cuda`, the test CI stalls on. So the trigger
needs something that lane lacks: newer code than `b634245`, or the CI machine, or the full CI
matrix. Causation in CI is nevertheless tight — the same test at the same position passes on
both control PRs and stalls only where `STOP` dispatches.

### Known gaps, not addressed here

- Defect ② is untouched. The same crash remains reachable from scripts that never call
  `clear_instance()` (`scripts/tutorials/04_sensors/run_usd_camera.py`,
  `scripts/demos/sensors/ppisp_camera.py`), because `STOP` never fires there at all.
- `IsaacRtxRenderer.cleanup()` only detaches annotators; the render product handle itself is
  discarded at creation, so it is still handed to Kit undestroyed.
- `sim.stop()` still releases nothing on Newton/OvPhysx, so a mid-run stop/play cycle leaves
  assets and sensors stale on those backends.
- Nothing in the repository asserts that `PhysicsEvent.STOP` is ever dispatched. #6193's own
  regression test binds a real class to `physics_manager` — the one input shape production
  cannot produce — which is why the gate could be dead for six weeks and stay green.

source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -189,7 +189,12 @@ def __init__(self, cfg: SimulationCfg | None = None):
             self.cfg.physics = self._physics
         self._physics = _resolve_auto_physx_cfg(self._physics, use_isaac_sim=has_kit())
         self.cfg.physics = self._physics
-        self.physics_manager: type[PhysicsManager] = self._physics.class_type
+        # A lazily declared ``"module:Class"`` config value proxies attribute access but is a
+        # ``str``, so resolve it here to keep identity checks against the class working.
+        class_type = self._physics.class_type
+        if isinstance(class_type, str):
+            class_type = string_to_callable(str(class_type))
```
ooctipus: this doesn't seem right, the resolable string should automatically resolve it and doesn't need manual resolution.
I would try to understand why it didn't auto resolve.

kellyguo11 (approved):

ooctipus (approved):

kellyguo11:
not sure why the rendering correctness tests are failing here

kellyguo11:
I think the rendering correctness failures are likely related to the change. the tests fail consistently

hujc7:
> I think the rendering correctness failures are likely related to the change. the tests fail consistently

it looks like the renderer data was released in other ways as physics event is not sent. so fixing the physics event will need renderer side to drop the previous fixes. Continue to look into ti.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1413,50 +1413,6 @@ def _render_legacy(self, render_data: OVRTXRenderData) -> None:
                 render_data.warp_buffers[str(RenderBufferKind.RGBA)],
             )
 
-    def _cleanup_legacy(self, render_data: OVRTXRenderData | None) -> None:
```
kellyguo11, line 1419: @rilei-nvidia do you mind reviewing the changes in this file?

AntoineRichard (approved):

AntoineRichard:
#6074 runs into similar problems. It cleaned-up the lifecycle of OVPhysX + OVRTX which is partly why it fails its rendering tests.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1413,50 +1413,6 @@ def _render_legacy(self, render_data: OVRTXRenderData) -> None:
...
-        def _safe_unbind(binding, name: str) -> None:
-            if binding is None:
-                return
-            try:
-                binding.unbind()
-            except Exception as e:
-                if "destroyed" not in str(e).lower():
-                    logger.warning("Error unbinding %s: %s", name, e)
-
-        _safe_unbind(self._camera_xform_binding, "camera transforms")
```
pbarejko, line 1432: Why we got rid of ovrtx unbinding? Is this intentional?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1413,50 +1413,6 @@ def _render_legacy(self, render_data: OVRTXRenderData) -> None:
...
-        self._camera_xform_binding = None
-        _safe_unbind(self._object_xform_binding, "object transforms")
-        self._object_xform_binding = None
-        _safe_unbind(self._deformable_points_binding, "deformable points")
-        self._deformable_points_binding = None
-        _safe_unbind(self._particle_points_binding, "particle points")
-        self._particle_points_binding = None
-
-        self._deformable_particle_offsets = []
```
pbarejko, line 1441: Why we got rid of internal variable reset? Is this intentional?

source/isaaclab_tasks/test/golden_images/franka_cloth/newton-isaacsim_rtx_renderer-motion_vectors.png
pbarejko, line 1: You probably don't need to regenerate them? Those are marked as xfail probably?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1522,11 +1478,17 @@ def render(self, render_data: OVRTXRenderData) -> None:
...
+
+        The stage queries, tensor bindings and render products this renderer holds are shared by
+        every camera that resolves to it, so they are not released here; ovrtx frees them with the
+        renderer once the last reference to it goes away.
+        """
+        if render_data is None:
+            return
+        render_data.warp_buffers.clear()
+        render_data.renderer_info.clear()
+        render_data.ppisp_pipeline = None
```
rilei-nvidia, line 1538: Clearing the render data is fine.  But we should also clean up the ovrtx and ovstage internal state?  I don't think we want to rely on python garbage collection we could run into issues with the destruction order

hujc7: Added renderer close

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
hujc7: Added back into `close`.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
hujc7: Added back

source/isaaclab_tasks/test/golden_images/franka_cloth/newton-isaacsim_rtx_renderer-motion_vectors.png
hujc7: it seems they are running.

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -162,3 +162,22 @@ def reset_stage_prepare_flag(self) -> None:
...
+        Called from :meth:`~isaaclab.sim.simulation_context.SimulationContext.clear_instance` after
+        cameras have released their render data and before the stage is torn down, so
+        :meth:`BaseRenderer.close` runs while the stage is still alive. A backend that raises is
+        logged and skipped so the remaining backends still close. Idempotent.
+        """
+        for _cfg, renderer in self._renderer_entries:
+            try:
+                renderer.close()
+            except Exception:
+                logger.exception("Error closing renderer %s.", type(renderer).__name__)
```
rilei-nvidia: If a renderer throws an exception on close it feels like a real error in the code that we should fix and address.  My worry about only logging it is it will be ignored.

rilei-nvidia (changes requested):
@hujc7 thanks for extending the base renderer and hooking up the .close() - now the teardown of the renderer internal state is logically much more clear.  Just had one comment about converting the captured exception into a log.

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -162,3 +162,22 @@ def reset_stage_prepare_flag(self) -> None:
     def reset_scene_state_cadence(self) -> None:
         """Clear per-step scene state update dedupe (e.g. a long pause with no physics)."""
         self._last_scene_state_step = None
+
+    def close(self) -> None:
```
pbarejko, line 166: Naming is hard. The `close` name isn't best choice in my option. In my mind close is more like closing a socket/connection/file/stream etc. I would probably suggest `teardown`, `destroy`. @rilei-nvidia what do you think?

source/isaaclab/isaaclab/renderers/render_context.py
pbarejko: I agree, we should leave exceptions to be thrown.

source/isaaclab/isaaclab/renderers/render_context.py
rilei-nvidia: `.destroy()` would be a stronger name at face value.  ovrtx & ovstage and CUDA API also uses "destroy" so that's another plus for consistency amongst NVIDIA products.  I can see why @hujc7 originally chose close as PhysicsManager uses `.close()`.  

Given these are new methods we are introducing I'd be in favor of renaming both `RenderContext.close` and `BaseRenderer.close` to `.destroy()` - and as a follow up PR update PhysicsManager to destroy as well for consistency.

source/isaaclab_tasks/test/golden_images/franka_cloth/newton-isaacsim_rtx_renderer-motion_vectors.png
pbarejko: They are, but I am quite confident those are allow to fail @nvsekkin ?

source/isaaclab/isaaclab/renderers/render_context.py
hujc7: Updated

source/isaaclab/isaaclab/renderers/render_context.py
hujc7: I think `close` is a more used form in similar situations such that trading the sightly naming diff for a more consistent pattern would make it more obvious that it's the function to call at the desired place.

rilei-nvidia: @hujc7 can you describe / provide examples of the common situations where .close() is preferred and used?

source/isaaclab/isaaclab/renderers/render_context.py
```diff
@@ -162,3 +162,28 @@ def reset_stage_prepare_flag(self) -> None:
...
+        regardless of failures, and the caller re-raises once teardown is complete. Idempotent.
+
+        Args:
+            caught_exceptions: A list to which any exceptions raised by backend
+                :meth:`BaseRenderer.close` calls are appended.
+        """
+        for _cfg, renderer in self._renderer_entries:
+            try:
+                renderer.close()
+            except Exception as exc:  # noqa: BLE001 - reported by the caller after teardown finishes
```
rilei-nvidia: While this is certainly one way to do it - why not just not handle the exception and let it raise?  Is there a concrete use case where a caller wants to collect all the exceptions?

source/isaaclab/isaaclab/renderers/render_context.py
hujc7: Looking at what the codebase already does, the split isn't by resource kind — it's by **who owns the object**.

**`close()` is what Isaac Lab defines on its own types**, released by their owner at a chosen moment. 31 classes today; `teardown` and `shutdown` appear on none. Four ABCs declare it, three of which are the same shape as this one — an abstract base with swappable backends, closed by whoever owns it:

| Base | Backends overriding `close()` |
|---|---|
| `PhysicsManager` | `PhysxManager`, `OvPhysxManager`, `NewtonManager` |
| `BaseVisualizer` | `KitVisualizer`, `NewtonVisualizer`, `RerunVisualizer`, `ViserVisualizer` |
| `DatasetFileHandlerBase` | `HDF5DatasetFileHandler` |
| `BaseRenderer` | `OVRTXRenderer` *(this PR)* |

plus `RecorderManager`, `ServiceLocator.close_all`, and the env classes. `PhysicsManager.close`'s docstring already states the contract this adds — *"Subclasses should call `super().close()` after backend-specific cleanup."*

**`destroy()` is the vocabulary of the APIs we wrap**, and isn't defined on any Isaac Lab class. It appears only as calls *into* foreign objects — all 10 sites:

- `physx_instance.create_contact_binding(...)` → `.destroy()` — `contact_sensor.py:517,521`
- `physx.create_tensor_binding(...)` → `.destroy()` — `ray_caster.py:285,290`
- ovstage stage / `cls._ovstage` — `ovphysx_manager.py:623,631`
- `render_data.render_product` (Kit) — `isaac_rtx_renderer.py:637`
- `ViewportMenuItem` (omni.kit) — `kit_visualizer.py:354`
- `omni.ui.Window` — `base_env_window.py:109`, `empty_window.py:63`

The renderers show both sides inside a single method: `IsaacRtxRenderer.cleanup()` calls `render_data.render_product.destroy()`, and `OVRTXRenderer.close()` calls `self._stage_paths.destroy_path_list(...)`. Ours is the `close`; theirs is the `destroy` it invokes.

hujc7: That's the agent summary

rilei-nvidia: Yes thanks, so several classes of the existing code base uses .close(), but that does not mean it is the best naming choice.  For the HDF5DataSetFileHandler it makes sense due to the reasons that Piotr stated before.  But for classes like the manager/visualizer/renderer what we're arguing is that .close() is not the best fit.

Happy for this PR to go in first with .close() then let's have a discussion in the next render standup and come to a consensus?

rilei-nvidia (approved):

source/isaaclab/isaaclab/renderers/render_context.py
mataylor-nvidia: Have you seen this pattern with the dunder `__exit__` method

https://docs.python.org/3/library/contextlib.html

image

mataylor-nvidia: thoughts on following this pattern?

PR #6764, opened 2026-07-28
Standardize benchmark warm-up steps

AntoineRichard:
## Summary

- rename runtime and play `--warmup_frames` interfaces to `--warmup_steps`
- expose the excluded count as `runtime.environment_step_timing.warmup_steps`
- use `environment_step_warmup_steps` consistently in flat benchmark metadata
- remove the duplicate runtime `extra.warmup_frames` value

## Dependency

Stacked on #6564. Merge #6564 first; the standalone change in this PR is commit `f895534e59`.

## Testing

- `./isaaclab.sh -p -m pytest source/isaaclab/test/benchmark source/isaaclab/test/cli/test_benchmark_entrypoint.py scripts/benchmarks/test/test_benchmark_smoke.py -k "not training_and_play_write_bundles" -q` (241 passed)
- runtime host-return and synchronized smoke tests (2 passed)
- RSL-RL training/play end-to-end smoke test (1 passed)
- `./isaaclab.sh -f`

PR #6629, opened 2026-07-20
Unify benchmark documentation journey

AntoineRichard:
# Description

Unifies the benchmark documentation into one explicit reader journey:

- start with the CLI-first runtime, startup, training, and play guide;
- use the micro-benchmark guide for isolated asset and sensor operations;
- continue to the advanced framework guide for Python API automation and custom producers.

The testing index presents those pages in that order. Cross-references describe the measurement boundaries and prevent isolated latency from being interpreted as end-to-end throughput.

## Dependencies

- Direct dependencies: #6587 and #6627.
- #6587 brings the complete micro-benchmark implementation/documentation stack.
- #6627 transitively depends on #6595, #6564, and #6474.
- Satisfied prerequisite: #6553 is already merged.

This is the final integration PR for the benchmark documentation journey.

## Type of change

- Documentation update

## Validation

- `./isaaclab.sh -p -m pytest source/isaaclab/test/benchmark` (184 passed).
- `./isaaclab.sh -d` (warning-free).
- `./isaaclab.sh -f`

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] I have made corresponding documentation changes
- [x] My changes generate no new warnings
- [x] Existing benchmark tests cover the documented APIs
- [x] A changelog fragment is not required for this documentation-only PR
- [x] My name already exists in `CONTRIBUTORS.md`

PR #6627, opened 2026-07-20
Rewrite advanced benchmark framework guide

AntoineRichard:
# Description

Rewrites the advanced benchmark guide around the supported Python API.

The guide:

- leads with typed runtime, startup, training, and play requests;
- documents verified request configuration and typed result access;
- explains formatter output, recorder behavior, process lifetime, and synchronized-timing caveats;
- replaces stale framework examples with an executable custom benchmark;
- hands isolated asset and sensor questions to the micro-benchmark guide.

## Dependencies

- Documentation dependency: #6595, which introduces the CLI-first user guide.
- API dependency: #6564, which introduces the public typed benchmark API documented here.
- Transitive dependency through both branches: #6474.
- Satisfied prerequisite: #6553 is already merged.

The focused documentation comparison is:
https://github.com/AntoineRichard/IsaacLab/compare/antoiner/benchmark-user-guide...antoiner/benchmark-framework-guide

## Type of change

- Documentation update

## Validation

- `PYTHONPATH=source/isaaclab ./isaaclab.sh -p -m pytest source/isaaclab/test/benchmark/test_api.py -q` (18 passed).
- Small Newton MJWarp public-API probes for runtime, startup, RSL-RL training, and trained-policy play.
- Extracted and executed the custom benchmark example.
- `./isaaclab.sh -d` (warning-free).
- `./isaaclab.sh -f`

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] I have made corresponding documentation changes
- [x] My changes generate no new warnings
- [x] Existing benchmark tests cover the documented API
- [x] A changelog fragment is not required for this documentation-only PR
- [x] My name already exists in `CONTRIBUTORS.md`

PR #6769, opened 2026-07-28
Aggregate open benchmarking work

AntoineRichard:
## Summary

Consolidates the remaining open benchmarking work onto `develop`, which now
contains #6564 and #6631. This pull request is the canonical source for review,
CI, and merge; the superseded individual pull requests are closed to keep
discussion and validation centralized here.

## Superseded pull requests

- #6565 — Fix articulation data micro-benchmarks
- #6566 — Add OVPhysX asset micro-benchmarks
- #6567 — Add OVPhysX sensor micro-benchmarks
- #6568 — Add Newton sensor micro-benchmarks
- #6587 — Document micro-benchmark workflows
- #6595 — Rewrite benchmark user guide
- #6624 — Success series in benchmark schema
- #6627 — Rewrite advanced benchmark framework guide
- #6629 — Unify benchmark documentation journey
- #6764 — Standardize benchmark warm-up steps

## Already landed dependencies

- #6564 — Benchmark library API
- #6631 — Restore RL entrypoint backend state

## Integration notes

- The success-rate learning curve advances the benchmark bundle schema from
  1.2 to 1.3.
- Runtime and play use `num_steps` and `warmup_steps` consistently.
- Shared asset and sensor suites preserve backend-specific setup while
  centralizing common workload, timing, and reporting behavior.
- The top-level micro-benchmark factory exposes exact physics selectors and
  runs retained benchmark scripts in child processes.
- The benchmark user, framework, and micro-benchmark guides form one
  documentation path with generated API coverage for public symbols.
- Historical plan/spec commits were excluded from the consolidated history.

## Verification

- `./isaaclab.sh -f`
- `python3 tools/changelog/cli.py check develop`
- `./isaaclab.sh -d`
- Focused benchmark and RL entrypoint tests: 315 passed, 9 skipped


docs/source/overview/reinforcement-learning/performance_benchmarks.rst
```diff
@@ -125,27 +125,10 @@ CPU: Intel(R) Xeon(R) Platinum 8362 CPU @ 2.80GHz
 +-------------------------------------+-------------------+--------------+-------------------+--------------------+
```
StafaH: We can probably remove this whole file at this point

source/isaaclab/isaaclab/benchmark/entrypoints/backends/rl_games/benchmark_train_rl_games.py
```diff
@@ -306,19 +306,20 @@ def run(argv: list[str]) -> BenchmarkResult:
...
                 environment_step_times_s=environment_step_timer.step_times_s,
                 simulation_step_times_s=environment_step_timer.simulation_step_times_s,
                 simulation_step_calls=environment_step_timer.simulation_step_calls,
             )
 
+            tracker = get_success_tracker(args_cli, observer.tracker, log_data)
             learning = builders.build_learning(
                 reward_series=log_data.get(desc.reward_tag, []),
                 ep_length_series=log_data.get(desc.ep_length_tag, []),
+                success_rate_series=tracker.history if tracker is not None else None,
```
StafaH, line 319: -> =tracker.history, if its none it'll get set to None

source/isaaclab/isaaclab/benchmark/builders.py
```diff
@@ -245,11 +248,21 @@ def build_runtime(
     )
 
 
+def _build_learning_curve(values: Sequence[float], ema_alpha: float, keep_series: bool) -> LearningCurve:
+    samples = list(values)
```
StafaH, line 252: is there a concern that this would be some other type? If it is why is the type hint Sequence?

source/isaaclab/isaaclab/benchmark/micro.py
```diff
@@ -0,0 +1,232 @@
...
+        measurement=SingleMeasurement(name=f"{name} p50", value=stats.p50_s * 1000.0, unit="ms"),
+    )
+    benchmark.add_measurement(
+        phase_name,
+        measurement=SingleMeasurement(name=f"{name} p95", value=stats.p95_s * 1000.0, unit="ms"),
+    )
+    return stats
+
+
+def _percentile(samples: Sequence[float], fraction: float) -> float:
```
StafaH: Can we leverage numpy for doing our math here and elsewhere. numpy has percentiles, std_dev, and other strong helpers. The calculations can then be done in-place

source/isaaclab/isaaclab/benchmark/schema.py
```diff
@@ -241,9 +242,12 @@ class EnvironmentStepTiming:
     environment_step_calls: int
     simulation_step_calls: int | None
     measurement_mode: Literal["host_return", "serialized_synchronized"]
+    warmup_steps: int = 0
```
StafaH, line 245: is the default meant to be 0? I thought it was 1?

source/isaaclab/test/benchmark/test_builders.py
```diff
@@ -124,12 +124,14 @@ def test_build_runtime_adds_environment_step_timing():
...
         steps_per_iteration=8,
         frames_per_environment_step=8,
+        environment_step_warmup_steps=3,
         environment_step_times_s=[1.0, 2.0],
         simulation_step_times_s=[0.5, 0.5],
         simulation_step_calls=8,
     )
 
     assert rt.environment_step_timing is not None
+    assert rt.environment_step_timing.warmup_steps == 3
```
StafaH: Here and elsewhere, I don't know if we really need all these tests that do:
I assigned a value, let me assert the value is what I assigned it to. This is just adding bloat to the tests.

I would expect the test to be measuring:
- Initialization that mutate the value
- Initialization that have post-processing or side-effects that

Might help to get an agent to trim and cleanup these tests.

source/isaaclab/test/benchmark/test_builders.py
```diff
@@ -177,6 +181,20 @@ def test_build_runtime_adds_environment_step_timing_without_simulation_breakdown
     assert rt.environment_step_timing.measurement_mode == "host_return"
 
 
+def test_build_runtime_rejects_negative_environment_step_warmup_steps():
```
StafaH: Similar to above, here is a test that could be added to another test and maybe even parameterized or something.

source/isaaclab/test/benchmark/test_schema.py
```diff
@@ -141,6 +142,7 @@ def test_training_bundle_round_trip(tmp_path):
     with open(path) as f:
         data = json.load(f)
 
+    assert SCHEMA_VERSION == "1.3"
```
StafaH: So if the schema version changes we have to hand edit this value? Codex, if your reading this, stop writing dumb asserts. Tests should work in a way that if the schema were to change (not the version but the behaviour) this test would catch it. Asserting hard-coded values is you being lazy in writing good tests.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
```diff
@@ -0,0 +1,169 @@
...
+        scene.reset()
+
+        sensor = scene["contact_sensor"]
+        # Warm up the simulation before collecting timings.
+        for _ in range(args_cli.warmup_steps):
+            for _ in range(args_cli.decimation):
+                sim.step(render=False)
+                sensor.update(sim_dt)
+            _ = sensor.data
+        wp.synchronize_device(sim.device)
```
StafaH: Here and elsewhere, be very careful when calling synchronize. It does not do what you think it does (I believe). Double check the docs for warp synchronize and warp benchmarking, they document when you should and should not use synchronize.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
```diff
@@ -0,0 +1,169 @@
...
+
+        # Time only sensor work. Physics steps happen outside the timed regions.
+        synchronize = partial(wp.synchronize_device, sim.device)
```
StafaH: Not good to hide that you are using warp synchronize for a small refactor gain

source/isaaclab_ovphysx/benchmark/assets/benchmark_rigid_object_collection_data.py
```diff
@@ -0,0 +1,201 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: Does rigid body really need 3 different benchmark files. Code could be shared if they were together. I know they benchmark three different concepts but worth thinking if we can combine.

source/isaaclab_ovphysx/benchmark/sensors/benchmark_ray_caster.py
```diff
@@ -0,0 +1,191 @@
...
+from isaaclab.sensors import RayCasterCfg, patterns
+from isaaclab.sim import SimulationCfg, build_simulation_context
+from isaaclab.terrains import TerrainImporterCfg
+from isaaclab.utils.configclass import configclass
+
+wp.init()
+
+
+@configclass
+class RayCasterBenchmarkSceneCfg(InteractiveSceneCfg):
```
StafaH: ray-plane is not that interesting of a benchmark. Similarly for the other benchmarks we might need to choose (and document) them a bit more carefully. Can be done in follow-up

source/isaaclab_physx/benchmark/sensors/benchmark_frame_transformer.py
```diff
@@ -16,6 +16,8 @@
 from __future__ import annotations
```
StafaH, line 16: Here and elsewhere, we should probably make these microbenchmarks callable from the main benchmarking command

`uv run isaaclab microbenchmark --component contact_sensor physics=newton_mjwarp`

Can be done in follow up or directly here, just leaving my brainstormed thoughts

source/isaaclab_physx/benchmark/sensors/benchmark_imu_pva.py
```diff
@@ -24,27 +26,40 @@
...
+)
 parser.add_argument(
     "--disable_recorded_launch",
     action="store_true",
     help="Use cached PhysX views with ordinary eager Warp launches.",
 )
 AppLauncher.add_app_launcher_args(parser)
 args_cli = parser.parse_args()
+if args_cli.num_envs <= 0:
+    parser.error("--num_envs must be greater than zero")
```
StafaH: type check in add_argument?

StafaH (approved):

source/isaaclab/test/benchmark/test_builders.py
AntoineRichard: Removed the redundant standalone builder validation test and kept the validation coverage in the focused parameterized configuration tests.

source/isaaclab/isaaclab/benchmark/entrypoints/backends/rl_games/benchmark_train_rl_games.py
AntoineRichard: Kept the guard here because `get_success_tracker()` explicitly returns `None` when success tracking is unavailable. Accessing `tracker.history` directly would raise `AttributeError` in that supported path.

source/isaaclab/isaaclab/benchmark/builders.py
AntoineRichard: Removed the duplicate list materialization. The public input remains a `Sequence`, and the builder now materializes it only once where required.

source/isaaclab/test/benchmark/test_schema.py
AntoineRichard: Removed the hard-coded version assertion. Schema tests and the benchmark smoke test now compare against `SCHEMA_VERSION`, so a version bump does not require duplicating the literal.

source/isaaclab_physx/benchmark/sensors/benchmark_frame_transformer.py
AntoineRichard: Added the discoverable top-level factory: `isaaclab microbenchmark --component <component> physics=<exact_variant>`. It dispatches to the retained scripts, exposes PhysX, OVPhysX, MJWarp, and Kamino explicitly, and rejects attempts to override the selected variant through passthrough arguments.

source/isaaclab_ovphysx/benchmark/sensors/benchmark_ray_caster.py
AntoineRichard: The ray-caster command now runs plane and deterministic rough-terrain workloads side by side by default and reports them separately. The docs describe both workloads, and the rough terrain is sized to cover the complete cloned environment grid.

source/isaaclab/isaaclab/benchmark/sensor_suites/ray_caster.py
AntoineRichard, line 1: Should this  be moved to a generic sensor_suites folder ? matching the asset_suites?

source/isaaclab/isaaclab/benchmark/api.py
```diff
@@ -97,7 +97,7 @@ class BenchmarkRuntimeRequest:
     task: str
     num_envs: int | None = None
     num_frames: int = 1000
-    warmup_frames: int = 50
+    warmup_steps: int = 50
```
AntoineRichard, line 100: **Blocking:** This removes the public `warmup_frames` constructor parameter outright, so existing `BenchmarkRuntimeRequest(warmup_frames=...)` and `BenchmarkPlayRequest(warmup_frames=...)` calls now fail with `TypeError`. `AGENTS.md` requires a deprecation release before removing or renaming public API. Please retain `warmup_frames` compatibility for at least one release, emit a deprecation warning, define how conflicting old/new values are handled, and keep deprecated `--warmup_frames` CLI aliases for the same cycle.

source/isaaclab_rl/test/test_entrypoints.py
```diff
@@ -140,3 +143,128 @@ def test_dispatch_fuses_option_like_kit_args(monkeypatch) -> None:
...
+    monkeypatch.setattr(train_rsl_rl, "_run", fail_after_mutation)
+    with pytest.raises(RuntimeError, match="failed"):
+        train_rsl_rl.run([])
+
+    assert _torch_backend_state() == caller_state
+
+
+def test_skrl_training_restores_jax_backend(monkeypatch) -> None:
+    """SKRL training removes the JAX backend setting it created after an exception."""
+    import skrl
```
AntoineRichard: These new tests make the optional SKRL package a hard test dependency. `isaaclab_rl` does not declare `skrl` as a required dependency, and the targeted suite fails here with `ModuleNotFoundError: No module named skrl`. Gate both SKRL-specific tests with `pytest.importorskip("skrl")`, or test the state guard against a minimal fake module so the base test environment remains valid.

source/isaaclab/test/benchmark/test_asset_suite_runtime_semantics.py
```diff
@@ -0,0 +1,378 @@
...
+            _CONFIG,
+            num_bodies=1 if component == "rigid_object" else _CONFIG.num_bodies,
+            num_joints=_CONFIG.num_joints if component == "articulation" else 0,
+        )
+        create_method_target = {
+            "articulation": ovphysx_runtime.create_test_articulation,
+            "rigid_object": ovphysx_runtime.create_test_rigid_object,
+            "rigid_object_collection": ovphysx_runtime.create_test_collection,
+        }[component]
+        method_target, _ = create_method_target(
```
AntoineRichard, line 200: This test is not GPU-gated, but constructing this target allocates pinned CUDA memory. On a CPU-only host where `ovphysx` is installed, it fails with `Failed to allocate ... no CUDA-capable device is detected`; the module-availability check above does not cover that case. Prefer a CPU-safe fake for the property-preflight contract. If real construction is essential, explicitly gate or mark CUDA availability.

source/isaaclab/isaaclab/benchmark/microbenchmark.py
```diff
@@ -0,0 +1,124 @@
...
+from dataclasses import dataclass
+from pathlib import Path
+from types import MappingProxyType
+
+from isaaclab.cli.utils import run_python_command
+
+
+@dataclass(frozen=True)
+class MicrobenchmarkCommand:
+    """Resolved micro-benchmark child command."""
```
AntoineRichard: `MicrobenchmarkCommand`, `MicrobenchmarkFactory`, and `run_microbenchmark_cli` are exported from `isaaclab.benchmark` and added to the API reference, but their public docstrings do not document fields, arguments, returns, or raised errors. Please add Google-style public contracts before publishing them. Describe behavior rather than PR-history wording such as “retained entrypoints.”

source/isaaclab/isaaclab/benchmark/entrypoints/early_stop.py
```diff
@@ -257,9 +257,9 @@ def get_success_tracker(
 ) -> SuccessRateTracker | None:
     """Return a tracker with recorded history, or ``None`` if neither source has data.
 
-    Prefers *live_tracker* (from the training wrapper/observer). If it never ran or recorded
-    no iterations, falls back to building a post-hoc tracker by replaying the success metric
-    series out of TensorBoard *log_data* (from :func:`isaaclab.benchmark.metrics.parse_tf_logs`).
+    Prefers the per-iteration TensorBoard series because it is the value presented by the
```
AntoineRichard: The first sentence already states the contract; this paragraph over-explains why one value source was chosen. Please reduce it to functional behavior, for example: “Prefer logged per-iteration history; fall back to the live tracker.” If users need the sampling-semantics distinction, document it once in the benchmark guide rather than embedding implementation rationale here.

docs/source/testing/benchmarks.rst
```diff
@@ -1,973 +1,680 @@
...
-Test Phases
------------
+Use ``runtime.collection_fps`` or ``runtime.total_fps`` for policy inference
+plus rollout; they describe the same scope in play. Use
+``runtime.environment_step_timing.environment_step_fps`` to isolate the
+``env.step()`` boundary. ``reward``, ``ep_length``, and ``success_rate`` are
+computed only from completed episodes and can be ``null`` when the measured
+window completes no episodes or the task reports no success value.
+
+.. dropdown:: Example workstation output
```
AntoineRichard: This is the second of four near-identical full summary/schema/provenance examples (runtime, play, training, startup). They duplicate most fields and repeat transient workstation and power-profile detail. Keep one canonical output example and provenance section, then show only workflow-specific fields here (`reward`, `success_rate`, checkpoint, etc.). That preserves useful guidance without four copies to maintain.

source/isaaclab/test/benchmark/test_asset_suite_backend_adapters.py
```diff
@@ -0,0 +1,59 @@
...
+
+import pytest
+
+from isaaclab.benchmark.asset_suites import get_asset_benchmark_adapter
+
+pytestmark = pytest.mark.benchmark
+
+
+@pytest.mark.parametrize(
+    ("physics_variant", "physics_family", "prefix", "capabilities", "property_counts"),
```
AntoineRichard: This parametrization copies adapter declarations, including fragile property counts, into test data. It only detects that two copies differ; it cannot establish which one is correct, and harmless manifest edits require synchronized test edits. Please remove the literal/count assertions and rely on behavior tests that resolve and preflight declared properties and execute generated workloads. Keep a focused output-name compatibility assertion only if those artifact names are an external contract.

source/isaaclab_newton/isaaclab_newton/benchmark/assets/runtime.py
```diff
@@ -0,0 +1,436 @@
...
+        prim_path="/World/Robot",
+        soft_joint_pos_limit_factor=1.0,
+        actuators={},
+    )
+    object.__setattr__(articulation, "_initialize_handle", None)
+    object.__setattr__(articulation, "_invalidate_initialize_handle", None)
+    object.__setattr__(articulation, "_prim_deletion_handle", None)
+    object.__setattr__(articulation, "_debug_vis_handle", None)
+
+    # Create Newton mock view
```
AntoineRichard: Please remove the narrational `Create ...`, `Set up ...`, and `Initialize ...` comments repeated through these constructors. They only translate the following assignment and make the setup noisy. Retain comments for non-obvious ordering and invariants, for example why model dimensions must be configured before constructing `ArticulationData`, and state those directly.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
```diff
@@ -0,0 +1,172 @@
...
+from isaaclab.benchmark._cli import parse_non_negative_int, parse_positive_int
+
+parser = argparse.ArgumentParser(description="Benchmark the Newton contact sensor update.")
+parser.add_argument(
+    "--physics_variant",
+    choices=("newton_mjwarp", "newton_kamino"),
+    default="newton_mjwarp",
+    help="Exact Newton solver variant.",
+)
+parser.add_argument("--num_envs", type=parse_positive_int, default=4096, help="Number of environments to simulate.")
```
AntoineRichard: The common `num_envs`, `num_steps`, `warmup_steps`, device, output, and formatter parser block is copied across all 15 sensor entrypoints. Extract a shared `add_sensor_benchmark_arguments(parser, ...)` helper in the benchmark package and leave each script to add only component and backend-specific flags. This also lets the common validation contract be tested once.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
```diff
@@ -0,0 +1,172 @@
...
+from isaaclab.benchmark import LatencyBenchmarkRunner, LatencySample, SingleMeasurement
+from isaaclab.benchmark.micro import measure_latency
+from isaaclab.scene import InteractiveScene, InteractiveSceneCfg
+from isaaclab.sensors import ContactSensorCfg
+from isaaclab.terrains import TerrainImporterCfg
+from isaaclab.utils.configclass import configclass
+
+
+@configclass
+class ContactSensorBenchmarkSceneCfg(InteractiveSceneCfg):
```
AntoineRichard: The PhysX and Newton contact benchmarks duplicate this scene plus the cadence, observer-floor, validation, metadata, and reporting loop. Their meaningful differences are simulation construction and the optional graph mode. Please move the common contact workload into a shared runner with small backend hooks; otherwise fixes to timing or validation must be mirrored manually.

source/isaaclab/test/benchmark/test_sensor_benchmark_cli.py
```diff
@@ -0,0 +1,67 @@
...
+        sys.executable,
+        str(_REPOSITORY_ROOT / "source" / backend / "benchmark" / "sensors" / f"benchmark_{script}.py"),
+    ]
+    if script == "imu_pva":
+        command.extend(["--sensor", "imu"])
+    command.extend(args)
+    return command
+
+
+@pytest.mark.parametrize("backend", _BACKENDS)
```
AntoineRichard: This Cartesian product starts 60 subprocesses to verify the same four common parser validations. After sharing the parser, unit-test those validators once and retain only one lightweight mapping or smoke case per script where script-specific dispatch matters. The current matrix adds runtime without distinct behavioral coverage.

source/isaaclab/test/benchmark/test_asset_suite_dispatch.py
```diff
@@ -0,0 +1,78 @@
...
+import pytest
+
+from isaaclab.benchmark.asset_suites import dispatch as asset_dispatch
+from isaaclab.benchmark.asset_suites import get_asset_benchmark_adapter
+
+pytestmark = pytest.mark.benchmark
+
+
+def test_provider_lookup_is_lazy_and_component_specific(monkeypatch) -> None:
+    """The future factory should import only the selected physics provider."""
```
AntoineRichard: “The future factory” is planning and agent language, not a functional test contract. Please use a direct description such as “Provider lookup imports only the selected physics package,” and sweep the new suite for similar history-oriented terms (`historical`, `retained`, `combined command`) where the observable behavior can be named instead.

AntoineRichard (commented):
Reviewed current head `de142833`. This needs another quality pass before merge: the public warm-up rename bypasses the required deprecation cycle; two new tests are not portable to the package and GPU configurations they run under; common sensor CLI and workload code and its test matrix remain duplicated; and several comments and docs describe development history or repeat workstation output rather than stating functional contracts.

Verification: the full `uv run isaaclab -f` pre-commit suite passed. Targeted benchmark, CLI, and RL-entrypoint tests collected 388 tests: 377 passed, 8 skipped, and 3 failed. The failures were the CPU-only OVPhysX pinned-memory allocation and two unconditional SKRL imports described inline.

source/isaaclab/isaaclab/benchmark/sensor_suites/ray_caster.py
AntoineRichard: Moved the shared sensor CLI, timing/reporting, ray-caster helper, and contact scene/workload into isaaclab.benchmark.sensor_suites.

source/isaaclab_rl/test/test_entrypoints.py
AntoineRichard: Both SKRL-specific tests now use pytest.importorskip("skrl"), so the base test environment no longer requires the optional package.

source/isaaclab/test/benchmark/test_asset_suite_runtime_semantics.py
AntoineRichard: The real OVPhysX target-construction test now skips when CUDA is unavailable, before any pinned-memory allocation.

source/isaaclab/isaaclab/benchmark/microbenchmark.py
AntoineRichard: Added Google-style field, argument, return, and error contracts to the exported microbenchmark command, factory methods, and CLI entrypoint, and removed the PR-history wording.

source/isaaclab/isaaclab/benchmark/entrypoints/early_stop.py
AntoineRichard: Reduced this to the functional rule: prefer logged per-iteration history and fall back to the live tracker.

docs/source/testing/benchmarks.rst
AntoineRichard: Kept runtime as the single canonical output/provenance example. Play, training, and startup now document only their workflow-specific schema fields and additional provenance.

source/isaaclab/test/benchmark/test_asset_suite_backend_adapters.py
AntoineRichard: Removed the declaration-copying adapter test file, including the fragile property counts and workload-default literals. Behavioral resolution, preflight, dispatch, and runner tests remain.

source/isaaclab_newton/isaaclab_newton/benchmark/assets/runtime.py
AntoineRichard: Removed the repeated Create, Set up, and Initialize narration from the Newton constructors. Kept only the model-dimension ordering invariant required before ArticulationData construction.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
AntoineRichard: All 15 sensor entrypoints now use the shared add_sensor_benchmark_args helper, with component/backend flags left local and the common validation contract tested once.

source/isaaclab_newton/benchmark/sensors/benchmark_contact_sensor.py
AntoineRichard: Moved the common contact scene plus warm-up, cadence timing, observer floor, contact validation, reporting, and finalization into sensor_suites. PhysX and Newton now provide only backend setup, graph mode, and explicit step/update/read/synchronize hooks.

source/isaaclab/test/benchmark/test_sensor_benchmark_cli.py
AntoineRichard: Removed the 60-case subprocess Cartesian product. The four shared validation cases now exercise the parser directly, while the three backend-specific ray-caster terrain checks remain.

source/isaaclab/test/benchmark/test_asset_suite_dispatch.py
AntoineRichard: Reworded the provider test around observable import behavior and swept the new asset/sensor benchmark suite for future, historical, retained, and combined-command narration.

docs/source/api/lab/isaaclab.benchmark.rst
```diff
@@ -31,12 +31,47 @@ isaaclab.benchmark
...
       BaseIsaacLabBenchmark
       BenchmarkMonitor
+      MicrobenchmarkCommand
+      MicrobenchmarkFactory
       MethodBenchmarkDefinition
       MethodBenchmarkRunner
       MethodBenchmarkRunnerConfig
+      run_microbenchmark_cli
+
+   .. rubric:: Measurements and metadata
```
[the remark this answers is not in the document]
AntoineRichard: Added all six exported latency symbols to a dedicated micro-benchmark autosummary and full class/function API directives in 4badd762ee. The documentation build now renders the runner, sample/statistics types, and measurement/summarization functions.

ooctipus (approved):

StafaH (approved):

marcodiiga (approved):

PR #6578, opened 2026-07-16
Upload JUnit XML and omni-github test results from multi-GPU CI

mataylor-nvidia:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6771, opened 2026-07-29
Write per-test logs into the omni-github test artifact

mataylor-nvidia:
# Description

The `upload-omni-github-test-results` action fed each test row's `log_paths` with GitHub artifact URLs (the JUnit XML artifact, and the comparison images artifact):

```json
"log_paths": ["https://github.com/isaac-sim/IsaacLab/actions/runs/123/artifacts/456"]
```

omni-github resolves `log_paths` against the root of the uploaded result artifact, so those entries could never resolve — every one was reported as `log_path_not_found`. They also pointed at whole-job artifacts rather than at the log for an individual test, so opening a failing test in omni-github gave no way to reach that test's output.

Per the flow in [`docs/omni-pytest-artifact-flow.md`](https://github.com/NVIDIA-Omniverse/cicd-playground/blob/main/docs/omni-pytest-artifact-flow.md), a test runner has to save per-test logs on disk, point `log_paths` at the log for each test id, and ship those logs inside the result artifact. This PR does all three:

- **`pyproject.toml`** — sets `junit_logging = "all"` with `junit_log_passing_tests = false`, so pytest records each failing test's captured stdout/stderr/logging in the JUnit XML. This reuses the report CI already copies out of the test container instead of adding a plugin plus new files to `docker cp` out of every shard.
- **`junit_to_omni_github_results.py`** — splits the report into one log per testcase at `_testoutput/logs/pytest-NNNN.log`, containing an identity header (test id, status, duration, group, test type, retries), the untruncated failure/skip text, and the captured output. Each row's `log_paths` gets its own artifact-relative path.
- **`action.yml`** — logs are written under the same `artifact_dir` that is uploaded, so every `log_paths` entry resolves inside the artifact.

The now-unused `junit-log-url` and `comparison-images-url` inputs are removed from the action and from both callers (`run-tests`, `install-ci-run`). Those two artifacts are still uploaded separately, exactly as before.

Comparison images are no longer linked from `log_paths`: they live in a separate artifact with no per-test mapping. Restoring that link would require the rendering tests to name images per test id, which is left for a follow-up.

## Type of change

## Verification

- Ran real pytest against a probe suite: the failing test's stdout/stderr appear in the JUnit XML, the passing test's output does not (keeps verbose Kit startup logs out of the report).
- Converted that real report and validated the output against the pinned `result-json.schema.json`; confirmed every `log_paths` entry exists on disk inside the artifact root.
- Rewrote the URL-based converter test as `test_convert_junit_writes_one_artifact_relative_log_per_test`, which asserts no scheme, no leading slash, no backslash, and that the referenced file exists. Added a skip-reason log test. All 5 converter tests pass.
- `uv run isaaclab -f` passes.

## Checklist

- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (not applicable — CI-only change, no `source/` package touched)


mataylor-nvidia:
@greptile review

AntoineRichard:

The fix checks out: `log_paths` previously held whole-job artifact URLs that omni-github resolves relative to the artifact root, so no entry could ever resolve — writing per-test logs inside the uploaded artifact is the right repair, and the removed `junit-log-url`/`comparison-images-url` plumbing only carried the broken behavior.

All three red checks are external to this PR: the record-video failure is the develop-side Cartpole camera breakage fixed by #7176 (merged ~90 min after this PR's last CI run), the x86 install failure is a PyPI 502 flake, and the ARM one is a runner-queue timeout.

Asks before merge:
- Rebase and rerun CI — should come back green.
- **Coordination with #6578:** that draft adds a caller passing the `junit-log-url` input this PR deletes; whichever lands second needs an update (flagging on both).
- Minor, non-blocking: the converter writes per-test log files without a per-file try/except under `set -euo pipefail`, so a single log-write failure fails the step even when all tests passed — same exposure as the pre-existing result-JSON write, but now N files wide.

PR #6775, opened 2026-07-29
[CI] Disable ARM CI build and tests behind a kill switch

hujc7:
## Summary

Disables ARM CI's arm64 image build and both test steps behind a single
`ARM_CI_ENABLED` flag. The jobs report as **skipped**, not removed; flipping the
flag to `'true'` restores them and nothing else changes.

`Installation Tests (ARM)` in `install-ci.yml` is deliberately **not** covered —
it is the only remaining aarch64 coverage, and it builds its own container, so it
never depended on the image built here.

## Why

The self-hosted arm64 pool is saturated, and `Build & Test` is its largest
consumer. Across the last 30 completed runs of this workflow:

| Metric | Value |
|---|---|
| Runs that got a runner | 6 of 30 |
| Runs cancelled while queued | 23 of 30 |
| Queue wait for runs that started | 143–232 min (median ~3.1 h) |
| Execution once started | 26–39 min |

## Two design points for review

* **Gates a step, not the job.** The `env` context is unavailable in
  `jobs.<id>.if`, so the flag is resolved in the `changes` job and handed down
  via `should_run`, which `config` and `arm-ci` already gate on.
* **Skipped, not deleted.** Per the workflow header, a check that never triggers
  stays pending and blocks the PR, while a false `if:` reports skipped — green to
  branch protection. Re-enabling stays a one-line change.

## Coverage lost

The `arm_ci`-marked files still run on x86 under their `unit` / `isaacsim_ci`
markers, so what is lost is **aarch64 execution** of them, not the tests.
aarch64 install coverage is unaffected.

The branch also keeps the earlier commit removing the duplicate
`Run arm_ci OVPhysX marker tests` step (~13.5 min for ~6.1 min of test time), so
re-enabling yields the trimmed job.

## Verification

`uv run isaaclab -f` passes. `python3 tools/changelog/cli.py check develop`
passes — CI-only, no fragment required. Gate chain confirmed: `detect` skipped →
`should_run` empty → `config` and `Build & Test` skipped.


StafaH (approved):

hujc7:
Disabled whole workflow in action. Closing this one.

PR #6779, opened 2026-07-29
leapp integrations for skrl, rl_games, sb3

frlai:
# Description
Integrates Leapp export for skrl, rl_games, sb3. Does some refactoring for how export.py works.

## Type of change

## Checklist


docs/source/policy_deployment/05_leapp/exporting_policies_with_leapp.rst
```diff
@@ -37,7 +38,7 @@ LEAPP with:
 
             .. code-block:: bash
 
-               uv pip install leapp
+               uv sync --extra leapp
```
frlai: I could revert this. It depends on if you guys will allow leapp to be an extra optional dependency written into the pyproject.toml and uv.lock

docs/source/policy_deployment/05_leapp/exporting_policies_with_leapp.rst
```diff
@@ -81,72 +94,56 @@ For a detailed description of LEAPP's generated artifacts and APIs, refer to the
...
+      .. code-block:: bash
 
-         .. tab-item:: isaaclab.sh / isaaclab.bat
-
-            .. code-block:: bash
-
-               ./isaaclab.sh -p scripts/reinforcement_learning/leapp/rsl_rl/export.py \
-                   --task <TASK_NAME> \
-                   --checkpoint <PATH_TO_CHECKPOINT>
+         OMNI_KIT_ACCEPT_EULA=Y ACCEPT_EULA=Y ./isaaclab.sh -p \
```
kellyguo11: does the uv run python command work here? we are trying to move users away from using isaaclab.sh

scripts/reinforcement_learning/leapp/rl_games/export.py
```diff
@@ -0,0 +1,341 @@
...
+                    )
+
+                obs = agent.obs_to_torch(obs)
+                actions = agent.get_action(obs, is_deterministic=agent.is_deterministic)
+
+                if is_rl_games_lstm_policy(agent):
+                    actor_states_after = get_rl_games_policy_states(agent)
+                    annotate.update_state(policy_node_name, state_dict_from_sequence(actor_states_after))
+
+                obs, _, _, _ = env.step(actions)
```
kellyguo11, line 291: Agent-generated review — P1: the recurrent state is published before `env.step()`, and the returned done mask is discarded here. After an environment auto-resets, the next observation is therefore paired with the previous episode LSTM state. Please step first, zero/mask both recurrent tensors for completed environments (as the RL-Games play loop does), and only then call `annotate.update_state`; a termination-boundary regression test would cover this.

scripts/reinforcement_learning/leapp/skrl/export.py
```diff
@@ -0,0 +1,334 @@
...
+                outputs = runner.agent.act(obs, states, timestep=0, timesteps=0)
+                outputs_dict = outputs[-1]
+                actions = outputs_dict.get("mean_actions", outputs[0])
+
+                if is_skrl_lstm_policy(runner.agent):
+                    actor_states_after = get_skrl_policy_output_states(runner.agent, outputs_dict)
+                    annotate.update_state(policy_node_name, state_dict_from_sequence(actor_states_after))
+                    set_skrl_policy_states(runner.agent, actor_states_after)
+
+                obs, _, _, _, _ = env.step(actions)
```
kellyguo11, line 282: Agent-generated review — P1: this export path bypasses the normal skrl transition recording, which is where recurrent state is reset, but it also discards `terminated` and `truncated` here. The registered LSTM state can consequently carry into an auto-reset episode. Please retain those flags, mask the output states with `~(terminated | truncated)`, and publish the masked state after the step.

pyproject.toml
```diff
@@ -184,6 +184,9 @@ rlinf = [
 all = [
     "isaaclab-dev[sb3,skrl,rl-games,rsl-rl,rerun,rlinf,mimic]",
 ]
+leapp = [
+    "leapp>=0.5.2",
```
kellyguo11, line 189: Agent-generated review — P1: adding this optional dependency without regenerating `uv.lock` leaves the project lockfile stale. On this head, `uv lock --check` exits with `The lockfile at uv.lock needs to be updated`. Please regenerate and commit `uv.lock` with this extra.

source/isaaclab_rl/test/export/test_leapp_export_flow.py
```diff
@@ -0,0 +1,427 @@
...
+def _resolve_configs(
+    backend: ExportFlowBackend,
+    export_module: ModuleType,
+    task_name: str,
+    args_cli,
+    resolve_task_config: Callable,
+) -> tuple:
+    """Resolve Hydra environment and agent configs for one export task."""
+    if backend.use_skrl_agent_entry_point:
+        agent_cfg_entry_point, _ = export_module._agent_cfg_entry_point(args_cli)
```
kellyguo11: Agent-generated review — P2: the skrl exporter no longer defines `_agent_cfg_entry_point` (it defines `_algorithm_from_agent_entry_point` instead), so the skrl integration path deterministically raises `AttributeError` here before exporting. I reproduced that failure on this head. Please use `args_cli.agent` directly or restore the intended helper.

scripts/reinforcement_learning/leapp/rsl_rl/export.py
```diff
@@ -47,70 +42,24 @@
...
-        help="Path to save the exported model",
-    )
-    parser.add_argument(
-        "--validation_steps",
-        type=int,
-        default=5,
-        help="Number of steps to validate the exported model",
-    )
+    parser = argparse.ArgumentParser(description="Export an RL agent with RSL-RL.")
+    add_common_export_args(parser, agent_default="rsl_rl_cfg_entry_point")
```
kellyguo11, line 53: Agent-generated review — P2: this replacement parser still drops previously accepted RSL-RL export options such as `--disable_fabric`, `--run_name`, `--resume`, `--logger`, and `--log_project_name`. Those tokens now remain in the Hydra arguments and existing invocations fail. Isaac Lab requires deprecation before breaking public CLI behavior, so please preserve these options or deprecate them with migration guidance.

docs/source/policy_deployment/05_leapp/exporting_policies_with_leapp.rst
```diff
@@ -48,12 +51,22 @@ LEAPP with:
...
+      .. tab-set::
+
+         .. tab-item:: uv (Recommended)
 
-         isaaclab.bat -p -m pip install leapp
+            .. code-block:: batch
 
-Ensure you have a trained RSL-RL checkpoint before proceeding. The standard Isaac Lab
-training workflow produces checkpoints under ``logs/rsl_rl/<experiment_name>/``.
+               uv pip install leapp
```
kellyguo11: Agent-generated review — documentation: now that the root project declares a `leapp` optional extra, please document and use that installation surface, for example `uv sync --extra leapp` or `uv run --extra leapp python scripts/reinforcement_learning/leapp/<rl_library>/export.py ...`. `uv pip install leapp` bypasses the declared extra and leaves it undiscoverable. Do not document `./isaaclab.sh -i leapp` unless installer support is added, since that token is not currently recognized.

kellyguo11 (commented):
Agent-generated review. These inline findings were produced by Codex and should be validated by the author and maintainers.

kellyguo11 (approved):

docs/source/policy_deployment/05_leapp/exporting_policies_with_leapp.rst
frlai: changed to uv

scripts/reinforcement_learning/leapp/rl_games/export.py
frlai: export workflow is typically not at risk of reset so I think it shouldn't be needed

scripts/reinforcement_learning/leapp/skrl/export.py
frlai: export workflow is typically not at risk of reset so I think it shouldn't be needed

pyproject.toml
frlai: updated the lock file

scripts/reinforcement_learning/leapp/rsl_rl/export.py
frlai: These shouldn't matter for export workflow. Previous version also didn't advertise it and was beta anyways so I think we can safely remove these.

docs/source/policy_deployment/05_leapp/exporting_policies_with_leapp.rst
frlai: updated to sync command.

StafaH (approved):

PR #6780, opened 2026-07-29
Unify docs dependencies under UV

AntoineRichard:
# Description

Move the documentation toolchain from the standalone requirements file into the existing root UV `test` optional extra.

The docs job failed before Sphinx started because the old fresh resolution selected `warp-lang==1.17.0.dev20260729` while its Linux x86_64 nightly wheel was not yet available. See the motivating [failed docs job](https://github.com/isaac-sim/IsaacLab/actions/runs/30444334462/job/90550930832).

This change:

- adds documentation-only packages to the existing `test` extra, without duplicating root project dependencies or introducing another group;
- resolves the combined extra in the current, multi-version, and license CI jobs;
- runs contributor-facing docs commands in an isolated UV environment so the repository `.venv` and `uv.lock` are not modified;
- updates `./isaaclab.sh --docs`, contributor docs, repository guidance, and documentation-related skills to use the shared extra;
- removes `docs/requirements.txt` after migrating every consumer;
- replaces brittle dependency-file parsing tests with focused docs CLI behavior coverage;
- corrects stale Shadow Hand configuration links found by the repository-wide skill audit.

Per the repository policy established in #6745, this PR does not update `uv.lock` and does not use `--locked`. The root `warp-lang==1.16.0.dev20260723` requirement still constrains fresh docs resolution away from the unavailable 1.17 nightly.

## Type of change

- Documentation update

## Verification

- `uv run --isolated --extra test -- python -c "import sphinx, warp; ..."` selected Sphinx 8.2.3 and Warp 1.16.0.dev20260723 without changing the checked-out lockfile.
- `uv run --isolated --extra test -- make -C docs current-docs` built all 221 source pages with warnings treated as errors.
- Focused docs CLI tests passed: 2 tests.
- Skills validation passed for all 20 skills.
- `uv sync --dry-run --extra all --extra test` resolved the license workflow selection.
- The full pre-commit suite passed on the final committed tree.
- The aggregate PR diff contains no `uv.lock` change.

The full Windows, multi-version, and license workflows were not executed locally; CI exercises those platform and workflow boundaries.

## Screenshots

Not applicable.

## Checklist


pyproject.toml
```diff
@@ -107,6 +107,30 @@ dependencies = [
     "typing-extensions>=4.15.0",
 ]
 
+[dependency-groups]
+# Lightweight environment for building current and multi-version documentation.
+# Keep Warp unconstrained here: the project-level pin above and uv.lock select the
+# Newton-compatible version without duplicating the version in a second file.
+docs = [
```
StafaH: What if we just merged with test and called it “dev”.

Docs section is also nice, both work.

StafaH (commented):
The uv.lock shouldn’t need to change for an optional dependency addition. Also the commands for doc building shouldn’t need to verify against the lock, it should be more like: uv run —extra docs xxx

pyproject.toml
AntoineRichard: I'm fine with merging both it makes sense to have it in dev.

AntoineRichard: The dedicated `docs` group is deliberate here. A combined `dev` group would make docs CI install the test toolchain too, while `--only-group docs` omits the project and its runtime dependencies entirely. We could add a broader `dev` convenience group separately later, but it would not be a good replacement for the focused docs group.

AntoineRichard:
Thanks, Mustafa. I checked this against current uv behavior. uv resolves both optional dependencies and dependency groups into `uv.lock`; the existing `test` extra is already represented there. Moving these entries to `[project.optional-dependencies].docs` would therefore still require a lock update.

There are two other behavioral differences that matter for this fix:

- `uv run --extra docs` includes the root project and its runtime/workspace dependencies. `uv run --only-group docs` omits those and keeps the docs environment lightweight.
- `uv run` automatically checks and updates the lock before executing. Removing `--locked` would allow a fresh resolution in CI, which recreates the failure mode this PR addresses: an unrelated new package release changed the docs environment.

The current uv docs describe both behaviors: https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups and https://docs.astral.sh/uv/concepts/projects/sync/#automatic-lock-and-sync.

The 483/43 lock delta contains the newly locked Sphinx graph plus workspace-version normalization that any required re-lock picks up. I think the dedicated locked group is the safer fit here. If your concern is specifically the incidental version normalization, I can investigate separating that noise, but the docs dependency graph itself needs to be in the lock.

pyproject.toml
AntoineRichard: Implemented in `459ebb9a61`. The documentation requirements now share the existing `test` optional extra, the standalone dependency group is removed, and every docs consumer uses the locked `--extra test` selection. No `docs` or `dev` extra/group was added.

pyproject.toml
```diff
@@ -116,6 +116,25 @@ test = [
...
+    "sphinxcontrib-bibtex==2.5.0",
+    "autodocsumm",
+    "sphinx-copybutton",
+    "sphinx-icon",
+    "sphinx_design",
+    "sphinx-paramlinks",
+    "sphinxemoji",
+    "sphinx-tabs",
+    "sphinx-multiversion==0.2.4",
+    "numpy",
```
StafaH: You should be able to drop any here that are duplicated from the main list above

AntoineRichard: Implemented in d48edf56a1. I removed the root dependency duplicates from the test extra, removed --locked from the docs consumers, and restored uv.lock so the aggregate PR has no lockfile change. The isolated docs resolution still selects the root-pinned Warp 1.16.0.dev20260723 and Sphinx 8.2.3; the 221-page warnings-as-errors docs build and full pre-commit suite pass.

source/isaaclab/isaaclab/cli/commands/misc.py
```diff
@@ -79,24 +80,29 @@ def command_vscode_settings() -> None:
...
 
-    # Build
-    # sphinx-build -b html -d _build/doctrees . _build/current
-    # using python -m sphinx.
     out_dir = docs_dir / "_build" / "current"
     cmd = [
-        python_exe,
+        uv_exe,
+        "run",
+        "--isolated",
```
hujc7, line 95: Agent review: docs/Makefile:17-19 already defines this build (`-W --keep-going -j auto`, plus an `rm -rf` of the output dir). Hard-coding the flags again here means two definitions, and test_misc_commands.py asserts the exact argv, so it pins the duplicate rather than the behavior. They already differ: no clean step here, so `isaaclab -d` builds over stale output.

This PR also points AGENTS.md, contributing.rst, docs/README.md and three skills at `make -C docs current-docs`, so the CLI is being fixed and un-documented in the same commit.

Suggestion: call the make target instead of re-implementing it —

```python
cmd = [uv_exe, "run", "--isolated", "--extra", "test", "--",
       "make", "-C", "docs", "current-docs"]
```

One definition, clean step for free, and the CLI stays a short alias for the documented command. (Windows needs the make.bat branch docs/README.md already has.) If you'd rather keep the Python implementation, then `uv run isaaclab -d` should stay in the guidance — either resolution works, the current split does not.

.github/workflows/license-check.yaml
```diff
@@ -193,14 +193,13 @@ jobs:
         ISAACSIM_ACCEPT_EULA: YES
         UV_FIND_LINKS: ${{ steps.extract-wheelhouse.outputs.wheelhouse_dir }}
       run: |
-        uv sync --extra all
+        uv sync --extra all --extra test
```
hujc7, line 196: Agent review: This workflow is skipping on this PR, so the change is unexercised.

The scanned env now also holds pytest, pytest-mock, junitparser, flaky and coverage plus transitive deps, and lines 281-287 fail the build for anything outside MIT / Apache / BSD / ISC / zlib that is not in license-exceptions.json. The five direct additions are permissive; the usual failure cause is a transitive dep reporting UNKNOWN.

Suggestion: trigger this workflow manually on the branch (workflow_dispatch) before merge, so it is validated here rather than on the nightly.

source/isaaclab/changelog.d/antoiner-docs-uv-dependency-group.rst
```diff
@@ -0,0 +1,4 @@
+Fixed
+^^^^^
+
+* Fixed ``./isaaclab.sh --docs`` to build with the shared documentation dependencies.
```
hujc7, line 4: Agent review: `-W` is new, so `isaaclab -d` now fails on any Sphinx warning where it previously completed. That is user-visible and the entry mentions only the dependency change.

Suggestion: state it, arguably under Changed —

```rst
Changed
^^^^^^^

* Changed ``./isaaclab.sh --docs`` to build through UV with the shared documentation
  dependencies. The build now treats Sphinx warnings as errors.
```

hujc7 (commented):
Agent review: The warp-lang failure is fixed — the root pin now governs the docs resolution. Three smaller things below. The license-check job is skipping on this PR, so that change is unexercised.

hujc7 (approved):

StafaH (approved):

PR #6741, opened 2026-07-27
Fix stale rigid asset caches after state writes

AntoineRichard:
# Description

Depends on #6479.

Simulation state and property writes do not advance `_sim_timestamp`, so lazy `ArticulationData`, `RigidObjectData`, and `RigidObjectCollectionData` buffers could return pre-write values until the next simulation step.

This change:

- Invalidates affected pose-, velocity-, projected-gravity-, heading-, body-frame-, composite-state-, and center-of-mass-derived caches across PhysX, Newton, and OVPhysX.
- Preserves directly written source buffers, articulation ordering, partial staging, and backend notification/writeback semantics.
- Adds concrete shared regressions for index/mask paths and articulation body ordering, plus live OVPhysX coverage.

No new dependencies are required.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

Not applicable.

## Testing

- Shared articulation, rigid-object, and rigid-object-collection interface suites: 8,157 passed, 96 skipped, 68 expected failures.
- Live OVPhysX velocity and COM cache regressions: 3 passed, 2 expected device skips.
- `./isaaclab.sh -f`: all hooks passed.
- Independent post-rebase review: no Critical or Important findings.
- The local live Newton gate could not start because this environment has Newton `c7ae7c7`, while current P9 pins `d7581b7`; the installed `ModelBuilder.replicate` lacks P9's `xforms` argument. The failing P9 cloner file is unchanged by this branch.
- The focused PhysX live test timed out twice during Kit startup before pytest output; its concrete PhysX shared-interface coverage passed in the complete shared suite.

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] Documentation changes are not required for this behavior-only fix.
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective.
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package.
- [x] My name already exists in `CONTRIBUTORS.md`.

PR #6740, opened 2026-07-27
Articulation Reordering Part 11: Cache proxy selectors

AntoineRichard:
# Description

Adds an opt-in cached selector path to asset `find_*` APIs. With
`as_proxy=True`, articulation, rigid-object, and rigid-object-collection
finders return a per-asset cached `ProxyArray` backed by device-local
`wp.int32` storage. The same allocation can be consumed directly by Warp
writers or accessed as a zero-copy Torch view.

The existing finder result types remain available during the deprecation
period: omitted `as_proxy` retains the legacy result and warns, while
`as_proxy=False` explicitly retains it without warning. The cache is a
per-asset 128-entry LRU and is cleared when the asset is invalidated.

PhysX, Newton, and OVPhysX selector boundaries consume the cached Warp
allocation without materializing Torch. Repeated actuator and action-term
writer paths now retain proxy selectors, and the articulation benchmark
grid covers list, Torch int32/int64, Warp int32/int64, and cached proxy
representations.

This is stacked after #6652.

> [!NOTE]
> The articulation reordering parts target `develop`, so until the preceding
> parts merge this PR's diff includes them. The true P11 delta is
> [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p10-index-dtypes...antoiner%2Farticulation-reordering-p11-cached-proxy-finders).

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Documentation update

## Validation

- `./isaaclab.sh -f`
- Focused cache, resolver, action-consumer, benchmark-contract, and
  documentation suites: 186 passed.
- Changelog parser, validator, and formatter suites: 57 passed.
- Independent code review found no remaining critical, important, or minor
  findings.

GPU-backed method-grid timings were not rerun for P11 because the shared GPU
was occupied; the benchmark registration, synchronization, completeness, and
reporting contracts are covered by the focused tests.

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks with `./isaaclab.sh -f`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove the fix and feature are effective
- [x] I have added one changelog fragment for every touched package
- [x] My name already exists in `CONTRIBUTORS.md`

Documentation screenshots are not applicable.

PR #6674, opened 2026-07-22
Run deformable demos and tasks with OvPhysX

AntoineRichard:
## Summary

- build on the merged OvPhysX deformable support from #6665
- add `ovphysx` presets for the Franka soft-volume and cloth-surface lift environments
- run the deformables demo and deformable-object tutorial through the existing multi-backend launcher
- expose OvPhysX link/COM Jacobians, mass matrices, and gravity compensation through the shared articulation data API
- preserve fixed/floating-base layouts plus configured body and joint ordering for computed dynamics
- run both public task presets with their configured arm actions and replicated deformables in every environment
- resolve automatic physics selection before stage creation so only the selected backend registers its schemas
- enable supported OvPhysX + OVRTX rendering cases and add one shared reviewed golden set for legacy and OvStage
- retain explicit `isaacsim_physx` catalog labels while documenting `ovphysx` for both Franka deformable tasks

## Rendering baselines

- 23 reviewed PNGs: 12 Franka soft and 11 Franka cloth; requesting RGB also validates the RGBA output
- the same files are consumed by both legacy and OvStage OVRTX paths
- no thresholds were widened
- retained skips are limited to soft/cloth instance segmentation (`NVBUG#6463802`) and cloth motion vectors (`NVBUG#6489754`)

## Validation

- legacy OvPhysX + OVRTX matrix: `21 passed, 3 skipped, 128 deselected`
- OvStage OvPhysX + OVRTX matrix against the same PNGs: `21 passed, 3 skipped, 128 deselected`
- focused config, two-environment live soft/cloth task, full OvPhysX manager, and articulation dynamics/reordering sweep: `40 passed, 3 skipped`
- live DexSuite state-task check after the cloner refactor: `Isaac-Lift-KukaAllegro` resolved the homogeneous `CuboidCfg`, created two environments, reset contact sensors, and completed the first OvPhysX step
- both live OvPhysX demo entry points reached setup and advanced simulation without tracebacks
- deformable-object tutorial produced finite CUDA root positions through 24 simulated seconds
- full repository pre-commit suite passed on final head `74ce113386`

Validated with OvPhysX 0.5.9. The local uv environment is kitless, so the Kit-backed `test_simulation_context.py` module was not runnable there because `isaacsim` is not installed; PR CI covers the Kit jobs.

@mmichelis could you take another look at the final rendering baselines and review fixes? @huidongc the shared legacy/OvStage baseline expectation is now implemented.

PR #6652, opened 2026-07-21
Articulation Reordering Part 10: Support int64 selectors

AntoineRichard:
# Description

Allows articulation item selectors to remain either signed 32-bit or signed 64-bit arrays through the PhysX,
Newton, and OVPhysX Warp worker kernels. A validated kernel dispatcher selects the concrete Warp specialization
without a Torch-side cast or temporary allocation, while legacy direct raw-kernel launches retain their int32
contract.

PhysX and OVPhysX environment selectors are still normalized to int32 at external backend boundaries that require
that width. Joint, body, fixed-tendon, and spatial-tendon selectors preserve their supported input width through the
working kernels.

This is stacked after #6479.

> [!NOTE]
> The articulation reordering parts target `develop`, so until the preceding parts merge this PR's diff includes
> them. The true P10 delta is [this compare](https://github.com/AntoineRichard/IsaacLab/compare/antoiner%2Farticulation-reordering-p9...antoiner%2Farticulation-reordering-p10-index-dtypes).

## Type of change

- New feature (non-breaking change which adds functionality)

## Performance

- Paired raw-kernel measurements found no measurable int64 selector penalty: observed deltas ranged from -1.11% to
  +0.17%, within run variance.
- The complete 4096-environment method grid confirmed that prebuilt int32 and int64 selectors are effectively tied.
- Repeated Python-list selectors were substantially slower because they allocate and copy a new device selector on
  every method call; across stable pairs, list mode was 1.94x slower on PhysX and 1.67x slower on Newton than
  prebuilt int32 tensors.

## Validation

- `./isaaclab.sh -f` passed twice after rebasing onto the current P9 tip.
- Core kernel/dispatcher regression suites: 51 passed.
- PhysX raw and benchmark-selector suites: 27 passed.
- Newton raw and mock-interface suites: 20 passed.
- OVPhysX raw kernel suite: 49 passed.
- Simulation-backed int64 selector suites: PhysX 4 passed, Newton 4 passed, OVPhysX CUDA 4 passed.
- Full benchmark grids completed with 238 PhysX and 309 Newton recorded measurements.

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks with `./isaaclab.sh -f`
- [x] My changes generate no new warnings
- [x] I have added tests that cover the feature
- [x] I have added one changelog fragment for each touched package
- [x] My name already exists in `CONTRIBUTORS.md`

Documentation screenshots are not applicable.

PR #6784, opened 2026-07-29
Integrate articulation ordering caches and selectors

AntoineRichard:
# Description

This aggregate PR preserves the ancestry of the articulation reordering and selector stack so it can be reviewed and merged as one coherent change.

Merge order preserved in this branch:

- #6507 — articulation ordering element-kind refactor
- #6479 — P9 cached articulation reads across Newton, PhysX, and OVPhysX, including OVPhysX computed dynamics
- #6741 — rigid asset cache invalidation
- #6652 — P10 signed `int32`/`int64` selector support inside Warp work kernels
- #6740 — P11 per-asset cached `ProxyArray` finder selectors and explicit legacy return modes

The P10 commits remain ancestors of P11. The source PRs should stay open and unchanged until this aggregate PR merges; they can then be closed as superseded.

Related overlap: #6674. This branch includes the OVPhysX Jacobian, mass-matrix, and gravity-compensation side folded into P9.

Key outcomes:

- Articulation work kernels accept signed 32-bit and signed 64-bit item selectors without Torch conversion allocations.
- Raw kernel APIs use dtype-aware factories while retaining compatibility aliases for deprecated kernels.
- Stable articulation reads are cached for Newton, PhysX, and OVPhysX.
- OVPhysX exposes computed dynamics through the backend-agnostic articulation data API.
- Asset finders provide opt-in cached, immutable `ProxyArray` selectors with zero-copy Torch and Warp views and per-asset LRU ownership.
- Implicit legacy finder returns are deprecated; `as_proxy=False` retains them explicitly during migration.
- Rigid asset state/property writes invalidate same-timestamp derived caches correctly.
- Changelog entries are consolidated to one fragment per touched package.

## Asset performance before and after this PR

The table compares the PR merge-base (`a050ceade7`) with the pushed PR head (`00c8c25cb1`) on an RTX 5090 using CUDA, 4,096 instances, 100 warmup calls, and 1,000 measured calls per operation. It contains only timed asset method/property calls; application and benchmark startup are excluded.

Values are the equal-weight mean of the per-operation mean latencies for operations present on both revisions. The pre-PR `torch_tensor` mode and post-PR `torch_tensor_int32` mode both generate CUDA `torch.int32` selectors. Lower is better.

| Backend | Asset | Matched methods / data properties | API methods, before → after | Data properties, before → after |
|---|---|---:|---:|---:|
| PhysX | Articulation | 24 / 65 | 129.12 → 90.82 µs (-29.7%) | 6.94 → 5.22 µs (-24.8%) |
| PhysX | Rigid object | 13 / 40 | 82.48 → 54.65 µs (-33.7%) | 16.36 → 7.86 µs (-51.9%) |
| PhysX | Rigid-object collection | 13 / 69 | 218.32 → 131.36 µs (-39.8%) | 64.13 → 27.86 µs (-56.6%) |
| Newton | Articulation | 24 / 64 | 108.64 → 62.01 µs (-42.9%) | 6.48 → 4.70 µs (-27.4%) |
| Newton | Rigid object | 8 / 40 | 66.34 → 43.16 µs (-34.9%) | 15.21 → 10.28 µs (-32.5%) |
| Newton | Rigid-object collection | 8 / 78 | 83.08 → 50.47 µs (-39.3%) | 37.99 → 12.12 µs (-68.1%) |
| OVPhysX | Articulation | 24 / — | 237.21 → 209.79 µs (-11.6%) | Baseline failed preflight¹ |
| OVPhysX | Rigid object | 8 / 40 | 144.43 → 142.74 µs (-1.2%) | 18.57 → 15.36 µs (-17.3%) |
| OVPhysX | Rigid-object collection | 8 / 78 | 185.65 → 191.17 µs (+3.0%) | 46.75 → 41.31 µs (-11.6%) |

API-method latency improved in 8 of 9 backend/asset pairs. Data-property latency improved in every comparable pair.

¹ The merge-base OVPhysX articulation data benchmark fails during `body_com_jacobian_w` preflight, while the post-PR suite passes. That data row is excluded instead of mixing unmatched measurements. The new `int64` and Warp selector modes are also excluded from this before/after table because they have no pre-PR equivalent; the post-PR all-mode matrix passes across all nine backend/asset pairs.

## Type of change

- Documentation update

## Verification

- Aggregate ordering, dtype-kernel, proxy-cache, consumer, documentation, and benchmark matrix: **868 passed**.
- Shared articulation, rigid-object, and rigid-object-collection interface suites: **6,059 passed, 71 skipped, 68 expected failures**.
- OVPhysX computed dynamics: **4 CUDA tests passed** and **4 CPU tests passed** in separate processes due the backend process-global device lock.
- Changelog tooling: **90 passed**.
- `./isaaclab.sh -f`: all hooks passed at the pushed HEAD.
- Independent aggregate review: no remaining Critical, Important, or Minor findings.

Full Kit-backed PhysX/Newton live asset suites were not completed locally because the available Kit environment timed out during startup; CI should provide that coverage.

## Screenshots

Not applicable.

## Checklist

- [x] My changes generate no unexpected new warnings
- [x] I have added tests that prove the fixes and features are effective
- [x] I have added one changelog fragment under `source/<pkg>/changelog.d/` for every touched package


source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
```diff
@@ -1093,6 +1148,95 @@ def body_com_pose_b(self) -> ProxyArray:
...
+                        self._jacobian_body_user_to_backend,
+                        self._jacobian_joint_user_to_backend,
+                        self._num_base_dofs,
+                        has_body_ordering,
+                        has_joint_ordering,
+                    ],
+                    outputs=[self._body_com_jacobian_w.data],
+                )
+            else:
+                self._binding_read(TT.JACOBIAN, self._body_com_jacobian_w.data)
```
marcodiiga, line 1166: Could we transform these outputs into the same public DoF basis as `joint_vel` and joint effort (or gate the API until the binding provides that basis)? With one physically equivalent USD joint authored with `body0`/`body1` and its local frames reversed, I reproduced `J @ joint_vel` missing the measured COM velocity by 2.2; applying the affected column sign restored agreement to 1.2e-7. The mass matrix has the corresponding row/column mismatch (`S M S` is required when only a subset of joints is reversed), so `M @ qdd = tau` also fails in public coordinates. This violates the new `BaseArticulationData` dynamics contract for a valid joint orientation. Please add a reversed-body regression covering both Jacobian velocity and mass-matrix dynamics.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
```diff
@@ -210,12 +237,23 @@ def _reset_pose(self, from_link: bool = True) -> None:
                 self._body_link_pose_w,
                 self._body_link_pose_w_backend,
                 self._body_com_pose_w,
+                self._root_link_vel_w,
+                self._body_link_vel_w,
```
marcodiiga, line 226: A root-pose write can change child COM velocities at unchanged joint velocities because the joint axes and lever arms rotate. This invalidates the derived link-velocity cache here but leaves `_body_com_vel_w` and `_body_com_vel_w_backend` marked fresh, so the next same-timestamp read recomputes from stale source data. I reproduced it by caching CartPole velocities, rotating the root 90 degrees with nonzero joint velocity, and reading again: `body_com_vel_w` stayed byte-for-byte unchanged until those source buffers were invalidated (corrected max delta 1.2167). Please invalidate both sources here and add a same-timestamp root-rotation regression.

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -118,7 +119,7 @@ def __init__(
         self,
         cfg: ActuatorBaseCfg,
         joint_names: list[str],
-        joint_ids: slice | torch.Tensor,
+        joint_ids: slice | torch.Tensor | ProxyArray,
```
marcodiiga: Could we preserve `ActuatorBase.joint_indices` as `slice | torch.Tensor` and retain the cached `ProxyArray` separately for backend writers? All three backends now pass a proxy into partial actuator groups, changing this generated public property without deprecation. On this exact head, the existing OVPhysX fixed-base initialization test fails because the partial group exposes `ProxyArray` instead of `torch.Tensor`; dtype/device semantics also differ, and one-dimensional Torch indexing with the proxy raises `IndexError`. Passing the already available zero-copy `.torch` view across the actuator boundary would preserve compatibility while keeping the proxy inside the backend.

source/isaaclab/isaaclab/managers/scene_entity_cfg.py
```diff
@@ -230,14 +232,18 @@ def _resolve_body_names(self, scene: InteractiveScene):
...
             # -- if both are not their default values, check if they are valid
             # use find_sensors/num_sensors for ContactSensor, find_bodies/num_bodies for others
-            _find_fn = entity.find_sensors if hasattr(entity, "find_sensors") else entity.find_bodies
-            _num_bodies = entity.num_sensors if hasattr(entity, "num_sensors") else entity.num_bodies
+            is_sensor = hasattr(entity, "find_sensors")
+            _find_fn = entity.find_sensors if is_sensor else entity.find_bodies
+            _num_bodies = entity.num_sensors if is_sensor else entity.num_bodies
+            _find_kwargs = {"preserve_order": self.preserve_order}
+            if not is_sensor:
+                _find_kwargs["as_proxy"] = False
```
marcodiiga: This passes `as_proxy` to every body finder that lacks `find_sensors`, but `BaseJointWrenchSensor` and `BaseFrameTransformer` both expose `find_bodies(name_keys, preserve_order=False)` without that keyword. `SceneEntityCfg(..., body_names=...)` therefore raises `TypeError` for these sensors; the current `isaaclab_rl` CI reproduces it while constructing Isaac-Ant in three wrapper tests. Please gate `as_proxy=False` on an actually proxy-aware asset interface rather than this capability check, and cover both sensor interfaces.

source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -693,23 +691,21 @@ def _resolve_env_ids(self, env_ids: wp.array | torch.Tensor | list | slice | Non
...
         Raises:
             TypeError: If ``body_ids`` is an unsupported type.
         """
         if body_ids is None:
             return self._ALL_BODY_INDICES
         if isinstance(body_ids, torch.Tensor):
-            if body_ids.dtype == torch.int64:
-                body_ids = body_ids.to(torch.int32)
-            return wp.from_torch(body_ids.contiguous(), dtype=wp.int32)
+            return wp.from_torch(body_ids.contiguous())
```
marcodiiga: Could we unwrap `ProxyArray` here before the Torch/Warp checks? This PR now accepts the cached finder proxy in `BaseRigidObject.set_external_force_and_torque(..., body_ids=...)`, but that method forwards it here and still raises `TypeError`; I reproduced this on the current head. Returning `body_ids.warp` preserves zero-copy behavior and lets the dtype-dispatched kernel handle `int32`/`int64`.

marcodiiga (changes requested):
I found five issues that need resolution before merge, detailed inline: two OVPhysX dynamics/cache correctness problems, two proxy API compatibility breaks, and a selector-dispatch regression reproduced by CI. I reviewed exact head `6f649e11`; please ping me after the follow-up.

source/isaaclab/isaaclab/utils/wrench_composer.py
AntoineRichard: I intentionally took the output-only route instead of implicitly unwrapping here. ProxyArray is now rejected at low-level input boundaries with an instruction to pass .warp or .torch, while internal finder consumers select the explicit zero-copy view. This keeps the low-level API Torch/Warp-only and avoids a hidden ProxyArray ingestion path. The change is in 59e9e37052 and c8c5a9d4b6.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
AntoineRichard: Fixed in 2a605d7918. PhysX and OVPhysX now cache the authored/backend DoF sign difference at startup and normalize Jacobians, mass matrices with S M S, and generalized-force reads into the public basis in the existing read kernels. The reversed-body live regression runs on both backends and checks J @ joint_vel against measured COM velocity and mass-matrix kinetic energy against body kinetic energy. Newton is unchanged because it rejects reversed joints.

source/isaaclab/isaaclab/actuators/actuator_base.py
AntoineRichard: Fixed in c8c5a9d4b6. Finder proxies remain backend-internal; all three articulation backends pass the zero-copy Torch view, or slice(None), across the actuator boundary. ActuatorBase.joint_indices is again slice | torch.Tensor.

source/isaaclab/isaaclab/managers/scene_entity_cfg.py
AntoineRichard: Fixed in 57dc783d45. SceneEntityCfg no longer passes as_proxy through body or sensor finder interfaces. Its bookkeeping needs the legacy list representation, so both find_sensors and find_bodies are called with only preserve_order; no capability gate or signature introspection is needed.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation_data.py
AntoineRichard: Fixed in fec8ce4bfb. Root-pose writes now invalidate the COM-velocity source cache in PhysX and both the public and backend COM-velocity caches in OVPhysX. The shared writer regression now asserts that these source buffers are stale immediately after a same-timestamp root-pose write.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -4850,15 +4927,14 @@ def _resolve_env_ids(self, env_ids: Sequence[int] | torch.Tensor | wp.array | No
...
-            # Convert int64 to int32 if needed, as warp expects int32
-            if env_ids.dtype == torch.int64:
-                env_ids = env_ids.to(torch.int32)
-            return wp.from_torch(env_ids, dtype=wp.int32)
+        if isinstance(env_ids, ProxyArray):
+            raise TypeError("ProxyArray is output-only; pass .warp or .torch explicitly.")
         if isinstance(env_ids, list):
             return wp.array(env_ids, dtype=wp.int32, device=self.device)
+        if isinstance(env_ids, torch.Tensor):
+            return wp.from_torch(env_ids)
```
marcodiiga: Could we keep simulator-bound environment selectors in a backend-supported dtype? This now preserves CUDA `int64` input. The width-specialized Warp work kernels accept it, but the subsequent simulator writes do not: on this head, `isaaclab_physx` fails `test_write_joint_state_accepts_int64_selector[False-cuda:0]` at `set_dof_positions` with `An int32 or uint32 tensor is required`, and `isaaclab_ov` fails the matching CUDA test with `int64 index_tensor must be CPU-resident; GPU int64 not supported`. The mirrored OVPhysX resolver has the same issue, and other reset/write paths also pass this output to int32-only kernels. Please normalize at those backend boundaries while retaining width-aware dispatch where it is supported.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection_data.py
```diff
@@ -300,7 +328,8 @@ def body_com_pose_w(self) -> ProxyArray:
         The orientation is provided in (x, y, z, w) format.
         """
         if self._body_com_pose_w.timestamp < self._sim_timestamp:
-            wp.launch(
+            self._read_launch_cache.launch(
```
marcodiiga, line 325: Could we avoid caching this launch across input rebinding? `_WarpLaunchCache` records the first command's array pointers, while `body_link_pose_w` replaces `_body_link_pose_w.data` with a new cloned `wp.array` on every refresh (and `body_com_vel_w` does the same). The cached CUDA command therefore keeps reading the first pointer. On this head, the PhysX job has 20 CUDA-only rigid-object-collection failures while the CPU cases pass, including every `test_write_object_state_functions_data_consistency` variant. Please keep stable backing arrays or re-record when an input pointer changes, with a regression that refreshes and then reuses the cached derived read.

source/isaaclab/isaaclab/utils/warp/kernels.py
```diff
@@ -680,6 +683,34 @@ def reset_wrench_composer_index(
...
+    set_forces_to_dual_buffers_index, ("env_ids", "body_ids")
+)
+_ADD_FORCES_TO_DUAL_BUFFERS_INDEX_DISPATCHER = IndexKernelDispatcher(
+    add_forces_to_dual_buffers_index, ("env_ids", "body_ids")
+)
+_RESET_WRENCH_COMPOSER_INDEX_DISPATCHER = IndexKernelDispatcher(reset_wrench_composer_index, ("env_ids",))
+
+
+def set_forces_to_dual_buffers_index_kernel(
+    env_ids: wp.array | torch.Tensor, body_ids: wp.array | torch.Tensor
```
marcodiiga: Could we defer evaluation of these annotations, for example with `from __future__ import annotations`? The docs build intentionally mocks `warp`, so `wp.array` is a mock instance rather than a type there. Importing this module then evaluates `wp.array | torch.Tensor` and raises `TypeError: unsupported operand type(s) for |: 'array' and 'Tensor'`. The current Docs job consequently fails multiple autosummary imports and exits under warnings-as-errors.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation.py
```diff
@@ -4035,6 +4093,40 @@ def has_articulation_root_api(prim) -> bool:
         # mark data as ready
         self._data.is_primed = True
 
+    def _resolve_joint_dof_signs(self, stage: Usd.Stage) -> tuple[int, ...]:
+        """Resolve joint directions once from the source USD."""
+        source_prim = sim_utils.find_first_matching_prim(self.cfg.prim_path, stage=stage)
+        if source_prim is None:
+            return (1,) * self.num_joints
+        joint_prims_by_name = {}
+        for prim in Usd.PrimRange(source_prim):
```
marcodiiga: `Usd.PrimRange(source_prim)` does not descend through an instance proxy. For an instanceable articulation containing a reversed joint, `joint_prims_by_name` is empty, so the unmatched fallback below silently assigns `+1` and leaves the Jacobian, mass matrix, and generalized-force reads in the solver basis. Could this traverse instance proxies from the concrete binding root, for example with `Usd.PrimRange(source_prim, Usd.TraverseInstanceProxies())`, and could the regression cover an instanceable reversed joint? The current reversed-joint fixture is non-instanceable.

marcodiiga (changes requested):
Thanks for slimming this down and addressing the earlier findings. I re-reviewed the current head. The earlier actuator, SceneEntityCfg, and WrenchComposer issues look addressed, but I found four blocking correctness/CI issues: simulator-bound int64 selectors, cached CUDA launches after buffer rebinding, a Sphinx import failure, and instance-proxy traversal for reversed joints. Details are inline.

kellyguo11 (approved):
have we added some documentation for the reordering? when it's useful, how to use it...

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
```diff
@@ -88,7 +89,9 @@ def __init__(self, root_view: physx.ArticulationView, device: str):
         self._sim_timestamp = 0.0
         self._is_primed = False
         self._fk_timestamp = 0.0
-        self._cached_read_launches: dict[object, wp.Launch] = {}
+        self._read_launch_cache = _WarpLaunchCache(device)
+        self._joint_dof_signs = wp.ones(root_view.max_dofs, dtype=wp.int32, device=device)
```
kellyguo11, line 93: Could we keep the shared PhysX view contract in sync with this new property, or derive the DoF count from an existing contract such as `shared_metatype.dof_count`? `MockArticulationViewWarp` does not expose `max_dofs`, so `ArticulationData` construction now fails before the shared PhysX interface tests can run. Across the current core shards this produces 1,188 failures with `AttributeError: 'MockArticulationViewWarp' object has no attribute 'max_dofs'`. Please update the mock and cover construction through the shared interface suite.

source/isaaclab_newton/isaaclab_newton/benchmark/assets/runtime.py
```diff
@@ -256,8 +256,8 @@ def _refresh_rigid_object_data(mock_view, data, config) -> None:
     num_instances, num_bodies, device = config.num_instances, config.num_bodies, config.device
     root_transforms = np.random.randn(num_instances, 1, 7).astype(np.float32)
     root_transforms[..., 3:7] /= np.linalg.norm(root_transforms[..., 3:7], axis=-1, keepdims=True)
-    mock_view.set_mock_root_transforms(wp.array(root_transforms, dtype=wp.transformf, device=device))
-    mock_view.set_mock_root_velocities(
+    mock_view._root_transforms.assign(wp.array(root_transforms, dtype=wp.transformf, device=device))
```
kellyguo11: Could we preserve the mock-view setter interface while keeping the backing arrays stable? This refresh helper now reaches directly into `_root_transforms`, `_root_velocities`, and `_attributes`, but the supported recording view exposes `set_mock_*` methods instead. Consequently both Newton benchmark runtime regressions fail, beginning with `test_newton_rigid_object_refresh_regenerates_kinematics_and_mass_properties` at this line. If pointer stability is the goal, the setters can assign into stable buffers so callers do not need to depend on private storage.

source/isaaclab/isaaclab/utils/warp/kernels.py
```diff
@@ -7,8 +7,11 @@
 
 from typing import Any
 
+import torch
 import warp as wp
 
+from .index_kernel import IndexKernelDispatcher
```
kellyguo11: Could we preserve the standalone loading path used by `test_ray_caster_kernels.py`? That test loads this file through `spec_from_file_location`, so this new relative import runs without a package context and raises `ImportError: attempted relative import with no known parent package` during collection. Please either make the dispatcher import standalone-safe or update the test to import through the package without reintroducing its heavyweight runtime dependencies.

kellyguo11 (dismissed):
Agent reviews: Requesting changes for three additional CI blockers called out inline. The four existing unresolved threads covering simulator-bound int64 selectors, recorded CUDA launches after buffer rebinding, Sphinx annotation evaluation, and instance-proxy traversal remain blocking as well. Pre-commit and the pure selector/cache utility tests pass, but the current integration matrix is not merge-ready.

AntoineRichard:
Yes. This PR includes the [sim-to-sim policy transfer guide](https://github.com/isaac-sim/IsaacLab/blob/a4c6c4b415/docs/source/overview/core-concepts/physical-backends/sim-to-sim-policy-transfer.rst). It covers when joint/body reordering is useful, configuration and CLI overrides, ordering presets, public/backend mapping semantics, and raw backend access.

source/isaaclab_newton/isaaclab_newton/benchmark/assets/runtime.py
AntoineRichard: Fixed in 19e72debaa. The Newton refresh helpers use the set_mock_* interface; articulation setters assign into lazily allocated stable buffers, and the collection mock exposes the same stable setter surface. The two runtime regressions and the 41-test mock-view suite pass.

source/isaaclab/isaaclab/utils/warp/kernels.py
AntoineRichard: Fixed in 19e72debaa with the absolute isaaclab.utils.warp.index_kernel import. The standalone test remains unchanged and all 12 ray-caster kernel tests pass on CPU.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation_data.py
AntoineRichard: Fixed in 19e72debaa. MockArticulationViewWarp now exposes max_dofs from its existing DOF count, and its initialization test asserts max_dofs alongside shared_metatype.dof_count. The focused PhysX/OVPhysX kernel and mock-interface group passes 61 tests.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/articulation.py
AntoineRichard: Fixed in 19e72debaa. OVPhysX now uses Usd.PrimRange(source_prim, Usd.TraverseInstanceProxies()) for the one-time sign resolution. The new kitless regression creates a referenced instanceable articulation in memory and verifies that its reversed joint resolves to -1; the existing live Jacobian and mass-matrix dynamics regression also passes.

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
AntoineRichard: Fixed in 19e72debaa. PhysX and OVPhysX retain the original int32 or int64 selector for dtype-dispatched IsaacLab Warp work, then provide int32 environment IDs only at simulator-write boundaries. Hot state writes fill preallocated int32 scratch views, avoiding a per-call conversion allocation. The audit covers articulation, rigid-object, and collection state and property writes in both backends.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection_data.py
AntoineRichard: Fixed in 19e72debaa. PhysX rigid-object-collection pose, velocity, and acceleration refreshes now assign into the existing TimestampedBuffer arrays instead of rebinding them, so cached launches and ProxyArray views retain stable pointers. The existing rigid-object-collection state/data-consistency grid exercises the refresh-then-reuse path, so I did not retain a duplicate test.

source/isaaclab/isaaclab/utils/warp/kernels.py
AntoineRichard: Fixed in 19e72debaa. Only the three new dispatcher-factory union annotations are quoted, which defers their evaluation under Sphinx mocks without changing annotation handling for the Warp kernels in the rest of the module. A mocked torch/warp Sphinx import probe and all 12 standalone ray-caster tests pass.

AntoineRichard:
@marcodiiga @kellyguo11 All outstanding inline findings are addressed on head 19e72debaa, with a response on each thread. Focused verification is 131 passed and 1 skipped, the mocked Sphinx import passes, and the full pre-commit hook set is clean. Ready for re-review.

source/isaaclab_ovphysx/isaaclab_ovphysx/assets/articulation/kernels.py
```diff
@@ -40,6 +47,120 @@ def compute_soft_joint_pos_limits_func(
...
+) -> None:
+    """Write joint positions and materialize simulator indices."""
+    i, j = wp.tid()
+    env_id = wp.int32(env_ids[i])
+    joint_id = wp.int32(joint_ids[j])
+    value = in_data[i, j]
+    user_pos[env_id, joint_id] = value
+    if has_ordering:
+        backend_pos[env_id, user_to_backend[joint_id]] = value
+    if j == 0:
```
marcodiiga, line 69: Could we handle the zero-selected-joints case before passing the fused simulator-index scratch to the backend? `sim_env_ids[i]` is populated only from the `j == 0` lane. With `joint_ids=[]`, the launch dimension is `(N, 0)`, so the scratch retains stale or uninitialized values and the following simulator setter receives those values as environment indices. I reproduced this through `write_joint_position_to_sim_index`: prefilled scratch `[77, 88]` remained unchanged and the binding raised `IndexError` on index 77. The state and velocity writers and the PhysX mirror have the same pattern; an all-false PhysX joint mask also resolves to this empty index path. Please make an empty joint selection a no-op or materialize simulator IDs independently of the joint dimension, with regression coverage.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/kernels.py
```diff
@@ -29,4 +34,12 @@ def resolve_view_ids(
...
     """
     i, j = wp.tid()
-    view_ids[j * num_query_envs + i] = body_ids[j] * num_total_envs + env_ids[i]
+    view_ids[j * num_query_envs + i] = wp.int32(body_ids[j]) * num_total_envs + wp.int32(env_ids[i])
+
+
+_RESOLVE_VIEW_IDS_DISPATCHER = IndexKernelDispatcher(resolve_view_ids, ("env_ids", "body_ids"))
+
+
+def resolve_view_ids_kernel(env_ids: wp.array | torch.Tensor, body_ids: wp.array | torch.Tensor) -> wp.Kernel:
```
marcodiiga, line 43: Could we defer these annotations as well? Under the same Sphinx `torch`/`warp` mocks, importing this module evaluates `wp.array | torch.Tensor` and raises `TypeError: unsupported operand type(s) for |: 'array' and 'Tensor'`. The equivalent dispatcher factories in the shared kernels were deferred in this follow-up, but this newly added factory was missed. Quoting the annotations or enabling postponed evaluation, plus a mocked-import regression for this module, would cover the collection path too.

marcodiiga (changes requested):
Re-reviewed exact head `19e72debaaf66573f1a0040ad5ffe340300dbd1f`. The nine issues from my earlier two reviews are addressed. I found two remaining regressions, detailed inline: the fused simulator-index path mishandles empty joint selections, and one new PhysX dispatcher annotation still evaluates under Sphinx mocks.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/kernels.py
AntoineRichard: I checked this against the actual documentation path before changing it. The fresh [Build Latest Docs](https://github.com/isaac-sim/IsaacLab/actions/runs/30631181877/job/91157651263) job for this head passed with `-W`. That environment installs and loads real Torch and Warp; Sphinx's mock finder does not replace modules already present in `sys.modules`. A forced pure-mock direct import does reproduce the `TypeError`, but that is not the configuration used by our documentation build. Since both the runtime import and supported docs build pass, I would prefer not to add a deferred-annotation change and a bespoke mocked-import regression for this hypothetical path.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection.py
```diff
@@ -1391,27 +1395,55 @@ def _env_body_ids_to_view_ids(
...
+        cpu_view_ids = self._cpu_view_ids_view(count)
+        wp.copy(cpu_view_ids, view_ids)
+        return cpu_view_ids
+
+    def _sim_view_ids_view(self, count: int) -> wp.array:
+        """Return a cached prefix of the device view-index scratch."""
+        if count not in self._sim_view_ids_views:
+            self._sim_view_ids_views[count] = wp.array(
+                ptr=self._sim_view_ids.ptr,
+                shape=(count,),
```
marcodiiga, line 1468: Could we bounds-check `count` against the backing allocation before creating this view? `_sim_view_ids` is sized to `num_instances * num_bodies`, while `count` is derived from selector lengths and the public APIs accept repeated selectors. On a 2x3 collection, `env_ids=[0, 0, 0, 1, 1]` with all bodies requests a 15-element view over a six-element allocation; the CPU repro terminated with a core dump and CUDA reported an invalid argument. The previous path allocated exactly `count`. Please fall back to a correctly sized allocation or reject oversized selectors, and apply the same guard to the mirrored prefix-view helpers.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection.py
```diff
@@ -1391,27 +1395,55 @@ def _env_body_ids_to_view_ids(
...
             inputs=[env_ids, body_ids, num_query_envs, self.num_instances],
             outputs=[view_ids],
-            device=device,
+            device=self.device,
         )
-        return view_ids
+        if device != "cpu" or self.device == "cpu":
+            return view_ids
+        cpu_view_ids = self._cpu_view_ids_view(count)
+        wp.copy(cpu_view_ids, view_ids)
```
marcodiiga, line 1459: Could we establish completion before returning this pinned host view? `wp.copy` from CUDA into pinned host memory is asynchronous on Warp's current stream, and this result is handed directly to a CPU simulator setter. With the destination prefilled, I observed the previous values immediately after `wp.copy` and the expected indices only after synchronization. In this path, the blocking data reshape happens before this copy, so it does not order the subsequent host read. Please add an explicit stream/event handoff or use a synchronous ownership boundary; the OVPhysX mirror has the same ordering.

docs/source/testing/micro_benchmarks.rst
```diff
@@ -228,13 +235,31 @@ Defaults differ by file; use ``--help`` before creating a comparison command.
 Asset Input Modes
 ~~~~~~~~~~~~~~~~~
 
+Methods indexed only by ``env_ids`` retain two modes:
```
marcodiiga: Could we update this mode contract and validate unknown `--mode` values? The environment-only generators now emit five modes and no `torch_tensor`; the item generators emit nine modes, and several vary both `env_ids` and the item selector rather than holding environment IDs at Torch int32. As written, the documented `--mode torch_tensor` matches no workloads and the CLI then fails during finalization with `KeyError: 'mean'`. The sample `[TORCH_TENSOR]` output is stale as well. Please align the names, grid, and docs, and reject unmatched modes so commands cannot benchmark nothing.

docs/source/testing/micro_benchmarks.rst
```diff
@@ -146,10 +146,10 @@ Python environment must contain the backend being measured.
      - Asset benchmarks
      - Sensor benchmarks
    * - PhysX
-     - Launch Isaac Sim and use mocked PhysX views
+     - Run kitless with mocked PhysX views
```
marcodiiga, line 149: Could we qualify this requirement? The prerequisites table now says PhysX asset benchmarks run kitless with mocked PhysX views, and the asset runtime no longer launches the app. Isaac Sim is still required for the PhysX sensor benchmarks, but not for the asset benchmarks described above.

source/isaaclab_physx/isaaclab_physx/benchmark/assets/runtime.py
```diff
@@ -7,38 +7,59 @@
...
+    omni_physics.__path__ = []
+    omni_physics.tensors = MagicMock()
+    omni.physics = omni_physics
+    modules = {
+        "isaaclab_physx.physics": physics,
+        "omni": omni,
+        "omni.physics": omni_physics,
+        "omni.physics.tensors": omni_physics.tensors,
+    }
+    with patch.dict(sys.modules, modules):
```
marcodiiga: Could we avoid restoring the entire `sys.modules` mapping after these imports? `patch.dict(sys.modules, modules)` removes every module first imported inside the block when it exits. After `_load_runtime_symbols()`, `isaaclab_physx.physics` and the imported asset modules are no longer registered; re-importing the articulation module in the same kitless process then fails on the real `carb` dependency, while the globals retained here still refer to the first class objects. Limiting cleanup to the injected mock keys would avoid failed later imports and duplicate class identities.

marcodiiga (approved):
Re-reviewed exact head `00c8c25cb102309f550081fb06ca352cd446055d`. The two issues from my previous review are cleared: empty joint selections are now no-ops in both backends, and the supported warnings-as-errors documentation build completes with zero warnings at this SHA. I found five follow-up issues in the latest scratch, benchmark, and documentation delta and left them inline. Approving this revision; the broader CI matrix is still in progress.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection.py
AntoineRichard: Fixed in a83b390624. Selector cardinality is now checked under the existing shape-validation gate before any prefix view or launch. This rejects selector lengths beyond the asset dimensions (environment IDs are required to be unique) and applies the environment/body bounds through the shared asset bases. The existing PhysX/OVPhysX collection scratch regression now covers the oversized case.

source/isaaclab_physx/isaaclab_physx/assets/rigid_object_collection/rigid_object_collection.py
AntoineRichard: Fixed in a83b390624. Both PhysX and OVPhysX now synchronize Warp's current device stream immediately after the CUDA-to-pinned-host index copy and before returning the CPU view. The existing mirrored collection test verifies copy-before-sync ordering. A matched 1,000-iteration GPU full-grid comparison at 4,096 instances measured a localized ~1–3% aggregate cost across masses, CoMs, and inertias.

docs/source/testing/micro_benchmarks.rst
AntoineRichard: Fixed in a83b390624. The documentation now describes the five environment-selector modes and nine paired item-selector modes, and the sample output uses a current mode name. The CLI derives exact choices from the selected backend/component definitions, so an unmatched mode exits through argparse instead of reaching empty finalization; this case was folded into the existing invalid-argument test.

docs/source/testing/micro_benchmarks.rst
AntoineRichard: Fixed in a83b390624. The troubleshooting text now states that PhysX sensor benchmarks require Isaac Sim; the kitless PhysX asset benchmark requirement remains as documented in the prerequisites table.

source/isaaclab_physx/isaaclab_physx/benchmark/assets/runtime.py
AntoineRichard: Fixed in a83b390624. The runtime now restores only the four injected mock-module keys and leaves modules imported during setup registered. The existing runtime test now removes the articulation module, reloads the symbols, and verifies that the registered module retains the same Articulation class identity.

PR #6791, opened 2026-07-29
Add teleop workstation capability check

rwiltz:
# Description

Teleoperation envs target 45 FPS at 120 Hz physics, and the docs record a recommended workstation spec for it, but nothing measured the host against that spec. On under-spec machines the session simply runs slow, and the cause is usually attributed to CloudXR, the network, or the headset.

Measure the workstation when a teleop session starts and report any unmet requirement, both to the terminal and to the connected XR client, where it appears as a banner the operator can actually see while wearing the headset. The check is advisory and never blocks a session.

Thresholds are numeric rather than a list of approved CPU and GPU models, so equivalent hardware passes. CPU is weighted most heavily and gated on a short single-thread benchmark rather than core count: Pink IK and CPU-side physics are single-thread bound, so a machine with fewer fast cores outperforms one with many slow cores, and a core-count rule would fail exactly the machines that teleoperate well. The reference score was measured on the Threadripper 7960X named in the spec.

Delivery reuses the existing teleop control channel, whose outbound sink was built and then discarded. Connect it behind a ValueInput and drain queued payloads into it each step. Because the channel only flushes once connected, a notice queued at startup is delivered when the headset connects rather than dropped. This also adds a general send_client_message() for any future host-to-client message.

Fixes # (issue)

## Type of change

## Checklist


rwiltz:
@greptile

rwiltz:
@greptile

hougantc-nvda:
**[P2] Emit strict JSON for non-finite floats**

At [`session_lifecycle.py:776`](https://github.com/isaac-sim/IsaacLab/blob/41bc976976fc1178a5cca1f0eee3b309be7685b3/source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py#L776), Python `json.dumps()` defaults to `allow_nan=True`, so a message containing `math.nan` or `±math.inf` is queued with `NaN`/`Infinity` tokens. Those tokens are not valid JSON, and the browser `JSON.parse()` rejects the entire message.

Please serialize with `allow_nan=False` so the existing `ValueError` path drops invalid payloads. If non-finite telemetry is expected, normalize it at the producer to `null` (rendered as `N/A` or `—`) before sending. A parameterized regression covering `nan`, `inf`, and `-inf` would lock down the boundary.

arussell-nvidia:
The current isaaclab_teleop failure is a deterministic test/code issue, not runner capacity. The JUnit artifact shows 16 tests in 	est_host_client_messaging.py failing during _make_lifecycle: get_eventdispatcher() returns None, then TeleopSessionLifecycle.__init__ calls .observe_event. The new test module's no-Kit assumption therefore is not yet isolated in this CI mode. The smallest scoped fix appears to be mocking session_lifecycle.get_eventdispatcher (or a return object with observe_event) in the fixture before lifecycle construction, then rerunning the focused package job. I did not retry while the failure remains deterministic.

arussell-nvidia:
Two additional failures have now completed, but their JUnit signatures match contemporaneous unrelated PR runs and should not be attributed to this branch:

- `rendering-correctness-kitless`: `test_rendering_dexsuite_kuka_hetero_kitless` exited 139, exactly matching [run 30489735586](https://github.com/isaac-sim/IsaacLab/actions/runs/30489735586).
- `isaaclab_tasks [3/3]`: `test_environments_newton` was killed by SIGABRT, exactly matching [run 30481271525](https://github.com/isaac-sim/IsaacLab/actions/runs/30481271525) and [run 30480608093](https://github.com/isaac-sim/IsaacLab/actions/runs/30480608093).

That leaves the `isaaclab_teleop` failure described above as the branch-relevant deterministic issue. I have not retried the matching baseline failures.

arussell-nvidia:
Validated the targeted fix at head `cdb8001a`: the `isaaclab_teleop` check completed successfully ([run](https://github.com/isaac-sim/IsaacLab/actions/runs/30564214524/job/90944694741)). This clears the earlier deterministic lifecycle-construction failure where Kit modules were importable but `get_eventdispatcher()` returned `None`. The remaining matrix checks are still in progress.

source/isaaclab_teleop/isaaclab_teleop/session_lifecycle.py
```diff
@@ -47,6 +49,18 @@ def __dlpack_device__(self) -> tuple[int, int]: ...
 
 _DLDEVICE_CPU = 1
 
+_MAX_PENDING_CLIENT_MESSAGES = 32
+"""Bound on host-originated messages awaiting delivery to the XR client.
+
+Small on purpose: these are user-facing notices, so dropping the oldest when a
+client never connects is preferable to growing without limit."""
+
+_MAX_CLIENT_MESSAGE_BYTES = 64 * 1024
```
hougantc-nvda: I believe when I tried to send 1k of opaque data it would not be able to handle it.
Is this a number you tested?

Back then, I tested empirically and it was somewhere in the area of 940 bytes in my scenario. 
Has this changed?

rwiltz: Good catch let me verifiy

rwiltz: Verified the limit is ~940 and updated the code here.

hougantc-nvda (approved):
Tested and it is working as expected for the benchmark AMD Threadripper 5995wx cpu and Ada RTX 6000 GPU.

kellyguo11 (approved):

PR #6793, opened 2026-07-30
Add zero agent and random agent into CLI

StafaH:
# Description

Adds the zero agent and random agent into our isaaclab cli workflow. Existing scripts are left behind as stubs.

`uv run isaaclab zero_agent --task`
`uv run isaaclab random_agent --task`

## Checklist


source/isaaclab_rl/isaaclab_rl/entrypoints/simple_agents.py
```diff
@@ -0,0 +1,119 @@
...
+    args_cli = _parse_args(argv, policy)
+
+    torch.manual_seed(_SEED)
+
+    # parse configuration via Hydra (supports preset selection, e.g. env.sim.physics=newton_mjwarp)
+    env_cfg, _ = resolve_task_config(args_cli.task, "")
+
+    with launch_simulation(env_cfg, args_cli):
+        # override with CLI arguments
+        env_cfg.scene.num_envs = args_cli.num_envs
```
hujc7: Agent found bug: `--num_envs` defaults to `4` (line 111) and is applied unconditionally here, so an explicit Hydra override is silently discarded.

```console
$ isaaclab zero_agent --task Isaac-Cartpole env.scene.num_envs=64
# runs 4 environments, not 64 -- no warning
```

**Why:** `resolve_task_config` applies Hydra overrides to `env_cfg` before this line, so the assignment overwrites them. That override form is live in the repo -- `test_preset_kit_decision.py` uses `env.scene.num_envs=123`, `benchmark_hydra_resolve.py` uses `env.scene.num_envs=256`.

**Divergence from the package convention:**

| Entrypoint | argparse default | assignment |
|---|---|---|
| `play_sb3.py` | `None` | guarded |
| `play_rl_games.py` | `None` | guarded |
| `play_skrl.py` | `None` | guarded |
| `simple_agents.py` (this PR) | `4` | unguarded |

**Fix** -- two lines; applying either alone leaves the bug live:

```diff
@@ line 111 @@
-    parser.add_argument("--num_envs", type=int, default=_DEFAULT_NUM_ENVS, help="Number of environments to simulate.")
+    parser.add_argument("--num_envs", type=int, default=None, help="Number of environments to simulate.")

@@ line 69 @@
-        env_cfg.scene.num_envs = args_cli.num_envs
+        env_cfg.scene.num_envs = args_cli.num_envs if args_cli.num_envs is not None else env_cfg.scene.num_envs
```

This also drops the 4-environment default, which reads as a change separate from moving the agents into the CLI.

<details>
<summary>Alternative -- keep the small default, but let it lose to an explicit override</summary>

Requires `_parse_args` to also return `hydra_args`:

```python
if args_cli.num_envs is not None:
    env_cfg.scene.num_envs = args_cli.num_envs
elif not any(arg.startswith("env.scene.num_envs=") for arg in hydra_args):
    env_cfg.scene.num_envs = _DEFAULT_NUM_ENVS
```

</details>

Done when `isaaclab zero_agent --task Isaac-Cartpole env.scene.num_envs=64` reports 64 environments.

source/isaaclab_rl/isaaclab_rl/entrypoints/simple_agents.py
```diff
@@ -0,0 +1,119 @@
...
+        print(f"[INFO]: Gym observation space: {env.observation_space}")
+        print(f"[INFO]: Gym action space: {env.action_space}")
+        # reset environment
+        env.reset()
+        # simulate environment
+        # keep running while any visualizer is open
+        sim = env.unwrapped.sim
+        device = env.unwrapped.device
+        zero_actions = torch.zeros(env.action_space.shape, device=device)
+        while True:
```
hujc7: Agent review: two observations on this loop.

**1. It re-implements an existing helper.** `SimulationContext.is_headless_or_exist_active_visualizer()` is the exact negation of the break condition, and is already the idiom in ~18 scripts under `scripts/demos/`:

```diff
-        while True:
-            if sim.visualizers:
-                # visualizer mode: run until the visualizer window is closed
-                if not any(v.is_running() and not v.is_closed for v in sim.visualizers):
-                    break
-            # run everything in inference mode
+        # step while a visualizer window is still open (or none exist, e.g. headless)
+        while sim.is_headless_or_exist_active_visualizer():
             with torch.inference_mode():
```

**2. With `--viz none` there is no exit path at all.** `_resolve_visualizer_cfgs` returns `[]` when visualizers are disabled, `initialize_visualizers` then returns early leaving `_visualizers` empty, so `if sim.visualizers:` is never true and the `break` is unreachable.

That was tolerable while these were only scripts. It matters more now: `zero_agent()` and `random_agent()` in `api.py` are documented to return a process exit code, and under `--viz none` they cannot return.

The deleted `MAX_STEPS = 100` was the vestige of that bound -- it was never wired up. A `--max_steps` argument would restore it, matching `scripts/demos/sensors/ppisp_camera.py` and the MPM demos. It would also make these commands cheap to smoke-test, since nothing in `.github/`, `tools/`, or the test suites currently executes either agent.

hujc7 (commented):
Agent review: nice consolidation -- one implementation behind three thin layers (typed request, CLI dispatcher, executables), matching the existing train/play shape.

**Verified locally:** `pre-commit` clean across all 15 changed files; the `cartpole/__init__.py` literalinclude line numbers land on both `gym.register` blocks.

> [!IMPORTANT]
> **Blocker** -- the new `--num_envs` default of `4` is applied unconditionally, so an explicit `env.scene.num_envs=<N>` on the command line is silently discarded. Details inline on `simple_agents.py`.

**Also inline:** a loop simplification that closes an exit-path gap in the new public API, and one typing nit.

**Beyond the diff:** no CI job or test exercises either agent end-to-end -- the delegation test asserts on file text only, so the new command paths are unexercised.

source/isaaclab_rl/isaaclab_rl/entrypoints/dispatch.py
```diff
@@ -40,6 +40,44 @@ def run_play_cli(argv: list[str] | None = None) -> int:
...
+    """Dispatch command-line arguments to the zero-action agent."""
+    return _run_simple_agent_cli("zero", argv)
+
+
+def run_random_agent_cli(argv: list[str] | None = None) -> int:
+    """Dispatch command-line arguments to the random-action agent."""
+    return _run_simple_agent_cli("random", argv)
+
+
+def _run_simple_agent_cli(policy: str, argv: list[str] | None) -> int:
```
hujc7: Agent suggestion: `policy` is typed `str` here, but `simple_agents.run()` declares `policy: PolicyName`.

The module already has `from __future__ import annotations`, so the alias imports under `TYPE_CHECKING` with no change to import cost:

```diff
 import argparse
 import importlib
 import runpy
 import sys
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .simple_agents import PolicyName
```

```suggestion
def _run_simple_agent_cli(policy: PolicyName, argv: list[str] | None) -> int:
```

kellyguo11 (approved):

PR #6804, opened 2026-07-30
[CI] Stabilize the Newton environment smoke tests and run them kit-less

hujc7:
# Description

The `isaaclab_tasks` shards intermittently aborted with `SIGSEGV` or `SIGABRT` while creating a
Newton environment — measured at **~5% per environment build**. The abort is glibc catching a
corrupted heap (`malloc(): unaligned tcache chunk detected`) at an unrelated later allocation, so
pytest recorded `CRASHED` with no Python traceback and no failing node.

Root cause is a race in OpenUSD's physics parser: collecting collider descriptors for several
colliders under one rigid body reallocates a shared `std::vector` across worker threads.

> *"Fixed a bug in UsdPhysicsParsingUtility which could cause a race condition and crash when parsing
> multiple colliders under the same rigid body in a multi-threaded context"* — OpenUSD CHANGELOG
> `[26.05] - 2026-04-24`, upstream issue 4002

This PR contains three related changes.

## 1. Work around the parser race

Limit OpenUSD's work-thread pool for this suite, mirroring the workaround already in
`source/isaaclab_tasks/test/rendering_test_utils.py` and carrying the same removal note.

Applied at **module scope** rather than in a fixture: kit-less, OpenUSD builds its task arena while
tests are collected, and `WorkSetConcurrencyLimit` does not shrink an arena that already exists
(measured — it still reported 128 workers and still crashed).

## 2. Run the suite kit-less

The MJWarp preset is a kit-less backend, so `AppLauncher` is dropped and the suite filters to tasks
that run without Kit. The split is derived from `scan(...).needs_kit` — the same decision
`launch_simulation()` makes — so a task follows its own config instead of a hand-maintained list.

Tasks carrying Kit camera sensors are excluded; `test_environments.py` already smoke-tests each of
them on its default backend, so no task loses coverage.

`get_isaac_sim_version()` in `_run_environments` is now called only when `create_stage_in_memory` is
set, since it imports `isaacsim` — that unconditional call was what previously forced Kit.

## 3. Make the physics preset selection actually apply

Selecting a preset in this suite had no effect. `parse_env_cfg` resolves every `PresetCfg` wrapper to
its `default` before returning, and the follow-up `apply_overrides` call had no choice node left to
act on, so each task ran whichever backend it defaulted to.

`parse_env_cfg` now takes the selection and applies it *while* the wrappers are resolved, which is
what `register_task_to_hydra` already does. Default `presets=()` keeps every existing caller
unchanged. The raw-config / `collect_presets` / `apply_overrides` dance in the test util collapses to
a single `parse_env_cfg(...)` call.

Consequence: `Isaac-Velocity-Flat-G1` and `Isaac-Velocity-Flat-Spot` now run **Newton** in this
suite rather than PhysX. Both pass.

## Evidence

Standalone reproducer — `pxr` only, no Isaac Lab, Newton, CUDA or Kit; 15 processes x 40 parses of
the Franka Reach stage:

| usd-core | Has the 26.05 fix | Crashed processes |
|---|---|---|
| 25.11 (current pin) | no | **14 / 15** |
| 26.3 | no | 10 / 15 |
| 26.8 | yes | **0 / 15** |

Crash rate against OpenUSD's thread count, which is what this PR limits:

| `PXR_WORK_THREAD_LIMIT` | 1 | 2 | 4 | 8 | 16 | 32 | 128 |
| crashed procs (of 15) | **0** | **0** | 10 | 10 | 12 | 11 | 14 |

Captured stack, previously unavailable because the process died without a Python traceback:

```
#0  moveDescsToDict<UsdPhysicsRigidBodyDesc>(...)        pxr/UsdPhysics/_usdPhysics.so
#1  LoadUsdPhysicsFromRange(stage, paths, callback, ...) libusd_ms
    newton/_src/utils/import_usd.py:546  parse_usd
    isaaclab_newton/cloner/newton_clone_utils.py:188  _build_source_builder
    isaaclab/scene/interactive_scene.py:187  __init__
```

Local verification:

| Check | Result |
|---|---|
| `test_parse_env_cfg_presets.py` (new) | 3 passed — fails on develop |
| kit-less collection, no `isaacsim` import | 58 tests in 7.3s |
| `Isaac-Reach-Franka`, `-OSC`, `Isaac-Reorient-Cube-Shadow-Direct` | 6 passed |
| `Isaac-Velocity-Flat-G1`, `-Spot` on Newton for the first time | 4 passed |

## Why not raise the usd-core pin

`usd-core>=26.5` is the real fix, but it cannot be delivered yet: `newton[importers]` declares
`usd-core>=25.5,<26.5`, whose exclusive bound excludes the fixing release. `[tool.uv]
override-dependencies` gets `uv sync` past that, but `isaaclab.sh --install` resolves
`[project.dependencies]` with **pip**, which has no override mechanism (see the comment at
`source/isaaclab/isaaclab/cli/commands/install.py:341`), so the Docker image build fails with
`ResolutionImpossible`.

**Raising `usd-core<26.5` to `<27` in `newton[importers]` is the prerequisite.** Once that lands, the
pin can move and both this workaround and the one in `rendering_test_utils.py` can be removed.

## Notes for review

- This is a mitigation for the crash, not a fix. Users creating Newton environments outside this
  suite remain exposed until the pin can move.
- Change 3 alters what CI exercises: any task whose default is not `newton_mjwarp` now genuinely runs
  Newton in this suite. G1 and Spot pass; other suites are unaffected because `presets` defaults to
  an empty selection.
- `apply_overrides` still accepts a `presets` argument that it never reads. It no longer has callers
  in `env_test_utils.py`, but `rendering_test_utils.py` and `test_hydra.py` still pass it. Removing
  it is a separate cleanup across those call sites.

## Type of change

## Checklist

- [x] Pre-commit hooks pass
- [x] Changelog fragment added
- [ ] Full CI validation

### DIFF
--- a/source/isaaclab_tasks/changelog.d/jichuanh-newton-env-tests-kitless.skip
+++ b/source/isaaclab_tasks/changelog.d/jichuanh-newton-env-tests-kitless.skip
@@ -1,0 +1,3 @@
+Test-only: ran the Newton environment smoke tests kit-less, split the Kit-requiring tasks into
+their own file, and limited OpenUSD's work-thread pool to work around the usd-core<26.5 physics
+parser race that aborted the process during the Newton USD import.
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/parse_cfg.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/parse_cfg.py
@@ -13,6 +13,7 @@ import inspect
 import os
 import re
 import warnings
+from collections.abc import Sequence
 from typing import TYPE_CHECKING
 
 import gymnasium as gym
@@ -144,7 +145,11 @@ def load_cfg_from_registry(task_name: str, entry_point_key: str) -> dict | objec
 
 
 def parse_env_cfg(
-    task_name: str, device: str = "cuda:0", num_envs: int | None = None, use_fabric: bool | None = None
+    task_name: str,
+    device: str = "cuda:0",
+    num_envs: int | None = None,
+    use_fabric: bool | None = None,
+    presets: Sequence[str] = (),
 ) -> ManagerBasedRLEnvCfg | DirectRLEnvCfg:
     """Parse configuration for an environment and override based on inputs.
 
@@ -155,6 +160,10 @@ def parse_env_cfg(
         use_fabric: Whether to enable/disable fabric interface. If false, all read/write operations go through USD.
             This slows down the simulation but allows seeing the changes in the USD through the USD stage.
             Defaults to None, in which case it is left unchanged.
+        presets: Preset names to select while resolving :class:`PresetCfg` wrappers, mirroring the
+            Hydra ``presets=`` CLI tokens. Defaults to an empty selection, which resolves to each
+            wrapper's default. A selection must be passed here rather than applied afterwards,
+            because the returned config no longer carries the wrappers to choose from.
 
     Returns:
         The parsed configuration object.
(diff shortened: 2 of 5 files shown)


AntoineRichard (approved):
Some comments could be shorter, but if it works it works.

AntoineRichard (changes requested):
Should we close this now that @maxkra15 PR got in?

PR #6554, opened 2026-07-16
Fix O(N)/O(N^2) camera-pose resolution stall under PhysX + Newton ren…

yts-nv:
## Description

At high environment counts (e.g. `Isaac-Lift-KukaAllegro-Camera` at 8192 envs) with
`presets=physx,newton_renderer`, the PhysX-backed camera-sensor pose update stalls for
minutes per training iteration, causing benchmarks to time out with no results.

Root cause: two per-prim Python hotspots on the PhysX -> Newton-renderer camera path,
both scaling with the *whole-scene* prim/body count rather than the handful of managed cameras:

1. `FabricFrameView._compute_fabric_indices_for` rebuilt a `{str(Sdf.Path): index}` map over
   **every** prim in the Fabric selection (~1.14M prims for an 8192-env scene) on **every env
   reset**, only to resolve the managed camera prims.
   -> Cache the map per selection; rebuild only when the selection size changes.

2. `SceneDataProvider.create_mapping` used an O(N^2) `list.index()` lookup per input transform
   (~200k rigid bodies).
   -> Replace with an O(N) dict lookup.

## Results

Measured on `Isaac-Lift-KukaAllegro-Camera`, `presets=physx,newton_renderer,single_camera,rgb64`,
`num_envs=8192` (RTX PRO 6000 Blackwell):

| | before | after |
|---|---|---|
| per-iteration | ~220 s | ~53 s |
| outcome | timeout, no results | completes, KPIs written |

Training rewards are unchanged (verified at `num_envs=256`; per-iteration mean reward identical).

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Performance improvement

## Checklist

- [ ] I have run the pre-commit checks with `./isaaclab.sh --format`
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml`
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6805, opened 2026-07-30
perf: scale FabricFrameView selections to the view, not the stage

pv-nvidia:
# Description

Camera world-pose resolution stalled at high environment counts on the PhysX backend, badly enough to time out benchmarks. There are two independent bottlenecks; the PR also adds a lifecycle hook that came out of review.

### 1. `FabricFrameView` resolved prim paths on the host, against the whole stage

Cameras read their poses from Fabric. To do that the frame view needs to know where each camera's data sits in Fabric memory, so it builds a lookup table from "camera number" to "Fabric slot".

Building that table was very slow. The view selected prims by requiring the Fabric world and local matrix attributes — but every xformable in the stage carries those, so the selection matched the **entire stage** (~1.1M prims at 8192 environments), not the view's handful of cameras. Finding its own prims in that list then meant building a Python dictionary over all of it, on the host:

```python
path_to_idx = {str(p): i for i, p in enumerate(selection.GetPaths())}
```

That ran twice per rebuild (once for the children, once for their parents), and rebuilt whenever `PrepareForReuse()` reported a bucket change — in practice on every environment reset.

The cost of that dictionary scales with the size of the whole scene, not with the number of cameras: it converts every path in the stage to a Python string and inserts it. Measured directly, one full rebuild takes 10 ms on a bare stage but **556 ms once 100k unrelated prims are present** (table below), and it grows linearly from there — seconds of host work per reset at the ~1.1M prims of an 8192-environment scene.

**Fix:** author a private per-view `uint` index attribute on each managed prim (and on each unique parent), holding that prim's view index. Every selection requires the matching index attribute, so **selections resolve to exactly the view's prims**. The view-to-Fabric slot mapping is then rebuilt by inverting that attribute in a single Warp kernel launch:

```python
fabric_slots[view_indices[fabric_slot]] = fabric_slot
```

That is `O(view)` device work with no host-side path resolution and no Python objects created per frame.

There is also **no cache**, so there is nothing to invalidate. The table is rebuilt from live Fabric data on every access, which means a bucket reorder can never leave a stale mapping behind. Selections are still checked with `GetCount()`, which is exact here: the index attribute belongs to one view, so the count can only change if one of that view's prims actually disappeared — and then the view raises a clear error instead of silently reading the wrong prim.

Attribute names embed a process-wide monotonic uid, so a dead view's leftover attributes can never satisfy a live view's selection — and since review, they are removed outright when the view is released (section 3). Parent reads get their own read-only selection, keeping the child RO/RW flip semantics from #5677 intact.

### 2. `SceneDataProvider.create_mapping` did a linear scan per item

```python
for i, path in enumerate(input_paths):
    mapping[i] = paths.index(path)   # O(N) scan, per path
```

**Fix:** build the reverse dict once (first occurrence wins, matching `list.index` semantics), then resolve each path in `O(1)`. Measured over 200k paths, preserving first-occurrence semantics costs nothing versus an unguarded dict comprehension, so the fix carries no behaviour change.

### 3. Index attributes are removed when a view is released (from review)

The per-view index attributes from fix 1 were initially never removed, so views recreated over the same prims on a long-lived stage accumulated attribute pairs, widening those prims' Fabric buckets. Measured over 8 successive views on one stage (512 prims), attributes on a single prim went 1 → 8 and first-access rose monotonically from 120 ms to 173 ms.

**Fix:** `FabricFrameView.close()` removes the view's tags (measured at 4.7 µs per prim, ~77 ms for an 8192-env view) and is safe to call more than once. `BaseFrameView` gains a no-op `close()` so callers can close any backend uniformly. Every in-repo callsite now closes its view: `Camera` when invalidated **and when dropped** (`__del__`), `prim_world_positions` after each per-env view it creates in a loop, and `SceneAsset` before replacing its cached view. Views dropped without `close()` are cleaned up best-effort from `__del__`, following the shutdown-safe idiom the env classes already use (`sys` bound as a default argument; nothing runs during interpreter finalization, where Fabric dies with the process anyway), and a warning names the view so the missing `close()` call can be fixed.

## How the problem was found

The stall was invisible in Tracy. Comparing captures before and after the commit that introduced it, Kit's own per-frame work is unchanged — `App Update` totals 65.1 s before and 62.3 s after, over ~120 frames. All the lost time sits in the gaps *between* frames, where the main thread is blocked in Python and Kit records nothing.

A `py-spy` capture found the cause: `_compute_fabric_indices_for` accounted for **11.86% of samples (23.2 s of a 195 s run)**, reached through `camera.reset()` → `get_world_poses()` → the selection rebuild. Every frame on that chain carries essentially the same share, so the whole cost is that one function.

Fixes nvbug 6535498.

## Type of change

- New feature (non-breaking change which adds functionality) — `FrameView.close()`

### End-to-end symptom

Task `Isaac-Lift-KukaAllegro-Camera`, 8192 environments, `presets=physx,isaacsim_rtx_renderer,duo_camera`. Commit `69888c34e471` (which introduced the problem) against its parent `e5f99d320338`:

| | Parent | With the problem |
| --- | ---: | ---: |
| Mean environment step | 699 ms | 7193 ms |
| Slow steps (over 3 s) | 1 of 100 | 28 of 100, averaging 18.9 s |
| GPU utilization | 23.4% | 4.7% |

GPU and CPU utilization both drop during the slow steps, which is what a stall looks like — the pipeline is waiting, not doing extra work.

With this PR, the same task at 8192 environments steps at **mean 1.554 s, peak 1.663 s** — peak within 7% of mean, i.e. the stalls are gone. (Measured on an RTX A6000, so the absolute step time is not comparable to the table above, which was captured on different hardware; the stall signature is the comparison that matters.)

### Selection size — the actual fix

L40, `cuda:0`, 1024-prim view (1024 children + 1024 parents). "Filler" is xformable prims **not** in the view but carrying Fabric matrices — i.e. the rest of a real scene.

| filler prims | `develop` selection | this PR |
| ---: | ---: | ---: |
| 0 | 2,048 | 1,024 |
| 10,000 | 12,048 | 1,024 |
| 100,000 | 102,048 | **1,024** |

`develop` scales with the stage; this PR is pinned to the view.

### Slot-mapping rebuild (mean ms, the path hit on every reset)

| filler prims | `develop` child | `develop` parent | `develop` full RO rebuild | PR child | PR parent |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 0 | 4.67 | 5.40 | 10.24 | 0.137 | 0.191 |
| 10,000 | 24.05 | 24.69 | 62.47 | 0.139 | 0.193 |
| 100,000 | 274.21 | 270.47 | **556.41** | **0.137** | **0.190** |

`develop` grows linearly with stage size — **556 ms per rebuild** at 100k filler prims. This PR is flat at **~0.14 ms**, a **~2000×** reduction, and unchanged from 0 to 100k filler prims. First access at 100k filler: 1824 ms → 383 ms.

### `create_mapping` (pure Python, reversed path order)

| N | `develop` | this PR | speedup |
| ---: | ---: | ---: | ---: |
| 1,000 | 0.0038 s | 0.0001 s | 27× |
| 5,000 | 0.0862 s | 0.0007 s | 120× |
| 20,000 | 1.3241 s | 0.0030 s | 434× |
| 50,000 | 8.5507 s | 0.0092 s | **929×** |
| 200,000 | (not run) | 0.0467 s | — |

Outputs verified identical at every N. At the ~200k rigid bodies of an 8192-environment scene the old path takes minutes.

## Tests

| Suite | Result |
| --- | --- |
| `isaaclab_physx/test/sim/test_views_xform_prim_fabric.py` | 79 passed, 4 skipped (`cpu` + `cuda:0` on an L40) |
| `isaaclab/test/sim/test_views_xform_prim.py` (USD contract) | 63 passed |
| `isaaclab_newton/test/sim/test_views_xform_prim_newton.py` | 58 passed |
| `isaaclab/test/sensors/test_camera.py` (main consumer, incl. the new `close()` call) | 35 passed, 0 missing-`close()` warnings |
| `isaaclab/test/utils/warp/test_proxy_array.py` | 81 passed |

New tests:

- `test_selections_match_only_the_view_prims` asserts each selection matches exactly the prims the view manages. Verified to fail without the fix — with the selections unscoped it reports `matched 8 prims, expected 4`, because the child selections pick up the parents too.
- `test_close_removes_index_attributes` asserts the tag-scoped selection drops to zero after `close()` and that a second `close()` is a no-op.
- `test_garbage_collection_removes_index_attributes_and_warns` asserts a view dropped without `close()` still removes its tags on collection and logs the warning.

`test_fabric_rebuild_after_topology_change` was updated to drive the new refresh paths instead of the removed `_rebuild_{ro,rw}_arrays`.

The 4 skips are environmental: empty device parameter sets, and Fabric hierarchy bindings that are unavailable in a headless experience.

**Coverage gap:** `test_physx_scene_data_backend.py` and `test_ovphysx_scene_data_backend.py` both skip at collection on the machine used here, so the `create_mapping` change is exercised only by the standalone benchmark above, not by the test suite. Worth a look in CI, where those backends are available.

## Screenshots

Not applicable — no visual change. The benchmarks above cover the behaviour change.

## Notes for reviewers

**Review feedback addressed so far:** the index-attribute accumulation is fixed by `close()` (section 3); the opposite-space recompute no longer refreshes the same selection twice on writer-scope exit; `create_mapping` keeps `list.index`'s first-occurrence semantics for duplicate paths (measured free at 200k paths); and docstrings/changelogs were trimmed to the observable contract.

**The `isaaclab`/`isaaclab_physx` changelog fragments are still named `fix-physx-newton-camera-pose-scaling.rst`**, inherited from #6554 — nothing here touches Newton or scaling. The new `isaaclab_mimic` fragment uses the branch slug. Happy to rename if reviewers prefer.

**A handful of tests still drop views without `close()`** and now log the new warning during runs; harmless, cleanup to follow.

EDIT: fixed — the only remaining warning is the one `test_garbage_collection_removes_index_attributes_and_warns` asserts on. `Camera` previously closed its view only on invalidation, so a dropped camera still went through `__del__`; it now closes in `Camera.__del__` too. `test_camera.py` went from 3 warnings to 0.

## Relationship to #6554

#6554 found the same two slow paths. This PR takes its `SceneDataProvider` fix unchanged (credited with a `Co-authored-by` line) and replaces its Fabric frame view fix.

#6554 kept the whole-stage selection and the Python dictionary, and cached the result, reusing it while the number of selected prims stayed the same. Two problems with that. The cache key is unsafe: an equal prim count does not mean the prims or their order stayed the same, so after a bucket reorder or a same-count membership change the cached indices point at the wrong prims and cameras silently read another prim's transform. And it treats the symptom — the underlying operation is still proportional to the size of the whole scene, just performed less often.

Making the selection small removes the need for a cache at all, so both problems go away.

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
+++ b/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
@@ -1,0 +1,6 @@
+Fixed
+^^^^^
+
+* Fixed a quadratic path lookup in :class:`~isaaclab.scene_data.SceneDataProvider` transform
+  mapping that stalled setup at high rigid-body counts (thousands of environments). The
+  per-item ``list.index`` scan is now an ``O(N)`` dictionary lookup.
--- a/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
+++ b/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
@@ -4,3 +4,10 @@ Fixed
 * Fixed :class:`~isaaclab.scene_data.SceneDataProvider` transform mapping stalling
   at high rigid-body counts, which delayed setup by minutes in scenes with
   thousands of environments.
+
+Added
+^^^^^
+
+* Added :meth:`~isaaclab.sim.views.BaseFrameView.close` to release backend state
+  authored by a frame view. Backends also release best-effort on garbage
+  collection, but only an explicit close is deterministic.
--- a/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
+++ b/source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
@@ -1,6 +1,6 @@
 Fixed
 ^^^^^
 
-* Fixed a quadratic path lookup in :class:`~isaaclab.scene_data.SceneDataProvider` transform
-  mapping that stalled setup at high rigid-body counts (thousands of environments). The
-  per-item ``list.index`` scan is now an ``O(N)`` dictionary lookup.
+* Fixed :class:`~isaaclab.scene_data.SceneDataProvider` transform mapping stalling
+  at high rigid-body counts, which delayed setup by minutes in scenes with
+  thousands of environments.
--- a/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
+++ b/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
@@ -212,14 +212,13 @@ class SceneDataProvider:
             paths or if no mapping is needed.
         """
         if input_paths := self.backend.transform_paths:
-            # Build a path -> output-index map once (first occurrence wins, to match
-            # ``list.index`` semantics), then resolve each input path in O(1).  This
-            # was an O(N^2) ``paths.index(path)`` linear scan per input path -- minutes
-            # at the ~200k rigid bodies of an 8192-env scene.
-            path_to_out: dict[str | None, int] = {}
-            for out_idx, out_path in enumerate(paths):
-                if out_path not in path_to_out:
-                    path_to_out[out_path] = out_idx
+            # Build a path -> output-index map once, then resolve each input path in
+            # O(1).  This was an O(N^2) ``paths.index(path)`` linear scan per input
+            # path -- minutes at the ~200k rigid bodies of an 8192-env scene.
+            # A plain dict comprehension keeps the LAST occurrence of a duplicate
+            # where ``list.index`` kept the first; duplicate paths are invalid input
+            # and produce a wrong mapping under either choice, so take the fast form.
+            path_to_out = {out_path: out_idx for out_idx, out_path in enumerate(paths)}
             mapping = [path_to_out.get(path, -1) for path in input_paths]
             if not np.array_equal(mapping, np.arange(len(input_paths))):
                 return wp.array(mapping, dtype=wp.int32)
--- a/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
+++ b/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
@@ -5,7 +5,6 @@
 
 from __future__ import annotations
 
-import contextlib
 import logging
 import re
 from collections import deque
@@ -213,10 +212,15 @@ class SceneDataProvider:
             paths or if no mapping is needed.
         """
         if input_paths := self.backend.transform_paths:
-            mapping = [-1] * len(input_paths)
-            for i, path in enumerate(input_paths):
-                with contextlib.suppress(ValueError):
-                    mapping[i] = paths.index(path)
+            # Build a path -> output-index map once (first occurrence wins, to match
+            # ``list.index`` semantics), then resolve each input path in O(1).  This
+            # was an O(N^2) ``paths.index(path)`` linear scan per input path -- minutes
+            # at the ~200k rigid bodies of an 8192-env scene.
+            path_to_out: dict[str | None, int] = {}
+            for out_idx, out_path in enumerate(paths):
+                if out_path not in path_to_out:
+                    path_to_out[out_path] = out_idx
+            mapping = [path_to_out.get(path, -1) for path in input_paths]
             if not np.array_equal(mapping, np.arange(len(input_paths))):
                 return wp.array(mapping, dtype=wp.int32)
         return None
--- a/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
+++ b/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
@@ -212,12 +212,8 @@ class SceneDataProvider:
             paths or if no mapping is needed.
         """
         if input_paths := self.backend.transform_paths:
-            # Build a path -> output-index map once, then resolve each input path
-            # against it.  This replaces an ``paths.index(path)`` linear scan per
-            # input path, which is quadratic overall and took minutes at the ~200k
-            # rigid bodies of an 8192-env scene.  First occurrence wins, preserving
-            # the ``list.index`` semantics for duplicate paths; measured over 200k
-            # paths the guard costs nothing versus an unguarded dict comprehension.
+            # The map keeps resolution linear in the number of paths. For duplicate
+            # paths the first occurrence wins, matching ``list.index``.
             path_to_out: dict[str | None, int] = {}
             for out_idx, out_path in enumerate(paths):
                 if out_path not in path_to_out:
--- a/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
+++ b/source/isaaclab/isaaclab/scene_data/scene_data_provider.py
@@ -212,13 +212,16 @@ class SceneDataProvider:
             paths or if no mapping is needed.
         """
         if input_paths := self.backend.transform_paths:
-            # Build a path -> output-index map once, then resolve each input path in
-            # O(1).  This was an O(N^2) ``paths.index(path)`` linear scan per input
-            # path -- minutes at the ~200k rigid bodies of an 8192-env scene.
-            # A plain dict comprehension keeps the LAST occurrence of a duplicate
-            # where ``list.index`` kept the first; duplicate paths are invalid input
-            # and produce a wrong mapping under either choice, so take the fast form.
-            path_to_out = {out_path: out_idx for out_idx, out_path in enumerate(paths)}
+            # Build a path -> output-index map once, then resolve each input path
+            # against it.  This replaces an ``paths.index(path)`` linear scan per
+            # input path, which is quadratic overall and took minutes at the ~200k
+            # rigid bodies of an 8192-env scene.  First occurrence wins, preserving
+            # the ``list.index`` semantics for duplicate paths; measured over 200k
+            # paths the guard costs nothing versus an unguarded dict comprehension.
+            path_to_out: dict[str | None, int] = {}
+            for out_idx, out_path in enumerate(paths):
+                if out_path not in path_to_out:
+                    path_to_out[out_path] = out_idx
             mapping = [path_to_out.get(path, -1) for path in input_paths]
             if not np.array_equal(mapping, np.arange(len(input_paths))):
                 return wp.array(mapping, dtype=wp.int32)
--- a/source/isaaclab/isaaclab/utils/warp/fabric.py
+++ b/source/isaaclab/isaaclab/utils/warp/fabric.py
@@ -52,15 +52,9 @@ def arange_k(a: ArrayUInt32_1d):
 def map_view_indices_to_fabric_slots(view_indices: FabricArrayUInt32, fabric_slots: ArrayInt32_1d):
     """Invert a selection's per-prim view-index attribute into a slot lookup table.
 
-    ``view_indices`` is the fabric array of a per-view ``uint`` index attribute:
-    one entry per selected prim, holding that prim's view-side index.  After the
-    launch, ``fabric_slots[view_index]`` is the fabric-side slot of that view
-    prim in the selection, suitable as :class:`wp.indexedfabricarray` indices.
-
-    The dtypes differ on purpose: the input is ``uint32`` because the Fabric
-    index attribute is authored as ``UInt``, while the output is ``int32``
-    because Warp only accepts ``int32`` index arrays (see
-    :data:`ArrayInt32_1d`).  This kernel is where that boundary is crossed.
+    Inverts a permutation: ``view_indices`` holds each selected prim's view-side
+    index, and after the launch ``fabric_slots[view_index]`` is that prim's
+    fabric-side slot, ready to use as :class:`wp.indexedfabricarray` indices.
 
     The launch dimension must equal the selection's prim count, and the stored
     view indices must cover ``0..dim-1`` exactly for the table to be complete.
@@ -72,12 +66,7 @@ def map_view_indices_to_fabric_slots(view_indices: FabricArrayUInt32, fabric_slo
 
 @wp.kernel(enable_backward=False)
 def gather_fabric_slots(slots: ArrayInt32_1d, gather_map: ArrayUInt32_1d, out_slots: ArrayInt32_1d):
-    """Gather ``slots`` entries through ``gather_map``: ``out_slots[i] = slots[gather_map[i]]``.
-
-    ``gather_map`` holds view-side indices (``uint32``), while ``slots`` and
-    ``out_slots`` hold Fabric slots for :class:`wp.indexedfabricarray`
-    (``int32``); see :data:`ArrayInt32_1d`.
-    """
+    """Gather ``slots`` entries through ``gather_map``: ``out_slots[i] = slots[gather_map[i]]``."""
     i = int(wp.tid())
     out_slots[i] = slots[int(gather_map[i])]
 
--- a/source/isaaclab/isaaclab/utils/warp/fabric.py
+++ b/source/isaaclab/isaaclab/utils/warp/fabric.py
@@ -29,10 +29,6 @@ else:
     IndexedFabricArrayMat44d = wp.indexedfabricarray(dtype=wp.mat44d)
     ArrayUInt32 = wp.array(ndim=1, dtype=wp.uint32)
     ArrayUInt32_1d = wp.array(dtype=wp.uint32)
-    # Signedness here is not a style choice: view indices are ``uint`` (matching
-    # the Fabric ``UInt`` index attributes they come from), but anything used as
-    # :class:`wp.indexedarray` / :class:`wp.indexedfabricarray` indices must be
-    # ``int32`` -- Warp's ``check_index_array`` raises on any other dtype.
     ArrayInt32_1d = wp.array(dtype=wp.int32)
     ArrayFloat32_2d = wp.array(ndim=2, dtype=wp.float32)
 
--- a/source/isaaclab/isaaclab/utils/warp/fabric.py
+++ b/source/isaaclab/isaaclab/utils/warp/fabric.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
     IndexedFabricArrayMat44d = Any
     ArrayUInt32 = Any
     ArrayUInt32_1d = Any
+    ArrayInt32_1d = Any
     ArrayFloat32_2d = Any
 else:
     FabricArrayUInt32 = wp.fabricarray(dtype=wp.uint32)
@@ -28,6 +29,7 @@ else:
     IndexedFabricArrayMat44d = wp.indexedfabricarray(dtype=wp.mat44d)
     ArrayUInt32 = wp.array(ndim=1, dtype=wp.uint32)
     ArrayUInt32_1d = wp.array(dtype=wp.uint32)
+    ArrayInt32_1d = wp.array(dtype=wp.int32)
     ArrayFloat32_2d = wp.array(ndim=2, dtype=wp.float32)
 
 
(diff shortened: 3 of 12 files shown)


source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -24,6 +25,23 @@
...
 
 
+def _parent_path(prim_path: str) -> str:
+    """Parent prim path of ``prim_path``.
+
+    Raises:
+        RuntimeError: If the prim is directly under the stage root and thus has
+            no non-pseudoroot parent to read Fabric matrices from.
+    """
+    parent = prim_path.rsplit("/", 1)[0]
```
pv-nvidia: No need to split all parts, we just need the last part! This can be optimized

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -577,39 +614,63 @@ def _initialize_fabric(self) -> None:
...
-                    continue
-                seen_paths.add(path)
+        # Per-view Fabric index attribute names (see ``_view_uid_counter``).
+        uid = next(FabricFrameView._view_uid_counter)
+        self._child_index_attr = f"isaaclab:fabricFrameView:{uid}:index"
+        self._parent_index_attr = f"isaaclab:fabricFrameView:{uid}:parentIndex"
+
+        # Unique parents in first-occurrence order; ``parent_ordinal`` maps a
+        # parent path to its position in that order.
+        self._unique_parent_paths = list(dict.fromkeys(_parent_path(p) for p in self.prim_paths))
```
pv-nvidia: We should cache _parent_path(p) for p in self.prim_paths) because we call _parent_path again later

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -577,39 +614,63 @@ def _initialize_fabric(self) -> None:
...
+        self._sel_ro = self._stage.SelectPrims(require_attrs=[child_tag, wm_ro, lm_ro], device=self._device)
+        self._sel_rw = self._stage.SelectPrims(require_attrs=[child_tag, wm_rw, lm_rw], device=self._device)
+        self._sel_parent = self._stage.SelectPrims(require_attrs=[parent_tag, wm_ro], device=self._device)
 
+        # View-side indices + kernel-built slot-mapping buffers.
         self._view_indices = wp.array(list(range(self.count)), dtype=wp.uint32, device=self._device)
-        self._rebuild_ro_arrays()
-        self._rebuild_rw_arrays()
+        self._child_parent_map = wp.array(
+            [parent_ordinal[_parent_path(p)] for p in self.prim_paths], dtype=wp.uint32, device=self._device
```
pv-nvidia: Here we should use the earlier cached parent paths

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
```diff
@@ -213,10 +212,15 @@ def create_mapping(self, paths: list[str | None]) -> wp.array(dtype=wp.int32) |
...
-            for i, path in enumerate(input_paths):
-                with contextlib.suppress(ValueError):
-                    mapping[i] = paths.index(path)
+            # Build a path -> output-index map once (first occurrence wins, to match
+            # ``list.index`` semantics), then resolve each input path in O(1).  This
+            # was an O(N^2) ``paths.index(path)`` linear scan per input path -- minutes
+            # at the ~200k rigid bodies of an 8192-env scene.
+            path_to_out: dict[str | None, int] = {}
+            for out_idx, out_path in enumerate(paths):
+                if out_path not in path_to_out:
```
pv-nvidia, line 219: Do we need this check at all?

Two cases exists here:
1) we have duplicate entries in the input_path. One of them will not be handled, which results in incorrect behavior either way
2) we have None or invalid paths. That will result in negative indices and the kernels won't like that, they will crash or give undefined behavior.

So it doesn't matter what we do, pick the first or last, we will get incorrect behavior anyway. So we better pick the most performant form. So could be just a dict comprehension

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
```diff
@@ -213,10 +212,14 @@ def create_mapping(self, paths: list[str | None]) -> wp.array(dtype=wp.int32) |
...
-            for i, path in enumerate(input_paths):
-                with contextlib.suppress(ValueError):
-                    mapping[i] = paths.index(path)
+            # Build a path -> output-index map once, then resolve each input path in
+            # O(1).  This was an O(N^2) ``paths.index(path)`` linear scan per input
+            # path -- minutes at the ~200k rigid bodies of an 8192-env scene.
+            # A plain dict comprehension keeps the LAST occurrence of a duplicate
+            # where ``list.index`` kept the first; duplicate paths are invalid input
+            # and produce a wrong mapping under either choice, so take the fast form.
+            path_to_out = {out_path: out_idx for out_idx, out_path in enumerate(paths)}
```
AntoineRichard: AI-generated review — Important: this silently changes the public method behavior for duplicate paths. The former `list.index` mapped to the first occurrence; this comprehension maps to the last (`input_paths=["/A"]`, `paths=["/A", "/A"]` produces `[1]` instead of `[0]`). Neither the type nor docstring rejects duplicates, and this repository requires deprecation before a breaking API change. Preserve first-occurrence semantics in O(N), or validate duplicates as a separate compatibility-managed change. Add a focused unit test for reordered, missing, and duplicate paths; the current backend tests do not exercise non-empty transform mappings.

source/isaaclab/isaaclab/utils/warp/fabric.py
```diff
@@ -46,6 +48,40 @@ def arange_k(a: ArrayUInt32_1d):
...
+@wp.kernel(enable_backward=False)
+def map_view_indices_to_fabric_slots(view_indices: FabricArrayUInt32, fabric_slots: ArrayInt32_1d):
+    """Invert a selection's per-prim view-index attribute into a slot lookup table.
+
+    ``view_indices`` is the fabric array of a per-view ``uint`` index attribute:
+    one entry per selected prim, holding that prim's view-side index.  After the
+    launch, ``fabric_slots[view_index]`` is the fabric-side slot of that view
+    prim in the selection, suitable as :class:`wp.indexedfabricarray` indices.
+
+    The dtypes differ on purpose: the input is ``uint32`` because the Fabric
```
AntoineRichard: AI-generated review — Comment hygiene: keep this kernel docstring functional. The UInt-versus-int32 rationale is repeated here, in `gather_fabric_slots`, in buffer declarations, and in the PR narrative. A concise contract such as “Invert a permutation of view indices into int32 Fabric slots” plus the precondition is sufficient; keep the `indexedfabricarray` dtype constraint once at the construction boundary. This avoids duplicated rationale and documentation of rejected alternatives.

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -24,6 +26,25 @@
...
 
 
+def _parent_path(prim_path: str) -> str:
+    """Parent prim path of ``prim_path``.
+
+    Raises:
+        RuntimeError: If the prim is directly under the stage root and thus has
+            no non-pseudoroot parent to read Fabric matrices from.
+    """
+    # Slice at the last separator instead of rsplit: no list, no tail string.
```
AntoineRichard: AI-generated review — Comment hygiene: remove this micro-optimization narrative. The code is self-evident, and “no list, no tail string” documents a rejected implementation rather than durable behavior. If absolute paths are the required precondition, state that once in the docstring and validate it if callers can violate it.

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -132,31 +153,38 @@ class FabricFrameView(BaseFrameView):
       :mod:`isaaclab.sim.views.xform_space_writer` for the full contract).
       The "torn data" concern is what motivates that no-step rule; it
       is separate from why the tracking pause exists.
-    * **Two persistent selections, flipped by the writer scope.**  Two
-      selections are built once during ``_initialize_fabric`` and kept for
-      the view's lifetime:
+    * **Per-view index attributes; selections match O(view), not O(stage).**
```
AntoineRichard: AI-generated review — Documentation duplication: this public class docstring now exposes private selection names, exact kernel launches, cache strategy, and error implementation already described by `_refresh_*` and the PR. It will become stale as internals change. Keep only the externally observable contract and lifecycle limitations here—selection is scoped to managed prims, topology changes are handled, membership loss raises—and keep implementation mechanics beside the helpers.

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -466,78 +507,84 @@ def _recompute_world_from_local_all(self) -> None:
...
-        return self._world_ifa_rw if self._is_rw else self._world_ifa_ro
+    def _get_world_ifa(self) -> wp.indexedfabricarray:
+        sel = self._refresh_child_selection()
+        return wp.indexedfabricarray(fa=wp.fabricarray(sel, self._WORLD_MATRIX_NAME), indices=self._child_slots_buf)
 
-    def _get_local_ifa(self):
-        self._refresh_active_bundle_if_needed()
-        return self._local_ifa_rw if self._is_rw else self._local_ifa_ro
+    def _get_local_ifa(self) -> wp.indexedfabricarray:
+        sel = self._refresh_child_selection()
```
AntoineRichard, line 554: AI-generated review — Moderate: both `_recompute_local_from_world_all` and `_recompute_world_from_local_all` call the world and local child accessors, so `_refresh_child_selection()` and its mapping kernel run twice against the same active selection on every writer-scope exit. Refresh once and construct both indexed arrays from that selection and `_child_slots_buf` (for example, with a helper returning the pair). This removes duplicated work from the hot path this PR is optimizing.

source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py
```diff
@@ -203,10 +204,15 @@ def test_fabric_rebuild_after_topology_change(device, view_factory):
     with view.xform_world_space_writer() as w:
         w.set_poses(positions=initial)
 
-    # Simulate topology change: rebuild both selection bundles, mirroring the
-    # lazy paths in the ``_refresh_active_bundle_if_needed`` accessor.
-    view._rebuild_ro_arrays()
-    view._rebuild_rw_arrays()
+    # Simulate topology change: refresh both child selections and the parent
+    # selection, mirroring the accessor paths.
+    view._refresh_child_selection()  # RO (steady state)
```
AntoineRichard, line 210: AI-generated review — Moderate: this does not simulate a topology change. It invokes refresh helpers without changing Fabric or making `PrepareForReuse` report a bucket change, then repeats a round trip already covered by the shared contract tests. It can pass when topology recovery is broken. Trigger a real membership-preserving bucket reorder and verify the mapping, or remove this test; direct execution of private helpers is not meaningful regression coverage.

source/isaaclab_physx/test/sim/test_views_xform_prim_fabric.py
```diff
@@ -297,6 +303,36 @@ def test_prepare_for_reuse_detects_topology_change(device, view_factory):
         assert not result, "PrepareForReuse should return False when no topology change"
 
 
+@pytest.mark.parametrize("device", test_devices())
+def test_selections_match_only_the_view_prims(device, view_factory):
```
AntoineRichard, line 308: AI-generated review — Test documentation: reduce this docstring to the functional contract, for example, `Selections contain only managed child prims and their unique parents.` The 8192-environment history and failure narrative duplicate the PR description and changelog and are not needed to understand or maintain the test.

source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
```diff
@@ -0,0 +1,6 @@
+Fixed
+^^^^^
+
+* Fixed a quadratic path lookup in :class:`~isaaclab.scene_data.SceneDataProvider` transform
+  mapping that stalled setup at high rigid-body counts (thousands of environments). The
+  per-item ``list.index`` scan is now an ``O(N)`` dictionary lookup.
```
AntoineRichard: AI-generated review — Documentation correctness: this is implementation-heavy and the complexity wording is inaccurate. A dictionary lookup is average O(1); building the complete mapping is O(N). State the user-visible outcome instead, for example, `Fixed transform mapping stalls at high rigid-body counts.` Keep algorithm and benchmark details in the PR.

source/isaaclab_physx/changelog.d/fix-physx-newton-camera-pose-scaling.rst
```diff
@@ -0,0 +1,8 @@
+Fixed
+^^^^^
+
+* Fixed camera world-pose resolution stalling (and benchmark timeouts) at high environment
+  counts under the PhysX backend. The Fabric frame view now tags its prims with per-view
```
AntoineRichard: AI-generated review — Documentation duplication: this changelog repeats the complete implementation—tags, selection scope, GPU rebuild, and host path resolution—already covered by code and the PR. Keep the durable user outcome, for example, `Fixed PhysX camera world-pose stalls at high environment counts.`

AntoineRichard (changes requested):
AI-generated review (Codex) — Requesting changes.

Two merge blockers remain:

- The existing unresolved inline finding on per-view Fabric index attributes documents a measured per-stage leak: recreated views permanently add attributes and progressively slow later initialization. A lifecycle cleanup or design that avoids permanent per-view stage state is needed; a documentation-only workaround is not sufficient for a normal recreation path. See https://github.com/isaac-sim/IsaacLab/pull/6805#discussion_r3700250511.
- `SceneDataProvider.create_mapping` now silently changes duplicate-path behavior from first occurrence to last occurrence without validation or deprecation. Preserve the public behavior or reject duplicates through a compatibility-managed change, with focused unit coverage.

I also left inline requests to remove a duplicated child-selection refresh from the writer hot path, replace or remove a topology test that never changes topology, and reduce implementation-history prose duplicated across comments, docstrings, tests, and changelog fragments.

Verification performed on head `dc22f1b97018d99be36223d50a374456a5b87c9f`:

- PhysX Fabric FrameView test file on CPU: 35 passed, 43 skipped; CUDA was unavailable.
- New selection-scope regression test: passed on the PR implementation and failed against the pre-fix installed implementation.
- Scene-data geometry tests: 5 passed.
- Direct duplicate-path probe confirmed the semantic change: one `/A` input maps to index 1 for `paths=["/A", "/A"]`, versus index 0 previously.

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
```diff
@@ -213,10 +212,17 @@ def create_mapping(self, paths: list[str | None]) -> wp.array(dtype=wp.int32) |
...
-            mapping = [-1] * len(input_paths)
-            for i, path in enumerate(input_paths):
-                with contextlib.suppress(ValueError):
-                    mapping[i] = paths.index(path)
+            # Build a path -> output-index map once, then resolve each input path
+            # against it.  This replaces an ``paths.index(path)`` linear scan per
+            # input path, which is quadratic overall and took minutes at the ~200k
+            # rigid bodies of an 8192-env scene.  First occurrence wins, preserving
+            # the ``list.index`` semantics for duplicate paths; measured over 200k
+            # paths the guard costs nothing versus an unguarded dict comprehension.
```
AntoineRichard: Could we cut this comment down?

source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -577,39 +674,72 @@ def _initialize_fabric(self) -> None:
...
+        child_parent_paths = [_parent_path(p) for p in self.prim_paths]
+        self._unique_parent_paths = list(dict.fromkeys(child_parent_paths))
+        parent_ordinal = {path: i for i, path in enumerate(self._unique_parent_paths)}
+
+        # Tag children and parents with their per-view index and ensure both
+        # carry the Fabric world+local matrix attributes (``Create*Attr`` calls
+        # are idempotent).  The index attribute doubles as the selection filter:
+        # the selections below match ONLY tagged prims, so their size is
+        # O(view), not O(stage).  A prim that is both a child and a parent of
+        # this view receives both index attributes.
```
AntoineRichard, line 694: Can we cut down that comment?

AntoineRichard (approved):
LGTM from a high level POV.

source/isaaclab/changelog.d/fix-physx-newton-camera-pose-scaling.rst
```diff
@@ -0,0 +1,13 @@
...
+^^^^^
+
+* Fixed :class:`~isaaclab.scene_data.SceneDataProvider` transform mapping stalling
+  at high rigid-body counts, which delayed setup by minutes in scenes with
+  thousands of environments.
+
+Added
+^^^^^
+
+* Added :meth:`~isaaclab.sim.views.BaseFrameView.close` to release backend state
```
pbarejko, line 11: How releasing the backend state helps the performance?

pv-nvidia: Each FabricFrameView assigns a unique index attribute to the selected prims, so these need to be removed when the frame view is deleted. Otherwise every newly created frame view would again create new attributes, and so on.  

In general, a frame view must be able to do cleanup, hence the close method.

pbarejko:
Blocked by https://github.com/isaac-sim/IsaacLab/pull/6945

kellyguo11 (approved):

PR #6810, opened 2026-07-30
Fix replay crash on trailing idle action in IK tasks

rwiltz:
# Description

  Replaying demonstrations for an absolute task-space (IK) task crashed *after* all
  episodes had replayed successfully, with a `torch.linalg.solve ... input matrix is
  singular` error from the differential IK controller instead of exiting cleanly.

  Root cause is a trailing step past the end of the recorded data. In
  `replay_episodes_loop`, `env.step(actions)` sat outside the `has_next_action`
  check, so once every environment had exhausted its episodes the loop applied the
  untouched `idle_action` one last time before terminating. No task defines
  `idle_action`, so it falls back to `torch.zeros(env.action_space.shape)` — and for
  an absolute-pose action (`[pos_xyz, quat_xyzw, gripper]`) a zeros action carries a
  **zero-norm quaternion**.

  `DifferentialIKController.set_command` renormalized it as `quat / norm(quat)`,
  i.e. `0/0`, producing a NaN target orientation. The NaN propagated into the joint
  position targets, diverged the articulation, and only surfaced on the next
  decimation sub-step as an unrelated "singular matrix" failure — which is why the
  traceback points at the solver rather than at the defect.

  Fixes:

  - `scripts/tools/replay_demos.py`: stop before stepping once every environment is
    exhausted, so the replay terminates without applying the idle action or running
    another IK solve.
  - `DifferentialIKController.set_command`: a degenerate (zero-norm) commanded
    quaternion now holds the current end-effector orientation (identity when none was
    supplied) instead of emitting NaN. Applied per-environment via a branchless
    `torch.where`, so there is no added host sync on the training hot path.
  - `adaptive_dls`: a non-finite Jacobian is reported by its actual cause instead of
    the opaque LAPACK singular-matrix / convergence failure. The check runs only on
    the failure path, so the happy path is unchanged.

  The controller change is not redundant with the script change: with `--num_envs > 1`,
  an environment that finishes early keeps receiving the zero-quaternion idle action
  every step while the others replay, which would diverge that environment and take
  down the whole batch.

  Reproduced and verified with `IsaacContrib-Stack-Cube-SO101-IK-Abs-v0`. On the
  unfixed code the QA command produces a byte-for-byte identical traceback; with the
  fix it prints `Finished replaying 1 episode.` and exits 0.

  Note (out of scope): `idle_action` defaulting to zeros is unsound for *any*
  absolute-pose task, since a zero quaternion is never a valid orientation. Seeding
  the quaternion to identity would be a broader change affecting every task using the
  replay script, so it is left for a separate PR.

Fixes # (issue)

## Type of change

- 
## Checklist

### DIFF
--- a/scripts/tools/replay_demos.py
+++ b/scripts/tools/replay_demos.py
@@ -216,9 +216,7 @@ def replay_episodes_loop(  # noqa: C901
                         has_next_action = True
                     actions[env_id] = env_next_action
                 if not has_next_action:
-                    # Every environment has exhausted its episodes. Stop before stepping so the
-                    # replay does not apply the idle action past the end of the recorded data --
-                    # for task-space (IK) tasks the zero idle action carries a zero-norm quaternion.
+                    # Stop before stepping once every environment has exhausted its recorded actions.
                     break
                 if first_loop:
                     first_loop = False
--- a/scripts/tools/replay_demos.py
+++ b/scripts/tools/replay_demos.py
@@ -215,6 +215,11 @@ def replay_episodes_loop(  # noqa: C901
                     else:
                         has_next_action = True
                     actions[env_id] = env_next_action
+                if not has_next_action:
+                    # Every environment has exhausted its episodes. Stop before stepping so the
+                    # replay does not apply the idle action past the end of the recorded data --
+                    # for task-space (IK) tasks the zero idle action carries a zero-norm quaternion.
+                    break
                 if first_loop:
                     first_loop = False
                 else:
--- a/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
+++ b/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
@@ -1,0 +1,11 @@
+Fixed
+^^^^^
+
+* Fixed :meth:`~isaaclab.controllers.DifferentialIKController.set_command` producing a NaN target
+  orientation when an absolute pose command carried a zero-norm quaternion. The NaN propagated into
+  the joint position targets and diverged the articulation, surfacing a step later as an unrelated
+  ``torch.linalg.solve ... input matrix is singular`` error. Degenerate quaternions now hold the
+  current end-effector orientation instead.
+* Fixed the ``adaptive_dls`` inverse-kinematics method reporting a non-finite Jacobian as an opaque
+  LAPACK singular-matrix or convergence failure. It now raises an error naming the actual cause: the
+  articulation state diverged before the solve.
--- a/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
+++ b/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
@@ -6,6 +6,7 @@ Fixed
   the joint position targets and diverged the articulation, surfacing a step later as an unrelated
   ``torch.linalg.solve ... input matrix is singular`` error. Degenerate quaternions now hold the
   current end-effector orientation instead.
-* Fixed the ``adaptive_dls`` inverse-kinematics method reporting a non-finite Jacobian as an opaque
-  LAPACK singular-matrix or convergence failure. It now raises an error naming the actual cause: the
+* Fixed the ``adaptive_dls`` inverse-kinematics method handling of a non-finite Jacobian, which
+  backends either reported as an opaque LAPACK singular-matrix or convergence failure, or propagated
+  silently into the joint position targets. It now raises an error naming the actual cause: the
   articulation state diverged before the solve.
--- a/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
+++ b/source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
@@ -1,12 +1,7 @@
 Fixed
 ^^^^^
 
-* Fixed :meth:`~isaaclab.controllers.DifferentialIKController.set_command` producing a NaN target
-  orientation when an absolute pose command carried a zero-norm quaternion. The NaN propagated into
-  the joint position targets and diverged the articulation, surfacing a step later as an unrelated
-  ``torch.linalg.solve ... input matrix is singular`` error. Degenerate quaternions now hold the
-  current end-effector orientation instead.
-* Fixed the ``adaptive_dls`` inverse-kinematics method handling of a non-finite Jacobian, which
-  backends either reported as an opaque LAPACK singular-matrix or convergence failure, or propagated
-  silently into the joint position targets. It now raises an error naming the actual cause: the
-  articulation state diverged before the solve.
+* Fixed demonstration replay stepping once after all episodes completed.
+* Fixed :meth:`~isaaclab.controllers.DifferentialIKController.set_command` handling of
+  unnormalizable absolute-pose quaternions, which produced a NaN target orientation. Such
+  commands now hold the current end-effector orientation, or identity when none is provided.
--- a/source/isaaclab/isaaclab/controllers/differential_ik.py
+++ b/source/isaaclab/isaaclab/controllers/differential_ik.py
@@ -121,11 +121,8 @@ class DifferentialIKController:
         It is up to the user to ensure that the command is given in the correct frame. The method only
         applies the relative mode if the command type is ``position_rel`` or ``pose_rel``.
 
-        For absolute ``pose`` commands the commanded quaternion is renormalized, so a slightly
-        non-unit quaternion is accepted. A *degenerate* quaternion -- one that cannot be normalized
-        to a finite value, i.e. a zero quaternion or one whose norm underflows to zero -- would
-        otherwise yield NaN, so those entries fall back per-environment to :attr:`ee_quat` (holding
-        the current orientation), or to identity when :attr:`ee_quat` is not provided.
+        Absolute ``pose`` commands normalize finite quaternions; unnormalizable entries use
+        :paramref:`ee_quat`, or identity when :paramref:`ee_quat` is omitted.
 
         Args:
             command: The input command in shape (N, 3) or (N, 6) or (N, 7).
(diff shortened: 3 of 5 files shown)


rwiltz:
@greptile

source/isaaclab/test/controllers/test_differential_ik_features.py
AntoineRichard, line 1: Do we need all these tests? Could you trim these to the strict minimum?

AntoineRichard (approved):
Looks reasonable to me, take a look at the other comments.

scripts/tools/replay_demos.py
```diff
@@ -215,6 +215,11 @@ def replay_episodes_loop(  # noqa: C901
                     else:
                         has_next_action = True
                     actions[env_id] = env_next_action
+                if not has_next_action:
```
AntoineRichard, line 218: **AI-generated review — Moderate (test coverage):** This is the primary bug fix, but no test asserts this loop contract. The new controller tests would still pass if this break were removed, and the fallback could mask the crash while the unwanted extra `env.step` remained. Please add a focused regression with a stub environment/dataset that asserts `env.step` is called exactly once per recorded action and never with the final idle action; per the repository rule, verify it fails on the base implementation.

scripts/tools/replay_demos.py
```diff
@@ -215,6 +215,11 @@ def replay_episodes_loop(  # noqa: C901
...
+                    # Every environment has exhausted its episodes. Stop before stepping so the
+                    # replay does not apply the idle action past the end of the recorded data --
+                    # for task-space (IK) tasks the zero idle action carries a zero-norm quaternion.
```
AntoineRichard: **AI-generated review — Moderate (comment quality):** Keep this functional and local to the branch, for example: `# Stop before stepping after every environment has exhausted its recorded actions.` The IK/quaternion failure chain is incident history already captured by the PR and does not belong in the implementation comment.

source/isaaclab/changelog.d/rwiltz-diff-ik-degenerate-quat.rst
```diff
@@ -0,0 +1,12 @@
...
+
+* Fixed :meth:`~isaaclab.controllers.DifferentialIKController.set_command` producing a NaN target
+  orientation when an absolute pose command carried a zero-norm quaternion. The NaN propagated into
+  the joint position targets and diverged the articulation, surfacing a step later as an unrelated
+  ``torch.linalg.solve ... input matrix is singular`` error. Degenerate quaternions now hold the
+  current end-effector orientation instead.
+* Fixed the ``adaptive_dls`` inverse-kinematics method handling of a non-finite Jacobian, which
+  backends either reported as an opaque LAPACK singular-matrix or convergence failure, or propagated
+  silently into the joint position targets. It now raises an error naming the actual cause: the
+  articulation state diverged before the solve.
```
AntoineRichard: **AI-generated review — Moderate (documentation):** This fragment omits the user-visible replay-loop fix while duplicating the internal NaN/LAPACK diagnosis from the PR and code comments. Please make it outcome-focused, for example: `Fixed demonstration replay stepping once after all episodes completed.` and `Fixed DifferentialIKController handling of unnormalizable absolute-pose quaternions.` If the adaptive-DLS validation remains, describe only that it now rejects non-finite Jacobians; the controller cannot establish that the articulation state was the source.

source/isaaclab/isaaclab/controllers/differential_ik.py
```diff
@@ -119,12 +121,20 @@ def set_command(
         It is up to the user to ensure that the command is given in the correct frame. The method only
         applies the relative mode if the command type is ``position_rel`` or ``pose_rel``.
 
+        For absolute ``pose`` commands the commanded quaternion is renormalized, so a slightly
+        non-unit quaternion is accepted. A *degenerate* quaternion -- one that cannot be normalized
+        to a finite value, i.e. a zero quaternion or one whose norm underflows to zero -- would
+        otherwise yield NaN, so those entries fall back per-environment to :attr:`ee_quat` (holding
+        the current orientation), or to identity when :attr:`ee_quat` is not provided.
```
AntoineRichard: **AI-generated review — Moderate (public docs):** State the contract directly instead of narrating the NaN mechanism: `Absolute pose commands normalize finite quaternions; unnormalizable entries use ee_quat, or identity when ee_quat is omitted.` Also use `:paramref:` for the parameter reference, not `:attr:`; there is no `ee_quat` attribute to resolve.

source/isaaclab/isaaclab/controllers/differential_ik.py
```diff
@@ -158,9 +168,20 @@ def set_command(
...
+                # silently into the joint position targets and only surfaces a step later as an
+                # unrelated solver failure. Hold the current end-effector orientation instead
+                # (identity when no current orientation was supplied) for those environments.
                 quat = self._command[:, 3:7]
-                self.ee_quat_des = quat / torch.linalg.norm(quat, dim=-1, keepdim=True)
+                normalized_quat = quat / torch.linalg.norm(quat, dim=-1, keepdim=True)
+                # Degeneracy is decided by whether the normalization produced a finite result, not by
+                # a magnitude threshold: a zero-norm quaternion gives 0/0, and a norm that underflows
+                # to zero gives Inf. Any quaternion that normalizes cleanly keeps its previous
+                # meaning, however small its norm (e.g. ``[0, 0, 0, 1e-7]`` is still identity).
```
AntoineRichard: **AI-generated review — Moderate (comment duplication):** These two blocks repeat the public docstring and PR history, including a specific `1e-7` example added to explain an earlier implementation choice. Replace them with one durable functional comment such as `# Normalize valid quaternions and use the fallback for non-finite results.` The tests should carry boundary examples.

source/isaaclab/isaaclab/controllers/differential_ik.py
```diff
@@ -274,6 +295,19 @@ def _compute_delta_joint_pos(self, delta_pose: torch.Tensor, jacobian: torch.Ten
...
+            # front rather than reacting to a decomposition error: depending on the backend, LAPACK
+            # either reports a misleading "matrix is singular"/"failed to converge" or propagates the
+            # NaN silently into the joint targets. Failing here names the real cause in both cases.
+            if not torch.isfinite(jacobian).all():
+                raise RuntimeError(
+                    "Differential IK received a non-finite Jacobian, so the articulation state has already"
+                    " diverged (NaN/Inf) before this solve. This is usually caused by a NaN or degenerate"
+                    " (zero-norm quaternion) task-space command applied on an earlier step -- check the"
+                    " commands feeding the IK action term."
+                )
```
AntoineRichard: **AI-generated review — Important (performance/correctness):** Please remove this secondary diagnostic from this fix or redesign it without a synchronous Python branch. `if not torch.isfinite(jacobian).all()` converts a device scalar to a Python boolean, forcing a host/device synchronization on every adaptive-DLS control step solely to improve an exceptional error message. The explanation is also too strong: a non-finite Jacobian does not prove that the articulation diverged or that a quaternion command caused it, and finite inputs do not guarantee finite downstream arithmetic (a finite `1e20` Jacobian already overflows `J @ J.T` in float32 and returns a NaN result). If validation remains, keep the message factual—e.g. `Differential IK requires a finite Jacobian; received NaN or Inf.`—and use an asynchronous or failure-path mechanism.

source/isaaclab/test/controllers/test_differential_ik_features.py
```diff
@@ -86,6 +86,98 @@ def test_set_command_renormalizes_quat():
     torch.testing.assert_close(stored, raw / torch.linalg.norm(raw), atol=1e-6, rtol=0.0)
 
 
+def test_set_command_zero_quat_holds_current_orientation():
+    """A zero-norm commanded quaternion falls back to the current orientation instead of NaN.
+
+    Regression: dividing by a zero norm produced a NaN target orientation that propagated into the
+    joint position targets, diverged the articulation, and only surfaced a step later as an opaque
+    ``torch.linalg.solve ... matrix is singular`` failure.
+    """
```
AntoineRichard: **AI-generated review — Moderate (test duplication):** Merge this case into `test_set_command_zero_quat_only_affects_degenerate_envs` below: give the degenerate row a non-identity current orientation and retain the valid row. One test then proves both per-environment selection and holding the current orientation. Keep its docstring to the one-line behavioral contract; the full failure chain belongs in the PR, not every regression test.

source/isaaclab/test/controllers/test_differential_ik_features.py
```diff
@@ -86,6 +86,98 @@ def test_set_command_renormalizes_quat():
...
+
+def test_adaptive_dls_finite_jacobian_is_unaffected_by_the_guard():
+    """The non-finite guard does not change results for a well-conditioned Jacobian."""
+    c = _make_controller(ik_method="adaptive_dls")
+    ee_pos = torch.tensor([[0.3, -0.1, 0.2]])
+    ee_quat = torch.tensor([_ID_QUAT])
+    c.set_command(torch.tensor([[0.31, -0.1, 0.2] + _ID_QUAT]), ee_pos, ee_quat)
+    jac = torch.eye(6, _NUM_JOINTS).unsqueeze(0).clone()
+    out = c.compute(ee_pos, ee_quat, jac, torch.zeros(1, _NUM_JOINTS))
+    assert torch.isfinite(out).all()
```
AntoineRichard: **AI-generated review — Moderate (unneeded test):** Delete this test. The existing `test_adaptive_dls_damps_singularity` already runs `adaptive_dls` with a finite Jacobian, asserts a finite result, and verifies the damping behavior. This only repeats a weaker subset of that coverage.

AntoineRichard (changes requested):
**AI-generated review — changes requested.**

The replay termination and quaternion fallback behave correctly in the focused tests, but this needs another cleanup pass for showroom quality:

- Add direct regression coverage for the primary replay-loop fix. The controller tests do not prove that the extra idle step stays removed.
- Remove or redesign the unconditional adaptive-DLS finiteness diagnostic. Converting the device reduction to a Python boolean synchronizes the host on every controller step, and its comments/error overstate what can be inferred from a non-finite Jacobian.
- Keep comments, public documentation, tests, and the changelog functional and concise. The same incident narrative is currently repeated at every layer.
- Consolidate overlapping quaternion tests and remove the finite-Jacobian guard test already covered by the existing adaptive-DLS behavior test.

Verification: `./isaaclab.sh -p -m pytest source/isaaclab/test/controllers/test_differential_ik_features.py -q` passed (21 tests) on the PR head; the targeted regression cases failed on the exact base commit; and `./isaaclab.sh -f` passed. The three red CI jobs are unrelated rendering/numerical flakes or a task-environment process crash; all core IsaacLab shards passed.

rwiltz:
@greptile

kellyguo11 (approved):

PR #6818, opened 2026-07-31
Add XR camera picture-in-picture feedback

hougantc-nvda:
# Description

Adds low-latency XR camera picture-in-picture (PiP) feedback to IsaacTeleop so teleoperators can see the same task-configured camera views captured in demonstration data.

The implementation:

- Selects existing task cameras declaratively through ordered `IsaacTeleopCfg.xr_camera_feeds`.
- Supports manual, horizontal, vertical, and grid layouts with viewer-start, head-locked, and world placement.
- Disables PiP when no feeds are configured or external cameras are disabled.
- Shares the task-configured feed path between `teleop_se3_agent.py` and `record_demos.py`.
- Uses a feed-owned CUDA Replicator annotator when available, with the existing Camera RGBA buffer as fallback.
- Keeps reference cameras on `MultiBackendRendererCfg` so other renderer backends remain valid.
- Loads Kit Scene UI lazily so kitless runs warn and continue without PiP or task-camera mutation.
- Applies optional DLSS Ray Reconstruction and execution-mode policy only to the selected feed RenderProduct after PiP attachment. The private Kit adapter authors transient session-layer opinions so later legacy/global synchronization cannot overwrite the selected feed or affect viewport/bystander RenderProducts.
- Falls back from Ray Reconstruction to classic DLSS on Isaac Sim versions before 6.1.
- Limits XR camera PiP to one environment while preserving existing multi-environment behavior for IsaacTeleop XR without PiP.
- Uses the default viewer-start panel placement for both G1 and GR1T2 reference tasks, vertically offset by `0.15 m`.

No production changes are made to backend-neutral IsaacLab core or `isaaclab_physx`; PiP-specific render policy remains on `XrCameraFeedCfg` and in the private IsaacTeleop Scene UI adapter.

The recursive PiP hall-of-mirrors behavior is intentionally deferred and is not claimed fixed by this PR.

Canonical example:

```bash
uv run python scripts/environments/teleoperation/teleop_se3_agent.py \
  --task IsaacContrib-PickPlace-GR1T2-Abs \
  --xr \
  --device cpu
```

## Type of change

- Documentation update

## Screenshots

Quest 3 and IsaacTeleop Web Client acceptance was completed. A headset demonstration video was recorded separately and can be attached to the PR.

## Validation

- Full `pre-commit run --all-files` on the final promoted head
- 88 focused camera-feed, configuration, layout, Scene UI, and task-preset tests on Horde
- 2/2 real Isaac Sim Kit integration tests
- 101/101 session-layer persistence, isolation, and cleanup probe checks under forced contrary root/global synchronization
- Clean G1 motion/reset validation with invariant camera attachment and changing valid RGB output
- Fresh IWER/OpenXR/CloudXR validation with live zero-copy CUDA PiP and stable XR session
- Quest 3 acceptance of G1 PiP, GR1T2 PiP, and the `--disable_external_cameras` no-PiP path
- Full Sphinx documentation build with warnings treated as errors

## Checklist


source/isaaclab_teleop/docs/README.md
```diff
@@ -86,9 +95,21 @@ The existing teleop scripts automatically detect `isaac_teleop` in the environme
...
-    --task My-IsaacTeleop-Env-v0
+    --task My-IsaacTeleop-Env-v0 --xr
+```
+
+`IsaacContrib-PickPlace-GR1T2-Abs` and
+`IsaacContrib-PickPlace-Locomanipulation-G1-Abs` are camera-feedback reference tasks. Both show
+their recorded `robot_pov_cam` observation in PiP:
+
+```bash
+./isaaclab.sh -p scripts/environments/teleoperation/teleop_se3_agent.py \
```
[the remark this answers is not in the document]
hougantc-nvda: Fixed in db52ef332f. The README example now uses uv run python, matching the adjacent example and the repository supported routine script workflow.

source/isaaclab_teleop/isaaclab_teleop/camera_feed.py
```diff
@@ -0,0 +1,463 @@
...
+def _configure_ray_reconstruction_compatibility(camera_cfg: Any, camera_name: str) -> None:
+    """Fall back to classic DLSS on runtimes without responsive denoising."""
+    renderer_cfg = getattr(camera_cfg, "renderer_cfg", None)
+    if getattr(renderer_cfg, "enable_dlss_ray_reconstruction", None) is not True:
+        return
+
+    from isaaclab.utils.version import get_isaac_sim_version
+
+    isaac_sim_version = get_isaac_sim_version()
+    if (isaac_sim_version.major, isaac_sim_version.minor) < _RESPONSIVE_DENOISING_MIN_ISAAC_SIM_VERSION:
```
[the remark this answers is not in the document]
hougantc-nvda: Fixed in db52ef332f. The pre-6.1 compatibility decision now lives in IsaacRtxRenderer._apply_render_product_settings, so every Isaac RTX camera consumer authors classic DLSS on older runtimes without mutating task configuration. Added a renderer contract regression covering Isaac Sim 6.0 (False) and 6.1 (True); the focused renderer and camera-feed suites pass (54 tests total).

source/isaaclab_teleop/isaaclab_teleop/camera_feed.py
```diff
@@ -0,0 +1,463 @@
...
+        if self._bound:
+            raise RuntimeError("XR camera feed session is already bound.")
+        self._env = env
+        self._bound = True
+        try:
+            if self.enabled:
+                self._manager = _XrCameraFeedManager(env, self._cfgs, self._layout_cfg, self._presenter)
+        except Exception:
+            self.close()
+            raise
```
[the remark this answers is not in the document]
hougantc-nvda: Addressed in db52ef332f by making the contract explicit rather than broadly swallowing initialization failures. Missing Scene UI imports still warn and continue without PiP. Configuration, camera-buffer, CUDA-upload, and panel-initialization failures remain fail-fast so production defects are not hidden; the feature documentation now states that distinction.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -18,7 +18,7 @@
 import warp as wp
 from packaging import version
 
-from pxr import Sdf, Usd, UsdGeom
+from pxr import Sdf, Usd, UsdGeom, UsdRender
```
AntoineRichard: AI-generated review — blocking: this top-level import regresses the supported wrapped-Python path. Running `./isaaclab.sh -p -m pytest source/isaaclab_physx/test/renderers/test_isaac_rtx_renderer_contract.py` on this checkout yields 15 failures because `pxr.UsdRender` raises `RuntimeError: extension class wrapper for ... UsdTyped has not been created yet`; the previous `Sdf, Usd, UsdGeom` imports succeed in the same environment. Avoid importing `UsdRender` at module load (for example, validate the prim's type name without loading that schema module, or defer the import to a Kit-initialized path) and keep the existing contract suite runnable.

source/isaaclab_teleop/isaaclab_teleop/camera_feed.py
```diff
@@ -0,0 +1,441 @@
...
+        return self
+
+    def refresh(self) -> None:
+        if not self._bound:
+            raise RuntimeError("XR camera feed session must be bound before refresh.")
+        if self._manager is not None:
+            # Request post-reset camera output before each render. Temporal
+            # RTX cameras need several frames to replace their pre-reset
+            # annotator contents, matching Isaac Lab's reset rerender path.
+            for _ in range(3):
```
AntoineRichard: AI-generated review — this creates a second reset-rerender policy immediately after `env.reset()`, even though `ManagerBasedEnvCfg.num_rerenders_on_reset` already owns exactly this behavior. Hard-coding three extra renders adds hidden reset cost, ignores the task's configured count, and the new unit test merely locks in the magic number. Configure/reuse `num_rerenders_on_reset` for the temporal-camera requirement and limit this method to rebinding/publishing the refreshed buffer.

source/isaaclab/test/controllers/test_pink_ik.py
```diff
@@ -7,8 +7,8 @@
 
 from isaaclab.app import AppLauncher
 
-# launch omniverse app
-simulation_app = AppLauncher(headless=True).app
+# Camera-enabled tasks under test require the rendering experience.
```
AntoineRichard: AI-generated review — this makes the entire Pink IK controller suite launch the camera-rendering experience solely because two referenced task configs now contain showcase cameras. None of these tests assert camera behavior. Strip camera configs/observation terms in the fixture (using the existing camera-removal utility) or use a camera-free task variant so this controller suite stays scoped and avoids unnecessary RTX startup cost.

source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/pickplace_gr1t2_env_cfg.py
```diff
@@ -27,8 +27,20 @@
...
 from isaaclab_teleop.haptic_feedback import GloveHapticFeedbackCfg  # isort: skip
-from isaaclab_teleop.isaac_teleop_cfg import IsaacTeleopCfg  # isort: skip
+from isaaclab_teleop.isaac_teleop_cfg import IsaacTeleopCfg, XrCameraFeedCfg  # isort: skip
 from isaaclab_teleop.xr_cfg import XrCfg  # isort: skip
+from isaaclab_physx.renderers import IsaacRtxRendererCfg  # isort: skip
+from isaaclab_tasks.utils.presets import MultiBackendRendererCfg  # isort: skip
+
+
+@configclass
+class _RobotPovCameraRendererCfg(MultiBackendRendererCfg):
```
AntoineRichard: AI-generated review — this renderer preset, including all three values, is duplicated verbatim in `locomanipulation_g1_env_cfg.py`, alongside a nearly identical 720×450 `CameraCfg`. Extract a shared robot-PoV camera/renderer preset (or a small config factory) and specialize only pose differences. These are showroom tasks; one source of truth prevents their render quality and device behavior from drifting.

source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/pickplace_gr1t2_env_cfg.py
```diff
@@ -633,7 +684,17 @@ def __post_init__(self):
             pipeline_builder=lambda: _build_gr1t2_pickplace_pipeline()[0],
             sim_device=self.sim.device,
             xr_cfg=self.xr,
+            xr_camera_feeds=[
+                XrCameraFeedCfg(
+                    camera_name="robot_pov_cam",
+                    # The 0.48 m-wide 720x450 image is 0.30 m tall; lower its
```
AntoineRichard: AI-generated review — this comment narrates arithmetic behind a literal and is repeated in the other PiP task configs. It will become stale as soon as panel width, aspect ratio, or placement changes. Replace the repeated feed block with a named shared placement/feed preset and remove the per-task value derivation; retain comments only for durable behavioral constraints that the names cannot express.

source/isaaclab_teleop/docs/README.md
```diff
@@ -86,9 +95,21 @@ The existing teleop scripts automatically detect `isaac_teleop` in the environme
 
 ```bash
 uv run python scripts/environments/teleoperation/teleop_se3_agent.py \
-    --task My-IsaacTeleop-Env-v0
+    --task My-IsaacTeleop-Env-v0 --xr
 ```
 
+`IsaacContrib-PickPlace-GR1T2-Abs` and
```
AntoineRichard: AI-generated review — the same reference-task list and runnable command are documented again in `docs/source/features/isaac_teleop.rst`, which this README links immediately below. Keep the README as a short package overview plus the canonical link, and keep task names/commands in the feature guide only; otherwise task renames or CLI changes require synchronized edits in two documentation trees.

source/isaaclab_teleop/test/test_xr_camera_feed_kit_scene_ui.py
```diff
@@ -0,0 +1,570 @@
...
+
+def _module(name: str) -> ModuleType:
+    module = ModuleType(name)
+    module.__path__ = []
+    return module
+
+
+@pytest.fixture
+def scene_ui_module(monkeypatch):
+    """Load the Scene UI implementation against fake Kit and OpenXR modules."""
```
AntoineRichard, line 216: AI-generated review — this 570-line suite recreates the Kit/XR API surface with hand-written modules and then loads the implementation against those fakes. It can verify local math/state transitions, but it cannot catch the signature or lifecycle mismatch that would break the actual Scene UI feature. Add a minimal `isaacsim_ci` integration smoke test against the real Scene UI imports/construction path, then trim these unit tests to behavior that does not simply mirror the test doubles.

source/isaaclab_teleop/test/test_xr_camera_feed.py
```diff
@@ -0,0 +1,426 @@
...
+    from isaaclab_tasks.contrib.locomanip_pick_place.locomanipulation_g1_env_cfg import LocomanipulationG1EnvCfg
+    from isaaclab_tasks.contrib.pick_place.pickplace_gr1t2_env_cfg import PickPlaceGR1T2EnvCfg
+
+    for cfg in (PickPlaceGR1T2EnvCfg(), LocomanipulationG1EnvCfg()):
+        assert cfg.isaac_teleop.xr_camera_feeds[0].camera_name == "robot_pov_cam"
+        assert hasattr(cfg.observations.policy, "robot_pov_cam")
+        assert isinstance(cfg.scene.robot_pov_cam.renderer_cfg, MultiBackendRendererCfg)
+
+
+def test_reference_camera_renderer_resolves_to_isaac_rtx_and_ovrtx():
```
AntoineRichard: AI-generated review — this teleop unit test imports `isaaclab_tasks` and `isaaclab_ov` only to assert literals in a task renderer preset; it does not exercise teleop behavior. Move any necessary preset contract test to `source/isaaclab_tasks/test` or remove it. Keeping it here reverses package boundaries, makes the focused teleop suite much heavier, and duplicates coverage of the preset resolver.

source/isaaclab_physx/test/renderers/test_isaac_rtx_render_product_settings.py
```diff
@@ -0,0 +1,113 @@
...
+        usd_prim = stage.GetPrimAtPath(render_product_path)
+
+        stage_id = UsdUtils.StageCache.Get().GetId(stage).ToLongInt()
+        fabric_stage = UsdRtUsd.Stage.Attach(stage_id)
+        fabric_prim = fabric_stage.GetPrimAtPath(render_product_path)
+
+        expected = ("quality", False)
+        assert _read_dlss_settings(usd_prim) == expected
+        assert _read_dlss_settings(fabric_prim) == expected
+        assert get_settings_manager().get("/rtx/dldenoiser/responsiveDenoising") is True
```
AntoineRichard: AI-generated review — this test sets a process-global Carb value to `True` but never restores its prior value, so later renderer tests in the same Kit process inherit order-dependent state. Snapshot the setting before constructing the renderer and restore it in `finally` (preferably through a fixture that restores every global setting changed by this test).

source/isaaclab_teleop/changelog.d/xr-camera-feedback.minor.rst
```diff
@@ -0,0 +1,19 @@
...
+* Added XR camera feedback to ``teleop_se3_agent.py`` and ``record_demos.py``, with task-configured
+  existing-camera selection, declarative layouts, and viewer-start, head-locked, or explicit-world
+  placement.
+* Added direct and staged CUDA image presentation paths so camera feedback works with CPU physics.
+
+Changed
+^^^^^^^
+
+* Changed Kit Scene UI loading to be lazy so kitless teleoperation warns and continues without PiP.
+* Changed ``--disable_external_cameras`` into the master camera-rendering and PiP gate for
```
AntoineRichard, line 18: AI-generated review — repository rules require migration guidance for every `Changed` entry. This changes the meaning of an existing CLI flag, so add the user action here—for example, explain how to keep external cameras enabled while disabling PiP via an empty/disabled `xr_camera_feeds` selection.

AntoineRichard (dismissed):
AI-generated review

Requesting changes. The renderer/output-device direction is sound, but the current head has one concrete compatibility blocker and several showroom-quality issues that should be resolved before merge:

- The new top-level `UsdRender` import breaks the renderer contract module under the repository's wrapped Python on this checkout; the focused contract run reports 15 failures at import time.
- PiP reset refresh duplicates the environment's existing reset-rerender mechanism with a hard-coded three-render loop.
- Reference-task camera/feed presets, explanatory value comments, and README material are duplicated and already create multiple maintenance surfaces.
- Several new tests are either coupled to higher-level packages, primarily validate hand-written fakes, or leak process-global renderer state.

Verification: full pre-commit passed twice. The focused suites produced 122 passes; the renderer contract file produced 15 failures rooted at `from pxr import UsdRender`. The GPU rendering integration file could not complete in this sandbox because no CUDA device is exposed. Current GitHub checks also show failing `isaaclab_tasks` shards, though I could not attribute those failures from the available summaries.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
hougantc-nvda: Fixed in 30d17335. UsdRender is no longer imported at module load; render-product validation now uses Prim.GetTypeName() == "RenderProduct", avoiding the schema-wrapper import on the wrapped-Python path.

source/isaaclab_teleop/isaaclab_teleop/camera_feed.py
hougantc-nvda: Fixed in 30d17335. The three refresh/render iterations were removed from _XrCameraFeedSession.refresh(). The two reference environment configs now set num_rerenders_on_reset = 3, leaving reset rerender ownership with the environment; the teleop unit test now expects publish only.

source/isaaclab/test/controllers/test_pink_ik.py
hougantc-nvda: Fixed in 30d17335. The Pink IK fixture now applies remove_camera_configs() before environment construction and launches the normal headless experience without enable_cameras=True, keeping this controller suite camera-free.

source/isaaclab_tasks/isaaclab_tasks/contrib/pick_place/pickplace_gr1t2_env_cfg.py
hougantc-nvda: Fixed in 30d17335. Both reference tasks now use the shared robot_pov_camera_cfg() factory, which owns the common camera and renderer settings. The task-package regression covers the default, isaacsim_rtx, and ovrtx resolutions.

source/isaaclab_teleop/docs/README.md
hougantc-nvda: Fixed in 30d17335. The package README no longer repeats the reference-task list or runnable command; it now points to the canonical XR Camera Feedback feature section for those details.

source/isaaclab_teleop/test/test_xr_camera_feed_kit_scene_ui.py
hougantc-nvda: Fixed in 30d17335. Added test_xr_camera_feed_kit_scene_ui_integration.py as an isaacsim_ci integration smoke. It launches the XR experience, imports the real Scene UI implementation, and constructs and closes a real world panel; the fake-based suite remains for isolated behavior.

source/isaaclab_teleop/test/test_xr_camera_feed.py
hougantc-nvda: Fixed in 30d17335. The task and OVRTX imports plus preset assertions were removed from the teleop unit suite. That contract now lives in source/isaaclab_tasks/test/core/test_xr_camera_feed_presets.py.

source/isaaclab_physx/test/renderers/test_isaac_rtx_render_product_settings.py
hougantc-nvda: Fixed in 30d17335. An autouse fixture now snapshots every process-global RTX setting modified by this integration module and restores each value in finally, destroying keys that were previously absent.

source/isaaclab_teleop/changelog.d/xr-camera-feedback.minor.rst
hougantc-nvda: Fixed in 30d17335. The Changed entry now gives the migration path: leave --disable_external_cameras unset and configure xr_camera_feeds as empty or disable every feed to keep task cameras enabled without PiP.

rwiltz:
@greptile

rwiltz:
## Code Review

**Scope**: commits `1fbd334`, `f0c96c2`, `921e213`, `b517542`. Core changes: new `camera_feed.py` and `camera_feed_kit_scene_ui.py` in `isaaclab_teleop`; per-camera DLSS settings in `isaaclab_physx`; `require_visual_shapes()` hook in `Camera.__init__`; `TeleopMessageProcessor` reset latch.

---

### 1. `_XrCameraFeedSession` is private but imported from public entry scripts

**Files**: `scripts/environments/teleoperation/teleop_se3_agent.py:274`, `scripts/tools/record_demos.py:814`

```python
from isaaclab_teleop.camera_feed import _XrCameraFeedSession
```

`_XrCameraFeedSession` has a leading underscore (private by convention) but is the only way for callers to use the PiP lifecycle. The `__init__.pyi` doesn't export it. Entry scripts importing private internals is a stable-API risk — any refactor of `camera_feed.py` breaks a documented entry point without a deprecation notice. The fix is either to remove the underscore and add it to `__init__.pyi`, or expose a module-level factory function (`prepare_xr_camera_feed_session(env_cfg, ...)`) as the public surface.

---

### 2. `require_visual_shapes()` now fires for every Camera, unconditionally

**File**: `source/isaaclab/isaaclab/sensors/camera/camera.py:178-183`

```python
sim_ctx = sim_utils.SimulationContext.instance()
if sim_ctx is not None:
    sim_ctx.require_visual_shapes()
```

The intent is correct (Newton backend needs this before cloning), but this fires in every `Camera.__init__` — including headless RL training tasks that don't need visual geometry at all. `require_visual_shapes()` is a one-way latch (`_visual_shapes_required = True`). A training run with a depth camera, a contact sensor, or any non-visual camera was previously able to skip loading visual shapes; now it can't. The PR comment says "every renderer backend draws the visual-only geometry", but that's only true for cameras that actually render. Consider gating on whether the renderer is an RTX renderer (e.g. `self.cfg.renderer_cfg is not None`), rather than firing for all cameras.

---

### 3. `_validate_layout_cfg` is called redundantly in the normal path

**File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`

`_XrCameraFeedManager.__init__` calls `_validate_layout_cfg(self._layout_cfg)` directly, and then `_layout_feed_cfgs(...)` (called from the same `__init__`) also calls `_validate_layout_cfg(layout_cfg)`. The second call is always a no-op if the first passed. Validation should live in one place — either in `_layout_feed_cfgs` alone (since that's where `layout_cfg` is consumed) or in `__init__` only.

---

### 4. `record_demos.py` applies responsive denoising in a separate call without DLSS antialiasing mode

**File**: `scripts/tools/record_demos.py:820-827`

```python
# separate call, no antialiasing_mode
if camera_feed_session.requires_responsive_denoising:
    apply_isaac_rtx_global_settings(
        IsaacRtxRendererGlobalSettingsCfg(
            carb_settings={"/rtx/dldenoiser/responsiveDenoising": True},
        )
    )
```

Compare to `teleop_se3_agent.py` which combines both in one call. This is logically correct today (the DLSS antialiasing mode is set in an earlier block), but fragile: the two conditions are evaluated independently and a future refactor could easily separate them further. Folding responsive denoising into the same `IsaacRtxRendererGlobalSettingsCfg` as `teleop_se3_agent.py` does would make the dependency explicit.

---

### 5. `_DLSS_EXEC_MODES` set and `Literal` annotation are two sources of truth

**File**: `source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py`

```python
_DLSS_EXEC_MODES = {"performance", "balanced", "quality", "auto", "rtxaa", "manual"}

dlss_exec_mode: Literal["performance", "balanced", "quality", "auto", "rtxaa", "manual"] | None = None
```

If a new mode is added to the `Literal`, `__post_init__` validation silently misses it (and vice versa). Standard fix: derive the set from a shared tuple so only one place needs updating.

---

### 6. `_XrCameraFeedSession._env` is dead state

**File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`

`_XrCameraFeedSession` stores `self._env = env` in `bind()` and clears it in `close()`, but never reads `self._env` anywhere in the session class. The manager receives `env` directly at construction. Should be removed.

---

### 7. `_feeds_require_responsive_denoising` triple-nested `getattr` is opaque

**File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`

```python
return any(
    getattr(getattr(getattr(scene, cfg.camera_name), "renderer_cfg", None), "enable_dlss_ray_reconstruction", None)
    is True
    for cfg in cfgs
)
```

`getattr(scene, cfg.camera_name)` has no fallback — it raises `AttributeError` if the camera is absent. This is safe only because `_prepare_camera_feed_cfgs` already validated presence, but the invariant is invisible here. Local variables per step plus an explicit `getattr(scene, cfg.camera_name, None)` guard would make this correct-at-a-glance.

---

### 8. Reset-latch fix in `TeleopMessageProcessor` — no test for latch edge cases

**File**: `source/isaaclab_teleop/isaaclab_teleop/teleop_message_processor.py`

The new `_channel_reset_latched` flag debounces hold-down resets — a real behavior fix, not just a refactor. Worth adding a unit test covering: first reset fires, second same-frame reset is suppressed, button release clears the latch, next press fires again.

---

### Minor nits

- `_KitSceneUiCameraFeedPresenter._validate_image` correctly rejects non-`cuda:0` CUDA devices but the error message could explain *why* (`set_bytes_data_from_gpu` requires device 0).
- `_CameraImageWidget` hardcodes style colors inline — not a bug, but makes theming harder later.
- `_world_panel_matrix` normalizes the quaternion inline — a short comment that USD's `Quatd` expects a pre-normalized value would help readers.

---

**Overall**: The architecture is clean and the two-phase `prepare` / `bind` lifecycle is well-suited to the XR teleoperation pattern. The Kit SceneUI viewer-start anchor logic is solid. Main items worth a follow-up before merge: the `require_visual_shapes()` regression risk for non-XR camera users (#2), the private session class leaking into public entry points (#1), and the duplicate validation call (#3).

rwiltz:
> ## Code Review
> **Scope**: commits `1fbd334`, `f0c96c2`, `921e213`, `b517542`. Core changes: new `camera_feed.py` and `camera_feed_kit_scene_ui.py` in `isaaclab_teleop`; per-camera DLSS settings in `isaaclab_physx`; `require_visual_shapes()` hook in `Camera.__init__`; `TeleopMessageProcessor` reset latch.
> 
> ### 1. `_XrCameraFeedSession` is private but imported from public entry scripts
> **Files**: `scripts/environments/teleoperation/teleop_se3_agent.py:274`, `scripts/tools/record_demos.py:814`
> 
> ```python
> from isaaclab_teleop.camera_feed import _XrCameraFeedSession
> ```
> 
> `_XrCameraFeedSession` has a leading underscore (private by convention) but is the only way for callers to use the PiP lifecycle. The `__init__.pyi` doesn't export it. Entry scripts importing private internals is a stable-API risk — any refactor of `camera_feed.py` breaks a documented entry point without a deprecation notice. The fix is either to remove the underscore and add it to `__init__.pyi`, or expose a module-level factory function (`prepare_xr_camera_feed_session(env_cfg, ...)`) as the public surface.
> 
> ### 2. `require_visual_shapes()` now fires for every Camera, unconditionally
> **File**: `source/isaaclab/isaaclab/sensors/camera/camera.py:178-183`
> 
> ```python
> sim_ctx = sim_utils.SimulationContext.instance()
> if sim_ctx is not None:
>     sim_ctx.require_visual_shapes()
> ```
> 
> The intent is correct (Newton backend needs this before cloning), but this fires in every `Camera.__init__` — including headless RL training tasks that don't need visual geometry at all. `require_visual_shapes()` is a one-way latch (`_visual_shapes_required = True`). A training run with a depth camera, a contact sensor, or any non-visual camera was previously able to skip loading visual shapes; now it can't. The PR comment says "every renderer backend draws the visual-only geometry", but that's only true for cameras that actually render. Consider gating on whether the renderer is an RTX renderer (e.g. `self.cfg.renderer_cfg is not None`), rather than firing for all cameras.
> 
> ### 3. `_validate_layout_cfg` is called redundantly in the normal path
> **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> 
> `_XrCameraFeedManager.__init__` calls `_validate_layout_cfg(self._layout_cfg)` directly, and then `_layout_feed_cfgs(...)` (called from the same `__init__`) also calls `_validate_layout_cfg(layout_cfg)`. The second call is always a no-op if the first passed. Validation should live in one place — either in `_layout_feed_cfgs` alone (since that's where `layout_cfg` is consumed) or in `__init__` only.
> 
> ### 4. `record_demos.py` applies responsive denoising in a separate call without DLSS antialiasing mode
> **File**: `scripts/tools/record_demos.py:820-827`
> 
> ```python
> # separate call, no antialiasing_mode
> if camera_feed_session.requires_responsive_denoising:
>     apply_isaac_rtx_global_settings(
>         IsaacRtxRendererGlobalSettingsCfg(
>             carb_settings={"/rtx/dldenoiser/responsiveDenoising": True},
>         )
>     )
> ```
> 
> Compare to `teleop_se3_agent.py` which combines both in one call. This is logically correct today (the DLSS antialiasing mode is set in an earlier block), but fragile: the two conditions are evaluated independently and a future refactor could easily separate them further. Folding responsive denoising into the same `IsaacRtxRendererGlobalSettingsCfg` as `teleop_se3_agent.py` does would make the dependency explicit.
> 
> ### 5. `_DLSS_EXEC_MODES` set and `Literal` annotation are two sources of truth
> **File**: `source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py`
> 
> ```python
> _DLSS_EXEC_MODES = {"performance", "balanced", "quality", "auto", "rtxaa", "manual"}
> 
> dlss_exec_mode: Literal["performance", "balanced", "quality", "auto", "rtxaa", "manual"] | None = None
> ```
> 
> If a new mode is added to the `Literal`, `__post_init__` validation silently misses it (and vice versa). Standard fix: derive the set from a shared tuple so only one place needs updating.
> 
> ### 6. `_XrCameraFeedSession._env` is dead state
> **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> 
> `_XrCameraFeedSession` stores `self._env = env` in `bind()` and clears it in `close()`, but never reads `self._env` anywhere in the session class. The manager receives `env` directly at construction. Should be removed.
> 
> ### 7. `_feeds_require_responsive_denoising` triple-nested `getattr` is opaque
> **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> 
> ```python
> return any(
>     getattr(getattr(getattr(scene, cfg.camera_name), "renderer_cfg", None), "enable_dlss_ray_reconstruction", None)
>     is True
>     for cfg in cfgs
> )
> ```
> 
> `getattr(scene, cfg.camera_name)` has no fallback — it raises `AttributeError` if the camera is absent. This is safe only because `_prepare_camera_feed_cfgs` already validated presence, but the invariant is invisible here. Local variables per step plus an explicit `getattr(scene, cfg.camera_name, None)` guard would make this correct-at-a-glance.
> 
> ### 8. Reset-latch fix in `TeleopMessageProcessor` — no test for latch edge cases
> **File**: `source/isaaclab_teleop/isaaclab_teleop/teleop_message_processor.py`
> 
> The new `_channel_reset_latched` flag debounces hold-down resets — a real behavior fix, not just a refactor. Worth adding a unit test covering: first reset fires, second same-frame reset is suppressed, button release clears the latch, next press fires again.
> 
> ### Minor nits
> * `_KitSceneUiCameraFeedPresenter._validate_image` correctly rejects non-`cuda:0` CUDA devices but the error message could explain _why_ (`set_bytes_data_from_gpu` requires device 0).
> * `_CameraImageWidget` hardcodes style colors inline — not a bug, but makes theming harder later.
> * `_world_panel_matrix` normalizes the quaternion inline — a short comment that USD's `Quatd` expects a pre-normalized value would help readers.
> 
> **Overall**: The architecture is clean and the two-phase `prepare` / `bind` lifecycle is well-suited to the XR teleoperation pattern. The Kit SceneUI viewer-start anchor logic is solid. Main items worth a follow-up before merge: the `require_visual_shapes()` regression risk for non-XR camera users (#2), the private session class leaking into public entry points (#1), and the duplicate validation call (#3).

I think the ones that are most important are 1, 6, adding unit tests for 8

rwiltz:
> > ## Code Review
> > **Scope**: commits `1fbd334`, `f0c96c2`, `921e213`, `b517542`. Core changes: new `camera_feed.py` and `camera_feed_kit_scene_ui.py` in `isaaclab_teleop`; per-camera DLSS settings in `isaaclab_physx`; `require_visual_shapes()` hook in `Camera.__init__`; `TeleopMessageProcessor` reset latch.
> > ### 1. `_XrCameraFeedSession` is private but imported from public entry scripts
> > **Files**: `scripts/environments/teleoperation/teleop_se3_agent.py:274`, `scripts/tools/record_demos.py:814`
> > ```python
> > from isaaclab_teleop.camera_feed import _XrCameraFeedSession
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `_XrCameraFeedSession` has a leading underscore (private by convention) but is the only way for callers to use the PiP lifecycle. The `__init__.pyi` doesn't export it. Entry scripts importing private internals is a stable-API risk — any refactor of `camera_feed.py` breaks a documented entry point without a deprecation notice. The fix is either to remove the underscore and add it to `__init__.pyi`, or expose a module-level factory function (`prepare_xr_camera_feed_session(env_cfg, ...)`) as the public surface.
> > ### 2. `require_visual_shapes()` now fires for every Camera, unconditionally
> > **File**: `source/isaaclab/isaaclab/sensors/camera/camera.py:178-183`
> > ```python
> > sim_ctx = sim_utils.SimulationContext.instance()
> > if sim_ctx is not None:
> >     sim_ctx.require_visual_shapes()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The intent is correct (Newton backend needs this before cloning), but this fires in every `Camera.__init__` — including headless RL training tasks that don't need visual geometry at all. `require_visual_shapes()` is a one-way latch (`_visual_shapes_required = True`). A training run with a depth camera, a contact sensor, or any non-visual camera was previously able to skip loading visual shapes; now it can't. The PR comment says "every renderer backend draws the visual-only geometry", but that's only true for cameras that actually render. Consider gating on whether the renderer is an RTX renderer (e.g. `self.cfg.renderer_cfg is not None`), rather than firing for all cameras.
> > ### 3. `_validate_layout_cfg` is called redundantly in the normal path
> > **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> > `_XrCameraFeedManager.__init__` calls `_validate_layout_cfg(self._layout_cfg)` directly, and then `_layout_feed_cfgs(...)` (called from the same `__init__`) also calls `_validate_layout_cfg(layout_cfg)`. The second call is always a no-op if the first passed. Validation should live in one place — either in `_layout_feed_cfgs` alone (since that's where `layout_cfg` is consumed) or in `__init__` only.
> > ### 4. `record_demos.py` applies responsive denoising in a separate call without DLSS antialiasing mode
> > **File**: `scripts/tools/record_demos.py:820-827`
> > ```python
> > # separate call, no antialiasing_mode
> > if camera_feed_session.requires_responsive_denoising:
> >     apply_isaac_rtx_global_settings(
> >         IsaacRtxRendererGlobalSettingsCfg(
> >             carb_settings={"/rtx/dldenoiser/responsiveDenoising": True},
> >         )
> >     )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Compare to `teleop_se3_agent.py` which combines both in one call. This is logically correct today (the DLSS antialiasing mode is set in an earlier block), but fragile: the two conditions are evaluated independently and a future refactor could easily separate them further. Folding responsive denoising into the same `IsaacRtxRendererGlobalSettingsCfg` as `teleop_se3_agent.py` does would make the dependency explicit.
> > ### 5. `_DLSS_EXEC_MODES` set and `Literal` annotation are two sources of truth
> > **File**: `source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py`
> > ```python
> > _DLSS_EXEC_MODES = {"performance", "balanced", "quality", "auto", "rtxaa", "manual"}
> > 
> > dlss_exec_mode: Literal["performance", "balanced", "quality", "auto", "rtxaa", "manual"] | None = None
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If a new mode is added to the `Literal`, `__post_init__` validation silently misses it (and vice versa). Standard fix: derive the set from a shared tuple so only one place needs updating.
> > ### 6. `_XrCameraFeedSession._env` is dead state
> > **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> > `_XrCameraFeedSession` stores `self._env = env` in `bind()` and clears it in `close()`, but never reads `self._env` anywhere in the session class. The manager receives `env` directly at construction. Should be removed.
> > ### 7. `_feeds_require_responsive_denoising` triple-nested `getattr` is opaque
> > **File**: `source/isaaclab_teleop/isaaclab_teleop/camera_feed.py`
> > ```python
> > return any(
> >     getattr(getattr(getattr(scene, cfg.camera_name), "renderer_cfg", None), "enable_dlss_ray_reconstruction", None)
> >     is True
> >     for cfg in cfgs
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `getattr(scene, cfg.camera_name)` has no fallback — it raises `AttributeError` if the camera is absent. This is safe only because `_prepare_camera_feed_cfgs` already validated presence, but the invariant is invisible here. Local variables per step plus an explicit `getattr(scene, cfg.camera_name, None)` guard would make this correct-at-a-glance.
> > ### 8. Reset-latch fix in `TeleopMessageProcessor` — no test for latch edge cases
> > **File**: `source/isaaclab_teleop/isaaclab_teleop/teleop_message_processor.py`
> > The new `_channel_reset_latched` flag debounces hold-down resets — a real behavior fix, not just a refactor. Worth adding a unit test covering: first reset fires, second same-frame reset is suppressed, button release clears the latch, next press fires again.
> > ### Minor nits
> > 
> > * `_KitSceneUiCameraFeedPresenter._validate_image` correctly rejects non-`cuda:0` CUDA devices but the error message could explain _why_ (`set_bytes_data_from_gpu` requires device 0).
> > * `_CameraImageWidget` hardcodes style colors inline — not a bug, but makes theming harder later.
> > * `_world_panel_matrix` normalizes the quaternion inline — a short comment that USD's `Quatd` expects a pre-normalized value would help readers.
> > 
> > **Overall**: The architecture is clean and the two-phase `prepare` / `bind` lifecycle is well-suited to the XR teleoperation pattern. The Kit SceneUI viewer-start anchor logic is solid. Main items worth a follow-up before merge: the `require_visual_shapes()` regression risk for non-XR camera users (#2), the private session class leaking into public entry points (#1), and the duplicate validation call (#3).
> 
> I think the ones that are most important are 1, 6, adding unit tests for 8

Looking more closely at this feedback, I think 2 is a red herring from develop, not your PR. #8 I think is wrong as well.
Also 4 and 7 are probably just readability issues.

hougantc-nvda:
Addressed the actionable items from the Claude/Rafael follow-up in [`8143d9e0`](https://github.com/isaac-sim/IsaacLab/commit/8143d9e0eaafeaf581ff0e81d73fe5de5e76b326):

- **#1:** Promoted `XrCameraFeedSession` to the supported public API, exported it from `isaaclab_teleop`, updated both entry scripts to use the public import, and added API docs/changelog/test coverage.
- **#3:** Removed the redundant manager-level layout validation; `_layout_feed_cfgs()` remains the single validation owner.
- **#5:** Made the DLSS execution-mode `Literal` the single source of truth and derive the runtime validation set from it.
- **#6:** Removed the unused session `_env` state.

No code change for the remaining items:

- **#2:** `require_visual_shapes()` is inherited from the PR base and was not introduced or modified by this PR.
- **#4:** The current behavior is intentional and correct: `record_demos.py` applies its DLSS mode during environment setup, then enables responsive denoising only after the selected feeds are known.
- **#7:** Camera presence/type is established by `_prepare_camera_feed_cfgs()` before this helper runs; this is readability feedback rather than a supported failure path.
- **#8:** `test_repeated_reset_frames_produce_one_pulse_until_released` already covers the initial press, held frames, release, and re-press.
- The minor nits are non-functional style/documentation preferences, so I left them unchanged.

Validation on this patch: 25 focused XR camera-feed tests and 29 renderer contract/config tests pass; `pre-commit` passes.

rwiltz (approved):
LGTM, testing working locally as well

source/isaaclab_tasks/isaaclab_tasks/contrib/robot_pov_camera_cfg.py
```diff
@@ -0,0 +1,44 @@
...
+
+from isaaclab_tasks.utils.presets import MultiBackendRendererCfg
+
+
+@configclass
+class _RobotPovCameraRendererCfg(MultiBackendRendererCfg):
+    """Renderer presets shared by the recorded robot-PoV cameras."""
+
+    default: IsaacRtxRendererCfg = IsaacRtxRendererCfg(
+        camera_output_device="cuda:0",
```
kellyguo11: is this expected behavior to always hard code this to cuda:0?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -82,6 +82,11 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
     "simple_shading_full_mdl": 3,
 }
 SIMPLE_SHADING_MODE_SETTING = "/rtx/minimal/mode"
+DLSS_RAY_RECONSTRUCTION_API_SCHEMA = "OmniRtxDebugSettingsAPI_1"
```
kellyguo11: can these be incorporated into IsaacRtxRendererGlobalSettingsCfg? not sure if we really need all these global variables here

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py
```diff
@@ -5,11 +5,14 @@
 
 """Configuration for Isaac RTX (Replicator) Renderer."""
 
-from typing import Any, Literal
+from typing import Any, Literal, get_args
 
 from isaaclab.renderers.renderer_cfg import RendererCfg
 from isaaclab.utils.configclass import configclass
 
+_DlssExecMode = Literal["performance", "balanced", "quality", "auto", "rtxaa", "manual"]
```
kellyguo11: can this also be included in IsaacRtxRendererGlobalSettingsCfg?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py
```diff
@@ -172,3 +201,28 @@ class IsaacRtxRendererCfg(RendererCfg):
     - ``"zero"``: Values are clipped to zero.
     - ``"none"``: No clipping is applied. Values will be returned as ``inf``.
     """
+
+    def __post_init__(self) -> None:
```
kellyguo11: is this really necessary?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -438,6 +451,62 @@ def create_render_data(self, spec: CameraRenderSpec) -> IsaacRtxRenderData:
             ppisp_pipeline=ppisp_pipeline,
         )
 
+    def _apply_render_product_settings(self, stage: Usd.Stage, render_product_paths: list[str]) -> None:
```
kellyguo11: this can also be combined together with IsaacRtxRendererGlobalSettingsCfg

source/isaaclab/isaaclab/renderers/base_renderer.py
```diff
@@ -44,6 +44,21 @@ def initialize(self) -> None:
         """Post-physics one-time initialization hook. Called only once."""
         return
 
+    def resolve_camera_output_device(self, simulation_device: str) -> str:
```
kellyguo11: Could we keep this device-placement workaround out of the core Camera/Renderer APIs? camera_output_device makes Camera.device ambiguous by placing camera state and pixel outputs on different devices, adds device parsing and a hook to every renderer, and hardcodes cuda:0 for teleop. 

Could we keep camera output on the simulation device and handle presentation in the teleop/SceneUI layer? The current implementation mixes camera and renderer contract on the device and it's quite confusing for other workflows.

kellyguo11:
@pbarejko @rilei-nvidia for viz on camera/renderer related changes

source/isaaclab_tasks/isaaclab_tasks/contrib/robot_pov_camera_cfg.py
hougantc-nvda: Addressed in 0bc08c6dd8. The shared robot-PoV task preset no longer sets camera_output_device or hard-codes cuda:0. Task camera outputs remain on the simulation device; the optional teleop/SceneUI presentation path now attaches its own CUDA annotator to the existing RTX render product and discovers the producing GPU dynamically, including multi-GPU configurations, with an explicit CPU fallback.

source/isaaclab/isaaclab/renderers/base_renderer.py
hougantc-nvda: Addressed in 0bc08c6dd8. BaseRenderer.resolve_camera_output_device, the Isaac RTX override and config field, and the Camera split-device allocation were removed. Camera state and outputs now stay on the simulation device, while the teleop/SceneUI presenter owns its feed-specific CUDA annotator on the existing render product and resolves the actual render GPU dynamically. The core, PhysX, task, and teleop package CI jobs are green on this commit.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
hougantc-nvda: I kept these settings on `IsaacRtxRendererCfg`. `IsaacRtxRendererGlobalSettingsCfg` writes process-global Carb settings, while these two values are USD API attributes authored on each individual `RenderProduct`. Making them global would also change the XR headset and bystander render products. The two-camera integration test gives the cameras different DLSS execution modes and Ray Reconstruction values and verifies that both remain isolated in USD and Fabric after annotator attachment; both GPU integration tests passed on Horde. Would you be comfortable keeping these settings render-product-local?

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py
hougantc-nvda: Addressed in [`5a4b5885c0`](https://github.com/isaac-sim/IsaacLab/pull/6818/commits/5a4b5885c099b9b67e8d93704dc906a6159e817f). I removed `_DlssExecMode` and `_DLSS_EXEC_MODES`; `dlss_exec_mode` now uses an inline `Literal`. The field remains on `IsaacRtxRendererCfg` because it is authored per `RenderProduct`, rather than as a process-global setting.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer_cfg.py
hougantc-nvda: Addressed in [`5a4b5885c0`](https://github.com/isaac-sim/IsaacLab/pull/6818/commits/5a4b5885c099b9b67e8d93704dc906a6159e817f). I removed the custom `__post_init__` validation and its invalid-value tests, matching the existing config-class convention. The typed field declarations remain the configuration contract without adding a separate eager runtime-validation layer.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
hougantc-nvda: Addressed in [`5a4b5885c0`](https://github.com/isaac-sim/IsaacLab/pull/6818/commits/5a4b5885c099b9b67e8d93704dc906a6159e817f). These were single-use USD schema/attribute names and a compatibility threshold, so I removed the module-level constants and inlined them at their use sites. I kept them out of `IsaacRtxRendererGlobalSettingsCfg` because that class represents process-global Carb settings, whereas these values describe per-`RenderProduct` USD attributes.

hougantc-nvda:
@AntoineRichard @kellyguo11 — the review fixes and purpose-scope cleanup are now on final head `43c83a16629407834311a566acdb7a4175b84049`.

- Antoine: all requested inline changes were implemented, replied to, and resolved; could you please re-review and clear the previous changes-requested state if satisfied?
- Kelly: three comments were implemented and resolved. One design thread remains open: the DLSS execution mode and Ray Reconstruction values are USD attributes authored per `RenderProduct`, while `IsaacRtxRendererGlobalSettingsCfg` owns process-global Carb settings. We kept them render-product-local so camera-specific settings do not change the XR headset or bystander render products. Could you please confirm whether that separation is acceptable?
- GPU validation on Horde passed: camera preset tests 5/5, renderer/config/contract tests 31/31, renderer integration 2/2, and the committed G1 camera follows the robot and remains stable across reset with valid RGB output.
- Local validation passed: 83 focused tests, full pre-commit, warning-as-error docs build, changelog check, and `git diff --check`. Fresh CI on this final head is running with no failures so far.

The recursive PiP hall-of-mirrors behavior is intentionally deferred and is not claimed fixed by this PR.

kellyguo11 (approved):

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
hougantc-nvda: Addressed in `b93b1083`. Runtime probing showed that ordinary-layer RenderProduct opinions could be overwritten by later legacy/global RTX synchronization, so the two settings and their authoring were removed entirely from `IsaacRtxRendererCfg` and `isaaclab_physx`. Optional PiP policy now lives on `XrCameraFeedCfg`, and the private Kit Scene UI adapter authors only the selected feed RenderProduct in the USD session layer after PiP annotator attachment. A forced contrary-sync probe remained isolated through update 140, while viewport/bystander RenderProducts were unchanged. The composed `isaaclab_physx` diff is now zero, making this renderer thread obsolete; I’m resolving it and requesting review of the new head.

hougantc-nvda:
@kellyguo11 @AntoineRichard — the final architecture is now on head `b93b1083b47dc4db53aa6c06d6070af90fcdb5e9`.

Runtime persistence testing after the previous review found that ordinary-layer RenderProduct opinions could be overwritten by later legacy/global RTX synchronization. The promoted head therefore:

- removes the DLSS/RR fields and all authoring from `IsaacRtxRendererCfg` / `isaaclab_physx`;
- keeps optional policy on `XrCameraFeedCfg` and applies it only to the selected PiP RenderProduct from the private Kit adapter;
- authors a transient session-layer opinion after PiP annotator attachment; a forced contrary-sync probe remained correct through update 140 while viewport/bystander RenderProducts were unchanged;
- leaves backend-neutral IsaacLab core and `isaaclab_physx` production code unchanged;
- uses the headset-tested viewer-start placement and `-0.15 m` offset for both G1 and GR1T2.

Validation completed before promotion: full pre-commit, 88 focused tests, 2/2 real Kit integration tests, 101/101 persistence/isolation/cleanup checks, clean G1 motion/reset RGB validation, fresh IWER/OpenXR/CloudXR validation, and Quest 3 acceptance of G1 PiP, GR1T2 PiP, and the disabled-camera/no-PiP path. Fresh CI is running on the promoted head.

Kelly’s renderer thread is now obsolete, replied to, and resolved. Since the architecture changed after the prior approval, could you both please re-review this exact head? Antoine’s earlier inline requests remain addressed and resolved.

The recursive PiP hall-of-mirrors behavior remains intentionally deferred and is not claimed fixed by this PR.

hougantc-nvda:
@kellyguo11 — a fifth Docker attempt was just queued on old head `dc3062a`; I canceled it before the runtime matrix materialized. Please stop retrying this run for now. Those retries cannot become the final required result because this branch must still refresh onto #6917, and each attempt competes with #6917 for the same GPU runners. I am actively monitoring both PRs and will start the replacement run on the refreshed head.

kellyguo11:
sounds good. I was going to get this one merged in first, but will leave it to you

kellyguo11 (approved):

PR #6460, opened 2026-07-10
[App] Add support for kitless URDF/MJCF importer

hujc7:
## 1. Summary

- URDF and MJCF command-line conversion no longer requires a full Isaac Sim installation: the importer APIs resolve from Kit extensions when Kit is running, and from the standalone `isaacsim-asset-isolated` wheel otherwise.
- Centralizes the Kit lifecycle facts that kitless-capable tools kept rediscovering: `AppLauncher.is_available()`, `AppLauncher.has_gui()`, and `sim.utils.show_stage_in_viewport()`.
- Adds a shared converter CLI so `convert_urdf.py` / `convert_mjcf.py` accept only their own arguments plus `--viz`, with previews on both backends (`kit` in the Isaac Sim viewport; `newton` / `rerun` / `viser` kitlessly).
- Converter flags move to `snake_case` (`--merge_joints`, `--merge_mesh`, ...) per the repository convention, with the hyphenated spellings kept as aliases so existing commands keep working.
- **52 converter tests now run against the standalone wheel** (20 URDF, 12 MJCF, 9 CLI, 11 importer API), where the wheel path previously had no correctness coverage.

## 2. What changed

| Area | Change |
|---|---|
| `sim/converters/_importer_api.py` (new) | `ImporterProvider` loads `isaacsim.asset.importer.{urdf,mjcf}` from Kit (via `sim.utils.enable_extension`, gated by `has_kit()`) or from the standalone wheel. Wheel availability and loadability are checked inline; the load check runs in a subprocess because a conflicting OpenUSD build can abort the interpreter instead of raising. |
| `sim/converters/_converter_cli.py` (new) | `ConverterCli` decides once per process whether conversion runs through Kit or the wheel, contributes `--viz`, and previews after conversion. |
| `app/app_launcher.py` | `is_available()` (Kit *launchable*) and `has_gui()` (resolved GUI state). Distinct from `has_kit()`, which reports whether Kit is *already running*. |
| `sim/utils/stage.py` | `show_stage_in_viewport()` opens a USD in the Kit viewport and blocks until closed. Consumed by `--viz kit` and `convert_mesh.py`. |
| `scripts/tools/convert_{urdf,mjcf}.py` | Bootstrap through `ConverterCli`; AppLauncher flags are no longer accepted. |
| Tests | Converter tests select the backend from the installed distribution — wheel when present, Isaac Sim otherwise — and use repository/`newton` assets so both backends convert the same inputs. |

## 3. Behavior notes

[...]

PR #6824, opened 2026-07-31
[App] Bootstrap URDF/MJCF converters via launch_simulation

ooctipus:
## 1. Summary

Folds the URDF/MJCF converter bootstrap into `launch_simulation`, the launcher every other tool
and demo script already uses. `ConverterCli` and `ImporterProvider` (added in #6460) are removed:
their launch decision, teardown, and visualizer handling already exist in `launch_simulation`,
and their `--viz` flag already exists as the launcher's `--visualizer` / `--viz`.

Net **+379 / −850**.

This is a follow-up to #6460, not a revert of it. The kitless importer support that PR added is
kept and extended — the standalone wheel is now the *preferred* backend rather than unreachable
whenever Isaac Sim happens to be installed.

## 2. What changed

| Area | Change |
|---|---|
| `app/sim_launcher.py` | `launch_simulation` reads a `require_kit` key from `launcher_args`. The one thing the launcher could not express is a tool that needs Kit for a reason its config does not describe. Additive: it can only turn a kitless launch into a Kit one, never the reverse. |
| `sim/converters/_converter_cli.py` | Deleted (204 lines). `launch_simulation` derives `needs_kit` from the scanned config, skips `AppLauncher` when Kit is already running, closes the app in a `finally` with exit-code propagation, validates backend combinations, and syncs the kitless visualizer selection to carb. |
| `sim/converters/_importer_api.py` | Deleted (121 lines). Loading the importer is a `has_kit()` guard around the extension enable plus an ordinary import — the wheel exposes the same `isaacsim.asset.importer.*` namespace and needs no code of its own. |
| `sim/converters/{urdf,mjcf}_converter.py` | Inline the guarded import in `_convert_asset`, 7 lines each. |
| `scripts/tools/convert_{urdf,mjcf}.py` | Standard `add_launcher_args` + `launch_simulation` preamble, matching their sibling `convert_mesh.py`. `args_cli.require_kit` is set from the presence of the `isaacsim-asset-isolated` wheel. |
| `app/app_launcher.py` | Separate commit: fix `--help` on scripts with required positional arguments (see below). |
| Tests | `test_converter_cli.py` and `test_importer_api.py` deleted; `test_launch_simulation_require_kit.py` added (3 tests); one `--help` regression test added to `test_argparser_launch.py`. |

## 3. Why the flag came back

`--viz` is not a new flag. `AppLauncher.add_app_launcher_args` has defined `--visualizer` / `--viz`
for a while, taking a validated CSV of `kit,newton,rerun,viser,none`, and 57 scripts get it from
one line. #6460 removed `add_app_launcher_args` from these two scripts and reintroduced `--viz`
locally with a different grammar, so `--viz kit,newton` worked everywhere except the converters.

Because the local flag's backend choice was driven by `AppLauncher.is_available()` — "is Isaac Sim
installed", not "is Kit needed" — the wheel path was unreachable whenever Isaac Sim was present,
and the kitless visualizers had to be rejected with `parser.error` to cover for it. Deriving
`needs_kit` from the wheel instead removes both the rejection and the flag divergence.

## 4. Why `require_kit` is a launcher argument, not a parameter

`launcher_args` is already the channel for launch signals a script contributes: `sim_launcher`
reads `physics` and `distributed` off it without `add_launcher_args` ever defining them, and writes
`visualizer_intent` into it as a purely config-derived signal. A dedicated keyword argument would
have been a second mechanism for the same class of information, plus a new symbol on the public
`launch_simulation` signature.

Carrying the extra key is safe: `AppLauncher` builds `_sim_app_config` by intersecting
`launcher_args` with `_SIM_APP_CFG_TYPES`, so `require_kit` is filtered out before `SimulationApp`
sees it — the same path `visualizer_intent` already takes.

## 5. `--help` fix (first commit)

`add_app_launcher_args` probed the command line with `parse_known_args()` to check for name
collisions with the SimulationApp config fields. That probe exits when a required argument is
missing, which is the case for every tool script that takes required positionals and is invoked
with `--help`. On `develop` today:

```console
$ python scripts/tools/convert_mesh.py --help
convert_mesh.py: error: the following arguments are required: input, output
```

The check validates the parser's configuration rather than a particular invocation -- it reports
arguments duplicating names `add_app_launcher_args` owns, and arguments whose type SimulationApp
cannot ingest -- so it now reads the declared arguments off the parser instead of parsing argv:

```python
config = {action.dest: action.default for action in parser._actions if action.dest != argparse.SUPPRESS}
```

`parse_known_args` returns each argument's default when it is absent from the command line, so this
is the same mapping for everything the user did not pass. The function already reaches into
`parser._actions` / `parser._option_string_actions` a few lines above, to pop and restore the help
action for this same eager-exit reason; this finishes that workaround's other half.

This fixes `convert_mesh.py` as well as the two rewritten scripts. Happy to split it into its own
PR if preferred -- it is self-contained and touches shared code.

## 6. Behavior changes

- **Breaking:** `--viz auto` and a bare `--viz` are no longer accepted. Name the backend
  (`--viz kit`, `--viz newton`). The spelling they replace shipped in 15.0.0.
- `--viz kit,newton`, `--device`, `--livestream`, and `--experience` are accepted again on both
  converters (#6460 dropped `add_app_launcher_args`, which the review bot flagged as removing CLI
  surface without a deprecation period).
- The `isaacsim-asset-isolated` wheel is preferred over Kit when both are installed; Kit is
  launched only when the wheel is absent or a Kit preview is requested. `--viz newton` with Isaac
  Sim installed is now possible rather than a hard error.
- The subprocess preflight that validated the wheel before every kitless conversion is gone. It
  cost a full interpreter start plus a heavy import per run to turn a crash into a nicer message,
  and could not prevent the crash it described.

Changelog fragment is filed as `.minor.rst` on the reasoning that the `--viz` spelling it breaks
shipped in the release dated the same day; say the word if `.major.rst` is preferred.

## 7. Validation

Unit tests and lint:

- [x] `test_launch_simulation_require_kit.py` (4 tests, dict and namespace launcher args) —
      verified failing without the `require_kit` change.
- [x] `--help` regression test in `test_argparser_launch.py` — verified failing without the
      `app_launcher.py` change.
- [x] `test_argparser_launch.py` + `test_kwarg_launch.py` + the new file: 46 passed. A pristine
      `upstream/develop` worktree in the same environment gives 41 passed with the *identical*
      1 failure (`test_matrix_headless_with_viz_names_takes_precedence`) and 2 errors (missing
      `pytest-mock`), so both are pre-existing and unrelated.
- [x] Full `pre-commit` suite on all changed files.

Both converters run against real assets (`source/isaaclab/test/sim/urdfs/test_merge_joints.urdf`
and newton's bundled `nv_ant.xml`) in a kitless environment — newton installed, no Isaac Sim, no
importer wheel:

- [x] **Neither importer source installed** — fails immediately after argument parsing, naming both
      Isaac Sim and the `isaacsim-asset-isolated` wheel, before any conversion work. An earlier
      revision of this branch regressed here: it fell through to the launcher's generic diagnostic,
      which blames the PhysX backend and Kit visualizer and never mentions the wheel. Fixed in
      `Report both importer sources when neither is installed`.
- [x] **Wheel present** (simulated with a `.dist-info` on `PYTHONPATH` so the distribution probe
      resolves) — `require_kit` is `False`, `launch_simulation` takes the kitless branch and starts
      no Kit app, `UrdfConverter` is constructed, the config is translated, `_convert_asset` is
      reached, and `has_kit()` correctly skips `enable_extension`. Execution stops only at
      `from isaacsim.asset.importer.urdf import ...`, i.e. the real wheel contents.
- [x] `--help`, `--viz kit,newton`, and `--device` on both scripts, plus `--help` on
      `convert_mesh.py`.

Not covered here — needs a reviewer with Isaac Sim or the wheel installed:

- [ ] A conversion that actually produces USD, on either backend.
- [ ] The `--viz kit` viewport preview and the kitless `newton` / `rerun` / `viser` previews.
- [ ] `test_urdf_converter.py` / `test_mjcf_converter.py`, which need one of the two installs.


kellyguo11 (approved):

ErenAta16 (commented):
A `-848` line removal is hard to review by reading, so I checked the two things a deletion of this shape usually gets wrong. Both are clean.

## Nothing is left pointing at the removed modules

```
_converter_cli    0 references
ConverterCli      0
_importer_api     0
ImporterProvider  0
```

across `.py`, `.rst` and `.md` on the branch. The removal is complete rather than partial.

## The CLI surface, including a false alarm worth recording

I diffed the `add_argument` calls in both converter scripts against `main`, and my first pass looked alarming:

```
convert_mjcf.py
  main : --fix-base, --import-sites, --make-instanceable
  PR   : --merge_mesh, --collision_from_visuals, --collision_type,
         --self_collision, --import_physics_scene
```

Three flags gone, five new ones, and the naming convention flipped from hyphen to underscore. Two checks later, neither is a problem:

- `migrating_to_isaaclab_3-0.rst:1858` already states those three flags "are no longer available", and that file is on `main` and untouched here. So the code was lagging the documented 3.0 migration and this PR closes the gap rather than opening one.
- Every new flag registers **both** spellings (`"--merge_mesh", "--merge-mesh"`), so the migration guide's `--merge-mesh` / `--self-collision` examples work against the implementation. `convert_urdf.py` does the same for `--fix_base` / `--fix-base`.

Recording this because a reviewer diffing the argparse surface will hit the same three-flags-missing result, and the reassurance is two files away.

`convert_urdf.py` lost nothing and gained five flags that previously came from `ConverterCli`, which matches the description.

## Two small things

**The usage block in `convert_urdf.py` lists one spelling.** Line 25 documents `--fix_base` while line 54 registers `--fix_base, --fix-base`. Same for the others. Harmless, but someone reading the header won't learn the hyphen form exists, and the migration guide uses hyphens throughout — so the two docs teach different spellings for the same flag.

**`require_kit` is overwritten unconditionally.**

```python
try:
    metadata.distribution("isaacsim-asset-isolated")
    args_cli.require_kit = False
except metadata.PackageNotFoundError:
    args_cli.require_kit = True
```

`sim_launcher.py:493` documents it as "whether the caller needs Kit for a reason *cfg* cannot express", and it's read at 524 through `_get_arg`. Today nothing registers it as a CLI flag, so there's no user value to discard and the assignment is the intended channel. Worth a short comment saying that, though — the name reads like an override the caller might have set, and if `add_launcher_args` ever grows a `--require_kit`, these two lines would silently win over it with no indication.

The early `ImportError` naming the wheel is a good touch. Reporting "neither Isaac Sim nor the standalone importer wheel is installed" is a much better first failure than the launcher's generic "Isaac Sim is absent", which is exactly the confusion the description mentions.

AntoineRichard (approved):

PR #6826, opened 2026-07-31
Set convex hull collision meshes for DR Legs

aserifi:
# Description

Until #6596 the Newton cloner imported with `skip_mesh_approximation=True` and convex-hulled every collision mesh, so DR Legs implicitly ran with convex-hull feet even though the USD authors `boundingCube` on the foot collision meshes. Now that authored approximations are fixed and the feet load as boxes, changing the contact geometry the task was tuned against, so this sets `convexHull` explicitly on the DR Legs spawn to keep the previous behavior.

### DIFF
--- a/source/isaaclab_assets/changelog.d/aserifi-set-drlegs-convexhull.rst
+++ b/source/isaaclab_assets/changelog.d/aserifi-set-drlegs-convexhull.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed the DR Legs feet colliding as bounding boxes by setting an explicit ``convexHull`` mesh
+  approximation on :data:`~isaaclab_assets.robots.dr_legs.DR_LEGS_IMPLICIT_PD_CFG`.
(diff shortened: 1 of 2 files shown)


rubengrandia:
Is there a reason we prefer ConvexHull over the boxes? The feet look like boxes to me

aserifi:
I’ve noticed some performance issues with the boxes. The sole grows by about 25%, and the simulator becomes unstable with the current solver settings. While we could tune this later, given the code freeze tonight, it’s better to set it to the old behavior.

rubengrandia:
> I’ve noticed some performance issues with the boxes. The sole grows by about 25%, and the simulator becomes unstable with the current solver settings. While we could tune this later, given the code freeze tonight, it’s better to set it to the old behavior.

Sounds reasonable. We should look into the wrong box sizing and the unstable sim separately.

rubengrandia (approved):

AntoineRichard (approved):

PR #6838, opened 2026-08-01
Fix standalone demo launch compatibility

kellyguo11:
# Description

Fix standalone demo launch compatibility across physics presets, visualizers, and the uv-managed environment.

- Let `h1_locomotion.py` accept and resolve the supported `isaacsim_physx` preset, load the versioned published H1 checkpoint, and fail clearly if that checkpoint is unavailable. The H1 demo remains documented and smoke-tested as Kit-only because its controls depend on the Kit viewport APIs.
- Make `multi_mesh_raycaster.py` open the requested Rerun viewer and allow the `isaacsim`, `rerun`, and `viser` extras to resolve together through a compatible WebSockets override.
- Replace legacy installer and pip-only runtime guidance with `uv run --extra ... <command>` hints for Rerun, Viser, OvPhysX, and OVRTX while preserving the original import error.
- Use the explicit `isaacsim_physx` selector in the PhysX/Newton policy-transfer documentation and agent skill.

The failures came from the H1 script not declaring or resolving its physics selector, the published checkpoint retaining the legacy versioned task name, Rerun using a non-browser-opening default in the interactive raycaster demo, and Isaac Sim's WebSockets pin conflicting with Viser's supported API range. No new package dependency is introduced.

## Type of change

- Documentation update

## Validation

- `uv run --extra test python -m pytest source/isaaclab/test/app/test_standalone_scripts.py source/isaaclab/test/cli/test_uv_run_pyproject.py source/isaaclab/test/visualizers/test_visualizer.py source/isaaclab_ov/test/test_ovrtx_runtime_imports.py source/isaaclab_ovphysx/test/test_runtime_imports.py` — 56 passed, 290 skipped.
- `uv run --extra test python -m pytest source/isaaclab_visualizers/test/test_viser_runtime_imports.py` — 1 passed.
- Confirmed all seven new regression assertions fail with the implementation changes temporarily removed and pass after restoration.
- `uv run --extra isaacsim,rerun,viser python scripts/demos/sensors/multi_mesh_raycaster.py --num_envs 1 --asset_type objects --physics=isaacsim_physx --viz rerun,viser` — reached scene setup and spawned both viewer servers without the reported Rerun binding error.
- `uv run --extra isaacsim python scripts/demos/h1_locomotion.py --physics=isaacsim_physx --viz kit --help` — accepted the supported launch options.
- `uv run isaaclab -d` — documentation build succeeded without warnings.
- `uv run --no-project python tools/skills/cli.py check` — validated 20 skills.
- `uv run python tools/changelog/cli.py check develop` and `compile --all --dry-run` — passed.
- `uv run isaaclab -f` — all hooks passed before commit and again before push.

## Screenshots

Not applicable; these changes affect launch behavior and error guidance.

## Checklist


source/isaaclab_ov/test/test_ovrtx_runtime_imports.py
AntoineRichard, line 1: Do we really need this test?

source/isaaclab_visualizers/test/test_viser_runtime_imports.py
AntoineRichard, line 1: Do we really need this test?

AntoineRichard (approved):
LGTM. An agent flagged that the websocket lib change may be a bit too broad.

pyproject.toml
```diff
@@ -391,18 +391,19 @@ override-dependencies = [
     "torchaudio==2.11.0",
     # Isaac Sim 6.0 pins 4.12.2, which lacks Sentinel required by the Newton viewer stack.
     "typing-extensions>=4.15.0",
+    # Isaac Sim 6.0 pins 12.0, while Viser requires the current WebSockets API.
+    "websockets>=13.1,<17.0.0",
```
AntoineRichard: The global WebSockets override also applies when Isaac Sim is selected without Viser, so this range replaces the exact WebSockets 12.0 pin from Isaac Sim 6.0.1.0 with the current highest allowed version, 16.1.1. WebSockets 14 switched its default asyncio implementation and documented backwards-incompatible API changes, which makes this a broader Isaac Sim compatibility change than the combined-viewer smoke test covers. Could we pin websockets==13.1 instead? That is the minimum supported by Viser 1.0.30 and stays closer to the dependency tested by Isaac Sim while retaining the required Viser API.

```suggestion
    "websockets==13.1",
```

ErenAta16 (commented):
`uv.lock` is about 60% of this diff, so I checked the parts a reviewer can actually reason about: the repeated demo block, and the library-side error paths.

## The repeated block is genuinely identical

Ten demo scripts gain a `--physics` argument. Hashing the added lines per file:

```
9 files, byte-identical block:
  pick_and_place.py, cameras.py, contact_sensor.py, frame_transformer_sensor.py,
  imu_sensor.py, multi_mesh_raycaster_camera.py, pva_sensor.py,
  raycaster_sensor.py, tacsl_sensor.py

1 file, larger block: h1_locomotion.py   (preset resolution + checkpoint handling)
1 file, unrelated:    multi_mesh_raycaster.py  (Rerun viewer wiring)
```

No copy-paste drift across the nine, which is the usual failure for a change of this shape.

## The flag is not dead code, and it's worth saying so

My first read was that the nine scripts add an argument nothing consumes — `args_cli.physics` appears zero times in all nine, and only once in `h1_locomotion.py`. It isn't dead: `sim_launcher.py:261` reads it off the namespace,

```python
physics_str = _get_arg(launcher_args, "physics", None)
```

so declaring the flag is how a script hands the launcher its override. Recording that because the next reviewer will grep the same way and reach the same wrong conclusion.

## What it does change for those nine

`_get_arg(..., "physics", None)` means a script that doesn't declare the flag passes no override and the config's own physics stands. These nine now declare it with `default="isaacsim_physx"`, so they always pass an explicit override — and with `choices=["isaacsim_physx"]` it can never be anything else.

That's a pin, not a user choice: the flag can only take its default, and passing anything else gets argparse's generic

```
error: argument --physics: invalid choice: 'newton' (choose from 'isaacsim_physx')
```

which says what is rejected but not why. The description explains the restriction for `h1_locomotion.py` — Kit viewport APIs — and says nothing about the other nine. If they are equally PhysX-only, a sentence saying so would help; if some would run under another backend, the pin now prevents it. Either way, a second supported backend means editing nine files rather than one shared helper.

## Library side reads correctly

`_VISUALIZER_EXTRAS` covers `kit`/`rerun`/`viser` but not `newton`, while the guard above tests four types. That's deliberate rather than an omission — `newton` falls to the `extra is None` branch and gets "Synchronize the project environment with: `uv sync`" instead of a `--extra newton` that doesn't exist. Worth a one-line comment saying newton is intentionally absent, since a dict missing one member of the tuple immediately above it reads like a bug.

Keeping the original error via `from exc` in the Viser path and in `create_visualizer` is the right call — the new message tells the user what to run, the chained exception still says what actually failed to import.

PR #6839, opened 2026-08-01
Refactor articulation actuator ownership

AntoineRichard:
# Description

This PR adds a backend-neutral actuator runtime around `ActuatorCollection` and
separates actuator-model state from simulated joint state. It keeps one scoped
collection per articulation while presenting one command, processed-command,
telemetry, group-access, and lifecycle API across PhysX, Newton, and OVPhysX.

The two main goals are:

1. make actuator ownership and runtime access consistent across backends;
2. batch compatible actuator groups to reduce launch overhead without changing
   their named configuration or access.

## Ownership and public API

- `Articulation` owns actuator application and backend submission.
- `ArticulationData` owns live solver joint properties, including joint gains,
  solver limits, armature, and friction.
- `ActuatorCollection` owns articulation-wide commands, processed commands,
  effort telemetry, group lookup, and actuator lifecycle.
- Each named group owns only its actuator-model parameters and state.
- Backend `ActuatorControl` implementations own ordering conversion,
  joint-property writes, command submission, and native-controller integration.
- Newton-native controllers retain controller parameters in Newton storage.

```python
# Commands received by actuator models.
robot.actuators.command.position
robot.actuators.command.velocity
robot.actuators.command.effort
robot.actuators.command.set_position_index(value=position_command)

# Commands produced for simulated joints on the Lab-managed path.
robot.actuators.joint_command.position
robot.actuators.joint_command.velocity
robot.actuators.joint_command.effort

# Articulation-wide effort telemetry.
robot.actuators.computed_effort
robot.actuators.applied_effort

# Named model parameters and state.
motor = robot.actuators["legs"]       # explicit group
motor.stiffness
motor.damping
motor.actuator_effort_limit

drive = robot.actuators["gripper"]    # implicit group
drive.joint_effort_limit
```

All collection-wide arrays use articulation joint order. Native paths bypass
`joint_command`, so that view is not submitted-command telemetry for them.

### Joint properties and actuator properties

`actuator_effort_limit` is the explicit model clipping limit.
`joint_effort_limit` and `joint_velocity_limit` configure the joint or solver.
The deprecated `effort_limit` alias resolves to the actuator limit for explicit
models and to the joint limit for implicit models. `effort_limit_sim` and
`velocity_limit_sim` remain deprecated configuration aliases through 3.x.

Explicit groups retain their model gains, rated velocity, delay, motor curves,
and clipping limits. They do not retain copies of solver limits or friction.
Implicit groups instead read stiffness, damping, and effort limits from live
articulation buffers because the backend executes their drive.

The former group-level joint-property accessors (`effort_limit_sim`,
`velocity_limit_sim`, armature, and friction variants) are removed. Read those
values through `ArticulationData` and update them with articulation joint
writers. The legacy backend gain writers remain deprecated 3.x forwarders;
managed randomization should use `randomize_actuator_gains`.

## Logical groups and execution batching

Named groups remain the configuration and access surface. Internally:

- disjoint stateless groups of the same exact Lab actuator class may execute as
  one batch even when their parameters differ;
- implicit batches read live articulation-wide gains and limits and publish
  processed commands and telemetry in one fused Warp launch;
- ideal-PD and DC-motor batches use pointer-stable staging, in-place Torch
  compute, cached Warp gathers, and fused output publication;
- a singleton full-articulation explicit group keeps direct views and avoids a
  redundant gather;
- stateful, neural, subclass-specialized, native, and incompatible groups stay
  separate.

Native runtime parsing aggregates structurally compatible controllers while
keeping per-DOF values. Unsupported custom explicit configurations raise before
USD actuator state is changed. Two ambiguous construction states are rejected:

- collection membership is fixed after construction;
- a joint cannot belong to more than one actuator group.

## Backend behavior

- **PhysX:** runs Lab models in the collection before submission and supported
  native explicit actuators through the shared host adapter.
- **OVPhysX:** also runs Lab models in the collection; its native path uses the
  shared host adapter while preserving eager tensor binding and partial writes.
- **Newton:** runs Lab models in the collection before submission and
  manager-owned native controllers inside the solver.
- **Ordering:** collection buffers stay in public joint order; conversion occurs
  only at backend boundaries.

The host adapter captures staging, native model execution, and telemetry on
CUDA when possible and falls back to eager execution otherwise. It manages the
capture for stateful native actuators. Neural checkpoints use Isaac Lab's shared
file cache before Newton metadata is added, so local and remote paths follow the
same loading path.

## Compatibility and documentation

- Deprecated articulation command setters forward to `actuators.command`.
- Deprecated `ArticulationData` command and torque-telemetry accessors forward
  to the collection.
- Lab execution of explicit actuator models is deprecated in favor of supported
  Newton-native actuators.
- The actuator concept page now documents ownership, commands, limits,
  explicit/implicit behavior, batching, native execution, and backend-specific
  constraints.
- The 3.0 migration guide and tutorials use the new command and property names.

LEAPP action terms retain the annotated articulation setters until the exporter
supports collection setters.

## Final performance validation

Both revisions used fresh, isolated environments and the same dependency lock.
Run order was counterbalanced by row and seed. Throughput values below are mean
± sample standard deviation. Throughput changes are paired geometric FPS ratios
with two-sided 95% t intervals; p-values are Holm-adjusted across all 12 rows.

### Checkpoint playback: five paired seeds

Protocol: fixed policy checkpoints, 4,096 environments, 50 warm-up steps, and
1,000 measured steps under inference mode. This uses the full benchmark task
configuration rather than the reduced interactive play configuration. The
canonical result set contains 60 paired comparisons and 120 unique bundles.

| Task | Physics | Actuator path | `develop` FPS | PR FPS | Paired change (95% CI) | Holm p |
|---|---|---|---:|---:|---:|---:|
| Cartpole | PhysX | Lab | 1,029,140 ± 8,301 | 1,269,172 ± 7,097 | +23.33% [+21.83%, +24.84%] | <0.0001 |
| Cartpole | Newton | Lab | 1,505,687 ± 18,142 | 1,981,901 ± 17,827 | +31.63% [+28.28%, +35.07%] | 0.0001 |
| G1 Flat | PhysX | Lab | 108,564 ± 3,465 | 117,091 ± 3,872 | +7.85% [+3.53%, +12.35%] | 0.0478 |
| G1 Flat | Newton | Lab | 233,893 ± 1,211 | 250,100 ± 776 | +6.93% [+5.96%, +7.90%] | 0.0003 |
| ANYmal-D Flat | PhysX | Lab neural | 268,397 ± 6,016 | 269,470 ± 5,798 | +0.40% [-0.79%, +1.61%] | 1.0000 |
| ANYmal-D Flat | Newton | Lab neural | 430,949 ± 2,785 | 434,384 ± 5,860 | +0.79% [-0.80%, +2.41%] | 1.0000 |
| Franka Reach | PhysX | Lab | 315,996 ± 9,606 | 332,548 ± 29,851 | +4.95% [-4.74%, +15.64%] | 1.0000 |
| Franka Reach | Newton | Lab | 563,375 ± 8,702 | 635,729 ± 6,690 | +12.85% [+11.66%, +14.05%] | <0.0001 |
| Go2 Flat | PhysX | Lab | 266,480 ± 37,831 | 280,768 ± 14,739 | +6.05% [-10.82%, +26.12%] | 1.0000 |
| Go2 Flat | PhysX | Newton native | 297,084 ± 23,028 | 294,879 ± 29,591 | -0.91% [-20.53%, +23.56%] | 1.0000 |
| Go2 Flat | Newton | Lab | 813,321 ± 5,481 | 836,744 ± 2,688 | +2.88% [+1.73%, +4.04%] | 0.0172 |
| Go2 Flat | Newton | Newton native | 905,438 ± 5,792 | 924,958 ± 10,496 | +2.15% [+0.90%, +3.42%] | 0.0520 |

There is no supported playback regression. Six rows remain significant after
correction. The negative Go2 PhysX-native point estimate is small relative to
its seed variance and its interval spans large gains and losses.

#### Environment creation time (exploratory)

The playback harness also records the time spent in `gym.make()`, including
articulation construction. This is not total process startup: imports, task
configuration, app launch, runner and checkpoint loading, and the first step
are outside this timer. Persistent driver and Warp caches can also affect these
measurements, so the intervals below are unadjusted and should be treated as
follow-up evidence rather than a release gate. A positive change means the PR
was slower to create the environment.

| Task | Physics | Actuator path | `develop` seconds | PR seconds | Paired PR/develop change (95% CI) |
|---|---|---|---:|---:|---:|
| Cartpole | PhysX | Lab | 4.633 ± 0.036 | 4.670 ± 0.162 | +0.76% [-2.94%, +4.60%] |
| Cartpole | Newton | Lab | 5.271 ± 0.055 | 5.294 ± 0.045 | +0.44% [-1.12%, +2.01%] |
| G1 Flat | PhysX | Lab | 48.631 ± 0.269 | 48.605 ± 0.135 | -0.05% [-0.80%, +0.70%] |
| G1 Flat | Newton | Lab | 8.379 ± 2.552 | 7.956 ± 0.017 | -2.14% [-29.19%, +35.23%] |
| ANYmal-D Flat | PhysX | Lab neural | 21.974 ± 0.646 | 22.753 ± 1.897 | +3.31% [-6.73%, +14.42%] |
| ANYmal-D Flat | Newton | Lab neural | 6.813 ± 0.456 | 6.852 ± 0.076 | +0.73% [-5.79%, +7.71%] |
| Franka Reach | PhysX | Lab | 32.516 ± 0.097 | 33.241 ± 1.758 | +2.12% [-4.16%, +8.81%] |
| Franka Reach | Newton | Lab | 8.554 ± 0.297 | 8.981 ± 0.791 | +4.74% [-7.75%, +18.92%] |
| Go2 Flat | PhysX | Lab | 24.657 ± 0.356 | 24.509 ± 0.086 | -0.59% [-2.19%, +1.03%] |
| Go2 Flat | PhysX | Newton native | 27.424 ± 1.349 | 26.848 ± 0.084 | -2.01% [-7.47%, +3.77%] |
| Go2 Flat | Newton | Lab | 6.281 ± 0.044 | 6.506 ± 0.093 | +3.58% [+2.24%, +4.93%] |
| Go2 Flat | Newton | Newton native | 6.329 ± 0.026 | 6.561 ± 0.054 | +3.66% [+2.35%, +4.99%] |

Most rows are unresolved. The consistent signal worth a dedicated startup
benchmark is the roughly 0.23-second increase for Go2 on Newton, on both the Lab
and native actuator paths.

### Full training: three paired seeds

Protocol: RSL-RL, 4,096 environments, 50 timing warm-up steps, and each task's
full schedule: Cartpole 150, ANYmal-D and Go2 300, Franka 1,000, and G1 1,500
iterations. The canonical set contains 36 paired comparisons and 72 unique
successful bundles. Native neural runs are excluded because that path is known
not to work on either revision.

| Task | Physics | Actuator path | `develop` FPS | PR FPS | Paired change (95% CI) | Holm p |
|---|---|---|---:|---:|---:|---:|
| Cartpole | PhysX | Lab | 609,337 ± 1,779 | 680,158 ± 3,496 | +11.62% [+10.98%, +12.27%] | 0.0018 |
| Cartpole | Newton | Lab | 755,488 ± 11,886 | 846,398 ± 6,748 | +12.04% [+8.82%, +15.36%] | 0.0354 |
| G1 Flat | PhysX | Lab | 100,722 ± 938 | 104,069 ± 1,021 | +3.32% [-1.42%, +8.29%] | 0.5742 |
| G1 Flat | Newton | Lab | 184,909 ± 425 | 194,393 ± 975 | +5.13% [+4.35%, +5.91%] | 0.0131 |
| ANYmal-D Flat | PhysX | Lab neural | 205,431 ± 716 | 205,932 ± 2,199 | +0.24% [-2.45%, +3.00%] | 1.0000 |
| ANYmal-D Flat | Newton | Lab neural | 270,330 ± 1,242 | 273,531 ± 2,486 | +1.18% [-1.54%, +3.98%] | 1.0000 |
| Franka Reach | PhysX | Lab | 241,489 ± 175 | 257,090 ± 2,820 | +6.46% [+3.51%, +9.48%] | 0.0855 |
| Franka Reach | Newton | Lab | 376,112 ± 2,892 | 399,159 ± 8,936 | +6.11% [+2.22%, +10.15%] | 0.1455 |
| Go2 Flat | PhysX | Lab | 228,745 ± 5,309 | 229,636 ± 10,642 | +0.34% [-9.65%, +11.43%] | 1.0000 |
| Go2 Flat | PhysX | Newton native | 227,488 ± 10,813 | 232,861 ± 3,340 | +2.43% [-5.57%, +11.12%] | 1.0000 |
| Go2 Flat | Newton | Lab | 443,121 ± 3,497 | 448,127 ± 4,156 | +1.13% [+0.73%, +1.53%] | 0.0607 |
| Go2 Flat | Newton | Newton native | 458,881 ± 4,562 | 459,165 ± 1,376 | +0.06% [-1.75%, +1.91%] | 1.0000 |

No row has a negative mean training-throughput change. Cartpole on both
backends and G1 on Newton remain significant after correction.

Each run's late-training reward is averaged over its final 100 iterations. The
table reports the mean ± sample standard deviation across three seeds:

| Task | Physics | Actuator path | `develop` reward | PR reward |
|---|---|---|---:|---:|
| Cartpole | PhysX | Lab | 4.917 ± 0.035 | 4.917 ± 0.035 |
| Cartpole | Newton | Lab | 4.922 ± 0.016 | 4.922 ± 0.016 |
| G1 Flat | PhysX | Lab | 28.252 ± 0.358 | 27.237 ± 0.502 |
| G1 Flat | Newton | Lab | 22.992 ± 0.629 | 23.254 ± 0.722 |
| ANYmal-D Flat | PhysX | Lab neural | 15.083 ± 2.840 | 15.083 ± 2.840 |
| ANYmal-D Flat | Newton | Lab neural | 7.803 ± 0.088 | 12.673 ± 4.248 |
| Franka Reach | PhysX | Lab | 0.033 ± 0.098 | 0.033 ± 0.098 |
| Franka Reach | Newton | Lab | 0.014 ± 0.092 | -0.031 ± 0.047 |
| Go2 Flat | PhysX | Lab | 32.038 ± 1.152 | 32.038 ± 1.152 |
| Go2 Flat | PhysX | Newton native | 31.253 ± 1.027 | 31.547 ± 0.825 |
| Go2 Flat | Newton | Lab | 34.071 ± 0.979 | 34.223 ± 1.019 |
| Go2 Flat | Newton | Newton native | 34.604 ± 0.318 | 34.742 ± 0.538 |

Episode length and success were also checked over the final 100 iterations. No
convergence regression is statistically established.
G1 PhysX has a lower mean reward on the PR, but the paired difference is
unresolved (`p = 0.157`) and success is unchanged. ANYmal-D on Newton is
bimodal across seeds; its apparent improvement is also unresolved (`p = 0.182`).

Compared commits:

- `develop`: `135cf98904cbd73f3b8bdd89188e1a463ebcbb97`
- PR: `9ee6296124687ec54ed13e102ac59b7667b08d8d`

## Validation

- [x] Complete core actuator suite on CPU and CUDA: **479 passed**
- [x] Focused Newton target-mode, native-gain, and friction integration: **15 passed**
- [x] Focused PhysX native current-state ordering: **2 passed**
- [x] OVPhysX native command, mixed-path, reset, gain, and state-refresh coverage
- [x] Warning-as-error Sphinx build succeeded
- [x] All repository pre-commit hooks passed before the final push
- [x] Independent final code review found no remaining Critical or Important findings
- [x] Independent benchmark audit accepted all **192** canonical final bundles

Fused implicit telemetry can differ from the previous Torch expression by one
float32 ULP (`4.768e-7`) because of arithmetic ordering. Processed position,
velocity, and feed-forward effort commands sent to the backend remain exact.

## Type of change

- New feature: backend-neutral actuator collection and Newton-native execution
- Performance: automatic batching and fused/cached Warp staging
- Bug fix: joint/actuator ownership, class validation, ordering, and remote checkpoints
- Documentation and migration update

## Checklist

- [x] I have run pre-commit checks with `./isaaclab.sh -f`
- [x] I have updated the documentation
- [x] I have added focused behavioral tests and verified regression tests fail before their fixes


docs/source/_static/actuators/armature-clip.webp
kellyguo11, line 1: we've been trying to enforce keeping only .jpg media files in the repo itself to keep things lean. how strict do we want to be with that now? perhaps some of the larger files could still be hosted on s3?

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -230,17 +224,11 @@ def reset(self, env_ids: Sequence[int] | None = None, env_mask: wp.array | None
...
-            actuator.reset(env_ids)
-        # Reset Newton-actuator per-env states (delay queues, neural hidden state, etc.).
-        # The adapter is per-articulation on PhysX and is not part of ``self.actuators``.
-        # ``getattr`` guards subclasses (e.g. ``Multirotor``) that override
-        # ``_process_actuators_cfg`` and never initialize these attributes.
-        if getattr(self, "_has_newton_actuators", False) and getattr(self, "newton_actuator_adapter", None) is not None:
-            self.newton_actuator_adapter.reset(env_ids)
+        # use ellipses object to skip initial indices.
+        if (env_ids is None) or (env_ids == slice(None)):
+            env_ids = slice(None)
```
kellyguo11, line 229: Could we preserve `None` for a full actuator reset (or make the delayed buffers accept `slice(None)`) here? With the current normalization, `robot.reset()` passes `slice(None)` through `ActuatorCollection.reset()` into `DelayedPDActuator.reset()`, and `CircularBuffer.reset()` calls `len(batch_ids)`, raising `TypeError: object of type 'slice' has no len()`. This is currently breaking `scripts/demos/quadrupeds.py` on both the PhysX and Newton CI jobs; the same normalization exists in the Newton articulation.

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -6,17 +6,20 @@
...
+from typing import TYPE_CHECKING, ClassVar, Literal
 
 import torch
 
 from isaaclab.utils import DelayBuffer, LinearInterpolation
 from isaaclab.utils.types import ArticulationActions
 
-from .actuator_base import ActuatorBase
+from .actuator_base import ActuatorBase, _effort_limits_equal
+from .actuator_base_cfg import _resolve_effort_limit_aliases
```
kellyguo11: Could this helper be moved to a runtime-safe module or imported lazily? This eager import makes `isaaclab.actuators.actuator_base_cfg` load while importing environment factories, and `test_factories_are_kitless_in_fresh_process` now fails with that module listed as forbidden. Keeping config-only modules out of this import path is required for kitless factory discovery.

source/isaaclab_physx/isaaclab_physx/assets/articulation/actuator_control.py
```diff
@@ -0,0 +1,235 @@
...
+        return self._host_actuator_runtime.get_gain(attr, joint_ids)
+
+    def write_native_actuator_gain(
+        self,
+        attr: str,
+        values: torch.Tensor,
+        env_ids: torch.Tensor,
+        joint_ids: torch.Tensor,
+    ) -> None:
+        if self._host_actuator_runtime is not None:
```
kellyguo11: This currently makes the gain write silently do nothing when only `newton_actuator_adapter` is present. `test_physx_native_actuator_gain_update_uses_public_joint_ids` installs that adapter and expects `kp == [99, 2, 3]`, but `_host_actuator_runtime` is `None`, so the CI test leaves `kp` unchanged. Should the compatibility path still forward to the adapter, or should the fixture/test be migrated to install a host runtime? Either way, can we reconcile the contract rather than returning successfully without applying the update?

source/isaaclab_physx/test/assets/test_newton_actuators_physx.py
```diff
@@ -218,19 +259,23 @@ def _run_simulation(
...
             sim.step()
             articulation.update(DT)
             recorded_pos.append(wp.to_torch(articulation.data.joint_pos).clone())
             recorded_vel.append(wp.to_torch(articulation.data.joint_vel).clone())
-            recorded_computed.append(wp.to_torch(articulation.data.computed_torque).clone())
-            recorded_applied.append(wp.to_torch(articulation.data.applied_torque).clone())
+            recorded_computed_effort.append(articulation.actuators.computed_effort.torch.clone())
+            recorded_applied_effort.append(articulation.actuators.applied_effort.torch.clone())
             if use_newton_actuators:
                 recorded_adapter_computed.append(wp.to_torch(articulation.data._sim_bind_joint_computed_effort).clone())
```
kellyguo11: Can this test assert through `articulation.actuators.computed_effort` or an explicit host-runtime test hook instead of the dynamically attached `_sim_bind_joint_computed_effort` private attribute? The PhysX native-actuator equivalence cases currently error here because that attribute is not guaranteed to exist. We already record collection telemetry above; if the test needs the pre-publication native value, accessing it through the binding would make that dependency explicit.

docs/source/concepts/actuators.rst
```diff
@@ -0,0 +1,718 @@
...
+By default, Isaac Lab runs explicit actuator models once per step outside the solver, usually on
+Torch or Warp. This path is deprecated. Set
+:attr:`~isaaclab.sim.SimulationCfg.use_newton_actuators` to ``True`` to use the native path for
+supported explicit models:
+
+.. code-block:: python
+
+    from isaaclab.sim import SimulationCfg
+
+    sim_cfg = SimulationCfg(use_newton_actuators=True)
```
kellyguo11, line 641: Could we add or convert at least one runnable example that uses a supported explicit actuator with this flag? The only task configs I find enabling it are `franka_pour` and `ur10_particle_push`, and both robots use `ImplicitActuatorCfg`, so neither actually exercises the native explicit path. At the moment the tests appear to be the only executable demonstration of this new path.

kellyguo11 (commented):
A few implementation and CI concerns inline.

docs/source/_static/actuators/armature-clip.webp
ooctipus: I have the same concern, something else could be github lfs

source/isaaclab/isaaclab/sim/simulation_cfg.py
```diff
@@ -85,19 +85,16 @@ class SimulationCfg:
     """
 
     use_newton_actuators: bool = False
```
kellyguo11, line 87: Should this now default to `True`? The Isaac Lab execution path for explicit actuator models is deprecated in this PR, but leaving this as `False` means existing users remain on the deprecated path unless they discover and opt into this flag. If Newton-native execution is the supported path going forward, it should be the default and the examples should validate that default; configurations that cannot use it need an explicit compatibility story.

docs/source/migration/migrating_to_isaaclab_3-0.rst
```diff
@@ -20,6 +20,82 @@ This guide covers the main breaking changes and deprecations you need to address
...
+
+.. list-table:: Actuator limit migration
+   :header-rows: 1
+   :widths: 38 38 24
+
+   * - Deprecated configuration field
+     - Canonical configuration field
+     - Runtime owner
+   * - ``effort_limit`` (explicit)
+     - ``actuator_effort_limit``
```
ooctipus, line 40: why somewhere is using joint, other is using actuator?

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -377,9 +302,34 @@ def _clip_effort(self, effort: torch.Tensor) -> torch.Tensor:
...
+        .. deprecated:: 3.0
+            Use :attr:`actuator_effort_limit` instead. This alias will be removed in 4.0.
+        """
+        warnings.warn(
+            "ActuatorBase.effort_limit is deprecated. Use actuator_effort_limit instead; "
+            "effort_limit will be removed in 4.0.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.actuator_effort_limit
```
ooctipus, line 338: sometime its called actuator_effort_limit, sometime its called joint_effort_limit, and sometime its called effort_limit, can all use the same naming? perhaps just effort_limit

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -377,9 +302,34 @@ def _clip_effort(self, effort: torch.Tensor) -> torch.Tensor:
...
+
+    @effort_limit.setter
```
ooctipus, line 340: our articulation data api has been set_limit_index rather than setter.
I don't think we should adopt to a new syntax just for actuator. If setter is a better syntax, we can have one pr that apply this syntax globally

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
 from __future__ import annotations
 
+import warnings
 from dataclasses import MISSING
 
 from isaaclab.utils.configclass import configclass
+from isaaclab.utils.string import resolve_matching_names_values, string_to_callable
+
+
+def _resolve_actuator_class(class_type: type | str) -> type:
```
ooctipus: this looks wrong way to resolve things.

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
+        try:
+            class_type = string_to_callable(str(class_type))
+        except (AttributeError, ImportError, ValueError) as error:
+            raise ValueError(f"Unable to resolve actuator class '{class_type}'.") from error
+    if not isinstance(class_type, type) or not issubclass(class_type, ActuatorBase):
+        raise ValueError(f"Actuator class must derive from ActuatorBase, got {class_type!r}.")
+    return class_type
+
+
+def _is_implicit_actuator_cfg(cfg: ActuatorBaseCfg) -> bool:
```
ooctipus, line 13: this looks could be inlined

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
+from isaaclab.utils.string import resolve_matching_names_values, string_to_callable
+
+
+def _resolve_actuator_class(class_type: type | str) -> type:
+    """Resolve and validate an actuator class reference."""
+    from .actuator_base import ActuatorBase  # noqa: PLC0415
+
+    if isinstance(class_type, str):
+        try:
+            class_type = string_to_callable(str(class_type))
```
ooctipus: something feels wrong with all the resolve here, we might need to understand whats going on.
class_type,
func,
should all be auto resolving

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
+
+
+def _is_implicit_actuator_cfg(cfg: ActuatorBaseCfg) -> bool:
+    """Return whether an actuator configuration resolves to an implicit actuator class."""
+    from .actuator_pd import ImplicitActuator  # noqa: PLC0415
+
+    return issubclass(_resolve_actuator_class(cfg.class_type), ImplicitActuator)
+
+
+def _resolve_limit_values(value: dict[str, float | int] | float | int, joint_names: list[str]) -> tuple[float, ...]:
```
ooctipus: this code seems like should be processed in actuator_base not actuator_base_cfg

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -61,47 +124,41 @@ class ActuatorBaseCfg:
...
+
+        Use :attr:`joint_velocity_limit` to request a solver-level hard clamp. A physical
+        actuator limits joint speed through its torque curve rather than a kinematic clamp,
+        so the two limits are resolved independently. When only
+        :attr:`joint_velocity_limit` is set, it also serves as the joint velocity limit.
     """
 
-    effort_limit_sim: dict[str, float] | float | None = None
-    """Effort limit of the joints in the group applied to the simulation physics solver. Defaults to None.
+    joint_effort_limit: dict[str, float] | float | None = None
```
ooctipus, line 73: why it has joint prefix?

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -163,3 +220,13 @@ class ActuatorBaseCfg:
     viscous_friction: dict[str, float] | float | None = None
     """The viscous friction coefficient of the joints in the group. Defaults to None.
     """
+
+    actuator_effort_limit: dict[str, float] | float | None = None
```
ooctipus: why prefix becomes actuator?

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
+        if alias_value is None:
+            continue
+        if warn_deprecated:
+            warnings.warn(
+                f"Actuator group '{actuator_name}' uses deprecated '{alias_name}'. Use "
+                f"'{canonical_name}' instead; '{alias_name}' will be removed in 4.0.",
+                DeprecationWarning,
+                stacklevel=3,
+            )
+        canonical_value = getattr(cfg, canonical_name)
```
ooctipus: canonical value is a very AI naming and I don't know what it really mean, can we make a better name?

source/isaaclab/isaaclab/actuators/actuator_base_cfg.py
```diff
@@ -5,9 +5,87 @@
...
+    if isinstance(value, (float, int)):
+        return (float(value),) * len(joint_names)
+    joint_ids, _, values = resolve_matching_names_values(value, joint_names)
+    resolved_values = [0.0] * len(joint_names)
+    for joint_id, resolved_value in zip(joint_ids, values, strict=True):
+        resolved_values[joint_id] = float(resolved_value)
+    return tuple(resolved_values)
+
+
+def _resolve_effort_limit_aliases(
```
ooctipus: where resolve_effort_limit_alias is in this file but resolve resolve_velocity_limit_alias is in ActuatorCollection?

source/isaaclab/isaaclab/actuators/actuator_collection.py
```diff
@@ -0,0 +1,1073 @@
...
+        self._computed_effort = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._applied_effort = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._soft_joint_vel_limits = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._gear_ratio = wp.ones(shape, dtype=wp.float32, device=self.device)
+        self._all_joint_ids = wp.array(list(range(self.num_joints)), dtype=wp.int32, device=self.device)
+
+        self._joint_pos_target_ta = ProxyArray(self._joint_pos_target)
+        self._joint_vel_target_ta = ProxyArray(self._joint_vel_target)
+        self._joint_effort_target_ta = ProxyArray(self._joint_effort_target)
+        self._joint_pos_target_sim_ta = ProxyArray(self._joint_pos_target_sim)
```
ooctipus, line 265: _ta doesn make sense , what does it suppose to mean

source/isaaclab/isaaclab/actuators/actuator_collection.py
```diff
@@ -0,0 +1,1073 @@
...
+
+    # Construction and property resolution.
+
+    def _allocate_buffers(self) -> None:
+        """Allocate articulation-wide command and telemetry buffers."""
+        shape = (self.num_instances, self.num_joints)
+        self._joint_pos_target = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._joint_vel_target = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._joint_effort_target = wp.zeros(shape, dtype=wp.float32, device=self.device)
+        self._joint_pos_target_sim = wp.zeros(shape, dtype=wp.float32, device=self.device)
```
ooctipus, line 249: I am not sure `_joint_pos_target_sim` and `_joint_vel_target_sim` are necessary

target_pos + target_vel + feedforward effort + target_effort should be everything actuator holds,
if implicit actuator, only target_pos + target_vel matters, 
if explicit actuator then only target_pos + target_vel + feedforward effort + target_effort

source/isaaclab/isaaclab/actuators/actuator_collection.py
```diff
@@ -0,0 +1,1073 @@
...
+        self._all_joint_ids = wp.array(list(range(self.num_joints)), dtype=wp.int32, device=self.device)
+
+        self._joint_pos_target_ta = ProxyArray(self._joint_pos_target)
+        self._joint_vel_target_ta = ProxyArray(self._joint_vel_target)
+        self._joint_effort_target_ta = ProxyArray(self._joint_effort_target)
+        self._joint_pos_target_sim_ta = ProxyArray(self._joint_pos_target_sim)
+        self._joint_vel_target_sim_ta = ProxyArray(self._joint_vel_target_sim)
+        self._joint_effort_target_sim_ta = ProxyArray(self._joint_effort_target_sim)
+        self._computed_effort_ta = ProxyArray(self._computed_effort)
+        self._applied_effort_ta = ProxyArray(self._applied_effort)
```
ooctipus, line 269: if effort_target_sim represents effort going into simulator we don't need an extra _applied_effort

source/isaaclab_physx/isaaclab_physx/assets/articulation/actuator_control.py
```diff
@@ -0,0 +1,235 @@
...
+            if exc.name not in {"isaaclab_newton", "isaaclab_newton.actuators"}:
+                raise
+            logger.warning(
+                "use_newton_actuators is enabled but 'isaaclab_newton.actuators' is not available."
+                " Newton-native actuators will be disabled and the simulation will fall back to the"
+                " Isaac Lab actuator path. Install the isaaclab_newton extension to enable the fast path."
+            )
+            return set()
+
+        from isaaclab.sim.schemas.schemas_actuators import _validate_newton_native_actuator_cfgs  # noqa: PLC0415
```
ooctipus: these shouldn't be local import

source/isaaclab_physx/isaaclab_physx/assets/articulation/actuator_control.py
```diff
@@ -0,0 +1,235 @@
...
+        articulation.root_view.set_dof_actuation_forces(effort_target, articulation._ALL_INDICES)
+        if articulation._has_implicit_actuators:
+            articulation.root_view.set_dof_position_targets(pos_target, articulation._ALL_INDICES)
+            articulation.root_view.set_dof_velocity_targets(vel_target, articulation._ALL_INDICES)
+
+    def reset_native_actuators(self, env_ids: Sequence[int] | slice) -> None:
+        if self._native_actuator_path_active and self._host_actuator_runtime is not None:
+            self._host_actuator_runtime.reset(env_ids)
+
+    def get_native_actuator_gain(
```
ooctipus: what does native mean?

source/isaaclab_physx/isaaclab_physx/assets/articulation/actuator_control.py
```diff
@@ -0,0 +1,235 @@
...
+from isaaclab.assets.articulation import ordering_kernels
+from isaaclab.sim.utils.queries import find_first_matching_prim
+
+if TYPE_CHECKING:
+    from .articulation import Articulation
+
+logger = logging.getLogger(__name__)
+
+
+class PhysxActuatorControl(ArticulationActuatorControl):
```
ooctipus, line 30: it feels a bit weird to have an ArticulationActuatorControl adapter in articulation to support the actuator collection. If physx and newton implements the same ActuatorCollection the interface should be unified. I don't know why ArticulationActuatorControl is needed

most of the api are a shim that resolve ids, and handle lifecylce of backend, it could just be disolved into ActuatorCollection, and if you really want a class it could be inner class of ActuatorCollection

source/isaaclab_physx/isaaclab_physx/benchmark/assets/runtime.py
```diff
@@ -186,6 +182,14 @@ def create_test_articulation(
         articulation, "_cpu_body_inertia", wp.zeros((N, B, 9), dtype=wp.float32, device="cpu", pinned=True)
     )
 
+    from isaaclab.actuators import ActuatorCollection
```
ooctipus: does this needs to be local import?

ooctipus (changes requested):
Octi hand reviewed and write his comment:

This PR is in the right direction of managing the execution path when different actuator group can perhaps fused together rather than a for loop. And its also in the right direction of refactoring actuator concern away from articulation.

In my mental picture this pr has this data movement

```
Policy action
  |
  Action term: scale / offset
  |
  Articulation.set_joint_*_target() — compatibility forwarder
  |
  ActuatorCollection.command.set_*_index()
  |
  Write desired commands: position, velocity, effort
  |
  Articulation.write_data_to_sim()
  |------------------------------------------------changes are everything below----------------------------------------------------------
  ActuatorCollection.compute() — uses prebuilt execution batches
  ├─ Implicit: desired commands → unchanged processed commands + effort telemetry
  └─ Explicit: desired commands + current state → batch compute/clip → processed effort
   |
  ActuatorCollection.submit_commands()
   |
  Backend ActuatorControl
   ├─ Optional public-order → backend-order conversion
   └─ Submit position target, velocity target, and joint force
```

and we basically decoupled ActuatorCollection, ActuatorControl from Articulation.

However, the current code contains a lot of stuff and I think we can do a bit better to be closer to the simplistic data flow model above step by step, to list some of them:
- fuse actuator groups (Speed)
- refactor actuator collection (Decouple Actuator and Articulation)
- refactor ActuatonControl to manage the Lifecy cle and interface to articulation and backend  (Decouple Actuator and Backend Lifecycle)
- doc update

and I believe that if each one of them implemented correctly its most likely to be net negative LOC rather than net positive. 


There are also some naming, convention changes I would be more considerate, and parts that could be polished further.

- lab never uses setter decorator and has been always set_xxx_index. I am not against using it, but if its good,  we should apply it repo wide not just for actuator for this PR

- velocity_limit changed to joint_velocity_limit, and effort_limit changed to actuator_effort_limit. They seem unecessary, and even in the PR, sometime its velocity_limit sometime its joint_velocity_limit, sometime its actuator_xxxx

- some ai naming is also quite confusing to read: for example instead of saying `default`, it says `canonical`, instead of say `backend`, it says `native - write_native_actuator_gain`, perhaps it could just be `write_actuator_gain` without mentioning .

- there are also more than needed vairable allocated in actuator collections, for example, there is `joint_position_target`, and also `joint_position_target_sim`. I checked in the code, `joint_position/velocity_target_sim` is never needed. together with the deprecation forwarding like set_joint_*_target, I feel like looking at the same thing with many different naming conventions.

- cfg resolution seems also to be unpolished and is not using all the automatic resolve tool lab currently provides.

- There are some deprecation, but others straight up breaking. Since its going to be a breaking change anyway, I would favor straight up removal to simplify our code before GA release.


Overall, I support the execution-planning direction, but I think the PR currently combines too many ownership, lifecycle, naming, compatibility, and backend changes around that relatively focused performance improvement. Narrowing the
  interfaces and separating the migrations would make both the architecture and the performance result much easier to review and maintain.

ooctipus:
```
robot.actuators.command.position
robot.actuators.command.velocity
robot.actuators.command.effort

robot.actuators.joint_command.position
robot.actuators.joint_command.velocity
robot.actuators.joint_command.effort
```

robot.actuators.joint_command.position
robot.actuators.joint_command.velocity
^^
these two seems to be redundant, most of the control seems to only represent
target_position, target_velocity, feed_forward_effort, and applied effort,

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -36,75 +53,193 @@
...
-                " To set the effort limit, please use 'effort_limit_sim' instead."
+    is_implicit_model: ClassVar[bool] = True
+
+    stiffness: torch.Tensor
+    """Live articulation joint stiffness [N/m or N·m/rad, depending on joint type]."""
+
+    damping: torch.Tensor
+    """Live articulation joint damping [N·s/m or N·m·s/rad, depending on joint type]."""
+
+    class _JointDrive:
```
hujc7: **Agent suggestion:** `_JointDrive` stores `_joint_indices`, which duplicates `ActuatorBase.joint_indices` (actuator_base.py:177), and its three members are one-line projections. It looks like the class exists mainly as the bound/unbound sentinel the getters probe via `self.__dict__.get("_joint_drive")`.

Could it be inlined, with the control reference as the sentinel?

```python
@property
def stiffness(self) -> torch.Tensor:
    if self._control is None:
        return self._stiffness
    return self._control.joint_stiffness.torch[:, self.joint_indices]
```

Also: `_bind_joint_drive` has one caller, right after construction in `_build_groups`. If constructing `ImplicitActuator` outside a collection isn't supported, passing `control` into `__init__` removes the unbound phase entirely.

Naming, separately: the class is read-only — writes are rejected by `_set_joint_drive_property` — so "drive" promises a capability it refuses. `_JointDriveView`?

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -1388,22 +1388,38 @@ def __init__(self, cfg: EventTermCfg, env: ManagerBasedEnv):
         self.default_joint_stiffness = self.asset.data.joint_stiffness.torch.clone()
         self.default_joint_damping = self.asset.data.joint_damping.torch.clone()
 
-        # For explicit Lab actuators the sim-level stiffness/damping is zeroed out,
-        # so patch the defaults with the actual actuator PD gains.
-        for actuator in self.asset.actuators.values():
-            if not isinstance(actuator, ImplicitActuator):
-                joint_ids = actuator.joint_indices
+        self._gain_actuators = {
```
hujc7, line 1397: **Agent suggestion:** `randomize_actuator_gains` now owns two things that arguably belong to whoever owns the gains: the dispatch and the baseline.

It branches three ways on gain ownership — solver buffers for implicit, controller storage for native, actuator tensors for explicit Lab — twice over, once for stiffness and once for damping. It also holds the per-group default snapshots, cloning and re-overwriting them each call so randomization can't compound across resets.

The collection already knows group types and `_native_group_names`, and already owns one third of the dispatch via `write_actuator_parameter`. Could it own all three?

```python
self.asset.actuators.write_gain(
    actuator_name, "stiffness", stiffness[:, actuator_indices],
    env_ids=env_ids, joint_ids=group_columns,
)
```

If it owned the write, it could also own the baseline — storing default plus sampled offset and combining at write time, rather than clone-then-overwrite in the caller. That makes "always start from default" a property of the representation instead of discipline every new branch has to repeat. The offset would need to carry the operation, since `add`, `scale` and `abs` can't share a plain delta.

source/isaaclab/isaaclab/actuators/actuator_base.py
hujc7: **Agent suggestion:** Agreed, and there may be a sharper framing. `_resolve_joint_properties` (actuator_collection.py:469-476) resolves eight cfg fields and writes all eight to the solver: stiffness, damping, armature, friction, dynamic_friction, viscous_friction, joint_effort_limit, joint_velocity_limit. Six carry no prefix.

So `joint_` isn't marking "solver-owned" — it's disambiguating the two names that also exist as actuator-model fields. A collision fix that reads as a classification.

The PR already carries these eight as one payload internally: `ActuatorJointProperties` in actuator_control.py, passed to `write_resolved_joint_properties`. Notably that struct prefixes `joint_effort_limit` but leaves `velocity_limit` bare — inside a struct where neither prefix is needed, which suggests the naming is load-bearing nowhere.

If the config mirrored that grouping, no prefix would be needed at all:

```python
ImplicitActuatorCfg(
    joint_properties=JointPropertiesCfg(
        stiffness=..., damping=..., armature=...,
        effort_limit=200.0, velocity_limit=10.0,
    ),
    effort_limit=40.0,      # actuator model, keeps its 2.x name
    velocity_limit=8.0,
)
```

The alternative — unprefixing the solver side — looks unsafe: `_compat.py:64` maps deprecated `effort_limit` to `joint_effort_limit` for implicit but `actuator_effort_limit` for explicit, so reusing that name for the solver clamp would flip meaning for explicit actuators.

source/isaaclab/isaaclab/assets/articulation/base_articulation.py
```diff
@@ -1611,173 +1632,270 @@ def set_inertias_mask(
...
+        self,
+        writer_name: Literal["write_actuator_stiffness_to_sim", "write_actuator_damping_to_sim"],
+        gain_name: Literal["kp", "kd"],
+        values: torch.Tensor,
+        env_ids: torch.Tensor,
+        joint_ids: torch.Tensor,
+    ) -> None:
+        """Warn and forward a legacy native-controller gain write."""
+        warnings.warn(
+            f"{writer_name} is deprecated in 3.x and will be removed in 4.0. Use "
```
hujc7, line 1686: **Agent found bug:** This message points at an API this PR removes. `ActuatorBase.write_parameter` was deleted by 4bdc886 ("Replace the parameter machinery with Newton's selection API") — a repo-wide grep finds this string as the only surviving reference to it, so anyone following the warning hits `AttributeError`.

The replacement introduced by the same commit is `ActuatorCollection.write_actuator_parameter(name, component, attr, ...)`.

source/isaaclab/isaaclab/envs/mdp/events.py
```diff
@@ -1388,22 +1388,38 @@ def __init__(self, cfg: EventTermCfg, env: ManagerBasedEnv):
...
+            name: actuator
+            for name, actuator in self.asset.actuators.items()
+            if isinstance(actuator, (ImplicitActuator, IdealPDActuator))
```
hujc7: **Agent review:** This re-derives `ActuatorBase.is_implicit_model`, which already exists as a `ClassVar` and is what `actuator_base_cfg.py:19` uses to classify a config (`getattr(cfg.class_type, "is_implicit_model", False)`).

Two classifiers for the same property can drift: add a third PD family and the flag keeps working while this tuple silently omits it.

The name is also hard to read against the filter — `_gain_actuators` deliberately holds *both* implicit and explicit groups, so it means "groups that have PD gains", which neither "implicit" nor "explicit" conveys. Would a `collection.gain_groups()` helper work, classifying off the flag so the definition sits next to what defines it?

source/isaaclab/isaaclab/benchmark/stepping.py
```diff
@@ -258,17 +260,20 @@ def run_runtime_loop(env, num_steps: int, *, reset: bool = True) -> list[float]:
...
 
     step_times: list[float] = []
 
-    for _ in range(num_steps):
-        actions = sample_random_actions(env)
-        t0 = time.perf_counter_ns()
-        env.step(actions)
-        t1 = time.perf_counter_ns()
-        step_times.append((t1 - t0) / 1e9)
+    with torch.inference_mode():
```
hujc7, line 267: **Agent review:** `torch.inference_mode()` is added around the measured loop here (and at stepping.py:267 / :375) in the same PR that reports a 4.52% mean step-time improvement over `develop`.

If the `develop` arm was measured on the pre-PR harness, some of that delta is `inference_mode` rather than the refactor. Could you confirm which harness produced the baseline numbers — or split the benchmark change out so the comparison is harness-invariant?

docs/conf.py
```diff
@@ -209,6 +209,7 @@ def _read_pinned_versions() -> dict:
     "README.md",
     "licenses/*",
     "plans",
+    "superpowers",
```
hujc7: **Agent found bug:** `"superpowers"` looks like it was added unintentionally — no such path is tracked or present in this branch. (`"plans"` on the line above is pre-existing.)

The repo's `AGENTS.md` asks that generated plans, scratch files and agent artifacts not be committed.

source/isaaclab/isaaclab/actuators/actuator_control.py
```diff
@@ -0,0 +1,485 @@
...
+
+        Returns:
+            The Newton actuator selection produced by the backend's execution
+            setup (view, actuators, and joint ordering), or ``None`` when no
+            Newton actuators are active. The collection's parameter door
+            consumes this; controls perform no parameter access themselves.
+        """
+        return None
+
+    def compute_native_actuators(self, collection: ActuatorCollection, dt: float) -> bool:
```
hujc7, line 291: **Agent suggestion:** The return value here is always exactly `native_actuator_path_active`, which the caller already has:

- Newton: `return self._native_actuator_path_active`
- PhysX / OVPhysX: `if not self._native_actuator_path_active or self._actuator_runtime is None: return False`, else `True`

So the boolean carries no information beyond the existing property. The name also reads differently per backend — on Newton nothing is computed (the solver does it and a post-actuator callback publishes telemetry), while on PhysX/OVPhysX the runtime really does step the actuators. From the single call site in `ActuatorCollection.compute` you can't tell which you got.

Would branching on `native_actuator_path_active` and making this `step_native_actuators(...) -> None` be clearer? It also makes the three uses of that property consistent.

source/isaaclab/isaaclab/actuators/actuator_pd.py
AntoineRichard: This needs to go it's useless

AntoineRichard: Even the bind is useless because it only "binds" if the indices are contiguous which is misleading to say the least...

source/isaaclab/isaaclab/actuators/actuator_base.py
```diff
@@ -301,18 +311,6 @@ def compute(
...
-        table = self.joint_property_resolution_table[actuator_param]
-
-        ids = joint_ids if isinstance(joint_ids, torch.Tensor) else list(range(len(joint_names)))
-        for idx, name in enumerate(joint_names):
-            cfg_val_log = "Not Specified" if cfg_val is None else float(new_val[idx])
-            default_usd_val = usd_val if isinstance(usd_val, (float, int)) else float(usd_val[0][idx])
-            applied_val_log = default_usd_val if cfg_val is None else float(new_val[idx])
-            table.append([name, int(ids[idx]), default_usd_val, cfg_val_log, applied_val_log])
-
     def _parse_joint_parameter(
```
ooctipus: this shim looks removable

ooctipus (approved):
Thanks this looks much nicer now.
only left a comment about shim layer but otherwise looks like a great rework :D 
Congrat

docs/source/migration/migrating_from_omniisaacgymenvs.rst
```diff
@@ -539,18 +539,18 @@ Task config in Isaac Lab can be split into the main task configuration class and
 |     gpu_max_rigid_patch_count: 81920                            |   actuators={                                                   |
 |     gpu_found_lost_pairs_capacity: 1024                         |     "cart_actuator": ImplicitActuatorCfg(                       |
 |     gpu_found_lost_aggregate_pairs_capacity: 262144             |        joint_names_expr=["slider_to_cart"],                     |
-|     gpu_total_aggregate_pairs_capacity: 1024                    |        effort_limit=400.0,                                      |
```
ooctipus, line 542: this should be actuator_limit?

docs/source/policy_deployment/02_gear_assembly/gear_assembly_policy.rst
```diff
@@ -369,8 +369,8 @@ For the UR10e and Flexiv Rizon 4s deployments, we use an impedance controller in
                 "arm": ImplicitActuatorCfg(
                     joint_names_expr=["shoulder_pan_joint", "shoulder_lift_joint",
                                     "elbow_joint", "wrist_1_joint", "wrist_2_joint", "wrist_3_joint"],
-                    effort_limit=87.0,           # From UR10e specifications
```
ooctipus, line 372: would this be actuator_(effort|velocity)_limit?

docs/conf.py
ooctipus: what is super powers lol

PR #6840, opened 2026-08-01
[Deps] Bump USD Core to 26.5

AntoineRichard:
# Description

Raises the kit-less `usd-core` dependency from `>=25.11,<26.0` to
`>=26.5,<27.0` so Isaac Lab picks up the upstream fix for crashes while
parsing colliders concurrently.

Newton's pinned importer metadata still declares `usd-core<26.5`, so this PR
adds a narrow uv override for x86_64/AMD64 until that metadata catches up. The
lock resolves `usd-core 26.8` for the supported Linux x86_64 and Windows
platforms. This intentionally excludes the unrelated Newton runtime changes
from #6301.

## Type of change

- Breaking change (extensions built against USD 25 must be rebuilt against
  USD 26)

## Screenshots

Not applicable.

## Verification

- `./isaaclab.sh -p -m pytest source/isaaclab/test/cli/test_source_package_metadata.py`
  (`2 passed`)
- `./isaaclab.sh -f`
- Fresh uv resolution (`472 packages`, `usd-core 26.8`)

No new test was added; the existing hard-coded USD dependency guard was
updated for the new ABI range.

## Checklist

- [x] I have run the pre-commit checks with `./isaaclab.sh --format`
- [x] I have made the corresponding changelog update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ dependencies = [
     # OpenUSD (kit-less mode). Both packages vendor a complete ``pxr`` runtime, so
     # they must never co-install: usd-core supplies pxr where it has wheels (x86_64),
     # usd-exchange supplies it on aarch64 (where usd-core has no wheel).
-    "usd-core>=25.11,<26.0 ; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
+    "usd-core>=26.5,<27.0 ; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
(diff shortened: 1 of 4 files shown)


kellyguo11:
We tried to do this in https://github.com/isaac-sim/IsaacLab/pull/6521 but decided against it because it causes a mismatch in versions with kit

AntoineRichard:
> We tried to do this in #6521 but decided against it because it causes a mismatch in versions with kit

Not anymore since Kit moved in this version no? Wasn't that the whole point of migrating to this Kit version?

PR #6841, opened 2026-08-01
Match prim path expressions as whole-path regular expressions

ooctipus:
# Description

`find_matching_prims` matches each `/`-separated token against prim names at that depth.
`find_first_matching_prim` compiles the same argument as one regex over the full path. Both
take a parameter named `prim_path_regex` and document the identical contract. So one
expression on one stage gives two answers:

```
"/World/Robot/.*link2"

find_matching_prims      -> []
find_first_matching_prim -> /World/Robot/link1/link2
```

This makes both read the argument as a plain Python regular expression over the whole path,
and has `find_first_matching_prim` delegate to `find_matching_prims`.

Two follow-on changes are needed, because `.*` had been carrying structure rather than
meaning what regex says.

**The clone slot comes from the template, not from the text.** `make_clone_plan` derived an
asset's destination with `prim_path.replace(".*", "{}")`. `str.replace` is positionally blind,
so a second wildcard below the environment produces two slots:

```
/World/envs/env_.*/Robot/.*  ->  /World/envs/env_{}/Robot/{}  ->  .format(3)  ->  IndexError
```

`CloneCfg` now carries `clone_template`, and the plan splits the cfg path at the known
environment depth. A template always yields a regex; recovering a template from a regex means
guessing which part of the text is the wildcard.

**The environment slot is segment-safe.** With `.*` free to mean what regex says, the
namespace spells its slot `[^/]+`, so `{ENV_REGEX_NS}/Robot` selects `env_0/Robot` and not a
`Robot` nested deeper.

Configurations do not spell that out. `{ENV_REGEX_NS}` previously resolved only for assets an
`InteractiveScene` collected, so a direct environment had to write the namespace by hand. It
is now expanded where an asset or a sensor is built, and used across the task configurations —
no configuration names `[^/]+` itself.

## Search scope and startup cost

An unscoped `find_matching_prims(expr, stage)` is deliberately an honest full-stage query: it
applies `re.fullmatch` to every traversed prim and does not inspect the expression to infer a
root or depth limit. That keeps Python regex semantics intact, including expressions whose
classes, groups, or alternations can consume `/`.

Clone-aware discovery has a narrower candidate scope for an architectural reason rather than a
regex heuristic. It rebases the destination expression through the active `ClonePlan`, then
walks only the concrete source/prototype subtree supplied by that plan. It never scans every
cloned destination environment merely to rediscover identical source structure.

The previous startup table measured the now-removed regex-bounded implementation, so it was
removed rather than presented as evidence for the current head.

## Breaking change

An expression whose trailing token meant "direct children only" now means "whole subtree", and
should spell that token `[^/]+`. Across production prim-path literals: 659 hold no wildcard and
cannot change; none end in a bare `.*`; 6 hold a mid-pattern `.*`, and those change only if a
same-named prim exists deeper under the environment.

`CloneCfg.clone_regex` is removed rather than deprecated — its value is
`clone_template.format("[^/]+")`, so the two can no longer disagree. Worth a maintainer's call
whether that needs a shim.

Green tests are weak evidence for the classification above: the suite does not construct the
stage shapes on which the mid-pattern class changes meaning.

## Not in this change

`resolve_matching_prims_from_source` still has a legacy no-clone-plan branch that infers an
environment boundary heuristically. Ad-hoc multi-instance test scenes depend on that inference,
so replacing it is a separate decision. ClonePlan-owned expressions do not use this branch.

## Type of change


source/isaaclab/isaaclab/sim/utils/queries.py
```diff
@@ -314,43 +335,48 @@ def find_first_matching_prim(prim_path_regex: str, stage: Usd.Stage | None = Non
...
+    segments.append(path_expr[start:])
+    return segments
 
 
 def matches_path_expr_prefix(path_expr: str, prim_path: str) -> bool:
     """Return whether ``prim_path`` matches ``path_expr`` up to ``prim_path`` depth."""
-    prefix_expr = "/".join(path_expr.split("/")[: prim_path.count("/") + 1])
-    return re.match(f"^{_normalize_legacy_wildcard_pattern(prefix_expr)}$", prim_path) is not None
+    prefix_expr = "/".join(split_path_expr(path_expr)[: prim_path.count("/") + 1])
+    return re.match(f"^{prefix_expr}$", prim_path) is not None
```
hujc7: **Agent found bug:** `fullmatch` landed in `find_matching_prims` but not in its sibling.

`find_matching_prims` now compiles the bare expression and calls `pattern.fullmatch(...)`, which anchors the complete expression. This line still builds `f"^{prefix_expr}$"` and calls `re.match`, so a top-level `|` splits it into `^A` and `B$` -- the first matches by prefix, the second can never match:

```python
matches_path_expr_prefix("/World/envs/env_[^/]+/Robot/cart|pole", "/World/envs/env_0/Robot/cartXX")
# -> True
```

The two functions now read the same expression differently, which is the situation this PR set out to remove.

```suggestion
    return re.fullmatch(prefix_expr, prim_path) is not None
```

<details><summary>Related, at the caller rather than here</summary>

`envs/mdp/events.py:2566` and `:2757` build `"|".join(body_names)` straight into a path expression without grouping. Under `fullmatch` that pattern no longer selects `<body>/visuals` for any of the names -- it selects the bare body prim for the first name only. Grouping the join as `(?:{body_names_regex})` fixes it at the caller.

</details>

- [ ] `matches_path_expr_prefix` uses `fullmatch`; `events.py` groups its joined body names

source/isaaclab/isaaclab/sim/utils/queries.py
```diff
@@ -361,39 +387,29 @@ def find_matching_prims(prim_path_regex: str, stage: Usd.Stage | None = None) ->
...
-    for index, token in enumerate(tokens):
-        token_compiled = re.compile(token)
-        for prim in all_prims:
-            for child in prim.GetAllChildren():
-                if token_compiled.match(child.GetName()) is not None:
-                    output_prims.append(child)
-        if index < len(tokens) - 1:
-            all_prims = output_prims
-            output_prims = []
+    for prim in Usd.PrimRange(stage.GetPseudoRoot(), Usd.TraverseInstanceProxies(Usd.PrimAllPrimsPredicate)):
```
hujc7: **Agent review:** removing the bound was right for correctness, but it moves two things worth stating.

**Traversal predicate.** `develop` reached prims via `GetAllChildren()`, which stops at instance boundaries; this walks `TraverseInstanceProxies(Usd.PrimAllPrimsPredicate)`.

| | `develop` | this PR |
|---|---|---|
| `find_matching_prims` | stops at instance boundaries | descends into instance proxies |
| `find_first_matching_prim` | `stage.Traverse()` -- active, defined, loaded, concrete | all prims, including inactive / `over`-only / class |

Descending into instance proxies looks intended. The knock-on to `find_first_matching_prim` may not be: it can now return a deactivated or `over`-only prim where it previously returned `None`, changing the exists-check at `sensors/camera/camera.py:171` and the schema read at `sim/schemas/schemas_actuators.py:128`. Worth a line in `Changed`, and worth deciding whether that delegation should keep the default predicate.

**Cost.** Every call is now a full-stage walk with a regex per prim, as the updated docstring says. That is the honest implementation; the concern is only that the evidence for it is stale -- see the comment on the description.

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -228,7 +230,7 @@ def _collect_asset_cfgs(self) -> list[Any]:
             )
             for child in children:
                 if hasattr(child, "prim_path"):
-                    child.prim_path = child.prim_path.format(ENV_REGEX_NS=self.cloner_cfg.clone_regex)
+                    child.prim_path = child.prim_path.format(ENV_REGEX_NS=self._env_regex_ns)
```
hujc7: **Agent found bug:** the macro now has two expansions, and the one the scene uses is not brace-safe.

`expand_env_regex_ns` uses `str.replace` and documents why -- *"a plain replace, not str.format: the rest of the expression may hold braces of its own"* (`cloner/cloner_cfg.py:35`). This line still uses `str.format`, so a config that hits the documented hazard fails depending on who builds it:

```
"{ENV_REGEX_NS}/Robot/link_[0-9]{2}"

AssetBase / SensorBase   ->  /World/envs/env_[^/]+/Robot/link_[0-9]{2}
InteractiveScene         ->  IndexError: Replacement index 2 out of range
```

A brace quantifier is a normal spelling in the regex dialect this PR promotes, so the divergence is reachable rather than theoretical. Routing this line through `expand_env_regex_ns` also removes the second place that knows how the macro expands.

- [ ] `InteractiveScene` expands the macro through `expand_env_regex_ns`

source/isaaclab/test/sim/test_utils_queries.py
```diff
@@ -90,13 +94,50 @@ def test_get_first_matching_ancestor_prim():
...
+    sim_utils.create_prim("/World/Robot/foo")
+    sim_utils.create_prim("/World/Robot/foo/bar")
+    sim_utils.create_prim("/World/Robot/Arm")
+
+    matches = sim_utils.find_matching_prims(r"/World/Robot/[^A]+")
+
+    assert [prim.GetPath().pathString for prim in matches] == ["/World/Robot/foo", "/World/Robot/foo/bar"]
+
+
+def test_find_matching_prims_has_no_inferred_traversal_bounds():
```
hujc7, line 132: **Agent suggestion:** this one asserts the implementation rather than the contract.

The two behavioural tests added alongside it are exactly right. This one parses the function's own source and asserts that `GetPrimAtPath`, `PruneChildren` and `_bound_search` do not appear in it -- so it fails on any future rewrite that keeps the contract, including a legitimate reintroduction of bounding that is provably sound. `AGENTS.md` asks for observable behaviour and public contracts rather than implementation details.

The property worth pinning is behavioural and needs no introspection: a match must be returned no matter where it sits. `test_find_matching_prims_uses_unbounded_full_path_regex` already does that for `[^A]+`; a case whose match lies outside the expression's literal prefix -- say `find_matching_prims(r"/World/[^/]+/foo")` with `/World/A/foo` and `/World/B/foo` present -- closes the same gap without reading the source.

Also still open: `path_expr_to_glob`, `split_path_expr` and `expand_env_regex_ns` have no test mention anywhere on the branch, and the first two are pure string functions that need no stage.

hujc7 (commented):
**Agent review:** six comments, all on the matcher and its call sites; the clone-template and macro-expansion groups read sound and I have no comments on them.

> [!IMPORTANT]
> The **Startup cost** section no longer describes the code.

It says whole-path matching *"does not walk the whole stage: the expression bounds the search. Its longest literal prefix is the traversal root..."* -- but `312fe8cd` removed `_bound_search`, and the docstring on `find_matching_prims` now says the opposite: *"Every prim on the stage is tested; the expression does not imply a traversal root or depth limit."*

That also leaves the benchmark as the only performance evidence for a change whose cost profile it no longer measures: `35.39 s -> 35.80 s` on `env_creation` was taken with the bound in place, on the arm that avoided full-stage walks. Worth re-running `isaaclab benchmark startup --task Isaac-Velocity-Rough-AnymalD --num_envs 4096` against the current head and updating both the prose and the table -- a full-stage walk per call at 4096 envs is exactly the case the original table was there to rule out.

source/isaaclab/isaaclab/sim/utils/queries.py
```diff
@@ -23,6 +23,27 @@
 # import logger
 logger = logging.getLogger(__name__)
 
+_CHARACTER_CLASS = re.compile(r"\[\^?[^]]*\]")
+"""Matches a regex character class, whose text may hold a ``/`` that is not a path separator."""
+
+_SEGMENT_WILDCARD = re.compile(r"\[\^/\][*+]|\.\*")
```
hujc7, line 29: **Agent review:** which segment-wildcard spellings are canonical? Two places answer differently.

`_SEGMENT_WILDCARD` recognises `.*`, `[^/]*` and `[^/]+`. The namespace and `clone_template` emit `[^/]+`; `resolve_matching_prims_from_source`, `scene_data/deformable_discovery.py:329` and most converted task cfgs emit `[^/]*`, which also matches the empty segment. Picking one and using it everywhere would make the intended spelling obvious to the next author writing a cfg.

Related: `path_expr_to_glob`'s docstring says every regex spelling of a segment wildcard maps onto `*`, but `[^/]` and `[^/]{2}` pass through into the glob verbatim. Either narrow the docstring or widen the pattern.

source/isaaclab_ov/test/sim/test_views_xform_prim_ovphysx.py
```diff
@@ -192,7 +192,7 @@ def _build(num_envs: int, device: str) -> ViewBundle:
             prim.GetAttribute("xformOp:orient").Set(Gf.Quatd(1.0, 0.0, 0.0, 0.0))
 
         sim.reset()
-        view = OvPhysxFrameView("/World/envs/env_.*/Cube/CameraMount", device=device)
+        view = OvPhysxFrameView("/World/envs/env_[^/]*/Cube/CameraMount", device=device)
```
hujc7: **Agent review:** the conversion sweep left four call sites behind, including one in this file.

This line converted; `:77` in the same file did not, and both go to `find_matching_prims`:

```
:77   FrameView("/World/envs/env_.*/WorldCamera")
:195  OvPhysxFrameView("/World/envs/env_[^/]*/Cube/CameraMount")
```

The other three all pre-date the branch point, so they are sweep misses rather than merge fallout:

```
isaaclab/test/sim/check_meshes.py:139                 clone() root "/World/Origin.*/Object{idx:02d}"
isaaclab/test/cloner/test_replicate_session.py:55     prim_path="/World/envs/env_.*/Robot"
isaaclab_visualizers/test/test_newton_adapter.py:219  prim_path="/World/envs/env_.*/Camera"
```

Nothing fails today -- I traced each. The last two are `SimpleNamespace` fakes that never reach the matcher, and neither `:77` nor `check_meshes.py` has a same-named prim nested deeper. Raising it because the safety argument for the whole change is that every pattern still selects what it selected before, and these four are the ones where that has to be argued from stage shape rather than read off the expression.

source/isaaclab/isaaclab/scene/interactive_scene.py
ooctipus: Addressed in a4be8bd4f29: InteractiveScene routes every ENV_REGEX_NS expansion through expand_env_regex_ns, as does the multi-mesh ray caster. A regression test verifies that regex quantifiers such as {2} survive macro expansion.

source/isaaclab/isaaclab/sim/utils/queries.py
ooctipus: Addressed in a4be8bd4f29: the all-authored-prims policy is now explicit in the API docs and pinned by tests for inactive and undefined prims. Unscoped queries remain honest full-stage walks; ClonePlan-owned discovery now walks only its concrete source subtree. I also removed the stale bounded-implementation benchmark and claim from the PR description.

source/isaaclab_ov/test/sim/test_views_xform_prim_ovphysx.py
ooctipus: Addressed all four sweep misses in a4be8bd4f29, using the required non-empty single-segment spelling [^/]+ for environment and generated-origin slots.

source/isaaclab/isaaclab/sim/utils/queries.py
ooctipus: Addressed in a4be8bd4f29: framework-owned environment slots now consistently use [^/]+, while arbitrary user regex is left untouched. The path_expr_to_glob documentation is narrowed to its explicitly supported adapter spellings rather than claiming general regex-to-glob translation.

source/isaaclab/test/sim/test_utils_queries.py
ooctipus: Applied the behavioral suggestions in a4be8bd4f29: literal-prefix-independent matching plus tests for path_expr_to_glob, split_path_expr, and expand_env_regex_ns. I retained the structural negative gate because regex-derived root or depth inference is a rejected architecture here, then extended the gate to cover the shared subtree matcher as well.

source/isaaclab/isaaclab/sim/utils/queries.py
ooctipus: Addressed in a4be8bd4f29: matches_path_expr_prefix now uses fullmatch, and both events callers group the body-name alternation. The visual-layout probes also use clone-aware source resolution instead of scanning every destination clone.

hujc7 (approved):

StafaH (approved):

PR #6835, opened 2026-07-31
Stabilize dexsuite camera training defaults

ooctipus:
# Description

Camera-based dexsuite policies could not be trained. They collapsed into
joint-velocity-limit terminations (`abnormal_robot`) or stopped moving
altogether, plateauing at roughly the fraction of episodes that start with the
object already in the gripper, while the state-based policies on the same task
trained to high success.

The cause is the magnitude of the camera encoder's updates rather than anything
task-specific. Two things follow from that:

* The KL-adaptive learning-rate schedule oscillates across the point where
  encoder features churn faster than the policy head can track. Its *mean* rate
  is fine — matched-mean comparisons still fail — it is the excursions that do
  the damage, and its KL signal is itself inflated by the feature churn it is
  reacting to.
* The previous encoder produced a latent large enough that no stable learning
  rate existed for it at a useful speed.

This PR pairs a smaller shared encoder with a fixed learning rate for both the
`single_camera` and `duo_camera` presets, and makes the image observations
stationary so the encoder sees a consistent input scale (RGB and depth now share
a fixed `[-0.5, 0.5]` affine map instead of per-frame mean subtraction, which
also restores the absolute-distance anchor for depth). State presets are
unchanged and keep the adaptive schedule, which works well for them.

Measured on `Isaac-Lift-KukaAllegro-Camera` at 4096 envs, all four camera
preset combinations now reach ~94-96% success where they previously plateaued
near 10%:

| preset | success | `abnormal_robot` |
| --- | --- | --- |
| `single_camera`, `rgb64` | 95.7% | 11.9% |
| `single_camera`, `depth64` | 95.5% | 11.7% |
| `duo_camera`, `rgb64` | 95.4% | 10.4% |
| `duo_camera`, `depth64` | 94.2% | 14.4% |

The PR also lands the reset-state machinery the above relies on: a success
monitor that restarts episodes from the states the policy solves about half the
time, a diversity feature that spreads the reset bank over the grasp geometry,
and progress-based tracking rewards that pay only for new ground gained on the
episode-best error.

A second, independent commit fixes a backend-validation gap found while
verifying these presets across renderer/physics combinations: OvPhysX is kitless
and cannot share a process with Kit, but the guard only covered the Kit
visualizer, so pairing it with the Isaac RTX renderer fell through to a raw
`Failed to initialize Carbonite and load PhysX plugins` from inside the OvPhysX
library instead of a readable error. Happy to split this into its own PR if
preferred.

## Type of change

[...]

PR #6842, opened 2026-08-01
Reject OvPhysX physics paired with a Kit-based renderer

ooctipus:
# Description

OvPhysX is kitless and cannot share a process with Kit. The runtime guard covered the Kit
*visualizer*, but not Kit-based *renderers*, so `presets=ovphysx,isaacsim_rtx` with no visualizer
passed validation and then failed deep inside the OvPhysX library:

```
RuntimeError: ovphysx_create_instance() failed: Failed to initialize Carbonite and load PhysX plugins
```

The check now also treats a Kit camera as a Kit runtime signal, so the combination is rejected up
front with the supported alternatives, matching the existing OVRTX renderer check in the same
function.

Found while validating dexsuite camera presets across renderer/physics combinations. Split out of
#6835 so it can be reviewed and tested on its own.

## Type of change

## Checklist

`test_ovphysx_plus_kit_camera_without_visualizer_raises` covers the case the previous guard missed.
Both existing OvPhysX tests pass `visualizer="kit"`, so they passed before this change and never
exercised `scan.has_kit_camera`. Verified the new test fails when the condition is reverted to the
old form and passes with it: `23 passed` on
`source/isaaclab_tasks/test/core/test_runtime_compatibility.py`.

### DIFF
--- a/source/isaaclab/changelog.d/zhengyuz-ovphysx-kit-renderer-guard.rst
+++ b/source/isaaclab/changelog.d/zhengyuz-ovphysx-kit-renderer-guard.rst
@@ -1,7 +1,7 @@
 Fixed
 ^^^^^
 
-* Fixed OvPhysX physics paired with a Kit-based renderer failing inside the OvPhysX
-  library with ``Failed to initialize Carbonite and load PhysX plugins``. The kitless
-  backend guard only covered the Kit visualizer, so the combination is now rejected up
-  front with the supported alternatives, matching the existing OVRTX renderer check.
+* Fixed kitless OvPhysX physics and OVRTX renderers accepting some configurations
+  that also require Isaac Sim / Kit, then failing during runtime initialization.
+  Compatibility validation now reports every component that requires Kit and the
+  supported alternatives.
--- a/source/isaaclab/changelog.d/zhengyuz-ovphysx-kit-renderer-guard.rst
+++ b/source/isaaclab/changelog.d/zhengyuz-ovphysx-kit-renderer-guard.rst
@@ -1,0 +1,7 @@
+Fixed
+^^^^^
+
+* Fixed OvPhysX physics paired with a Kit-based renderer failing inside the OvPhysX
+  library with ``Failed to initialize Carbonite and load PhysX plugins``. The kitless
+  backend guard only covered the Kit visualizer, so the combination is now rejected up
+  front with the supported alternatives, matching the existing OVRTX renderer check.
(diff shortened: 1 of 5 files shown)


ErenAta16 (commented):
The fix is right and the error message is a real improvement — naming both the offending combination and the two supported alternatives is much better than the old "use a kitless visualizer" ending. The regression test also targets the actual gap (no visualizer, so the camera is the only Kit signal), which is the case a visualizer-shaped test would have missed.

Two structural things, both about the guard's shape rather than this change.

## `needs_kit` already means what the guard is hand-assembling

Line 248:

```python
config_scan.needs_kit = config_scan.has_kit_camera or config_scan.has_kit_physics or not has_physics
```

and the `Scan` docstring calls it "the headline launch decision after automatic selections are resolved".

The new guard builds its own version of that:

```python
if scan.has_ovphysx_physics and (has_kit_visualizer or scan.has_kit_camera):
```

which is `needs_kit` minus `has_kit_physics`, plus the visualizer. The missing term matters: `has_kit_physics` and `has_ovphysx_physics` are separate fields on `Scan`, so a config tree carrying both an `OvPhysxCfg` and a `PhysxCfg` sets both — and that combination reaches OvPhysX initialization with Kit already required, with no visualizer and no camera needed to trigger it.

More to the point, this is the same failure mode twice. The guard had one Kit term, a second Kit signal appeared, and the guard didn't grow with it — which is exactly what this PR is fixing. Expressing it against `needs_kit` (or a helper that is `needs_kit or has_kit_visualizer`) means the next Kit signal is picked up automatically instead of after someone hits `Failed to initialize Carbonite` again.

## The OVRTX check below has the identical omission

```python
if not scan.has_ovrtx or (not scan.has_kit_physics and not has_kit_visualizer):
    return
```

`has_kit_physics` and `has_kit_visualizer`, but not `has_kit_camera` — the term this PR just added to the branch above.

Whether that's reachable I can't confirm from outside, but the scan looks like it permits it: `has_ovrtx = True` and `has_kit_camera = True` are set in the same per-camera walk (lines 298 and 300), so a config with one OVRTX camera and one `IsaacRtxRendererCfg` camera sets both, and line 362 sets `has_ovrtx = True` unconditionally during RTX placeholder resolution. In that state, with no Kit physics and no Kit visualizer, the function returns early and the kitless OVRTX renderer shares a process with Kit — the same Carbonite conflict, from the other side.

If it isn't reachable, a comment saying why would be worth more than the check. If it is, it belongs in this PR: it's four lines below the one being fixed and the same one-line addition.

## Smaller

`kit_sources` reads well, and joining with `' and '` is right for two entries. If a third Kit signal is ever added, `' and '.join` on three items gives "a and b and c" — an Oxford-comma join would age better, but that's genuinely trivial.

The changelog entry and the split from #6835 both make this easy to review on its own, which I appreciated — the diff is small enough that I could check the surrounding function rather than just the changed lines.

kellyguo11 (approved):

PR #6851, opened 2026-08-02
Add OpenCV lens distortion rendering to Newton

liorbenhorin:
## Description

PR #6608 introduced renderer-agnostic OpenCV camera calibration configs and native RTX/OVRTX rendering. The Newton path was intentionally left as a documented no-op that emitted a warning and rendered using a centered, square-pixel pinhole projection.

This PR completes backend parity by making `NewtonWarpRenderer` consume the same OpenCV calibration and generate distorted per-pixel camera rays.

## This PR adds native OpenCV lens-distortion rendering to Newton:

- **OpenCV pinhole:** Adds a Warp kernel that inverts the OpenCV forward model per pixel using fixed-point iteration. It supports rational radial (`k1..k6`), tangential (`p1`, `p2`), and thin-prism (`s1..s4`) distortion - conritbuted by @AntoineRichard in https://github.com/liorbenhorin/IsaacLab/pull/1 
- **OpenCV fisheye:** Uses Newton’s native `compute_camera_rays_fisheye_opencv` helper rather than maintaining a duplicate Isaac Lab implementation.
- **Calibrated intrinsics:** Both paths honor `fx/fy/cx/cy`, including non-square focal lengths and an off-center principal point.
- **Disabled distortion:** `apply_lens_distortion=False` mutes the coefficients while retaining the calibrated intrinsics, matching RTX/OVRTX behavior.
- **Renderer behavior:** Replaces the previous Newton warning/no-op path with distortion-aware ray generation.
- **Tests:** Adds CPU unit tests for pinhole ray generation and CUDA integration tests for pinhole rendering, fisheye rendering, and intrinsic-matrix readback.

This is a stacked follow-up to #6608. It does not change the public camera configuration or USD authoring path introduced there; it only adds Newton renderer consumption of the existing model.

## Release backport

## Images
image222

## Checklist


source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -145,6 +146,9 @@ def __init__(
         else:
             self.clear_color = 0xFFEEEEEE
 
+        # OpenCV lens-distortion model (``spawn.distortion``), consumed by :meth:`_build_distortion_rays`
+        # to trace distorted per-pixel rays instead of the centered, square-pixel pinhole field.
+        self._distortion = getattr(spawn, "distortion", None)
```
StafaH: is getattr required?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -356,14 +360,86 @@ def update(self, positions: ProxyArray, orientations: ProxyArray, intrinsics: Pr
...
+        cfg = self._distortion
+        device = self.newton_sensor.model.device
+        image_width, image_height = float(cfg.image_size[0]), float(cfg.image_size[1])
+        # ``apply_lens_distortion=False`` keeps the intrinsics but mutes the distortion coefficients.
+        apply = bool(getattr(cfg, "apply_lens_distortion", True))
+
+        def _coeff(name: str) -> float:
+            return float(getattr(cfg, name, 0.0)) if apply else 0.0
+
+        if cfg.model == "opencvFisheye":
```
StafaH, line 393: if the newton utility exists, why do we need to also define our own kernel? If there are limitations in the newton kernel, I would suggest sending a PR there instead of building a seperate kernel here.

source/isaaclab_newton/changelog.d/opencv-lens-distortion-cameras.rst
```diff
@@ -0,0 +1,11 @@
...
+^^^^^
+
+* Added OpenCV lens-distortion support to the Newton renderer: a camera cfg carrying an OpenCV
+  pinhole (``k1..k6``, ``p1``, ``p2``, ``s1..s4``) or fisheye (``k1..k4``) distortion model on
+  ``spawn.distortion`` is now rendered through the distortion instead of as a centered, square-pixel
+  pinhole. Fisheye cameras use Newton's native OpenCV fisheye ray helper, while pinhole cameras use
+  an Isaac Lab kernel supporting rational radial, tangential, and thin-prism distortion. Both paths
+  honor the calibrated ``fx/fy/cx/cy`` intrinsics (including non-square focal lengths and an
+  off-center principal point). With ``apply_lens_distortion=False`` the distortion coefficients are
+  muted while the intrinsics are still applied, matching the RTX/OVRTX behavior.
```
AntoineRichard: **[AI-generated] Changelog — keep this user-facing**

This repeats the implementation and PR description (kernel/helper choices and the coefficient inventory). The changelog guidelines explicitly exclude internal details. Condense it to the behavior, for example: `Added OpenCV pinhole and fisheye lens-distortion rendering to the Newton backend, including calibrated intrinsics and coefficient muting through apply_lens_distortion.`

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -356,14 +360,86 @@ def update(self, positions: ProxyArray, orientations: ProxyArray, intrinsics: Pr
...
+                float(cfg.cy),
+                image_width=image_width,
+                image_height=image_height,
+                k1=_coeff("k1"),
+                k2=_coeff("k2"),
+                k3=_coeff("k3"),
+                k4=_coeff("k4"),
+                # Match the PR kernel's forward-facing camera hemisphere and avoid validating the
+                # OpenCV polynomial outside its physically meaningful calibration range.
+                max_fov=math.pi,
```
AntoineRichard: **[AI-generated] Comment — make it functional**

`PR kernel` is review-history language, and `max_fov` caps the accepted ray angle; it does not validate the polynomial. Replace both lines with the durable statement `# Limit fisheye rays to the forward hemisphere.`

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -473,20 +549,12 @@ def prepare_cameras(self, stage: Any, spec: CameraRenderSpec) -> None:
 
         Also captures the USD ``stage`` so the segmentation mapper can read the scene's
         :class:`UsdSemantics.LabelsAPI` labels when a segmentation output is requested.
+
+        OpenCV lens distortion (``spawn.distortion``) needs no preparation here: it is consumed at
+        ray-generation time by :meth:`RenderData._build_distortion_rays`, which inverts the OpenCV
+        forward model per pixel to trace the distorted camera-space rays.
```
AntoineRichard: **[AI-generated] Documentation — update the source of truth**

This adds renderer-local implementation prose while the public docstrings for `OpenCvDistortionCfg` and `PinholeCameraCfg.distortion` in `sensors_cfg.py` still state that Newton does not apply the model. Remove this duplicated paragraph and update those public docstrings to describe Newton support; otherwise the documented API contradicts the implementation.

source/isaaclab_newton/isaaclab_newton/renderers/opencv_distortion_rays.py
```diff
@@ -0,0 +1,114 @@
...
+Newton uses the OpenGL camera convention (``+X`` right, ``+Y`` up, looking down ``-Z``).
+
+To honor an OpenCV ``fx/fy/cx/cy`` + distortion-coefficient calibration, for each output pixel the
+kernel below inverts the OpenCV forward distortion model to recover the *undistorted* normalized
+image coordinates ``(x_u, y_u)`` (OpenCV image ``y`` points down), then emit the camera-space ray
+``normalize(vec3(x_u, -y_u, -1))`` -- the negation of ``y`` and ``z`` maps OpenCV camera space
+(``+Z`` forward, ``+Y`` down) onto Newton's OpenGL camera space.
+
+OpenCV fisheye ray generation is provided directly by Newton's
+``SensorTiledCamera.utils.compute_camera_rays_fisheye_opencv`` helper.
```
AntoineRichard: **[AI-generated] Documentation — remove duplication**

Keep the module docstring to a one-sentence scope. Ray layout and coordinate conversion are repeated in the kernel documentation/body, while fisheye delegation is not implemented in this module. These copies add maintenance surface without a second contract.

source/isaaclab_newton/isaaclab_newton/renderers/opencv_distortion_rays.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+
+from __future__ import annotations
+
+import warp as wp
+
+# Number of fixed-point iterations used to invert the distortion model. OpenCV's own
+# ``undistortPoints`` defaults to a similar iteration budget; the inversion converges well within
+# this for realistic calibrations.
+_INVERSION_ITERATIONS = 20
```
AntoineRichard: **[AI-generated] Comment — correct the iteration claim**

OpenCV documents the default `undistortPoints` path as 5 iterations, so 20 is not a similar budget, and this PR has no convergence test supporting “well within.” Keep the comment functional (what the constant controls), or add evidence for the selected bound and failure behavior.

source/isaaclab_newton/test/renderers/test_opencv_distortion_rays.py
```diff
@@ -0,0 +1,165 @@
...
+def test_pinhole_off_center_principal_point_is_honored():
+    """The ray through the principal-point pixel looks straight ahead (down -Z)."""
+    rays = _launch_pinhole({k: 0.0 for k in _PINHOLE_COEFFS})
+    directions = rays[0, :, :, 1, :]
+    # pixel closest to the principal point on the render grid
+    px = int(round(_CALIB["cx"] / _IMAGE_W * WIDTH - 0.5))
+    py = int(round(_CALIB["cy"] / _IMAGE_H * HEIGHT - 0.5))
+    direction = directions[py, px]
+    assert direction[0] == pytest.approx(0.0, abs=2e-2)
+    assert direction[1] == pytest.approx(0.0, abs=2e-2)
```
AntoineRichard: **[AI-generated] Test — remove redundant weak coverage**

`test_pinhole_zero_coeffs_matches_ideal_projection` already checks sampled rays against the full off-center calibration. This test adds no independent oracle, and its `2e-2` tolerance cannot detect sub-pixel principal-point errors. Remove it, or use a calibration whose principal point lands exactly on a sampled pixel and assert the expected ray tightly.

source/isaaclab_newton/test/sensors/test_camera_opencv_distortion_newton.py
```diff
@@ -0,0 +1,247 @@
...
+WIDTH, HEIGHT = 640, 480
+WARMUP_STEPS = 4
+
+# Example real-world OpenCV pinhole calibration (fx != fy, off-center principal point).
+_CALIB = dict(fx=339.26592887, fy=338.82010626, cx=323.55809091, cy=250.27360914)
+_COEFFS = dict(k1=0.07702322, k2=-0.13605453, k3=0.05163219, p1=-0.00024938, p2=-0.00175006)
+# scale the (mild) real coefficients so the barrel effect is unambiguous in the assertion
+_K_SCALE = 15.0
+# OpenCV fisheye (equidistant) coefficients; the base fisheye projection alone differs strongly from pinhole
+_FISHEYE_COEFFS = dict(k1=0.1, k2=-0.05, k3=0.0, k4=0.0)
```
AntoineRichard: **[AI-generated] Important — use an invertible calibration fixture**

Multiplying these radial coefficients by 15 makes the forward radial mapping non-monotonic inside this frame: its derivative first becomes negative near normalized radius 0.899, while the corners reach about 1.2. The inverse is therefore ambiguous, yet the test only requires a different image, so invalid rays can pass. Use a synthetic coefficient set that stays monotonic over the full frame and assert representative rays or render samples against an oracle.

source/isaaclab_newton/test/sensors/test_camera_opencv_distortion_newton.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+        anchor = RigidObjectCfg(
+            prim_path="{ENV_REGEX_NS}/Anchor",
+            spawn=sim_utils.CuboidCfg(
+                size=(0.01, 0.01, 0.01),
+                rigid_props=sim_utils.RigidBodyPropertiesCfg(),
+                mass_props=sim_utils.MassPropertiesCfg(mass=0.001),
+                collision_props=sim_utils.CollisionPropertiesCfg(),
+                physics_material=sim_utils.RigidBodyMaterialCfg(),
+                visual_material=sim_utils.PreviewSurfaceCfg(diffuse_color=(0.0, 0.0, 0.0)),
```
AntoineRichard: **[AI-generated] Test hygiene — do not add deprecated APIs**

These `RigidBodyPropertiesCfg`, `CollisionPropertiesCfg`, and `RigidBodyMaterialCfg` constructors emit three deprecation warnings during collection. New tests should use the solver-common `RigidBodyBaseCfg`, `CollisionBaseCfg`, and `RigidBodyMaterialBaseCfg` equivalents (or omit properties the off-screen anchor does not need).

source/isaaclab_newton/test/sensors/test_camera_opencv_distortion_newton.py
```diff
@@ -0,0 +1,247 @@
...
+    distorted, _ = _render_distance(_pinhole_distortion(True), device=device)
+    reference, _ = _render_distance(_pinhole_distortion(False), device=device)
+
+    assert distorted.shape == (HEIGHT, WIDTH, 1)
+    # both frames render geometry (the ground plane fills the frame)
+    assert np.isfinite(distorted).mean() > 0.9
+    assert np.isfinite(reference).mean() > 0.9
+    # the renderer applied the lens distortion: the distance maps warp well beyond render noise
+    mean_abs_diff = _mean_abs_distance_diff(distorted, reference)
+    assert mean_abs_diff > 0.05, f"distorted vs reference distance maps differ by only {mean_abs_diff:.4f} m"
```
AntoineRichard: **[AI-generated] Comments — remove assertion narration**

The two comments restate the assertions immediately below them. Remove them and let the test/helper names carry the contract; apply the same cleanup to the repeated comments in the fisheye test.

source/isaaclab_newton/test/sensors/test_camera_opencv_distortion_newton.py
```diff
@@ -0,0 +1,247 @@
...
+    """The Newton camera reports intrinsics matching the authored, non-square, off-center calibration."""
+    _distance, k = _render_distance(_pinhole_distortion(True), device=device)
+
+    assert k[0, 0] == pytest.approx(_CALIB["fx"], abs=1e-2)
+    assert k[1, 1] == pytest.approx(_CALIB["fy"], abs=1e-2)
+    assert k[0, 2] == pytest.approx(_CALIB["cx"], abs=1e-2)
+    assert k[1, 2] == pytest.approx(_CALIB["cy"], abs=1e-2)
+    # not the stock fx == fy / centered-principal-point collapse
+    assert k[0, 0] != k[1, 1]
+    assert k[0, 2] != pytest.approx(WIDTH / 2)
```
AntoineRichard: **[AI-generated] Test — remove unrelated duplicate integration coverage**

The Newton renderer does not compute `Camera.data.intrinsic_matrices`; `Camera._update_intrinsic_matrices` does. `source/isaaclab/test/sensors/test_opencv_distortion.py::test_readback_uses_authored_fx_fy_cx_cy` already checks these same values without a CUDA scene. Remove this test and simplify `_render_distance` to return only data needed to verify the changed renderer path.

AntoineRichard (dismissed):
> **AI-generated review (Codex).**

Requesting changes for test validity, stale public documentation, and code hygiene. The renderer path is directionally sound, but the integration fixture is non-invertible over the tested frame, the new integration test emits deprecation warnings, and the public camera configuration still says Newton ignores this model. I also found avoidable documentation/test duplication and comments that narrate review history or unsupported rationale.

I did not repeat the two existing inline threads about full coefficient coverage and pixel-center mapping; both remain unresolved. Before changing the latter, reconcile it with Newton's native fisheye helper, which currently uses the same half-pixel mapping.

Verification at this SHA: `./isaaclab.sh -f` passed in an isolated checkout; all 4 CPU ray tests passed; the 3 CUDA integration tests collected and skipped because CUDA is unavailable here. The PR's `isaaclab_newton` CI job passed; a separate `isaaclab_tasks [3/3]` job is failing.

liorbenhorin:
Newton provides an OpenCV fisheye helper, but no equivalent pinhole helper supporting rational, tangential, and thin-prism coefficients. Can we keep the [local OpenCV pinhole kernel](https://github.com/liorbenhorin/IsaacLab/blob/d68aae706/source/isaaclab_newton/isaaclab_newton/renderers/opencv_distortion_rays.py#L17-L98) as a temporary measure until both projections are supported upstream?

Also, I retained the +0.5 mapping because it matches Newton’s fisheye convention, as [documented alongside the pixel mapping](https://github.com/liorbenhorin/IsaacLab/blob/d68aae706/source/isaaclab_newton/isaaclab_newton/renderers/opencv_distortion_rays.py#L74-L78) - Changing only the local pinhole path would make the two projections inconsistent.

Other comments have been resolved.
@AntoineRichard @StafaH

AntoineRichard:
@jcarius-nv I think we'd like to avoid taking these Kernels into lab, even temporarily. Could you make sure Newton is ready for this? @daniela-hase

jcarius-nv:
@liorbenhorin , @daniela-hase , do you already have a corresponding Newton PR open for review?

liorbenhorin:
@daniela-hase would you like me to make this PR for Newton or will you handle it?

daniela-hase:
> Newton provides an OpenCV fisheye helper, but no equivalent pinhole helper supporting rational, tangential, and thin-prism coefficients. Can we keep the [local OpenCV pinhole kernel](https://github.com/liorbenhorin/IsaacLab/blob/d68aae706/source/isaaclab_newton/isaaclab_newton/renderers/opencv_distortion_rays.py#L17-L98) as a temporary measure until both projections are supported upstream?

I am no lens expert, so this may be a silly question: Is this a common enough lens that we should add a helper function for it upstream in Newton? The idea of the helper functions in Newton was to provide helpers for the most commonly used lenses - but the goal was not to have a helper for literally every lens in the world live in there I think :)

If someone has a fancy custom lens, the idea was that they can just compute the rays on their end.

But, if this is a common enough lens, then of course - fine by me to add it to Newton. :)

> @daniela-hase would you like me to make this PR for Newton or will you handle it?

If you could make the PR, that'd be great! Thank you :)
(assuming the answer to the above question is "Yes, this is a common model we want to have a helper for in Newton)

liorbenhorin:
@daniela-hase Yes, this is common for an upstream helper. It is OpenCV’s standard pinhole calibration family rather than a custom lens model:

[cv::calibrateCamera](https://docs.opencv.org/4.13.0/d9/d0c/group__calib3d.html) estimates this pinhole model’s intrinsics and radial/tangential distortion coefficients. The rational (k4–k6) and thin-prism (s1–s4) terms are optional extensions of the same model.
[ROS sensor_msgs/CameraInfo](https://docs.ros.org/en/ros2_packages/humble/api/sensor_msgs/msg/CameraInfo.html) describes the corresponding radial-tangential plumb_bob model as sufficient for most cameras.
This would remain separate from the existing upstream OpenCV fisheye helper. 
Pinhole is commonly used for conventional lenses, while the fisheye model is typically selected for very wide-angle lenses.
A camera/lens assembly can potentially be calibrated against either model, but users must use the helper matching the calibration model that produced their intrinsics and coefficients... the two coefficient sets are not interchangeable.

One OpenCV pinhole helper can cover the common 5-coefficient form and the optional rational/thin-prism extensions by leaving unused terms at zero.

I'll make the PR

liorbenhorin:
[Add OpenCV pinhole camera rays](https://github.com/newton-physics/newton/pull/3901) - @jcarius-nv @daniela-hase

kellyguo11:
run-ci

kellyguo11:
run-ci

kellyguo11 (commented):
do we need any updates for this PR now that the newton-side PRs landed?

liorbenhorin:
@kellyguo11 updated and ready

kellyguo11 (approved):

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -357,14 +361,74 @@ def update(self, positions: ProxyArray, orientations: ProxyArray, intrinsics: Pr
...
+
+        Uses Newton's native OpenCV pinhole and fisheye ray helpers. Both paths honor calibrated
+        ``fx/fy/cx/cy`` (non-square, off-center) intrinsics. When
+        :attr:`OpenCvDistortionCfg.apply_lens_distortion` is ``False``, the coefficients are treated
+        as zero while the calibrated intrinsics remain active, matching the RTX/OVRTX behavior.
+        """
+        cfg = self._distortion
+        image_width, image_height = float(cfg.image_size[0]), float(cfg.image_size[1])
+        # ``apply_lens_distortion=False`` keeps the intrinsics but mutes the distortion coefficients.
+        apply = bool(getattr(cfg, "apply_lens_distortion", True))
```
StafaH: I prefer not using getattr where we can avoid, is there a reason apply_lens_distortion would not be part of the config?

source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
```diff
@@ -357,14 +361,74 @@ def update(self, positions: ProxyArray, orientations: ProxyArray, intrinsics: Pr
...
+
+        def _coeff(name: str) -> float:
+            return float(getattr(cfg, name, 0.0)) if apply else 0.0
```
StafaH: Same comment as above, is there a reason _coeff needs to use getattr like this?

source/isaaclab_newton/test/sensors/test_camera_opencv_distortion_newton.py
```diff
@@ -0,0 +1,257 @@
...
+from __future__ import annotations
+
+import importlib.util
+
+import numpy as np
+import pytest
+
+pytestmark = [pytest.mark.integration, pytest.mark.rendering]
+
+_REQUIRED_MODULES = ("isaaclab_newton", "newton", "warp", "torch")
```
StafaH: This test file looks really weird. Why is it set-up in this way? Can we simplify this down. newton and warp should always be there, and we shouldn't need to check if a cuda device is available. I think the whole test file can probably be stripped down to a more basic simple test.

StafaH (approved):

kellyguo11:
run-ci

kellyguo11:
looks like there's a filing test that could be relevant - source/isaaclab/test/renderers/test_camera_output_contract.py

StafaH:
run-ci

kellyguo11:
run-ci

StafaH:
run-ci

ooctipus (approved):

PR #6862, opened 2026-08-03
Complete the L40S runner stability qualification

Neil4561:
## Summary

The [previous qualification run](https://github.com/isaac-sim/IsaacLab/actions/runs/30377922938) collected 57 of 135 samples and never reached a verdict. This finishes it, scoped to the buckets that actually failed.

### Why it fell short

Two unrelated faults, both confirmed from the run logs:

1. **All five Newton-touching buckets crashed** on `from newton.solvers import SolverNotifyFlags`. Staging source was 145 commits behind `develop` while the image's Newton was newer, and that symbol had already been replaced by `ModelFlags`. That accounts for 75 of the 78 missing samples.
2. **One allocation hit a `PermissionError`** on `/tmp/jit-cache/warp/...`. All five parallel allocations shared a single Warp JIT cache directory and raced on creating it. That accounts for the remaining 3.

### What changed

- **Caught staging up to `develop`.** The merged source now uses `ModelFlags` and imports only `SolverBase, SolverKamino`, which is the API the image already provides.
- **Isolated the JIT cache per allocation**, keyed on run id, attempt, and run label.
- **Scoped the run to the Newton buckets**, 75 samples instead of 135. The four PhysX-only buckets already produced clean evidence and re-measuring them would double the pool time for no new information. Both the seeder and the qualifier take the same backend allowlist, so the expected scope matches what is collected; a test keeps the two lists in sync and an unknown backend key is rejected rather than silently narrowing the scope.
- **Benchmark execution health is always enforced.** A crash could previously exit zero while FPS regressions were advisory.
- **Added a stale-image guard** so that hardening cannot fail PRs for reasons outside their control. When a crash log shows the image missing a symbol the source pins, the gate reports a stale CI image and stays advisory for those tasks. Only packages the image installs (`newton`, `warp`, `isaacsim`, ...) qualify; a missing Isaac Lab symbol is still a real defect and still fails.
- **Qualification no longer writes baselines** or chains a second gate run. It measures the pool and reports a verdict.

### Scope note

The verdict will cover the five Newton buckets. The four PhysX buckets are not re-qualified at this commit, since their July 28 samples carry the old staging SHA and cannot be mixed with new ones. Runner-to-runner spread is largely a hardware property and the Newton buckets sample the same five runners, so this still answers the pool-stability question.

The gate remains in advisory mode.

## Test plan

- [ ] Merge triggers `Performance Smoke - L40S Runner Stability`, which waits for this commit's `Performance Smoke Test` to pass and for the L40 pool to go quiet.
- [ ] All five allocations complete with no `PermissionError` in the JIT cache.
- [ ] 75 of 75 samples collected (5 allocations x 3 samples x 5 Newton buckets).
- [ ] The qualification report reaches a verdict instead of `INCONCLUSIVE` from missing evidence.
- [ ] 104 unit tests pass and `pre-commit` is clean (verified locally).


Neil4561:
Closing: the develop catch-up merge made this PR show 220 commits / 2024 files against the stale staging base. Reopening as a focused PR containing only the stability changes.

PR #6863, opened 2026-08-03
Fix and rescope the L40S runner stability qualification

Neil4561:
## Summary

The [previous qualification run](https://github.com/isaac-sim/IsaacLab/actions/runs/30377922938) collected 57 of 135 samples and never reached a verdict. This fixes the causes and narrows the re-run to the buckets that actually failed.

### Why it fell short

Two unrelated faults, both confirmed from the run logs:

1. **All five Newton-touching buckets crashed** on `from newton.solvers import SolverNotifyFlags`. The staging source was far behind `develop` while the image's Newton was newer, and that symbol had already been replaced by `ModelFlags`. This accounts for 75 of the 78 missing samples.
2. **One allocation hit a `PermissionError`** on `/tmp/jit-cache/warp/...`. All five parallel allocations shared a single Warp JIT cache directory and raced on creating it. This accounts for the remaining 3.

### What changed

- **Isolated the JIT cache per allocation**, keyed on run id, attempt, and run label.
- **Scoped the run to the Newton buckets**, 75 samples instead of 135. The four PhysX-only buckets already produced clean evidence, so re-measuring them would double the pool time for no new information. The seeder and the qualifier take the same backend allowlist, so the expected scope matches what is collected. A test keeps the two lists in sync, and an unknown backend key is rejected rather than silently narrowing the scope.
- **Benchmark execution health is always enforced.** A crash could previously exit zero while FPS regressions were advisory.
- **Added a stale-image guard** so that hardening cannot fail PRs for reasons outside their control. When a crash log shows the image missing a symbol the source pins, the gate reports a stale CI image and stays advisory for those tasks. Only packages the image installs (`newton`, `warp`, `isaacsim`, ...) qualify; a missing Isaac Lab symbol is still a real defect and still fails.
- **Qualification no longer writes baselines** or chains a second gate run. It measures the pool and reports a verdict.

### Depends on a staging catch-up

This PR deliberately contains no merge from `develop`, so the diff stays reviewable. It does **not** on its own fix cause 1: at the current staging tip, `newton_manager.py` still imports `SolverNotifyFlags`, so the Newton buckets would crash again and collect nothing.

**Staging must be caught up to `develop` before this qualification is worth running.** On current `develop`, `newton_manager.py` uses `ModelFlags` and imports only `SolverBase, SolverKamino`, which matches what the image provides. With this PR's guard in place, the crash would at least be reported as a stale image and stay advisory rather than reading as a performance failure.

### Scope note

The verdict will cover the five Newton buckets. The four PhysX buckets are not re-qualified at this commit, since their July 28 samples carry the old staging SHA and cannot be mixed with new ones. Runner-to-runner spread is largely a hardware property and the Newton buckets sample the same five runners, so this still answers the pool-stability question.

The gate remains in advisory mode.

## Test plan

- [ ] Land the staging catch-up to `develop` first.
- [ ] Merge triggers `Performance Smoke - L40S Runner Stability`, which waits for this commit's `Performance Smoke Test` to pass and for the L40 pool to go quiet.
- [ ] All five allocations complete with no `PermissionError` in the JIT cache.
- [ ] 75 of 75 samples collected (5 allocations x 3 samples x 5 Newton buckets).
- [ ] The qualification report reaches a verdict instead of `INCONCLUSIVE` from missing evidence.
- [ ] 104 unit tests pass and `pre-commit` is clean (verified locally).

### DIFF
--- a/.github/workflows/perf-smoke-runner-stability.yaml
+++ b/.github/workflows/perf-smoke-runner-stability.yaml
@@ -8,6 +8,10 @@
 # the commit out to five independent allocations so per-runner and run-to-run FPS
 # spread can be measured separately. Qualification only reports a verdict; it
 # never writes baselines.
+#
+# Scope is the Newton-touching buckets. The PhysX-only buckets already produced
+# clean evidence on 2026-07-28; these five crashed on a Newton import and are the
+# ones still lacking a verdict. Halving the matrix also halves the pool time.
 
 name: Performance Smoke - L40S Runner Stability
 
@@ -28,6 +32,11 @@ permissions:
 concurrency:
   group: perf-smoke-runner-stability-${{ github.ref }}
   cancel-in-progress: false
+
+env:
+  # Every backend_key whose benchmark loads Newton. Covers five task/backend
+  # buckets, since `newton` applies to both Cartpole and Velocity-Flat-G1.
+  STABILITY_BACKENDS: "newton,newton_rtx_renderer,newton_newton_renderer,physx_newton_renderer"
 
 jobs:
   wait_for_quiet_pool:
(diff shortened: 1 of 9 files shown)


Neil4561:
Superseded by #6864, which lands the gate directly on `develop`.

The fixes here are all carried over. Targeting `develop` also removes the `SolverNotifyFlags` crash at its root: that failure came from benchmarking three-week-old staging source inside a container built from current `develop`, and on `develop` the source and image move together.

PR #6864, opened 2026-08-03
Add the perf-smoke performance regression gate

Neil4561:
Adds a performance regression gate for pull requests. It benchmarks a fixed matrix of tasks on the L40S runner pool, compares each result against a rolling baseline stored on the `perf-baselines` branch, and posts a per-task verdict as a sticky PR comment and per-task commit statuses.

The gate is advisory. `gate_config.json` sets `blocking: false`, so it never fails a PR on its own. Making it an enforcing required check would be a separate rollout.

## How a run works

`perf-smoke-test.yaml` runs six jobs.

`config` reads the Isaac Sim version from `config.yaml`, and `validate` runs static checks on `tasks.json` so a malformed matrix fails on a free runner in seconds rather than after an hour of GPU time. Both are skipped for draft PRs.

`bench` fans out one GPU job per task/backend bucket. Each job resolves and pulls the CI image, restores a Warp JIT cache keyed on the Warp version, runs `perf_runtime.py` inside Docker with the repo bind-mounted, retries once on failure, normalizes the output into `perf_smoke_test_result.json`, and posts a per-task commit status.

`aggregate` downloads every bench artifact, loads the matching baselines, runs the oracle, writes the summary comment, and uploads results to omni-github. It runs with `always()` so a crashed bucket still gets reported instead of vanishing.

`baseline_update` appends the run's samples to `perf-baselines`, and only on pushes to `main`, `develop`, or a release branch. Pull requests are strictly read-only against the baseline branch, so nothing a PR does can move the numbers it is judged against.

`reseed` fills buckets that came out under-filled, by calling the seeding workflow for just those tasks at the pushed commit. Without it, a bucket that resets (say, after a launch-config change) would take five separate pushes to become useful again. It is self-limiting: once a bucket is full it stops being reported as under-filled, and seeding writes to `perf-baselines`, which triggers no workflow.

## How the verdict is decided

`oracle.py` produces one of `PASS`, `WARN`, `BLOCK`, or `HARD_FAILURE` per bucket, and the most severe signal wins.

A run is a `HARD_FAILURE` if the launch configuration did not match what was requested, if the result file is missing, or if the reported FPS is missing or not positive. A run reporting no forward progress is a dead run rather than a slow one, so it fails regardless of what the baseline says.

Otherwise the result is compared against the rolling window. The baseline contributes a median and a median absolute deviation, and the gate blocks when the measured FPS falls below `median - 4.0 x MAD` **and** the drop is at least 3 percent. The percentage floor exists so that a bucket with an unusually tight MAD cannot block on a fraction of a percent. Between `median - 2.5 x MAD` and the block line the verdict is `WARN`. There is also a noise floor so that a very tight baseline does not make the thresholds hypersensitive.

Buckets can additionally declare fixed FPS floors in `tasks.json`. A crossed gating floor forces a `BLOCK` on its own; a reporting-only floor is recorded without changing the verdict.

With no baseline, or fewer than five samples in the window, the verdict is `WARN` and the reason is stated as `NO_BASELINE` or `INSUFFICIENT_WINDOW`. The gate says it does not know rather than implying a pass.

## What gets measured

Nine task/backend buckets:

- `Isaac-Cartpole-Direct` on physx and on newton
- `Isaac-Velocity-Flat-G1` on physx and on newton
- `IsaacContrib-Factory-GearMesh-Direct` on physx
- `Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct` across four physics/renderer combinations (physx and newton, each with the RTX renderer and the Newton renderer)

`perf_runtime.py` drives each task with random actions and no policy, discards warmup frames at the source, and reports steady-state mean FPS.

## Baselines and how samples are matched

Baselines live on the orphan `perf-baselines` branch as append-only NDJSON, one file per GPU model, task, and backend.

A stored sample is only eligible if it matches on GPU model, task, backend, target branch, launch-config hash, baseline epoch, benchmark contract hash, and runtime contract hash, and if its commit is an ancestor of the PR's merge base. Ancestry matters because a sample from an unmerged branch describes code that may never land. The window holds between 5 and 20 samples, nearest commits first.

The hashes are the reason a verdict can be trusted: change the number of environments, the seed, the frame count, or the CUDA/driver pairing, and the fingerprint changes, so old samples stop matching instead of being silently compared against a different experiment.

## Keeping the environment stable

`image_era.py` derives a key from the container's inputs and looks it up in a manifest on `perf-baselines`. When an immutable image is recorded for that key the gate pins it; otherwise it falls back to the published `latest-develop` tag, and builds only if that pull fails. Pinning matters because a nightly rebuild of a floating tag would otherwise shift the environment underneath a baseline that was measured in the old one.

## When something goes wrong

Benchmarks retry once, since a single flaky container start should not read as a regression.

A crash always exits non-zero. Advisory mode suppresses regression failures, not execution failures, so a benchmark that never ran cannot be mistaken for one that ran fine.

Skew detection in `aggregate.py` handles one specific case that would otherwise blame the wrong person. The gate mounts PR source over a prebuilt image, so between a dependency pin landing and the next image publish, source can reference a symbol the installed package does not have yet. When a crash log shows a missing name from a package the image provides (`newton`, `warp`, `isaacsim`, `mujoco`, and so on), the summary reports a stale CI image and those tasks stay advisory. Isaac Lab's own source is mounted from the PR, so a missing symbol there is a real defect and still fails.

## Workflows

`perf-smoke-test.yaml` is the gate itself, described above.

`perf-smoke-seed-baselines.yaml` walks a slice of a branch's history, re-runs each commit's own benchmark, and appends the results, so a new bucket can be filled without waiting for five organic pushes. It runs on manual dispatch, defaulting to a dry run, and is called by the gate's reseed job. It resolves its image through the same code path the gate uses, so seeded samples and live runs share an environment.

`perf-smoke-unit-tests.yaml` runs the gate's own tests on `ubuntu-latest`.

## The Python

Under `tools/perf_smoke_test/`, roughly by role:

- Deciding: `oracle.py`, `gate_types.py`, `gate_config.py`
- Baselines: `baseline_manager.py`, `verify_baselines.py`, `seed_baselines.py`
- Identity and reproducibility: `backend_identity.py`, `gpu_identity.py`, `launch_config.py`, `runtime_contract.py`, `contracts.py`, `hashing.py`, `image_era.py`
- Running and normalizing: `perf_runtime.py`, `build_bench_result.py`, `benchmark_result_adapter.py`, `write_launch_config.py`
- Matrix and config: `tasks.json`, `task_config.py`, `tasks_to_ci_matrix.py`, `validate_tasks.py`, `gate_config.json`
- Reporting: `aggregate.py`, `omni_github.py`, `github_gate_context.py`

`tools/subprocess_runner.py` classifies which phase a failed subprocess died in, which is what lets a config mismatch be told apart from a crash mid-benchmark.

## Tests

72 tests covering the oracle's verdicts and thresholds, baseline matching and ancestry, contract hashing, seeding and cache isolation, skew detection, and the aggregate reporting path.

They are pure Python: no GPU, no simulator, no Isaac Lab install. `tools/perf_smoke_test/pyproject.toml` makes pytest root there so the isaaclab-importing `tools/conftest.py` above it is not loaded, which is the same arrangement `tools/skills/` and `tools/changelog/` already use.

## What to know before merging

It runs on every PR into develop from day one. The trigger list already covers `pull_request: [main, develop, release/**]`, so that is real L40S pool time on every PR. More of a capacity question than a code question.

Baselines start empty. Every sample on `perf-baselines` today is stamped `target_branch: perf-smoke/develop-staging`, and matching requires an exact target-branch match, so none carry over. All nine buckets begin with no baseline and report `NO_BASELINE` or `INSUFFICIENT_WINDOW` until five samples accumulate, which happens on its own from `develop` pushes. It sorts itself out after a few, but the gate is not useful on day one.

Six of the nine buckets have inert thresholds. Every Newton bucket and three camera buckets carry a `0.0` hard floor, which can never fire. Harmless while the gate is advisory, but they need real values before it could block anything.

There is no runner-stability verdict yet. The PhysX buckets produced clean samples on July 28; the Newton buckets never ran, because the staging branch had drifted 216 commits behind and its source called `newton.solvers.SolverNotifyFlags`, renamed to `ModelFlags` in the meantime. That was the frozen branch showing its age rather than a problem with the gate, and it does not arise on `develop`, where source and image move together. Landing here rather than on staging is what retires that failure mode. This supersedes #6863, which carried the same fixes against staging.

## Test plan

- [x] 72 passed, 1 skipped via `python3 -m pytest tools/perf_smoke_test/`
- [x] Tests collect and pass with no Isaac Lab install and no pytest flags
- [x] All pre-commit hooks pass
- [x] All three workflows parse as valid YAML
- [x] No changelog fragment required; the check only demands one for packages touched under `source/`
- [ ] After merge: first `develop` push runs the gate and begins filling baselines
- [ ] Confirm no image build or seeding run is triggered unprompted


fatimaanes (approved):

tools/perf_smoke_test/perf_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Perf-smoke-test runtime benchmark driver (random actions, no policy).
+
+Thin wrapper over the public Isaac Lab benchmark library that produces a
+schema-v1 :class:`~isaaclab.benchmark.schema.RuntimeBundle`.  It imports only
+the stable building blocks (:mod:`~isaaclab.benchmark.stepping`,
+:mod:`~isaaclab.benchmark.builders`, :mod:`~isaaclab.benchmark.capture`) rather
+than driving ``isaaclab benchmark runtime``, so the gate controls its own
```
AntoineRichard, line 12: **AI review — simplification:** This custom driver duplicates the supported isaaclab benchmark runtime entrypoint and the typed BenchmarkRuntimeRequest / run_runtime_benchmark() API. The existing implementation already performs random-action stepping, exact warm-up exclusion, backend presets, schema output, and provenance capture. Please call it directly and remove this fork. To preserve the current 300-total/100-warm-up behavior, pass 200 measured steps plus 100 warm-up steps, or redefine the matrix field as measured steps explicitly.

.github/workflows/perf-smoke-test.yaml
```diff
@@ -0,0 +1,1022 @@
...
+# (main/develop/release). Pull requests and merge queue candidates are read-only
+# and post verdicts without updating the rolling baselines.
+#
+# backend_key: "{physics_backend}_{render_backend}" when render_backend is
+# set, otherwise just physics_backend.  Matches TaskConfig.backend_key.
+
+name: Performance Smoke Test
+
+on:
+  pull_request:
```
AntoineRichard, line 32: **AI review — operational scope:** There is no path/change filter here, so every PR into these branches launches nine L40S jobs. The latest run consumed about 145 combined GPU runner-minutes and still produced no usable regression comparison. Please reuse the repository's existing detect-changes action, or start with label/manual/nightly execution until the gate is calibrated and reliable.

.github/workflows/perf-smoke-test.yaml
```diff
@@ -0,0 +1,1022 @@
...
+            echo "::notice::No known login for registry '${REGISTRY}'; attempting anonymous pull"
+            ;;
+        esac
+
+        echo "🔵 Pulling prebuilt image ${CI_IMAGE_REF}..."
+        docker pull "${CI_IMAGE_REF}"
+        docker tag "${CI_IMAGE_REF}" "${{ env.CI_IMAGE_TAG }}"
+        echo "🟢 Tagged ${CI_IMAGE_REF} as ${{ env.CI_IMAGE_TAG }}"
+
+    - name: Build CI image (fallback)
```
AntoineRichard, line 290: **AI review — simplification:** Pulling a moving image, overlaying PR source, and conditionally building a fallback creates the source/dependency skew that later requires image eras, runtime-contract hashes, and stale-image classification. The normal Docker + Tests workflow already produces a PR-SHA-specific image through ecr-build-push-pull. Please integrate this job with that workflow and consume the exact PR image; most of this image lifecycle and its compensating machinery should then disappear.

tools/perf_smoke_test/gate_config.json
```diff
@@ -0,0 +1,6 @@
...
+  "blocking": false,
```
AntoineRichard, line 3: **AI review — result semantics:** A failing but non-required GitHub check is already advisory. Returning success for BLOCK/HARD_FAILURE and manufacturing a separate red commit status adds another protocol and allowed the latest Aggregate + Verdict job to stay green despite a hard failure and a missing bucket. Let the aggregate check reflect the real verdict; make it required later through branch protection when the rollout is ready.

.github/workflows/perf-smoke-test.yaml
```diff
@@ -0,0 +1,1022 @@
...
+    #
+    # It carries the VERDICT, independently of whether the aggregate job passed:
+    # in advisory mode a BLOCK/HARD_FAILURE paints this status red while the job
+    # stays green, so the signal is visible without an unexplained red check on an
+    # unrelated pull request. If the aggregate step could not produce a verdict at
+    # all, the status reports that rather than silently claiming success.
+    # Distinct from the per-task `perf-smoke (...)` statuses, which only report
+    # whether each benchmark ran.
+    # Skipped for fork pull requests; see the per-task status step for why.
+    - name: Report aggregate status
```
AntoineRichard, line 845: **AI review — reporting:** This custom status path is unavailable for fork PRs, including this PR; the live run produced no commit statuses or sticky verdict comment. Prefer the native Actions check conclusion plus job summary/JUnit/Omni ingestion, which work without write access to the PR. If a PR comment is essential, it should be produced by a separate trusted workflow_run consumer rather than PR-controlled code.

.github/workflows/perf-smoke-test.yaml
```diff
@@ -0,0 +1,1022 @@
...
+  # adds one sample -- the bucket would stay advisory (NO_BASELINE /
+  # INSUFFICIENT_WINDOW) for several more pushes. This fills it in one shot by
+  # reseeding exactly the flagged tasks for the pushed commit through the same
+  # era-image resolver the gate used, so the samples share the gate's environment.
+  #
+  # Self-limiting: once filled the bucket is no longer under-filled, so the next
+  # push emits no reseed_tasks. No loop risk -- seeding writes to perf-baselines,
+  # which triggers neither this gate nor the seed workflow.
+  # ---------------------------------------------------------------------------
+  reseed:
```
AntoineRichard, line 995: **AI review — scope:** Automatic historical reseeding is a separate product concern, not a prerequisite for the first advisory gate. The baseline branch is currently populated for perf-smoke/develop-staging and was unusable by this develop-targeted run. Please first prove the simple protected-branch write/read path (preferably through the existing internal performance store), then add seeding only if organic target-branch runs are demonstrably insufficient.

AntoineRichard (commented):
## AI review — architecture and simplification

This review was generated by AI at the request of the reviewer. It focuses primarily on simplification, reuse of IsaacLab's existing benchmark/CI infrastructure, and the maintenance cost of the proposed design.

### Overall finding

I do not think this should land in its current shape. The underlying requirement is useful and comparatively small:

1. Define a task/backend matrix.
2. Run the supported runtime benchmark.
3. Read one FPS metric from its RuntimeBundle.
4. Compare it with an authoritative baseline.
5. Publish one visible result.
6. Record trusted target-branch measurements for future comparisons.

The PR instead adds a largely self-contained performance-CI platform: 10,017 lines across 41 new files, with no deletions.

| Area | Added lines |
|---|---:|
| Baseline database, historical seeding, verification, and image eras | 2,023 |
| Oracle, aggregation, GitHub/Omni reporting | 1,676 |
| Benchmark runner, adapters, and subprocess management | 1,407 |
| GitHub workflows | 1,426 |
| Configuration, identities, hashes, and contracts | 881 |
| Tests | 2,479 |
| JSON/TOML/support files | 125 |

Tests are valuable, but they do not reduce the long-term ownership of roughly 7,500 lines of new production/configuration code. This is not just wiring up a benchmark gate; it creates another benchmark runner, result schema, compatibility system, baseline database, image lifecycle, reporting stack, and historical execution engine inside IsaacLab.

### The benchmark runner is duplicated

The clearest simplification is tools/perf_smoke_test/perf_runtime.py. Its stated reason for bypassing the supported isaaclab benchmark runtime command is control over warm-up and output layout. Those capabilities already exist:

- isaaclab benchmark runtime supports random-action stepping, exact warm-up exclusion, measured-step count, backend/renderer presets, schema output, and provenance.
- BenchmarkRuntimeRequest and run_runtime_benchmark() expose the same workflow directly to Python and return the typed bundle plus output paths.
- RuntimeBundle already contains run configuration, versions, hardware, runtime metrics, and resources.

The gate should call that API or CLI rather than maintaining a fork of the runtime entrypoint. To preserve this PR's current contract, note that num_frames=300 with warmup_frames=100 currently produces 200 measured steps; the supported API would use num_steps=200 and warmup_steps=100. Alternatively, define num_steps as the measured window explicitly and remove this ambiguity.

The current pipeline then translates the supported schema through several additional representations:

RuntimeBundle → RuntimeSample → BenchResult → baseline record → OracleResult → Omni-GitHub row

Most of the identity and provenance carried through those types already exists in RuntimeBundle. A small consumer can parse the supported bundle directly. If some field is genuinely missing, it should normally be added to the shared benchmark schema rather than reconstructed in a gate-specific parallel contract.

### Several large subsystems solve complexity introduced by this architecture

The workflow pulls a published image, overlays PR source on it, and builds a fallback image when pulling fails. That makes source/dependency skew possible, which then motivates image-era manifests, runtime-contract hashes, stale-image classification, fallback behavior, and historical image resolution.

The normal Docker + Tests workflow already builds and publishes a PR-SHA-specific image using the shared ecr-build-push-pull action. The performance job should consume that exact image, preferably from the same workflow or as a dependent workflow/job. Doing so should remove most or all of:

- image_era.py
- the moving-image manifest on perf-baselines
- the pull/build/retag implementation duplicated in this workflow
- the stale-image heuristic in aggregate.py
- a substantial part of runtime-contract matching

The same issue exists in reporting. This PR publishes per-task statuses, an aggregate status, a sticky comment, a job summary, normal artifacts, and a custom Omni-GitHub artifact. IsaacLab already has shared JUnit summary and Omni-GitHub upload actions. Prefer one native check result plus the existing result-ingestion path. If structured performance fields are missing from the shared uploader, extend that shared boundary once rather than adding another serializer and validation path here.

### Advisory should be a repository setting, not a second status protocol

The current blocking=false mode forces aggregate.py to exit successfully for BLOCK and HARD_FAILURE, then attempts to communicate failure through a separate commit status. A failing but non-required GitHub check is already advisory. When the gate is ready to enforce, repository branch protection can make that check required without changing the result protocol.

This would eliminate the blocking-mode split, custom aggregate status, and the risk of a green Actions check hiding an unhealthy benchmark.

The latest run demonstrates this problem:

- Nine GPU jobs used about 145 combined L40S runner-minutes.
- Seven completed measurements produced WARN without a usable baseline comparison.
- One produced HARD_FAILURE during initialization.
- One timed out without producing an artifact.
- Aggregate + Verdict still passed.
- This is a fork PR, so the custom statuses and sticky comment were skipped; the meaningful result remained in the job summary/artifact.

That is a large operational cost without answering the core reviewer question: did this PR regress performance?

### Baseline storage needs an explicit design decision

Before maintaining an append-only NDJSON database on an orphan Git branch, please establish why the existing internal performance storage/comparison tooling cannot be authoritative. The benchmark framework already emits OmniPerf and typed schema formats, and this PR also uploads to Omni-GitHub.

The perf-baselines implementation brings worktree management, transactional pushes, retries, ancestry filtering, stable sample IDs, branch matching, compatibility hashes, historical checkout execution, automatic reseeding, and verification. Some of these may be legitimate future requirements, but they should not all be prerequisites for an initial advisory smoke test.

The existing baseline branch is populated with records targeted at perf-smoke/develop-staging, while this PR targets develop, so the current live run cannot use them. Automatic historical reseeding should be deferred until the simple protected-branch record/read path is proven useful.

Questions that need explicit answers:

1. What existing internal service is intended to store and compare IsaacLab benchmark results, and why can it not be called here?
2. What exact metric is authoritative: runtime.total_fps.mean or runtime.environment_step_timing.environment_step_fps.mean?
3. Why is an ancestry-filtered rolling MAD window required for the initial gate instead of a service-provided baseline or a simple calibrated threshold?
4. Which identity fields materially affect comparability, and which are merely diagnostic?
5. Why must all nine L40S jobs run on every PR, including changes unrelated to runtime performance?
6. Is a sticky comment/custom status actually required when it cannot run for fork PRs?

### Suggested replacement architecture

A much smaller first version would be:

- Reuse the existing change detector or an explicit label/manual/nightly trigger.
- Reuse the PR-SHA CI image from Docker + Tests.
- Keep a small declarative matrix containing task, presets, num_envs, measured steps, warm-up steps, and timeout.
- Invoke isaaclab benchmark runtime or run_runtime_benchmark() directly.
- Read the selected metric from RuntimeBundle.
- Query the existing internal baseline/comparison tooling.
- Emit one normal check/JUnit result and reuse the existing Omni-GitHub upload action.
- Record results only from trusted protected-branch runs.
- Let regressions fail a non-required check during the advisory period.
- Add rolling statistics, automated seeding, sticky comments, or blocking only after actual data shows they are needed.

If the internal baseline/comparison service covers storage and policy, this should plausibly be hundreds of lines rather than ten thousand. If a comparator is genuinely missing, it should still be possible to keep the change to a small, well-tested comparator and a thin workflow integration.

### Requested direction

Please reconsider this as a replacement rather than trying to incrementally trim the current implementation. A good next revision would prove one or two buckets end-to-end using the existing benchmark API and normal CI image, then expand the matrix once the baseline lookup and reviewer-visible verdict are demonstrably working.

I inspected the full 41-file change, surrounding benchmark/CI APIs, PR discussion, current baseline branch contents, and the latest live workflow artifacts. I did not run the GPU benchmarks locally. The PR's remote pre-commit and pure-Python gate test checks passed.

mataylor-nvidia:
It might be good to have some way to enable this once a PR makes it to develop. 

If it is run on every PR it may take a long time. If we could tag a commit to get the perf numbers it would be very useful and only exercise CI runners on demand

PR #6871, opened 2026-08-04
Cleanup MDP for locomotion

StafaH:
# Clean up ant and humanoid locomotion, and make the direct and manager workflows identical

## Summary

Cleans up the ant and humanoid locomotion tasks against the conventions used by `core/cartpole`, and
makes the direct-workflow and manager-based environments define the same MDP so both converge to the
same reward.

The cleanup surfaced four genuine bugs, one of which was already breaking humanoid training on
`main` before this branch existed.

## Motivation

`Isaac-Ant` / `Isaac-Ant-Direct` and `Isaac-Humanoid` / `Isaac-Humanoid-Direct` are meant to be two
presentations of the same task, but they had drifted into two different problems. The direct
environments were not scaling rewards by `step_dt` the way `RewardManager` does, so the two
workflows were not even on a common reward scale — 9641 versus 91 on the same task.

The direct environments also carried a fair amount of dead machinery inherited from the IsaacGym
port, and encoded joint gears as backend-ordered lists that had to be duplicated per physics
backend.

### Direct/manager parity

The manager-based environments were the better-specified MDP, so they are the basis. The direct
environments gained:

- feet joint wrench observations (`observation_space` 36 → 60 for ant, 75 → 87 for humanoid)
- reset joint randomization (±0.2 rad, ±0.1 rad/s, clamped into the joint limits)
- `step_dt`-scaled rewards, matching `RewardManager.compute()`
- the gear-weighted energy and joint-limit penalties
- `episode_length_s` 15.0 → 16.0 and `env_spacing` 4.0 → 5.0

Where the direct environments were the better-specified side, the manager-based tasks were changed
instead rather than copying the weaker behaviour:

- the terminal death cost is now a `terminating` reward term on both workflows
- the joint effort action is now clipped on both workflows

### Cleanup

- `joint_gears` is now a dict keyed by joint name expression instead of a backend-ordered list, so a
  single table is correct for every physics backend and the `isinstance(cfg.sim.physics, ...)`
  branch is gone
- removed dead IsaacGym-era machinery from `LocomotionDirectEnv`: the start rotation was the identity
  quaternion, so `inv_start_rot`, `basis_vec0/1` and the `quat_mul` inside `compute_heading_and_up`
  were all no-ops; `motor_effort_ratio` was hardcoded to ones; `up_vec`, `heading_vec` and `pitch`
  were computed but never read
- replaced the local `normalize_angle` helper with `isaaclab.utils.math.wrap_to_pi`, and used it in
  the manager MDP terms too so both workflows wrap angles identically
- dropped a redundant `self.robot.reset()` (the base `_reset_idx` already calls `scene.reset`) and a
  private `_ALL_INDICES` access
- absolute imports in `mdp/rewards.py` and `mdp/__init__.pyi`, two stray `# isort: skip` markers
  removed, and the humanoid gear table hoisted to a single `JOINT_GEARS` constant instead of being
  written out three times
- split the `Metrics/success_rate` logging out of `progress_reward` into a `survival_success_rate`
  term, matching the cartpole convention
- removed the pass-through `__init__` on `AntEnv` / `HumanoidEnv` and an unused `contact_force_scale`
  field (later reintroduced as the feet wrench observation scale)

## Bugs fixed

**Unbounded joint efforts drove the MJWarp solver to `NaN`.** Humanoid training aborted partway
through a run with `The observation group 'policy' ... contains NaN values`. Reproduced in 10 steps
with large actions; the corruption originates in the raw articulation state (`joint_pos`,
`joint_vel`, `root_lin_vel_w` all non-finite while the quaternion is still unit-norm), not in any
observation or reward math.

| action scale | clamped | result |
|---|---|---|
| 3.0 | no | NaN at step 10 |
| 1.0 | no | clean 6000 steps |
| 3.0 | yes | clean 6000 steps |
| 5.0 | yes | clean 6000 steps |

This is **pre-existing on `main`** — confirmed by stashing this entire branch and reproducing on
unmodified code, where `Isaac-Humanoid` failed at iteration 872. The manager-based tasks never
clipped their joint efforts. Fixed by clipping on both workflows.

**The humanoid direct agent trained at a different rate.** `HumanoidDirectPPORunnerCfg` overrode
`learning_rate` (1.0e-4 versus 5.0e-4), `desired_kl` and `value_loss_coef`, so the two workflows
trained differently despite sharing an MDP. It now inherits the algorithm settings and overrides only
the experiment name, as the ant and cartpole configs do. This closed a 10.4% reward gap to 0.4%.

**The manager-based tasks could not run on OvPhysX.** `presets=ovphysx` failed with
`Unknown preset(s): ovphysx` because their physics preset configurations were missing the `ovphysx`
entry the direct configurations already declared. Also pre-existing.

**`progress_reward` was inconsistent across reset.** `reset()` computed the distance to the target
without zeroing the vertical component, while `__call__` zeroed it, so the potential recorded at
reset did not match any subsequent step.

**The manager-based walk target was an absolute world position.** `(1000, 0, 0)` was shared by every
environment, so robots in different grid cells aimed along measurably different directions. It is now
offset by each environment origin, matching the direct environments.

## Verification

Parity was checked by driving both workflows from identical states and comparing tensors directly,
which is independent of RL seed noise:

| | ant | humanoid |
|---|---|---|
| observation dimension | 60 = 60 | 87 = 87 |
| step-0 observation max abs difference | 0.0 | 0.0 |
| reward max abs difference, 60 steps | 0.0 | 4e-9 |

The humanoid comparison was repeated with actions large enough that the new effort clip binds on
every joint, to exercise that path.

Training, `newton_mjwarp`, seed 42, 1000 iterations:

| task | manager | direct |
|---|---|---|
| ant | 80.36 | 74.68 |
| humanoid | 142.11 | 141.54 |

The ant difference sits inside its own run-to-run spread: the same config and seed produced 81.98 and
74.68 on two separate runs, so roughly 9% of nondeterminism between runs.

Smoke tested (3 iterations, all four tasks) on `newton_mjwarp`, `isaacsim_physx` and `ovphysx` —
12/12 pass. `uv run isaaclab -f` is clean.

## Breaking changes

- `observation_space` for the direct tasks changed from 36 to 60 (ant) and 75 to 87 (humanoid), and
  `episode_length_s` from 15.0 to 16.0. **Existing direct-workflow checkpoints will not load.**
- `joint_gears` changed from an ordered list (or a `{"physx": [...], "newton": [...]}` dict) to a dict
  keyed by joint name expression. Migrate by replacing the list with regex entries, e.g.
  `joint_gears = [15, 15, 15, 15, 15, 15, 15, 15]` becomes `joint_gears = {".*": 15.0}`.
- Reward magnitudes for the direct tasks changed by roughly 60x because they are now `step_dt`-scaled
  like the manager-based tasks. Logged reward curves are not comparable across this change.
- The joint effort clip changes the reachable effort envelope for the manager-based tasks, so
  manager-based checkpoints will behave differently even though they still load.

## Known gaps

- **The experimental warp frontend is not ported.** Its observation kernel writes a fixed 36/75-value
  layout with no wrench block and unscaled rewards. Rather than let it silently zero-fill the tail of
  a 60/87 buffer, it now raises with an explicit message. `--frontend warp` on these tasks does not
  run until it is ported.
- **Convergence is only established on `newton_mjwarp`.** PhysX and OvPhysX are smoke tested but not
  trained to convergence.
- The direct configurations now set `bounce_threshold_velocity=0.2` on the Isaac Sim PhysX preset to
  match the manager-based tasks, instead of leaving the 0.5 default. This changes PhysX behaviour for
  the direct tasks and has not been validated by a convergence run.
- The `NaN` fix addresses the trigger, not the solver behaviour. Whether MJWarp should emit `NaN`
  rather than erroring on a diverging articulation is a separate question, and a sufficiently bad
  contact could still diverge inside the clipped envelope.

## Checklist


kellyguo11 (dismissed):
I don't think breaking --frontend warp is a good idea. we should still get them working with the new updated mdp

source/isaaclab_tasks/isaaclab_tasks/core/locomotion/locomotion_direct_env.py
```diff
@@ -19,98 +17,45 @@
...
-            elif isinstance(self.cfg.sim.physics, NewtonCfg):
-                joint_gears = self.cfg.joint_gears["newton"]
-            else:
-                raise ValueError(f"Invalid physics type: {self.cfg.sim.physics}")
-        else:
-            joint_gears = self.cfg.joint_gears
-        self.joint_gears = torch.tensor(joint_gears, dtype=torch.float32, device=self.sim.device)
-        self.motor_effort_ratio = torch.ones_like(self.joint_gears, device=self.sim.device)
+        # resolve the gears by joint name, since the joint ordering differs across physics backends
+        self.joint_gears = torch.zeros(self.robot.num_joints, device=self.sim.device)
```
kellyguo11: Could we validate that the resolved `joint_ids` cover every robot joint, or use the same documented fallback as the manager path? `resolve_matching_names_values` requires every regex key to match something, but it does not require every joint name to be matched. Unmatched joints therefore remain zero here (and in the Warp implementation), while `JointAction` and the manager reward terms initialize unmatched entries to one. A partial but otherwise valid mapping would silently produce different efforts and penalties between the direct and manager workflows.

source/isaaclab_tasks/isaaclab_tasks/core/locomotion/ant/ant_manager_env_cfg.py
```diff
@@ -201,6 +202,11 @@ class RewardsCfg:
     joint_pos_limits = RewTerm(
         func=mdp.joint_pos_limits_penalty_ratio, weight=-0.1, params={"threshold": 0.99, "gear_ratio": {".*": 15.0}}
     )
+    # (8) Penalty for falling over. The manager scales every term by ``step_dt``, so the weight is
+    #     the -2.0 terminal cost divided by the 1/60 s step, giving a one-off -2.0 on the dying step.
+    terminating = RewTerm(func=mdp.is_terminated, weight=-120.0)
```
kellyguo11: Could this terminal penalty be made independent of the configured `step_dt` rather than hard-coding the inverse of the current 1/60 s step? The direct environment applies `death_cost` as a one-off constant, whereas `RewardManager` multiplies this weight by `step_dt`; overriding `sim.dt` or `decimation` therefore changes the manager death cost and breaks parity again. For example, a terminating term normalized by `env.step_dt` could retain a natural `-2.0` weight. The Humanoid `-60.0` term has the same issue.

kellyguo11 (commented):
Two additional parity concerns.

AntoineRichard (approved):
Reviewed the direct/manager locomotion parity changes at this head. Kelly’s Warp compatibility, joint-gear fallback, and step-dt-independent terminal penalty concerns are addressed, along with the follow-up reset-state and configuration parity fixes.

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/core/locomotion/locomotion_env_warp.py
```diff
@@ -388,6 +383,18 @@ def __init__(self, cfg: DirectRLEnvCfg, render_mode: str | None = None, **kwargs
         self.root_vel_w = self.robot.data.root_vel_w.warp
         self.soft_joint_pos_limits = self.robot.data.soft_joint_pos_limits.warp
 
+        # The observation kernel writes a fixed proprioceptive layout and has no feet-wrench block,
+        # unlike the stable LocomotionDirectEnv it mirrors. Fail loudly rather than silently leaving
+        # the tail of the buffer zeroed if the shared config expects the richer observation.
+        expected_observation_space = 12 + 3 * self.robot.num_joints
+        if self.cfg.observation_space != expected_observation_space:
```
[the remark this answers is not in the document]
AntoineRichard: Addressed in `10639a73ca`. The Warp observation kernel now writes the feet force/torque block, and the guard uses `12 + 3 * num_joints + 6 * num_feet`. This yields 60 observations for Ant and 87 for Humanoid, matching the shared configurations, so the registered Warp locomotion tasks no longer fail this check.

source/isaaclab_tasks/isaaclab_tasks/core/locomotion/locomotion_direct_env.py
AntoineRichard: Addressed in `10639a73ca`. The stable direct environment now initializes the gear tensor with `torch.ones`, and the Warp frontend initializes its list with `1.0` before applying the resolved mappings. Unmatched joints therefore use the same unit-gear fallback as `JointAction` and the manager reward terms.

source/isaaclab_tasks/isaaclab_tasks/core/locomotion/ant/ant_manager_env_cfg.py
AntoineRichard: Addressed in `10639a73ca`. `mdp.terminated_penalty` now returns `terminated / env.step_dt`, while the Ant and Humanoid weights are `-2.0` and `-1.0`. `RewardManager` then multiplies by `step_dt`, producing the intended one-off terminal costs independently of `sim.dt` and `decimation`.

kellyguo11 (approved):

PR #6876, opened 2026-08-04
So101 keyboard

ooctipus:
# Description

Adds the `IsaacContrib-Keyboard-SO101` manager-based reinforcement-learning task under `isaaclab_tasks.contrib.keyboard`. The task trains an SO-101 arm to type target words while preserving the training and evaluation behavior validated by the accompanying checkpoint.

The change includes:

- Procedurally generated, logo-agnostic keyboard families with varied layouts, geometry, materials, and articulated keys. Both PhysX and Newton use the full keyboard pool rather than a single fixed keyboard.
- Typing commands, observations, rewards, terminations, reset-state replay, success-rate curriculum sampling, and visualization markers.
- A single RSL-RL actor/critic configuration with a shared MLP encoder for the one-hot typing command.
- Physics presets for concrete Isaac Sim PhysX (`isaacsim_physx`, the default), automatic PhysX selection (`physx`), and Newton MJWarp (`newton_mjwarp`).
- Focused configuration and architecture tests, plus environment-catalog documentation.

Implementation reference: [Emergent Dexterity via Diverse Resets and Large-Scale Reinforcement Learning (OmniReset)](https://arxiv.org/abs/2603.15789).

### Asset dependency

The SO-101 USD is hosted separately and is intentionally not included in this PR. Before merge, it should be published under the official IsaacLab Nucleus `Robots/SO101/` directory and resolved through `ISAAC_NUCLEUS_DIR`.

### Training and evaluation

The policy can be evaluated with:

```bash
./isaaclab.sh play --task=IsaacContrib-Keyboard-SO101 --num_envs=32 physics=newton_mjwarp --rl_library=rsl_rl
```

Manual checkpoint playback validates that the policy solves the typing task reliably.

## Type of change

## Screenshots

A task screenshot still needs to be attached and added to the environment catalog at:

`docs/source/_static/tasks/manipulation/so101_keyboard.jpg`

## Checklist


docs/source/overview/environments.rst
```diff
@@ -1030,6 +1030,10 @@ including disabling runtime perturbations used for training.
       - Direct
       - **rl_games** (PPO), **rsl_rl** (PPO), **skrl** (PPO)
       - **physics=** ``isaacsim_physx``, ``newton_mjwarp``, ``ovphysx``, ``physx``
+    * - Isaac-Keyboard-SO101
```
kellyguo11: should this be IsaacContrib?

docs/source/_static/css/environment-browser.js
```diff
@@ -127,6 +128,7 @@
             [/Reorient-Franka/, "tasks/manipulation/franka_lift.jpg"],
             [/Reorient-KukaAllegro/, "tasks/manipulation/kuka_allegro_reorient.jpg"],
             [/Shadow-Handover/, "tasks/manipulation/shadow_hand_over.jpg"],
+            [/Keyboard-SO101/, "tasks/manipulation/so101_keyboard.jpg"],
```
AntoineRichard, line 134: **AI-generated review — Important (documentation):** This rule points to `docs/source/_static/tasks/manipulation/so101_keyboard.jpg`, but that file is absent from both the PR and the base tree. The browser will request a missing `_images/so101_keyboard.jpg`; the default image only applies when no regex matches and does not handle this 404. Add the licensed screenshot in this PR, or remove this rule until the asset exists.

source/isaaclab_tasks/isaaclab_tasks/core/lift/lift_env_cfg.py
```diff
@@ -86,6 +86,7 @@ class ObjectCfg(PresetCfg):
         mass_props=sim_utils.MassPropertiesCfg(mass=0.2),
     )
     default = shapes
+    ovphysx = cube
```
AntoineRichard, line 92: **AI-generated review — Important (API/scope):** This unrelated preset move is coupled to deleting the top-level `KukaAllegroObjectCfg` class in `kuka_allegro_env_cfg.py`. That is a public symbol removal without a prior deprecation, contrary to the repository's breaking-change rule. Revert the three `core/lift` edits from this keyboard PR. If the consolidation is needed, handle it separately with deprecation and focused regression coverage.

source/isaaclab_tasks/isaaclab_tasks/core/lift/__init__.py
```diff
@@ -23,4 +23,15 @@
   year={2024}
 }
 
+The MDP implementation is based primarily on OmniReset:
```
AntoineRichard: **AI-generated review — Moderate (documentation duplication):** OmniReset is already cited by the new keyboard package. Adding the same bibliography to the unrelated core lift package duplicates documentation and incorrectly implies that the whole lift MDP is based on this work. Remove this block with the other out-of-scope `core/lift` changes.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_pool.py
```diff
@@ -0,0 +1,54 @@
...
+    """Spawner configurations and key metadata consumed by the task."""
+
+    spawners_partitioned: tuple[KeyboardSpawnerCfg, ...]
+    spawners_single: tuple[KeyboardSpawnerCfg, ...]
+    backspace_slot: int
+    active_slots: tuple[int, ...]
+    slot_labels: tuple[str, ...]
+
+
+_BASE_SPAWNER = KeyboardSpawnerCfg(family="ansi_full", topology_mode="exact", partition_dof=_PARTITION_DOF)
```
AntoineRichard, line 31: **AI-generated review — Important (scope/documentation):** Every task variant derives from `family="ansi_full"`, so the environment does not vary layouts as the PR description claims. The other layout families, topology modes, and tail policy add a large unexercised surface. Either make layout variation real while preserving the task's stable slot/DOF contract and test it, or correct the description and remove/split the unused generalization.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,353 @@
...
+    typing = mdp.LetterTypingCommandCfg(
+        asset_name="robot",
+        object_name="keyboard",
+        resampling_time_range=(10.0, 10.0),
+        debug_vis=False,
+        letter_length=(1, 5),
+        max_len=5,
+        # letter_full shows both target and typed letters; letter_left shows only the remaining target
+        command_mode="letter_full",
+        typeable_slots=tuple(
```
AntoineRichard, line 148: **AI-generated review — Important (behavioral contract):** This excludes only Backspace, so targets include Ctrl, Shift, F-keys, arrows, media keys, and punctuation. That is an arbitrary key sequence, not the `letter` / `target word` behavior documented throughout the task. Either restrict targets to the intended alphabetic slots (and test the exact set), or consistently rename the command and documentation to key-sequence typing.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/rewards.py
```diff
@@ -0,0 +1,150 @@
...
+    r"""High-water-mark typing-progress reward: a per-episode ratchet on the correct-prefix length.
+
+    Fetches the :class:`LetterTypingCommand` term and returns a sparse per-step signal driven only by
+    *records* in the length of the contiguous correct prefix (``prefix_len``):
+
+    * ``+1`` on any step that sets a new episode maximum prefix length (genuine forward progress),
+    * ``-1`` on any step that sets a new episode minimum prefix length (destroying correct progress),
+    * ``0`` otherwise - staying put, or re-reaching a prefix length already seen this episode.
+
+    Since one control step registers at most one keystroke, ``prefix_len`` moves by at most one, so each
```
AntoineRichard, line 51: **AI-generated review — Important (correctness):** This premise contradicts `_update_command`, which explicitly registers every newly pressed key in a control step. If simultaneous presses happen to match consecutive target slots in slot order, `prefix_len` can jump by more than one while the boolean `new_high` reward still returns only `+1`. Decide the intended contract: use a prefix delta for per-key progress, or enforce one registered keystroke per step. Add a simultaneous-press regression test; then shorten this docstring to the resulting functional rule.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,353 @@
...
+        self.episode_length_s = 6.0
+        self.sim.render_interval = self.decimation
+        self.sim.default_visualizer_cfg = VisualizerCfg(
+            eye=(0.85, -0.75, 1.0), lookat=(0.25, 0.0, 0.1), focal_length=28.0
+        )
+
+    def play_mode(self):
+        """Enable typing markers for every playback environment."""
+        super().play_mode()
+        self.num_envs = 36
```
AntoineRichard, line 340: **AI-generated review — Important (configuration):** Environment count lives at `self.scene.num_envs`; `ManagerBasedRLEnvCfg.play_mode()` already clamps that field to 50. Assigning a new `self.num_envs` attribute here does not reduce the scene to 36, so this override does not do what its comment implies. Set `self.scene.num_envs = min(self.scene.num_envs, 36)` (or rely on the base clamp) and cover the play-mode result in the consolidated config test.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands.py
```diff
@@ -0,0 +1,1121 @@
...
+            # Iteration budget: an int is a fixed count; a (min, max) tuple is sampled per env each solve so
+            # poses land at varying convergence (a reach-difficulty gradient). Normalize to (min, max).
+            it = cfg.reset.ik_iters
+            self._ik_iters = (int(it), int(it)) if isinstance(it, int) else (int(it[0]), int(it[1]))
+            self._reset_ik = DifferentialIKController(ik_cfg.controller, num_envs=self.num_envs, device=self.device)
+
+        # Success-conditioned reset curriculum (buffer built lazily on the first reset; see _build_buffer).
+        # Needs reset-IK to synthesize the snapshot poses, so it stays off when reset.ik is unset.
+        cur = cfg.reset
+        self._cur_enabled = bool(cur.enabled) and self._reset_ik is not None
```
AntoineRichard, line 271: **AI-generated review — Important (error handling):** `enabled=True` with no IK is an invalid curriculum configuration, but this silently changes the requested training distribution and continues. Fail fast with `ValueError` during configuration/term construction instead of printing and disabling the feature. A log message is not a substitute for enforcing this invariant.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands_cfg.py
```diff
@@ -0,0 +1,240 @@
...
+from .typing_commands import LetterTypingCommand
+
+
+@configclass
+class LetterTypingCommandCfg(CommandTermCfg):
+    """Configuration for :class:`LetterTypingCommand`."""
+
+    @configclass
+    class ResetCfg:
+        """All reset behavior for the typing task: the per-reset differential-IK arm snap (the ``ik*`` fields;
```
AntoineRichard, line 26: **AI-generated review — Moderate (comment/doc hygiene):** This public config docstring and many field docs narrate implementation history and tuning rationale (`no hand-tuned`, why particular weights/defaults were chosen) instead of stating the durable contract. Reduce `ResetCfg` to its purpose, dependencies, constraints, and units; keep each field to one or two functional sentences. Move algorithm derivation to one design document if it must be retained. While doing so, remove invalid `~...` Sphinx targets and references to private methods.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,353 @@
...
+class KeyboardAssetCfg(PresetCfg):
+    """Backend-dependent keyboard articulation, selected by the ``physics=`` preset.
+
+    PhysX flattens all per-env articulation instances onto one axis, so the ``fixed_dof`` partition
+    (18 roots/env) exposes every key. IsaacLab's Newton :class:`ArticulationData` reads only the first
+    articulation per env (``[:, 0]``), so Newton uses a single 108-DOF articulation instead.
+    """
+
+    # PhysX: 18 articulation roots/env under ``parts/part_*`` (its ArticulationView flattens them).
+    # (Octi) partitioned keyboard investigate if topologies can be harnessed to improve performance
```
AntoineRichard, line 46: **AI-generated review — Moderate (comment hygiene):** This owner-directed investigation note is neither a functional invariant nor actionable repository documentation. Delete it or track the investigation in an issue. Keep only the preceding concise explanation of the PhysX partition. The nearby checkpoint-history comment should likewise describe current behavior directly (for example, placing the keyboard in the robot workspace) rather than why a historical value was chosen.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,353 @@
...
+        func=mdp.illegal_contact,
+        params={"sensor_cfg": SceneEntityCfg("robot_contact"), "threshold": 20.0},
+    )
+
+    success = DoneTerm(func=mdp.typing_complete, params={"command_name": "typing"})
+
+
+@configclass
+class PhysicsCfg(PresetCfg):
+    # Octi: note
```
AntoineRichard, line 287: **AI-generated review — Moderate (comment hygiene):** Replace the author-addressed `Octi: note` with functional documentation, e.g. a `PhysicsCfg` docstring stating that PhysX is evaluation-only and Newton MJWarp is intended for training. If this is a user-visible performance limitation, document it in the environment docs with evidence; do not leave an agentic note in showroom code.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands.py
```diff
@@ -0,0 +1,1121 @@
...
+
+    def _log_buffer_stats(self):
+        """Print one-time coverage/distribution stats for the freshly built snapshot buffer."""
+        cap = self._cur_buffer_size
+        labels = self.cfg.slot_labels
+        typeable = self._typeable.long()
+        total = int(typeable.numel())
+
+        # Mirror target_key_slot() on the buffered commands: the key each snapshot's arm was posed over.
+        pos = torch.arange(self.max_len, device=self.device)
```
AntoineRichard, line 565: **AI-generated review — Moderate (duplication):** This reimplements `_prefix_len()` for buffered tensors. Extract a single helper that computes prefix length from `(target, typed, target_len, typed_len)` and use it for live state, snapshots, and statistics. Keeping two copies of the state-machine rule makes diagnostic/curriculum behavior drift when one copy changes.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_usd.py
```diff
@@ -0,0 +1,449 @@
...
+        part_root.CreateAttribute("keyboard:partitionName", Sdf.ValueTypeNames.String).Set(part_name)
+
+        part_base_path = f"{part_path}/base_link"
+        part_keys_scope = f"{part_path}/keys"
+        part_joints_scope = f"{part_path}/joints"
+        UsdGeom.Xform.Define(stage, part_base_path)
+        UsdGeom.Scope.Define(stage, part_keys_scope)
+        UsdGeom.Scope.Define(stage, part_joints_scope)
+
+        part_base_prim = stage.GetPrimAtPath(part_base_path)
```
AntoineRichard, line 180: **AI-generated review — Moderate (duplication):** The rigid-body, mass, tiny-inertia, and world-fixed-joint setup duplicates the single-articulation block at lines 127–137. Extract one helper parameterized by base path, joint scope, and mass; keep only the partition-specific policy here. This is physics-authoring code, so a single owner for those authored properties is preferable to parallel branches.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands.py
```diff
@@ -0,0 +1,1121 @@
...
+                ee_pos = self.robot.data.body_pos_w.torch[:, self._ik_body_idx]
+                ee_quat = self.robot.data.body_quat_w.torch[:, self._ik_body_idx]
+                tip = ee_pos + quat_apply(ee_quat, self._ik_offset)
+                reach = torch.linalg.norm(tip - (self.target_key_pos_w() + self._ik_hover), dim=-1)
+                self._buf_reach[start : start + n] = reach[:n]
+                pbar.update(n)
+        self._buffer_built = True
+        self._log_buffer_stats()
+
+    def _log_buffer_stats(self):
```
AntoineRichard, line 557: **AI-generated review — Moderate (production hygiene):** This unconditionally runs roughly 75 lines of diagnostic reductions and `print()` calls during the first reset, including multiple device-to-host synchronizations and verbose histograms. Keep essential metrics through the repository logging/metrics facilities and gate any detailed coverage dump behind an explicit debug option. Production task startup should not emit a debug report by default.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/rewards.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Args:
+        env: The environment instance.
+        command_name: Name of the :class:`LetterTypingCommand` term to read completion from.
+    """
+    command: LetterTypingCommand = env.command_manager.get_term(command_name)  # type: ignore
+    return (command.distance == 0).float()
+
+
+class reach_key(ManagerTermBase):
```
AntoineRichard, line 84: **AI-generated review — Moderate (dead public surface):** Neither `reach_key` nor `typing_mistake` is referenced by this environment configuration, yet both are exported from `mdp/__init__.pyi` and neither is tested. Remove these unused terms and exports, or wire them into a documented behavior with distinct tests. Shipping unused MDP terms expands the public surface and, for `typing_mistake`, contradicts the configured backspace-recovery curriculum.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_geometry.py
```diff
@@ -0,0 +1,527 @@
...
+
+def inset_box(size: tuple[float, float, float], margin: float) -> tuple[float, float, float]:
+    return (
+        max(size[0] - 2.0 * margin, size[0] * 0.25),
+        max(size[1] - 2.0 * margin, size[1] * 0.25),
+        max(size[2] - 2.0 * min(margin, size[2] * 0.2), size[2] * 0.5),
+    )
+
+
+def is_identity_quat(quat_xyzw: tuple[float, float, float, float]) -> bool:
```
AntoineRichard, line 462: **AI-generated review — Moderate (dead code):** `is_identity_quat` has no callers; `quat_to_matrix` and `mat3_rotate` below are also unused. Remove all three rather than carrying an untested parallel math utility layer. Keep only the quaternion helpers actually needed by keyboard generation.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands_cfg.py
```diff
@@ -0,0 +1,240 @@
...
+    """``(min, max)`` target sequence length sampled per episode."""
+
+    max_len: int | None = None
+    """Fixed width [key slots] of the target/typed buffers and the one-hot observations. ``None`` uses
+    ``letter_length[1]``. Set it explicitly (must be ``>= letter_length[1]``) to keep the observation size -
+    and hence a trained policy's input layer - fixed while varying ``letter_length``. For example, evaluate
+    a policy trained with ``letter_length=(1, 2)`` on single letters by keeping ``max_len=2`` and setting
+    ``letter_length=(1, 1)``."""
+
+    command_mode: str = "letter_full"
```
AntoineRichard, line 189: **AI-generated review — Moderate (type design):** Only two values are valid, but the type is unrestricted `str`, and any typo silently takes the `letter_full` branch. Use `Literal["letter_full", "letter_left"]` and validate at construction/config boundaries. Include one negative test that proves invalid modes fail clearly.

AntoineRichard (changes requested):
## AI-generated review — request changes

I reviewed the current head (`ba2bb025e9`) with emphasis on functional comments, duplication, test value, documentation, and strict showroom-quality hygiene. The new inline comments below are additional findings; I did not duplicate the existing precise threads.

The following existing unresolved findings remain blocking on this head: the [standard RSL-RL export path bypasses the custom adapters](https://github.com/isaac-sim/IsaacLab/pull/6876#discussion_r3708936560), [partitioned PhysX reset snapshots index flattened articulation state incorrectly](https://github.com/isaac-sim/IsaacLab/pull/6876#discussion_r3708950715), [the robot uses an internal Nucleus URL](https://github.com/isaac-sim/IsaacLab/pull/6876#discussion_r3708950721), [reset-state helpers depend on private scene collections](https://github.com/isaac-sim/IsaacLab/pull/6876#discussion_r3708950724), and [keyboard randomization is configured twice](https://github.com/isaac-sim/IsaacLab/pull/6876#discussion_r3708950727). The earlier changelog finding is addressed by `so101-keyboard.minor.rst` and can be resolved.

### Tests

This 5.3k-line change currently adds or modifies **no test files**, despite the PR description claiming focused configuration and architecture tests. The existing contributed-environment smoke test should not be duplicated; it is already useful and CI shows it failing for this task with `FileNotFoundError` on the internal SO-101 USD. Add a small consolidated suite with distinct contracts:

- one CPU model/export test that exercises the exact standard JIT and ONNX entry points, not only `as_jit()` / `as_onnx()` in isolation;
- one parametrized typing-state test covering simultaneous presses, backspace, buffer-full behavior, padding, and partial `env_ids`;
- one reset-state round-trip test covering both single-articulation Newton layout and partitioned PhysX layout;
- one parametrized pure geometry/pool test only if the unused general generator remains.

Use shared fixtures and parametrization. Do not add tests that merely restate constants, inspect config fields, or repeat task registration/instantiation already covered by the generic smoke test. Each test should fail for a named behavioral regression.

### Verification

`CUDA_VISIBLE_DEVICES='' ./isaaclab.sh -f` passes all hooks. I did not run local simulator/GPU tests because the machine is running benchmarks and GPU use was explicitly excluded. Current CI has failures, including the task smoke failure above. GitHub also reports the PR as non-mergeable (`mergeable_state: dirty`).

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands.py
```diff
@@ -0,0 +1,1121 @@
...
+        # "distance" metric at reset; stays None (falls back to terminal) when the buffer is disabled.
+        self._buf_avg_distance: float | None = None
+
+        # Reset-pose IK: a differential-IK controller (built once) snaps the arm so its jaw tip hovers
+        # above the first key on every reset. See :meth:`_solve_reset_pose`.
+        self._reset_ik: DifferentialIKController | None = None
+        if cfg.reset.ik is not None:
+            ik_cfg = cfg.reset.ik
+            self.robot: Articulation = env.scene[cfg.asset_name]
+            self._ik_joint_ids, _ = self.robot.find_joints(list(ik_cfg.joint_names))
```
AntoineRichard, line 242: **AI-generated review — important (performance): cache device-native joint indices**

`find_joints()` returns a Python list here, and `_solve_reset_pose()` passes it to `write_joint_position_to_sim_index()` on every IK iteration. The articulation backends resolve a list by constructing a device index array each call, adding allocation/conversion overhead inside this reset hot loop. Request the cached proxy (`as_proxy=True`), store its device-native tensor once (for example, `.torch` because these IDs also index torch data), and derive `_ik_jacobi_joint_ids` as a tensor as well. Asset writes in MDPs should receive pre-allocated tensor/Warp indices—or `None`/a slice for the full selection—rather than Python lists.

AntoineRichard (commented):
## AI-generated review — performance follow-up

I audited the new MDP asset-write paths for reusable indices. The reset-state helpers already receive tensor `env_ids`; the reset IK joint selection is the concrete hot-path violation and is called repeatedly inside the solve loop.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/states.py
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+from collections.abc import Sequence
+
+import torch
+import warp as wp
+from torch import Tensor
+
+
+def get_reset_state(env, env_ids: Tensor, reset_assets: Sequence[str], is_relative: bool = False) -> Tensor:
```
AntoineRichard: **AI-generated review — moderate (duplication/LOC): reuse the shared reset-state utility**

This 62-line module reimplements `core/lift/mdp/utils.py::{get,set}_reset_state`, even though the command already depends on `core.lift.mdp` for `SuccessMonitor`. The copy has already diverged to private scene dictionaries and different root-state APIs, contributing to the partitioned-state correctness issue raised in the existing review. Add the required partition-aware mapping once to the shared utility, import it here, and delete this module instead of maintaining two reset serialization formats.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_schema.py
```diff
@@ -0,0 +1,233 @@
...
+        return tuple(key for key in self.keys if key.active)
+
+    def key_partition(self, slot: int) -> tuple[int, int]:
+        """Return ``(partition_index, local_slot)`` for a global key slot."""
+
+        if self.partition_mode == "fixed_dof":
+            return slot // self.partition_dof, slot % self.partition_dof
+        return 0, slot
+
+    def manifest(self) -> dict[str, Any]:
```
AntoineRichard, line 185: **AI-generated review — moderate (dead code/LOC): remove the unused manifest surface**

Neither `manifest()` nor the preceding `active_keys` property has a caller in this PR or the repository, and the USD authorer writes its metadata independently. They are also not part of the package's declared exports. Delete both methods (and the resulting unused `Any` import), saving roughly 55 lines. If a concrete consumer is added later, introduce the smallest serialization contract it actually needs with a behavioral test.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_styles.py
```diff
@@ -0,0 +1,609 @@
...
+        "cream_classic",
+        "mint_minimal",
+    ),
+}
+
+
+def sample_style(name: str | None, family: str, rng: random.Random) -> KeyboardStyle:
+    """Sample a coherent keyboard style for a layout family."""
+
+    styles = style_archetypes(rng)
```
AntoineRichard, line 112: **AI-generated review — moderate (performance/LOC): instantiate only the selected style**

This builds and jitters all 13 `KeyboardStyle` objects before retaining one. Because `keyboard_pool.py` generates 32 keyboards at import time, it constructs 416 style objects plus all discarded random values. Represent archetypes as compact immutable specs with shared defaults, choose the compatible style name first, and instantiate/jitter only that spec. If the task remains ANSI-full-only, also retain only its supported specs. Fold deterministic seed/output coverage into the single parametrized generator/pool test already requested rather than adding one test per style.

AntoineRichard (commented):
## AI-generated review — simplification follow-up

There is substantial room to reduce this PR without weakening the task. The largest cut remains the earlier inline finding to remove the unused layout/topology framework: the runtime pool is always a 108-key `ansi_full` layout. The earlier comments also cover deleting unused MDP terms/math helpers, consolidating duplicated USD authoring and prefix logic, removing the verbose startup report, shortening configuration prose, and eliminating the duplicated keyboard reset.

The additional inline findings below identify three non-overlapping reductions: reuse the existing reset-state implementation, remove an unused manifest surface, and avoid eagerly constructing every style archetype. Together with the earlier findings, this should remove several hundred lines and reduce the amount of behavior that needs tests.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/mdp/commands/typing_commands.py
```diff
@@ -0,0 +1,1121 @@
...
+    quat_apply,
+    quat_conjugate,
+    quat_from_angle_axis,
+    quat_mul,
+    skew_symmetric_matrix,
+)
+
+from isaaclab_tasks.core.lift.mdp.events import SuccessMonitor
+from isaaclab_tasks.core.lift.mdp.events_cfg import SuccessMonitorCfg
+from isaaclab_tasks.core.lift.mdp.utils import get_reset_state, set_reset_state
```
hujc7, line 33: **Agent found bug:** the partitioned-reset fix moved the bug rather than fixing it. Deleting `mdp/states.py` in favour of the shared helper resolved two of the sub-findings, but not the correctness one.

`core/lift/mdp/utils.py::get_reset_state` reads `articulation.data.joint_pos.torch[env_ids]`, and `joint_pos` is `(num_instances, num_joints)` — not `(num_envs, ...)` (`isaaclab/assets/articulation/base_articulation_data.py:925`). On the default `isaacsim_physx` preset the keyboard has 18 articulation roots per env, as this PR documents at `so101_env_cfg.py:40-42` and as `partition_dof=6` over 108 slots produces.

So `env_ids = [0..k)` selects `part_0..part_{k-1}` of environment 0, and each snapshot stores `13 + 2*6 = 25` floats — 6 of 108 key joints. `set_reset_state` writes them back with the same conflation. The shapes stay self-consistent, so nothing raises.

`reset.enabled=True` (`so101_env_cfg.py:154`) with `reset.ik` set, so the replay curriculum is on by default on the preset this breaks. Newton (1 instance/env, 108 DOF) is unaffected, which is why a Newton-only training run does not surface it.

Two sub-findings **are** fixed by the move, and those threads can be resolved: the shared helper uses the public `scene.articulations` accessor, and it uses `root_link_pose_w` / `write_root_link_pose_to_sim_index` instead of the deprecated `root_state_w` / `write_root_state_to_sim`.

What remains is the partition-aware mapping requested on the old `states.py:13`. `LetterTypingCommand.__init__` already builds `_inst_idx` / `_col_idx` for exactly this layout; the shared helper needs the same awareness, or an explicit per-asset instance-index argument, before either caller can round-trip a multi-root articulation.

- [ ] Reset-state round-trip test over both layouts — Newton single-articulation and partitioned PhysX — asserting `restore(capture(s)) == s` for a non-trivial `env_ids` subset.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,341 @@
...
+            ls_iterations=15,
+            use_mujoco_contacts=False,
+        ),
+        collision_cfg=NewtonCollisionPipelineCfg(),
+        default_shape_cfg=NewtonShapeCfg(),
+        num_substeps=2,
+        debug_mode=False,
+    )
+    physx = PhysxAutoCfg(isaacsim_physx=isaacsim_physx)
+    default = isaacsim_physx
```
hujc7, line 315: **Agent suggestion:** the default physics preset is the one this file advises against, and the one the partitioned-reset issue breaks.

`:288` states "physx is usable but extremely slow for this task and is only for evaluation purposes / for training please only use newton_mjwarp", yet `:315` sets `default = isaacsim_physx`. Anyone omitting `physics=` gets the slow path, which is also the 18-roots-per-env layout the reset snapshot mishandles.

Worth noting the fix is not a one-line flip. `resolve_presets` picks per `PresetCfg` by selected name and otherwise falls back to that preset's own `default` (`isaaclab_tasks/utils/hydra.py:419-420`). With no `physics=` on the command line, `KeyboardAssetCfg` would still resolve to its `default` at `:47` — the partitioned PhysX articulation — while the sim ran Newton. Per this class's own docstring, Newton reads only the first articulation per env, so the other 17 key banks would be invisible. Changing the default means moving `KeyboardAssetCfg.default` to the single-articulation variant at the same time.

If PhysX has to remain the default for checkpoint parity, that is a reasonable call — but please state it in the `PhysicsCfg` docstring. As written, the comment reads as advice against the default the class sets.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/so101_env_cfg.py
```diff
@@ -0,0 +1,341 @@
...
+                "roll": [0.0, 0.75],
+            },
+            "velocity_range": {"x": [-0.0, 0.0], "y": [-0.0, 0.0], "z": [-0.0, 0.0]},
+            "asset_cfg": SceneEntityCfg("keyboard"),
+        },
+    )
+
+
+@configclass
+class SO101ReorientRewardCfg:
```
hujc7, line 259: **Agent review:** `SO101ReorientRewardCfg` names the reorient task, not this one. It reads as a copy-paste from the lift/reorient family and will misdirect anyone grepping for this task's reward config.

```suggestion
class SO101TypingRewardCfg:
```

The reference in `SO101KeyboardEnvCfg.rewards` needs the matching rename.

hujc7 (commented):
**Agent review:** three findings on head `f100f4f`, checked against the existing threads so they do not restate them.

The main one is that deleting `mdp/states.py` in favour of the shared `core/lift` helper looks like it closes the partitioned-PhysX reset thread, but the incorrect indexing came along with the move — details inline. Two sub-findings on that thread (private scene collections, deprecated root-state APIs) genuinely are fixed and can be resolved.

The other two are small: the default physics preset contradicts this file's own guidance, and a reward config class carries the reorient task's name.

docs/source/overview/environments.rst
```diff
@@ -1368,6 +1368,10 @@ including disabling runtime perturbations used for training.
       - Direct
       - **skrl** (AMP)
       -
+    * - IsaacContrib-Keyboard-SO101
+      - Manager Based
+      - **rsl_rl** (PPO)
+      - **physics=** ``isaacsim_physx``, ``newton_mjwarp``, ``physx``
```
hujc7: this seems wrong?

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/__init__.pyi
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+__all__ = ["KeyboardSpawnerCfg", "TYPING_KEYBOARD_POOL"]
+
+from .keyboard_gen_cfg import KeyboardSpawnerCfg
```
StafaH, line 8: absolute imports here and elsewhere

StafaH (approved):
LGTM. The entire keyboard setup section looks extremely complex though, but I don't have any ideas on how to clean it up.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_usd.py
```diff
@@ -0,0 +1,449 @@
...
+from .keyboard_geometry import base_profile_visuals, generate_keyboard, inset_box
+from .keyboard_labels import label_mesh_data
+from .keyboard_schema import KeyboardStyle, ResolvedKey, ResolvedKeyboard
+
+if TYPE_CHECKING:
+    from .keyboard_gen_cfg import KeyboardSpawnerCfg
+
+
+@lru_cache(maxsize=1)
+def _usd_modules():
```
hujc7, line 22: **Agent suggestion:** consider authoring through a class rather than 15 free functions. `TerrainGenerator` (`isaaclab/terrains/terrain_generator.py:32`) is the in-repo pattern for procedural asset generation from a cfg, and this module is the same shape.

The evidence is `_usd_modules()` itself — destructured at 11 call sites, with 28 of the returned bindings discarded:

```python
_, _, UsdGeom, _, _             = _usd_modules()   # 58
_, Sdf, _, _, _                 = _usd_modules()   # 80, 106
Gf, Sdf, UsdGeom, UsdPhysics, _ = _usd_modules()   # 116, 154, 269
Gf, _, _, _, _                  = _usd_modules()   # 336
Gf, _, UsdGeom, _, Vt           = _usd_modules()   # 381, 432
```

An `lru_cache`d 5-tuple re-unpacked in every function is standing in for instance attributes. The same values also thread through the signatures: `stage` ×7, `resolved` ×7, `cfg` ×6, `prim_path` ×5.

```python
class KeyboardAuthor:
    def __init__(self, stage, prim_path, resolved, cfg): ...   # + Gf/Sdf/UsdGeom/UsdPhysics/Vt
    def author(self) -> Usd.Prim: ...
    def _spawn_single(self): ...
    def _spawn_partitioned(self): ...
```

`spawn_keyboard` stays a free function — the spawner protocol needs a callable — and becomes an adapter: `KeyboardAuthor(stage, prim_path, generate_keyboard(cfg), cfg).author()`.

Offered as the structural cause behind the duplication you raised at `:180`, not a second report of it: the rigid-body / mass / tiny-inertia / fixed-joint block is duplicated across the single and partitioned paths largely because there is no receiver to hang the shared step on. With both paths as methods that extraction is a few lines.

Not a blocker — but it should shrink the diff rather than grow it.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_geometry.py
```diff
@@ -0,0 +1,527 @@
...
+
+from .keyboard_layouts import sample_layout
+from .keyboard_schema import BaseVisual, KeyboardStyle, ResolvedKey, ResolvedKeyboard, UnitKey
+from .keyboard_styles import sample_style
+
+if TYPE_CHECKING:
+    from .keyboard_gen_cfg import KeyboardSpawnerCfg
+
+
+def generate_keyboard(cfg: KeyboardSpawnerCfg) -> ResolvedKeyboard:
```
hujc7, line 23: **Agent suggestion:** the same `TerrainGenerator` point for the generation half of this module. `cfg` appears in 5 signatures and `style` in 5, and `_slot_count` / `_partition_slot_count` / `_partition_count` always travel together as one topology decision.

```python
class KeyboardGenerator:
    def __init__(self, cfg: KeyboardSpawnerCfg): ...   # self.cfg, self.style, self.rng
    def generate(self) -> ResolvedKeyboard: ...
```

`generate_keyboard(cfg)` stays the public entry point — `keyboard_pool.py` depends on it being a pure function of `cfg`, which is what lets the pool build and validate 32 variants — and becomes `return KeyboardGenerator(cfg).generate()`.

Scoped deliberately to the generation functions only. The quaternion helpers below (`_quat_from_euler`, `_quat_mul`, `_quat_conj`, `quat_rotate`) should stay free functions: pure math with no shared state. They are a separate concern sharing a file, and the fix there is a module boundary rather than a receiver. The three unused ones are already covered by your `:462` comment.

Not a blocker.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_pool.py
```diff
@@ -0,0 +1,54 @@
...
+from .keyboard_gen_cfg import KeyboardSpawnerCfg
+from .keyboard_geometry import base_profile_visuals, generate_keyboard
+
+_KEYBOARD_DOF = 108
+_KEYBOARD_VARIANTS = 32
+_PARTITION_DOF = 6
+
+
+@dataclass(frozen=True)
+class _TypingKeyboardPool:
```
hujc7, line 21: **Agent suggestion:** `_TypingKeyboardPool` is already the right container — the construction that fills it is what sits outside it. Consider a `build()` classmethod so the pipeline has a body.

Measured on this head (3 warm runs, PR tree extracted standalone): importing this module takes 1.57–1.59 s and retains 90.1 MB. All 32 `ResolvedKeyboard` objects stay in `_KEYBOARDS` for the process lifetime, though only `_KEYBOARDS[0]` is read after validation. Seven scratch names survive as module globals, including the `keyboard` loop variable.

Two secondary effects: the validation `raise` fires during import, so a bad variant breaks `load_cfg_from_registry` and environment-doc generation rather than env construction; and the invariants cannot be exercised without importing a module for its side effects.

```python
@dataclass(frozen=True)
class TypingKeyboardPool:
    ...
    @classmethod
    def build(cls, variants: int = 32, partition_dof: int = 6) -> TypingKeyboardPool: ...
```

Locals die at return, the 108-slot and uniform-trim invariants become directly testable, and `_KEYBOARD_DOF` / `_KEYBOARD_VARIANTS` / `_PARTITION_DOF` become parameters instead of module constants a reader must cross-check against the `raise`.

Separate from your `:31` point about every variant being `ansi_full` — this is about where the construction lives, not what it produces. Not a blocker.

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_usd.py
```diff
@@ -0,0 +1,449 @@
...
+    drive.CreateTargetPositionAttr(0.0)
+    # The authored style spring is often too weak to hold the keycap against gravity, so the cap sags to
+    # the bottom and reads as permanently pressed (and leaves no travel for the robot to press into).
+    # Raise stiffness so the cap rests within ~10% of the top, well above the 50% actuation point. Keep
+    # the spring underdamped (zeta ~0.35) so it actively springs the cap back up with a little bounce on
+    # release instead of just creeping back. The rest pose and the spring-back overshoot both stay above
+    # the actuation point, so releasing a key never re-triggers a keystroke.
+    travel = float(key.upper_limit - key.lower_limit)
+    if travel > 1.0e-6:
+        hold_stiffness = key.mass * 9.81 / (0.1 * travel)
```
hujc7, line 321: **Agent found bug:** `KeyboardStyle.stiffness` and `.damping` never reach the sim.

Measured across all 32 pool variants: **3456/3456 keys (100%)** have the authored stiffness replaced by `hold_stiffness`; `key.damping` is recomputed unconditionally. Ten style archetypes, zero authored gains surviving.

```python
hold_stiffness = key.mass * 9.81 / (0.1 * travel)
stiffness = max(float(key.stiffness), hold_stiffness)   # floor always binds
damping   = 0.7 * (stiffness * key.mass) ** 0.5         # key.damping unused
```

The clamp's reasoning is sound (a weak spring sags past the 50% actuation point). The problem is that `KeyboardStyle` still advertises both as tunables — adjusting key feel there has no effect at any value. Either derive the gains in `keyboard_geometry` so `ResolvedKey` carries what is authored, or rename the field `min_stiffness` and drop `damping`, noting it follows a fixed damping ratio (ζ ≈ 0.35).

Wider point, take or leave: same shape as the press axis at `:302` and the unconditional collision box at `:290`. The three properties that decide how a pressable behaves live in this function rather than on `ResolvedKey`, so the record only looks like a complete element spec. Moving `axis`, collision shape and the resolved gains onto it would also make the element reusable outside keyboards — a machine panel with a few buttons is the same physics, and the runtime press detection is already axis-agnostic. Complementary to the `KeyboardAuthor` note at `:22`.

Not a blocker.

hujc7 (commented):
Not exact understanding the construction but i think overall the keyboard construction can be better structured so there's more reuse opportunity?

source/isaaclab_tasks/isaaclab_tasks/contrib/keyboard/keyboards/keyboard_pool.py
ooctipus: no this is not isaaclab convention

docs/source/overview/environments.rst
ooctipus: its correct

Issue #6873, opened 2026-08-04
[Bug Report] MPMObject particles are never emitted under CouplerProxyCfg because the coupler skips the per-world builder hooks


hong-nvidia:
### Describe the bug

An `MPMObject` in a scene whose physics is a `CouplerProxyCfg` never gets its particles
into the Newton model. `model.particle_count` stays `0` and the run dies during
`sim.reset()`.

The cause is an inherited builder path. `NewtonCouplerManager` extends
`NewtonVBDManager`, which overrides `instantiate_builder_from_stage`
(`source/isaaclab_contrib/isaaclab_contrib/deformable/vbd_manager.py:127`). That override
walks only the deformable registry and never runs `cls._per_world_builder_hooks`, which
`NewtonManager.instantiate_builder_from_stage` does at
`source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py:1787`.

`MPMObject` publishes its particle emitter through exactly that hook list
(`source/isaaclab_newton/isaaclab_newton/assets/mpm_object/mpm_object.py:105-106`):

```python
if add_registered_mpm_objects_to_builder not in SimulationManager._per_world_builder_hooks:
    SimulationManager._per_world_builder_hooks.append(add_registered_mpm_objects_to_builder)
```

So any manager that inherits the VBD builder path silently drops every MPM object. The
same `MPMObject` works correctly under the standalone `MPMSolverCfg` path, which confirms
the asset is configured properly and the coupler is what differs.

This blocks coupling implicit MPM to any other solver through
`isaaclab_contrib.coupling`, which appears to be the intended way to build MPM/rigid
multiphysics scenes.

### Steps to reproduce

Minimal example: a single MPM entry, one `MPMObject`, no rigid bodies and no proxies.

```python
import argparse

parser = argparse.ArgumentParser()
parser.add_argument("--device", type=str, default="cuda:0")
parser.add_argument("--standalone", action="store_true", help="Use MPMSolverCfg directly instead of the coupler.")
args_cli = parser.parse_args()

import isaaclab.sim as sim_utils
from isaaclab.sim import SimulationContext
from isaaclab_contrib.coupling import CouplerEntryCfg, CouplerProxyCfg
from isaaclab_newton.assets.mpm_object import MPMObject, MPMObjectCfg
from isaaclab_newton.physics import MPMSolverCfg, NewtonCfg, NewtonManager
[...]

PR #6878, opened 2026-08-04
Fix coupled MPM reset mask handling

### LINKED ISSUE #6874: [Bug Report] Coupled solver reset forwards Isaac Lab's per-world mask to SolverImplicitMPM, which rejects it (ValueError: world_mask has shape (1,), expected (2,))
This bug can only be reproduced after [this](https://github.com/isaac-sim/IsaacLab/issues/6873) issue is fixed.

### Describe the bug

A scene whose physics is a `CouplerProxyCfg` containing an `MPMSolverCfg` entry raises on
the first `sim.step()`:

```
ValueError: world_mask has shape (1,), expected (2,).
```

`NewtonManager._reset_solver_internals`
(`source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py:1946`) calls
`cls._solver.reset(cls._state_0, world_mask=world_mask, flags=0)`. When the solver is a
`SolverCoupled`, that forwards the same mask to every entry, and
`SolverImplicitMPM.reset` only honors a per-world mask when it runs one FEM environment
per world, so it rejects the mask Isaac Lab builds.

`NewtonMPMManager` already handles this for the standalone case by overriding
`_reset_solver_internals` as a no-op
(`source/isaaclab_newton/isaaclab_newton/physics/mpm_manager.py:149`), with a docstring
that describes this exact incompatibility. `NewtonCouplerManager` does not inherit that
specialization, so a coupled MPM entry hits the very problem `NewtonMPMManager` documents.

There is also a design question behind the crash. `SolverCoupled.reset` resets all entries
together, so even with the mask problem solved there is no way to reset only the rigid
entry of a coupled model. As it stands, a coupled model containing an MPM entry cannot
support per-environment resets at all, which matters for RL use.

### Steps to reproduce

Minimal example: a single MPM entry, one `MPMObject`, no rigid bodies and no proxies.

Reaching this issue requires working around a separate bug first — under
`CouplerProxyCfg` the `MPMObject` particles are never emitted, because the coupler
inherits `NewtonVBDManager.instantiate_builder_from_stage`, which skips the per-world
builder hooks. That is filed separately; the `NewtonPatchedCouplerManager` subclass below
contains only that workaround so this issue can be reached.

```python
import argparse

parser.add_argument("--device", type=str, default="cuda:0")
parser.add_argument("--standalone", action="store_true", help="Use MPMSolverCfg directly instead of the coupler.")

from isaaclab.sim import SimulationContext
from isaaclab_contrib.coupling import CouplerEntryCfg, CouplerProxyCfg, NewtonCouplerManager
from isaaclab_newton.assets.mpm_object import MPMObject, MPMObjectCfg
from isaaclab_newton.physics import MPMSolverCfg, NewtonCfg
from isaaclab_newton.physics.newton_manager import NewtonManager
from isaaclab_newton.sim.spawners.mpm import MPMGridCfg, MPMParticleMaterialCfg

VOXEL_SIZE = 0.05


class NewtonPatchedCouplerManager(NewtonCouplerManager):
    """Coupler carrying only the builder-hook workaround, so this issue can be reached.

    The ``Newton`` prefix is required: Isaac Lab derives the asset backend from the
    physics manager's class name in ``FactoryBase._get_backend``.
    """

    @classmethod
    def instantiate_builder_from_stage(cls):
        NewtonManager.instantiate_builder_from_stage.__func__(cls)


def main() -> None:
    if args_cli.standalone:
        # Control: the same scene under the standalone MPM manager.
        solver_cfg = MPMSolverCfg(voxel_size=VOXEL_SIZE, grid_type="sparse")
    else:
        solver_cfg = CouplerProxyCfg(
            entries=[
                CouplerEntryCfg(
                    name="gravel",
                    solver_cfg=MPMSolverCfg(voxel_size=VOXEL_SIZE, grid_type="sparse"),
                    all_particles=True,
                    in_place=True,
                    include_static_shapes=True,
                )
            ]
        )
        solver_cfg.class_type = NewtonPatchedCouplerManager

    sim_cfg = sim_utils.SimulationCfg(
        dt=1.0 / 100.0,
        device=args_cli.device,
        physics=NewtonCfg(solver_cfg=solver_cfg),
        visualizer_cfgs=[],
    )
    sim = SimulationContext(sim_cfg)

    ground_cfg = sim_utils.GroundPlaneCfg()
    ground_cfg.func("/World/defaultGroundPlane", ground_cfg)
    sim_utils.create_prim("/World/Env_0", "Xform", translation=(0.0, 0.0, 0.0))

    bed = MPMObject(
        MPMObjectCfg(
            prim_path="/World/Env_.*/Gravel",
            spawn=MPMGridCfg(
                lower=(-0.4, -0.4, 0.0),
                upper=(0.4, 0.4, 0.2),
                voxel_size=VOXEL_SIZE,
                particles_per_cell=3.0,
                material=MPMParticleMaterialCfg(
                    density=2500.0,
                    friction=0.8,
                    young_modulus=1.0e15,
                    viscosity=1.0e4,
                    damping=0.02,
                    yield_stress=1.0e6,
                    yield_pressure=1.0e7,
                ),
            ),
        )
    )

    sim.reset()
    print(f"[RESULT]: reset ok, particles = {bed.num_instances * bed.particles_per_object}")
    for step in range(3):
        sim.step()
        print(f"[RESULT]: step {step} ok")


if __name__ == "__main__":
    main()
```

Run it through the coupler and the first step fails:

```
python repro_mpm_coupler_reset.py
```

```
Traceback (most recent call last):
  File "repro_mpm_coupler_reset.py", line 90, in main
    sim.step()
  File ".../isaaclab/isaaclab/sim/simulation_context.py", line 680, in step
    self.physics_manager.step()
  File ".../isaaclab_newton/isaaclab_newton/physics/newton_manager.py", line 926, in step
    cls._reset_solver_internals_delegate(cls._world_reset_mask)
  File ".../isaaclab_newton/isaaclab_newton/physics/newton_manager.py", line 1961, in _reset_solver_internals
    cls._solver.reset(cls._state_0, world_mask=world_mask, flags=0)
  File ".../newton/_src/solvers/coupled/solver_coupled.py", line 1934, in reset
    entry.solver.reset(entry.state_0, world_mask=world_mask, flags=flags)
  File ".../newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py", line 1577, in reset
    self._validate_reset_inputs(state, world_mask)
  File ".../newton/_src/solvers/implicit_mpm/solver_implicit_mpm.py", line 1453, in _validate_reset_inputs
    raise ValueError(f"world_mask has shape {world_mask.shape}, expected {expected_shape}.")
ValueError: world_mask has shape (1,), expected (2,).
```

Run the same scene under the standalone MPM manager and it steps fine, because
`NewtonMPMManager._reset_solver_internals` is a no-op there:

```
python repro_mpm_coupler_reset.py --standalone
```

```
[RESULT]: reset ok, particles = 31213
[RESULT]: step 0 ok
[RESULT]: step 1 ok
[RESULT]: step 2 ok
```

This is with a single environment, so it is not specific to a multi-environment layout.

### System Info

- Commit: `05f68ac3621ff2ae8dbdc2eec74cc4d7cae1a686` (`develop`, `perf-2026-07-06-315-g05f68ac36`)
- Isaac Sim Version: not used — kit-less Isaac Lab with the Newton backend
  (`isaaclab` 15.1.1, `isaaclab_newton` 2.4.3, `isaaclab_contrib` 1.0.0,
  `newton` 1.5.0.dev0 @ `831076f5`)
- OS: Ubuntu 24.04.4 LTS
- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition
- CUDA: 13.0
- GPU Driver: 580.173.02

### Additional context

`NewtonMPMManager._reset_solver_internals` already spells out why a masked reset cannot
work for implicit MPM (`mpm_manager.py:150-161`):

> :meth:`SolverImplicitMPM.reset` only honors a per-world mask when the solver runs one
> FEM environment per world (``Config.separate_worlds``). With the shared topology Isaac
> Lab builds, a masked reset is rejected, and a full reset would clear the particle
> history of environments that were not reset.

Suppressing the reset in a local subclass is enough to get a coupled MPM scene stepping,
mirroring what `NewtonMPMManager` does:

```python
class NewtonMPMCouplerManager(NewtonCouplerManager):
    @classmethod
    def _reset_solver_internals(cls, world_mask) -> None:
        """Skip the solver-internal reset, as NewtonMPMManager does."""
```

That is fine as a stopgap for a scene that never resets, but it also suppresses the reset
for the rigid entry, which is why the per-entry question above seems worth a decision
rather than just inheriting the no-op. A per-entry mask, or letting `SolverCoupled.reset`
target named entries, would let the rigid entry reset normally while MPM keeps its
particle history.

### Checklist

### Acceptance Criteria

- [ ] The minimal example above steps without raising, with no local
      `_reset_solver_internals` workaround.
- [ ] A coupled model containing an MPM entry can reset its rigid entry per environment
      without either raising or clearing MPM particle history.
- [ ] The behaviour is consistent between the standalone `MPMSolverCfg` path and an MPM
      entry under `CouplerProxyCfg`.
- [ ] A regression test resets a coupled model containing an MPM entry.

hong-nvidia:
## Summary

- Reset compatible entries in coupled Newton solvers while skipping implicit MPM entries.
- Preserve MPM particle history while retaining per-world reset behavior for rigid entries.
- Add regression coverage for a coupled rigid/MPM configuration.

## Root cause

`SolverCoupled.reset` forwards Isaac Lab's per-world reset mask to every nested solver. `SolverImplicitMPM` rejects that mask for shared topology, which causes the coupled scene to fail on its first step.

Fixes #6874

## Validation

- `uv run --extra test python -m pytest source/isaaclab_contrib/test/coupling/test_coupler.py` (50 passed)
- `uv run isaaclab -f`

## Checklist

- [x] I have read and understood the contribution guidelines.
- [x] I have run the pre-commit checks.
- [ ] I have made corresponding changes to the documentation.
- [x] I have added tests that prove my fix is effective.
- [ ] I have added a changelog fragment for the touched package (intentionally omitted by request).
- [x] I have added my name to `CONTRIBUTORS.md` or it is already present.

### DIFF
--- a/source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
+++ b/source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from functools import partial
 
+import warp as wp
 from isaaclab_newton.physics import (
     KaminoSolverCfg,
     MJWarpSolverCfg,
@@ -208,6 +209,41 @@ class NewtonCouplerManager(NewtonVBDManager):
         super()._initialize_contacts()
         if cls._contacts is not None and hasattr(NewtonManager._solver, "prepare_contacts"):
             NewtonManager._solver.prepare_contacts(cls._contacts)
+
+    @classmethod
+    def _reset_solver_internals(cls, world_mask: wp.array | None) -> None:
+        """Reset compatible coupled entries while preserving implicit MPM history.
+
+        :meth:`SolverCoupled.reset` applies the same world mask to every
+        nested solver. :class:`SolverImplicitMPM` rejects Isaac Lab's
+        per-world mask for its shared topology, and resetting it without a
+        mask would also clear particle history for worlds that did not reset.
+        Reset the remaining entries through the coupled solver's reset
+        lifecycle so their per-world solver state and coupled transient state
+        remain synchronized.
+
+        Args:
+            world_mask: Per-world reset mask, or ``None`` when no simulation
+                state is available.
+        """
+        if world_mask is None:
+            return
+
+        solver = NewtonManager._solver
+        solver._distribute_state(cls._state_0, iteration_restart=True)
+        solver_cfg = PhysicsManager._cfg.solver_cfg
+        for entry_cfg in solver_cfg.entries:
+            if isinstance(entry_cfg.solver_cfg, MPMSolverCfg):
+                continue
+            entry = solver._entries[entry_cfg.name]
+            entry.solver.reset(entry.state_0, world_mask=world_mask, flags=0)
+            solver._sync_entry_reset_state(entry)
+
+        solver._reconcile_state(cls._state_0)
+        solver._reset_coupling_state(cls._state_0, world_mask=world_mask, flags=0)
+        solver._clear_entry_contact_buffers()
+        solver._rebuild_entry_solver_state_caches()
+        solver._entry_output_state_valid = False
 
     @classmethod
     def _supports_cuda_graph_capture(cls) -> bool:
--- a/source/isaaclab_contrib/test/coupling/test_coupler.py
+++ b/source/isaaclab_contrib/test/coupling/test_coupler.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 from types import SimpleNamespace
+from unittest.mock import MagicMock
 
 import numpy as np
 import pytest
@@ -31,6 +32,8 @@ from isaaclab_newton.physics.newton_manager import NewtonManager
 from newton import ShapeFlags
 from newton.solvers.experimental.coupled import SolverCoupledADMM, SolverCoupledProxy
 
+from isaaclab.physics import PhysicsManager
+
 from isaaclab_contrib.coupling import (
     CouplerAdmmCfg,
     CouplerCfg,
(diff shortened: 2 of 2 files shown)


maxkra15:
seems to be a duplicate of #6877

hong-nvidia:
> seems to be a duplicate of #6877

You're right. I told my agent to file a PR for each of #6873 and #6874 , and I'm at fault for not catching the duplication

PR #6881, opened 2026-08-04
Add physics variant selection to asset converters

hujc7:
## Summary

The URDF and MJCF importers emit their physics description as payloads behind a `"Physics"` variant set on the generated asset, and leave that variant set unselected. USD has no implicit default, so an unselected variant set contributes nothing to composition and the converted USD opens with geometry only — no joints, no articulation root, no mass properties.

The converters already author a `"physx"` selection to compensate, but return early when the asset offers no such variant. A URDF whose joints are all fixed yields `["none", "physics"]` and no `"physx"`, so those assets still convert physics-free.

## What changed

- `AssetConverterBaseCfg.physics_variant` (new, defaults to `"physics"`) chooses which variant the converter selects on the generated USD file.
- Conversion raises when the asset does not offer the requested variant, naming the ones it does. Substituting another variant would hand back an asset authored for a different backend than the caller asked for.
- Fixed the standalone importer install command in the import how-to. The wheel is published on PyPI, so the package index substitution it described is unnecessary.

Every variant stays in the generated USD file, so the selection is not destructive — it only decides what composes when the asset is opened without saying anything. `UsdFileCfg.variants` still overrides it at spawn time.

## Why `"physics"` is the default

The variants are a shared base plus two supersets: `physx.usda` and `mujoco.usda` both carry `subLayers = [@./physics.usda@]`. Measured on the converted Franka:

| variant | what it adds over the base |
|---|---|
| `physics` | this *is* the base — `UsdPhysics` joints / articulation root / mass, plus `NewtonMimicAPI`, `NewtonArticulationRootAPI`, `NewtonCollisionAPI` |
| `physx` | `PhysxJointAPI` and `physxJoint:maxJointVelocity` per joint |
| `mujoco` | one `MjcActuator` prim per joint |

The base is the backend-portable layer by construction: the importer routes schemas matching `Physics.*` and `Newton.*` into it, `Physx.*` into the PhysX variant, and `Mjc.*` into the MuJoCo variant. Everything structural lives in the base — the importer no longer authors PhysX mimic joints or articulation roots — so the backend variants carry solver tuning only.

Defaulting to `"physics"` therefore produces an asset that runs on either backend, and leaves the backend choice to the caller who knows which one they want. Selecting `"physx"` or `"mujoco"` bakes that choice into the asset, which is the right thing to do explicitly and the wrong thing to do by default.

## Testing

Isaac Sim 6.1.0-alpha.35, Kit 110.3.0:

```
./isaaclab.sh -p -m pytest source/isaaclab/test/sim/test_urdf_converter.py -q
```

| Run | Result |
|---|---|
| With this change | 23 passed (20 existing + 3 new) |
| With the converter change reverted | `test_physics_variant_override` and `test_physics_variant_raises_when_requested_absent` fail |

The new tests cover the default selection, an explicit `"mujoco"` override, and the failure raised for an asset that offers no `"physx"` variant.

Tracked as nvbug 6527494.


hujc7:
Folding this into #6935. That PR already contains these commits, and splitting the review across two PRs was costing more than it saved — the physics variant selection and the importer packaging both hinge on the same question of which importer serves a conversion.

Everything here is in #6935 unchanged:
- `physics_variant` on the converter cfg, selecting the variant on the generated USD
- `REQUIRED_VARIANT_SETS` in `select_usd_variants`, so a missing or unoffered `"Physics"` variant raises instead of silently composing nothing
- the enum round-trip fix in `class_to_dict` / `update_class_from_dict`

Closing in favour of #6935.

PR #6883, opened 2026-08-04
Add OSMO workflow for multi-GPU cluster training

ooctipus:
# Description

Isaac Lab's cluster tooling currently only covers SLURM and PBS via singularity. Both paths require converting the Docker image to a `.sif` and copying the source tree to the cluster on every submission.

This PR adds an [NVIDIA OSMO](https://developer.nvidia.com/osmo) workflow spec at `docker/cluster/multi_gpu.yaml`. OSMO runs the published Isaac Lab container image directly, so there is no singularity conversion and no code copy step — the image is the unit of deployment.

The workflow requests a single node and scales training across that node's GPUs through the existing `train_multigpu` command. Everything site-specific is a submission-time parameter (`image`, `rl_library`, `num_gpu`, `num_cpu`, `memory`, `storage`, `platform`, `master_port`, `args`), so the file carries no cluster- or user-specific defaults. When `num_gpu` is `1` it falls back to the single-process `train` entry point, so one file covers both cases.

Submitting a 4-GPU run:

```bash
osmo workflow submit docker/cluster/multi_gpu.yaml \
  --set num_gpu=4 \
  --set args="--task Isaac-Reorient-KukaAllegro --num_envs 4096"
```

The container startup script does two things beyond launching training:

- Hides the system Vulkan ICDs when the image ships its own, which otherwise makes the loader pick up a conflicting driver on the compute node.
- Splits the allocated CPUs across the per-GPU workers via `OMP_NUM_THREADS`, so each worker does not size its thread pool for the whole node.

The spec is a Jinja template and only becomes valid YAML after parameter substitution, so it is excluded from the `check-yaml` pre-commit hook. Rendering was verified for both the `num_gpu=1` and `num_gpu>1` branches.

Fixes # (issue)

## Type of change

- Documentation update

## Checklist

<sub>No changelog fragment is included because this PR does not touch any package under `source/`; `tools/changelog/cli.py check` passes as-is. No tests are included because the change is a cluster job spec plus documentation.</sub>

### DIFF
--- a/docker/cluster/multi_gpu.yaml
+++ b/docker/cluster/multi_gpu.yaml
@@ -1,0 +1,81 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# NVIDIA OSMO workflow for single-node, multi-GPU Isaac Lab training.
+#
+# Submit with:
+#   osmo workflow submit docker/cluster/multi_gpu.yaml \
+#     --set num_gpu=4 \
+#     --set args="--task Isaac-Reorient-KukaAllegro --num_envs 4096"
+#
+# When ``num_gpu`` is 1 the workflow falls back to the single-process ``train``
+# entry point, so the same file covers both cases. No visualizer is requested,
+# so training runs headless.
+
+workflow:
+  name: isaac-lab
+  resources:
+    default:
+      gpu: {{ num_gpu }}
+      cpu: {{ num_cpu }}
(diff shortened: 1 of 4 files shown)


kellyguo11 (approved):

docker/cluster/multi_gpu.yaml
```diff
@@ -0,0 +1,81 @@
...
+            --master_port {{ master_port }} \
+            {{ args }}
+          {% else %}
+          /workspace/isaaclab/isaaclab.sh train \
+            --rl_library {{ rl_library }} \
+            {{ args }}
+          {% endif %}
+
+default-values:
+  image: nvcr.io/nvidia/isaac-lab:latest
```
kellyguo11: we probably need to update this based on the released docker image tag - likely 3.0.0

PR #6889, opened 2026-08-04
[Fix] Stop OvPhysX writing its collider cache into the Python interpreter directory

hujc7:
## 1. Summary

- Kitless `physics=ovphysx` runs wrote their UJITSO derived-data cache into the **directory holding the Python interpreter**. The location is now a config field defaulting to a per-user directory under the system temp directory, created owner-only.
- Fixes nvbug 6537980: where that directory is not writable (uv reusing a system `/usr/bin/python3.12`), every run logged two `[Error] [omni.datastore]` lines.
- Where it *was* writable there was no error at all, and the cache was silently written inside the Python installation — shared by every project using that interpreter, and discarded on an interpreter upgrade.

## 2. Root cause

The `ovphysx` wheel bundles its own Carbonite, which resolves the UJITSO cache from `realpath("/proc/self/exe")` + `/cache/DerivedDataCache` when no path is configured.

That assumption — *the executable's directory belongs to me* — holds for Kit, whose `kit` binary lives in its own application package next to a `cache/` sibling. It breaks for every embedder, where the "executable" is a shared Python interpreter whose directory the program does not own.

## 3. Fix

`OvPhysxCfg.cooked_collider_cache_dir` selects the directory and `_create_physx_instance` passes it to `ovphysx.PhysXConfig(cooked_collider_cache_dir=...)`, mirroring how the sibling kitless renderer threads `OVRTXRendererCfg.log_file_path` into `RendererConfig`.

The field is `str | None`: `None` is forwarded unchanged so OVPhysX applies its own resolution, which `PhysXConfig` has accepted since the pinned 0.5.9.

The default is `<temp>/ovphysx_derived_data_cache_<uid>`. Because that lives in a world-writable root, `_prepare_default_cache_dir` creates it with `os.makedirs(..., mode=0o700)` and, when the path already exists, `os.lstat`-rejects a symlink, a non-directory, or a directory owned by another user — a symlink planted there is otherwise followed and written through. A *configured* directory is passed through untouched, so a site can still point the field at a provisioned shared cache.

## 4. Validation

```
uv run isaaclab train --rl_library rsl_rl --task Isaac-Cartpole-Camera-Direct \
    physics=ovphysx renderer=ovrtx presets=rgb --num_envs 32 --max_iterations 2
```

Run three ways, varying only the fix and whether the interpreter directory is writable:

| Arm | Fix | Interpreter dir | `omni.datastore` errors | Cache written to |
|---|---|---|---|---|
| A0 | reverted | writable | 0 | **interpreter's `bin/cache/DerivedDataCache`** |
| A1 | reverted | read-only | **2** | not created |
| B | applied | read-only | **0** | the configured directory |

Arm A1 reproduces both reported lines verbatim. Arm A0 shows this is a wrong-location bug rather than a permissions bug: no error, cache still in the wrong place. All three arms exit 0 with training completing.

## 5. Test

- `test_manager_supports_pinned_runtime_api` — the cache assertion and a strict `PhysXConfig` fake fold into the existing CPU/GPU parametrization; both fail when the keyword is removed.
- `test_construct_physx_passes_the_configured_cooked_collider_cache_dir` — pins the config-to-runtime seam, with a no-config fallback case. Verified to fail when `_construct_physx` stops reading the config.
- `test_construct_physx_forwards_an_unset_cooked_collider_cache_dir` — `None` reaches `PhysXConfig` unchanged.
- `test_default_cache_dir_is_created_owner_only`, `..._rejects_a_planted_symlink`, `..._rejects_a_directory_owned_by_another_user`.

## 6. Notes

- Earlier revisions also configured a pre-0.5.9 legacy runtime. Develop removed that code in #6074, so this branch targets the pinned runtime only.
- The default path is per-user but predictable, so a TOCTOU window remains between the ownership check and OVPhysX opening the path. A home-rooted parent would close it structurally at the cost of the temp-rooted convention the ov packages already use.
- The datastore is created once per process, so the first OVPhysX construction fixes the cache location.
  The runtime garbage collects the cache against a disk budget; deleting the directory reclaims it immediately.


source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -1011,10 +1029,11 @@ def _create_physx_instance(ovphysx: Any, ovphysx_device: str, gpu_index: int) ->
         if "active_cuda_gpus" in physx_parameters and ovphysx_device == "gpu":
             physx_kwargs["active_cuda_gpus"] = str(gpu_index)
             physx_kwargs["config"] = ovphysx.PhysXConfig(
```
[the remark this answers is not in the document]
hujc7: Confirmed and fixed in 48e4935cbef.

The legacy branch now builds `PhysXConfig` unconditionally, so legacy CPU runs and the `gpu_index` constructor both receive the cache setting. The GPU-only `/physics/suppressReadback` and `/physics/suppressFabricUpdate` keys stay conditional, since those are correctly GPU-gated.

One correction worth recording: on that path the typed field was not merely omitted, it was unsafe. The declared runtime`'s `PhysXConfig` predates `cooked_collider_cache_dir` (verified against an older `ovphysx` wheel: `carbonite_overrides` present, the typed field absent), so `cooked_collider_cache_dir=` would have raised `TypeError` on legacy GPU startup. The declared runtime now goes through the generic `/UJITSO/datastore/localCachePath` override instead, which it accepts and which 0.5.9 rejects as conflicting with its typed field. The two routes are mutually exclusive by wheel version.

Test coverage went from 2 to 6 combinations (two legacy constructors x two devices, plus the current runtime x two devices), and each stubbed `PhysXConfig` now mirrors its runtime`'s real signature, so a keyword the real runtime would reject fails the test instead of being absorbed by a permissive `lambda **kwargs`.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -1011,10 +1029,11 @@ def _create_physx_instance(ovphysx: Any, ovphysx_device: str, gpu_index: int) ->
...
+                cooked_collider_cache_dir=OvPhysxManager._cooked_collider_cache_dir(),
```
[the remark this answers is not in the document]
hujc7: Confirmed and fixed in 48e4935cbef, along the lines you suggested.

The cache-bearing config is now built unconditionally on the legacy path, with the GPU selectors (`active_cuda_gpus` / `gpu_index`) and the GPU-only `/physics/suppress*` overrides kept conditional:

```python
legacy_overrides = {"/UJITSO/datastore/localCachePath": OvPhysxManager._cooked_collider_cache_dir()}
if "active_cuda_gpus" in physx_parameters and ovphysx_device == "gpu":
    physx_kwargs["active_cuda_gpus"] = str(gpu_index)
    legacy_overrides["/physics/suppressReadback"] = True
    legacy_overrides["/physics/suppressFabricUpdate"] = True
elif "gpu_index" in physx_parameters:
    physx_kwargs["gpu_index"] = gpu_index
physx_kwargs["config"] = ovphysx.PhysXConfig(carbonite_overrides=legacy_overrides)
```

One deviation from the suggestion: the declared runtime cannot take the typed field at all. Its `PhysXConfig` predates `cooked_collider_cache_dir` (verified against an older `ovphysx` wheel), so the typed keyword would raise `TypeError` there rather than merely being skipped. It therefore uses the generic `/UJITSO/datastore/localCachePath` override, while the current runtime keeps the typed field, which is the only route it accepts (0.5.9 rejects the generic key as conflicting with the typed field).

The regression test was extended as you asked, to two legacy constructors x two devices plus the current runtime x two devices, and its stubs now mimic each runtime`'s real `PhysXConfig` signature so this class of mistake fails the test.

hujc7:
@greptileai review

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -967,6 +967,28 @@ def _atexit_release_and_exit():
...
+            cache_dir = os.path.join(base, "ov", "cache", "DerivedDataCache")
+        else:
+            base = os.environ.get("XDG_CACHE_HOME") or os.path.join(os.path.expanduser("~"), ".cache")
+            cache_dir = os.path.join(base, "ov", "DerivedDataCache")
+        if not os.path.isabs(cache_dir):
+            # ``expanduser`` leaves "~" unchanged when the home directory cannot be resolved
+            # (e.g. an arbitrary-UID container with no passwd entry); a relative path would
+            # put the cache under the working directory.
+            cache_dir = os.path.join(tempfile.gettempdir(), "ov", "DerivedDataCache")
+        return cache_dir
```
AntoineRichard: Should we use this, if IsaacSim is not installed does this even exists?

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
```diff
@@ -967,6 +967,28 @@ def _atexit_release_and_exit():
...
+            base = os.environ.get("LOCALAPPDATA") or os.path.join(os.path.expanduser("~"), "AppData", "Local")
+            cache_dir = os.path.join(base, "ov", "cache", "DerivedDataCache")
+        else:
+            base = os.environ.get("XDG_CACHE_HOME") or os.path.join(os.path.expanduser("~"), ".cache")
+            cache_dir = os.path.join(base, "ov", "DerivedDataCache")
+        if not os.path.isabs(cache_dir):
+            # ``expanduser`` leaves "~" unchanged when the home directory cannot be resolved
+            # (e.g. an arbitrary-UID container with no passwd entry); a relative path would
+            # put the cache under the working directory.
+            cache_dir = os.path.join(tempfile.gettempdir(), "ov", "DerivedDataCache")
```
AntoineRichard: 🤖 **AI-generated review:** This fallback uses the same predictable `<temp>/ov/DerivedDataCache` directory for every user. On a shared Linux host, the first user to create `/tmp/ov` will normally own it, so another arbitrary-UID process may be unable to write the cache and hit the same datastore failure this branch is intended to avoid; another user can also pre-populate that cache. Could we use a securely created temporary directory (for example, `tempfile.mkdtemp(...)`) or another safely per-user location here?

source/isaaclab_ovphysx/test/physics/test_ovphysx_scene_data_backend.py
```diff
@@ -434,6 +434,78 @@ def wait_op(self, operation):
     assert physx.calls == [("step_sync", 0.02), ("reset_stage",), ("wait_op", 23)]
 
 
+@pytest.mark.skipif(sys.platform == "win32", reason="cache root is LOCALAPPDATA-derived on Windows")
+def test_manager_passes_explicit_derived_data_cache_dir(monkeypatch, tmp_path):
```
AntoineRichard: 🤖 **AI-generated review:** The regression is worth covering: the neighboring runtime-API tests do not assert the cache setting, and a focused test would have caught the earlier legacy-CPU omission. The full 3 × 2 matrix seems larger than the distinct behavior, though, while the decorator skips the only platform-specific branch added by this PR. Could we fold strict `PhysXConfig` fakes and cache assertions into the existing current/legacy API tests, retain focused cases for legacy CPU and `gpu_index`, and make the expectation conditional on `LOCALAPPDATA` versus `XDG_CACHE_HOME` instead of skipping Windows? That preserves the useful regression coverage with less duplicated fixture code.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
hujc7: Fixed in 703f20cc298. The path is now a public config field, `OvPhysxCfg.cooked_collider_cache_dir`, defaulting to `<system temp>/ovphysx_derived_data_cache_<uid>`.

Measured with the 0.5.9 wheel: the runtime creates the directory 0755 owned by whoever runs first, so a fixed shared name really does lock out the second UID — with the same `omni.datastore` error this PR removes. Windows has no `getuid` and its temp root is already per-user, so the suffix applies only where it exists.

Not `mkdtemp`: it returns a fresh directory per process, so the cache would never be reused and the directories would accumulate. Caveat worth stating — per-user is not *securely created*; the path is predictable. Sites needing a hardened or shared location set the config field.

source/isaaclab_ovphysx/test/physics/test_ovphysx_scene_data_backend.py
hujc7: Done in 703f20cc298. The skip is gone because the platform branch is gone: the directory is injected into `_create_physx_instance` now, not derived from `LOCALAPPDATA`/`XDG_CACHE_HOME`.

The matrix collapsed further than proposed — develop removed the pre-0.5.9 runtime in #6074, so two of the three runtimes no longer exist. The cache assertion and a strict `PhysXConfig` fake fold into develop's `test_manager_supports_pinned_runtime_api` as you suggested, replacing its `lambda **kwargs` stub. Removing the keyword fails both parametrizations, verified.

Your point generalised: `test_ovphysx_manager_lifecycle`'s fake had the opposite drift — it rejected a keyword the real wheel accepts — and now carries the field too.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
hujc7: Moot as of 703f20cc298 — the code no longer targets an Omniverse-managed location at all. The directory is a config field defaulting under the system temp directory, so nothing depends on an Isaac Sim install having created anything.

On the underlying question: measured against the real 0.5.9 wheel, pointed at an empty temp root, `_create_physx_instance` created the whole cache tree from nothing.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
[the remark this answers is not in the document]
hujc7: Follow-up: develop removed the legacy runtime in #6074, so the branch this thread is about no longer exists. `_create_physx_instance` has one path now — the pinned 0.5.9 runtime — taking the cache directory as an argument from `OvPhysxCfg.cooked_collider_cache_dir`. The legacy `PhysXConfig` and `/UJITSO/datastore/localCachePath` override from my earlier reply are gone with it.

source/isaaclab_ovphysx/isaaclab_ovphysx/physics/ovphysx_manager.py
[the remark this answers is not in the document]
hujc7: Follow-up: the legacy-path fix in my earlier reply was removed — not because the finding was wrong, but because develop deleted the legacy runtime in #6074. The CPU/`gpu_index` gap is moot: one construction path now, receiving the cache directory unconditionally.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
```diff
@@ -7,9 +7,18 @@
...
+import os
+import tempfile
+
 from isaaclab.physics import PhysicsCfg
 from isaaclab.utils.configclass import configclass
 
+# POSIX temp roots are shared between users; Windows already gives each user a private one.
+_CACHE_DIR_NAME = f"ovphysx_derived_data_cache_{os.getuid()}" if hasattr(os, "getuid") else "ovphysx_derived_data_cache"
+
+DEFAULT_COOKED_COLLIDER_CACHE_DIR: str = os.path.join(tempfile.gettempdir(), _CACHE_DIR_NAME)
```
marcodiiga, line 19: Following up on the earlier shared-temp discussion: could we close the remaining pre-creation hole before making this the default? I verified with both the runtime pinned by this branch and ovphysx 0.5.10 that a pre-existing cache-file symlink under this predictable directory is followed and the linked file is modified during an otherwise successful cook. The UID suffix prevents accidental cross-user collisions, but it does not establish ownership of the path on a shared POSIX host. Please use a per-user cache root whose parent is not writable by other users, or atomically create and validate an owner-only, non-symlink directory (including ownership of an existing entry) before passing it to ovphysx.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
```diff
@@ -919,7 +921,9 @@ def _construct_physx(cls, ovphysx_device: str, gpu_index: int) -> None:
         """
         ovphysx = import_ovphysx()
         ovphysx.bootstrap()
-        cls._physx = cls._create_physx_instance(ovphysx, ovphysx_device, gpu_index)
+        # The runtime is also constructed outside a configured simulation, where no cfg exists.
+        cache_dir = getattr(PhysicsManager._cfg, "cooked_collider_cache_dir", DEFAULT_COOKED_COLLIDER_CACHE_DIR)
```
marcodiiga: Could we pin this config-to-runtime seam in a test and avoid the string-based fallback? test_manager_supports_pinned_runtime_api passes cache_dir directly into _create_physx_instance, so it would stay green if _construct_physx stopped reading OvPhysxCfg.cooked_collider_cache_dir. A lifecycle test can set PhysicsManager._cfg to an OvPhysxCfg with a custom path, invoke _construct_physx, and assert that the fake PhysXConfig receives it; a no-config case can assert the fallback. Since _cfg is already OvPhysxCfg | None, direct access after an explicit None check would also make a field rename fail instead of silently substituting the default.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
```diff
@@ -22,6 +31,14 @@ class OvPhysxCfg(PhysicsCfg):
 
     class_type = "{DIR}.ovphysx_manager:OvPhysxManager"
 
+    cooked_collider_cache_dir: str = DEFAULT_COOKED_COLLIDER_CACHE_DIR
```
marcodiiga: Could this docstring describe the reachable storage contract? The field is a non-optional str with a concrete default, so "Left unset" is not a state OvPhysxCfg can express. This also enables an on-disk cache by default; ovphysx does not evict entries automatically, and users reclaim space by deleting the directory. Please document that lifecycle. If disabling cross-run persistence is intended after rebasing onto ovphysx 0.5.10, the config type and construction path need to preserve None rather than describe it as an existing option.

source/isaaclab_ov/changelog.d/jichuanh-ovphysx-derived-data-cache-dir.minor.rst
```diff
@@ -0,0 +1,12 @@
+Added
+^^^^^
+
+* Added :attr:`~isaaclab_ov.physics.OvPhysxCfg.cooked_collider_cache_dir` to select where OVPhysX
+  writes its cooked-collider cache. It defaults to a per-user directory under the system temporary
+  directory, and cooked colliders now persist across runs.
```
marcodiiga: The A0 result in this PR shows that the old behavior already wrote the cache under the interpreter when that location was writable, so it already persisted across runs. A system temp directory may also be purged. Could we describe the actual change instead, for example: "cooked colliders are now reusable from the configured cache directory while that directory is retained"?

marcodiiga (changes requested):
Requesting changes for one blocking issue in the default cache location: on shared POSIX systems, the predictable path can be pre-created and redirect datastore writes before IsaacLab starts. The behavior is reproducible with both the runtime pinned by this branch and ovphysx 0.5.10. I also left focused comments on the config-to-runtime seam, the storage lifecycle contract, and the release note. Current develop now pins ovphysx 0.5.10, so please rebase and revalidate there after addressing the comments.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
hujc7: Fixed in 6ea4ab2f2ff. `_prepare_default_cache_dir` creates the default with `os.makedirs(..., mode=0o700)` and, if the path exists, `os.lstat`-rejects a symlink, a non-directory, or another uid's directory before ovphysx sees it.

Took your second option to keep the temp-rooted default consistent with `OVRTXRendererCfg.log_file_path`. A configured directory passes through untouched, so pointing the field at a site-provisioned shared cache still works.

Residual: a TOCTOU window between our `lstat` and ovphysx opening the path. Your home-rooted-parent option closes it structurally — say the word and I'll switch.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
hujc7: Both done in 6ea4ab2f2ff. `_construct_physx` reads `PhysicsManager._cfg` through an explicit `None` check, so a rename raises instead of substituting the default.

`test_construct_physx_passes_the_configured_cooked_collider_cache_dir` asserts a configured path reaches the fake `PhysXConfig`, with a no-config fallback case. Verified it fails exactly as you predicted: hardcoding the default leaves the other ten green.

The gap existed because the fixture patches `_cfg` on `OvPhysxManager` while the code reads it on `PhysicsManager`.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
hujc7: Fixed in 6ea4ab2f2ff and c61f3926d39. The fix is the type, not the wording: `PhysXConfig` has accepted `str | None` since the pinned 0.5.9, so the non-optional `str` removed a reachable option. Field and `_create_physx_instance` are now `str | None`, `None` is forwarded unchanged, and a test covers it.

Docstring now states the contract plus the lifecycle: ovphysx never evicts, the directory grows and is reclaimed by deletion, and a temp directory may be purged.

source/isaaclab_ov/changelog.d/jichuanh-ovphysx-derived-data-cache-dir.minor.rst
hujc7: Fixed in 6ea4ab2f2ff — A0 was my own evidence against that sentence. It now reads that cooked colliders are reusable from the configured directory for as long as it is retained, and mentions the `None` option.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
```diff
@@ -22,6 +31,18 @@ class OvPhysxCfg(PhysicsCfg):
...
 
+    cooked_collider_cache_dir: str | None = DEFAULT_COOKED_COLLIDER_CACHE_DIR
+    """Directory for the OVPhysX cooked-collider cache, defaulting to a per-user directory under the
+    system temporary directory.
+
+    Set to ``None`` to let OVPhysX resolve the location itself, which places the cache beside the running
+    executable -- for an embedded runtime that is the Python interpreter, whose directory the program does
+    not own.
+
+    OVPhysX never evicts entries, so the directory grows with every distinct cooked collider and is
```
marcodiiga: Non-blocking: two backend lifecycle details would make this public contract precise. The UJITSO datastore is constructed once, so the first ovphysx construction in a process fixes the cache path; later configs do not relocate it. The datastore also performs budget-driven disk garbage collection, so "never evicts" is inaccurate. Could we document the first-use/process scope, describe the bounded-cache behavior while retaining deletion as a manual reclamation option, and phrase None as "use the runtime default" so the wording does not depend on the current interpreter-directory fallback?

source/isaaclab_ov/test/physics/test_ovphysx_manager_lifecycle.py
```diff
@@ -419,3 +420,74 @@ def test_retained_binding_preserves_uncaught_failure_exit_status():
...
+def test_default_cache_dir_rejects_a_directory_owned_by_another_user(manager_module, tmp_path, monkeypatch):
+    """A pre-existing directory this user does not own is refused."""
+    import os as _os
+
+    target = tmp_path / "ovphysx_derived_data_cache_1000"
+    target.mkdir(mode=0o700)
+    monkeypatch.setattr(manager_module, "DEFAULT_COOKED_COLLIDER_CACHE_DIR", str(target))
+    monkeypatch.setattr(_os, "getuid", lambda: _os.stat(target).st_uid + 1)
+
+    with pytest.raises(RuntimeError, match="owned"):
```
marcodiiga, line 495: Non-blocking portability: os.getuid does not exist on Windows, so this monkeypatch raises before the assertion. Please make this test POSIX-only or patch with raising=False. The exact 0o700 mode assertion above is also POSIX-specific and should be guarded.

marcodiiga (approved):
Re-reviewed the current head against the IsaacLab ovphysx construction path and the pinned ovphysx 0.5.10 behavior. The previously blocking default-path issue is addressed: the default leaf is created owner-only, and a pre-existing symlink, non-directory, or foreign-owned directory is rejected before ovphysx receives it. The config-to-runtime seam and the optional None path now have focused coverage; the relevant backend tests pass.

No major P0 found, so approving. I left two non-blocking follow-ups on the public cache-lifecycle wording and Windows portability of the new POSIX-specific test. The backend overview also still says OvPhysxCfg exposes only GPU buffer sizes, so it should mention the new field. Required CI is still running, so this is not a merge-readiness claim.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager_cfg.py
hujc7: Fixed in f6e649b14a1. The docstring now states that the datastore is created once per process, so the first construction fixes the location and a later config does not relocate it; that the runtime garbage collects the cache against a disk budget, with deletion as immediate manual reclamation; and phrases `None` as "use the runtime default" rather than naming the interpreter-directory fallback.

You were right about eviction, and I had the evidence backwards — `cachelru.dat` is in the on-disk layout I had already measured, and an LRU index is exactly the mechanism I claimed did not exist. The same sentence had reached the changelog fragment and the PR description; both are corrected.

The branch has since merged develop, so it now runs against the `ovphysx==0.5.10` pin. `PhysXConfig`'s signature is unchanged there and the cache tree is still created from nothing at the configured path.

source/isaaclab_ov/test/physics/test_ovphysx_manager_lifecycle.py
hujc7: Fixed in f6e649b14a1 — all three POSIX-specific tests now carry `@pytest.mark.skipif(sys.platform == "win32", ...)`, covering both the `os.getuid` monkeypatch and the exact `0o700` mode assertion.

Two notes. This class cannot be caught here by CI: Installation Tests run ARM and x86, not Windows, so review is the only gate. And I had asserted the `os.getuid` availability rather than verified it; confirmed since against the CPython docs ("Availability: Unix") and locally via `os.getuid is posix.getuid`, which is why it has no `nt` counterpart.

hujc7:
@AntoineRichard for review if there's additional things that should be fixed.

AntoineRichard (approved):

kellyguo11:
does this need to be cherry picked to release branch?

PR #6891, opened 2026-08-04
Add the Factory contact-rich assembly task on Newton

ooctipus:
# Description

> [!IMPORTANT]
> **Depends on #7078 — that PR must be merged first.** This branch is stacked on the effort-limit mirror of #6481, so the diff includes its single actuator commit. Review the Factory commits on top; after #7078 lands, rebasing will remove the parent commit from this diff.

Adds **`IsaacContrib-Factory-Franka`**, a preset-driven contact-rich assembly task under `isaaclab_tasks.contrib.nist`.

The task covers 12 NIST taskboard variants:

- M16 nut threading
- round peg insertion at 4, 8, 12, and 16 mm
- rectangular rod insertion at 4, 8, 12, and 16 mm
- small, medium, and large gear meshing

Assembly geometry, reset strategy, curriculum, robot, and physics are composed through presets rather than separate environment classes. The task supports explicit `isaacsim_physx` and `newton_mjwarp` presets, with `physx` retaining automatic backend selection.

## Dependency

#7078 separates the two implicit-actuator effort fields:

- `effort_limit` retains the actuator-facing rated force or torque.
- `effort_limit_sim` controls the physics-solver clamp.

The Factory Franka configuration intentionally authors distinct rated and solver limits, so it cannot construct on unmodified `develop`.

## Changes

- Added the Factory task, assembly profiles, reset-state generation, sampling, rewards, observations, terminations, and curriculum.
- Added the 12 task presets and RSL-RL configuration.
- Added explicit Isaac Sim PhysX and Newton physics configuration.
- Kept task assets on uploaded asset paths; no Factory-specific Nucleus root constant is introduced.
- Bounded precollected reset states in the existing environment smoke harness so one- and two-environment CI cases remain smoke tests rather than building the full training bank.

## Type of change

- New feature (non-breaking)

## Validation

- [x] `uv run isaaclab -f`
- [x] Preset decision suite: 19 passed
- [x] Exact Factory two-environment CI smoke case: 1 passed in 35.8 seconds
- [x] Five-iteration `isaacsim_physx` training run completed
- [x] Changelog fragment validation
- [x] No new test files were added while addressing the CI failures


AntoineRichard (changes requested):
## AI-generated review

This PR is not ready to merge. The latest head still has functional blockers in task startup and collision validation, and it carries substantial unused/copied infrastructure that should be removed before this becomes a showroom task.

### Merge blockers

- The task hard-codes NVIDIA's internal `isaac-dev` Nucleus, so external users cannot load its assets.
- `SuccessMonitorCfg` is imported from `core/lift/mdp/events_cfg.py`, which has been deleted on current `develop`; the merged tree will fail to import this task.
- Reset precollection calls `CollisionAnalyzer` with the predicate signature instead of constructing it from its config.
- The SDF kernel addresses globally packed collider data with subset-local strides/indices.
- The process-global collider cache retains stage-owned prims and an old environment count across environment reconstruction.

The prior automated comments about the analyzer constructor, SDF subset indexing, nonexistent `.util.pose_offset` import, undefined `RESET_STRATEGIES`, and lift-layer coupling remain valid at this head.

### Required PR reduction and consolidation

- Remove/defer the 514-line Warp sampler and the Frontier strategy: no production config in this PR enables `warp=True` or uses Frontier; only their tests do. Keep the Torch Beta/Uniform path used by Factory.
- Remove the terrain/multi-task command observations and unused reward/termination terms copied into `contrib.nist`; they are not referenced by the Factory config and create a false public surface.
- Generate the repeated socket/plug asset configs through a small typed helper, consolidate identical keypoint/preset classes, and delete commented-out alternatives.
- Fold the duplicated Hydra import test into the existing shared test and remove config-only smoke/import tests already covered by the shared Factory runtime test and other unit tests. Keep the two scheduler behavior tests in the curriculum test module.
- Add focused regressions for reordered/subset `env_ids` and for analyzer reconstruction across stages/environment counts. Those tests catch the concrete failures; many current tests only assert trivial properties or implementation strings.

The comments/docstrings also need a full editorial pass: keep them functional and durable, remove author tags/debugging narratives/value-selection postmortems, and delete stale references to terrain, connectors, old layouts, and “backwards compatibility” for new symbols.

I did not run project code, pre-commit, or GPU tests locally because benchmarks are running on this machine. Static `git diff --check` passed. GitHub CI at this head is currently blocked with failures in two `isaaclab_tasks` shards, docs, `isaaclab_rl`, and visualizers.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
```diff
@@ -0,0 +1,521 @@
...
+from isaaclab.actuators.actuator_cfg import ImplicitActuatorCfg
+from isaaclab.assets import ArticulationCfg, AssetBaseCfg, RigidObjectCfg
+from isaaclab.sim.spawners.materials import UsdPhysicsRigidBodyMaterialCfg
+from isaaclab.utils.configclass import configclass
+
+from isaaclab_tasks.utils import PresetCfg, preset
+
+from .assembly_keypoints import NIST_BOARD_CFG
+
+ISAACLAB_NUCLEUS_DIR = "omniverse://isaac-dev.ov.nvidia.com/Isaac/IsaacLab"
```
AntoineRichard: AI-generated review — **Blocker:** this replaces the standard public asset root with NVIDIA's internal `isaac-dev` server. External users cannot resolve/authenticate these USDs, so the showroom task cannot launch. Import `ISAACLAB_NUCLEUS_DIR` from `isaaclab.utils.assets` and publish the assets under that root (or package them locally); do not embed an internal endpoint.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/reset_env_cfg.py
```diff
@@ -0,0 +1,287 @@
...
+from __future__ import annotations
+
+from isaaclab.managers import EventTermCfg as EventTerm
+from isaaclab.managers import SceneEntityCfg
+
+from isaaclab_tasks.contrib.nist.utils import (
+    BetaSamplingStrategyCfg,
+    SamplerCfg,
+)
+from isaaclab_tasks.core.lift.mdp.events_cfg import SuccessMonitorCfg
```
AntoineRichard: AI-generated review — **Blocker after merging current `develop`:** `core/lift/mdp/events_cfg.py` was deleted by the lift/dexsuite consolidation. Since this branch does not modify that deleted file, the merge keeps the deletion while these new imports remain, producing `ModuleNotFoundError`. Move `SuccessMonitor` and `SuccessMonitorCfg` into a stable shared module owned by this reset/sampling infrastructure and import it from there in both this file and `event_combinators.py`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/event_combinators.py
```diff
@@ -0,0 +1,349 @@
...
+            precollect_capacity = self.state_data.shape[0]
+            target_size = min(self._state_target_size, precollect_capacity)
+            state_size = 0
+            pbar = tqdm(total=precollect_capacity, desc="reset_accumulator")
+            while state_size < precollect_capacity:
+                prev_size = state_size
+                reset_term.func(env, all_env_ids, **reset_term.params)
+                valid_mask = torch.ones(len(all_env_ids), dtype=torch.bool, device=env.device)
+                for condition in self.acceptance_conditions.values():
+                    condition_func = condition if callable(condition) else condition.func
```
AntoineRichard: AI-generated review — **Blocker:** each value here is a `CollisionAnalyzerCfg`; its resolved `func` is the `CollisionAnalyzer` class whose constructor is `(cfg, env)`. Calling that class as `(env, all_env_ids)` fails on the first precollection reset. Resolve and instantiate configured conditions once in `__init__` (for example, `cfg.func(cfg, env)`), store the resulting predicates, and call those with `(env, env_ids)`. Add a regression that actually enters precollection with a `CollisionAnalyzerCfg`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sdf.py
```diff
@@ -0,0 +1,268 @@
...
+    best = max_dist
+    for obs_idx in range(num_obstacles):
+        root_mat = compose_root_mat(
+            obs_root_quat[obs_idx, env_id],
+            obs_root_scale[obs_idx, env_id],
+        )
+        root_mat_inv = wp.inverse(root_mat)
+        root_pos = obs_root_pos[obs_idx, env_id]
+
+        prim_base = obs_idx * num_envs_subset * max_prims + env_local * max_prims
```
AntoineRichard: AI-generated review — **Important correctness:** `CollisionAnalyzer` packs these buffers as `obs * global_num_envs * max_prims + absolute_env_id * max_prims`, but this kernel uses the query subset size and local subset index. Partial or reordered IDs therefore read another environment's colliders/counts. Pass the global environment count and use `env_id` in both offsets. Please add a regression with reordered IDs such as `[3, 1]`; the full ordered range cannot expose this bug.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/rigid_object_hasher.py
```diff
@@ -0,0 +1,217 @@
...
+HASH_STORE = {"warp_mesh_store": {}}
+
+
+class RigidObjectHasher:
+    """Compute per-root and per-collider hashes and full 3x3 transforms."""
+
+    def __init__(self, num_envs, prim_path_pattern, device="cpu"):
+        self.prim_path_pattern = prim_path_pattern
+        self.device = device
+        if prim_path_pattern in HASH_STORE:
```
AntoineRichard: AI-generated review — **Important lifecycle bug:** this process-global cache is keyed only by prim-path text, but stores `Usd.Prim` objects, `num_roots`, environment IDs, and tensors from one stage. Reconstructing the same task with a different environment count/stage (the shared Factory test does 2 envs and then 1) reuses stale prims and incompatible shapes. Scope the cache to the active stage/`SimulationContext` plus geometry/environment layout, or clear it on context teardown. Add a same-process reconstruction regression.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
```diff
@@ -0,0 +1,290 @@
...
+
+    range_list = [pose_range.get(key, (0.0, 0.0)) for key in ["x", "y", "z", "roll", "pitch", "yaw"]]
+    ranges = torch.tensor(range_list, device=env.device)
+    samples = math_utils.sample_uniform(ranges[:, 0], ranges[:, 1], (len(env_ids), 6), device=env.device)
+    new_pos = nominal_pos + samples[:, 0:3]
+    new_quat = math_utils.quat_mul(
+        nominal_quat, math_utils.quat_from_euler_xyz(samples[:, 3], samples[:, 4], samples[:, 5])
+    )
+
+    fixed_asset.write_root_pose_to_sim(torch.cat([new_pos, new_quat], dim=1), env_ids=env_ids)
```
AntoineRichard: AI-generated review — **Performance/API:** all new writes use deprecated convenience methods that route through the indexed API and emit warnings. During accumulator precollection, `env_ids` is the full ordered `torch.arange(num_envs)`, so every reset still pays the slower indexed path. Use explicit `*_index` methods for genuine subsets and centralize a full-environment fast path using the `*_mask` methods with `env_mask=None`; apply the same treatment to the root/joint writes in this file and `reset_state.py`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
```diff
@@ -0,0 +1,173 @@
...
+    time_left_frac = 1 - env.episode_length_buf / env.max_episode_length
+    return time_left_frac.view(env.num_envs, -1)
+
+
+# ---------------------------------------------------------------------------
+# Multi-task command observation accessors.
+# ---------------------------------------------------------------------------
+
+
+def command_progress(env, command_name: str = "goal_point"):
```
AntoineRichard: AI-generated review — **Remove unused copied surface:** Factory never references `time_left` or any `command_*` observation; this NIST module also documents terrain and a `MultiTaskCommand` that the package does not contain. The same issue exists for `command_task_reward`, `mechanical_power`, `contact_penalty`, `abnormal_robot_state`, and `illegal_contact_ratio`. Delete these unused terms and their stub exports rather than publishing unrelated generic APIs from a showroom task.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sampling/impl/sampler_warp.py
```diff
@@ -0,0 +1,514 @@
...
+    while lo < hi:
+        mid = (lo + hi) // 2
+        if u <= cdf[mid]:
+            hi = mid
+        else:
+            lo = mid + 1
+    samples[i] = wp.int64(lo)
+
+
+class SamplerWarp:
```
AntoineRichard: AI-generated review — **PR-size reduction:** no production config in this PR sets `SamplerCfg(warp=True)`; the only caller is the equivalence test. Likewise, Frontier is referenced only by tests/stubs, not Factory. Defer/remove this 514-line backend, Frontier, and their tests from this PR, and keep the Torch Beta/Uniform implementation that the task actually uses. This removes a large GPU-only maintenance surface without changing task behavior.

source/isaaclab_tasks/test/test_factory_hydra_imports.py
```diff
@@ -0,0 +1,109 @@
...
+
+        result = {{
+            'load_error': load_error,
+            'violations': violations,
+        }}
+        print("__RESULTS__" + json.dumps(result))
+    """)
+
+
+@pytest.mark.parametrize("task_name", _FACTORY_TASKS)
```
AntoineRichard: AI-generated review — **Duplicate test:** `test/core/test_env_cfg_no_forbidden_imports.py` already owns this import-hook/subprocess check. Its discovery currently filters with `startswith("Isaac-")`, which misses `IsaacContrib-*`; extend that single canonical test's discovery and delete this 109-line copy. Maintaining two versions will make their forbidden sets and startup behavior drift.

source/isaaclab_tasks/test/test_factory_smoke.py
```diff
@@ -0,0 +1,189 @@
...
+from types import SimpleNamespace
+
+import gymnasium as gym
+import pytest
+import torch
+
+import isaaclab_tasks  # noqa: F401 -- registers the gym tasks
+
+
+@pytest.mark.parametrize("task_name", ["IsaacContrib-Factory-Franka"])
```
AntoineRichard: AI-generated review — **Consolidate tests:** `test/contrib/test_factory_environments.py` already discovers Factory tasks and actually constructs, resets, and steps them. This config-only construction test and `test_env_cfg_to_dict_serialises` repeat the Hydra/config path, while `test_relocated_module_imports` imports modules already exercised by the sampler/state tests. Remove those redundant cases; move the two meaningful `DifficultyScheduler` behavior tests into `test_curriculum.py`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
```diff
@@ -0,0 +1,521 @@
...
+        usd_path=f"{ASSET_DIR}/NIST/nut_m16.usd",
+        rigid_props=ASSEMBLY_PLUG_RIGID_BODY_PROPS_CFG,
+        mass_props=sim_utils.MassPropertiesCfg(mass=0.03),
+        collision_props=ASSEMBLY_PLUG_COLLISION_PROPS_CFG,
+        physics_material=ASSEMBLY_CONTACT_MATERIAL_CFG,
+    ),
+)
+
+
+HOLE_16MM_CFG: RigidObjectCfg = RigidObjectCfg(
```
AntoineRichard: AI-generated review — **Consolidate repeated configuration:** the following hole/rod/peg/gear blocks repeat the same `RigidObjectCfg` shape and differ only in name, USD filename, mass, and socket-vs-plug properties. Use a small typed factory/helper to build the named constants. This can remove roughly 200 lines and prevents one size variant from silently drifting in collision/material settings.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_presets.py
```diff
@@ -0,0 +1,560 @@
...
+    peg_insert_4mm: kpts.Offset = kpts.RECTANGULAR_HOLE_4MM.hole_tip_offset
+    peg_insert_8mm: kpts.Offset = kpts.RECTANGULAR_HOLE_8MM.hole_tip_offset
+    peg_insert_12mm: kpts.Offset = kpts.RECTANGULAR_HOLE_12MM.hole_tip_offset
+    peg_insert_16mm: kpts.Offset = kpts.RECTANGULAR_HOLE_16MM.hole_tip_offset
+
+    default: kpts.Offset = nut_thread_m16
+
+
+@configclass
+class FixedAssetTipCfg(PresetCfg):
```
AntoineRichard: AI-generated review — **Exact duplication:** `FixedAssetTipCfg` is field-for-field identical to `HeldAssetTipCfg` for every preset. Replace both with one semantically named `AssemblyTipCfg` (or alias one without duplicating fields) and update the two consumers. Keeping parallel classes guarantees future variants can update one mapping and forget the other.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/__init__.pyi
```diff
@@ -0,0 +1,62 @@
...
+    "progress_reward",
+    "success_reward",
+    "action_rate_l2_clamped",
+    "action_l2_clamped",
+    "out_of_bound",
+    "abnormal_robot_state",
+    "progress_context",
+    "success_termination",
+    "CollisionAnalyzerCfg",
+    "RESET_STRATEGIES",
```
AntoineRichard: AI-generated review — **Broken lazy export:** `RESET_STRATEGIES` is not imported or defined anywhere in this package. Advertising it in `__all__` makes `mdp.RESET_STRATEGIES`/star imports fail. Remove the entry; do not add another alias unless there is an actual supported API that needs it.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
```diff
@@ -0,0 +1,173 @@
...
+import isaaclab.utils.math as math_utils
+from isaaclab.managers import SceneEntityCfg
+
+from isaaclab_tasks.contrib.nist.utils import get_reset_state
+
+if TYPE_CHECKING:
+    from isaaclab.assets import Articulation, RigidObject
+    from isaaclab.envs import ManagerBasedEnv, ManagerBasedRLEnv
+
+    from .util.pose_offset import Offset
```
AntoineRichard: AI-generated review — **Broken type/doc reference:** there is no `mdp.util` package. `Offset` lives in `..utils.pose_offset`, so type checkers and documentation cannot resolve these annotations despite runtime postponement. Import it from the real module.

source/isaaclab/isaaclab/actuators/actuator_pd.py
```diff
@@ -57,25 +57,19 @@ class ImplicitActuator(ActuatorBase):
 
     def __init__(self, cfg: ImplicitActuatorCfg, *args, **kwargs):
         # effort limits
+        # 'effort_limit' is the motor's rated torque/force reflected at the joint: it feeds the
```
AntoineRichard: AI-generated review — **Comment hygiene:** this repeats the public field docstrings and embeds the Factory-specific `87 N-m` tuning story in a core actuator. Keep only the durable invariant (rated limit and solver clamp resolve independently); the two fallback assignments are self-explanatory. Task-specific values and why they were chosen belong with the task configuration or PR evidence.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_env_cfg.py
```diff
@@ -0,0 +1,359 @@
...
+
+    policy: PolicyCfg = PolicyCfg()
+    critic: PolicyCfg = PolicyCfg()
+
+
+@configclass
+class FactoryEventCfg:
+    """Events specifications for Factory"""
+
+    # when nut dropped right above the bolt, it sometime can immediately success due to high speed it falls
```
AntoineRichard: AI-generated review — **Rewrite/remove these comments:** this text is ungrammatical, describes gravity rather than the material term it precedes, and the author-tagged inertia note below records a debugging narrative and tuning values instead of a functional contract. Put one concise comment beside `variable_gravity` and describe the inertia event functionally, e.g. that it increases diagonal inertia to prevent contact-induced angular instability.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_env_cfg.py
```diff
@@ -0,0 +1,359 @@
...
+##
+# Environment configuration
+##
+
+
+@configclass
+class FactoryPhysicsCfg(PresetCfg):
+    """Physics-backend preset for Factory tasks.
+
+    Selected via ``presets=physx`` (default) or ``presets=newton_mjwarp``. The PhysX
```
AntoineRichard: AI-generated review — **Documentation hygiene:** this docstring is a tuning log explaining why individual solver values differ from a demo. Keep the stable user contract and selectors here; move benchmark comparisons and value-selection rationale to the PR/issue. The current prose will become stale as soon as any knob changes.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
```diff
@@ -0,0 +1,521 @@
...
+        sdf_narrow_band_outer=0.005,
+    )
+    physx = default
+
+
+ASSEMBLY_SOCKET_COLLISION_PROPS_CFG = _SocketCollisionPropsCfg()
+
+ASSEMBLY_PLUG_COLLISION_PROPS_CFG = _PlugCollisionPropsCfg()
+
+# Newton-only stiff contact material for the assembly pair. Raising stiffness globally
```
AntoineRichard: AI-generated review — **Comment hygiene:** this ten-line solver postmortem is too implementation-specific for a config module. Preserve the actionable invariant in one or two lines (the Newton material must bind friction because omitted friction resolves to zero and destabilizes contact); move derivations, warning transcripts, and chosen-value rationale to the PR/issue.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
```diff
@@ -0,0 +1,521 @@
...
+        velocity_limit_sim=2.0,
+        stiffness=7500.0,
+        damping=173.0,
+        friction=0.1,
+        armature=0.0,
+    ),
+}
+
+
+# FRANKA_ACTUATORS_CFG = {
```
AntoineRichard: AI-generated review — **Delete dead code:** this 27-line commented alternative duplicates the active actuator configuration. Git already preserves the previous version; leaving it here obscures which limits are supported and invites edits to the wrong block.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/collision_analyzer_cfg.py
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+# Copyright (c) 2024-2025, The Isaac Lab Project Developers.
```
AntoineRichard: AI-generated review — **File hygiene:** this new file contains a second stale 2024–2025 copyright/license header immediately after the required 2026 header. Remove the duplicate header.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/pose_offset.py
```diff
@@ -0,0 +1,123 @@
...
+        return self.pos + self.quat
+
+    def __post_init__(self):
+        self._pos_t: torch.Tensor | None = None
+        self._quat_t: torch.Tensor | None = None
+        self._inv_pos_t: torch.Tensor | None = None
+        self._inv_quat_t: torch.Tensor | None = None
+
+    def _ensure_tensors(self, device: str | torch.device):
+        """Materialize and cache forward + inverse tensors on first use."""
```
AntoineRichard: AI-generated review — **Device correctness:** the cache is materialized only once, but every public method accepts a device. Reusing an `Offset` first on CPU and then CUDA (or across CUDA devices) returns tensors on the first device and fails in frame operations. Cache per device or rematerialize when `self._pos_t.device != device`; apply the same device-aware rule to `AssemblyProfile`/segment tensor caches.

source/isaaclab_tasks/changelog.d/zhengyuz-factory-newton.rst
```diff
@@ -0,0 +1,6 @@
+Added
+^^^^^
+
+* Added the ``IsaacContrib-Factory-Franka`` contact-rich assembly task under
```
AntoineRichard, line 4: AI-generated review — **Stale user documentation:** the current task presets also expose three gear-mesh and four round-rod variants, but the only user-facing changelog entry says nut threading and peg insertion. Update the concise canonical description to match the shipped variants, and remove duplicated/stale variant descriptions from module docstrings.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/agents/rsl_rl_ppo_cfg.py
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+from isaaclab.utils.configclass import configclass
+
+from isaaclab_rl.rsl_rl import RslRlMLPModelCfg, RslRlOnPolicyRunnerCfg, RslRlPpoAlgorithmCfg
+
+from isaaclab_tasks.utils import PresetCfg, preset
+
+# Shared PPO hyper-parameters reused by both the plain-PPO and value-shift variants.
+_FACTORY_PPO_KWARGS = dict(
```
StafaH: Can't you just inherit the config instead? Why extract out just the PPO args?

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+    uv run isaaclab train --task IsaacContrib-Factory-Franka \
+        presets=peg_insert_4mm physics=newton_mjwarp
+
+See :mod:`..factory_presets` for the robot registry and how to add new robots.
+"""
+
+import gymnasium as gym
+
+from . import agents
```
StafaH: Absolute imports here and elsewhere

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+
+# Imported for its side effect: the module sets each robot-specific field on the
```
StafaH: This comment is actually pretty scary.  ⚠️ Magical things happen during import ⚠️

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/franka_factory_cfg.py
```diff
@@ -0,0 +1,65 @@
...
+    EndEffectorBodyCfg,
+    GripperGraspOffsetCfg,
+    GripperJointNamesCfg,
+    IKJointNamesCfg,
+    JointEffortNamesCfg,
+    RobotActionsCfg,
+    RobotArticulationCfg,
+)
+
+EndEffectorBodyCfg.default = "panda_fingertip_centered"
```
StafaH: More magical things happening here?

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/franka_factory_cfg.py
```diff
@@ -0,0 +1,65 @@
...
+    arm_action = mdp.RelativeJointPositionActionCfg(
+        asset_name="robot",
+        joint_names=["panda_joint.*"],
+        scale={"(?!panda_joint7).*": 0.02, "panda_joint7": 0.2},
+        use_zero_offset=True,
+    )
+
+    gripper_action = mdp.BinaryJointPositionActionCfg(
+        asset_name="robot",
+        joint_names=["panda_finger.*"],
```
StafaH: `_.*`? To match the expressions below

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/curriculums.py
```diff
@@ -0,0 +1,107 @@
...
+        return mdp.modify_env_param.NO_CHANGE
+
+    # convert iv/fv to tensors, but we'll peel them apart in recursion
+    initial_value_tensor = torch.tensor(initial_value, device=env.device)
+    final_value_tensor = torch.tensor(final_value, device=env.device)
+
+    return _recurse(initial_value_tensor.tolist(), final_value_tensor.tolist(), data, frac)
+
+
+def _recurse(iv_elem, fv_elem, data_elem, frac):
```
StafaH, line 20: nit: recurse should be at above the MDP that uses it.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
```diff
@@ -0,0 +1,290 @@
...
+        asset_map: Mapping from scene entity key to :class:`NistBoardKeyPointsCfg` attribute name.
+            The ``"fixed_asset"`` entry selects the board keypoint used as the nominal center.
+        pose_range: Per-axis ``(min, max)`` sample ranges, keys ``x``/``y``/``z`` [m] and
+            ``roll``/``pitch``/``yaw`` [rad]. Missing axes stay at the nominal value.
+    """
+    nistboard: RigidObject = env.scene["nistboard"]
+    fixed_asset: Articulation | RigidObject = env.scene["fixed_asset"]
+    keypoint: Offset = getattr(NIST_BOARD_CFG, asset_map["fixed_asset"])
+
+    board_default = wp.to_torch(nistboard.data.default_root_state)[env_ids]
```
StafaH: nistboard is not a ProxyArray? if not, then should it be?

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Args:
+        asset_map: Mapping from scene entity key to :class:`NistBoardKeyPointsCfg` attribute name.
+            ``"fixed_asset"`` selects the keypoint solved against; the rest ride along on the board.
+    """
+    nistboard: RigidObject = env.scene["nistboard"]
+    fixed_asset: Articulation | RigidObject = env.scene["fixed_asset"]
+    keypoint: Offset = getattr(NIST_BOARD_CFG, asset_map["fixed_asset"])
+
+    fixed_pos = wp.to_torch(fixed_asset.data.root_pos_w)[env_ids]
```
StafaH: Here and elsewhere, can we use proxy array and not call wp.to_torch?

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+# ---------------------------------------------------------------------------
```
StafaH: Let's get rid of the comment blocks? (especially if theyre empty!)

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sampling/sampler.py
```diff
@@ -0,0 +1,55 @@
...
+
+from .impl.sampler_torch import SamplerTorch
+from .impl.sampler_warp import SamplerWarp
+
+if TYPE_CHECKING:
+    from ..state_layout import StateLayout
+    from .sampler_cfg import SamplerCfg
+
+
+class Sampler:
```
StafaH, line 20: Could this be important enough to actually be in isaaclab core?

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/mesh_ops.py
```diff
@@ -0,0 +1,324 @@
...
+@cache
+def _load_mesh_tensors(prim):
+    tm = prim_to_trimesh(prim)
+    verts = torch.from_numpy(tm.vertices.astype("float32"))
+    faces = torch.from_numpy(tm.faces.astype("int64"))
+    return verts, faces
+
+
+@wp.kernel
+def _sample_triangle_mesh_surface_kernel(
```
StafaH, line 65: Depending on how committed you are to this, warp has a series of really efficient mesh ops you can use after constructing a wp.Mesh. Those would improve the performance alot (they use BVH internally).

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/event_combinators.py
```diff
@@ -0,0 +1,349 @@
...
+                    valid_mask &= condition_func(env, all_env_ids)
```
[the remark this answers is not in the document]
ooctipus: Addressed in `2f7dda8cbc8`. The event manager now recursively resolves the nested `ManagerTermBaseCfg` values before constructing `reset_accumulator`; its constructor stores the resolved predicate instances once, and precollection only calls those predicates with `(env, env_ids)`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/__init__.pyi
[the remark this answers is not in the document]
ooctipus: Resolved in `2f7dda8cbc8` by removing the nonexistent `RESET_STRATEGIES` export rather than adding an alias for an unsupported API.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sdf.py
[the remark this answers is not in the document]
ooctipus: Fixed in `2f7dda8cbc8`. The kernel now receives the global environment count and indexes collider slots and counts with the absolute `env_id`, so partial and reordered subsets use the packed layout correctly.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
```diff
@@ -0,0 +1,173 @@
...
+import isaaclab.utils.math as math_utils
+from isaaclab.managers import SceneEntityCfg
+
+from isaaclab_tasks.core.multi_task.utils import get_reset_state
+
+if TYPE_CHECKING:
+    from isaaclab.assets import Articulation, RigidObject
+    from isaaclab.envs import ManagerBasedEnv, ManagerBasedRLEnv
+
+    from .util.pose_offset import Offset
```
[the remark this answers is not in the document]
ooctipus: Fixed in `2f7dda8cbc8`. The type-only import now points to `isaaclab_tasks.contrib.nist.utils.pose_offset.Offset`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/event_combinators.py
```diff
@@ -0,0 +1,349 @@
...
+
+from collections.abc import Callable
+from typing import TYPE_CHECKING
+
+import torch
+from tqdm import tqdm
+
+from isaaclab.managers import EventTermCfg, ManagerTermBase
+
+from isaaclab_tasks.core.lift.mdp.events_cfg import SuccessMonitorCfg
```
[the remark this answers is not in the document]
ooctipus: Resolved in `2f7dda8cbc8`. `SuccessMonitor` and `SuccessMonitorCfg` now live in `isaaclab_tasks.utils.success_monitor`; NIST imports that shared owner directly, and Lift re-exports it for its existing callers.

source/isaaclab/isaaclab/utils/assets.py
```diff
@@ -95,6 +95,11 @@ def _resolve_asset_root() -> str:
 ISAACLAB_NUCLEUS_DIR: str = f"{ISAAC_NUCLEUS_DIR}/IsaacLab"
 """Path to the ``Isaac/IsaacLab`` directory on the NVIDIA Nucleus Server."""
 
+LOCAL_ASSET_PATH_DIR: str = os.path.normpath(
```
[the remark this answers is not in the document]
ooctipus: Resolved in `2f7dda8cbc8`. The Factory assets no longer use `LOCAL_ASSET_PATH_DIR`; they now derive from the canonical public `ISAACLAB_NUCLEUS_DIR`, so wheel installs do not depend on a source-checkout path.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sdf.py
ooctipus: Fixed in `2f7dda8cbc8`. `get_signed_distance_mega` now takes global `num_envs` and uses absolute `env_id` for both offsets. I validated the reordered `[3, 1]` case directly, while keeping that validation out of a new test file as requested.

source/isaaclab_newton/changelog.d/zhengyuz-newton-cloner-no-approximation.major.rst
```diff
@@ -0,0 +1,8 @@
+Changed
+^^^^^^^
+
+* **Breaking:** Removed ``NewtonCfg.simplify_meshes``. Newton replication no longer
```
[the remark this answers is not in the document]
ooctipus: This removal is no longer part of the PR after the rebase onto current `develop`. The current branch leaves the `simplify_meshes` public API unchanged.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/reset_env_cfg.py
ooctipus: Resolved in `2f7dda8cbc8`. The monitor implementation and config moved to the stable shared module `isaaclab_tasks.utils.success_monitor`, and both NIST reset consumers import it there. Lift retains only a compatibility re-export.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
ooctipus: Resolved in `2f7dda8cbc8`. Factory now imports the canonical `ISAACLAB_NUCLEUS_DIR`. I also verified every one of the 26 referenced USD URLs is already uploaded and reachable from the public IsaacLab asset root.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/event_combinators.py
ooctipus: Resolved in `2f7dda8cbc8`. Nested acceptance configs are resolved by the manager before `reset_accumulator` construction; the constructor now stores those resolved predicate instances once and precollection calls them directly.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
ooctipus: Addressed in `2f7dda8cbc8`. All NIST root and joint writes now use the explicit `*_index` APIs, eliminating the deprecated convenience calls. I did not add a full-env index/mask dispatcher because that would add branching and a second write path without a measured benefit.

source/isaaclab_tasks/test/test_factory_hydra_imports.py
ooctipus: Addressed in `2f7dda8cbc8`. The duplicate Hydra subprocess test was deleted, and the canonical `test_env_cfg_no_forbidden_imports.py` discovery now includes the Factory registration.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
ooctipus: Fixed in `2f7dda8cbc8`. The annotation imports `Offset` from the real `isaaclab_tasks.contrib.nist.utils.pose_offset` module.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_presets.py
ooctipus: Addressed in `2f7dda8cbc8`. `FixedAssetTipCfg` and `HeldAssetTipCfg` were replaced by one `AssemblyTipCfg`, and both consumers now use it.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/__init__.pyi
ooctipus: Addressed in `2f7dda8cbc8` by deleting `RESET_STRATEGIES` from the lazy public surface. No compatibility alias was added because no supported implementation exists.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
ooctipus: Addressed in `2f7dda8cbc8`. The unused command/time observations, copied reward terms, copied termination terms, related imports, and lazy exports were deleted; the NIST MDP surface now contains only terms consumed by Factory.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/rigid_object_hasher.py
ooctipus: Resolved in `2f7dda8cbc8` by deleting the process-global cache entirely. Each `RigidObjectHasher` reads the active stage, and each `CollisionAnalyzer` owns its Warp meshes and stage-derived tensors, so reconstruction cannot reuse stale prims or environment layouts.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. A single `_assembly_asset_cfg` now owns the shared socket/plug physics policy and the named constants provide only name, USD file, mass, and role. This removed roughly 350 lines of repeated configuration.

source/isaaclab_tasks/test/test_factory_smoke.py
ooctipus: Addressed in `2f7dda8cbc8`. The config-only Factory smoke module was deleted; its two meaningful scheduler cases now live in the existing `test_curriculum.py`, while simulator construction remains owned by `test_factory_environments.py`.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sampling/impl/sampler_warp.py
ooctipus: Addressed in `2f7dda8cbc8`. The 514-line Warp sampler backend, backend wrapper, `Frontier`, `StateLayout`, and their unused test surface were removed. The sampler is now the direct Torch Beta/Uniform implementation used by Factory.

source/isaaclab/isaaclab/actuators/actuator_pd.py
ooctipus: Addressed in `2f7dda8cbc8`. The core actuator now keeps one durable comment about independent model-facing and solver-facing limits; the Factory-specific values and fallback narration were removed.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/agents/rsl_rl_ppo_cfg.py
ooctipus: Addressed in `2f7dda8cbc8` by removing both the extracted PPO kwargs and the one-use wrapper. `FactoryPPORunnerCfg` now directly owns its `RslRlPpoAlgorithmCfg`, so an extra inheritance layer is unnecessary.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_env_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. `FactoryPhysicsCfg` now documents only the stable backend selector contract; the tuning history and demo comparisons were removed.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/collision_analyzer_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. The duplicate stale license header was removed.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. The material postmortem was reduced to the actionable invariant that Newton friction must be authored because omission resolves to zero and destabilizes contact.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/__init__.py
ooctipus: Addressed in `2f7dda8cbc8`. Regular NIST Python modules now use absolute imports throughout. The `.pyi` lazy-export stubs retain relative imports because that is the repository-required stub format.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_assets_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. The commented alternative actuator block was deleted.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/factory_env_cfg.py
ooctipus: Addressed in `2f7dda8cbc8`. The inertia term now has one functional comment about preventing contact-induced angular instability, and the gravity comment sits beside `variable_gravity` and describes the curriculum behavior.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/pose_offset.py
ooctipus: Fixed in `2f7dda8cbc8`. `Offset`, assembly segments, and `AssemblyProfile` now rematerialize cached tensors when the requested device changes.

source/isaaclab_tasks/changelog.d/zhengyuz-factory-newton.rst
ooctipus: Addressed in `2f7dda8cbc8`. The changelog now covers all 12 shipped variants: nut threading, three gear-mesh tasks, four round-rod insertions, and four rectangular-peg insertions. Duplicate variant prose was also reduced.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/__init__.py
ooctipus: Addressed in `2f7dda8cbc8`. Registration is now a pure composition root pointing directly at `FactoryEnvCfg`; the import-time class mutation and warning comment were removed.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/curriculums.py
ooctipus: Addressed in `2f7dda8cbc8`. `_recurse` now appears above `initial_final_interpolate_fn`, its consumer.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/sampling/sampler.py
ooctipus: I kept the sampler under `contrib.nist`: Factory is its only production consumer today, so promoting a one-task helper to core would broaden API ownership without a second use case. The implementation was reduced to the direct Torch path instead.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/observations.py
ooctipus: Addressed in `2f7dda8cbc8`. The empty section blocks and stale copied observation commentary were removed along with the unused terms.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/utils/mesh_ops.py
ooctipus: I kept the current surface sampler because it runs only during analyzer preprocessing, while runtime collision queries already use `wp.Mesh` and Warp mesh queries. A BVH-based sampling rewrite would add substantial scope without a measured startup bottleneck; I would revisit it with profiling evidence.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/franka_factory_cfg.py
ooctipus: The file and mutation path were deleted in `2f7dda8cbc8`, so this catch-all regex no longer exists. The direct action configuration uses the actual Panda joint expressions at its owner.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
ooctipus: Addressed in `2f7dda8cbc8`. Asset-state reads now use ProxyArray `.torch` views. The only remaining `wp.to_torch` converts the collision analyzer output allocated by Warp itself, rather than simulator asset data.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/config/franka_factory_cfg.py
ooctipus: Addressed in `2f7dda8cbc8` by deleting `franka_factory_cfg.py`. Robot and action configuration are now owned directly by the Factory scene/environment composition root, with no import-time mutation layer.

source/isaaclab_tasks/isaaclab_tasks/contrib/nist/mdp/events.py
ooctipus: Addressed in `2f7dda8cbc8`. NIST board and other asset data reads now use their ProxyArray `.torch` views directly.

ooctipus:
Extracted the implicit-actuator effort-limit semantics into #7078, the effort-limit mirror of #6481, and rebased this branch on that single prerequisite commit. The direct diff from #7078 to this head now touches only `isaaclab_tasks`.

Revalidated after the split:
- preset decision suite: 19 passed
- exact `IsaacContrib-Factory-Franka-2-cuda` smoke case: 1 passed in 35.8 seconds
- repository formatter and changelog checks: passed

StafaH (approved):

PR #6892, opened 2026-08-04
Stop the Newton cloner from approximating mesh colliders

ooctipus:
# Description

Newton replication ran a blanket convex-hull pass over every mesh collider, so a USD-authored `physics:approximation` was discarded the moment the asset was cloned. Assembly geometry — threads, chamfered holes, anything concave that matters — cannot survive that, and the only escape hatch was `NewtonCfg.simplify_meshes`, a scene-wide flag that turned approximation off for *every* collider at once.

Approximation belongs on the asset, not in the cloner. This PR drops the pass and the flag: the cloner imports colliders as authored (`newton.ModelBuilder.add_usd(skip_mesh_approximation=...)`), and a config that wants a convex hull asks for it per collider instead of getting one imposed.

## Changes

**`isaaclab_newton`**
* `newton_clone_utils.py` — the cloner no longer remeshes anything. Removes the `physics:approximation` → remeshing-method table and the two passes built on it.
* `replicate.py`, `newton_manager_cfg.py`, `visualization_builder.py` — `simplify_meshes` plumbing removed.
* `test/cloner/test_collision_approximation.py` — covers each authored mode (`convexDecomposition`, `boundingSphere`, `boundingCube`, `meshSimplification`, `none`) mapping to its shape type, an unauthored mesh staying a trimesh, mixed stages, and sources that resolve to different shape types being left alone.

**`isaaclab`**
* `CollisionBaseCfg.mesh_collision_property` — the replacement mechanism. A file-spawned USD asset exposes no approximation knob through `collision_props`, so without this there is no way to author an approximation from a spawner config once the cloner stops doing it. The dispatch into `modify_mesh_collision_properties` already existed in `schemas.py` behind a `getattr`; this declares the field that feeds it.

**`isaaclab_tasks`**
* `lift_env_cfg.py`, `reorient_manager_env_cfg.py` — author `convexHull` on the two assets that were relying on the implicit hulling (see the audit below).

## Collider audit: every task, before vs after

Rather than reason about which assets were affected, every registered task was launched at `num_envs=1` on both revisions and each colliding shape's resolved Newton `GeoType` was dumped and diffed.

| | before (`develop`) | after |
|---|---|---|
| tasks probed | 132 | 132 |
| comparable (`ok` on both sides) | 61 | 61 |
| uncovered | 65 skipped · 6 error | 65 skipped · 6 error |

Uncovered: 35 physx-only tasks, 8 with no `sim.physics`, 15 camera/visuomotor variants whose non-render siblings were probed, 5 pre-existing Newton limitations (surface grippers, reversed gripper joints), 1 missing optional tetrahedralization dependency.

**Three tasks changed**, all `CONVEX_MESH → BOX` on the same prim:

* `Isaac-Lift-KukaAllegro`, `Isaac-Reorient-Franka`, `Isaac-Reorient-KukaAllegro` — `/World/envs/env_0/Object/geometry/mesh`

Both root causes are tessellated mesh primitives that authored no approximation: `lift_env_cfg.ObjectCfg` is a `MultiAssetSpawnerCfg` of `MeshCuboidCfg` / `MeshSphereCfg` / `MeshCapsuleCfg` / `MeshConeCfg` variants sharing one `collision_props`, and the Reorient object is the dex cube. Authoring `mesh_collision_property=MeshCollisionPropertiesCfg(mesh_approximation_name="convexHull")` on the shared props — which the multi-asset spawner propagates to every variant — restores all three to **byte-identical** collider dumps.

Four other tasks (`IsaacContrib-Stack-Cube-SO101-{v0,IK-Abs-v0,Joint-Teleop-v0}`, `IsaacContrib-Stack-Cube-Bin-Franka-IK-Rel-Mimic`) showed shape-count deltas, but a same-revision control run reproduced the same spread with no code change at all — SO101 gave 2305 vs 2287 convex meshes on two runs of `develop`, the bin 72 vs 70. Their convex decomposition is nondeterministic run to run and the observed deltas sit inside that. A control re-run of `Isaac-Reorient-Franka` was byte-identical, confirming the control discriminates.

One measurement caveat: at `num_envs=1` a multi-asset spawner instantiates one of its variants, so the sweep witnessed one shape per Lift task rather than all sixteen. The fix is on the shared props that propagate to every variant.

## Type of change

### Note on the breaking change

`NewtonCfg.simplify_meshes` is removed outright rather than deprecated first, which departs from the deprecation rule in `AGENTS.md`. The field only ever selected between "approximate everything" and "approximate nothing"; a deprecated shim that still forced convex hulls would keep the geometry-destroying path alive for another release, and any scene that wants convex hulls can now express that per collider. Happy to add a shim instead if reviewers prefer.

Out-of-tree scenes that relied on the implicit hulling and author no approximation will now load raw trimeshes. USD defaults `physics:approximation` to `none`, so the fix for those assets is the same one applied to Lift and Reorient here: author the approximation they actually want.

## Screenshots

n/a

## Checklist


ooctipus:
#6891 (Factory contact-rich assembly task on Newton) is stacked on this PR and needs it merged first.

kellyguo11 (approved):

kellyguo11:
image
this seems incorrect in the rendering-correctness test

ooctipus:
> image this seems incorrect in the rendering-correctness test

https://github.com/isaac-sim/IsaacLab/pull/6896 is also failing, this doesn't look like this pr's issue

PR #6894, opened 2026-08-04
Fix XR startup abort on unshipped omni.kit.xr.bundle.generic

rwiltz:
# Description

  --xr aborted at startup with a dependency solver failure wherever the Kit extension registry was unreachable — reported by QA in the Isaac Sim 6.1 container:

  No versions of omni.kit.xr.bundle.generic that satisfies: isaaclab.python.xr.openxr-3.0.0
    Available packages for omni.kit.xr.bundle.generic version *: (none found)
  [Error] [omni.kit.app.plugin] Exiting app because of dependency solver failure...

  apps/isaaclab.python.xr.openxr.kit depended on omni.kit.xr.bundle.generic, a meta-extension Isaac Sim does not ship (0 copies on disk in any install). It resolved only by downloading from the Kit SDK
  registry, so every XR run required registry reachability.

  The 2.3 line declared omni.kit.xr.system.openxr + omni.kit.xr.profile.ar. Kit 110 removed profile.ar; its successor is omni.kit.xr.ui.window.profile, which is shipped. The Isaac Sim 6.0 update
  (0049bb5ca44) pointed at the bundle instead — a one-line change — so the migration to discrete XR extensions stalled halfway. This PR completes it, matching what isaacsim.exp.base.xr.vr.kit declares.

  This is latent since Isaac Sim 6.0, not a regression on develop — release/3.0.0-beta2 carries the identical line. A workstation that reaches the registry caches the bundle on first use and loads it
  locally forever after, which is why it only ever surfaced in containers.

  Two related fixes:

  - --xr without --visualizer kit now always resolves headless. Previously a task config declaring a Kit visualizer left XR non-headless, selecting the GUI experience and leaving /isaaclab/xr/auto_start
  false — so the session waited for a Start XR click that headless users cannot make. This wires up _xr_auto_start, which was computed but never read.
  - AttributeError: 'Namespace' object has no attribute 'headless' in record_demos.py and teleop_se3_agent.py, which still read the --headless flag removed in #6656. Fixed by exposing a public
  AppLauncher.headless.

Fixes # (issue)

## Type of change

## Checklist


rwiltz:
@greptile

rwiltz:
@greptile

source/isaaclab/test/app/test_kwarg_launch.py
```diff
@@ -64,6 +64,55 @@ def test_explicit_experience_requires_isaac_sim_runtime():
...
+    """Test that XR runs headless without an explicit Kit visualizer, and auto-starts when it does.
+
+    A task config declaring a Kit visualizer must not leave ``--xr`` non-headless, since the XR
+    session auto-starts. Conversely ``HEADLESS=1`` and livestreaming run windowless even when
+    ``--visualizer kit`` was requested, so auto-start must follow the *resolved* headless state
+    rather than the visualizer selection alone -- otherwise the session waits forever for a
+    "Start XR" click that no window can deliver. Resolution is exercised directly to avoid
+    launching Isaac Sim.
+    """
+    monkeypatch.setenv("HEADLESS", str(headless_env))
```
hougantc-nvda, line 106: **`HEADLESS` is pinned now, but `XR` still leaks in from the environment.**

`_resolve_xr_settings` reads `XR` straight from `os.environ`, and this test doesn't pin it. The last row — `({}, 0, 0, False)` — is the one that breaks: with `XR=1` exported, `_xr` resolves True, `_xr_implies_headless` becomes True, and `_headless` comes out `True` against an expected `False`.

Re-probed against this head:

| ambient env | resolved `_headless` | expected | result |
| --- | --- | --- | --- |
| `HEADLESS=1` | `False` | `False` | pass (fixed by the new `setenv`) |
| `XR=1` | `True` | `False` | fail |

`XR=1` is routinely exported by people working on exactly this feature, so this fails locally for the most likely readers of the test and passes in CI — an annoying one to chase down.

`test_limit_cpu_threads_forwarded_to_simulation_app` (line 204) already pins all three:

```python
monkeypatch.setenv("HEADLESS", "0")
monkeypatch.setenv("LIVESTREAM", "0")
monkeypatch.setenv("XR", "0")
```

Following that precedent here would just mean adding `monkeypatch.setenv("XR", "0")` next to the existing `setenv`.

scripts/tools/record_demos.py
```diff
@@ -656,7 +656,7 @@ def stop_recording_instance():
     # turns it into one env reset) and pauses the session -- binding it straight to
     # ``reset_recording_instance`` would reset the env twice.
     control_keyboard = None
-    if use_isaac_teleop and not args_cli.headless:
+    if use_isaac_teleop and not app_launcher.headless:
```
hougantc-nvda: **`headless` is the wrong predicate for "can the user press keys" — this silently drops the control keyboard under `--livestream`.**

Livestreaming always forces `_headless = True` (`app_launcher.py:857`), so `--livestream 1|2` gets no B/P/R bindings at all — even though a streamed Kit window with input forwarding is exactly where an operator would want them. Same at `teleop_se3_agent.py:442`.

What convinces me this is a real gap rather than a judgement call is that the new docstring in this same PR now says it outright:

> Note that a headless host may still drive a remote UI, as livestreaming does.

So the property's own documentation warns against the inference these two call sites are making. The rest of the codebase agrees: `hide_ui` stays `False` under livestream (`app_launcher.py:1009`), and `/isaaclab/has_gui` is published as `not headless or livestream >= 1 or xr` (`app_launcher.py:1285`).

The underlying problem is that `AppLauncher` exposes only `app` and `headless`, so a consumer has no way to express "is there a UI". Two options:

1. Expose `has_gui` as a property alongside `headless`, mirroring the expression already published at line 1285, and gate both call sites on that. My preference — the new public surface then matches what consumers actually need to branch on, and this is the moment to get it right while `headless` is still new.
2. Or gate on `not app_launcher.headless or app_launcher.livestream` locally, which needs a `livestream` accessor too.

Happy for this to be a follow-up if you'd rather keep this PR to the XR startup fix — it's pre-existing in the sense that nothing here made livestream worse. But the new property is what future callers will reach for, so it's cheap to settle now.

hougantc-nvda (commented):
Reviewed at head `8891a945` ("Key XR auto-start off the resolved headless state"). I'd started against `da682bdc` and re-ran everything after the push.

**Both P1s from the previous head are fixed, and I verified the fixes rather than taking them on faith.** I re-ran the resolvers standalone against this head:

- `/isaaclab/xr/auto_start` is now `self._xr and self._headless`, so `HEADLESS=1` (and `headless=True`) with `--xr --visualizer kit` resolves to `headless=True, auto_start=True`. That was `auto_start=False` before — the case `isaaclab-review-bot` flagged. Fixed, and the two new matrix rows cover it.
- The class-level `_xr_implies_headless: bool = False` at line 651 fixes a breakage that hadn't been reported anywhere: at the previous head, `_xr_auto_start` was assigned only in `_resolve_xr_settings`, so `_resolve_headless_for_case()` (which skips XR resolution) raised `AttributeError` — 10 test functions through that helper, plus 6 parametrized `_load_extensions` cases. Re-probing this head, that path now resolves cleanly. The comment explaining the default is the right call.

The docstring rewrite also addresses the bot's other thread.

Three smaller things left, below, plus nits. Caveat on my verification: this is all static/control-flow analysis — I have **not** run a real offline XR launch, since that needs a headset, so the end-to-end claim in the changelog is still unverified from my side.

### Nits

- **`app_launcher.py:890`** — the XR force-headless path logs at DEBUG, while the sibling force-headless branch at `:909` logs at INFO. A user whose task config requests Kit gets a headless run with no visible explanation of why their window vanished. Suggest matching the INFO precedent.
- **`test_teleop_scripts_smoke.py:182`** — the regex scans the whole `.kit` file rather than scoping to the `[dependencies]` table. Harmless today (all six matches happen to be dependencies), but it'll pick up unrelated quoted keys if the file grows.
- **`test_kwarg_launch.py:99`** — `object.__new__(AppLauncher)` and `livestream_env=-1`; every other test in this file uses `AppLauncher.__new__(AppLauncher)` and `livestream_env=0`. Worth matching for consistency.

Also confirmed fine, for the record: the extension-dependency fix propagates correctly (`isaaclab.python.xr.openxr.headless.kit` inherits from the patched experience, so the headless XR path is covered); the `.skip` changelog fragment matches existing convention; and `_assert_started_cleanly` does use `any(...)`, so the new smoke test's "either marker is accepted" docstring is accurate.

scripts/tools/record_demos.py
hougantc-nvda: Correction to this comment: I claimed `AppLauncher` exposed only `app` and `headless` and suggested adding a `has_gui` property. That was wrong — `AppLauncher.has_gui()` already existed at `app_launcher.py:395`; my search filtered on `@property` and skipped the `@classmethod`.

It was also the wrong suggestion on the substance. `has_gui()` is `True` for headless XR, so gating on it would have constructed a keyboard in a run with no Kit window to capture keys. The `has_window` property you added is the correct predicate, and the livestream case is now handled at both call sites. Sorry for sending you after something that already existed.

source/isaaclab/test/app/test_kwarg_launch.py
```diff
@@ -64,6 +64,68 @@ def test_explicit_experience_requires_isaac_sim_runtime():
...
+
+    launcher._resolve_visualizer_settings(args)
+    launcher._resolve_xr_settings(args)
+    launcher._resolve_headless_settings(args, livestream_arg=-1, livestream_env=0)
+
+    assert launcher._headless is expected_headless
+    assert launcher.has_window is expected_has_window
+    # This is the expression published as ``/isaaclab/xr/auto_start``: the session must start
+    # itself exactly when XR is on and no window exists to start it from.
+    assert (launcher._xr and launcher._headless) is (bool(launcher_args.get("xr")) and expected_headless)
```
hougantc-nvda: **Nothing asserts the published `/isaaclab/xr/auto_start` value.**

This assert re-derives `(launcher._xr and launcher._headless)` from resolver state and compares it against the same expression rebuilt from the parameters, so it never reaches `_load_extensions` — the setting this PR exists to get right has no regression test at its publication site.

Demonstrated by replacing the publisher in memory so it always wrote `False`: all seven rows here, plus all four `test_load_extensions_publishes_has_gui_setting` rows, still passed — headless XR included. A regression at `app_launcher.py:1309` would escape the suite entirely.

`test_load_extensions_publishes_has_gui_setting` already uses `_DummySettings` and already parametrizes exactly `headless x livestream x xr`, so this is a one-liner there:

```python
assert settings.values["/isaaclab/xr/auto_start"] is (xr and headless)
```

hougantc-nvda (commented):
Re-reviewed at `d78e6584`. Everything from my earlier review is addressed — `has_window` at both call sites, the `XR` env pin, the INFO log with the `--viz kit` hint, the section-scoped `_kit_dependencies()` parser, and the fixture normalization. I've corrected and resolved the `record_demos.py` thread, where my `has_gui` suggestion was wrong on both counts.

One finding left, inline.

Test runs, for the record: `test_kwarg_launch.py` gives **44 passed / 2 failed** at `d78e6584` and **37 passed / 2 failed** at the merge base — the same two failures both times (`test_livestream_launch_with_kwargs`, `test_matrix_headless_with_viz_names_takes_precedence`). Pre-existing, not attributable to this PR, but worth knowing they're red on a file this PR touches.

On the core premise, two independent passes verified it further than I did: the pinned `isaacsim-extscache-kit==6.0.1.0` wheel was opened directly and ships `omni.kit.xr.core`, `omni.kit.xr.system.openxr`, and `omni.kit.xr.ui.window.profile` while not shipping `omni.kit.xr.bundle.generic`; and Isaac Sim's own `source/apps/isaacsim.exp.base.xr.vr.kit` declares exactly the same three. A real offline XR session is still unverified here — that needs a headset — so CI wasn't treated as proof of it.

hougantc-nvda (approved):
LGTM

kellyguo11 (approved):

PR #6895, opened 2026-08-04
Defer USD imports until Kit startup

TWpoint:
## Summary

- defer importing `UsdReplicateContext` until replication runs against a live stage
- keep resolving `queue_replication` free of `pxr` imports before Kit startup
- add a changelog fragment for the startup fix

## Root cause

Environment configuration imports resolve `queue_replication` before `SimulationApp` starts. `replicate_session` previously imported `.usd` at module load time, which eagerly loaded the kit-less OpenUSD `pxr` runtime. When Kit subsequently initialized `omni.usd.libs`, both USD runtimes were present in the process and startup could fail with `free(): invalid pointer`.

The USD replication context is only needed when `replicate()` operates on an initialized stage, so importing it at that boundary prevents the premature runtime load without changing the public API or replication behavior.

## Impact

Kit-backed applications can import task and asset configuration modules before startup without resolving `queue_replication` preloading the kit-less OpenUSD runtime. Kit-less behavior and the replication queue API remain unchanged.

## Validation

- `uv run python -m pytest source/isaaclab/test/cloner/test_clone_plan_algebra.py -q` (`81 passed`)
- `uv run python tools/changelog/cli.py check develop`
- full repository pre-commit suite via `uv run isaaclab -f`
- downstream Newton MJWarp evaluation with the Kit visualizer successfully initialized the scene and loaded its checkpoint

### DIFF
--- a/source/isaaclab/changelog.d/twpoint-pxr-import.rst
+++ b/source/isaaclab/changelog.d/twpoint-pxr-import.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed resolving :func:`~isaaclab.cloner.queue_replication` before Kit startup
+  so it no longer preloads the kit-less OpenUSD runtime and corrupts Kit's USD runtime.
--- a/source/isaaclab/isaaclab/cloner/replicate_session.py
+++ b/source/isaaclab/isaaclab/cloner/replicate_session.py
@@ -17,7 +17,6 @@ from isaaclab.utils.version import has_kit
 
 from .clone_plan import make_clone_plan
 from .cloner_strategies import sequential
-from .usd import UsdReplicateContext
 
 if TYPE_CHECKING:
     import torch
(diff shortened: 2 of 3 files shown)


AntoineRichard (changes requested):
Thank you for tracking this down and for the focused fix. Deferring the USD import until replication runs is a useful change.

Please remove the newly added test_queue_replication_imports_without_kit test. Although it detects this specific transitive import through sys.modules, it is a proxy for the reported failure and does not exercise the actual Kit startup/runtime conflict. I would prefer not to add this implementation-detail test as part of the fix.

TWpoint:
Hi @AntoineRichard, I’ve addressed your feedback and removed the runtime metadata test. Could you please take another look? Thanks!

TWpoint:
Hi maintainers — I checked the remaining merge blockers on this PR.

- The change requested by @AntoineRichard has been addressed by commit `1d9135c`: the proxy runtime metadata test was removed. The current diff contains only the deferred production import and changelog fragment, with no unresolved review threads.
- The PR is mergeable with no conflicts, but the old `CHANGES_REQUESTED` review is still active pending re-review.
- The latest fork workflows are all `action_required`; contributors cannot approve or run them in the upstream repository.
- The repository rules also require two approvals, including the designated team approval.

Could @AntoineRichard please update the review, and could a maintainer approve the pending workflow runs and provide the remaining required team review? Thanks!

PR #6900, opened 2026-08-04
[Warp] Restore --frontend warp on Reach and gate the supported task set in CI

hujc7:
## 1. Summary

- `--frontend warp` fails at env construction for `Isaac-Reach-Franka` and `Isaac-Reach-UR10`: `rewards.success` and `terminations.success` resolve to `is_terminated_term` and `pose_command_success`, which had no warp twins. Adds both, matched to the stable terms and validated by training.
- The warp-supported task list was hand-maintained and had already drifted — sweeping the registry finds 4 contrib velocity tasks that adapt but were never listed (10 → 14). It is now derived, not declared.
- Adds a **CUDA-graph capture-safety gate** for warp MDP terms. Managers run captured by default, so capture safety is the normal path; the gate captures a term, overwrites its inputs in place, replays, and compares against stable on the new data.
- `source/isaaclab_experimental/` had no CI job at all: all 9 of its test files were collected by nothing. Test files matched by no job: 13 → 4.
- Also restores `Isaac-Ant` / `Isaac-Humanoid`, which lost warp support on develop when the locomotion rewards gained `terminated_penalty` and `survival_success_rate` without twins — the sweep in this PR is what surfaced it.
- Package suite: 203 passed, 1 skipped, ~10 s. NVBug 6558734.

## 2. Why the existing gate did not catch it

`test_frontend_cfg_conversion.py` already asserted both Reach tasks adapt, and has since #5504 — it just never ran. Jobs either name a package with a positive `filter-pattern`, or pass `not isaaclab_`, which `run_tests.sh` converts into an exclude on `isaaclab_`. This package matched neither form.

## 3. Twin semantics

`pose_command_success` recomputes the pose error in a kernel rather than copying the stable result, whose buffer is reallocated per call and would go stale under graph replay. Three points where a naive twin diverges from the stable term, each with a regression test:

| Behaviour | Stable | Naive twin |
|---|---|---|
| Sticky per-episode success | ORed inside `compute_success()`, before reset | never set; `reset()` reads and clears first, so `Metrics/success_rate` reads 0 |
| Non-finite error | `error < bound` is false → denies success | reject form `error >= bound` is false → **reports success** |
| Negative threshold | configured (`is not None`) → denies success | read as "unset" → ignored |

The sticky write happens inside the kernel, not after the launch: host-side work between launches is not replayed under capture.

Thresholds are resolved at init. A curriculum mutating them at runtime is not honoured, and cannot be while the manager is captured — a kernel scalar is baked into the graph. Recorded in the class docstring.

## 4. Capture-safety gate

Terms are discovered from the shared warp MDP modules **and every per-task mirror** under `isaaclab_tasks_experimental`, rather than listed, so a new term without a declaration fails the gate. Enumerating separated three groups convention had blurred: terms already covered by the hand-written mutation classes, three observation terms with no capture coverage anywhere, and eight event terms that cannot be captured at all (variable `env_ids` launch dim at reset cadence).

Capture-and-replay alone does not prove safety — replaying against unchanged memory re-reads the same valid bytes. Injecting a stale-view bug (`wp.clone` instead of a zero-copy view) fails only `test_term_is_capture_safe[pose_command_success]`; all 17 other tests pass.

## 5. Test plan

- [x] `pytest source/isaaclab_experimental/test/` — 201 passed, 1 skipped
- [x] Reverting the twins fails the sweep, naming both tasks and both missing symbols
- [x] Mutation checks: stale view → only the capture test fails; reject-form predicate → only the NaN test; dropped sticky write → the threshold tests; wxyz quaternion → only the orientation cases
- [x] Training, 4096 envs / 150 iters / `--frontend warp presets=newton_mjwarp`:

| Task | Training time | Reward start → end | `Metrics/success_rate` |
|---|---|---|---|
| `Isaac-Reach-Franka` | 91.76 s | −0.63 → −0.49 | 0.78 |
| `Isaac-Reach-UR10` | 72.25 s | −2.43 → 0.03 | 0.99 |

  With the sticky write reverted, Franka trains identically (episode length still falls 360 → 252, so terminations fire) but `Metrics/success_rate` stays 0.0000 for all 150 iterations — the metric bug in isolation.
- [x] `uv run isaaclab -f` clean

## 6. Out of scope

- `source/isaaclab_ppisp/` has the same CI hole (4 uncollected test files).
- Three observation terms have no capture coverage yet: `joint_pos_rel`, `joint_vel_rel`, `body_incoming_wrench`.
- `isaaclab_tasks_experimental` has no test directory, so its twenty pre-existing per-task twins have no parity or capture coverage anywhere. They are recorded in the gate's backlog rather than assumed correct.
- Warp orientation-error twins use `2*acos(|w|)` where the stable path normalizes first; the two differ by ~7e-4 rad at float32, far below any configured threshold. Pre-existing in `orientation_command_error`; left consistent rather than changed in one place.


source/isaaclab_experimental/isaaclab_experimental/envs/mdp/rewards.py
```diff
@@ -68,6 +69,62 @@ def is_terminated(env: ManagerBasedRLEnv, out) -> None:
...
+        )
+
+    def __call__(
+        self, env: ManagerBasedRLEnv, out: wp.array(dtype=wp.float32), term_keys: str | list[str] = ".*"
+    ) -> None:
+        wp.launch(
+            kernel=_is_terminated_term_kernel,
+            dim=env.num_envs,
+            inputs=[
+                env.termination_manager.term_dones_wp,
```
StafaH, line 119: Let's avoid the _wp suffiex (it might be already there from before but it's cleaner without it)

source/isaaclab_experimental/isaaclab_experimental/envs/mdp/rewards.py
```diff
@@ -68,6 +69,62 @@ def is_terminated(env: ManagerBasedRLEnv, out) -> None:
     )
 
 
+@wp.kernel
+def _is_terminated_term_kernel(
```
StafaH, line 73: This could actually be a 2D kernel, or even a tiled kernel. The tile is of shape tile_mask, and we can use tile_sum() or a tile_map(). 

(You can ignore this comment for now, I noticed that many of the kernels could be written in a much more optimized way, which we can tackle later)

source/isaaclab_experimental/isaaclab_experimental/envs/mdp/terminations.py
```diff
@@ -49,6 +50,114 @@ def time_out(env: ManagerBasedRLEnv, out) -> None:
...
+    Each threshold is an accept predicate, matching the stable term's ``error < threshold``: a
+    non-finite error compares false and denies success, where a reject form would let it through.
+    """
+    i = wp.tid()
+    success = bool(True)
+    if check_position:
+        des_b = wp.vec3f(cmd[i, 0], cmd[i, 1], cmd[i, 2])
+        des_w = root_pos_w[i] + wp.quat_rotate(root_quat_w[i], des_b)
+        cur_w = body_pos_w[i, body_idx]
+        dx = cur_w[0] - des_w[0]
```
StafaH, line 79: cur_w - des_w?

source/isaaclab_experimental/isaaclab_experimental/envs/mdp/terminations.py
```diff
@@ -49,6 +50,114 @@ def time_out(env: ManagerBasedRLEnv, out) -> None:
...
+    """
+
+    def __init__(self, cfg: TerminationTermCfg, env: ManagerBasedRLEnv):
+        super().__init__(cfg, env)
+        command_name = cfg.params["command_name"]
+        command = env.command_manager.get_term(command_name)
+        # carry "configured" as its own flag: a threshold is any float, so no value can encode absence
+        position_threshold = command.cfg.position_success_threshold
+        orientation_threshold = command.cfg.orientation_success_threshold
+        self._check_position = position_threshold is not None
```
StafaH, line 122: One thing you can do here and elsewhere is that thing like _check_position can actually be used as wp.static inside the kernel definition. To do that you would need to define the kernel here inside init or in _call__.

source/isaaclab_experimental/test/envs/mdp/parity_helpers.py
```diff
@@ -542,6 +550,109 @@ def get_term(self, name: str):
...
+
+    Returns:
+        The command term, whose ``pose_command_b`` is the (num_envs, 7) command buffer.
+    """
+    from isaaclab.envs.mdp.commands.commands_cfg import UniformPoseCommandCfg
+    from isaaclab.envs.mdp.commands.pose_command import UniformPoseCommand
+
+    rng = np.random.RandomState(seed)
+    data = articulation.data
+    root_pos = data.root_pos_w.torch.cpu().numpy()
```
StafaH, line 602: This looks weird. Whats wrong with root_pos_w.numpy()?

source/isaaclab_experimental/test/envs/mdp/parity_helpers.py
```diff
@@ -542,6 +550,109 @@ def get_term(self, name: str):
...
+    reimplementation of it.
+
+    The command is derived from the body's current pose plus a known error, so the
+    resulting position and orientation errors straddle
+    :data:`POSE_CMD_POSITION_SCALE` and :data:`POSE_CMD_ORIENTATION_SCALE`.
+
+    Returns:
+        The command term, whose ``pose_command_b`` is the (num_envs, 7) command buffer.
+    """
+    from isaaclab.envs.mdp.commands.commands_cfg import UniformPoseCommandCfg
```
StafaH, line 597: Can we put all imports at the top of file.

source/isaaclab_experimental/test/envs/mdp/parity_helpers.py
```diff
@@ -542,6 +550,109 @@ def get_term(self, name: str):
...
+    term.body_idx = body_idx
+    term.pose_command_b = torch.tensor(command_np, device=device)
+    term.pose_command_w = torch.zeros_like(term.pose_command_b)
+    term._env = SimpleNamespace(num_envs=num_envs, device=device)
+    term._track_success = position_success_threshold is not None or orientation_success_threshold is not None
+    term._succeeded = torch.zeros(num_envs, dtype=torch.bool, device=device)
+    return term
+
+
+class MockPoseCommandManager:
```
StafaH, line 643: HIgh level question, why do we need mocks? I prefer that tests use the original code if it is cheap to do so. The only exception are really expensive classes like Newton Model

source/isaaclab_experimental/test/envs/mdp/parity_helpers.py
```diff
@@ -564,12 +675,34 @@ def __init__(self, name="contact_sensor", body_ids=None, device=DEVICE):
...
+
+    def __init__(self, num_envs=NUM_ENVS, device=DEVICE, term_names=("success", "fell_over", "time_out"), seed=31):
         self.terminated = torch.zeros(num_envs, dtype=torch.bool, device=device)
         self.terminated_wp = wp.from_torch(self.terminated)
 
+        self.active_terms = list(term_names)
+        rng = np.random.RandomState(seed)
+        dones_np = rng.rand(num_envs, len(self.active_terms)) < 0.3
+        self.term_dones = torch.tensor(dones_np, dtype=torch.bool, device=device)
+        self.term_dones_wp = wp.from_torch(self.term_dones)
```
StafaH, line 692: Proxy Arrays here and elsewhere?

source/isaaclab_experimental/test/envs/test_frontend_cfg_conversion.py
```diff
@@ -33,22 +34,33 @@
...
+from isaaclab_tasks.utils.parse_cfg import load_cfg_from_registry
+
+# Stable manager-based tasks resolve to this env class; direct tasks provide their own and
+# take the :meth:`WarpFrontend._resolve_direct_warp_class` path instead of cfg adaptation.
+_STABLE_MANAGER_ENTRY_POINT = "isaaclab.envs:ManagerBasedRLEnv"
+
+# Stable task ids that adapt cleanly under ``--frontend warp``, as produced by
+# :func:`_sweep_warp_support`. Do not curate this by hand: when the test fails it prints the
+# exact set to paste back.
+_WARP_SUPPORTED_TASKS = frozenset(
```
StafaH, line 46: Any way we can track the warp supported envs automatically?

source/isaaclab_experimental/test/envs/test_frontend_cfg_conversion.py
```diff
@@ -82,6 +94,76 @@ def _load_adapted_cfg(cfg_entry_point: str):
...
+            incompatible[task_id] = reason
+    return frozenset(supported), incompatible, unimportable
+
+
+def _format_task_set(task_ids) -> str:
+    """Render a task-id set as a paste-ready literal for :data:`_WARP_SUPPORTED_TASKS`."""
+    return "{\n" + "".join(f"    {task_id!r},\n" for task_id in sorted(task_ids)) + "}"
+
+
+def test_warp_supported_task_set_matches_the_registry():
```
StafaH, line 137: Ignore above comment, it looks like we already are searching automatically. So why do we need to compare to a hardcoded list

source/isaaclab_tasks_experimental/isaaclab_tasks_experimental/core/locomotion/mdp/rewards.py
```diff
@@ -325,3 +325,106 @@ def __call__(
...
+    """Atomically count masked (just-reset) envs and how many of them timed out."""
+    env_index = wp.tid()
+    if env_mask[env_index]:
+        wp.atomic_add(counts, 0, 1)
+        if time_outs[env_index]:
+            wp.atomic_add(counts, 1, 1)
+
+
+@wp.kernel
+def _survival_rate_kernel(
```
StafaH, line 379: This is not even a real kernel?

StafaH (approved):

source/isaaclab_experimental/test/envs/test_frontend_cfg_conversion.py
hujc7: The goal is to enforce support. If it's resolved dynamically, it also means that torch mdp changes could break warp support. So the goal is to gradually add all parity mdps. And if we have all the envs supported, we can flip the test to ensure all envs are supported.

PR #6905, opened 2026-08-05
Add OVRTX shader cache for CI rendering tests

mataylor-nvidia:
## Summary

Each CI run starts with an empty `nv_shadercache` directory, so the NVIDIA driver recompiles every pipeline state object (PSO) before the first render — roughly 600 s, which is what `COLD_CACHE_BUFFER = 700` in `tools/conftest.py` exists to absorb. The cost is paid on every rendering job, twice per run (kit-based and kitless).

This persists the driver's PSO blobs across runs with GitHub Actions cache, so warm runs skip the compile entirely.

Two independent cache trees live under `${RUNNER_TEMP}/isaaclab-ovrtx-shader-cache/`, because the two render paths are compiled by different runtimes and a bump to one must not invalidate the other:

| Tree | Produced by | Reaches the container as |
|---|---|---|
| `kit/` | RTX renderer inside the Isaac Sim image | nested bind mount over `/isaac-sim/kit/cache/nv_shadercache`, which the enclosing `kit/cache` tmpdir mount would otherwise present as empty |
| `kitless/` | pip-installed `ovrtx` wheel | `OVRTX_SHADER_CACHE_PATH`, applied by `OVRTXRenderer` as a carb setting |

### `.github/actions/ovrtx-shader-cache/` (new)

One action owns the whole lifecycle — key computation, restore, reporting and writeback — so callers only decide *when* each phase runs. A composite action cannot span the caller's test step, so the phases are separate invocations selected by `mode`:

- `restore` — reads the newest compatible snapshot of each tree, never writes
- `report` — summarises how far the run compiled beyond what was restored
- `save` — reports, then writes each **populated** tree back as a new snapshot; only the warmer uses it

Every mode recomputes keys from the same `key.sh`, so the collection a job reads and the one it writes cannot drift.

Cache key composition:

| Component | Source | Why |
|---|---|---|
| `RUNNER_OS`, `RUNNER_ARCH` | GHA env | platform |
| `sm<cc>` | `nvidia-smi --query-gpu=compute_cap` | PSO blobs are GPU-architecture specific |
| `drv<version>` | `nvidia-smi --query-gpu=driver_version` | PSO blobs are driver specific |
| `isaacsim<tag>` (kit only) | `isaacsim-version` input | identifies the Kit RTX build that compiled `kit/` |
| `ovrtx<version>` (kitless only) | `uv.lock` pin | keeps the key a function of the commit, so an upstream wheel release cannot re-key open PRs onto a cold collection |

`verify.sh` fails the warmer when either tree came out empty, which would otherwise publish a silent half-warm snapshot that every consumer prefers over the last good one.

### CI wiring

- `run-package-tests`: new `ovrtx-shader-cache` input (`''` / `restore` / `save`); restores before the tests, reports or publishes after. Publish is gated on `!cancelled()` rather than `success()` — compiled PSO blobs stay valid when a golden-image assertion fails.
- `run-tests` / `run_tests.sh`: new `ovrtx-shader-cache-host-dir` input; owns the canonical mount layout and runs the pre-test mount check.
- `build.yaml`: `ovrtx-shader-cache: restore` on `rendering-correctness`, `rendering-correctness-kitless-legacy` and `rendering-correctness-kitless-ovstage`; new post-merge `warm-ovrtx-cache` job (`continue-on-error`, publishes only on `main` / `develop` / `release/*`) that renders one Kit and one kitless cartpole case to fill both trees.

### Source

- `source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_shader_cache.py` (new): reads `OVRTX_SHADER_CACHE_PATH` and points both `/rtx/shaderDb/driverShaderCachePath` and `/rtx/shaderDb/driverAppShaderCachePath` at it via the ovrtx settings extension. No-ops when the variable is unset; warns when the runtime has no settings extension; raises when a setting is rejected, since a partial redirect still compiles cold. Kept out of `ovrtx_renderer` so the policy imports without the ovrtx runtime.
- `ovrtx_renderer.py`: calls `redirect_shader_cache(OVRTX_CONFIG)` before constructing `Renderer()` — the settings are read at renderer-creation time and cannot be changed on a live instance.
- `tools/verify_ovrtx_shader_cache.py` (new): in-container check that the mounted trees are present and writable, run before tests start. A shadowed `kit/` mount is otherwise invisible — the restore step still reports a hit and the only symptom is a slower job.
- `tools/conftest.py`: `COLD_CACHE_BUFFER` docstring condensed to its functional contract; the value is unchanged.

### Tests

`source/isaaclab_ov/test/test_ovrtx_shader_cache_redirect.py` covers the redirect boundary with a fake settings applier — no GPU, no ovrtx runtime, no renderer: both settings redirected, rejection raises and is not logged as success, unset env var queries nothing, the renderer config reaches the applier (it is what initializes the library), and a runtime without the extension degrades to a warning.

## Type of change

## Release backport

## Test plan

- [ ] `warm-ovrtx-cache` runs post-merge and publishes both entries — step summary shows a non-zero file count for `kit/` and `kitless/`
- [ ] Subsequent `rendering-correctness` and `rendering-correctness-kitless-*` runs restore the cache and report `fully covered`
- [ ] The `COLD_CACHE_BUFFER` allowance is no longer consumed on warm runs — the first camera test completes within its normal budget
- [ ] Cold run (ovrtx bump or driver bump) still succeeds: cache miss → compile → warmer publishes into the new collection

## Checklist

- [x] I have added a changelog fragment under `source/isaaclab_ov/changelog.d/`


mataylor-nvidia:
@greptile review

mataylor-nvidia:
How cache key is generated

| Component | Source | Purpose |
|---|---|---|
| `RUNNER_OS` | GHA env | Linux vs Windows |
| `RUNNER_ARCH` | GHA env | CPU arch (X64, ARM64) |
| `sm${gpu_arch}` | `nvidia-smi --query-gpu=compute_cap` | GPU SM generation |
| `ovrtx{version}` | `uv.lock` pinned version | Full semver of the `ovrtx` package |
| `drv{driver_ver}` | `nvidia-smi --query-gpu=driver_version` | NV driver version |

.github/actions/ovrtx-shader-cache-key/action.yml
```diff
@@ -0,0 +1,68 @@
...
+        # nvidia-smi must be available on every runner that has a GPU.
+        driver_ver="$(nvidia-smi --query-gpu=driver_version --format=csv,noheader 2>/dev/null | head -1 || echo unknown)"
+
+        # SM compute capability (e.g. "8.0", "9.0") identifies the GPU
+        # architecture. PSO blobs compiled for one SM generation are not
+        # compatible with another, so heterogeneous runner fleets require
+        # this in the key to prevent cross-architecture cache hits.
+        gpu_arch="$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -1 | tr -d '.' || echo unknown)"
+
+        collection="ovrtx-v1-${RUNNER_OS}-${RUNNER_ARCH}-sm${gpu_arch}-${ovrtx_ver}-drv${driver_ver}"
```
AntoineRichard: **AI-generated review — Important: key each cache by the runtime that produces it.**

This collection uses the uv.lock OVRTX version for both subtrees. However, kit/ is produced by Kit/Isaac RTX (test-rendering-correctness does not install OVRTX), while kitless jobs install the ovrtx>=0.4.0,<0.5.0 range emitted by resolve-ov-pins, not necessarily the lockfile version. A Kit upgrade with a stable OVRTX entry can reuse the wrong collection, and a newly released OVRTX wheel can be saved under the old version key.

Please split the cache entries: key kit/ by the exact Isaac Sim/Kit RTX build and kitless/ by the exact OVRTX wheel installed, retaining driver and SM in both keys.

.github/workflows/build.yaml
```diff
@@ -890,6 +892,50 @@ jobs:
...
+        # Kit-based cartpole rendering warms kit/nv_shadercache; kitless ovrtx
+        # cartpole rendering warms kitless/nv_shadercache. One combined job
+        # covers both sub-trees in a single container run.
+        filter-pattern: "isaaclab_tasks"
+        extra-pip-packages: "${{ steps.ov_pins.outputs.ovrtx }} ${{ steps.ov_pins.outputs.ovphysx }}"
+        include-files: >-
+          test_rendering_cartpole.py,
+          test_rendering_cartpole_kitless.py
+        test-k-expr: >-
+          (newton_warp or ovrtx) and (rgb or depth)
```
AntoineRichard: **AI-generated review — Important: this warmer selects the wrong renderer cases.**

The newton_warp IDs map to newton_renderer, not Kit RTX or OVRTX. Consequently the Kit file never exercises isaacsim_rtx, while the kitless file runs four Newton Warp RGB/depth cases that cannot populate the OVRTX subtree. That both misses the intended Kit render path and spends GPU time on unrelated tests.

Use (isaacsim_rtx or ovrtx) and (rgb or depth) or exact node IDs so every selected case contributes to one of the two target caches.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -125,6 +125,79 @@
...
+            f"--/rtx/shaderDb/driverShaderCachePath={cache_path}",
+            f"--/rtx/shaderDb/driverAppShaderCachePath={cache_path}",
+        ):
+            r = vtable.apply_settings(_ovrtx_bindings.ovx_string_t(setting))
+            if r.status != _ovrtx_bindings.OVRTX_API_SUCCESS:
+                logger.warning(
+                    "Failed to apply OVRTX carb setting %r — shader cache may still go to default path.", setting
+                )
+        logger.info("OVRTX kitless nv_shadercache redirected to %r.", cache_path)
+    except Exception as exc:
```
AntoineRichard: **AI-generated review — Important: do not hide redirect regressions.**

Any programming error or private-binding ABI change in this block is downgraded to a warning, after which rendering succeeds cold. verify_ovrtx_shader_cache.py only proves that the mount is writable, so CI has no assertion that either setting was applied. The unconditional success log on line 193 is also emitted when one or both calls returned failure.

Catch only expected extension-unavailable errors, return or propagate a failure when the environment variable explicitly requests redirection, and log success only after both calls succeed. Add a non-GPU unit test with fake bindings that asserts both setting strings and covers query/apply failure paths.

.github/actions/run-package-tests/action.yml
```diff
@@ -381,6 +436,47 @@ runs:
...
+        else
+          verdict="compiled $(( grew * 100 / (before > 0 ? before : 1) ))% beyond the restored cache"
+        fi
+
+        echo "OVRTX shader cache: restored ${before} MB, ended at ${after} MB (+${grew} MB) - ${verdict}"
+        echo "🔵 OVRTX shader cache: ${before} -> ${after} MB (+${grew}) - ${verdict}" >> "$GITHUB_STEP_SUMMARY"
+        python3 .github/actions/run-package-tests/ovrtx_shader_cache_inventory.py \
+          "$HOST_DIR" "OVRTX shader cache final"
+
+        [ "$after" -lt 1 ] && echo "empty=true" >> "$GITHUB_OUTPUT" || echo "empty=false" >> "$GITHUB_OUTPUT"
```
AntoineRichard: **AI-generated review — Important: validate both subtrees before publishing.**

This total-size check allows save when only one subtree contains data. A broken kitless redirect can therefore be masked by a non-empty Kit cache (or vice versa), and the warmer publishes a permanently partial snapshot whose missing path still compiles cold.

Have the inventory step expose per-subtree file/byte counts and require both kit/ and kitless/ to be non-empty in save mode. This also gives the warm job a functional assertion that its selected tests actually exercised both cache paths.

.github/actions/run-package-tests/action.yml
```diff
@@ -232,6 +245,47 @@ runs:
 
         python3 .github/actions/run-package-tests/warp_cache_inventory.py "$HOST_DIR" "Warp cache restored"
 
+    - name: Resolve OVRTX shader cache key
```
AntoineRichard: **AI-generated review — Moderate: factor the duplicated cache state machine.**

The resolve/restore/hit-report block here and the growth/empty/save block at lines 439–478 duplicate the adjacent Warp implementation with label substitutions. Cache correctness fixes now require coordinated edits in two long YAML paths, which is precisely where subtle drift occurs.

Extract a reusable composite action or script for restore metadata, growth reporting, empty gating, and save semantics, parameterized by cache name, host directory, initialization callback/directories, and inventory command. Keep only OVRTX-specific mount/layout behavior in this action.

.github/actions/run-package-tests/action.yml
```diff
@@ -107,6 +107,19 @@ inputs:
       post-merge a cumulative writeback; only warm-warp-cache uses it.
     default: ''
     required: false
+  ovrtx-shader-cache:
+    description: >-
+      OVRTX shader cache mode. Empty disables the cache. 'restore' reads the
+      newest compatible snapshot and never writes. 'save' restores and writes
+      back a new snapshot; only warm-ovrtx-cache uses it.
+
+      The cache covers two sub-trees:
```
AntoineRichard: **AI-generated review — Moderate: keep this input description to its public contract.**

The kit/ and kitless/ mount layout is repeated in this input, the key action description, the run-tests input, the shell mount comments, and the verification script. These copies describe implementation rather than how a caller uses this input and will drift together.

Keep only the disabled/restore/save semantics here. Document the directory/mount layout once beside the implementation (or in one design document) and reference that source from the other boundaries.

tools/conftest.py
```diff
@@ -36,10 +36,18 @@ def pytest_ignore_collect(collection_path, config):
...
-run (~600 s).  This buffer prevents that from being misreported as a test
-timeout.  Only the first such test gets the extension — after it runs, the
-on-disk cache is populated.
+The first test that uses ``enable_cameras=True`` may compile NVIDIA Vulkan
+driver PSO shaders (nv_shadercache/GLCache) during its run (~600 s on a cold
+runner).  This buffer prevents that from being misreported as a test timeout.
+Only the first such test gets the extension — after it runs, the on-disk cache
+inside the container is populated and subsequent tests are fast.
+
+When the CI OVRTX shader cache is warm (i.e. ``ovrtx-shader-cache: restore``
```
AntoineRichard: **AI-generated review — Moderate: keep the timeout documentation functional.**

This constant now documents a workflow input, container path, cache invalidation cases, and a measured compile duration. Those details do not affect how COLD_CACHE_BUFFER is used and duplicate the cache action/workflow documentation.

A concise contract is sufficient, for example: Extra seconds granted to the first camera-enabled test for cold shader-cache compilation. Keep cache plumbing and measured timings with the CI cache implementation.

source/isaaclab_ov/changelog.d/mataylor-ovrtx-shader-cache.rst
```diff
@@ -0,0 +1,10 @@
+Added
+^^^^^
+
+* Added ``OVRTX_SHADER_CACHE_PATH`` environment variable support to
+  :class:`~isaaclab_ov.renderers.OVRTXRenderer`. When set, the NVIDIA Vulkan
```
AntoineRichard: **AI-generated review — Moderate: make the changelog entry user-facing and durable.**

The private Carb setting, constructor timing, CI motivation, and ~600 s measurement repeat implementation details already described elsewhere and will age independently. Reduce this to the supported behavior, for example: Added OVRTX_SHADER_CACHE_PATH support to OVRTXRenderer for selecting the NVIDIA driver shader-cache directory.

AntoineRichard (dismissed):
**AI-generated review**

Requesting changes. The main blockers are that the shared collection is not keyed by the exact runtime that produces each subtree, and the warmer selects Newton Warp cases while omitting the Kit RTX renderer it is intended to warm. The current failure handling can also publish a partial cache or silently continue after the renderer redirect breaks.

I also left focused inline requests to remove duplicated cache orchestration and repeated implementation documentation, shorten the timeout/changelog text to functional contracts, and add a non-GPU regression test around the redirect boundary.

Verification: reviewed all 11 changed files plus the renderer matrices, cache helpers, workflow callers, prior review threads, and current CI logs. GitHub pre-commit passed. The two rendering jobs populated their individual cold-cache subtrees, but failed golden-image assertions; the combined warm job is skipped on pull requests, so this PR has not demonstrated that the proposed warmer populates both subtrees. I did not run local tests or pre-commit because benchmarks are active on the machine, and I ran no GPU workload.

.github/actions/run-package-tests/action.yml
```diff
@@ -381,6 +454,105 @@ runs:
...
+    # a golden-image assertion fails. Cancellation still skips, so a tree the
+    # driver was mid-write on is never published.
+    # The count is empty rather than '0' when the growth step aborted before it
+    # counted. Both mean "nothing was measured", so both have to block the
+    # save - an unmeasured tree publishes under a key every consumer prefers over
+    # the last good snapshot. This matches how the verify step below reads the
+    # same outputs.
+    - name: Save OVRTX kit shader cache
+      if: >-
+        !cancelled() && inputs.ovrtx-shader-cache == 'save'
```
StafaH: So this would save even if the PR is failing?

.github/actions/ovrtx-shader-cache-key/action.yml
```diff
@@ -0,0 +1,106 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH: I wonder if we should move more of the logic from run-package-tests into this action to keep more isolation.

Something like:
ovrtx-shader-cache/
  action.yml
  key.sh
  verify.sh

Where the action does:
compute keys
→ restore kit
→ restore kitless
→ prepare directories
→ report hit/miss

then it can be used in run package tests like:

uses: ./.github/actions/ovrtx-shader-cache
with:
  mode: restore   # restore | save
  isaacsim-version: ...
outputs:
  host-dir: ...


just some more seperation

.github/actions/run-tests/run_tests.sh
```diff
@@ -347,7 +376,12 @@ run_tests() {
       fi
       if [ -n \"\${TEST_EXTRA_PIP_PACKAGES:-}\" ]; then
         echo \"Installing extra pip packages: \${TEST_EXTRA_PIP_PACKAGES}\"
-        ./isaaclab.sh -p -m pip install \${TEST_EXTRA_PIP_PACKAGES}
+        # Without the NVIDIA index, ovrtx/ovphysx resolve to the wheel-stub
+        # placeholder sdist on public PyPI, whose build hook re-downloads the
+        # real wheel unretried and hard-asserts its sha256 - a truncated
+        # multi-gigabyte transfer then fails the job with a misleading
+        # \"index not reachable\" error. Naming the index installs the wheel directly.
+        ./isaaclab.sh -p -m pip install --extra-index-url https://pypi.nvidia.com/ \${TEST_EXTRA_PIP_PACKAGES}
```
StafaH: I'm not sure about this, it might cause issues, nvidia packages might differ from pip pacakges

StafaH (commented):
Thanks @mataylor-nvidia! Looks like a good prototype. I would maybe change the architecture to be more like:

build.yaml

── rendering tests
    └── restore-ovrtx-cache
                   ── outputs host-dir

└── warm-ovrtx-cache
              ── restore-ovrtx-cache
              ── run warming workload
              ── verify both shader paths produced actual files
              ── save-ovrtx-cache


Current prototype is also good, but alot of logic is in different places, we could collect all ovrtx shader cache into one action, keep it nice and containted.

We can do similar to the warp cache if theres a mismatch between their implementations.

WDYT?

.github/actions/run-tests/run_tests.sh
mataylor-nvidia: I think I was testing without this and couldn't get past CI

mataylor-nvidia:
Yes I think I can try to unify / condense some of the logic to keep it self contained

warp cache pr for context: https://github.com/isaac-sim/IsaacLab/pull/6809

.github/actions/run-tests/run_tests.sh
mataylor-nvidia: removed

AntoineRichard:
🤖 **AI-assisted review** (batch triage of open CI/infra PRs, with automated code-review/silent-failure/test-coverage passes on a checkout of this branch)

Strong PR — cache staleness is designed out (keys cover OS/arch/`sm<cc>`/driver plus image digest or the uv.lock ovrtx pin, and driver PSO blobs are content-validated so a mismatch degrades to a recompile, not a wrong render), and the restore path is proven in this PR's own CI: the kit job filled `kit/` (+149 MB) and the kitless job filled `kitless/` (+121 MB) with the mount check passing — which also proves the private-ovrtx ctypes redirect works on a real runner.

Two blocking asks, both small:
1. **`.github/actions/ovrtx-shader-cache/action.yml:42-47` — the compute-keys step needs `if: always()`.** The comment at :78-80 states the rule and applies `always()` to the growth/save/verify steps, but not to the compute step they all depend on. On a failed job the caller's `always()`-gated report step runs, compute is skipped, `host-dir` is empty, and `report.sh:22` aborts; in save mode the gates then see empty counts and publish nothing — defeating the `!cancelled()` rationale at `run-package-tests/action.yml:433-436` ('one failing assertion should not discard a whole warm run').
2. **Port the warp-style unchanged-content gate before publishing.** The warp sibling at `run-package-tests/action.yml:404-425` fingerprints and only saves on `changed == 'true'` ('avoid filling the repository quota with duplicate snapshots'); the OVRTX save gates only on non-empty, so the warm job would publish ~270 MB of fresh immutable entries per push to develop/main/release. The Actions cache is already at 10.34 GB / 83 entries against the 10 GB LRU budget, so un-deduplicated pushes would accelerate eviction of the warp cache these same jobs restore (and contend with #7287's new pip/uv entries).

Non-blocking findings:
- `tools/verify_ovrtx_shader_cache.py:33-45` only probes writability, so a shadowed nested `kit/` bind mount still prints 'cache OK' and `report.sh:74` renders 0 growth as 'fully covered' — a host-written sentinel file would cover the case the docstring claims.
- `report.sh:74` uses `-le 0`, conflating 'nothing new compiled' with 'the tree shrank'; the restore path never calls `count_files`, so 'hit reported, tree empty' passes silently.
- `key.sh:35` reads the ovrtx version from `uv.lock` while the container installs the `resolve-ov-pins` value — identical today, diverges if the pin becomes a range spec.
- `ovrtx_shader_cache.py:71` sets `.argtypes` unguarded, so a build lacking the symbol raises AttributeError instead of the documented graceful-None path; the hand-rolled vtable at :58-64 has no version guard (blast radius is CI-only). All five tests monkeypatch `_acquire_settings_applier` away — a cheap importorskip-guarded symbol-shape test would catch wheel bumps.

.github/actions/ovrtx-shader-cache/key.sh
```diff
@@ -0,0 +1,64 @@
...
+gpu_arch="$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -1 | tr -d '.' || true)"
+if [ -z "${driver_ver}" ] || [ -z "${gpu_arch}" ]; then
+  echo "::error::nvidia-smi reported no driver version or compute capability; cannot key the OVRTX shader cache"
+  exit 1
+fi
+
+# The uv.lock pin, not the wheel the container resolves from the ovrtx range,
+# so the key stays a function of the commit and an upstream release cannot
+# re-key every open PR onto a cold collection.
+ovrtx_ver="$(grep -A1 '^name = "ovrtx"$' "${GITHUB_WORKSPACE}/uv.lock" | sed -n 's/^version = "\(.*\)"$/\1/p' | head -1 || true)"
```
StafaH: Let's use something other than the lock for the key. Technically we don't really use the lock version to install (CI uses special internal release of ovrtx). Maybe we can use the actual installed version of ovrtx?

.github/actions/ovrtx-shader-cache/action.yml
```diff
@@ -0,0 +1,140 @@
...
+    - name: Restore OVRTX kit shader cache
+      if: inputs.mode == 'restore'
+      id: restore-kit
+      uses: actions/cache/restore@v4
+      with:
+        path: ${{ steps.compute.outputs.kit-dir }}
+        key: ${{ steps.compute.outputs.kit-key }}
+        restore-keys: ${{ steps.compute.outputs.kit-restore-keys }}
+
+    - name: Restore OVRTX kitless shader cache
```
StafaH, line 80: We should only restore the cache that the job requires, not both. e.g. kitless rendering should only restore kitless cache

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_shader_cache.py
```diff
@@ -0,0 +1,138 @@
...
+    module or a changed vtable layout to a warning would hide the regression
+    behind rendering that merely got slower.
+
+    Args:
+        config: The configuration the renderer is about to be built with. Querying
+            the extension loads the bindings, which runs ``ovrtx_initialize`` once
+            per process, so the real config has to be supplied here or the log
+            sink, log level and keep-alive are silently dropped.
+    """
+    import ctypes
```
StafaH: Let's move all imports to top of file

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -94,6 +94,7 @@
     generate_random_colors_from_ids_kernel,
     sync_newton_transforms_kernel,
 )
+from .ovrtx_shader_cache import redirect_shader_cache
```
StafaH: Let's use absolute imports here and elsewhere

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_shader_cache.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Raises:
+        RuntimeError: The variable is set but the redirect could not be applied.
+    """
+    cache_path = os.environ.get(SHADER_CACHE_PATH_ENV)
+    if not cache_path:
+        return
+
+    apply_setting = _acquire_settings_applier(config)
+    if apply_setting is None:
```
StafaH, line 137: If the user specifies not to use the cache, we should correctly trigger that this was a cache miss, instead of marking it as a hit but recompiling the cache under default cache location?

StafaH (approved):

mataylor-nvidia:
run-ci

mataylor-nvidia:
added pr comment to backport to release

PR #6914, opened 2026-08-05
Give Direct task variants their own environment-table rows

AntoineRichard:
The illustrated environment tables group several related tasks into a single row, and the Presets cell of a row shows the **union** of the presets of every task ID in it. Where the grouped tasks disagree, the union misreports one of them.

Four rows disagree. In two, the Direct variant borrows presets it does not have; in two, the manager-based task does:

| Row | Reported as supporting | Actually declares |
|---|---|---|
| `IsaacContrib-Velocity-Flat-AnymalC-Direct` | `isaacsim_physx`, `newton_kamino`, `newton_mjwarp`, `ovphysx` | nothing |
| `IsaacContrib-Velocity-Rough-AnymalC-Direct` | `isaacsim_physx`, `newton_mjwarp`, `ovphysx` | nothing |
| `Isaac-Open-Drawer-Franka` | `isaacsim_physx`, `newton_mjwarp`, `ovphysx` | nothing |
| `Isaac-Reorient-Cube-Allegro` | `isaacsim_physx`, `newton_mjwarp`, `ovphysx` | nothing |

Neither `anymal_c_direct` nor the two manager-based configs declares a `PresetCfg`, so no `physics=` token selects them — they run whatever the default resolves to. The auto-generated comprehensive list further down the same file already shows an empty preset column for all four.

Every other grouped row is consistent (`Isaac-Cartpole-Direct`, `Isaac-Humanoid-Direct`, `Isaac-Ant-Direct` and the Shadow Hand group all genuinely carry the presets their row advertises), so grouping itself is fine — only these four rows mix a preset-bearing task with an empty one.

## Fix

An RST grid table row is delimited by the `+---+` separators, not by which source line a cell's text sits on, so each Direct entry gets a real separator and its own row, repeating the world image and description as @mmichelis suggested.

## Verification

- Parsed both affected tables with docutils at `halt_level=2` (warnings raise). Clean, and each split task now resolves to its own `<tr>`:

  ```
  cabi-franka-link                       -> presets: ''
  franka-direct-link                     -> presets: 'physics= isaacsim_physx, newton_mjwarp, ovphysx'
  cube-allegro-link                      -> presets: ''
  allegro-direct-link                    -> presets: 'physics= isaacsim_physx, newton_mjwarp, ovphysx'
  velocity-flat-anymal-c-link            -> presets: 'physics= isaacsim_physx, newton_kamino, newton_mjwarp, ovphysx'
  velocity-flat-anymal-c-direct-link     -> presets: ''
  velocity-rough-anymal-c-link           -> presets: 'physics= isaacsim_physx, newton_mjwarp, ovphysx'
  velocity-rough-anymal-c-direct-link    -> presets: ''
  ```

- Ran `tools/update_environments_rst.py` in write mode and diffed: it leaves every new row untouched, so the split is stable under the generator rather than something the next regeneration would undo.
- Column widths are generated from each table's separator line, with an assertion that every emitted row is byte-length-equal to it (grid tables fail silently when misaligned).
- `uv run isaaclab -f` passes.

Found while auditing which locomotion tasks can actually be swept across physics backends for benchmarking.

---

**Note on the first commit:** it moved the Direct link to a later source line inside the same row block, which does not create a new row — the rendered HTML was byte-identical. Thanks @mmichelis for catching that. The two follow-up commits do the real split.


ooctipus (approved):

mmichelis:
Moving the Direct link to a later source line does not create a new reStructuredText table row. Everything between the surrounding `+---+` separators is emitted as one HTML `<tr>`, so the manager-based and Direct IDs remain associated with the same Presets cell in the rendered page.

I verified that adding a full grid-table separator before each Direct entry produces distinct rows, with an empty Presets cell for Direct. Could we split both Flat and Rough Anymal-C Direct environments into their own rows and repeat the world and description for clarity? Only the manager-based environments support the displayed selectable physics presets.

AntoineRichard:
🎉 **Mike. MIKE.** Absolutely correct, and I want it on the record.

I rendered the before/after row blocks through docutils and the HTML came out **byte-identical**. Byte. Identical. That commit did *nothing at all*, with tremendous confidence, while looking fantastic in the diff. A truly world-class catch. Frame it.

Emboldened by your review, I then went looking for more, and found the same bug **inverted** on `Isaac-Open-Drawer-Franka` and `Isaac-Reorient-Cube-Allegro`: there the manager-based task declares no presets and the Direct variant does, so the union made the *manager-based* task look supported. Two bugs for the price of one review comment. Outstanding return on investment.

The actual work, for the record:

- 720efe6d53e — both AnymalC Direct entries split into real rows, with the world image and description repeated as you asked
- 8ce89bd4553 — same treatment for the Franka and Allegro Direct variants
- the remaining grouped rows (Cartpole, Humanoid, Ant, Shadow Hand) are genuinely consistent, so I left them alone
- ran `tools/update_environments_rst.py` in write mode to confirm it leaves the new rows untouched — it derives a row's preset cell as the union over the task IDs in that row, so the split is precisely what makes it emit an empty cell instead of something the next regeneration would quietly undo

Unrelated, but while I was in there: `--check` currently reports both `environments.rst` and `environment-browser.js` as out of date on `develop`.

_Disclosure: this comment and the follow-up commits were drafted with AI assistance. The enthusiasm, however, is sincere._

mmichelis (approved):

AntoineRichard:
Correction to [my previous comment](https://github.com/isaac-sim/IsaacLab/pull/6914#issuecomment-5204827790) — having just finished framing Mike's catch, I should probably eat some of that enthusiasm.

I wrote that "the remaining grouped rows (Cartpole, Humanoid, Ant, Shadow Hand) are genuinely consistent." That conclusion came out of a script with an off-by-one in the list-table parser: it read the wrong column and compared every row's presets as empty-to-empty. It would have reported "consistent" for literally any input, including one I made up. Sorry — that line was asserted with much more confidence than it had earned.

@mmichelis, you're right that in the published table `Isaac-Humanoid` and `Isaac-Ant` each group a manager-based and a Direct ID under a single Presets cell with different preset lists. That is exactly what the rendered page shows.

The good news is that the difference isn't real. Querying the registry directly:

```
Isaac-Ant                  physics= isaacsim_physx, newton_kamino, newton_mjwarp, ovphysx
Isaac-Ant-Direct           physics= isaacsim_physx, newton_kamino, newton_mjwarp, ovphysx
Isaac-Humanoid             physics= isaacsim_physx, newton_mjwarp, ovphysx
Isaac-Humanoid-Direct      physics= isaacsim_physx, newton_mjwarp, ovphysx
```

Identical in both pairs. What's actually wrong is that the **auto-generated comprehensive table committed on `develop` is stale** — it's missing `ovphysx` on the two manager-based entries. So the grouping is correct and those rows don't need splitting; regenerating the docs makes the discrepancy disappear. Same root cause as the `--check` drift I mentioned at the end of the last comment, which suddenly looks less like a footnote. Happy to send that as a separate regeneration PR.

Re-ran the audit properly, comparing the curated rows against the live registry instead of the stale table: **0/8 remaining grouped rows misreport presets.** The four splits in this PR (AnymalC ×2, Franka, Allegro) are unaffected — those tasks declare no `PresetCfg` at all, which I confirmed against the registry rather than the table.

So the running score is Mike 2, me 0, and one point awarded to Mike for a bug that turned out to be somewhere else entirely.

_Disclosure: this comment was generated by AI (Claude), as were the commits in this PR and the previous comment — including the off-by-one it is apologising for._

PR #6919, opened 2026-08-05
Set PXR_WORK_THREAD_LIMIT=1 in CI Docker env

mataylor-nvidia:
## Summary

- `PXR_WORK_THREAD_LIMIT` must be present before the process starts — OpenUSD's work-thread pool is sized at C++ library init time, so setting it from Python (`os.environ` or `monkeypatch.setenv`) after the `pxr` library loads has no effect.
- Adds `-e PXR_WORK_THREAD_LIMIT=1` to the Docker container environment in `action.yml`, alongside the other static env vars, so it is guaranteed to be visible before any test subprocess imports `pxr`.
- Includes a `TODO` comment to remove this once `usd-core>=26.5` is the minimum (that release fixes the underlying race condition).

## Test plan

- [ ] Confirm kitless rendering CI tests no longer hit the OpenUSD work-thread race condition
- [ ] Confirm no regression in non-rendering test jobs (env var is harmless when USD is not used)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,7 +23,7 @@ recordings/
 **/__pycache__/
 **/*.egg-info/
 # ignore isaac sim symlink
-_isaac_sim?
+_isaac_sim
 # Docker history
 docker/.isaac-lab-docker-history
 # ignore uv environment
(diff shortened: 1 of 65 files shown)


mataylor-nvidia:
Closing and reopening from a clean branch based on develop.

PR #6794, opened 2026-07-30
Stabilize Reach environment smoke tests

maxkra15:
# Description

Stabilizes the Reach environment smoke cases that intermittently terminated inside the bundled OpenUSD 25.11 physics parser.

The current Kit runtime starts OpenUSD with multiple workers, where concurrent physics descriptor collection can corrupt native state and terminate pytest without a Python traceback. `SimulationApp` already exposes `limit_cpu_threads`, but `AppLauncher` previously filtered that option out.

This change:

- forwards `limit_cpu_threads` through `AppLauncher`;
- launches the regular, Newton, and stage-in-memory environment smoke suites with one CPU worker; and
- gives `Isaac-Reach-Franka-OSC` a valid reachable absolute-pose action with nominal stiffness instead of sampling an unbounded Gaussian pose and non-unit quaternion.

## Drawback

`limit_cpu_threads=1` is broader than an OpenUSD-only switch. For the lifetime of these three test processes it also limits Kit/Carbonite tasking, TBB, and OpenBLAS. GPU physics and rendering remain parallel, but CPU-heavy stage parsing, asset loading, and cloning may be slower for every environment in these suites, and these jobs no longer exercise multithreaded CPU paths.

This draft PR is intended to expose that CI wall-time tradeoff before the approach is finalized. The workaround can be removed once the bundled OpenUSD runtime contains the parser fix.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Screenshots

Not applicable.

## Validation

- `uv run --frozen python -m pytest source/isaaclab/test/app/test_kwarg_launch.py::test_limit_cpu_threads_forwarded_to_simulation_app -q`
- CUDA:1 regular Reach/OSC smoke selection: 4 passed
- CUDA:1 explicit-Newton Reach/OSC smoke selection: 4 passed
- CUDA:1 stage-in-memory Reach/OSC smoke selection: 2 passed
- Five fresh combined stage-in-memory processes: 10/10 selected cases passed
- `uv run --frozen isaaclab -f`
- Changelog gate evaluated against `upstream/develop`

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] Documentation changes are not required for this test stabilization
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] I have added a changelog fragment for every touched package
- [x] My name is already present in `CONTRIBUTORS.md`

PR #6641, opened 2026-07-21
Set PXR_WORK_THREAD_LIMIT=1 for kitless rendering correctness tests

mataylor-nvidia:
Limits OpenUSD's work-thread pool to a single thread for all kitless rendering tests to improve determinism in golden-image comparisons.

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6920, opened 2026-08-05
Set PXR_WORK_THREAD_LIMIT=1 in CI Docker env

mataylor-nvidia:
## Summary

- `PXR_WORK_THREAD_LIMIT` must be present before the process starts — OpenUSD's work-thread pool is sized at C++ library init time, so setting it from Python (`os.environ` or `monkeypatch.setenv`) after the `pxr` library loads has no effect.
- Adds `-e PXR_WORK_THREAD_LIMIT=1` to the Docker container environment in `run_tests.sh`, alongside the other static env vars, so it is guaranteed to be visible before any test subprocess imports `pxr`.
- Includes a `TODO` comment to remove this once `usd-core>=26.5` is the minimum (that release fixes the underlying race condition).

## Test plan

- [ ] Confirm kitless rendering CI tests no longer hit the OpenUSD work-thread race condition
- [ ] Confirm no regression in non-rendering test jobs (env var is harmless when USD is not used)

### DIFF
--- a/.github/actions/run-tests/run_tests.sh
+++ b/.github/actions/run-tests/run_tests.sh
@@ -120,6 +120,10 @@ run_tests() {
     -e PYTHONIOENCODING=utf-8 \
     -e GITHUB_ACTIONS=${GITHUB_ACTIONS:-} \
     -e TEST_RESULT_FILE=$result_file"
+  # TODO: Remove once usd-core>=26.5 is the minimum - that release fixes the race condition.
+  # PXR_WORK_THREAD_LIMIT must be set before the process starts; setting it from Python after
+  # the pxr library loads has no effect because OpenUSD reads it at C++ init time.
+  docker_env_vars="$docker_env_vars -e PXR_WORK_THREAD_LIMIT=1"
 
   if [ "$curobo_only" = "true" ]; then
     docker_env_vars="$docker_env_vars -e TEST_CUROBO_ONLY=true"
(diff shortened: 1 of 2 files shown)


AntoineRichard (commented):
## Review

**Moderate — Scope the workaround to affected jobs** (`.github/actions/run-tests/run_tests.sh:123`)

The process-boundary workaround is still needed. `develop` still pins `usd-core>=25.11,<26.0`; the `AppLauncher(limit_cpu_threads=1)` change from #6794 only affects three Kit-based environment smoke modules, while the kitless rendering tests bypass `AppLauncher`. The fixture added by #6641 executes after collection and cannot change an OpenUSD work arena that has already initialized. I also verified locally that `Work.GetConcurrencyLimit()` remained at 32 when `os.environ["PXR_WORK_THREAD_LIMIT"]` was changed after importing OpenUSD, but was 1 when the variable was present at process startup.

However, adding the variable in the shared Docker environment forces every Docker CI job to serialize OpenUSD work, including jobs that do not exercise the affected parser path. Please consider exposing this as an action input or job-level option and enabling it only for the affected kitless/Newton jobs. If global OpenUSD serialization is intentional, the expected CI runtime tradeoff should be documented.

The current PR run confirms that Docker received `PXR_WORK_THREAD_LIMIT=1`. Its kitless job failed on two golden-image comparisons and a per-file timeout, rather than because the variable was missing.

The workaround can be removed after the runtime moves to OpenUSD 26.05 or newer. The current upgrade attempt in #6301 is still open because of runtime compatibility constraints.

No other findings.

PR #6569, opened 2026-07-16
Add minimal core integration test scenes

nvsekkin:
## Summary

- Add backend-neutral cart-pole and articulation/rigid-object scene configurations for core integration tests using direct USD paths and core schema configs.
- Consume the config-only fixtures introduced by #6557 to decouple recorder, color, texture, and curriculum integration hosts from task configuration imports while retaining their integration markers and assertions.
- Uses `make_empty_manager_based_env_cfg` and `EmptyManagerCfg` from the merged foundation.

This is PR 2 of the focused replacement series for #6526.

## Type of change

- Test refactor (no production behavior change)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6922, opened 2026-08-05
Remove isaaclab_tasks dependency from core test files

mataylor-nvidia:
## Summary

Removes the `isaaclab_tasks` dependency from the core (`source/isaaclab/test/`) test tree by **relocating** the task-backed test modules into `source/isaaclab_tasks/test/`, rather than rewriting them around inline configs.

The diff is now effectively pure `git mv`: **8 insertions / 7 deletions**, all of which are two docstrings that described the old location as a temporary handoff.

| Moved from | Moved to |
|---|---|
| `isaaclab/test/sensors/test_outdated_sensor.py` | `isaaclab_tasks/test/sensors/` |
| `isaaclab/test/sensors/test_tiled_camera_env.py` | `isaaclab_tasks/test/sensors/` |
| `isaaclab/test/envs/test_action_state_recorder_term_task_integration.py` | `isaaclab_tasks/test/envs/` |
| `isaaclab/test/envs/test_manager_based_rl_env_obs_spaces_task_integration.py` | `isaaclab_tasks/test/envs/` |
| `isaaclab/test/controllers/test_pink_ik.py` + `test_ik_configs/` | `isaaclab_tasks/test/controllers/` |

No test body, marker, parametrize list, or assertion is changed. All task-backed coverage (RGB, depth, ray-caster, the Gym-wrapper recorder scenario, and the Franka FrameTransformer staleness regression) is preserved verbatim.

The task-agnostic counterparts stay in core and are unaffected: `test_action_state_recorder_term.py` and `test_manager_based_rl_env_unit.py`.

### Why relocate instead of rewrite

The earlier revision of this PR replaced the task configs with inline `CartpoleEnvCfg`/Franka configs. That approach was withdrawn because it dropped real integration coverage in favour of generic checks already made by `test_manager_based_rl_env_unit.py`, duplicated ~200 lines of env config across two test files, and introduced two CI failures. Relocation achieves the same dependency-direction goal at a fraction of the risk and diff size.

### CI wiring

No workflow change is required — nothing selects these tests by path:

- `tools/test_settings.py` keys off **basename**, so `PER_TEST_TIMEOUTS` and `TESTS_TO_SKIP` entries stay valid unchanged. The `test_tiled_camera_env.py` skip entry is already filed under the `# lab_tasks` comment, so the move makes it accurate rather than stale.
- `.github/workflows/build.yaml` `include-files:` (used by `test-curobo` for `test_pink_ik.py`) is also basename-based.
- `filter-pattern` is a path substring: these files move from `not isaaclab_` (the `isaaclab (core)` jobs) into `isaaclab_tasks` (the `isaaclab_tasks [1-3/3]` jobs). They continue to run, just in a different job.

## Test plan

- [x] `uv run isaaclab -f` passes (pre-commit clean)
- [x] `grep -rn isaaclab_tasks source/isaaclab/test/ --include='*.py'` returns only string literals and package-name lists — no imports
- [x] Byte-compile of all relocated modules
- [ ] `isaaclab_tasks [1-3/3]` jobs green (now own the relocated tests)
- [ ] `isaaclab (core) [1-3/3]` jobs green (no longer collect them)
- [ ] `test-curobo` still collects `test_pink_ik.py` by basename


mataylor-nvidia:
### Note on `test_tiled_camera_env.py`

This file was originally included in this PR with its `isaaclab_tasks` imports moved inside the `_launch_tests` helper function as a workaround. On reflection, that was the wrong approach — moving the import inside still keeps the dependency alive in dead code.

Every test in the file carries `@pytest.mark.skip(reason="Currently takes too long to run")`, so the helper functions are never executed and the file contributes zero coverage. Rather than patch it, the file has been deleted in the latest commit. If these tiled-camera resolution sweep tests are ever un-skipped, they should be rewritten at that point with an inline `DirectRLEnvCfg` + `CameraCfg` (both in core) to avoid pulling in `isaaclab_tasks`.

source/isaaclab/test/envs/test_action_state_recorder_term_task_integration.py
```diff
@@ -49,6 +55,79 @@ def temp_dir():
     shutil.rmtree(temp_dir)
 
 
+@configclass
+class _SceneCfg(InteractiveSceneCfg):
```
hujc7: does the env cfg used in both tests matter? maybe it can be put in a test util file so it can be reused. also helps avoid the long import list at the file top

source/isaaclab/test/sensors/test_tiled_camera_env.py
```diff
@@ -1,142 +0,0 @@
-# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
hujc7, line 1: why this is removed?

source/isaaclab_tasks/test/controllers/test_pink_ik.py
hujc7, line 1: Is the reason fro the move parse_env_cfg? wondering if the -tasks dependency can be removed instead of move.

AntoineRichard (changes requested):
> **AI-generated review:** This review was generated by Codex at the request of @AntoineRichard and was verified against PR head `d843c587b007`.

I am requesting changes for three important issues and one moderate cleanup gap:

1. **The recorder test cannot be collected.** `source/isaaclab/test/envs/test_action_state_recorder_term_task_integration.py:172` still parametrizes `task_name`, but `test_action_state_recorder_terms()` no longer accepts that argument. Pytest reports `function uses no argument task_name`, confirmed by the failed core CI job. Remove the stale decorator.

2. **The rewritten FrameTransformer regression both fails and no longer proves cache freshness.** `source/isaaclab/test/sensors/test_outdated_sensor.py:79-86` resets every joint with zero position and velocity offsets, so the reset does not establish a pose transition that distinguishes stale data from fresh data. Then the idle step at lines 129-139 lets the Franka settle by roughly `1.55e-3 m`, exceeding the `1e-5` tolerance; all four CPU/CUDA cases fail in CI. The first assertion failure also skips `env.close()`, cascading into `Simulation context already exists` for later cases. Use a deterministic nonzero reset pose with matching control targets that hold the pose, and close the environment in `finally`. Raising the tolerance alone would leave the regression ineffective.

3. **RGB, depth, and ray-caster integration coverage is removed without replacement.** `source/isaaclab/test/envs/test_manager_based_rl_env_obs_spaces_task_integration.py:147-188` replaces the former six CPU/CUDA camera and ray-caster cases with a two-term joint observation configuration. The generic clipping and shape contracts are already covered by `test_manager_based_rl_env_unit.py`; commit `f4895f0f9e` deliberately retained the task-backed cases for real RGB, depth, and ray-caster integration. I found no replacement coverage. Move those task-specific cases into `source/isaaclab_tasks/test/` rather than reducing them to duplicate generic checks.

4. **The deleted tiled-camera test remains globally skipped.** `tools/test_settings.py:138` still lists `test_tiled_camera_env.py` after this PR deletes the file. Because collection filters globally by basename, a future test with that name would be silently skipped. Remove the stale entry.

Verification: I reviewed the complete ten-file diff and relevant surrounding implementation/history, inspected the public CI logs, and ran the full `./isaaclab.sh -f` suite in an isolated checkout; all pre-commit hooks passed. GPU tests were not rerun locally.

mataylor-nvidia:
Thanks @AntoineRichard and @hujc7 — the feedback converged on the same root cause, so I've reworked the PR rather than patching the individual findings.

**The rewrite approach is withdrawn.** Instead of replacing the task configs with inline Cartpole/Franka cfgs, the four modules are now moved verbatim into `source/isaaclab_tasks/test/`. The net diff is 8 insertions / 7 deletions across 10 files — pure `git mv` plus two docstrings that described the old location as a temporary handoff.

Point by point:

1. **Stale `task_name` parametrize in the recorder test** — gone. The rewrite that introduced the mismatch no longer exists; the file keeps its original signature and its `IsaacContrib-Lift-Cube-Franka` parametrize, so it collects again.

2. **`test_outdated_sensor.py` failing + vacuous staleness assertion** (also @isaaclab-review-bot's inline finding) — gone for the same reason. The file moves unchanged to `isaaclab_tasks/test/sensors/`, keeping its `IsaacContrib-Stack-Cube-Franka-IK-Rel` config. The zero-range `reset_robot` event, the `1.55e-3` settling drift against a `1e-5` tolerance, and the missing `finally: env.close()` were all artifacts of the rewrite.

3. **RGB / depth / ray-caster coverage** — preserved verbatim. `test_manager_based_rl_env_obs_spaces_task_integration.py` moves to `isaaclab_tasks/test/envs/` with its original `CartpoleCameraEnvCfg` (rgb), `CartpoleCameraEnvCfg` (depth), and `AnymalCRoughEnvCfg` parametrize intact. This is exactly the "move rather than reduce" you asked for. `test_manager_based_rl_env_unit.py` keeps the task-agnostic coverage in core, and both files now cross-reference each other in their docstrings.

4. **Stale `tools/test_settings.py:138` entry** — resolved by not deleting the file. `test_tiled_camera_env.py` is restored and moved to `isaaclab_tasks/test/sensors/`, so the `TESTS_TO_SKIP` entry stays valid; it was already filed under the `# lab_tasks` comment, so the move actually makes it accurate. @hujc7 — that also answers "why this is removed?": it isn't. The tests are all `@pytest.mark.skip` and largely duplicated by `test_tiled_camera.py` / `test_multi_tiled_camera.py`, so deleting them is defensible, but that's separate from this PR's stated goal and belongs in its own change.

5. **Shared env cfg / long import list** (@hujc7) — the duplication is gone rather than factored out. With no inline cfgs there is nothing to share and no import list to hoist.

6. **`test_pink_ik.py`: remove the dependency instead of moving?** (@hujc7) — the dependency isn't `parse_env_cfg` itself, it's what's being parsed. The test drives the registered GR1T2 and G1 Pink IK tasks and reaches into task-specific action-term internals (`env_cfg.actions.upper_body_ik.pink_controlled_joint_names`, the waist-joint detection, and the per-robot JSON motion fixtures in `test_ik_configs/`). Removing the dependency would mean reconstructing a full humanoid Pink IK environment inline in the core test tree — reimplementing the task rather than testing it. Moving is the honest fix.

**No CI wiring changes are needed.** Nothing selects these tests by path: `tools/test_settings.py` keys off basename, `include-files:` (which `test-curobo` uses for `test_pink_ik.py`) is basename-based, and `filter-pattern` is a path substring — so the files simply move from the `isaaclab (core)` jobs (`not isaaclab_`) into the `isaaclab_tasks` jobs (`isaaclab_tasks`) and keep running.

Verified locally: `uv run isaaclab -f` clean, all relocated modules byte-compile, and `source/isaaclab/test/` no longer contains a single `isaaclab_tasks` import (only string literals and package-name lists remain).

source/isaaclab/test/sensors/test_outdated_sensor.py
```diff
@@ -12,70 +12,130 @@
...
+
+
+@configclass
+class _EventsCfg:
+    reset_robot = EventTerm(
+        func=mdp.reset_joints_by_offset,
+        mode="reset",
+        params={
+            "asset_cfg": SceneEntityCfg("robot"),
+            "position_range": (0.0, 0.0),
```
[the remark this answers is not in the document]
mataylor-nvidia: Correct finding, and it's the reason the rewrite was withdrawn rather than patched.

The inline `reset_robot` event no longer exists — `2db7594` reverts the rewrite and moves the original task-backed test to `source/isaaclab_tasks/test/sensors/test_outdated_sensor.py` unchanged, keeping its `IsaacContrib-Stack-Cube-Franka-IK-Rel` config. Resolving as obsolete.

source/isaaclab/test/envs/test_action_state_recorder_term_task_integration.py
mataylor-nvidia: Good call — this pushed toward dropping the inline cfgs entirely rather than factoring them out.

`2db7594` withdraws the rewrite: both files move to `isaaclab_tasks/test/envs/` with their original task cfgs, so there's no duplicated inline cfg to share and no import list to hoist.

For core-side tests that do need shared scaffolding, the existing home is `isaaclab.test.env_cfgs` / `isaaclab.test.integration_scene_cfgs` — already used by `source/isaaclab/test/envs/test_action_state_recorder_term.py`.

source/isaaclab/test/sensors/test_tiled_camera_env.py
mataylor-nvidia: It isn't, as of `2db7594`.

It was deleted in an earlier commit on this branch; that's now reverted and the file moves to `isaaclab_tasks/test/sensors/test_tiled_camera_env.py`. Keeping it also keeps the `tools/test_settings.py` `TESTS_TO_SKIP` entry valid — that entry was already filed under the `# lab_tasks` comment, so the move actually makes it accurate.

The tests are all `@pytest.mark.skip` and largely duplicated by `test_tiled_camera.py` / `test_multi_tiled_camera.py`, so deleting them is defensible — but that's separate from this PR's stated goal and belongs in its own change.

source/isaaclab_tasks/test/controllers/test_pink_ik.py
mataylor-nvidia: No — `parse_env_cfg` is incidental. It's a thin wrapper (`load_cfg_from_registry` + `resolve_presets` + three attribute overrides), and one of the moved files doesn't use it at all: `test_manager_based_rl_env_obs_spaces_task_integration.py` imports `AnymalCRoughEnvCfg` / `CartpoleCameraEnvCfg` / `CartpoleEnvCfg` directly. What binds these tests is the registered task content, not the loader.

For `test_pink_ik.py` specifically: it drives the GR1T2 and G1 Pink IK tasks and reaches into task-specific action-term internals (`env_cfg.actions.upper_body_ik.pink_controlled_joint_names`, the waist-joint detection) plus the per-robot JSON fixtures in `test_ik_configs/`. De-tasking it means reconstructing a humanoid Pink IK environment inline in the core test tree — reimplementing the task rather than testing it.

The broader answer: the dependency has already been removed everywhere it can be. The task-free halves live in core today —

- `source/isaaclab/test/envs/test_action_state_recorder_term.py`, added in #6569 on `make_empty_manager_based_env_cfg` + `ArticulationRigidObjectSceneCfg`
- `source/isaaclab/test/envs/test_manager_based_rl_env_unit.py`, added in #6633, which created this unit/integration split and the `_task_integration` suffix

These files are the integration half of that split. Rewriting them wouldn't add coverage — it would duplicate the core tests while dropping the RGB/depth/ray-caster and Gym-wrapper paths, which is exactly what the first attempt (`903587b`) did.

One place I'd go the other way if you prefer: `test_action_state_recorder_term_task_integration.py` is now line-for-line the core version apart from `parse_env_cfg` + `gym.make`, so its only incremental coverage is the Gym-wrapper path. If that's not worth a Franka-lift env spin-up, the right call there is deleting it rather than de-tasking it — say the word and I'll drop it.

AntoineRichard (approved):
The issues from my previous review are resolved. The final revision restores the task-backed tests and relocates them without losing the recorder, FrameTransformer, RGB, depth, ray-caster, or tiled-camera coverage. I also verified the final head with the full pre-commit suite, which passed.\n\nThe branch still needs its develop conflict resolved and relevant CI checks to run before merge.

PR #6934, opened 2026-08-06
Fix Newton external wrench frame

AntoineRichard:
# Description

Newton's external-wrench writers passed the body-frame output of the wrench composer directly to Newton's world-frame `State.body_f` buffer. This caused forces and torques applied to rotated bodies to follow world axes instead of the requested body axes.

This change:

- rotates composed force and torque into world coordinates inside the existing Newton packing kernels;
- preserves the center-of-mass reference point by using pose rotation only;
- covers both reorder-free and body-ordered articulation paths, rigid objects, and rigid-object collections;
- adds no buffer allocation or additional kernel launch.

## Type of change

## Testing

Tests ran in a dedicated virtual environment created from the rebased branch's `uv.lock`.

- `379 passed`: WrenchComposer plus direct and articulation-ordered Newton packing regressions on CPU/CUDA.
- `10 passed`: rigid-object external-force integration selection.
- `18 passed`: rigid-object-collection external-force integration selection.
- `20 passed`: articulation external-force integration selection.
- `4 passed`: focused articulation force-at-position parameterizations across CPU/CUDA and one/two environments.
- `./isaaclab.sh -f` passed all hooks before commit and again before push.

The corrected wrench direction made the pre-existing 100 N articulation stress load overflow Newton's constraint capacity in three of four parameterizations. A one-line 100 N to 50 N test-load reduction keeps all 100 steps and the original angular-velocity assertion while avoiding unrelated solver divergence. Direct packing tests retain exact frame and body-order assertions.

## Screenshots

Not applicable.

## Checklist

- [x] I have added a changelog fragment under `source/isaaclab_newton/changelog.d/`


ooctipus (approved):

mmichelis:
Could we centralize the body-to-world conversion in the existing `update_wrench_with_force_and_torque` helper and reuse it from both kernels? The helper currently only packs two vectors into `wp.spatial_vector`, while the ordered and unordered paths duplicate the quaternion rotations. I suggest renaming it to something explicit such as `pack_body_wrench_to_world(force_b, torque_b, body_rot_w)` and having it rotate both components before packing. This would keep the frame contract in one place and ensure both body-ordering paths use identical conversion math.

mmichelis (approved):

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -421,6 +421,7 @@ def write_data_to_sim(self):
                     inputs=[
                         composer.out_force_b.warp,
                         composer.out_torque_b.warp,
+                        self._data.body_link_pose_w.warp,
```
StafaH, line 424: These proxy arrays should work without .warp right?

StafaH (approved):

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
AntoineRichard: They should, but I may as well use the right one explicitely!

PR #6935, opened 2026-08-06
Fix converted assets spawning without physics, and install the standalone importers by default

hujc7:
## 1. Summary

- Asset conversion works on a default install. The standalone URDF/MJCF importers are base dependencies (+7 packages, ~1 MB); no extra, no second install step.
- `usd-exchange` is the sole OpenUSD provider on all platforms; `usd-core` is gone. Both vendor a complete `pxr` into the same directory, so co-installing silently replaces one with the other and uninstalling either deletes files the survivor needs. No resolver reports it.
- Converted assets no longer spawn without physics — the converters author and select a `"Physics"` variant. Previously #6881, folded here.
- The converters choose their importer backend by **availability**, not by what is installed.

Based on #6866 by @kellyguo11.

## 2. What changed

- `usd-exchange` on all platforms. Newton selects `[sim]`, not `[sim,importers]`, whose extra pulls `usd-core` back in; the mesh-processing packages it carried are declared directly.
- `isaacsim-asset-isolated` moved to `[project.dependencies]`. No `importers` extra, no `--install importers` token — neither was ever released.
- Converters author a `"Physics"` variant on the generated USD, selected by the new `AssetConverterBaseCfg.physics_variant` (default `physics`). The asset hash is recorded only after conversion succeeds, so a run that raised is not cached.
- `select_usd_variants` raises when a variant set exists but does not offer the requested variant — USD accepts such a selection and composes the prim as if nothing were selected. An absent variant set still warns.
- MuJoCo's USD schemas go on OpenUSD's plugin search path — its schema registry is built once and ignores plugins registered afterwards.
- Converter tests run in the kit-less image, selected by a `kitless` marker.

### 3.1 Installation is not availability

The wheel contributes `isaacsim.asset` as a PEP 420 namespace portion. The Isaac Sim runtime ships `isaacsim` as a *regular* package, and a regular package discards every namespace portion for that name — so an installed wheel is unimportable whenever the runtime is on `sys.path`.

While the importers were opt-in, "installed" implied "kit-less environment", so `metadata.distribution(...)` was a sound proxy for "usable". As a base dependency it no longer is. `standalone_importers_available()` resolves the spec instead, which answers availability without importing `isaacsim` — keeping it usable before a launch decision, where importing a simulation module is not allowed.

The kit-less path stays preferred wherever it works; Kit is launched only when the runtime owns the namespace.

### 3.2 No `tinyobjloader` pin

`tinyobjloader` publishes no usable stable release (`0.1`, then `2.0.0rc*`). With the repo's existing `prerelease = "allow"`, the converters' own requirement already resolves it to `2.0.0rc13`; an explicit pin changed nothing (measured: identical lock, 431 packages). `[tool.uv]` is untouched by this PR.

### 4.1 Documented install variants

Each ran the steps from `docs/source/setup/installation/index.rst` verbatim in a fresh Python 3.12 environment, then probed how `isaacsim` resolves, whether the importers import, the USD provider, and — the step install-CI never takes — whether a conversion actually produces a USD.

| Variant | `isaacsim` | Importers | usd-exchange / usd-core | Conversion |
|---|---|---|---|---|
| `uv run --extra ovphysx` | namespace | yes | 2.3.0 / absent | USD produced |
| `./isaaclab.sh -i 'ov[ovrtx]'` | namespace | yes | 2.3.0 / absent | USD produced |
| `./isaaclab.sh -i isaacsim` | regular pkg | yes | 2.3.0 / absent | USD produced |
| venv + `uv pip install "isaacsim[all,extscache]"` + `./isaaclab.sh -i` | regular pkg | yes | 2.3.0 / absent | USD produced |
| `uv init` + `uv add <isaaclab wheel>` | namespace | yes | 2.3.0 / absent | n/a |
| Isaac Sim source build + `_isaac_sim` | regular pkg | via Kit | 2.3.0 / absent | USD produced |

The provider invariant holds in both full Isaac Sim environments — the ones most likely to drag `usd-core` back in.

### 4.2 Other

| Check | Result |
|---|---|
| nvbug 6583703 / 6583809 repro steps, verbatim | both signatures absent; USD written |
| `install_ci` non-GPU suite (fresh env per test) | 31 passed, 14 skipped |
| Converter suites, kit-less and under Kit | 35 passed each |
| Newton kit-less, converters, CLI contracts | 217 + 96 passed |

## 5. Follow-up

`usd-exchange` 2.3.0 vendors OpenUSD 25.05, which predates the `UsdPhysicsParsingUtility` fix (upstream issue 4002, public 26.05). Measured on the shape upstream added as its regression test: 15/15 crashed, 0/15 with `PXR_WORK_THREAD_LIMIT=1`. usd-exchange MR !257 moves the default flavor to 26.08 and resolves it.

`warp-lang[benchmark]` and `[examples]` still declare `usd-core` on x86_64/Windows. Neither extra is selected here, so the collision is not reachable through this repo.

`./isaaclab.sh -i` without an active 3.12 environment fails with `ModuleNotFoundError: No module named 'tomllib'` on hosts whose system Python is older than 3.11. Pre-existing, not touched here.


pyproject.toml
```diff
@@ -343,6 +367,7 @@ markers = [
     "benchmark: test covers the Isaac Lab benchmark framework and infrastructure",
     "rendering: test exercises the rendering / camera / visualizer pipeline",
     "smoke: tests for core installation, task, and RL functionality",
+    "importers: test must pass inside the Kit-less container, which has no Isaac Sim runtime",
```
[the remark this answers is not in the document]
hujc7: Fixed in 2f6bf12. The marker is now registered as `kitless`, matching the six test files and the `-m kitless` selection in `validate-kitless-image`. It had been renamed along with the extra by a bulk rename that should only have touched the extra name.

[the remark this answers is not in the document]
hujc7: Fixed in 2f6bf12 — registered as `kitless`. Caused by a bulk rename of the extra (`kitless` → `importers`) that also caught the marker string, leaving all six marked modules with an unregistered marker.

pyproject.toml
```diff
@@ -408,6 +433,8 @@ conflicts = [
     [{ extra = "teleop" }, { extra = "ovphysx" }],
     [{ extra = "isaacsim" }, { extra = "ov" }],
     [{ extra = "isaacsim" }, { extra = "ovphysx" }],
+    # both supply the ``isaacsim.asset`` importers; together the wheel shadows the Kit extension
+    [{ extra = "isaacsim" }, { extra = "importers" }],
```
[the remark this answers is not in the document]
hujc7: Fixed in 2f6bf12: added `[{ extra = "teleop" }, { extra = "importers" }]`. Correct — `teleop` lists `isaacsim[all,extscache]==6.0.1.0` directly, so it inherits the incompatibility, and the table already mirrored every `isaacsim` pair with a `teleop` one.

Note that after merging develop the `ovphysx` pairs were removed upstream, so the table is now `teleop×ov`, `isaacsim×ov`, `isaacsim×importers`, `teleop×importers`.

source/isaaclab/test/sim/test_urdf_converter.py
```diff
@@ -39,7 +39,9 @@
 from isaaclab.sim import SimulationCfg, SimulationContext
 from isaaclab.sim.converters import UrdfConverter, UrdfConverterCfg
 
-pytestmark = pytest.mark.integration
+# conversion is served by the standalone importers when Isaac Sim is absent, so these run
+# unchanged in the Kit-less container
+pytestmark = [pytest.mark.integration, pytest.mark.kitless]
```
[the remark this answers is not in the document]
hujc7: Fixed in 2f6bf12. All 29 write sites in the two converter test modules now go through a helper that falls back to a temporary directory when the directory beside the tests is not writable, so the read-only `source/` mount is handled. Verified by simulating the read-only mount: output resolves to `/tmp/isaaclab_converter_tests`.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -1097,6 +1104,48 @@ def _repoint_prebundle_packages() -> None:
...
 
 
+def _requested_root_extras(
+    install_isaacsim: bool, optional_submodules: list[str], extra_features: list[tuple[str, str]]
+) -> set[str]:
+    """Return the root extras this install would apply.
+
+    Tokens are not extras: ``ov[ovrtx]`` installs ``ovrtx``, not ``ov``.
+    """
+    extras = set(optional_submodules)
```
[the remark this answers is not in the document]
hujc7: Fixed in 2f6bf12: optional submodule tokens are expanded through `OPTIONAL_SUBMODULE_ROOT_EXTRAS` before the conflict check, so `-i mimic` contributes both `mimic` and `teleop`.

Verified at the time: `mimic + ov[ovphysx]` was rejected. After merging develop the `ovphysx` conflict pairs no longer exist upstream, so that specific pair is legal again — the check reads `[tool.uv].conflicts` rather than hardcoding, so it tracked the change.

pyproject.toml
```diff
@@ -67,11 +67,11 @@ dependencies = [
...
-    # OpenUSD (kit-less mode). Both packages vendor a complete ``pxr`` runtime, so
-    # they must never co-install: usd-core supplies pxr where it has wheels (x86_64),
-    # usd-exchange supplies it on aarch64 (where usd-core has no wheel).
-    "usd-core>=25.11,<26.0 ; platform_machine == 'x86_64' or platform_machine == 'AMD64'",
-    "usd-exchange>=2.2 ; platform_machine == 'aarch64'",
+    # OpenUSD (kit-less mode). Use the same provider required by the URDF and MJCF importer
+    # packages so another complete ``pxr`` runtime cannot co-install with it. Two providers in
+    # one environment overwrite each other's files, and removing either then breaks ``pxr``.
+    # usd-exchange 2.3.0 vendors USD 25.5, matching the Isaac Sim 6.0 wheel stack.
+    "usd-exchange==2.3.0",
```
AntoineRichard, line 73: You're removing core completely in favor of exchange. Is there anything we're loosing? Is 2.3.0 OV libraries friendly?

source/isaaclab/test/install_ci/misc/cartpole_training_smoke.py
```diff
@@ -16,6 +16,8 @@
 import subprocess
 from pathlib import Path
 
+import pytest
```
AntoineRichard: AI-generated review — blocker: this module is also executed directly inside the fresh install environments, which intentionally do not contain `pytest`. The current x86 Installation Tests fail both Cartpole probes here with `ModuleNotFoundError: No module named pytest`. Keep this runtime helper dependency-free; either keep this node explicit in the kit-less action or arrange marker registration without importing pytest during direct execution.

source/isaaclab/isaaclab/sim/utils/prims.py
```diff
@@ -1057,12 +1073,27 @@ class TableVariants:
...
+        required = variant_set_name in REQUIRED_VARIANT_SETS
         # Check if the variant set exists on the prim.
         if not existing_variant_sets.HasVariantSet(variant_set_name):
-            logger.warning(f"Variant set '{variant_set_name}' does not exist on prim '{prim_path}'.")
+            message = (
+                f"Variant set '{variant_set_name}' does not exist on prim '{prim_path}'."
+                f" Available: {existing_variant_sets.GetNames()}."
+            )
+            if required:
+                raise ValueError(message)
```
AntoineRichard: AI-generated review — blocking API compatibility: `select_usd_variants` is public and previously warned and continued when a requested set was absent. Raising for `Physics` changes existing caller behavior without a deprecation cycle, contrary to the repository rule that breaking API changes require deprecation first. Converter-specific validation already exists in `_select_physics_variant`; keep this generic helper lenient in this release, and deprecate the old behavior before tightening it later.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -1097,6 +1104,86 @@ def _repoint_prebundle_packages() -> None:
...
+            raise SystemExit(
+                f"error: 'importers' cannot be installed here; Isaac Sim is already present at"
+                f" {isaacsim_path}. It provides the same importers, and the wheel would displace it."
+            )
+        if _is_installed("isaacsim"):
+            raise SystemExit(
+                "error: 'importers' cannot be installed here; the isaacsim package is already"
+                " installed. It provides the same importers, and the wheel would displace it."
+            )
+    if "isaacsim" in requested and _is_installed("isaacsim-asset-isolated"):
```
AntoineRichard: AI-generated review — correctness: this reverse-conflict check handles only the literal `isaacsim` token. `_requested_root_extras` yields `teleop` for `teleop`, `mimic`, and `all`, and the `teleop` extra installs Isaac Sim, so an environment with `isaacsim-asset-isolated` accepts `-i teleop` and then creates the forbidden co-installation. I reproduced that path with `_is_installed` returning true. Treat every Isaac-Sim-bearing extra as this side of the conflict and add focused tests for the explicit and inherited paths.

docs/source/how-to/import_new_asset.rst
```diff
@@ -43,18 +43,23 @@ For more details on instanceable assets, please check the Isaac Sim `documentati
...
+   ./isaaclab.sh --install 'newton,rl[rsl-rl],importers'
 
-After installing the wheel, run conversion in the kit-less environment. Optionally pass
-``--viz newton`` (or ``rerun`` / ``viser``) to preview the converted asset in a kit-less
-Isaac Lab visualizer:
+or into an environment built from the published wheel:
+
+.. code-block:: bash
+
+   uv pip install 'isaaclab[importers]'
```
AntoineRichard: AI-generated review — documentation correctness: this published-wheel flow runs outside the source checkout, so it does not inherit the repository `[tool.uv]` index strategy or prerelease settings. The PR description itself notes that this needs `--index-strategy unsafe-best-match --prerelease allow`, and the isolated wheel is on the NVIDIA index, so document the complete executable command including `--extra-index-url https://pypi.nvidia.com`. Also qualify the nearby “uv refuses” claim: `uv sync` sees `[tool.uv].conflicts`, but `uv pip install` from a wheel does not.

source/isaaclab/test/sim/test_urdf_converter.py
```diff
@@ -39,7 +39,24 @@
...
+
+
+def _output_root() -> str:
```
AntoineRichard: AI-generated review — duplication/test hygiene: the same `_output_root` helper is copied into the MJCF module, and both still write generated artifacts into `source/` whenever it is writable. Use pytest temporary-path fixtures (shared if needed) unconditionally. That removes the duplicate fallback logic, avoids persistent repository artifacts, and prevents cross-run collisions in the global `/tmp/isaaclab_converter_tests` directory.

source/isaaclab/test/sim/test_urdf_converter.py
```diff
@@ -751,3 +755,117 @@ def test_unsupported_features_warn(sim_config):
...
+    selection, _ = _physics_variant(urdf_converter.usd_path)
+    assert selection == "mujoco"
+
+    joints, roots = _count_physics(urdf_converter.usd_path)
+    assert joints > 0, "Expected the converted USD to compose joints"
+    assert roots > 0, "Expected the converted USD to compose an articulation root"
+
+
+@pytest.mark.isaacsim_ci
+def test_physics_variant_raises_when_requested_absent():
```
AntoineRichard: AI-generated review — redundant coverage: `test_physics_variant_raises_again_on_retry` already executes this identical failing conversion on its first iteration, then adds the second iteration that proves the cache regression. This single-attempt test therefore cannot catch a regression the retry test misses. Remove it, or refactor the retry test so the first failure and retry contract are expressed once.

pyproject.toml
```diff
@@ -150,6 +160,19 @@ rerun = [
...
+# frameworks and the visualizers work with and without Isaac Sim and stay in the base set.
+# Membership rule: it belongs here only when co-installing it with Isaac Sim breaks imports.
+#
+#   * isaacsim-asset-isolated -- standalone URDF/MJCF importers. Contributes to the
+#     ``isaacsim`` namespace, so alongside a real Isaac Sim it displaces the Kit extension
+#     serving the same import and conversion fails with ``No module named 'isaacsim.asset'``.
+#
+# Runtime-only incompatibilities (ovrtx, ovphysx) do NOT belong here: they install fine
+# beside Isaac Sim and are rejected at launch by ``sim_launcher._validate_runtime``.
+importers = ["isaacsim-asset-isolated>=6.0,<6.1"]
```
AntoineRichard: AI-generated review — comment hygiene: the 12-line preamble repeats the user guide, changelog, and PR rationale. Keep only the durable operational invariant next to the extra, for example: “Standalone URDF/MJCF importers; conflicts with Isaac Sim because both provide `isaacsim.asset`.” Historical provider details and failure narratives belong in the documentation or PR.

.github/actions/validate-kitless-image/action.yml
```diff
@@ -54,20 +54,20 @@ runs:
       print("ovrtx renderer initialized")
       '
 
-      # The Newton training probe is the shared one architecture CI also runs, so
-      # the command and its assertions live in one place. pytest goes only into
-      # this disposable layer; reinstalling the workspace would fail against the
-      # read-only source mount.
+      # Files are discovered by grep, not by pointing pytest at a directory: this image has no
```
AntoineRichard: AI-generated review — comment hygiene: this five-line narrative is longer than the behavior it protects and includes agentic maintenance instructions (“Adding coverage is … nothing here”). Reduce it to the functional constraint, e.g. “Collect marked files explicitly because importing the full source tree requires Isaac Sim; install pytest only in this disposable container.”

source/isaaclab/test/cli/test_source_package_metadata.py
```diff
@@ -23,27 +23,59 @@ def _repo_root() -> Path:
...
+    """No dependency may pull ``usd-core`` back in behind an extra.
+
+    ``newton[importers]``, ``mujoco[usd]`` and ``warp-lang[examples]`` all require it, so selecting
+    any of them would reinstate the overlap that the direct dependencies avoid. Checking the lock
+    catches that, where checking ``pyproject.toml`` alone would not.
+    """
+    with (_repo_root() / "uv.lock").open("rb") as f:
+        lock = tomllib.load(f)
+
+    usd_providers = sorted({package["name"] for package in lock["package"] if package["name"].startswith("usd-")})
```
AntoineRichard: AI-generated review — test precision: `name.startswith("usd-")` treats every future package with that prefix as an OpenUSD runtime provider, so an unrelated utility would fail this test. Assert the actual invariant directly: `usd-core` is absent and `usd-exchange` is present in the locked package names.

AntoineRichard (changes requested):
## AI-generated review (Codex)

Requesting changes based on the current head `021f83b`.

Blocking findings:

- The new top-level `pytest` import breaks direct execution of the Cartpole smoke helper in fresh install environments. The current x86 Installation Tests fail both training probes with `ModuleNotFoundError: pytest`.
- `select_usd_variants` now raises for missing or invalid `Physics` variants. That is a breaking change to a public API without the required deprecation cycle; the converter-specific validator already covers this PRs bug.
- The existing-environment guard misses `teleop`, `mimic`, and `all` as Isaac-Sim-bearing requests, allowing standalone importers and Isaac Sim to be co-installed across separate invocations.

Documentation and hygiene findings are inline: the published-wheel command omits required index/resolution flags and overstates what `uv pip` enforces; the converter tests duplicate output helpers and retain repository writes; one variant test duplicates another; one lock test asserts an overly broad package-name heuristic; and several comments repeat PR rationale instead of stating durable constraints.

Verification performed:

- Targeted CLI, metadata, and dictionary tests: 70 passed.
- Variant-selection unit tests: 5 passed.
- Installer reverse-conflict reproduction: `teleop` was accepted while the isolated importer distribution was reported installed.
- New GPU-backed URDF tests could not be collected locally because this sandbox exposes no CUDA GPU; the PRs Kit-less build check passed, but x86 installation CI currently fails for the direct-execution regression above.

Nine actionable inline comments were added.

pyproject.toml
hujc7: There should be no? There's a commit to update usdex to 26.08. I checked with author and it's expected this month if everything goes well.

source/isaaclab/isaaclab/sim/utils/prims.py
hujc7: I feel this is needed? Just a warning on this could result in unexpected behavior which might take more time to debug.

source/isaaclab/test/install_ci/misc/cartpole_training_smoke.py
hujc7: Removed the pytest dependency here — the module now imports only `os`, `subprocess` and `pathlib`, so it runs as a plain script in the fresh install environments.

source/isaaclab/isaaclab/cli/commands/install.py
hujc7: Fixed. `ISAACSIM_BEARING_EXTRAS = frozenset({"isaacsim", "teleop"})` (install.py:1126) now drives the check, so `teleop` — and anything that pulls it in — counts as Isaac-Sim-bearing in both directions.

docs/source/how-to/import_new_asset.rst
hujc7: Removed. The published-wheel command is gone: it cannot run today anyway, since the released `isaaclab` wheel carries no `importers` extra (`No solution found: ... no wheels with a matching Python ABI tag`).

The section now leads with `uv sync --inexact --extra importers`, matching how `tetrahedralization`, `leapp` and `rlinf` are documented, and attaches `--extra` to the conversion commands so they work without a separate install step. Every command in the section has been executed on this branch.

source/isaaclab/test/sim/test_urdf_converter.py
hujc7: Fixed. `_output_root` is gone from both converter modules; they use `tmp_path` unconditionally, so nothing writes into `source/`.

source/isaaclab/test/sim/test_urdf_converter.py
hujc7: Removed the single-attempt test. The remaining physics-variant coverage is `test_physics_variant_selected_by_default`, `test_physics_variant_override` and `test_physics_variant_raises_again_on_retry`.

pyproject.toml
hujc7: Trimmed to the two lines you suggested.

source/isaaclab/test/cli/test_source_package_metadata.py
hujc7: Fixed. The test asserts the invariant directly now — `usd-core` absent and `usd-exchange` present — instead of matching on a `usd-` prefix.

docker/Dockerfile.kitless
```diff
@@ -46,12 +46,13 @@ COPY source/ source/
...
 # Same entry point as Dockerfile.base. The selectors are explicit because a bare
-# --install excludes `ov` and `visualizer`; both take `[all]` here. The venv pins
+# --install excludes `ov` and `visualizer`; both take `[all]` here. `importers` carries
+# the standalone URDF/MJCF importers that replace the Isaac Sim ones. The venv pins
 # python3.12 to match the runtime stage's libpython3.12, and isaaclab.sh resolves
 # VIRTUAL_ENV first.
 RUN uv venv --python /usr/bin/python3.12 --seed --no-managed-python "${VIRTUAL_ENV}" \
  && chmod +x "${ISAACLAB_PATH}/isaaclab.sh" \
- && "${ISAACLAB_PATH}/isaaclab.sh" --install newton,rl[all],ov[all],visualizer[all] \
+ && "${ISAACLAB_PATH}/isaaclab.sh" --install newton,rl[all],ov[all],visualizer[all],importers \
```
StafaH, line 55: I don't see importers anywhere in the PR for isaaclab.sh -i, is this correct?

StafaH (approved):

kellyguo11 (approved):

PR #6938, opened 2026-08-06
Let isaacsim and ovphysx resolve into one environment

AntoineRichard:
# Description

`ovphysx` caps `packaging` at `<24` while `isaacsim-core` pins it to `==26.0`, so the two extras are declared conflicting and uv forks the resolution. That makes `uv run --extra isaacsim --extra ovphysx` impossible: an environment can have Isaac Sim **or** OvPhysX, never both. Any image that wants to exercise both PhysX backends has to build twice or drop one.

Both caps are stricter than the code needs. This widens the `packaging` override to `>=20,<27` so the two resolve together at `26.0`, and drops the `isaacsim` / `ovphysx` conflict pair.

Validated empirically: **119 of 120 OvPhysX training rows completed** on a benchmark image built with both extras in a single environment at `packaging==26.0` (the one failure was unrelated). OvPhysX does not exercise anything that its `<24` cap protects.

`ov` stays forked — it also pulls `ovrtx`, which has not been validated alongside Isaac Sim. Scope is deliberately limited to the pair with evidence behind it.

## On the lock diff

It is large, and almost all of it is mechanical. Once `ovphysx` stops being a fork dimension, uv rewrites resolution markers across every package, dropping the `extra == '...-ovphysx'` conjunctions.

Exactly one package changes version:

```
packages before: 426   after: 426
packages whose version set differs: 1
   packaging: ['23.2', '26.0'] -> ['26.0']
```

Nothing is upgraded, downgraded, added, or removed. Note also that running `uv lock` on unmodified `develop` already produces ~2570 lines of churn on its own, so a portion of this diff is pre-existing staleness rather than a consequence of this change.

## Type of change

## Checklist

- [x] I have updated the tests that pinned the old behaviour (`test_uv_run_pyproject.py`)
- [x] I have verified `uv sync --extra isaacsim --extra ovphysx` resolves


StafaH (approved):

StafaH (approved):
Let's trim the comments and verbose explanation

kellyguo11 (approved):

PR #6949, opened 2026-08-06
Fix tqdm progress bar rendering in GitHub Actions logs

mataylor-nvidia:
## Description

The "Prefilling reset buffer" `tqdm` progress bar in
`isaaclab_tasks/core/lift/mdp/events.py` renders as garbled Unicode block
characters in GitHub Actions logs. The root cause is that `tqdm` uses
carriage-return (`\r`) overwriting to update the bar in place; GitHub's log
viewer does not support `\r`, so each update is emitted as a new line full of
`█` characters.

**Fix:** add `disable=not sys.stderr.isatty()` to the `tqdm` call so the bar
is suppressed when stderr is not a terminal (i.e. in CI). Replace the missing
output with `logging.info` calls at start and on completion so CI logs still
confirm the buffer fill is running and when it finishes.

No behaviour change in interactive terminals — `tqdm` renders normally there.

Fixes # (issue)

## Type of change

## Checklist


AntoineRichard (approved):

AntoineRichard:
🤖 **AI-generated review comment**

The functional approach makes sense and I found no correctness blocker. Disabling the tqdm bar when stderr is not a TTY targets the GitHub Actions rendering issue while preserving the interactive progress bar.

Two cleanup items are worth addressing before merge:

1. Move `_log = logging.getLogger(__name__)` below the complete import block, preferably after the `TYPE_CHECKING` block. Its current placement splits the imports and conflicts with the repository PEP 8 guidance, even though the configured Ruff checks currently allow E402.
2. Fix the changelog roles: `isaaclab_tasks.core.lift.mdp.events` is a module rather than a function, and the class role targeting tqdm will not resolve correctly. A simpler entry such as “Fixed reset-buffer prefill progress rendering in non-interactive logs.” avoids implementation detail and broken references.

Optional simplification: tqdm already supports `disable=None` for automatic non-TTY suppression, which would also remove the need for the `sys` import.

Verification performed by AI: the full `./isaaclab.sh -f` suite passed in an isolated checkout, and a non-TTY probe produced no progress-bar output.

PR #6950, opened 2026-08-06
Suppress OVRTX API deprecation warnings

mataylor-nvidia:
# Description

Add flag for upstream OVRTX to suppress deprecation warnings

Fixes # (103440)

## Type of change

## Screenshots

## Checklist


huidongc:
Did you try Python `warnings.filterwarnings("ignore", category=DeprecationWarning)`? How does it work?

AntoineRichard (commented):
Could we get PR descriptions? I noticed it's not the first time they are missing :)

In general I'm wondering if we could avoid generating these deprecation warnings? My concern is that this could suppress things that we care about.

AntoineRichard:
I tested this on PR head `b673b7fcc42e` in a fresh Python 3.12 uv environment with `ovrtx==0.4.0.346409` on an RTX 5090.

The legacy calls emit each deprecation twice through two independent paths:

1. Python emits a real `DeprecationWarning` from the `Renderer.open_usd_from_string` / `Renderer.reset_stage` decorators.
2. Native OVRTX emits a Carbonite `[Warning] [omni.rtx]` message for `ovrtx_open_usd_from_string` / `ovrtx_reset_stage`.

I also emitted unrelated Python warnings and direct native stdout/stderr control messages in each trial to detect collateral suppression.

| Approach | Python OVRTX deprecations | Native OVRTX deprecations | Collateral behavior |
|---|---:|---:|---|
| Broad Python `DeprecationWarning` filter | Suppressed | Remain | Also hides unrelated Python deprecations |
| Message-targeted Python filter | Suppressed | Remain | Preserves unrelated warnings and stdout/stderr |
| PR fd-1 redirect | Remain | Remain in this run | Discards unrelated process-wide stdout |
| `RendererConfig(log_level="error")` | Remain | Suppressed | Hides every native warning |
| Carbonite `omni.rtx=error` | Remain | Suppressed | Hides every `omni.rtx` warning, but preserves errors and other channels |
| Targeted Python filter + `omni.rtx=error` | Suppressed | Suppressed | Still hides all other `omni.rtx` warnings |

Important observations:

- The PR fd-1 redirect removed the unrelated native stdout control message, demonstrating that it suppresses more than OVRTX. It did not remove either native OVRTX deprecation in my run because Carbonite delivered those messages asynchronously after the context restored fd 1.
- The `omni.rtx=error` trial preserved `omni.rtx` errors during a forced GPU initialization failure and preserved a warning from the separate `omni.platforminfo.plugin` channel. However, useful `omni.rtx` warnings such as the CPU governor and IOMMU diagnostics disappeared.
- The current public Python `RendererConfig` exposes the global `log_level`, but not Carbonite channel rules.

My clear suggestion is:

1. Remove `suppress_native_logging()` and the fd-redirection call sites from this PR. The behavior is process-wide, can hide unrelated concurrent output, and is unreliable with asynchronous Carbonite delivery.
2. If removing the duplicate Python layer is still useful, use a scoped message-specific `DeprecationWarning` filter for only `Renderer.open_usd_from_string` and `Renderer.reset_stage`.
3. Do not set `omni.rtx=error` in Isaac Lab as the final solution, because it still suppresses warnings we may care about. The native duplicate should instead be addressed in OVRTX, ideally by avoiding duplicate Python/native deprecation reporting or by exposing a message-specific filtering mechanism.
4. Longer term, migrate the remaining legacy path to the already available ovstage APIs, which is the migration requested by the deprecation itself.

Until the native side offers narrower control or the legacy path is migrated, keeping the native deprecation visible is safer than suppressing the entire output stream or the whole `omni.rtx` warning channel.

AntoineRichard:
For reference, this is the narrowly targeted Python option I tested:

```python
import contextlib
import warnings


@contextlib.contextmanager
def _suppress_ovrtx_legacy_python_deprecations():
    with warnings.catch_warnings():
        warnings.filterwarnings(
            "ignore",
            message=r"Renderer\.(open_usd_from_string|reset_stage) is deprecated in ovrtx 0\.4\.",
            category=DeprecationWarning,
        )
        yield
```

It can be scoped directly around the two legacy calls:

```python
with _suppress_ovrtx_legacy_python_deprecations():
    self._renderer.open_usd_from_string(combined_usd_string)

# ...

with _suppress_ovrtx_legacy_python_deprecations():
    self._renderer.reset_stage()
```

The message regex is important: unlike a blanket `category=DeprecationWarning` filter, this preserved an unrelated `DeprecationWarning` in the same block during my probe.

This only suppresses the Python decorator warnings. The native Carbonite duplicates still appear. For comparison, the per-channel experiment used the following OVRTX Carbonite configuration:

```json
{
  "log": {
    "channels": {
      "omni.rtx": "error"
    }
  }
}
```

That retained `omni.rtx` errors and warnings from other channels, but removed every warning from `omni.rtx`, not only the deprecations. It therefore demonstrates that channel filtering works, but I would not recommend shipping that rule as the final Isaac Lab fix. The preferable native-side solution is for OVRTX to avoid the duplicate native warning or expose a narrower filter.

AntoineRichard (dismissed):
See my AI's comments, I would not merge as is.

AntoineRichard:
@mataylor-nvidia what's the status on this?

mataylor-nvidia:
> Did you try Python `warnings.filterwarnings("ignore", category=DeprecationWarning)`? How does it work?

Yes tried this it does not work :(

mataylor-nvidia:
I tested your suggestion @AntoineRichard but it was not work after the addition of the IsaacLab Loading screen

huidongc (approved):

kellyguo11 (approved):

pbarejko (approved):

PR #6961, opened 2026-08-07
[Docker] Accept the container profile before the command, and start from enter

hujc7:
## 1. Summary

- ``./docker/container.py <profile> <command>`` is now accepted, so stepping a container through its lifecycle only changes the last word: ``kitless build`` → ``kitless start`` → ``kitless enter``. **The existing ``<command> <profile>`` order keeps working.**
- ``enter`` now starts the container when it is not running, instead of failing with ``The container '<name>' is not running.``
- 3 lines of parsing logic, 15 tests, 14 documentation examples updated.

## 2. Why

Repeating the profile on every command puts the only varying word in the middle of the line, so the previous command cannot be reused by editing its tail. This was invisible while ``base`` was the default and usually omitted; the kit-less profile is the first one typed on every invocation.

The command set comes from argparse's own ``subparsers.choices`` rather than a second hardcoded list, so adding a subcommand needs no change here. Every subcommand takes exactly one positional — the profile — which is what makes the swap unambiguous; a test pins that profile names and command names stay disjoint.

## 3. ``enter`` behaviour change

``stop`` deletes the temporary ``.xauth`` file, so ``enter`` afterwards failed twice over: ``x11_refresh`` errored with ``X11 forwarding is enabled but the temporary .xauth file does not exist`` before the "container is not running" check was even reached. Running the same setup ``start`` does resolves both.

**Note for reviewers:** this means ``enter`` can now trigger a build on a cold cache, so it is no longer guaranteed-fast. It prints ``[INFO] Container '<name>' is not running. Starting it first...`` so the pause is explained. A test pins both directions, so a running container is never restarted.

## 4. Validation

```
15 passed
container.py base enter --info     → Profile: base      container.py enter base --info  → Profile: base
container.py kitless build --info  → Profile: kitless   container.py build kitless --info → Profile: kitless
container.py base start --suffix custom --info → Container Name: isaac-lab-base-custom
```

## 5. Type of change

- Documentation update

## 6. Checklist

- [x] I have added a changelog fragment for the touched Isaac Lab packages


AntoineRichard:
The new regression tests are not currently executed by CI. [docker/test/test_container_cli.py](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/docker/test/test_container_cli.py#L40-L101) covers both the profile-first parser and enter-from-stopped behavior, but the [Kit-less Docker workflow](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/.github/workflows/kitless-docker.yml#L44-L80) runs only test_container_profiles.py and test_dockerfile_nonroot.py. The general test runner defaults to source/, so this new file is not collected elsewhere either. Please add test_container_cli.py to both the workflow change-detection patterns and its pytest command, so these regressions remain enforced and test-only changes trigger the job.

AntoineRichard (commented):
## 🤖 AI-generated review (OpenAI Codex)

This review was generated by an AI coding agent and verified against commit `719e5e4c13992613de924d81e8dc33a21ffb1028`.

### Findings

1. **Important — New regression tests are not run by CI.**

   [`docker/test/test_container_cli.py`](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/docker/test/test_container_cli.py#L40-L101) is absent from both the change-detection patterns and pytest invocation in the [Kit-less workflow](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/.github/workflows/kitless-docker.yml#L44-L80). Consequently, neither new behavior is continuously enforced, and test-only changes would not trigger the job. Please add this file to both locations. This was also raised in [the existing PR comment](https://github.com/isaac-sim/IsaacLab/pull/6961#issuecomment-5238782823).

2. **Moderate — The X11 recovery behavior is not actually tested.**

   The stopped-container test forces [`x11_check()` to return `None`](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/docker/test/test_container_cli.py#L72-L93). Removing the new X11 setup would therefore leave this test green even though `enter` after `stop` would again terminate in `x11_refresh()` when X11 is enabled. Please add a stopped/X11-enabled case that returns the overlay and environment, then verifies they are applied before `start()`.

3. **Moderate — The `Changed` fragment lacks required migration guidance.**

   The repository guidelines require migration guidance for every `Changed` entry, but the [fragment](https://github.com/isaac-sim/IsaacLab/blob/719e5e4c13992613de924d81e8dc33a21ffb1028/source/isaaclab/changelog.d/jichuanh-container-profile-first.rst#L8-L14) only describes the new behavior. A concise addition could tell scripts that relied on `enter` failing for stopped containers to check container state explicitly before invoking it.

I found no implementation-level correctness defect in the argument rewrite or auto-start control flow.

### Verification

- `28 passed` across `test_container_cli.py` and `test_container_profiles.py`.
- Full `./isaaclab.sh -f` passed.
- With the base implementation temporarily restored, the new test file produced `12 failed, 3 passed`; restoring the PR implementation produced `15 passed`.
- No live Docker lifecycle test was run.

PR #6916, opened 2026-08-05
Restructure KaminoSolverCfg with nested sub-configs and enable DVI backend support

rubengrandia:
# Description

Restructures the unreleased `KaminoSolverCfg` API to mirror Newton's nested Kamino configuration layout and adds DVI support through separate concrete solver configuration classes.

## Summary

- **Breaking:** Replaced `KaminoSolverCfg` with `KaminoPADMMSolverCfg` and `KaminoDVISolverCfg`.
- **Configuration:** Shared settings live in nested sub-configurations (`dynamics`, `constraints`, `fk`, `collision_detector`, and `materials`); backend-specific settings live in `dynamics_solver_cfg`.
- **DVI:** `KaminoDVISolverCfg` selects Newton's DVI backend. When `dynamics` is not supplied, Newton selects dynamics defaults appropriate to DVI and the sparsity options.
- **Tasks:** Updated Cartpole, Ant, Fourbar Pole, DR Legs, and locomotion task presets to use the renamed concrete config classes. The task import changes are mechanical migration work.
- **Docs:** Updated the Kamino solver guide, solver comparison, Hydra documentation, and API reference for the nested fields and concrete config classes.
- **Tests:** Coverage verifies nested configuration values are forwarded to the selected Newton solver backend.

## Migration

```python
# Before
KaminoSolverCfg(padmm_max_iterations=200, padmm_rho_0=0.1)

# After
from isaaclab_newton.physics import KaminoPADMMCfg, KaminoPADMMSolverCfg

KaminoPADMMSolverCfg(
    dynamics_solver_cfg=KaminoPADMMCfg(max_iterations=200, rho_0=0.1),
)
```

For DVI:

```python
from isaaclab_newton.physics import KaminoDVICfg, KaminoDVISolverCfg

KaminoDVISolverCfg(
    dynamics_solver_cfg=KaminoDVICfg(max_alternating_iterations=20),
)
```

Hydra overrides use the nested path:

```bash
uv run isaaclab train --rl_library rsl_rl --task Isaac-Cartpole \
  env.physics.solver_cfg.dynamics_solver_cfg.max_iterations=200
```

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (existing functionality will not work without user modification)
- [x] Documentation update

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
[...]

PR #6964, opened 2026-08-07
Fix Kamino camera crash: disable internal collision detector for cartpole

AntoineRichard:
## Summary

- Fixes `RuntimeError: Cannot perform collision detection: a collision pipeline has not been created` when running `Isaac-Cartpole-Camera` with `physics=newton_kamino`
- Sets `use_collision_detector=False` in the cartpole `newton_kamino` preset — cartpole bodies never contact each other or the ground, so Kamino's internal collision detector serves no purpose
- Adds `collision_detector_max_contacts_per_world` field to `KaminoSolverCfg` for tasks that do need the internal detector on models without pre-computed explicit broadphase pairs

## Root cause

The updated Newton library only creates the internal Kamino collision pipeline when `model_minimum_contacts > 0` (pre-computed explicit broadphase pairs). The cartpole has none, so the pipeline stays `None`. However, `ContactsKamino` silently substitutes zero capacities with a default of 128 per world, leaving `model_max_contacts_host > 0`. When `detector.collide()` runs, it bypasses the early-return guard but finds no pipeline → crash.

This only manifested with cameras because `cameras_enabled=True` sets `_clone_physics_only=False` → `_usdrt_stage` is set → CUDA graph capture is deferred to the first `step()`, where the warmup executes `_simulate_physics_only()` eagerly and hits the broken state. Without cameras the standard `wp.ScopedCapture` path is used instead.

## Test plan

- [x] `uv run --extra isaacsim isaaclab train --rl_library rsl_rl --task Isaac-Cartpole-Camera physics=newton_kamino` — confirmed no longer crashes
- [x] `uv run isaaclab -f` — all pre-commit checks pass

### DIFF
--- a/source/isaaclab_newton/changelog.d/antoiner-kamino-collision-detector-max-contacts-per-world.rst
+++ b/source/isaaclab_newton/changelog.d/antoiner-kamino-collision-detector-max-contacts-per-world.rst
@@ -1,0 +1,8 @@
+Added
+^^^^^
+
+* Added :attr:`~isaaclab_newton.physics.KaminoSolverCfg.collision_detector_max_contacts_per_world`
+  to :class:`~isaaclab_newton.physics.KaminoSolverCfg`. When set, this overrides Newton's
+  geometry-based contact capacity estimation for Kamino's internal collision detector and guarantees
+  the collision pipeline is created regardless of whether the model has pre-computed explicit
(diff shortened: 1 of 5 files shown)


AntoineRichard:
Closing in favor of https://github.com/isaac-sim/IsaacLab/pull/6916

PR #6971, opened 2026-08-07
[Docker][Kit]Let OmniHub start in the Isaac Sim based containers

hujc7:
## 1. Summary

- Fixes nvbug 6564035 / internal 926: ~39 ``Hub failed to launch`` warnings and a
  ~10 s stall at every Kit startup in the Isaac Sim based containers.
- The Isaac Sim image sets ``HUB__ARGS__DETECT_ONLY=true``, forbidding OmniHub to
  start, while ``omni.client`` defaults to ``shared`` mode and asks it to launch
  anyway. The retries are those two settings contradicting each other.
- ``ENV HUB__ARGS__DETECT_ONLY=false`` in ``Dockerfile.base`` + ``Dockerfile.curobo``,
  and ``/var/cache/hub`` chowned to the runtime user so OmniHub can use its cache.
- Cartpole training in ``isaac-lab-base``: 39 → 0 warnings, exit 0 both runs.

## 2. Notes

- The value must be ``false``, not empty: ``--detect-only`` takes a value, so
  ``ENV HUB__ARGS__DETECT_ONLY=`` aborts hub with ``a value is required`` instead.
  A Dockerfile cannot remove an inherited ``ENV``. A test pins the exact string.
- Removes the ad-hoc ``unset HUB__ARGS__DETECT_ONLY`` from ``run_tests.sh`` and
  ``multi_gpu_shard_runner.sh``, now superseded by the image.

## 3. Type of change


AntoineRichard:
This conflicts with https://github.com/isaac-sim/IsaacLab/pull/6985 in the direction we want to take. Do we want to remove omnihub? If we don't need it I'd be for removing it. @ooctipus was saying this is possible.

hujc7:
> This conflicts with #6985 in the direction we want to take. Do we want to remove omnihub? If we don't need it I'd be for removing it. @ooctipus was saying this is possible.

It should not conflict. The other one is kitless, and this is in isaacsim base docker. My understanding is  that it does some caching? no strong opinion on keep or just disable.

kellyguo11 (approved):

PR #6953, opened 2026-08-07
Fix kit-less container bugs and ship OVPhysX plus all Newton viewers

hujc7:
## 1. Summary

- Fixes 4 QA-reported kit-less container bugs (nvbug 6569001 / 6569031 / 6569217 / 6569864), each verified against QA's verbatim command.
- `container.py` no longer reports success on a failed Docker command — it exited `0` after printing `Finished building`, which is why 6569864 arrived as a screenshot with no error text.
- Kit-less image gains OVPhysX and the Viser/Rerun viewers: `physics=ovphysx` and `--viz newton,viser,rerun` now work out of the box. Image 14.3 GB → 15.4 GB (+7.7%).
- `libxrender1` added: `--viz newton` crashed with `AttributeError: 'NoneType' object has no attribute 'XRenderFindVisualFormat'`.

## 2. Why OVPhysX was excluded, and why it no longer needs to be

`ovphysx` caps `packaging<24`; `ray` (via `rl[all]`) needs `>=24.2`. Installing OVPhysX downgraded `packaging` to 23.2 and broke `ray` — a real conflict, so the exclusion was correct.

The cap is over-strict: ovphysx's entire use of the library is one `from packaging.version import Version` in `ovphysx/api.py`. A `packaging>=26.0` override reconciles it, as the neighbouring overrides already do for other over-strict pins. The in-file comment claimed an OmniClient pin instead; `omniclient` appears nowhere in `uv.lock`.

## 3. Validation

Run against the rebuilt image with QA's verbatim commands:

| nvbug | Command | Result |
|---|---|---|
| 6569001 | `train … physics=ovphysx renderer=ovrtx` | rc 0, 0 tracebacks, `Training time: 155.44 s` |
| 6569031 | `play …`, empty log dir | `ValueError: No runs present in the directory` (was a bare `FileNotFoundError`) |
| 6569217 | `play … --viz newton,viser,rerun` | rc 124 (full 300 s window), 0 tracebacks; viser `:8080`, rerun `:9090` serving |
| 6569864 | forced Docker failure | exits `1`, no `Finished building` (was exit `0`) |

Newton regression: `presets=newton_mjwarp` → `Training time: 2.06 s`. Build gated by in-image assertions on `ovphysx`, `ovrtx`, `viser`, `rerun-sdk`.

## 4. Notes

- nvbug 6569864's reported symptom is a host `docker.sock` permission error, not an Isaac Lab defect — but it surfaced the silent-success bug fixed here.
[...]

PR #6972, opened 2026-08-07
Scan kitless Docker dependency licenses

kellyguo11:
## Summary

- reuse one Docker dependency-license policy for the Isaac Sim-based and kitless images
- scan the kitless runtime image and its `ubuntu:24.04` base with Trivy
- separate inherited Ubuntu packages from dependencies distributed by Isaac Lab
- reject unreviewed non-permissive licenses, in-process GPL, and unsupported static LGPL linkage
- fail closed when Trivy inventories a package but reports no license metadata
- retain the standalone FFmpeg guard and upload the complete kitless scan reports

## Motivation

The kitless image is a separately distributed runtime artifact, but only the Isaac Sim-based Docker image currently receives the strict dependency-license check. This is especially important as #6953 proposes adding OVPhysX, Viser, Rerun, and another runtime system library to the kitless image.

The scan runs in the existing kitless build job, so it reuses the already-built image instead of adding another large Docker build.

## #6953 audit evidence

A temporary branch combining #6953 with this scanner completed the image build, runtime validation, Ubuntu scan, final-image scan, and standalone FFmpeg guard in [run 31210321067](https://github.com/isaac-sim/IsaacLab/actions/runs/31210321067).

That audit used #6953 head `e78eead0`. The current head `f0995a92` leaves `docker/Dockerfile.kitless` and #6953's added viewer extras unchanged. Its dependency-lock delta only updates Newton, Newton USD schemas, and Warp versions from the subsequently merged `develop`, so it does not add another #6953-specific package to the audited set.

The final-image delta contained eight new Python license findings: ImageIO (BSD-2-Clause), msgspec (BSD-3-Clause), OVPhysX (NVIDIA Omniverse), PyArrow (Apache-2.0), Rerun (MIT OR Apache-2.0), Viser (MIT), websockets (BSD-3-Clause), and zstandard (BSD-3-Clause). It contained no new GPL/LGPL finding.

The audit also found two new inventory records with no Trivy license finding: `libxrender1` and Viser's embedded private Node package. Viser 1.0.30 also has inconsistent top-level metadata: PyPI declares MIT while the release-tag `LICENSE` is Apache-2.0. Both are permissive, but the provenance and third-party notices for the bundled web client need to be reconciled. The shared policy now reports any package without scanner license metadata as `UNKNOWN` and requires explicit review instead of silently omitting it.

## Validation

- `bash -n tools/license/check_docker_licenses.sh`
- parsed both modified workflows with PyYAML
- replayed the policy against the #6953 Trivy artifact; the expected fail-closed report contained 1,481 findings and explicitly flagged both missing-license records

The PR's self-hosted kitless job provides the authoritative image scan.

## Type of change

- New feature (CI license enforcement)

## Checklist

- [x] My changes generate no new local warnings
- [x] Documentation changes are not required for this CI-only change
- [x] A changelog fragment is not required because no package source changed


AntoineRichard (changes requested):
I reviewed this against base `3fbd73278639981ffbf091bee8fd70824acb8429` and head `557aa7c9f8d3424c0d0381b56277e5c0e01234a1`, including both uploaded Trivy report artifacts.

The direction is good: one policy for both distributed Docker images, explicit treatment of missing scanner metadata, and retaining the standalone FFmpeg guard are all useful. I am requesting changes because the current policy is not yet mergeable and two code paths can allow inventory or LGPL obligations to escape the intended fail-closed behavior.

## 1. Important: establish a passing reviewed baseline

Location: `.github/workflows/kitless-docker.yml:226` and `.github/workflows/license-check.yaml:301`.

Both Trivy scans complete, but the shared policy rejects the images produced at this head:

- Kit-less Docker job: 428 violations.
  - 248 have no reviewed exception.
  - 171 are inventoried packages with no Trivy license metadata.
  - 8 match an exception that lacks the newly required `usage`.
  - 1 license finding cannot be associated with a version.
- Existing Isaac Sim-based Docker job: 176 violations.
  - 170 have no Trivy license metadata.
  - 5 match an exception that lacks `usage`.
  - 1 license finding cannot be associated with a version.

This is therefore not incidental runner noise: both authoritative jobs fail specifically in `check_docker_licenses.sh`. Since the PR does not update the exception baseline or provide an alternate source for inventories that Trivy cannot license (notably compiled Go modules), the new enforcement cannot pass for either currently distributed image.

Please make both current images a reviewed, passing baseline before merge:

1. Deduplicate the reports by a stable package identity, not merely by display row.
2. Classify each non-permissive or missing-license identity.
3. For a reviewed exception, provide `package_type`, a precise license or aliases, `usage`, `interaction`, and linkage/relinking data where applicable.
4. Where Trivy inventories a component but cannot produce license evidence, use a documented alternate evidence/normalization path or make an explicit, narrowly scoped inventory decision. Do not add blanket package-prefix or category-wide exceptions merely to reduce the count.
5. Rerun both Docker jobs and require both generated summaries to report zero violations.

Because this workflow and policy script are themselves kit-less change-detection inputs, leaving the baseline red would also block future relevant PRs.

## 2. Important: preserve individual package identities

Location: `tools/license/check_docker_licenses.sh:81-120`, especially the `$versions` map at line 91 and `$licensed_packages` map at line 120.

Both maps are keyed only by lower-cased `package_type + package name`. `from_entries` retains one version for a duplicated name, while the presence of any license row marks every same-named inventory row as licensed.

The actual kit-less artifact demonstrates the loss:

- Inventory contains `certifi` 2026.6.17 and 2026.7.22. The checker maps the name to 2026.7.22 and emits no `UNKNOWN` row for 2026.6.17.
- Inventory contains `typing_extensions` 4.15.0 and 4.16.0. The checker maps the name to 4.16.0 and emits no `UNKNOWN` row for 4.15.0.
- `aiohttp` 3.14.1 and 3.14.3 have distinct Trivy license records and paths, but the generated summary retains only a 3.14.3 finding after sorting.

This is fail-open. If the arbitrarily selected version is inherited from the base, an additional version introduced by the final image may be counted as inherited or disappear entirely.

Please retain one finding per stable inventory identity. Prefer Trivy's `FilePath` when both package and license rows provide it; use PURL/UID/ID plus version as fallbacks. A same-named row that cannot be correlated safely must remain a separate `UNKNOWN` finding rather than borrowing another version's license.

Required regression fixture:

- final inventory: `example` versions 1 and 2;
- base inventory: only `example` version 2;
- license metadata associated only with version 2.

The result must independently report/evaluate version 1. Also cover two installation paths containing different versions where both paths have license rows.

## 3. Important: make static interaction authoritative for LGPL

Location: `tools/license/check_docker_licenses.sh:254-289`, especially line 282.

The accepted interaction values include `same_process_static`, but the relinking requirement runs only when the separate `linkage` field equals `static`. This exception currently reaches `EXCEPTED_FINDINGS` without relinking material:

```json
{
  "package": "example-lgpl",
  "package_type": "Python",
  "license": "LGPL-3.0-only",
  "usage": "runtime_dependency",
  "interaction": "same_process_static"
}
```

Here `is_lgpl=true`, but `linkage` is empty, so the static branch is skipped. Declaring `interaction: same_process_static` together with `linkage: dynamic` also passes despite the contradictory state. That contradicts the stated guarantee that unsupported static LGPL linkage is rejected.

Please enforce one invariant:

- `same_process_static` implies static linkage even when `linkage` was omitted;
- contradictory interaction/linkage values are violations;
- every LGPL record determined to be static requires non-empty `relinking_materials` and `relinking_instructions`.

Add negative fixtures for omitted and contradictory linkage, plus positive fixtures for explicitly dynamic LGPL and fully documented static LGPL.

## 4. Moderate: share the Docker-license orchestration

Locations: `.github/workflows/kitless-docker.yml:164-246` and `.github/workflows/license-check.yaml:229-324`.

These jobs repeat the base pull, two Trivy scans, FFmpeg inspection, policy invocation, and artifact upload. They already differ slightly in entrypoint handling and Python probing, which makes scanner-version and fail-closed changes prone to drift.

Recommended boundary:

1. Add a local composite action at `.github/actions/check-docker-licenses/action.yml`.
2. Inputs should include `image-ref`, `base-image-ref`, `summary-title`, `base-image-label`, and `artifact-name`.
3. Use fixed internal report filenames. The action should pull/scan both images, invoke the policy, and upload every report under `if: always()`.
4. Move the FFmpeg inspection into a shared host-side script. It can consistently use `docker run --entrypoint bash` and choose `./isaaclab.sh -p` versus direct `python` inside the container based on the image contract.
5. Reduce each workflow caller to one local-action step after ensuring its final image is locally available.

A reusable workflow is a worse fit because the built image is already in the current job's Docker daemon; a new job may run on another runner and require another multi-gigabyte transfer. Keep the top-level Python `pip-licenses` job separate for now because it has a different inventory format and policy semantics.

## Autonomous completion checklist

An implementation agent should be able to complete the revision from this checklist without further design decisions:

- Add fixture-driven policy tests covering:
  - an allowed permissive license;
  - an unreviewed rejected license;
  - an exactly inherited base identity;
  - a package with two versions where only one exists in the base;
  - one same-named inventory record with license metadata and another without it;
  - GPL with same-process interaction;
  - dynamic LGPL;
  - static LGPL with absent, contradictory, incomplete, and complete relinking data;
  - missing inventory version;
  - missing license metadata.
- Ensure every unique current-image identity is classified as inherited, allowed, reviewed, or violating. No identity may disappear merely because another package shares its name.
- Review the current 428 and 176 violations and provide precise evidence/metadata rather than broad suppression.
- Extract the shared Docker orchestration while preserving `if: always()` report upload behavior on policy failure.
- Rerun both self-hosted Docker license jobs and attach passing summaries. Passing synthetic fixtures alone are insufficient.
- Run `bash -n tools/license/check_docker_licenses.sh`.
- Parse both modified workflow YAML files.
- Run `./isaaclab.sh -f`.
- Review any formatter changes, rerun pre-commit, and only then commit/push.

## Verification performed for this review

- `git diff --check`: passed.
- `bash -n tools/license/check_docker_licenses.sh`: passed.
- Both modified workflow YAML files parsed successfully.
- I inspected both CI-uploaded Trivy JSON reports and generated Markdown summaries.
- I did not rebuild either large Docker image locally; the self-hosted CI artifacts are the runtime evidence.

PR #6973, opened 2026-08-07
[Audit only] Scan #6953 kitless image licenses

kellyguo11:
## Temporary license audit

This draft combines the head of #6953 with the kitless license-scanning change from #6972 so GitHub's self-hosted runner can produce an image-level Trivy report.

Do not merge. This PR will be closed after the report is collected and compared with #6972's baseline scan.


kellyguo11:
Temporary audit complete. The #6953 image built and validated, its Trivy artifacts were compared against the kitless baseline, and the results are documented on permanent scanning PR #6972.

PR #6932, opened 2026-08-06
Enable synchronous texture streaming for IsaacSim RTX

huidongc:
# Description

Disable async texture streaming in the headless Kit app while keeping texture streaming itself enabled:

```ini
'rtx-transient'.resourcemanager.enableTextureStreaming = true
'rtx-transient'.resourcemanager.texturestreaming.async = false
```

**Motivation:** With async texture streaming enabled, Isaac RTX camera renders could vary across runs, which breaks golden-image / pixel-diff rendering correctness tests. Synchronous texture streaming restores stable pixel output.

**Scope:** `apps/isaaclab.python.headless.kit` only. Texture streaming remains on; only the async path is turned off.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

### Pixel determinism

Re-ran Isaac RTX rendering correctness tests (`lift_kuka_hetero`, `lift_kuka_homo`, `shadow_hand`; AOVs: `rgb` / `rgba` / `albedo` / `simple_shading_diffuse_mdl` / `simple_shading_full_mdl`) **20 times** after this change.

- All cases: **0.00%** PixelDiff vs golden on every run (`N=20`, std=`0`)
- Actual images were **byte-identical** across runs (same SHA256)

This confirms the setting restores deterministic Isaac RTX render output for these tests.

### Performance / memory impact

Dexsuite runtime benchmark on `Isaac-Lift-KukaAllegro-Camera` (`newton_mjwarp` + `isaacsim_rtx`, seed `42`, 1000 measured iterations) before (`develop` / `62e1433e`) vs after (this change).

| Variant | Envs | FPS Δ% | Env create Δ% | First step Δ% | GPU mem peak Δ% | RAM peak Δ% |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| homogeneous | 1024 | −2.98% | +2.99% | +0.03% | +2.54% | −0.12% |
| homogeneous | 2048 | +1.04% | −0.52% | +1.24% | −1.11% | +0.43% |
| homogeneous | 4096 | +0.33% | +1.54% | +1.25% | −2.31% | −0.09% |
| homogeneous | 8192 | −0.95% | −1.65% | −6.06% | −1.51% | +0.06% |
| heterogeneous | 1024 | −0.77% | +1.74% | +3.77% | −4.58% | +0.12% |
| heterogeneous | 2048 | −0.59% | +1.35% | +3.51% | −3.38% | +0.64% |
[...]

PR #6976, opened 2026-08-07
Decouple rendering tests from task environments

ooctipus:
# Description

Rendering correctness tests previously instantiated complete registered task environments. That loaded manager stacks and task assets the tests did not exercise, coupled golden images to reset/task implementation changes, and repeated expensive setup for visually redundant scenes.

This PR makes rendering tests own their scenes and moves cross-package coverage to a private downstream testing composition root.

## Architecture

- Adds the private, test-only `source/_isaaclab_testing` project. It depends downstream on the runtime packages it integrates; production packages do not import or depend on it.
- Keeps the empty `isaaclab-testing` distribution non-public and non-importable. Rendering is one suite under `test/rendering`, leaving room for future integration suites without renaming the project again.
- Consolidates collection into one Kit renderer module, one Kitless renderer module, and one visualizer module. Those thin roots select cases; shared runners own execution and validation.
- Makes `RenderingSceneSpec` the scene composition record. It owns the direct scene type, physics/runtime inputs, tolerances, and expected segmentation instances. `RenderCase` owns only the backend/renderer/AOV combination.
- Keeps reset policy in `mdp.reset_scene_to_default`. `InteractiveScene` gains only the lifecycle operation needed by direct scene owners to release callbacks before simulation teardown.
- Adds architecture gates for the intended file tree, downstream dependency direction, forbidden task imports/old test roots, exact case partition coverage, unique identities, and golden inventory derivation.

The result removes the former per-task/per-visualizer test implementations and their duplicated helpers. Defining a new rendering scene now requires scene configuration and registry data, not another complete test.

## Direct scenes and fidelity

No rendering test imports `isaaclab_tasks` or constructs an RL environment. The suite locally declares the visible facts that rendering is meant to cover:

- The shared full-AOV scene uses the textured standard grid asset, an explicit-size procedural table, varied primitive geometry, a cartpole articulation at its true default joint state, semantic labels, and deliberate non-overlapping placement.
- Cartpole remains explicit coverage because it has caught scene-partitioning failures. Its asset, ground, lighting, camera pose, and default joint state mirror the task; only output resolution is raised for reviewability.
- Franka Soft and Franka Cloth mirror the current task asset, actuators, failure-state table, deformables, camera, spacing, supports, and Newton contact/material tuning. This includes the latest cloth density and support-friction retune from `develop`.
- Kuka Allegro heterogeneous mirrors the task robot, table, failure color, camera, and representative cuboid/sphere/capsule/cone object topology. Its tiled output is raised to the same reviewable scale as OVRTX.
- Shadow Hand mirrors the backend-specific task robot, task camera/orientation, and the full-size DexCube at `(0.0, -0.39, 0.6)`. The known backend-dependent hand/cube proximity remains tracked separately as issue 6593722.
- Visualizer coverage reuses the same direct lifecycle and now views the scene from above the table.

The full-AOV camera renders at 256×256. Task probes use their task camera composition with reviewable resolution. Direct PreviewSurface and MDL authoring preserves authored colors in Kitless OVRTX as well as Isaac RTX; Franka Cloth therefore has the same pink table and yellow cloth in both paths.

## Golden and motion contracts

- The suite owns 121 references: 113 renderer goldens and 8 visualizer goldens.
- Deprecated OVRTX and ovstage reader paths are both tested, but `legacy` and `ovstage` are pytest/artifact identities only. They resolve to the same stage-independent golden filenames, so a disagreement is a failure rather than two accepted images.
- Case IDs are scene-first: `<scene>-<physics>-<renderer>[-<real-variant-or-single-AOV>]`. The former `canonical`, `probe`, and no-op `standard` labels are removed.
- Camera outputs are captured together. Motion-vector cases then advance physics exactly once and replace only the motion output with the second capture.
- Deterministic PhysX/OVPhysX motion retains image comparison. Newton motion is behavior-checked for finite, non-zero, localized vectors because its magnitude is not stable enough for a golden.
- Instance segmentation validates rendered colors, reserved entries, both metadata maps, and prim-path/class pairings against each scene's declared instances.
- AOV names use `RenderBufferKind`; scene-specific tolerances remain beside the scene facts that justify them.
- OVRTX golden-test setup disables texture streaming before renderer construction because that path has no Kit USD-context streaming wait; Isaac RTX keeps its normal streaming synchronization. Isaac RTX RGB/albedo and Kit visualizer comparisons use a shared cross-GPU color-noise floor while retaining scene-owned spatial-difference and SSIM gates. Legacy OVRTX and ovstage therefore keep one shared expected image.
- The shared semantic scene requires a minimum visible table area, so a backend scale regression fails even when its semantic label is still present.

## CI organization

Rendering jobs target `_isaaclab_testing`, preserving the dependency direction `isaaclab -> isaaclab_ov -> _isaaclab_testing`. Cold-cache selection is marker-driven. Process-isolated partitions are derived from the case matrices and gated so every case is selected exactly once; a future unmatched case cannot silently run in zero CI passes. Kitless USD parsing sets `PXR_WORK_THREAD_LIMIT=1` before USD import.

## Screenshots

These previews are pinned to commit `867555fe37170ef7487cad0dd79e03cb07b72d07`.

| Shared full-AOV scene | Kuka Allegro heterogeneous |
| --- | --- |
| Shared full-AOV scene | Kuka Allegro heterogeneous |
| Franka Cloth — Isaac RTX | Franka Cloth — OVRTX |
| Franka Cloth Isaac RTX | Franka Cloth OVRTX |
| Franka Soft — Isaac RTX | Franka Soft — OVRTX |
| Franka Soft Isaac RTX | Franka Soft OVRTX |
| Cartpole | Shadow Hand |
| Cartpole | Shadow Hand |

## Validation

- Defined matrix: 20 Kit cases, 63 Kitless cases, and 8 visualizer cases.
- Full Kit matrix: 20/20 passed across the intended isolated partitions.
- Visualizer matrix: 8/8 passed.
- Latest architecture and golden-image regression suite: 19/19 passed.
- Existing Kit material spawning suite: 6/6 passed.
- Legacy and ovstage OVRTX RGB/albedo pairs: 2/2 passed in each path against the same 256×256 references.
- Real Kitless ovstage instance-segmentation render and metadata contract: passed.
- Franka Cloth Isaac RTX and colored OVRTX comparisons: passed.
- Franka Soft Kit cases, including tetrahedralization-dependent cases: passed.
- Repository pre-commit/format suite, changelog validation, and `git diff --check`: passed.
- Rebased onto `develop` at `ed1f83d939c`; no unresolved files or upstream back-references remain.

## Type of change

- Test architecture refactor and rendering-coverage improvement (non-breaking)

## Checklist

- [x] Documentation changes are not required for this internal test refactor
- [x] I have added tests that prove the architecture and behavior
- [x] The private test project has a test-only changelog skip fragment


StafaH (approved):

source/_isaaclab_testing/test/rendering/rendering_cases.py
```diff
@@ -0,0 +1,163 @@
...
+    *_single_aov_cases("newton", "ovrtx", OVRTX_AOVS),
+    RenderCase("ovphysx", "newton_warp", NEWTON_WARP_AOVS),
+    RenderCase("newton", "newton_warp", NEWTON_WARP_AOVS),
+    *_simple_shading_cases("ovphysx", "ovrtx"),
+    RenderCase("ovphysx", "ovrtx", ("rgb",), profile="yellow-background", background_color=(1.0, 1.0, 0.0)),
+]
+KITLESS_CASES = tuple(
+    (stage, case)
+    for case in _KITLESS_CASES
+    for stage in (("legacy", "ovstage") if case.renderer == "ovrtx" else ("legacy",))
```
hujc7, line 120: So this is only for ovphysx and ovrtx right? for newton, it should be empty?

tools/test_settings.py
```diff
@@ -16,6 +16,72 @@
 """The default timeout for each test in seconds."""
 
 
+COLD_CACHE_TESTS = {
```
hujc7: can this be based on pytest marker?

source/isaaclab/test/test_golden_scene_architecture.py
```diff
@@ -0,0 +1,534 @@
...
+            "fill_light",
+            "camera",
+            "robot",
+            "object",
+            "table",
+        },
+        "ShadowHandRenderingSceneCfg": {"ground", "key_light", "fill_light", "camera", "robot", "object"},
+    }
+
+    required_signatures = (
```
hujc7: seem a bit verbose, what's the purpose?

source/isaaclab/test/renderers/rendering_scene_cfgs.py
```diff
@@ -0,0 +1,443 @@
...
+
+    ground = None
+    key_light = None
+    fill_light = _SHADOW_LIGHT.copy()
+    camera: CameraCfg = _SHADOW_CAMERA.copy()
+    robot: ArticulationCfg = _SHADOW_HAND_NEWTON.copy()
+    object: ArticulationCfg | RigidObjectCfg = _SHADOW_OBJECT_NEWTON.copy()
+
+
+def make_rendering_scene_cfg(
```
hujc7: **Agent review:** this returns six values that `run_rendering_case` destructures and re-threads as six keyword arguments into `build_rendering_scene` — a grouping that is real but unnamed, so a reader has to reconstruct it.

Naming it is also what would let one registry own the per-scene facts that are currently spread across this branch, three dicts in `rendering_cases.py`, two conditionals in `rendering_runner.py`, and five sets in the architecture gate. `isaaclab.renderers.CameraRenderSpec` is a nearby precedent for the shape.

source/isaaclab/isaaclab/test/utils/rendering.py
```diff
@@ -0,0 +1,244 @@
...
+    else:
+        raise ValueError(f"Unknown renderer: {renderer!r}")
+    if hasattr(cfg, "semantic_filter"):
+        cfg.semantic_filter = ["class"]
+    if hasattr(cfg, "semantic_segmentation_mapping"):
+        cfg.semantic_segmentation_mapping = SEMANTIC_COLORS.copy()
+    return cfg
+
+
+def make_physics_cfg(physics_backend: str) -> PhysicsCfg:
```
hujc7: **Agent found bug:** this shadows `isaaclab.app.sim_launcher.make_physics_cfg`, which already ships in this package, with different semantics — `"physx"` returns `PhysxAutoCfg(...)` there and `PhysxCfg(...)` here. Core also takes `"newton_mjwarp"` / `"isaacsim_physx"` / `"newton_vbd"` and rejects `"newton"`; this one is the reverse.

Two importable functions with one name where the same argument yields a different backend is easy to import wrongly. Develop's harness translated `"newton"` into core's vocabulary via `_physics_preset_name()`; this replaces the translation with a fork.

- [ ] rename to `make_rendering_physics_cfg`, or take core's labels and layer the test-only determinism flags on its result

source/_isaaclab_testing/test/rendering/rendering_cases.py
```diff
@@ -0,0 +1,163 @@
...
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+SIMPLE_SHADING_AOVS = (
+    "simple_shading_constant_diffuse",
+    "simple_shading_diffuse_mdl",
+    "simple_shading_full_mdl",
+)
+RTX_AOVS = (
```
hujc7, line 19: **Agent suggestion:** these AOV names already exist as a typed enum.

`isaaclab.renderers.output_contract.RenderBufferKind` is a `StrEnum` enumerating exactly these strings — `rgb`, `albedo`, `motion_vectors`, `instance_id_segmentation_fast`, `simple_shading_full_mdl` and the rest — and its docstring states the values match `CameraCfg.data_types`.

Because it is a `StrEnum` it substitutes anywhere a `str` is accepted today, so this is a typing change rather than a refactor. The gain: a typo becomes an `AttributeError` at import instead of an AOV that silently matches nothing, and the four AOV tuples plus `_NO_SSIM`, `_SEMANTIC_ONLY_AOVS` and the inline `alpha_only` set all key off one declared vocabulary.

source/isaaclab/test/renderers/rendering_runner.py
```diff
@@ -0,0 +1,160 @@
...
+_ARTIFACT_DIR = Path.cwd() / "tests" / "comparison-images" / "images"
+_NO_SSIM = {
+    "depth",
+    "distance_to_camera",
+    "distance_to_image_plane",
+    "instance_segmentation",
+    "instance_id_segmentation_fast",
+}
+_SEMANTIC_ONLY_AOVS = {"motion_vectors"}
+_MAX_DIFF_PCT = {
```
hujc7: **Agent suggestion:** these tolerances read as properties of the scenes rather than of the runner.

`_MAX_DIFF_PCT` is keyed by scene name here, and two more scene-conditionals follow inside `run_rendering_case` — the `canonical_rtx_rgb` branch and `0.95 if case.scene == "kuka_heterogeneous"`. So the shared runner has to know every scene by name, which makes it harder to read as generic machinery.

If the scene configuration grows a home (see the comment on `make_rendering_scene_cfg`), these would sit next to the geometry that justifies them, where the measured-variance comments are checkable against the scene they describe.

source/isaaclab/test/test_golden_scene_architecture.py
```diff
@@ -0,0 +1,534 @@
...
+    rendering_entrypoints.add("source/isaaclab_visualizers/test/test_visualizer_rendering.py")
+    assert rendering_entrypoints <= assignments["COLD_CACHE_TESTS"]
+    implicit_camera_token = "enable_cameras" + "=True"
+    former_implicit_callers = {
+        path.relative_to(_REPO_ROOT).as_posix()
+        for path in (_REPO_ROOT / "source").rglob("test_*.py")
+        if implicit_camera_token in path.read_text()
+    }
+    assert former_implicit_callers <= assignments["COLD_CACHE_TESTS"]
+    assert assignments["PROCESS_ISOLATED_TESTS"] == {
```
hujc7: **Agent review:** these assert the config equals a literal copy of itself — the only way to fail is to change the config, which is the intended operation.

Same pattern twice more in this file: `len(KIT_CASES) == 8` / `len(KITLESS_CASES) == 46` assert the length of module-level literals, and the `"selector in target for selector in selectors..."` check pins verbatim source text from `tools/conftest.py`, so reformatting that file breaks a test in another package. If the counts are meant as a cost tripwire, a budget says so: `assert len(KIT_CASES) <= 10, "each case costs ~60 s of GPU time and one committed golden"`.

`test_golden_inventory_matches_case_matrix` shows the alternative — it computes its expectation from `RenderCase.golden_id`, so it needs no edit when the matrix grows.

Scope: the three above only, not the `sum(map(len, partitions))` / `set().union(*partitions)` relations, which are derived and should not change.

tools/test_settings.py
```diff
@@ -67,22 +131,41 @@
...
+    "test_rendering_scene_probes_kitless_ovstage_ovphysx.py": 1200,
     "test_contact_sensor.py": 2000,
 }
 """A dictionary of tests and their timeouts in seconds.
 
 Note: Any tests not listed here will use the default timeout.
 """
 
+
+PROCESS_ISOLATED_TESTS = {
```
hujc7, line 79: **Agent found bug:** nothing asserts these partitions cover every case, so a future case can run in zero passes while the file reports green.

The runner invokes a partitioned file once per partition under `-k`, matched as a substring against the test ID. Add a scene whose name matches none of the selectors and its cases are collected by no pass — and the gate's `PROCESS_ISOLATED_TESTS == {...}` equality still passes, because this dict did not change. Latent rather than live: today's selectors do cover every case (CI: 10/10 and 4/4).

- [ ] assert every case in a partitioned file is selected by exactly one partition, computed from the case tuples — which also retires `tuple(map(len, partitions)) == (4, 10, 3, 0)` in the gate

**Agent suggestion**, on the `COLD_CACHE_TESTS` question above: `tools/_device_split.py` already reads a file's module-level `pytestmark` from source *without importing it*, and `run_individual_tests` already loads `test_content` for that call — so markers work with the mechanism already here, and would drop the 60-path list.

ooctipus:
Review follow-up is pushed in dc68b7a2c19.

- Consolidated the rendering suite to one Kit file and one Kit-less file; native palette/lifecycle isolation remains in process-partition metadata.
- Added a named RenderingSceneSpec and one SceneProbe registry, moved scene tolerances beside scene composition, adopted RenderBufferKind, and renamed the test-only physics factory.
- Replaced the 60-path cold-cache set with a pytest marker.
- Replaced literal counts/source copies with derived coverage: every case must select exactly one nonempty partition, with a 15-scene budget.
- Clarified that ovstage selects the OVRTX USD reader and is valid with both Newton and OVPhysX; only Newton-Warp has no ovstage cases.

The follow-up removes 326 net lines. Architecture 11/11, marker tests 8/8, representative Kit and Kit-less renders, and full pre-commit all pass.

ooctipus:
Addressed in 863ec8d0423. Both OVRTX execution paths remain in the 63-case Kit-less matrix; legacy still exercises the deprecated clone_usd/open_usd_from_string path and ovstage still exercises the replacement API, but both now resolve to the same stage-independent golden filename. I removed the 23 duplicate ovstage PNGs and dropped legacy-/ovstage- from all Kit-less baselines, including Newton-Warp. The retained RGB pairs are within the existing visual tolerances. OVRTX semantic IDs can receive different arbitrary colors from the two USD readers, so semantic images compare their rendered mask while the existing idToLabels assertions validate label presence and reject unexpected labels. Canonical means the broad synthetic scene used for the full AOV matrix; probe means a small task-faithful specialized scene. That distinction is used only in pytest IDs to form process-isolated groups under the 15-scene compilation budget and never participates in golden identity.

source/_isaaclab_testing/test/rendering/rendering_runtime.py
```diff
@@ -0,0 +1,244 @@
...
+    """Resolve a physics backend label without task or Hydra presets."""
+    if physics_backend == "physx":
+        from isaaclab_physx.physics import PhysxCfg
+
+        return PhysxCfg(enable_enhanced_determinism=True, enable_external_forces_every_iteration=True)
+    if physics_backend == "ovphysx":
+        from isaaclab_ovphysx.physics import OvPhysxCfg
+
+        return OvPhysxCfg(enable_enhanced_determinism=True, enable_external_forces_every_iteration=True)
+    if physics_backend == "newton":
```
rilei-nvidia: It doesn't make a difference since each conditional returns, but i'd make this an elif anyway for best practices.

source/_isaaclab_testing/test/rendering/test_rendering_kit.py
```diff
@@ -0,0 +1,17 @@
...
+from isaaclab.app import AppLauncher
+
+simulation_app = AppLauncher(headless=True, enable_cameras=True).app
+
+import pytest  # noqa: E402
+from rendering_runner import make_kit_test  # noqa: E402
+
+pytestmark = [pytest.mark.isaacsim_ci, pytest.mark.cold_cache]
+
+test_rendering = make_kit_test()
```
rilei-nvidia: I'd re-name make_kit_test() to something like generate_kit_test_cases() for clarity that it is not a single test

source/_isaaclab_testing/test/rendering/test_rendering_kitless.py
```diff
@@ -0,0 +1,13 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""All golden scenes rendered through Kit-less renderer/backend pairs."""
+
+import pytest
+from kitless_rendering_runner import make_kitless_test
+
+pytestmark = [pytest.mark.isaacsim_ci, pytest.mark.arm_ci, pytest.mark.cold_cache]
+
+test_rendering = make_kitless_test()
```
rilei-nvidia: I'd re-name make_kitless_test() to something like generate_kitless_test_cases() for clarity that it is not a single test

ooctipus:
Good catch. `standard` was only spelling “no special render variant,” so it carried no useful information. I removed it in e356f0812f6, superseding my earlier explanation of the `canonical`/`probe` labels.

Case IDs are now scene-first:

- `canonical-physx-newton_warp-standard` → `rendering_scene-physx-newton_warp`
- `probe-physx-isaac_rtx-shadow_hand-standard` → `shadow_hand-physx-isaac_rtx`
- single-AOV cases append the typed AOV, while only a real visual variant such as `yellow-background` adds a variant suffix

The `suite` and `profile` abstractions are gone; the specialized matrix is described as scene coverage. I also removed `standard` from all 96 affected golden filenames as byte-identical renames. `legacy`/`ovstage` remain only on Kit-less pytest IDs to identify which OVRTX USD-reader path is exercised; both paths still resolve to the same stage-independent golden.

CI selectors and process partitions were updated, and a derived architecture gate now checks that IDs are unique and that these rejected labels cannot return. Validation: 13 targeted tests passed, all 63 Kit-less IDs collect, all 18 exact post-merge selectors resolve, pre-commit passes, and changelog coverage passes. The PR description and preview-image paths are updated too.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1,1910 +0,0 @@
...
-
-    assert info is not None, "camera.data.info is None; renderer did not populate segmentation metadata."
-    assert info["semantic_segmentation"]["idToLabels"] == expected, (
-        f"semantic_segmentation idToLabels mismatch.\n"
-        f"  expected: {expected}\n"
-        f"  actual:   {info['semantic_segmentation']}"
-    )
-
-
-def maybe_validate_instance_segmentation(
```
rilei-nvidia, line 1193: @ooctipus I see that you ported over validation of `semantic_segmentation` it would be great to continue to also validate `instance_segmentation`

source/_isaaclab_testing/test/rendering/rendering_runner.py
```diff
@@ -0,0 +1,145 @@
...
+    RenderBufferKind.DISTANCE_TO_CAMERA,
+    RenderBufferKind.DISTANCE_TO_IMAGE_PLANE,
+    RenderBufferKind.INSTANCE_SEGMENTATION,
+    RenderBufferKind.INSTANCE_ID_SEGMENTATION_FAST,
+}
+_SEMANTIC_ONLY_AOVS = {RenderBufferKind.MOTION_VECTORS}
+_ALPHA_ONLY_AOVS = {RenderBufferKind.INSTANCE_SEGMENTATION, RenderBufferKind.INSTANCE_ID_SEGMENTATION_FAST}
+
+
+def run_rendering_case(
```
hujc7, line 37: **Jichuan:** Personal preference rather than a request — I'd like the harness to have a model that groups things into namespaces instead of free functions in a module. A file gives a location but not a scope: nothing declares what belongs together or what each part owns. As a general preference: where two or more functions only make sense together, or one exists only to serve another, I'd rather they lived on a type than sat side by side in a module. Genuine adapters with no shared context are fine as free functions — `frame_image` and `camera_output_image` read well exactly as they are. The goal is maintainability as this grows.

**Agent suggestion:** this function is the clearest place to see it. It reads as the orchestrator, but only about eight of its sixty-seven lines are sequencing; the rest is calculation belonging to the things being sequenced:

- twelve lines of tensor math for the motion-vector check — `amax`, the 0.99 support-pixel ratio, the half-view bound
- four threshold rules resolved at the call site: `scene.image_tolerance(...)`, the `newton_warp` ternary, `_NO_SSIM`, and `_ALPHA_ONLY_AOVS` now or-ed with an OVRTX/`semantic_segmentation` special case
- a three-line inline join to build `artifact_label`
- nine arguments handed to `build_rendering_scene` that the case and scene spec already know
- two namespace keywords (`golden_namespace`, `artifact_namespace`) threaded through 14 sites across three files, where one field on the case would do — and would also let the two test factories become one

`case.golden_filename(aov, ...)` is the counter-example and shows the direction already works — that identity used to be assembled here and now belongs to the case. A thin orchestrator would read as the sequence and nothing else:

```python
def execute(self, request):
    with self.case.build() as runtime:                   # 9 args -> the case already knows them
        runtime.stabilize_camera()
        outputs, info = runtime.camera_outputs()
        if self.case.needs_motion_step:                  # stepping is sequence, stays here
            outputs |= self._step_once_and_recapture(runtime)
        self._assert_segmentation(outputs, info)
        for aov in self.case.aovs:
            self.case.judgement(aov).assert_ok(outputs[aov], request)
```

On the motion math specifically: `_SEMANTIC_ONLY_AOVS` already declares that motion vectors are judged semantically rather than against a baseline — it just doesn't carry the check, which sits twenty lines above behind a separate `if`. Putting the two together makes both verdict kinds reachable the same way, and a second semantically-checked AOV becomes a table row instead of another branch here. The four threshold rules would sit naturally on `RenderCase`, which already holds the scene, renderer and AOV they depend on, and `compare_to_golden` with its four private helpers on the `ImageComparison` they already construct.

A sample from the current design, not a prescription. The data side has landed (`RenderCase`, `SceneProbe`, `RenderingSceneSpec`, and the gate even asserts this runner does not hardcode scene names); it is the behavior that has no owner yet — 25 module-level functions against 11 classes. Flow-wise the change is a no-op: same operations in the same order. It buys legibility, so it is worth it only if this keeps growing — adding one probe scene currently costs about seven edits across five files. Fine to note for later or skip.

hujc7 (commented):
**Agent review:** nothing blocking from me this round — the restructure landed and I re-checked it after the package move. One preference noted inline, freely skippable.

ooctipus:
Correction after checking both the exact develop images and develop's motion-vector behavior:

**Motion vectors**

Develop does not omit motion-vector goldens globally. It retains them for deterministic PhysX/OVPhysX cases and enables the corresponding determinism flags; only Newton motion-vector tests are skipped because their per-step magnitudes are not deterministic enough. My earlier wording generalized the Newton exception too far.

[`f91b828c3ab`](https://github.com/isaac-sim/IsaacLab/commit/f91b828c3abb3feafecd7f366ab297500a61ce24) restores that contract without restoring task duplication:

- every backend still steps physics exactly once and must produce finite, non-zero, localized motion;
- PhysX/Isaac RTX and OVPhysX/OVRTX now compare the post-step output against two shared motion goldens;
- Newton remains behavior-only instead of being skipped;
- legacy and ovstage pass against the same OVPhysX/OVRTX motion golden;
- `RenderCase.golden_aovs` owns this backend policy, and the architecture gate rejects another global omission.

Validation: PhysX/Isaac RTX 1/1, legacy+ovstage OVPhysX/OVRTX 2/2, architecture/golden inventory 13/13, full pre-commit, and changelog coverage passed.

**Franka RGB comparison**

I downloaded and compared the exact PNGs rather than inferring from configuration:

- Develop's Franka Cloth OVRTX image is already grayscale: dark cloth, gray ground, and no pink/yellow PreviewSurface rendering. Develop's Isaac RTX image is pink/yellow. Franka Soft shows the same backend distinction.
- The PR's `e356f` images did contain one real mismatch: direct scenes used task-training spacing instead of the former golden harness's `3.0`, allowing neighboring table markers into some views. `704c4a46f22` fixed that and regenerated the affected RGB references.
- The current direct OVRTX scene includes the table geometry because it is authored before OVRTX captures the stage, but its PreviewSurface colors remain gray exactly as on develop. Camera, pose, geometry, and authored material values match the current task.

The “same image” requirement applies to the deprecated legacy OVRTX reader versus the ovstage OVRTX reader. Those paths now share `newton-ovrtx-rgb.png`. It does not imply pixel identity between OVRTX and Isaac RTX, which are different renderers and already render these materials differently on develop. Cross-renderer PreviewSurface parity would be a separate renderer issue rather than a regression from this test rearchitecture.

AntoineRichard (commented):
Would it make sense to rename this from `isaaclab_rendering_tests` to `_isaaclab_tests` or something of the like?

ooctipus:
Renamed the downstream project to the final requested name in a27cd508eb8:

- Private checkout directory: `source/_isaaclab_testing`
- Empty distribution metadata: `isaaclab-testing`
- Rendering remains an isolated suite under `test/rendering`, leaving room for future sibling test suites
- Rendering CI filters, exact post-merge node IDs, process isolation, CODEOWNERS, and the changelog fragment now use the new path
- The architecture gate enforces the suite boundary and rejects production-package back-references

Final-name validation passed: package build, 13 architecture/golden tests, 18 Kit and 63 kitless cases collected, full pre-commit, and changelog coverage. Representative real Kit and legacy OVRTX renders passed immediately before this final path-only rename. All 114 PNGs are exact 100% moves; this rename regenerated no goldens.

AntoineRichard (approved):

ooctipus:
Addressed these review notes in `2054709a99a`:

- changed the backend selection branches to `elif`;
- renamed the generated test factories to `generate_kit_test_cases()` and `generate_kitless_test_cases()`;
- restored instance-segmentation metadata validation in the shared runner.

The instance PNGs were already golden-compared, but the former `idToLabels` / `idToSemantics` checks had indeed been lost during the move. The shared validator now checks that both metadata maps match the rendered color set, validates BACKGROUND/UNLABELLED, and compares every rendered prim path with its scene-owned semantic class. This keeps expectations with `RenderingSceneSpec` rather than putting scene names into generic runner logic.

Validation: 15 focused helper/architecture tests, full pre-commit, and real instance-segmentation renders through Isaac RTX, Newton-Warp, legacy OVRTX, and ovstage OVRTX. The Kuka heterogeneous and Shadow Hand real-render cases also passed. No golden images changed in this follow-up.

ooctipus:
Rebased onto current `develop` and force-pushed with lease at `103f77a0586`. GitHub now reports the PR mergeable.

The two failing jobs came from the restored instance-segmentation metadata validator being stricter than Newton-Warp's actual contract:

- `kuka_heterogeneous-physx-newton_warp`: Newton-Warp's scene-static map contained colors for compiled shapes not present in the rendered pixels.
- both legacy Shadow Hand + Newton-Warp cases: legacy cloning exposed only env_0 prim paths in metadata, while the scene declaration correctly lists all four environments.

Isaac RTX and OVRTX still require exact pixel/color/path/class metadata. Newton-Warp now requires both maps to agree, every rendered color and reserved entry to be present, and every reported prim/class pair to be a non-empty subset of the scene declaration. This retains useful validation without asserting completeness that the renderer does not promise.

The rebase also brought in four newly added registered-task Shadow goldens after this PR had removed their owning test. The architecture gate caught that duplicate ownership, so those four were removed; the downstream Shadow scene already owns the replacement coverage.

Validated locally on the rebased tree:

- focused helper/architecture tests: 17 passed;
- exact former CI failures: Kuka Newton-Warp, legacy Shadow OVPhysX/Newton-Warp, and legacy Shadow Newton/Newton-Warp all passed;
- strict contract representatives: shared full-AOV Isaac RTX, Shadow Hand Isaac RTX, and legacy OVRTX instance segmentation passed;
- full pre-commit, changelog coverage, conflict/file ownership audit, and range-diff audit passed.

No downstream golden image was regenerated. I am leaving the newly triggered CI run unmonitored.

ooctipus:
Rebased again onto develop at 7ae2312610e and force-pushed d1d6f3041b8. The conflict came from #6932 regenerating 97 task-owned rendering goldens that this PR removes; I kept develop's synchronous texture-streaming setting and resolved those files to the downstream ownership model (deleted from isaaclab_tasks, retained under _isaaclab_testing). Range-diff preserves all 22 patches; only deletion-side LFS hashes changed. Validation: 17 focused architecture/helper tests passed, shared full-AOV + Franka Cloth Isaac RTX renders passed unchanged, full pre-commit passed, and changelog coverage passed. GitHub now reports MERGEABLE. I am not monitoring the newly triggered CI runs.

ooctipus:
Rebased onto develop at 2ca3b886b6e and force-pushed 54152a9178b. The conflict was .github/workflows/build.yaml: develop replaced the Kit-less matrix with literal legacy/ovstage jobs to restore required check names, while this PR rewired and consolidated the rendering suite. The resolution preserves both static jobs and routes each to _isaaclab_testing/test/rendering/test_rendering_kitless.py with its matching selector. I also fixed the non-conflicting ovstage filter that would otherwise have collected zero tests. Validation: 17 focused architecture/helper tests passed, both static jobs passed structural YAML routing checks, full pre-commit passed, changelog coverage passed, and the branch is 0 commits behind. GitHub reports MERGEABLE. I am not monitoring the newly triggered CI runs.

ooctipus:
Rebased again onto develop `22526de4ca2` and force-pushed `370322aa7a1`. The conflicts were modify/delete overlaps in former task/visualizer Shadow Hand files; I preserved this PR’s downstream `_isaaclab_testing` ownership. Develop also unified the Shadow Hand object across physics backends, so the independently declared rendering scene now uses the same full-size rigid DexCube at `(0.0, -0.39, 0.6)` for every backend, without importing task code; the 11 affected Newton baselines were regenerated. Local validation: helper/architecture 17/17, affected Kit renders 2/2, affected legacy Kit-less renders 4/4, pre-commit/formatting passed, and changelog coverage passed. GitHub reports the PR mergeable; required checks have restarted.

ooctipus:
Rebased again onto develop `a3b6411e138` and force-pushed `47a5b097963`. The conflicts were all modify/delete overlaps in the former task-owned rendering suite. I preserved `_isaaclab_testing` ownership, migrated its OVPhysX dependency/import to develop’s merged `isaaclab_ov` package, added a negative architecture gate for the retired package, and regenerated the two deterministic motion references for develop’s new absolute encoding. Validation: helper/architecture 17/17, PhysX/Isaac RTX motion bundle 1/1, legacy+ovstage OVPhysX/OVRTX motion 2/2 against one shared golden, frozen dependency sync passed, and the full pre-commit/changelog/LFS suite passed against refreshed upstream develop. GitHub reports the PR mergeable; required checks have restarted.

ooctipus:
Fixed the latest legacy correctness failure in 44b35c09490. The only failure was Franka Soft + Newton physics + Newton-Warp RGB: 0.81% differing pixels versus the blanket 0.75% limit, with SSIM 0.9965. The uploaded diff was confined to deformable-surface edges; four earlier/local samples measured 0.23-0.41%. The global Newton-Warp limit remains 0.75%; Franka Soft RGB now owns a narrow 1.00% override, the generic runner consumes the scene-resolved value, and an architecture gate enforces that ownership. Replayed the exact failed artifact successfully, reran the GPU case (passed), focused helper/architecture tests (17 passed), and full pre-commit (passed).

.github/test-subsets/postmerge-rendering.toml
```diff
@@ -5,38 +5,29 @@
 
 # Stable rendering correctness tests used by post-merge CI.
 rendering-correctness = [
-  "source/isaaclab_tasks/test/core/test_rendering_cartpole.py::test_rendering_cartpole[newton-isaacsim_rtx-albedo]",
```
pbarejko, line 8: Why did we remove cartpole? There were some rendering issues i.e. scene partitioning issues that came up with cart pole. Its proven its value. I guess you created a scene with cartpole + random objects  on the table to test the asset?

.github/test-subsets/postmerge-rendering.toml
```diff
@@ -5,38 +5,29 @@
...
-  "source/isaaclab_tasks/test/core/test_rendering_registered_tasks.py::test_rendering_registered_tasks[Isaac-Cartpole-Camera-Direct-None-cartpole]",
-  "source/isaaclab_tasks/test/core/test_rendering_registered_tasks.py::test_rendering_registered_tasks[Isaac-Cartpole-Camera-Direct-rgb-cartpole]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_architecture.py",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[rendering_scene-physx-isaac_rtx]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[rendering_scene-newton-isaac_rtx]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[rendering_scene-physx-newton_warp]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[shadow_hand-physx-isaac_rtx]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[kuka_heterogeneous-physx-isaac_rtx]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[franka_cloth-newton-isaac_rtx]",
+  "source/_isaaclab_testing/test/rendering/test_rendering_kit.py::test_rendering[franka_soft-newton-newton_warp-rgb]",
```
pbarejko: What is so special about `rgb` case. Previously we explicitly listed AOVs and now we have it listed here only.

source/_isaaclab_testing/test/rendering/golden_images/renderers/kuka_heterogeneous/kit-physx-isaac_rtx-instance_segmentation.png
pbarejko, line 1: I think we should have same resolution for IsaacRTX and OvRTX, when an image is so small it's really hard to tell what is correct and what is incorrect. OvRTX resolution is 262x262 and IsaacRTX is 134x134.

source/_isaaclab_testing/test/rendering/golden_images/visualizers/newton-kit-viewport.png
pbarejko, line 1: Is this correct? It seems like we are looking from underneath the table? Is this intentional?

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-newton-isaac_rtx-albedo.png
pbarejko, line 1: I see you separated tiled rendering vs aov rendering. This is a nice change. Can we have aov rendering higher resolution as well?

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-newton-isaac_rtx-rgb.png
pbarejko, line 1: Can we add texture asset to the table plane? In the past we had experienced texture loading multiple times. Having assets with textures helps intercept those issues.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-physx-isaac_rtx-motion_vectors.png
pbarejko, line 1: I wonder why pole is in different position here in comparison to other images from physx-isaac-rtx?

Is this intentional? I see in some golden images pole is leaning to the left and on other it is straight up.

source/_isaaclab_testing/test/rendering/golden_images/renderers/kuka_heterogeneous/kit-physx-isaac_rtx-rgb.png
pbarejko, line 1: The ground asset used to cause a lot of problems with texture loading. It would be great if we use it for the avos rendering scene.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/ovphysx-ovrtx-albedo.png
pbarejko, line 1: We could add some verity to the objects in the scene.

source/_isaaclab_testing/test/rendering/golden_images/renderers/shadow_hand/ovphysx-newton_warp-instance_segmentation.png
pbarejko, line 1: The is an issue with position of the hand, in some renders cube is at the tips and for some closer to the middle. Just pointing it out here that this is a known issue. Internally logged as 6593722.

source/isaaclab/changelog.d/rendering-scene-reset.rst
```diff
@@ -0,0 +1,17 @@
...
+^^^^^^^
+
+* Reworked rendering correctness tests around direct task-free scenes that locally preserve task
+  assets, cameras, layouts, and default state while sharing one runner and bundled camera outputs.
+
+* Extended :func:`isaaclab.envs.mdp.reset_scene_to_default` with an optional fixed-root
+  preservation list and reused that MDP policy from direct rendering scenes.
+
+* Rendering probes resolve task HDR skies locally and reset RTX temporal accumulation between
+  scenes, avoiding asynchronous texture uploads and cross-test renderer history.
```
pbarejko: Is this really true? I thought we have a separate PR for synchronous texture loading, it's a setting for RTX and OVRTX

source/isaaclab_tasks/test/golden_stages/cartpole/newton-isaacsim_rtx_renderer-rgb.usda
pbarejko, line 1: @huidongc what did we do with those stages? Is it safe to just remove them?

source/isaaclab_visualizers/test/golden_images/cartpole/.gitkeep
pbarejko, line 1: do we want to keep this directory?

ooctipus:
Rebased onto current develop and pushed the consolidated review cleanup in 5f8bdd15a31. The final structure keeps two renderer collection roots plus one visualizer root in the private downstream _isaaclab_testing project; direct scenes remain task-free and task-faithful. Cartpole coverage is restored, AOV/Kuka images are larger, the shared scene now exercises textured assets and varied geometry, the visualizer camera is above the table, legacy and ovstage share exact goldens, and Kitless USD materials preserve the Franka colors. Motion advances exactly once, instance metadata is validated, cold-cache selection is marker-driven, and partitions are exhaustively derived. I also aligned Franka Cloth with the latest develop density/friction retune and removed the stale texture-loading/changelog claims. Validation: full Kit 20/20, visualizers 8/8, architecture/golden/Kitless-material 19/19, Kit materials 6/6, shared legacy+ovstage OVRTX golden checks passed, and the real ovstage instance-segmentation check passed. Pre-commit and changelog validation pass.

.github/test-subsets/postmerge-rendering.toml
ooctipus: Good call — the cartpole inside the shared full-AOV scene is not a substitute for the partition-sensitive task composition. I restored a dedicated `cartpole-physx-isaac_rtx` post-merge case. Its direct scene locally mirrors the task asset, four-environment layout, grid ground, light, camera, and default joint state without importing or constructing the task. The case bundles RGB, albedo, depth, and semantic segmentation.

.github/test-subsets/postmerge-rendering.toml
ooctipus: Nothing is special about RGB here; that selector was an artifact of representing a bundled camera capture as a single-AOV case. Fixed: post-merge now selects the bundled specialized cases (for example, `franka_soft-newton-isaac_rtx`) and validates every compatible AOV declared for the scene. An AOV suffix remains only when a case genuinely contains one AOV, chiefly the isolated Kitless OVRTX cases.

source/_isaaclab_testing/test/rendering/golden_images/renderers/kuka_heterogeneous/kit-physx-isaac_rtx-instance_segmentation.png
ooctipus: Agreed. I raised the Kuka per-tile camera to 128×128, so the 2×2 tiled references are now 262×262, matching the reviewable OVRTX scale you called out. Isaac RTX and Newton-Warp use the same camera configuration, and the affected goldens were regenerated.

source/_isaaclab_testing/test/rendering/golden_images/visualizers/newton-kit-viewport.png
ooctipus: It was not intentional. I corrected the shared visualizer view to look from above the table, regenerated all eight visualizer references, and ran the full 8/8 visualizer matrix.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-newton-isaac_rtx-albedo.png
ooctipus: Done. The shared full-AOV camera is now 256×256 for all renderer backends. The AOV references were regenerated at that resolution and remain driven by one shared camera specification.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-newton-isaac_rtx-rgb.png
ooctipus: Done. The shared scene now uses the authored `Props/Blocks/DexCube/dex_cube_instanceable.usd` asset, scaled as the table, instead of a procedural table plane; its asset material is left intact. Together with the standard grid-ground USD, this keeps asset/material/texture loading in the rendering contract.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/kit-physx-isaac_rtx-motion_vectors.png
ooctipus: Yes, this difference is intentional, and the sequence is now explicit in the runner. It snapshots all ordinary AOVs at the reset pose, asserts the physics-step count is zero, advances physics exactly once, renders again, and replaces only the motion-vector output; it then asserts the count is one. The motion reference therefore represents the second-frame transforms while RGB and the other AOVs remain first-frame captures. The dedicated Cartpole probe separately remains at the task's default joint pose.

source/_isaaclab_testing/test/rendering/golden_images/renderers/kuka_heterogeneous/kit-physx-isaac_rtx-rgb.png
ooctipus: Done. The Kuka scene now uses `GroundPlaneCfg`, which loads Isaac Sim's standard `Environments/Grid/default_environment.usd` asset. The white tint preserves the visible grid while exercising the asset and texture-loading path that previously caused regressions.

source/_isaaclab_testing/test/rendering/golden_images/renderers/rendering_scene/ovphysx-ovrtx-albedo.png
ooctipus: Done. The shared AOV scene now contains a cartpole articulation, a moving blue cuboid, a scaled USD table, a red cylinder, and a green sphere, with non-overlapping poses and distinct semantic labels. That gives albedo, normals, depth, and segmentation materially different geometry to validate.

source/_isaaclab_testing/test/rendering/golden_images/renderers/shadow_hand/ovphysx-newton_warp-instance_segmentation.png
ooctipus: Thanks for flagging 6593722. I deliberately did not hide it with a test-only pose correction. Each direct Shadow Hand scene uses the current task's backend-specific robot configuration, camera/orientation, and full-size DexCube at `(0.0, -0.39, 0.6)`; backend-specific goldens keep the known proximity difference visible while still catching further regressions.

source/isaaclab/changelog.d/rendering-scene-reset.rst
ooctipus: You're right; this PR does not own synchronous texture loading. I removed that claim from the changelog and narrowed the camera helper documentation. The harness only resets RTX temporal accumulation and primes renderer history; synchronous texture-loading behavior remains owned by the separate setting/change.

source/isaaclab_tasks/test/golden_stages/cartpole/newton-isaacsim_rtx_renderer-rgb.usda
ooctipus: I audited the references before removing them. Those `.usda` files were owned only by the retired `maybe_save_stage_golden` helper/test; no remaining runtime or test reads them. The new composition root constructs the direct scene and compares renderer outputs, so keeping stage goldens would create a second, stale scene owner. The helper and stages were removed together, and an architecture gate now rejects reintroducing the old `golden_stages` directory.

source/isaaclab_visualizers/test/golden_images/cartpole/.gitkeep
ooctipus: No — I removed the `.gitkeep` and the old visualizer golden directories. Visualizer references now live exclusively under `_isaaclab_testing/test/rendering/golden_images/visualizers`, and the architecture gate rejects the former ownership paths.

ooctipus:
@pbarejko Follow-up on the shared-scene image review:

- The scaled instanceable DexCube was the cause of the table-size disagreement across readers. It is now an explicit-size procedural table, while the standard grid ground remains the textured-asset probe.
- The cartpole now starts at its true default joint state, so RGB, albedo, depth, and segmentation all show the same pre-step pole pose. Only motion vectors take one physics step.
- Semantic validation now enforces a minimum visible table area, so a tiny-table regression fails even if its label is still present.
- The legacy/ovstage albedo discrepancy was a texture-streaming race, not a stable reader difference: the identical legacy case flipped between textured and placeholder output. Texture streaming is now disabled before renderer construction, and legacy plus ovstage RGB/albedo pairs pass the same shared goldens. There is no extra legacy baseline.

Latest validation: shared Kit scene 8/8, legacy RGB/albedo 2/2, ovstage RGB/albedo 2/2, architecture/golden regressions 18/18, and full pre-commit passed. The PR screenshots are pinned to 64b9763004e.

ooctipus:
Pushed a follow-up for the latest `rendering-correctness` failure and rebased onto current `develop` (`ed1f83d939c`).

I downloaded and compared the failing job's artifacts. Cartpole, Kuka, and all four failed Kit visualizer captures have the same camera, pose, geometry, and texture content as their goldens; the failures are low-amplitude RTX shading/color variation across GPU families. For example, using the existing per-pixel L2 threshold of 10 produced 14.10% Cartpole RGB and 17.97% Kuka RGB differences, while an L2 color-noise floor of 20 reduces those to 6.78% and 4.38% respectively. SSIM remains 0.9611 and 0.9765, so the structural signal is still strong.

The fix is deliberately narrow:

- disable texture streaming only for OVRTX, which lacks the Kit USD-context streaming wait;
- use one shared L2=20 color-noise floor only for Isaac RTX RGB/albedo and Kit visualizer output;
- retain the scene-owned spatial-difference budgets and SSIM checks, including a Cartpole-specific 8% / 0.95 gate;
- add unit coverage for small-versus-large color deltas and an architecture gate keeping the streaming workaround OVRTX-owned;
- do not regenerate any golden image.

Post-rebase local validation used the worktree's `_isaac_sim` symlink and `uv run --extra ovrtx`:

- Kit rendering matrix: 20/20 across the four CI process partitions;
- visualizer matrix: 8/8;
- comparison + architecture regressions: 19/19;
- full pre-commit suite: passed.

ooctipus:
Fixed the failed `isaaclab (core) [2/3]` job in `e33e3849f4f`.

The direct MDL authoring code previously resolved built-in modules only from the pip layout (`<isaacsim package>/kit/mdl/...`). CI uses the archive layout, where `python.sh` exposes the Kit root through `CARB_APP_PATH`, so `OmniPBR.mdl` remained a relative unresolved asset.

The material boundary now checks the archive Kit root first and retains the pip fallback. Added a regression that forces the archive layout.

Validation:
- kitless visual-material suite: 3/3;
- actual archive-root resolution: passed;
- Franka Cloth Isaac RTX golden: passed;
- Franka Cloth legacy OVRTX RGB golden: passed;
- full pre-commit: passed.

PR #6981, opened 2026-08-07
OMPE-103001: Render and animate Newton cables on all three renderers

jmart-nv:
# Description

- Newton cables (`UsdGeom.BasisCurves`) now **render and animate** on `newton_warp`, `ovrtx` and `isaac_rtx`. Previously they drew at their spawn pose and never moved.
- Cable curve points are computed on device from the Newton segment bodies and written **GPU/ASYNC**, zero-copy, once per render frame.
- Drops the CPU mirror in `sync_cables_to_usd` — it copied the whole model's `body_q` device→host on every dirty render frame, for a cable that reads only its own segments.
- Adds the first rendering coverage for cables: 9 cells across the three renderers, gated on **motion**, not just presence.
- Fixes three defects found alongside: `view_count`/camera-prim mismatch, the Isaac RTX sensor pump refreshing transforms only, and the Fabric stage being resolved exactly once at startup.

Same scene, same seed, three renderers: cables fall and settle. A cable that isn't driven renders frozen at its spawn pose — it does not disappear — which is why the tests gate on centroid displacement.

**Requires:**

Cable rendering depends on renderer-side fixes that are **not** in the pinned Kit/OVRTX:

| | needed |
|---|---|
| Kit | `!47946`, `!48359`, `!48511` |
| OVRTX | `>= 0.4.1` |

The rendering cells **skip** below those floors rather than fail, so a green run on stock deps proves nothing. Check for `0 skipped`.

**Changes:**

| package | what |
|---|---|
| `isaaclab` | `CameraRenderSpec` rejects a `view_count` that disagrees with `camera_prim_paths`; `Camera` sizes it from the prims it found. Shared rendering-test harness in `isaaclab.test.utils.cable_rendering`. |
| `isaaclab_newton` | `sync_cables_to_usd` runs on device and invalidates each curve's GPU data afterwards. New `collect_cable_segment_shapes()` for kit-less renderers. Lazy Fabric-stage recovery. |
| `isaaclab_ov` | New OVRTX cable point binding (legacy zero-copy + ovstage). |
| `isaaclab_physx` | Sensor pump uses `pre_render()` instead of `forward()`; empty-annotator-buffer guard. |
| `scripts/demos/cables.py` | `--camera`, `--seed`, `--frames_dir`. Camera pose baked into the cfg; cables moved under `/World/envs/env_0/`. |

**Testing:**

Three scenarios, each run on all three renderers — 9 cells:

| test | asserts |
|---|---|
| `test_cable_renders` | a cable produces geometry at all |
| `test_cable_render_tracks_simulation` | the rendered image follows the cable as it falls |
| `test_cable_renders_across_environments` | both of the above, replicated across 4 envs |

```bash
export OMNI_KIT_ACCEPT_EULA=yes
python -m pytest source/isaaclab_ov/test/ -q
python -m pytest source/isaaclab/test/renderers/test_cable_rendering.py -q -ra
python -m pytest source/isaaclab_physx/test/renderers/test_cable_rendering_isaac_rtx.py -q -ra
python scripts/demos/cables.py --visualizer none --camera isaac_rtx --seed 7 --max_steps 60
```

| check | expected |
|---|---|
| `isaaclab_ov` contract | 112 passed |
| kit-less cells | 6 passed, **0 skipped** |
| `isaac_rtx` cells | 3 passed |
| demo | exit 0, 60 frames |

Confirmed on four machines (1 Windows/RTX, 3 Linux/L40).

The `isaac_rtx` cells live in their own module because Kit must start before anything imports USD — **do not merge them into the kit-less module**, and do not run both in one pytest process.

**Handoff notes for reviewers:**

- **`isaac_rtx_renderer.py` empty-buffer guard**: never fires in any cell, ships without a test, and its `getattr(buf, "size", None) == 0` check cannot fire for a torch tensor (`.size` is a method there). The condition it was written for is now rejected earlier by `CameraRenderSpec.__post_init__`. Suggest dropping it from this PR and filing separately.
- **ovstage cable path** hands over via host copy: ovstage 0.1.0 accepts the `points` lanes=3 dtype override only on numpy, not on DLPack producers. The legacy path — the GA one — stays zero-copy.
- `collect_cable_segment_shapes` duplicates the discovery half of `_initialize_fabric_cable_prims`. Deliberate: the Kit path **raises** on a malformed cable, the kit-less path skips it. Merging them would change one of those behaviours.
- The demo's cable prims moved from `/World/Env_0/` to `/World/envs/env_0/` — required for scene partitions, but it is a user-visible path change.

**Appendix - required Kit MRs:**

All three are merged to `release/ovrtx-0.4-ovstage-0.1` and labelled `ovrtx-0.4.1` — which is where the OVRTX ≥ 0.4.1 floor comes from.

- **[!47946](https://gitlab-master.nvidia.com/omniverse/kit/-/merge_requests/47946)** — *Allow Points and Curves to be created from GPU interop points data.* Without it a GPU-authored curve resolves zero points and logs `Not enough points, N is required but 0 was found`. Merged 2026-08-01.
- **[!48359](https://gitlab-master.nvidia.com/omniverse/kit/-/merge_requests/48359)** — *Fix for curves empty bounding boxes in scenePartition.* A `BasisCurves` in a partition published an inverted empty bbox, blanking **every** prim in that partition. Merged 2026-08-03.
- **[!48511](https://gitlab-master.nvidia.com/omniverse/kit/-/merge_requests/48511)** — *Update geometry group BB when ProceduralGeometry is updated.* Animated point writes kept the creation-time bbox, so a falling cable was culled once it moved about its own thickness. Merged 2026-08-06.

## Type of change

## Screenshots

cables_all_renderers

## Checklist

### DIFF
--- a/scripts/demos/cables.py
+++ b/scripts/demos/cables.py
@@ -16,12 +16,17 @@
     # Usage without a visualizer and with a larger cable pile.
     uv run python scripts/demos/cables.py --visualizer none --num_cables 40 --num_segments 15
 
+    # Usage with a rendering camera, to observe cable rendering on a given renderer.
+    uv run python scripts/demos/cables.py --visualizer none --camera ovrtx
+    uv run python scripts/demos/cables.py --visualizer none --camera newton_warp
+
 """
 
 from __future__ import annotations
 
 import argparse
 import math
+import os
 import random
 
 from isaaclab.app import add_launcher_args, launch_simulation
@@ -30,6 +35,25 @@ parser = argparse.ArgumentParser(description="Spawn a pile of cables with Newton
 parser.add_argument("--num_cables", type=int, default=25, help="Number of cables to spawn.")
 parser.add_argument("--num_segments", type=int, default=20, help="Number of segments per cable.")
 parser.add_argument("--max_steps", type=int, default=-1, help="Stop after this many steps; negative runs forever.")
+parser.add_argument(
+    "--seed",
+    type=int,
+    default=None,
+    help="Seed the cable spawn RNG so two runs are comparable frame-for-frame.",
+)
+parser.add_argument(
+    "--camera",
+    nargs="?",
+    const="ovrtx",
+    choices=["ovrtx", "newton_warp", "isaac_rtx"],
+    default=None,
+    help="Attach a camera that renders the cable pile with the given renderer.",
+)
+parser.add_argument(
+    "--frames_dir",
+    default=None,
+    help="Write each rendered frame as a PNG into this directory, for assembly into an animation.",
+)
 parser.add_argument("--physics", default="newton_vbd", choices=["newton_vbd"], help="Physics backend.")
 add_launcher_args(parser)
 parser.set_defaults(visualizer=["kit"])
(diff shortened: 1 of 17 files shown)


huidongc:
The work is being continued on #7016

jmart-nv:
Closing - superseded by #7016

PR #6768, opened 2026-07-28
Re-generate lock, bump pins, fix platforms

StafaH:
# Description

Bump the newton pin. Regenerate the lock and fix previous issues and pins from the pyproject and lock.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #6986, opened 2026-08-08
Fix usd-core marker to resolve correctly on aarch64 with uv

jpsousaai96-ux:
## Summary

- `SUPPORTED_ARCHS` in `source/isaaclab/setup.py` gates the `usd-core`
  dependency behind a list-membership PEP 508 marker
  (`platform_machine in 'x86_64,AMD64'`). This is semantically correct —
  `usd-core` has no aarch64 wheels on PyPI, and IsaacLab already uses
  `usd-exchange` instead on ARM via the separate `SUPPORTED_ARCHS_ARM`
  marker.
- However, `uv`'s resolver does not prove this marker disjoint from an
  aarch64-only resolution fork, so `uv sync` on aarch64 (e.g. NVIDIA DGX
  Spark) fails with a spurious "no aarch64-compatible wheels for
  usd-core" error, even though the marker should exclude `usd-core`
  entirely on that platform.
- Rewriting the marker with explicit equality/`or` instead of list
  membership is semantically identical but resolves correctly under
  `uv`.

## Test plan

- [x] Verified via `packaging.markers.Marker(...).evaluate(...)` that
      both the old and new marker forms evaluate identically for
      `x86_64`, `AMD64`, and `aarch64`.
- [x] From `isaac-sim/IsaacLab-Arena` (`isaaclab-from-source` flavor,
      which installs this package editable from the submodule), `uv
      sync` fails with this error before the fix and succeeds after it,
      on aarch64 Linux (DGX Spark, GB10).
- [ ] Not independently tested against pip/conda installs — the marker
      is still valid PEP 508 syntax so behavior there is expected to be
      unaffected.


AntoineRichard:
Thank you for taking the time to investigate this issue and submit the fix, Joao! The clear aarch64 reproduction and test notes are especially appreciated.

The underlying issue has already been addressed on `develop` by #6768. Dependencies have since moved from `source/isaaclab/setup.py` to the root `pyproject.toml`, where `usd-core` now uses explicit equality/or platform markers and `usd-exchange` is restricted to `aarch64`. The current `uv.lock` also confirms that the two packages resolve into separate architecture-specific forks.

Because this change no longer applies to `develop`, I am closing the PR as already resolved. Thank you again for contributing to Isaac Lab!

PR #6997, opened 2026-08-10
Read OVRTX render vars without a GPU-side wait

pv-nvidia:
## Description

Reading an OVRTX render var normally orders the read against render completion with a GPU-side wait.
On Linux that hits a driver GPU-context scheduling issue where most of the graphics work runs while
the render-output copy waits, costing roughly **20% of end-to-end throughput**
(`Isaac-Lift-KukaAllegro-Camera`, 64 envs: 1621 → 2133 total FPS).

So on Linux the render var is now mapped with no GPU-side barrier and the calling thread blocks on
the render-completion event instead. Other platforms keep the GPU-side wait, and Linux uses it too
when `ISAAC_LAB_DISABLE_LINUX_CUDA_CPU_SYNC`  disables the driver behaviour.

All mapping sites now go through one helper, `OVRTXRenderer._map_render_var_to_dlpack()`.

Camera outputs are unchanged — only the read/render ordering differs. 

Reported in https://nvbugspro.nvidia.com/bug/6566453

## Type of change

- Performance improvement

## Screenshots

None — no visual change.

## Checklist

### DIFF
--- a/source/isaaclab_visualizers/changelog.d/maxkra15-fix-newton-rtx-ui-callbacks.rst
+++ b/source/isaaclab_visualizers/changelog.d/maxkra15-fix-newton-rtx-ui-callbacks.rst
@@ -1,0 +1,4 @@
+Fixed
+^^^^^
+
+* Fixed duplicate Simulation panels in the Newton RTX viewer.
--- a/source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
+++ b/source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
@@ -425,13 +425,6 @@ class _NewtonViewerUIMixin:
                 " training\nhigher values -> less responsive visualizer but faster training"
             )
 
-    def _render_physics_panel(self, imgui):
-        """Render Simulation collapsing section at the top of the Newton viewer panel."""
-        imgui.set_next_item_open(True, imgui.Cond_.appearing)
-        if imgui.collapsing_header("Simulation"):
-            imgui.separator()
-            imgui.text(f"Physics: {self._backend_display}")
-
     def _draw_streaming_view_controls(self) -> None:
         """Render streaming image panel selector in the HUD sidebar.
 
@@ -573,10 +566,9 @@ class NewtonViewerRTX(_NewtonViewerUIMixin, ViewerRTX):
         # UI patches must be deferred: ViewerRTX creates self.gui lazily in
         # _init_window() (called from _init_ovrtx() on the first end_frame()).
         # _patch_viewer_panel() sets gui._render_left_panel, which requires gui to
-        # exist.  Register the callbacks now (they are buffered by ViewerRTX until
+        # exist.  Register the training controls now (they are buffered by ViewerRTX until
         # the GUI is available); the panel patch is applied in _init_window() below.
         self.register_ui_callback(self._render_training_controls, position="side")
-        self.register_ui_callback(self._render_physics_panel, position="panel")
 
     def _init_window(self) -> None:
         """Create the viewer window and immediately apply Isaac Lab UI patches."""


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -126,6 +127,17 @@
...
 
+# Runtime environment variables controlling how a CUDA render-var mapping is synchronized against the
+# render that produced it: the first picks the GPU-side barrier, the second the host-side wait.
+# See :meth:`OVRTXRenderer._map_render_var_to_dlpack`.
+#
+# ``"0"`` is OVRTX's "no sync" sentinel, not the NULL CUDA stream: the field encodes
+# ``0=no sync, 1=default stream, >1=specific stream``, so omitting it means ``1``, not ``0``.
+_RENDER_VAR_SYNC_STREAM_ENV = "ISAAC_LAB_OVRTX_RENDER_VAR_SYNC_STREAM"
+_RENDER_VAR_SYNC_STREAMS = ("0", "default", "warp")
+_RENDER_VAR_MAP_WAIT_ENV = "ISAAC_LAB_OVRTX_RENDER_VAR_MAP_WAIT"
```
rilei-nvidia: Hey @pv-nvidia - are both the sync stream and render var map wait exposed because you want to experiment with them at runtime?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1086,6 +1122,56 @@ def _generate_random_colors_from_ids(self, input_ids: wp.array, output_colors: w
...
+          is the barrier the OVRTX API is designed around and the correct GPU-side ordering here.
+
+        With a GPU-side barrier, :data:`_RENDER_VAR_MAP_WAIT_ENV` (default ``1``) selects whether the
+        host additionally blocks on the render-completion event. With ``"0"`` the host wait is the
+        only ordering available, so it is always applied and that variable is ignored.
+
+        The default (``"0"`` plus the host wait) is the faster ordering on every platform. A GPU-side
+        barrier followed by OVRTX's render-output copy hits a Linux driver GPU-context scheduling issue
+        in which most of the graphics work runs while the copy waits, costing around 20% of end-to-end
+        throughput; blocking the host instead sidesteps that, and is also marginally faster on Windows.
```
rilei-nvidia: Shall we tag the NVBUG # here?  6566453

rilei-nvidia (approved):
Looks good to me.  Had a question of whether we need to expose those env vars or not.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Good point, these hardcore technical flags are complicated, you would need a good reason for exposing them.

In case you do really want to expose these env vars, "0" and "default" is not very informative regarding what it means. It seems to be the three modes are:
- NO_SYNC (this is the default according to `_render_var_sync_stream_setting`)
- CUDA_DEFAULT_STREAM (this is not the default value though, so maybe just CUDA_STREAM)
- WARP_STREAM

huidongc: Also I keep thinking why is this workaround not implemented inside ovrtx? It seems like an issue for all ovrtx clients.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1086,6 +1122,56 @@ def _generate_random_colors_from_ids(self, input_ids: wp.array, output_colors: w
...
+
+        :data:`_RENDER_VAR_SYNC_STREAM_ENV` picks the GPU-side barrier OVRTX inserts:
+
+        - ``"0"`` (default): none. This is OVRTX's "no sync" sentinel, *not* the NULL CUDA stream --
+          it encodes the field as ``0=no sync, 1=default stream, >1=specific stream``, so omitting
+          the argument entirely means ``1``, not ``0``.
+        - ``"default"``: a ``cudaStreamWaitEvent`` on the CUDA default stream, which is what OVRTX
+          does when the argument is omitted. Our consuming kernels run on the Warp stream rather than
+          the default stream, so this only orders them via the default stream's implicit
+          synchronization, not by OVRTX's documented contract.
```
huidongc: "0" and "default" are separated values, and "0" is the default...

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: Sorry guys, this was meant to be a draft PR, I must have clicked the wrong button. Making it draft again.

I added these env vars to benchmark all possible combinations.

But good point @huidongc this should be inside ovrtx...

huidongc: No worry, Peter.

PR #6999, opened 2026-08-10
Add lightweight performance smoke comparisons

AntoineRichard:
# Description

This is a focused alternative to #6864. It uses IsaacLab's existing
`benchmark runtime` CLI to run the performance workloads and adds only the
missing comparison and CI wiring.

The PR:

- runs eight supported benchmark configurations against the exact PR CI image;
- compares total FPS, total startup time, peak GPU memory, and peak
  benchmark-process RSS;
- keeps the established arithmetic mean of per-step FPS by default, while
  exposing a single opt-in `--aggregate_throughput` flag;
- requires an FPS regression to cross both its percentage threshold and a
  two-standard-error significance threshold computed from coherent per-step
  FPS statistics;
- downloads a read-only authoritative baseline table from the repository
  variable `PERF_BASELINE_S3_URI`;
- publishes a Markdown workflow summary, a machine-readable JSON comparison,
  and the raw benchmark bundle;
- matches exact provider-qualified Isaac Sim or kit-less Newton runtime
  identities and fails closed on malformed, missing, ambiguous, or non-finite
  input; and
- adds no dependency, baseline-writing path, reseeding workflow, custom GitHub
  status, or parallel benchmark implementation.

Until the S3 bucket is provisioned, leaving `PERF_BASELINE_S3_URI` unset makes
the comparison explicitly report `SKIP`; benchmarks and artifacts are still
produced. Once configured, the workflow only requires `s3:GetObject` access.
Every matching baseline row must provide all four metrics. The FPS row also
stores the reference per-step FPS standard deviation and sample count so the
comparison can reject noisy percentage regressions that are not statistically
significant.

The implementation is 799 additions across eight files, including exactly six
focused tests. This is approximately 92% smaller than the current proposal
while retaining the gating and reporting behavior needed for CI.

Local validation:

- focused CPU-only checks covered both CLI modes, typed-request forwarding,
  and all five comparator scenarios;
- `./isaaclab.sh -f` passed immediately before commit and again immediately
  before push;
- real downloaded Newton CI bundles for Cartpole and G1 produce complete
  four-metric `SKIP` reports while S3 is unconfigured, including startup,
  memory, and measured per-step FPS deviation; and
- the earlier GPU smoke exercised the real runtime bundle and comparison
  outputs. No GPU benchmark was rerun for these reporting-only trims because
  the shared machine is currently busy.

## Type of change

## Screenshots

Not applicable; this change affects CI reports and artifacts.

## Checklist

- [x] Public API and CLI documentation are updated in the runtime request docstring and CLI help
- [x] I added the required `source/isaaclab/changelog.d/` fragment

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -914,16 +914,19 @@ jobs:
           -e ISAAC_SIM_HEADLESS=1 \
           -v "$PWD/benchmark-output:/benchmark-output" \
           --entrypoint bash "$CI_IMAGE_TAG" -lc \
-          "./isaaclab.sh benchmark runtime \
+          "set -euo pipefail
+          mkdir -p /tmp/benchmark-output
+          ./isaaclab.sh benchmark runtime \
             --task '$BENCHMARK_TASK' \
             --num_envs '$NUM_ENVS' \
             --num_steps 200 \
             --warmup_steps 100 \
             --seed 42 \
             --benchmark_formatter schema \
-            --output_path /benchmark-output \
+            --output_path /tmp/benchmark-output \
             --visualizer none \
-            $BENCHMARK_ARGS"
+            $BENCHMARK_ARGS
+          cp /tmp/benchmark-output/*.json /benchmark-output/"
     - name: Download authoritative baseline
       if: vars.PERF_BASELINE_S3_URI != ''
       env:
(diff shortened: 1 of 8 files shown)


AntoineRichard:
Closing in favor of https://github.com/isaac-sim/IsaacLab/pull/7137

PR #7004, opened 2026-08-10
Warn when XR mode silently falls back to CPU physics

maxwbuckley:
# Description

Passing `--xr` without `--device` overrides the usual `cuda:0` default and runs physics on the CPU:

https://github.com/isaac-sim/IsaacLab/blob/develop/source/isaaclab/isaaclab/app/app_launcher.py#L1073-L1079

```python
device_explicitly_passed = launcher_args.pop("device_explicit", False)
if self._xr and not device_explicitly_passed:
    # If no device is specified, default to the CPU device if we are running in XR
    device = "cpu"
    launcher_args["device"] = "cpu"
```

The fallback is deliberate — CPU physics can reduce latency when only a single environment is
present, which is the common teleoperation case. The problem is that it is **silent**. It emits no
console output, and on `develop` the CloudXR how-to does not mention the XR device default at all,
so the only ways to discover it are reading `_resolve_device_settings` or noticing `cpu` in the
environment banner.

On scenes where the GPU pipeline is the faster choice this is an easy, invisible performance loss.
I hit it running `Isaac-Stack-Cube-Franka-IK-Abs-v0` teleoperation on an RTX 5090 and spent a while
chasing latency before finding the fallback in the launcher source.

This PR does not change the default — only makes it discoverable:

- `AppLauncher._resolve_device_settings` logs a warning when the fallback is applied, naming both
  the behavior and the override.
- The `--device` help text mentions the XR fallback.
- The CloudXR teleoperation how-to documents it, with the override command and how to confirm the
  selection in the startup banner.

The warning text deliberately avoids claiming GPU is always faster, since which device wins is
scene-dependent:

```
XR mode is enabled and no device was specified explicitly: running physics on the CPU. Outside XR
the default is 'cuda:0'. CPU physics can reduce latency for single-environment teleoperation, but
is typically slower for larger or contact-rich scenes. Pass --device explicitly (for example
'--device cuda:0') to override this fallback.
```

Fixes # (no issue filed — happy to open one if preferred)

## Type of change

- Documentation update

## Screenshots

N/A — console warning text is quoted above.

## Checklist

### Notes on the two unchecked boxes

**Tests.** There is no existing test module for `AppLauncher`, so there was no harness to extend and
adding one felt out of scope for a warning. I verified the behavior by loading `app_launcher.py`
with the `isaaclab` package stubbed and calling `_resolve_device_settings` directly:

| case | resolved device | warnings |
| --- | --- | --- |
| `--xr`, no `--device` | `cpu` | 1 |
| `--xr --device cuda:0` | `cuda:0` | 0 |
| no `--xr` | `cuda:0` | 0 |

Glad to add a proper test module if you would like one.

**CONTRIBUTORS.md.** Not added yet — let me know if you would like me to add an entry.

### DIFF
--- a/docs/source/how-to/cloudxr_teleoperation.rst
+++ b/docs/source/how-to/cloudxr_teleoperation.rst
@@ -257,6 +257,32 @@ To verify that the headset and controller tracking poses are reaching Isaac Lab,
 hand joints and RGB axes at tracked controller aim poses. See
 :ref:`isaac-teleop-tracking-debug-visualization` for details.
 
+.. note::
+
+   **In XR mode, physics runs on the CPU unless you pass** ``--device`` **explicitly.** Passing
+   ``--xr`` without ``--device`` overrides the usual ``cuda:0`` default and selects ``cpu``
+   instead. Isaac Lab warns when this happens:
+
+   .. code-block:: text
+
+      XR mode is enabled and no device was specified explicitly: running physics on the CPU.
+
+   CPU physics can reduce latency when only a single environment is present, which is why it is
+   the default for teleoperation. It is typically slower for larger or contact-rich scenes, where
+   the GPU pipeline wins instead. Which one is faster is scene-dependent, so measure both. To
+   force GPU physics, pass the device explicitly:
+
+   .. code-block:: bash
+
+      uv run --extra teleop isaaclab teleop run \
+          --task <task> \
+          --visualizer kit \
+          --xr \
+          --device cuda:0
+
+   Confirm the selection in the startup banner, which prints
+   ``Environment device    : cuda:0`` (or ``cpu``).
+
 .. attention::
 
    **First run — EULA acceptance required.**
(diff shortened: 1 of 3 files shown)


maxwbuckley:
Thanks — both findings were correct, fixed in ea6f7154a.

**Warning could misreport the device under `--distributed`.** Confirmed: the warning sat inside the XR fallback branch, while the `distributed` branch further down rewrites `device` to `cuda:<rank>`. An `--xr --distributed` run without `--device` would have reported CPU physics while resolving to CUDA. The fallback is now tracked in a local flag and the warning is emitted after final resolution, gated on `device == "cpu"`.

Verified by calling `_resolve_device_settings` directly (`isaaclab` package stubbed, plus a `torch` stub so the distributed branch is reachable):

| case | resolved device | warnings |
| --- | --- | --- |
| `--xr`, no `--device` | `cpu` | 1 |
| `--xr --device cuda:0` | `cuda:0` | 0 |
| no `--xr` | `cuda:0` | 0 |
| `--xr --distributed`, no `--device` | `cuda:0` | 0 |

The last row is the regression — it warned before this fix.

**Changelog category and wording.** Moved both entries under `Added` to match their wording, and dropped the `AppLauncher._resolve_device_settings` reference from the user-facing note.

`./isaaclab.sh --format` passes.

AntoineRichard (approved):

PR #7009, opened 2026-08-10
Restore Newton coverage in rigid-object-collection interface tests

hujc7:
# Description

`_rigid_object_collection_iface_test_utils.py` imports `WrenchComposer` from
`isaaclab_newton.test.fixtures.mock_newton`, which defines only `MockNewtonModel` and
`create_mock_newton_manager`. The import raises `ImportError`, the surrounding
`except ImportError` swallows it, and `"newton"` is never appended to `BACKENDS`.

Because the suite parametrizes over that filtered list, an empty list becomes skipped
parameters rather than a failure — so the Newton collection tests stopped running while the
suite stayed green. From `isaaclab (core) [3/3]` logs, same job, same container:

| Suite | physx | newton |
|---|---|---|
| `test_articulation_iface.py` | 1151 | 915 |
| `test_rigid_object_iface.py` | 384 | 398 |
| `test_rigid_object_collection_iface.py` | 657 | **0** (was 659) |

## Changes

- Use the `WrenchComposer` already imported at module scope. It is the class production
  Newton constructs (`isaaclab_newton/.../rigid_object_collection.py:1256`), so no
  backend-specific alias is warranted.
- Record the cause in `BACKEND_UNAVAILABLE_REASONS` in the two utilities that discarded it,
  matching `_articulation_iface_test_utils.py`.
- Add `test_iface_utilities_agree_on_backend_availability`: the three utilities gate physx
  and newton on the same package, so a backend registered by some and not others is a broken
  import and now fails the build.

## Validation

- `test_rigid_object_collection_iface.py -k "newton and cpu"` → 309 passed, 12 xfailed (was 651 skipped)
- Four interface suites + guard → 1887 passed, 0 failed
- Guard verified both ways: fails with the import reinstated, naming the module and the
  `ImportError`; passes once fixed
- `uv run isaaclab -f` clean

## Checklist

- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/`

### DIFF
--- a/source/isaaclab/test/assets/test_iface_test_utils.py
+++ b/source/isaaclab/test/assets/test_iface_test_utils.py
@@ -72,6 +72,38 @@ assert all(not any(backend.lower() == "physx" for backend in module.BACKENDS) fo
     assert result.returncode == 0, result.stdout + result.stderr
 
 
+@pytest.mark.parametrize("backend", ["physx", "newton"])
+def test_iface_utilities_agree_on_backend_availability(backend: str) -> None:
+    """Fail when one utility drops a backend that its siblings kept.
+
+    All three utilities gate ``physx`` and ``newton`` on the same backend package being
+    importable, so a backend registered by some and not others means that utility's own imports
+    are broken rather than the backend being unavailable. ``ovphysx`` is excluded because the
+    collection utility additionally gates it on ``_create_buffers``.
+    """
+    script = f"""
+import importlib
+import sys
+
+sys.path.insert(0, {_ASSET_TEST_DIR.as_posix()!r})
+backend = {backend!r}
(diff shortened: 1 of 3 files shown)


source/isaaclab/test/assets/test_iface_test_utils.py
AntoineRichard, line 1: I don't think this test is needed?

AntoineRichard (approved):
LGTM otherwise

source/isaaclab/test/assets/test_iface_test_utils.py
hujc7: This is to ensure that an unavailable backend is not due to import error which could silently drop tests

PR #7016, opened 2026-08-11
OVRTX rendering support for cables

huidongc:
# Description

Adds rendering support for Newton cable objects with OVRTX.

- Adds render-agnostic cable segment discovery and ordered shape-ID mapping.
- Adds OVRTX curve bindings and point updates for both legacy and OVStage paths.
- Introduces camera rendering coverage and golden images for franka lift cable env.

Fixes OMPE-103001

## Type of change

## Screenshots

| OVRTX | IsaacSim RTX | Newton Warp |
| ------ | ----- |  ----- |
| franka_cable-newton-ovrtx_renderer-rgb | franka_cable-newton-isaacsim_rtx_renderer-rgb | franka_cable-newton-newton_renderer-rgb  |

## Checklist


huidongc:
@greptile-apps

huidongc:
@isaaclab-review-bot

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -216,6 +216,20 @@ class _ParticleVisualPrim:
     frames_since_sync: int
 
 
+@dataclass
+class CableRegistryEntry:
```
mmichelis: Is this registry abstraction still needed? `collect_cable_segment_shapes()` already returns the complete `curve path -> ordered shape IDs` mapping. At the current head, `_discover_cables_into_registry()` wraps every mapping in singleton instance lists and `_initialize_fabric_cable_prims()` immediately unwraps them; `curve_prim_path` and `segments_per_cable` are never read, while both OVRTX paths consume `collect_cable_segment_shapes()` directly. Could we remove `CableRegistryEntry`, `_cable_registry`, and `_discover_cables_into_registry()`, then have Fabric iterate `collect_cable_segment_shapes().items()` as well? This appears to be leftover scaffolding from the asset-side registration design removed in `7856688110`.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -725,45 +743,68 @@ def sync_transforms_to_usd(cls) -> None:
...
+            ],
+            device=PhysicsManager._device,
+        )
+        # The points were written in place on the device. A render delegate caches its curve
+        # geometry and cannot notice that, so without an explicit invalidation it keeps drawing
+        # the points it last read and the cable renders frozen.
+        for cable_path in cls._cable_prim_paths:
+            points_attr = cls._usdrt_stage.GetPrimAtPath(cable_path).GetAttribute("points")
+            if points_attr:
+                points_attr.InvalidateGpuData()
```
mmichelis: Could we remove this per-prim invalidation loop? `SelectPrims` already requests GPU `ReadWrite` access for `points`; creating a writable `RtPrimSelection` marks every selected attribute dirty in Fabric, and `PrepareForReuse()` is the bulk mechanism if the selection is retained across frames. `InvalidateGpuData()` instead obtains a CPU write handle and marks the GPU copy invalid, which reverses authority immediately after the GPU kernel wrote these points. With this loop removed, `_cable_prim_paths` also has no remaining consumer. If RTX still misses the curve update, that needs the appropriate renderer/change-tracking or bounds notification rather than invalidating the freshly written GPU data.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -710,6 +711,66 @@ def _save_comparison_image(img: Image.Image, filename: str) -> str:
...
+    tensor = output if isinstance(output, torch.Tensor) else output.torch
+    condition = torch.logical_or(torch.isinf(tensor), torch.isnan(tensor))
+    corrected = torch.where(condition, torch.zeros_like(tensor), tensor)
+    normalized = normalize_camera_output_for_display(corrected, data_type)
+    grid = make_camera_output_grid(normalized)
+    ndarr = grid.mul(255).add_(0.5).clamp_(0, 255).permute(1, 2, 0).to("cpu", torch.uint8).numpy()
+    return Image.fromarray(ndarr).convert("RGB")
+
+
+def save_rendering_gif(
```
mmichelis, line 709: Are these new gif rendering functions needed? Why were they not used before?

mmichelis (commented):
Thanks for the PR! I've added some comments already mainly on the additions in the newton manager, would be great if newton manager can be as minimal as possible.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
huidongc: Good call out!

huidongc: Fixed with f226e03c

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
huidongc: If I remove the per-prim invalidation loop, then the curves will disappear.

NVBUG#6502662 is still not fixed?

I am scoping this GPU interop out of this PR. We can improve this together with particles (OMPE-106051).

source/isaaclab_tasks/test/rendering_test_utils.py
huidongc: They are opt-in utilities to create gif for local manual validation. I didn't upload them and kept them local. I think it might be useful for other devs.

huidongc:
> would be great if newton manager can be as minimal as possible.

Yeah, good point. I pushed a few other commits to make it a minimal change. The GPU interop for Kit RTX will need to be tackled separately.

huidongc:
@greptile-apps

huidongc:
@greptile-apps

huidongc:
@isaaclab-review-bot Review

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
huidongc: GPU interop for Kit RTX is scoped out.

huidongc:
@greptile-apps
@isaaclab-review-bot Review

huidongc:
> get_current_stage(fabric=True)

@greptile-apps Good point but this is pre-existing code not related to this pr, better not to scope it in this PR.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -845,58 +905,59 @@ def _update_transforms_legacy(self) -> None:
...
-                self._apply_particle_workaround(particle_q)
-                self._particle_workaround_applied = True
-            else:
-                self._write_particle_q_slices(
-                    self._particle_points_binding,
-                    particle_q,
-                    self._particle_visual_offsets,
-                    self._particle_visual_counts,
-                )
+            self._write_particle_q_slices(
```
mmichelis, line 982: This disconnects the existing first-frame MPM workaround: `_apply_particle_workaround()` and `_particle_workaround_applied` remain, but this branch now always performs the GPU ASYNC write. The focused suite reports 3 failures out of 18, including both tests that require the first update to use host SYNC before later GPU ASYNC updates. Could you double check this PR does not change existing particle behavior?

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -1891,3 +1955,88 @@ def rendering_test_franka_soft(
...
+        if gif_steps is not None:
+            frames: list[Image.Image] = []
+            for _ in range(gif_steps):
+                env.step(zero_actions)
+                frames.append(_camera_outputs_to_pil_image(env.scene.sensors["base_camera"].data.output))
+            save_rendering_gif(frames, test_name, physics_backend, renderer, data_type)
+            return
+
+        # Let the cable settle under gravity so golden frames are not first-frame spawn poses.
+        env.step(zero_actions)
```
mmichelis, line 2008: A single post-step golden only validates one rendered pose; it does not verify that the curve continues to follow the simulated cable instead of remaining at its spawn geometry. Could this assert a visible change between frames captured before and after multiple simulation steps, so the test fails when cable points are not updated over time?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1420,6 +1492,90 @@ def _setup_particle_bindings(self) -> None:
         else:
             self._setup_particle_bindings_legacy()
 
+    def _setup_cable_bindings(self) -> None:
```
mmichelis, line 1568: This dispatcher is redundant: its only production caller is legacy initialization, where `_use_ovstage` is necessarily false, while ovstage initialization calls `_setup_cable_bindings_ovstage()` directly. Please call `_setup_cable_bindings_legacy()` directly and remove this wrapper, or consistently route both initialization paths through it. The current half-used dispatch adds branching without behavior.

mmichelis:
Please update the title and description to match the final scope. At the current head, Newton Warp already renders Newton geometry directly, and Kit RTX retains the existing CPU Fabric cable sync and CPU mirror because GPU interop was explicitly scoped out. The novel runtime feature here is live Newton cable synchronization for the legacy OVRTX and ovstage paths, plus shared discovery and renderer coverage. A title such as **Render Newton cables with OVRTX renderers** would be more precise. The description currently says cable points are sent through GPU/ASYNC each frame and that the CPU mirror is dropped, which contradicts the final Kit implementation. Please describe the renderer-specific paths and remove those outdated claims.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1420,6 +1492,90 @@ def _setup_particle_bindings(self) -> None:
...
+        self._cable_counts = wp.array(counts, dtype=wp.int32, device=device)
+        self._cable_segment_counts = counts
+        self._cable_max_points = max(counts) + 1 if counts else 0
+        self._cable_point_offsets = []
+        total_points = 0
+        for count in counts:
+            self._cable_point_offsets.append(total_points)
+            total_points += count + 1
+        self._cable_points = wp.zeros(total_points, dtype=wp.vec3f, device=device)
+        self._cable_point_offsets_wp = wp.array(self._cable_point_offsets, dtype=wp.int32, device=device)
```
mmichelis: Optional simplification: could we avoid storing a second set of offsets here? Since `offsets[i]` is the prefix sum of segment counts and each preceding curve contributes one additional point, the point start is always `offsets[i] + i`. The kernel and host slices could derive this, removing `_cable_point_offsets`, `_cable_point_offsets_wp`, and one kernel argument.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Once we merge PR #7108 , I will rebase this branch. The failure will be gone because they have been fixed by OVRTX 0.4.1.

source/isaaclab_tasks/test/rendering_test_utils.py
huidongc: Yeah, this is the bare minimum. The cons to add an additional frame is that the rendering test will take longer. 

Octi has a PR that refactors the rendering test and gets obvious cut in the test runtime, that PR is targeting `develop`. I think we can do something over there. We can leave this as is for the release branch.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: I just mirrored it from the existing code. We will do cleanup once ovstage becomes the only code path.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Good point.

source/isaaclab_ov/test/test_ovrtx_deformable_bindings.py
```diff
@@ -479,3 +486,123 @@ class _FakeStream:
...
+    assert backend.calls[0]["shape"] == (3,)
+    assert backend.calls[0]["flags"] is BindingFlag.OPTIMIZE
+    assert renderer._cable_points_binding is backend.bindings["points"]
+
+    # World-space points are written directly, so the inherited env transform must be neutralised
+    # or it is applied twice -- the same contract the deformable path relies on.
+    assert [write["attribute_name"] for write in backend.writes] == ["omni:resetXformStack", "omni:xform"]
+
+
+def test_setup_cable_bindings_reads_discovery_without_clone_plan(monkeypatch: pytest.MonkeyPatch):
```
mmichelis: Non-blocking: could we simplify the cable test block? This case only verifies that `_clone_plan` is unused, while the private-offset test below overlaps the runtime write test. I suggest keeping the binding and no-cable cases, folding unequal cable lengths into one `update_geometries()` write test, and asserting the resulting slices plus ASYNC stream behavior there. That preserves the observable contracts with less overlap and avoids coupling coverage to `_cable_point_offsets`.

mmichelis (approved):
Thanks for your hard work, and updating the PR description and simplifying the code! It looks good to me for cable rendering support!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
huidongc: Fixed with 4b1aebb

ooctipus (approved):

source/isaaclab_tasks/test/golden_images/franka_cable/newton-isaacsim_rtx_renderer-distance_to_image_plane.png
pbarejko, line 1: What are those white spots on the image? They appear on other images too, but why they are white on distance to image plane? This seems like target that we render?

pbarejko (approved):

source/isaaclab_tasks/test/golden_images/franka_cable/newton-isaacsim_rtx_renderer-distance_to_image_plane.png
huidongc: Yes, the Franka cable environment creates command visual markers:

- CommandsCfg.cable_pose.debug_vis=True enables the goal and current cable-segment markers.
- The cable command’s debug callback renders them at the sampled target and tracked middle segment.
- The rendering helper does not disable cable_pose.debug_vis; unlike the soft/cloth helper, it only applies the generic camera overrides.

StafaH (approved):

kellyguo11 (approved):

PR #7020, opened 2026-08-11
Add animated Isaac Lab 3 greetings to the loading screen

AntoineRichard:
# Description

This PR adds a lightweight, static greeting beside the loading-screen summary.

- Retains the static Isaac Lab 3 and NVIDIA terminal artwork, selected once at random for each run.
- Removes animated greetings, animation containers, environment overrides, packaging, authoring/viewer tooling, and their dedicated tests.
- Removes the “The way it’s meant to be trained” tagline.
- Starts the wide layout at 130 columns so the Isaac Lab 3 wordmark fits.

## Validation

- `uv run isaaclab -f`
- Focused loading-screen tests: 14 passed

## Checklist

- [x] Added a changelog fragment.
- [x] Added focused regression coverage.
- [x] Updated the PR branch against current `develop`.


AntoineRichard:
Thanks both — all four findings were real and are fixed in 5508e4b.

**Pinned greeting ignored on wide terminals** (`isaaclab-review-bot`) — confirmed and the most serious of the four. Reproduced with `ISAACLAB_LOADING_ANIM=hello-small`: 22 columns of greeting at 80 columns, 63 at 140, i.e. the random `isaac-wide` rather than the pinned art. A pin now offers that greeting and nothing else, which also deletes the slot-matching loop — there is nothing to match when there is only one.

Worth calling out that the existing test asserted on `_animations[0]` only, so it passed against the broken behaviour throughout. It now renders the header at both widths and measures what actually landed beside the box; reverting the fix fails it at both.

**Shutdown clock keeps falsy start-time check** (`isaaclab-review-bot`) — correct, and it was the third of three sites. The other two were changed to `is None` earlier and this one was missed, so the changelog claimed a fix that was only two-thirds done. All three now spell the condition identically, and the changelog fragment says so.

**Validate playback frame rates** (`greptile`) — `--fps 0` divided by zero inside the loop. Now an argparse type that reports `must be greater than zero, got 0`.

**Reject negative frame counts** (`greptile`) — confirmed the failure mode exactly as described: `range(-1)` is empty, the container is written with `frames: 0`, and the body then parses as one empty frame, so `unpack` rejects a file that was written without complaint. Guarded in `_resample` so the API path cannot produce one either, with an argparse type for a clean CLI message.

StafaH (changes requested):
I've been trying to keep loading screen very very lightweight. Can we clean up this PR for a more lightweight approach.

- We really don't need tests for loading screen related stuff, maybe 1 simple test but we don't need a whole test suite.
- Reduce files/code down to more lightweight.

StafaH (approved):
I think we can strip out the animated from the PR and just keep the green logo. Let's do IsaacLab 3 when we have space, and IL3 when we have less space.

I think we can bring back the animated stuff, let's make a mini-lib like we said for ASCII visualizer and loading screen, and we'll re-introduce it in 3.1, and add some easter eggs for christmas/halloween

kellyguo11 (commented):
could we quickly test this on windows to make sure it also works there? we had a QA bug with our previous implementation that caused issues on windows terminals.

AntoineRichard:
image
image
Done!

marcodiiga (approved):

PR #7021, opened 2026-08-11
Seal the OVStage write ordinal before attaching

AntoineRichard:
# Description

Seals the OVStage write ordinal before `attach_ovstage()`. **No version pins change in this PR** — it is the standalone fix, so QA can test against it directly.

`ovstage.population.open_usd_from_string()` completes its writes but never commits the ordinal it wrote to. `OvPhysxManager._attach_ovstage()` populated at ordinal 1 and attached at `read_ordinal=1` without ever sealing it.

Newer `ovphysx` releases require `read_ordinal` to be sealed by a completed write-floor advance. From the `attach_ovstage` docstring:

> Reads target sealed data, so attaching at an unsealed ordinal fails the parse and yields an empty scene.

Because it yields an empty scene rather than raising, the failure surfaces far from its cause. Every articulation, rigid body, and sensor binding resolves to zero prims, showing up as either:

```
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'
```

deep in asset initialization, or where a guard exists:

```
RuntimeError: OVPhysX could not create any articulation bindings for pattern '/World/Env_*/Robot'.
```

Both are the same root cause at different call sites.

## The fix

```python
stage.advance_write_floor(ordinal=1).wait()
cls._physx.attach_ovstage(stage, read_ordinal=1)
```

`isaaclab_ov/renderers/ovrtx_renderer.py` already did this before its own attach — the OvPhysX manager was the only site missing it. `advance_write_floor()` exists in the currently pinned ovstage, and on the currently pinned ovphysx the seal is simply redundant, so **this is safe to land ahead of any version bump**.

## Verification

Verified against a newer ovphysx where the bug reproduces. All three train for 5 iterations with no traceback; all three fail at env creation with the `NoneType + int` TypeError when the line is reverted:

| Task | Invocation | Result |
| --- | --- | --- |
| `Isaac-Velocity-Flat-AnymalD` | `presets=ovphysx` | `Training time: 7.19s` |
| `Isaac-Cartpole` | `physics=ovphysx` | `Training time: 2.36s` |
| `Isaac-Velocity-Flat-G1` | `presets=ovphysx` | `Training time: 16.08s` |

Also confirmed AnymalD still trains on the **currently pinned** ovphysx with this change applied, confirming backward compatibility.

`test_ovphysx_scene_data_backend.py` — 34 passed. The manager lifecycle test now asserts the seal lands **between** population and attach, so the ordering is guarded rather than just the call being present.

## Not verified

- The full `isaaclab_ovphysx` unit suite has not been run to completion. These files carry `pytest.mark.device_split` and must be invoked once per device in separate processes.
- **OvRTX is separately broken** with a newer ovstage — `Isaac-Cartpole-Camera` with `presets=ovphysx,ovrtx` fails during population with `write_attributes op failed for [metersPerUnit, kilogramsPerUnit, upAxis]: Stage reader/writer legacy interface is unavailable`. This happens *before* the attach, so it is unrelated to and unaffected by this fix. Whether it also reproduces on the currently pinned versions was not established. Tracked separately from this PR.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_ov/changelog.d/ovphysx-attach-write-floor.rst
+++ b/source/isaaclab_ov/changelog.d/ovphysx-attach-write-floor.rst
@@ -1,0 +1,10 @@
+Fixed
+^^^^^
+
+* Fixed :class:`~isaaclab_ov.physics.OvPhysxManager` attaching its OVStage at an
+  unsealed write ordinal. ``ovstage.population.open_usd_from_string()`` only
+  completes population; it never commits the ordinal it wrote to. Newer
+  ``ovphysx`` releases fail the parse when attaching at an unsealed ordinal and
+  yield an empty scene, so every articulation, rigid body, and sensor binding
+  resolved to zero prims. The manager now calls ``advance_write_floor().wait()``
+  to seal the ordinal before ``attach_ovstage()``.
--- a/source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
+++ b/source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
@@ -602,6 +602,10 @@ class OvPhysxManager(PhysicsManager):
                 # dependencies in physics-only population.
                 domains=ovstage.PopulationDomain.ALL,
             )
+            # ovphysx reads sealed data only: population completes the writes but never
+            # commits the ordinal, so attaching at an unsealed ordinal fails the parse
+            # and silently yields an empty scene.
+            stage.advance_write_floor(ordinal=1).wait()
             cls._physx.attach_ovstage(stage, read_ordinal=1)
         except Exception:
             stage.destroy()
--- a/source/isaaclab_ov/test/physics/test_ovphysx_scene_data_backend.py
+++ b/source/isaaclab_ov/test/physics/test_ovphysx_scene_data_backend.py
@@ -459,9 +459,19 @@ def test_manager_attaches_and_releases_owned_ovstage(monkeypatch):
 
     events = []
 
+    class FakeWriteFloorOp:
+        def __init__(self, ordinal):
+            self._ordinal = ordinal
+
+        def wait(self):
+            events.append(("seal", self._ordinal))
+
     class FakeStage:
         def __init__(self, name):
             events.append(("stage", name))
+
+        def advance_write_floor(self, ordinal):
+            return FakeWriteFloorOp(ordinal)
 
         def destroy(self):
             events.append(("destroy",))
@@ -508,9 +518,12 @@ def test_manager_attaches_and_releases_owned_ovstage(monkeypatch):
         OvPhysxManager._physx = previous_physx
         OvPhysxManager._ovstage = previous_ovstage
 
+    # The seal must land between population and attach: ovphysx reads sealed data
+    # only, so attaching at an unsealed ordinal silently yields an empty scene.
     assert events == [
         ("stage", "isaaclab"),
         ("populate", stage, "#usda 1.0", 1, "all"),
+        ("seal", 1),
         ("attach", stage, 1),
         ("close_views", physx),
         ("reset",),


pyproject.toml
```diff
@@ -147,9 +147,9 @@ rerun = [
 ]
 
 isaacsim = ["isaacsim[all,extscache]==6.0.1.0"]
-ov = ["ovphysx==0.5.9", "ovrtx>=0.4.0,<0.5.0", "ovstage==0.1.0.346039"]
-ovphysx = ["ovphysx==0.5.9", "ovstage==0.1.0.346039"]
-ovrtx = ["ovrtx>=0.4.0,<0.5.0", "ovstage==0.1.0.346039"]
+ov = ["ovphysx==0.5.10", "ovrtx>=0.4.0,<0.5.0", "ovstage==0.1.1.355824"]
```
marcodiiga: Blocking: [`rendering-correctness-kitless (legacy)`](https://github.com/isaac-sim/IsaacLab/actions/runs/31489093723/job/93777717614) is required and fails all 16 non-xfail OVPhysX cases at this head: nine with OVRTX and seven with Newton Warp. The failures occur during `open_usd_from_string()`, before the new floor advance; the reported population error is `write_attributes: Stage reader/writer legacy interface is unavailable`. The [base run of this lane passed with the prior package set](https://github.com/isaac-sim/IsaacLab/actions/runs/31470132905/job/93712576858). Please select a public OV package set that is mutually compatible at runtime, or rescope IsaacLab's extras so they do not select this combination, then rerun this lane.

source/isaaclab/test/cli/test_uv_run_pyproject.py
```diff
@@ -139,7 +139,7 @@ def test_version_single_source_matches_literal_pins():
     optional = pyproject["project"]["optional-dependencies"]
     overrides = pyproject["tool"]["uv"]["override-dependencies"]
 
-    assert versions["ovphysx"] == "0.5.9"
+    assert versions["ovphysx"] == "0.5.10"
```
marcodiiga: The required [changelog check](https://github.com/isaac-sim/IsaacLab/actions/runs/31489093819/job/93771114086) treats this changed test as an `isaaclab` package change and fails because that package has no fragment. Since this is a test-only pin assertion, please add a unique `source/isaaclab/changelog.d/*.skip`.

source/isaaclab_ov/pyproject.toml
```diff
@@ -19,7 +19,7 @@ requires-python = ">=3.12"
 # Only interpackage dependencies should go here.
 dependencies = [
     "isaaclab",
-    "ovstage==0.1.0.346039",
+    "ovstage==0.1.1.355824",
```
marcodiiga: The 0.1.1 wheel reports `LicenseRef-NvidiaProprietary`, while the reviewed exception currently recognizes `NVIDIA Proprietary Software`. Both the [Python dependency license check](https://github.com/isaac-sim/IsaacLab/actions/runs/31489093899/job/93771114446) and [Docker dependency license check](https://github.com/isaac-sim/IsaacLab/actions/runs/31489093899/job/93771114393) flag OVStage, with the Docker report classifying it as having `No reviewed exception`. Please have the new license-metadata identifier reviewed and align the checked-in exception as appropriate for this exact wheel.

marcodiiga (changes requested):
The seal-before-attach change is correct and the ordering test is good, but I can't approve this dependency set while its required legacy kitless lane is regressed.

At this head, every non-xfail OVPhysX case in the required legacy renderer check fails during OVStage population, before the new seal/attach sequence. Please restore a public package/runtime combination compatible with this lane, or rescope the extras so IsaacLab's extras do not select the incompatible combination, then rerun the required check.

The missing `isaaclab` changelog marker and the new OVStage wheel's license metadata also need resolution. The kitless-image 503 appears unrelated.

Non-blocking wording correction: the sealing rule predates this version bump. The floor advance is backward-compatible with 0.5.9, but it is not a no-op because it advances the write floor. Please adjust the PR description accordingly.

The PR description's detailed-contact impact note also appears unsupported for Isaac Lab: the OVPhysX contact sensor does not currently call the detailed-contact APIs affected by that 0.5.10 change. Please remove that impact claim unless it can be demonstrated with a supported Isaac Lab configuration.

kellyguo11 (approved):

marcodiiga (approved):

PR #7023, opened 2026-08-11
Fix kernel-only Isaac Sim installations

mmichelis:
# Description

Fixes NVBug 6587433

A kernel-only `isaacsim` pip installation is importable and has package metadata, but it does not expose `isaacsim.SimulationApp`. Isaac Lab previously reported this incomplete runtime as a missing Isaac Sim installation at launch, while `./isaaclab.sh -i isaacsim` treated it as complete and could not repair the environment.

This change:

- reports that Isaac Sim is installed but its full runtime is unavailable when `SimulationApp` cannot be loaded;
- preserves local binary-install activation guidance and recommends the platform-neutral `uv run isaaclab -i isaacsim` repair command;
- makes the installer verify the `SimulationApp` capability in addition to package metadata; and
- completes kernel-only installations with `isaacsim[all,extscache]` at the installed version, avoiding an unintended version change.

No new dependencies are introduced.

## Type of change

## Screenshots

Not applicable; this change affects CLI diagnostics and installation behavior.

## Validation

- Verified that the installer regression test fails against unmodified `develop` and passes with this fix.
- `uv run --extra test python -m pytest source/isaaclab/test/cli/test_install_commands.py -q`: 60 passed.
- `uv run isaaclab -f`: passed.
- `uv run python tools/changelog/cli.py check develop`: passed.

Coverage includes completing an incomplete Isaac Sim installation at its existing version.

## Checklist

- [x] Documentation changes are not required; CLI guidance and the changelog fragment are updated

### DIFF
--- a/source/isaaclab/test/app/test_launch_simulation_require_kit.py
+++ b/source/isaaclab/test/app/test_launch_simulation_require_kit.py
@@ -94,14 +94,14 @@ def test_kernel_only_isaac_sim_reports_missing_runtime(
     from isaaclab.app import AppLauncher
 
     monkeypatch.setattr(AppLauncher, "is_available", lambda: False)
-    monkeypatch.setattr(sim_launcher.importlib.metadata, "version", lambda _name: "6.1.0rc3")
+    monkeypatch.setattr(sim_launcher.importlib.metadata, "version", lambda _name: "1.2.3")
     monkeypatch.setattr(sim_launcher.sys, "platform", platform)
 
     with caplog.at_level("ERROR"), pytest.raises(SystemExit) as exc_info:
         sim_launcher._ensure_isaac_sim_available()
 
     assert exc_info.value.code == 1
-    assert "Isaac Sim 6.1.0rc3 is installed, but its full runtime is unavailable" in caplog.text
+    assert "Isaac Sim 1.2.3 is installed, but its full runtime is unavailable" in caplog.text
     assert f"{installer} -i isaacsim" in caplog.text
     assert "Isaac Sim is not installed or not found on PYTHONPATH" not in caplog.text
 
--- a/source/isaaclab/test/app/test_launch_simulation_require_kit.py
+++ b/source/isaaclab/test/app/test_launch_simulation_require_kit.py
@@ -16,6 +16,7 @@ taken. No Kit/GPU required.
 """
 
 import argparse
+import importlib.metadata
 
 import pytest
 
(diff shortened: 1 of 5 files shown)


kellyguo11:
when would this become the case? is it actually a bad installation of isaac sim?

source/isaaclab/test/app/test_launch_simulation_require_kit.py
AntoineRichard, line 1: Do we need these tests?

AntoineRichard:
Same question as Kelly! What is this used for?

mmichelis:
> when would this become the case? is it actually a bad installation of isaac sim?

if users install just isaacsim without isaacsim[all,extscache]. Technically if the isaaclab installation is followed this should never happen. The nvbug repro reports this install command:

`uv pip install "isaacsim==6.1.0rc3" --extra-index-url https://pypi.nvidia.com/ --index-strategy unsafe-best-match   # TRIGGER: no [all,extscache] extras => kernel-only install`

We could close the bug and say the documented way of installing should not cause this error. This PR would simply give a more informative error even if the user happens to not follow the installation instructions.

kellyguo11:
> > when would this become the case? is it actually a bad installation of isaac sim?
> 
> if users install just isaacsim without isaacsim[all,extscache]. Technically if the isaaclab installation is followed this should never happen. The nvbug repro reports this install command:
> 
> `uv pip install "isaacsim==6.1.0rc3" --extra-index-url https://pypi.nvidia.com/ --index-strategy unsafe-best-match # TRIGGER: no [all,extscache] extras => kernel-only install`
> 
> We could close the bug and say the documented way of installing should not cause this error. This PR would simply give a more informative error even if the user happens to not follow the installation instructions.

I guess it doesn't hurt to have an extra check, though perhaps we should point the error message to using uv run instead of isaaclab.sh/bat. but in general, yes we would also expect users to follow the installation commands in the docs

AntoineRichard (changes requested):

source/isaaclab/test/app/test_launch_simulation_require_kit.py
mmichelis: removed in [656652d](https://github.com/isaac-sim/IsaacLab/pull/7023/commits/656652ddca42fe14852f0a823ce64060d6f0adbf)

mmichelis:
> > > when would this become the case? is it actually a bad installation of isaac sim?
> > 
> > 
> > if users install just isaacsim without isaacsim[all,extscache]. Technically if the isaaclab installation is followed this should never happen. The nvbug repro reports this install command:
> > `uv pip install "isaacsim==6.1.0rc3" --extra-index-url https://pypi.nvidia.com/ --index-strategy unsafe-best-match # TRIGGER: no [all,extscache] extras => kernel-only install`
> > We could close the bug and say the documented way of installing should not cause this error. This PR would simply give a more informative error even if the user happens to not follow the installation instructions.
> 
> I guess it doesn't hurt to have an extra check, though perhaps we should point the error message to using uv run instead of isaaclab.sh/bat. but in general, yes we would also expect users to follow the installation commands in the docs

Yes sounds good! In 03077d6 I've fixed this!

AntoineRichard (approved):

kellyguo11 (approved):

PR #7024, opened 2026-08-11
docs(odin): update skrl failures handoff with confirmed root causes

AntoineRichard:
## Summary

Updates `docs/odin/HANDOFF-skrl-classic-failures.md` with confirmed root causes for both skrl failure clusters found in dispatch `20260811-124406`.

**Both issues trace to the same gap: Odin dispatch never injects `--agent` or `--algorithm` to select the right skrl config for non-default entries.**

### Issue A (192 rows — non-box space presets)

- **Was**: "model output is not reshaped for non-box action spaces"
- **Is**: Dispatch emits `presets=box_discrete` but no `--agent skrl_box_discrete_cfg_entry_point`. `benchmark_train_skrl.py` falls back to `skrl_cfg_entry_point` → box_box GaussianMixin → continuous `(num_envs, 3)` actions into a `Discrete(3)` env → shape mismatch crash.
- **Fix**: `discover.py` must read `agent_preset_compatibility` from gym kwargs and propagate matching `--agent <ep>` into dispatch row overrides.

### Issue B (9 rows — Humanoid-AMP)

- **Was**: error not captured; possible missing motion file
- **Is**: AMP tasks only register `skrl_amp_cfg_entry_point`. Dispatch defaults to `--algorithm PPO` → tries `skrl_cfg_entry_point` → not found → crash. Motion `.npz` files are bundled in the package.
- **Fix**: Same discover.py gap — detect tasks with no `skrl_cfg_entry_point` and inject `--agent skrl_amp_cfg_entry_point` (or `--algorithm AMP`).

Also adds corrected local reproducers with explicit failing vs. passing vs. workaround variants for both issues.

## Test plan

- [ ] Confirm reproducer A fails without `--agent` and passes with `--agent skrl_box_discrete_cfg_entry_point`
- [ ] Confirm reproducer B fails without `--algorithm AMP` and passes with it
- [ ] Implement unified fix in `discover.py` / `plan.py` and re-dispatch to validate


AntoineRichard:
Closing — accidentally pushed internal Odin documentation to the public repo. This content should not have been committed here.

PR #7026, opened 2026-08-11
Report the resolved backends in the run startup summary

mataylor-nvidia:
# Description

The train and play startup summary was printed before `launch_simulation` scanned the config, so a
run reached through an automatic selector was reported by the family it asked for rather than the
backend it resolves to. `physics=physx` resolves to `PhysxAutoCfg` and `renderer=rtx` to an
`auto_rtx` placeholder; both are resolved inside `scan()`, using the same "does this run need Kit"
decision, well after the panel was drawn.

```
uv run isaaclab train --rl_library rsl_rl --task Isaac-Cartpole-Camera-Direct renderer=rtx physics=ovphysx
```

reported `Renderer rtx` and then launched OVRTX.

`show_run_summary` now scans the config before drawing the panel, so every row names the backend
that will run alongside the choice the run stopped at, following the existing
`default (<resolved>)` form: a backend reached through a family the command line named is reported
as `<family> (<resolved>)`, and one the run named neither directly nor by family stays
`default (<resolved>)`.

Two supporting changes:

* `pre_launch_video_config` moves above the summary in the four train entrypoints (the play
  entrypoints already had this order). It injects a headless Kit visualizer for `--video` runs, so
  scanning before it ran would resolve `auto_rtx` to OVRTX and bake that in. As a side effect the
  summary of a `--video` run no longer reports `Visualizer none (headless)`.
* Requesting the livestream Kit visualizer moves from `launch_simulation` into `scan()`, so a
  caller that scans directly reaches the same launch decision the launcher does.

`launch_simulation` still scans, and does not need a cached or precomputed result: resolving a
placeholder consumes it, so the second walk observes the same signals. `test_scanning_twice_reaches_the_same_launch_decision`
pins that property.

## Type of change

## Screenshots

`--task Isaac-Cartpole-Camera-Direct renderer=rtx physics=ovphysx`, headless:

| Before | After |
| ------ | ----- |
| <pre>│ Physics       ovphysx           │<br>│ Renderer      rtx               │<br>│ Presets       none              │</pre> | <pre>│ Physics       ovphysx           │<br>│ Renderer      rtx (ovrtx)       │<br>│ Presets       none              │</pre> |

The run launches OVRTX in both cases; only the report changes.

## Checklist


source/isaaclab/isaaclab/app/sim_launcher.py
```diff
@@ -257,7 +257,14 @@ def scan(cfg, launcher_args: argparse.Namespace | dict | None = None) -> Scan:
     place). Automatic PhysX configurations and RTX
     renderer placeholders (``renderer_type="auto_rtx"``) are also resolved
     at this stage using the full *launcher_args* context.
+
+    The walk mutates *cfg* in place, and resolving a placeholder consumes it, so
+    a second walk of the same config observes the same signals and reaches the
+    same launch decision.
     """
+    # Livestreaming implies a Kit visualizer; make that visible to auto RTX resolution.
+    _ensure_livestream_kit_visualizer(launcher_args)
```
mataylor-nvidia, line 266: Livestreaming requires a Kit viewport, but nothing in the config says so. `scan` needs to be self sufficient so that when it is called from the show_summary this information is already known

source/isaaclab_rl/test/test_entrypoints_common.py
```diff
@@ -356,3 +356,53 @@ def test_additional_preset_names_lists_presets_without_a_row_of_their_own(
...
+
+    def __init__(self) -> None:
+        self.fields: dict[str, str] = {}
+
+    def summary(self, title: str, fields: dict[str, str]) -> None:
+        self.fields = fields
+
+
+@pytest.mark.parametrize(
+    "selectors, expected_physics, expected_renderer, expected_presets",
```
mataylor-nvidia, line 372: marking which selectors were used should be more descriptive for this test

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -577,9 +595,23 @@ def show_run_summary(
...
+        resolved: Preset name of the backend that will run.
+        selected: Preset names the command line asked for.
+
+    Returns:
+        The resolved name alone when the run asked for that backend or reached it
+        through an automatic selector, and ``default (<resolved>)`` when the task
+        config pinned it.
+    """
+    if requested != resolved:
+        return resolved
```
mataylor-nvidia, line 612: if the name was resolved by auto selection output the name that was resolved

StafaH (approved):
Thanks Matthew for fixing this.

Can we remove the renderer/physics from preset row, I'd prefer not to repeat there.

Either we show the resolved name, or we show the actual name and resolved in brackets (similar to how default is handled).

I think the second option suits us.

Example:
`renderer: rtx (ovrtx)`

because it follows the default logic which currently does:

`renderer: default (newton_renderer)`

mataylor-nvidia:
> Thanks Matthew for fixing this.
> 
> Can we remove the renderer/physics from preset row, I'd prefer not to repeat there.
> 
> Either we show the resolved name, or we show the actual name and resolved in brackets (similar to how default is handled).
> 
> I think the second option suits us.
> 
> Example: `renderer: rtx (ovrtx)`
> 
> because it follows the default logic which currently does:
> 
> `renderer: default (newton_renderer)`

done, thanks this makes sense @StafaH

kellyguo11 (approved):

PR #7032, opened 2026-08-12
Fix duplicate Newton RTX Simulation panel

maxkra15:
# Description

Remove a redundant Newton RTX physics-panel callback that rendered a second `Simulation` header alongside Isaac Lab’s patched viewer panel. Both headers shared the same ImGui ID, producing the visible conflicting-ID warning.

The patched panel remains the single source for the Physics, Up Axis, Gravity, pause, and reset controls. GL behavior is unchanged.

## Type of change

## Validation

- Regression test verified to fail with the redundant panel callback and pass after its removal.
- `uv run --extra test python -m pytest source/isaaclab_visualizers/test/test_newton_adapter.py -q` (47 passed)
- Changelog validation against the exact upstream base
- `uv run isaaclab -f`
- Inspected the live RTX callback registry during initialization

## Checklist

### DIFF
--- a/source/isaaclab_visualizers/test/test_newton_adapter.py
+++ b/source/isaaclab_visualizers/test/test_newton_adapter.py
@@ -22,7 +22,7 @@ from isaaclab_visualizers.newton import (
 )
 from isaaclab_visualizers.newton import newton_visualization_markers as newton_markers
 from isaaclab_visualizers.newton import newton_visualizer as newton_visualizer_module
-from isaaclab_visualizers.newton.newton_visualizer import NewtonViewerGL, _eye_lookat_to_pitch_yaw
+from isaaclab_visualizers.newton.newton_visualizer import NewtonViewerGL, NewtonViewerRTX, _eye_lookat_to_pitch_yaw
 from isaaclab_visualizers.newton_adapter import (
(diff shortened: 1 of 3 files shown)


source/isaaclab_visualizers/test/test_newton_adapter.py
AntoineRichard, line 1: Do we need this test?

AntoineRichard (approved):
LGTM

source/isaaclab_visualizers/test/test_newton_adapter.py
```diff
@@ -512,6 +512,23 @@ def test_newton_visualizer_forwards_and_neutralizes_picking():
     assert visualizer._viewer_picking_binding._retained_picking is None
 
 
+def test_newton_rtx_viewer_does_not_register_redundant_physics_panel(monkeypatch):
```
matthewtrepte: i feel like this test is very specific, likely we can remove it. as long as the UI panel looks correct with the patch.

maxkra15: yeah, agent giberish

kellyguo11 (approved):

PR #7034, opened 2026-08-12
Fix raycast_mesh distance and face-id shapes for rays with any leading dimensions

### LINKED ISSUE #2416: Bug in source/isaaclab/isaaclab/utils/warp/ops.py ray_distance.view(shape[0], shape[1]) causes shape mismatch
I am trying to use raycast_mesh(..., return_distance=True) with the following input:

```
ray_starts = torch.randn(4, 3)
ray_directions = torch.randn(4, 3)
```

This results in:

```
ray_distance = ray_distance.to(device).view(shape[0], shape[1])
RuntimeError: shape '[4, 3]' is invalid for input of size 4
```

It seems ray_distance is initialized with shape (N,) where N = 4, but the code attempts to reshape it to (4, 3) which is incorrect for scalar distances.

Shouldn’t the correct line be:

`ray_distance = ray_distance.to(device).view(shape[0])`

instead of:

`ray_distance = ray_distance.to(device).view(shape[0], shape[1])`

samirbhattarai135:
# Description

`raycast_mesh` flattens the rays with `.view(-1, 3)`, but reshaped the distance and face-id
outputs with `.view(shape[0], shape[1])`. That reshape is only correct when the rays have
exactly two leading dimensions, so every other input shape was mishandled — including the
`(N, 3)` shape the docstring advertises:

| `ray_starts` shape | `ray_distance` before | after |
| --- | --- | --- |
| `(N, 3)` (documented) | `RuntimeError` | `(N,)` |
| `(B, N, 3)` | `(B, N)` | `(B, N)` — unchanged |
| `(3, 1, 2, 3)` | `RuntimeError` | `(3, 1, 2)` |
| `(2, 3, 1, 3)` | `(2, 3)` — **silently wrong shape, no error** | `(2, 3, 1)` |

The silent case is the one worth highlighting: when the leading dimensions happen to multiply
out to `shape[0] * shape[1]`, the old reshape succeeded and returned a tensor whose shape did
not match the rays it came from.

Both reshapes now use `.view(shape[:-1])`, which is the same generic idiom the neighbouring
`ray_normal` line already used. For any 3-D input `shape[:-1]` *is* `(shape[0], shape[1])`, so the
reshape change alone is a strict superset of the current behaviour. The direction-shape guard in the
second commit is not — see below.

`raycast_dynamic_meshes` is deliberately untouched: it documents `(B, N, 3)` explicitly, so
`view(shape[:2])` is correct for its contract.

Fixes #2416

### On the second commit

`fix(warp): require raycast_mesh directions to match ray starts` is adjacent to the issue and can be
dropped if you would rather keep this PR to the reshape alone.

The kernel launches with `dim=num_rays` derived from `ray_starts` and indexes
`ray_directions_wp[tid]` unconditionally. Passing fewer directions than rays therefore reads past the
end of the direction array, and the values it reads are not stable between calls — four rays as
`(1, 4, 3)` with two directions, all starting at `z = -1` above a plane at `z = 0` so every distance
should be `1.0`, returns `[1.0, 1.0, 136.86, inf]`, then `[1.0, 1.0, 136.59, inf]`, then
`[1.0, 1.0, 137.91, inf]` on successive calls. On CUDA an out-of-bounds read of a Warp array is an
illegal memory access.

The guard requires `ray_directions` to have the same shape as `ray_starts`, and runs before the
`.view(-1, 3)` flatten. Comparing counts after the flatten is not enough: the flatten reinterprets
whatever it is handed, so `(4, 3)` starts with `(4, 2)` or `(2, 2, 3)` directions both yield four
rows and pass a count check.

This is a behaviour change. Calls that passed surplus directions, or a direction tensor whose layout
differed from the ray starts, worked before and now raise. No in-tree caller is affected:
`terrains/utils.py:245` is the only call site outside tests and passes `points.view(-1, 3)` and
`dirs.view(-1, 3)`, which always have the same shape. The changelog fragment carries this as a
`Changed` entry with migration guidance; I left it at the patch tier on the reading that this closes
undefined behaviour rather than removing a feature, but that is a maintainer call and moving it to
`.major.rst` is a rename away.

### Note

`isaaclab/terrains/utils.py` already works around this bug: it calls
`raycast_mesh(points.view(-1, 3), dirs.view(-1, 3), wp_mesh)` and then restores the shape itself
with `ray_hits.view(points.shape)`. It avoids the crash only because it leaves `return_distance`
at its default. That workaround becomes unnecessary with this change, but I have left it alone to
keep the diff focused.

## Type of change

## Testing

Added `source/isaaclab/test/utils/warp/test_ops.py`. It needs neither a GPU nor Isaac Sim — the
module under test imports only numpy, torch and warp — so it runs as a `unit` test alongside the
existing `test/utils/warp/test_proxy_array.py`.

The shape test is parametrized over `(2,)`, `(1, 2)`, `(3, 1, 2)` and `(2, 3, 1)` ray dimensions.
With the two reshapes reverted, three of the four fail; `(1, 2)` passes either way and is kept as
the control, since it is the only shape the existing suite covered. Ray depths are staggered so
the assertions also pin down the row-major ordering between the flatten and the unflatten, not
just the output shape.

The guard is covered by a second test parametrized over four mismatched direction shapes — too few,
too many, a wrong trailing dimension, and a different layout with a matching element count. Deleting
the guard fails all four; weakening it to a flattened-count comparison still fails the last.

I verified separately that for 3-D input the new reshape is byte-identical to the old one: 4344
output elements compared across batch shapes `(1, 1)`, `(1, 2)`, `(4, 7)` and `(16, 32)`, 0
differences, 403 misses (`inf` distances, `-1` face ids) included.

Please note what I could not run: I do not have an NVIDIA GPU, so I have not executed
`test/sensors/test_ray_caster.py` (it launches Kit) or anything on CUDA. The change is a pure
`torch.Tensor.view` on the host side plus a shape comparison, and is device-independent, but a GPU CI
run is worth having.

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/utils/warp/ops.py
+++ b/source/isaaclab/isaaclab/utils/warp/ops.py
@@ -93,8 +93,8 @@ def raycast_mesh(
     and data types to ensure proper execution. Additionally, they all must be in the same frame.
 
     Args:
-        ray_starts: The starting position of the rays. Shape (N, 3).
-        ray_directions: The ray directions for each ray. Shape (N, 3).
+        ray_starts: The starting position of the rays. Shape (..., 3).
+        ray_directions: The ray directions for each ray. Shape (..., 3).
         mesh: The warp mesh to ray-cast against.
         max_dist: The maximum distance to ray-cast. Defaults to 1e6.
         return_distance: Whether to return the distance of the ray until it hits the mesh. Defaults to False.
@@ -102,17 +102,19 @@ def raycast_mesh(
         return_face_id: Whether to return the face id of the mesh face the ray hits. Defaults to False.
 
     Returns:
-        The ray hit position. Shape (N, 3).
+        The ray hit position. Shape (..., 3).
             The returned tensor contains :obj:`float('inf')` for missed hits.
-        The ray hit distance. Shape (N,).
+        The ray hit distance. Shape (...,).
             Will only return if :attr:`return_distance` is True, else returns None.
             The returned tensor contains :obj:`float('inf')` for missed hits.
-        The ray hit normal. Shape (N, 3).
+        The ray hit normal. Shape (..., 3).
             Will only return if :attr:`return_normal` is True else returns None.
             The returned tensor contains :obj:`float('inf')` for missed hits.
-        The ray hit face id. Shape (N,).
+        The ray hit face id. Shape (...,).
             Will only return if :attr:`return_face_id` is True else returns None.
             The returned tensor contains :obj:`int(-1)` for missed hits.
+
+    The leading dimensions of the rays are arbitrary and are preserved in every output.
     """
     # extract device and shape information
     shape = ray_starts.shape
@@ -175,11 +177,11 @@ def raycast_mesh(
     wp.synchronize()
 
     if return_distance:
-        ray_distance = ray_distance.to(device).view(shape[0], shape[1])
+        ray_distance = ray_distance.to(device).view(shape[:-1])
     if return_normal:
         ray_normal = ray_normal.to(device).view(shape)
     if return_face_id:
-        ray_face_id = ray_face_id.to(device).view(shape[0], shape[1])
+        ray_face_id = ray_face_id.to(device).view(shape[:-1])
 
     return ray_hits.to(device).view(shape), ray_distance, ray_normal, ray_face_id
 
--- a/source/isaaclab/isaaclab/utils/warp/ops.py
+++ b/source/isaaclab/isaaclab/utils/warp/ops.py
@@ -114,10 +114,16 @@ def raycast_mesh(
             Will only return if :attr:`return_face_id` is True else returns None.
             The returned tensor contains :obj:`int(-1)` for missed hits.
 
-    The leading dimensions of the rays are arbitrary and are preserved in every output.
+    Raises:
+        ValueError: If :attr:`ray_directions` does not have the same shape as :attr:`ray_starts`.
     """
     # extract device and shape information
     shape = ray_starts.shape
+    if ray_directions.shape != shape:
+        raise ValueError(
+            "Expected ray_directions to have the same shape as ray_starts. Received ray_starts"
+            f" {tuple(shape)} and ray_directions {tuple(ray_directions.shape)}."
+        )
     device = ray_starts.device
     # device of the mesh
     torch_device = wp.device_to_torch(mesh.device)
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -182,6 +182,7 @@ Guidelines for modifications:
 * Ryan Gresia
 * Ryley McCarroll
 * Sahara Yuta
+* Samir Bhattarai
 * Sergey Grizan
 * Shafeef Omar
 * Shane Reetz
--- a/source/isaaclab/changelog.d/raycast-mesh-ray-dimensions.rst
+++ b/source/isaaclab/changelog.d/raycast-mesh-ray-dimensions.rst
@@ -1,0 +1,8 @@
+Fixed
+^^^^^
+
+* Fixed :func:`~isaaclab.utils.warp.raycast_mesh` shaping its distance and face-id outputs from only
+  the first two ray dimensions. Whenever :attr:`return_distance` or :attr:`return_face_id` was
+  enabled, rays with any shape other than ``(B, N, 3)`` either raised an error — including the
+  documented ``(N, 3)`` input — or, when the leading dimensions happened to multiply out, silently
+  returned a wrong-shaped tensor. Both outputs now keep the ray dimensions of the input.
--- a/source/isaaclab/changelog.d/raycast-mesh-ray-dimensions.rst
+++ b/source/isaaclab/changelog.d/raycast-mesh-ray-dimensions.rst
@@ -1,3 +1,12 @@
+Changed
+^^^^^^^
+
+* Changed :func:`~isaaclab.utils.warp.raycast_mesh` to raise a :class:`ValueError` unless
+  ``ray_directions`` has the same shape as ``ray_starts``. Fewer directions than rays previously read
+  past the end of ``ray_directions`` and returned undefined results, while surplus directions and
+  differently shaped direction tensors were silently reinterpreted. Callers that relied on either
+  must reshape ``ray_directions`` to match ``ray_starts``.
+
 Fixed
 ^^^^^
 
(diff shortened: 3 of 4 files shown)


AntoineRichard:
Thanks for contributing! @pascal-roth will take a look :).

source/isaaclab/isaaclab/utils/warp/ops.py
```diff
@@ -123,6 +125,12 @@ def raycast_mesh(
     ray_starts = ray_starts.to(torch_device).view(-1, 3).contiguous()
     ray_directions = ray_directions.to(torch_device).view(-1, 3).contiguous()
     num_rays = ray_starts.shape[0]
+    # the kernel reads one direction per ray, so a shorter tensor would be indexed out of bounds
+    if ray_directions.shape[0] < num_rays:
+        raise ValueError(
+            f"Expected at least one direction per ray. Received {num_rays} ray starts and"
+            f" {ray_directions.shape[0]} ray directions."
+        )
```
pascal-roth: I think we should raise an error when the number of directions is not equal to the number of ray starting positions, everything else is an undefined behavior

source/isaaclab/isaaclab/utils/warp/ops.py
```diff
@@ -175,11 +183,11 @@ def raycast_mesh(
     wp.synchronize()
 
     if return_distance:
-        ray_distance = ray_distance.to(device).view(shape[0], shape[1])
+        ray_distance = ray_distance.to(device).view(shape[:-1])
```
pascal-roth, line 186: That change makes it more general, which is good. I think the prev. shape is the typical one in IsaacLabs parallel environment structure, which assumes (num_envs, num_rays_per_env, ...) and enforces a certain structure. Unsure whether that change will make it more general or more complex.

source/isaaclab/isaaclab/utils/warp/ops.py
samirbhattarai135: Agreed. Switched to a shape check:

```python
shape = ray_starts.shape
if ray_directions.shape != shape:
    raise ValueError(
        "Expected ray_directions to have the same shape as ray_starts. Received ray_starts"
        f" {tuple(shape)} and ray_directions {tuple(ray_directions.shape)}."
    )
```

I compared shapes rather than counts, and moved the check above the `.view(-1, 3)`. A count check
placed after the flatten still lets undefined input through, because the flatten reinterprets whatever
it is handed: `(4, 3)` ray starts with `(4, 2)` directions, or with `(2, 2, 3)` directions, both come
out as four rows and pass a count comparison. Checking before the flatten also lets the message quote
the shapes the caller actually passed, and rejects the call before two host-to-device copies.

On the "undefined behavior" framing — it is worse than a wrong number, the values are not stable
between calls. Four rays as `(1, 4, 3)` with only two directions, every start at `z = -1` above a
plane at `z = 0`, so every distance should be `1.0`. Five consecutive calls in one process on
`develop`:

```
[1.0, 1.0, 136.86, inf]
[1.0, 1.0, 136.59, inf]
[1.0, 1.0, 137.91, inf]
[1.0, 1.0, 136.86, inf]
[1.0, 1.0, 136.59, inf]
```

Nothing in the tree relies on the old tolerance. `terrains/utils.py:245` is the only non-test caller
and builds its directions with `torch.zeros_like(points)` a few lines above, so the shapes are equal
by construction; `test/sensors/test_ray_caster.py` passes matching `(1, 2, 3)` tensors. It also brings
this in line with `raycast_dynamic_meshes`, whose `.view(n_envs, n_rays_per_env, 3)` on the direction
tensor (`ops.py:310`) already rejects short, surplus and differently shaped input — the difference is
only that this is an explicit `ValueError` rather than a torch `RuntimeError`.

Calls that passed surplus directions did work before, so the changelog now carries this as a `Changed`
entry with migration guidance rather than a `Fixed` one, and I dropped the test that asserted the old
tolerance. I left the fragment at the patch tier on the reading that this closes undefined behaviour
rather than removes a feature — say the word and I will move it to `.major.rst`.

source/isaaclab/isaaclab/utils/warp/ops.py
samirbhattarai135: This one is narrower than it looks. Before this PR the docstring already documented `ray_starts` as
`Shape (N, 3)` and the distance output as `Shape (N,)`, but that exact call raised
`RuntimeError: shape '[4, 3]' is invalid for input of size 4` as soon as `return_distance` was set,
because `shape[1]` there is the xyz axis rather than a ray count. The change makes the code match a
contract that was already published, rather than widening one.

For the `(num_envs, num_rays_per_env, ...)` layout you are protecting, the two expressions are the
same value — `shape[:-1]` *is* `(shape[0], shape[1])` whenever `shape` has length three. I ran both
versions side by side on a bumpy 8×8 grid mesh over batch shapes `(1, 1)`, `(1, 2)`, `(4, 7)` and
`(16, 32)`, with the ray origins spread wide enough to force misses: 4344 output elements compared,
0 differences, 403 misses (`inf` distances, `-1` face ids) included. Nothing is traded away for that
layout.

The structure is still enforced where it is actually assumed. `raycast_dynamic_meshes` reshapes to
`(n_envs, n_rays_per_env, 3)` and launches `dim=[n_meshes, n_envs, n_rays_per_env]`, and the ray-caster
sensors bypass both wrappers entirely — they allocate their own `(num_envs, num_rays)` buffers and
launch `raycast_mesh_masked_kernel` directly. `raycast_mesh` flattens to `view(-1, 3)` and launches
`dim=num_rays`, so it never reads a batch or a per-env count; the two-dimension assumption only ever
lived in those two reshape lines. `ray_hits` and `ray_normal`, on either side of them, already used
`.view(shape)` and already preserved arbitrary leading dimensions. So this is the two remaining outputs
agreeing with the other two, not a new rule — same one-liner, not a more complex one.

Its one non-test caller already works around the current behaviour: `terrains/utils.py:245` passes a
flat `(N, 3)` and unflattens the hits itself, and escapes the crash only because it requests neither
distance nor face id.

If you would rather this layer required `(num_envs, num_rays_per_env, 3)`, that is a defensible
position, but it is a separate and breaking change — it would reject `terrains/utils.py:245` as
written. I would rather open it as its own issue than fold it into this fix.

AntoineRichard:
@samirbhattarai135 could you kindly provide use-cases / justifications why this feature would be required? Is it just to solve #2416 ?

samirbhattarai135:
@AntoineRichard, the reshape commit is the fix for #2416 and changes nothing for existing callers. `shape[:-1]` is the same expression as `(shape[0], shape[1])` for 3-D input, verified over 4344 output elements with 0 differences. 
The second commit, the direction-shape guard, is not from #2416. I added it at @pascal-roth request in review, and it is the only reason the PR is marked as a breaking change. I'm happy to drop it if you'd prefer to keep this to the bug fix. The PR then has no behaviour change for any call that works today. 
Could you also approve the workflow runs when you get a chance? All ten are still sitting at action_required behind the first-time-contributor gate.

PR #7035, opened 2026-08-12
Stop routing OVRTX renderer logs through stdout for tests

mataylor-nvidia:
# Description

Fixes large blocks of NUL bytes being written to CI logs during kitless rendering tests, at the source.

The kitless rendering tests overrode `OVRTXRendererCfg.log_file_path` to `/dev/stdout` (`CON` on Windows), added in #5749 so the RTX shader-compilation timeline would be visible while a test appeared to hang.

OVRTX opens that path natively and, running with `keep_system_alive=True`, retains the handle for the lifetime of the process. Under pytest's file-descriptor capture, fd 1 is a temporary file that pytest rewinds and truncates between tests, so the retained handle keeps writing at a stale offset and the resulting sparse gap is read back as a block of NUL bytes.

This drops the override entirely rather than working around it. `log_file_path` already defaults to `<system temp>/ovrtx_renderer.log`, so `_redirect_ovrtx_renderer_log_to_stdout()` only ever moved the log somewhere less safe — and its hardcoded camera-name list (`tiled_camera`, `base_camera`, `wrist_camera`) silently missed any camera named otherwise.

To keep the log visible, an autouse fixture in `source/isaaclab_tasks/test/conftest.py` replays the bytes each test appends. It is a no-op for tests that never build an OVRTX renderer, records only the byte range written during the test, and replays from the beginning when the 64 bytes in front of that range changed — the log was truncated and rewritten, by OVRTX opening one an earlier pytest process left behind, say.

Net effect: 48 lines deleted, 36 added, and no test-owned file descriptor competing with pytest for stdout.

## Alternative considered

Running the child pytest with `--capture=sys` (#7031) avoids the sparse-file mechanism, but breaks `SimulationApp.__init__`, which calls `faulthandler.enable()` — that needs a real fd, and under `sys` capture `sys.stderr` is a `CaptureIO` with no `fileno()`. It also sends every native fd-1/fd-2 write and all subprocess output from passing tests straight to the job log, reversing #6462.

## Type of change

## Screenshots

Failure output with the fixture in place:

```
--------------------------- Captured stdout teardown ---------------------------

----- OVRTX renderer log: test_failing_ovrtx_test -----
[omni.rtx] BOOM
```

## Checklist

### DIFF
--- a/tools/conftest.py
+++ b/tools/conftest.py
@@ -261,17 +261,6 @@ def _get_diagnostics(pre_kill_diag=""):
     return diag
 
 
-def _report_ovrtx_log(label):
-    """Log what the subprocess left in the renderer log and return it, for attaching to a failure report.
-
-    The process that wrote it is gone by now, so this is the only report its renderer output reaches.
-    """
-    section = ovrtx_log.format_log_section(ovrtx_log.LOG_PATH, label)
-    if section:
-        logger.info(section)
-    return section
-
-
 def _capture_system_diagnostics():
     """Capture system diagnostics (GPU, memory, processes) for crash investigation.
 
@@ -708,7 +697,6 @@ def _run_one_pass(
             if len(diag) > 10000:
                 diag = diag[:10000] + "\n... (truncated)"
             logger.info(diag)
-            _report_ovrtx_log(pass_file_label)
             continue
 
         if kill_reason == "timeout" and not has_report and timeout_attempts < TIMEOUT_RETRIES:
@@ -729,7 +717,6 @@ def _run_one_pass(
             if len(diag) > 10000:
                 diag = diag[:10000] + "\n... (truncated)"
             logger.info(diag)
-            _report_ovrtx_log(pass_file_label)
             continue
         break
 
@@ -740,7 +727,7 @@ def _run_one_pass(
         diag = _get_diagnostics(pre_kill_diag)
         logger.warning(f"⚠️  {ctx.test_file}{suffix}: startup hang after {STARTUP_HANG_RETRIES + 1} attempt(s)")
         logger.info(diag)
-        ovrtx_log_section = _report_ovrtx_log(pass_file_label)
+        ovrtx_log_section = ovrtx_log.format_log_section(ovrtx_log.LOG_PATH, pass_file_label)
 
         msg = f"Startup hang after {ctx.startup_deadline}s (retried {STARTUP_HANG_RETRIES} time(s))"
         details = f"{msg}\n\n=== SYSTEM DIAGNOSTICS ===\n{diag}\n\n"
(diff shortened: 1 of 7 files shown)


mataylor-nvidia:
@greptile review

StafaH (approved):

AntoineRichard (dismissed):
The root-cause analysis looks credible, but the replacement still has important diagnostic gaps and no regression coverage. Requesting changes for the following:

1. **Renderer diagnostics are unavailable on hangs and native crashes** (`source/isaaclab_tasks/test/conftest.py:57-75`). The log is replayed only after the fixture resumes from `yield`. That code cannot run after a segfault, abort, OOM, or when the outer CI supervisor kills a hung process with `SIGKILL`. These are precisely the failures where the native renderer log is most useful. Please have the outer supervisor include a bounded tail of `ovrtx_renderer.log` in its crash/timeout diagnostics, or otherwise collect it outside the process being diagnosed.

2. **There is no regression test for the capture bug this PR fixes.** Existing rendering jobs completing successfully establish neither that retained native writes no longer introduce NUL bytes nor that renderer output is preserved on failure. The repository guidelines require a regression test to fail without the fix. Please add behavioral coverage that reproduces the retained-fd/capture interaction and asserts that the resulting output contains no NUL block while the expected renderer diagnostics remain available. Coverage for truncation/rewrite and abnormal termination would also pin the new failure paths.

3. **The 64-byte fingerprint can still mistake a rewrite for an append** (`source/isaaclab_tasks/test/conftest.py:42-54`). OVRTX startup output is substantially deterministic. If another process truncates the shared file and rewrites the same bytes around the previous offset before appending new output, the fingerprint is unchanged, so the fixture seeks to the stale offset and omits the beginning of the current log. A process-unique log path or explicit ownership/session cleanup would avoid inferring file generation from content.

4. **Failure replay is unbounded** (`source/isaaclab_tasks/test/conftest.py:72-75`). `handle.read()` prints the entire appended verbose renderer log into pytest capture and potentially JUnit output. A long failing rendering test can therefore recreate the oversized-log problem with real text rather than sparse NUL bytes. Please cap the replay to a documented tail size and state when earlier bytes were omitted.

I verified these findings against head commit `e7155265df5350ccda8644fd92e87e7ee8190e1d`. The passing kitless-rendering CI job contains no NUL bytes, but it also contains no `OVRTX renderer log:` replay section because no failure exercised that path.

mataylor-nvidia:
Added fixes for the 4 comments [AntoineRichard](https://github.com/AntoineRichard) posted

tools/conftest.py
```diff
@@ -260,6 +261,17 @@ def _get_diagnostics(pre_kill_diag=""):
...
 
 
+def _report_ovrtx_log(label):
+    """Log what the subprocess left in the renderer log and return it, for attaching to a failure report.
+
+    The process that wrote it is gone by now, so this is the only report its renderer output reaches.
+    """
+    section = ovrtx_log.format_log_section(ovrtx_log.LOG_PATH, label)
+    if section:
+        logger.info(section)
```
pbarejko: It seems you redirect all ovrtx logging into info, should we redirect ovrtx logged messages to appropriate loggers?

pbarejko (approved):

tools/conftest.py
mataylor-nvidia: I removed this

mataylor-nvidia:
I created another PR ontop of this one and broke the test on purpose to see we are indeed still receiving ovrtx logs:

https://github.com/isaac-sim/IsaacLab/actions/runs/31644824736/job/94280076646#step:5:695

PR #7041, opened 2026-08-12
Fix relative deformable scene state

mmichelis:
# Description

Fix relative deformable nodal positions in `InteractiveScene.get_state()` and `InteractiveScene.reset_to()` by broadcasting each environment origin across all nodes. Add focused regression tests for both state serialization and restoration, including reordered environment indices.

## Type of change

## Screenshots

Not applicable.

## Validation

- Before the fix, both new regression tests fail with tensor broadcasting errors.
- After the fix, both focused tests pass.
- `source/isaaclab/test/scene/test_interactive_scene.py`: 13 passed.
- `uv run isaaclab -f`: passed.
- `uv run --extra test python tools/changelog/cli.py check develop`: passed.

## Checklist

- [x] No documentation changes are required

### DIFF
--- a/source/isaaclab/test/scene/test_interactive_scene.py
+++ b/source/isaaclab/test/scene/test_interactive_scene.py
@@ -107,6 +107,66 @@ def test_relative_flag(device, setup_scene):
     assert_state_equal(prev_state, scene.get_state(is_relative=True))
 
 
+def test_get_relative_deformable_state_offsets_all_nodes():
+    env_origins = torch.arange(12, dtype=torch.float32).reshape(4, 3)
+    nodal_position = torch.arange(60, dtype=torch.float32).reshape(4, 5, 3)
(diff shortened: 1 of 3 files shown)


source/isaaclab/test/scene/test_interactive_scene.py
AntoineRichard, line 1: Are these tests really needed. Is this not exercised somewhere else?

mmichelis: There is a test_relative_flag in this file as well, but that only tests articulation and rigid objects. Since the bug was not caught before, it is quite important to also have the equivalent test for deformables which we add here. I reduced it down a bit so it's more concise in a4cbe5e.

AntoineRichard (approved):

kellyguo11 (approved):

PR #7042, opened 2026-08-12
Deprecate IO descriptors

AntoineRichard:
# Description

Deprecate IO descriptors in Isaac Lab 3.0 ahead of their removal in Isaac Lab
3.2.

This change:

- marks the stable and Warp-specific descriptor APIs as deprecated;
- emits a `FutureWarning` from public descriptor collection/export entrypoints;
- marks the shared `--export_io_descriptors` training option as deprecated;
- preserves compatibility behavior through the 3.0 deprecation window;
- removes the IO descriptor tutorial and generated YAML examples; and
- directs supported RSL-RL/PyTorch deployments to the LEAPP export workflow.

Warp observation-dimension metadata remains temporarily because the Warp
frontend still consumes it for runtime buffer planning. LEAPP does not yet
replace that Warp-first path.

## Type of change

- API deprecation (non-breaking compatibility change)
- Documentation update

## Release backport

## Screenshots

Not applicable.

## Verification

- `uv run python -m pytest source/isaaclab/test/managers/test_observation_manager_unit.py`
- `uv run python -m pytest source/isaaclab_rl/test/test_entrypoints_common.py`
- Python compile checks for the affected modules
- `uv run --isolated --extra dev -- make -C docs current-docs`
- `uv run isaaclab -f` (formatting/lint hooks)
- `uv run python tools/changelog/cli.py check --include-worktree` against the canonical `develop` base

No deprecation-specific tests were added; the public behavior remains available
during the deprecation window.

## Checklist


source/isaaclab/isaaclab/envs/utils/io_descriptors.py
```diff
@@ -21,10 +29,23 @@
...
+        FutureWarning,
+        stacklevel=stacklevel,
+    )
+
+
 @configclass
 class GenericActionIODescriptor:
     """Generic action IO descriptor.
 
+    .. deprecated:: 3.0
```
StafaH, line 46: Can't we just deprecate it altogether without warning, and just keep it in the migration guide?

StafaH (changes requested):

source/isaaclab/isaaclab/envs/utils/io_descriptors.py
AntoineRichard: It should only warn if the users collect the descriptors and the docstring lets the users know this is getting removed. It feels like the right balance to me.

kellyguo11:
are we sure nobody is using these? I think Isaac Sim team might still be depending on the IO descriptors for their policy deployment work.

StafaH (approved):

AntoineRichard:
@kellyguo11 it's not removing them, just removing them from the doc, and pushing folks to use Leapp instead!

ooctipus (approved):

StafaH:
run-ci

PR #7044, opened 2026-08-12
[Fix] Handle asset cleanup after partial initialization

mmichelis:
# Description

Guard asset callback handles during cleanup so destructors tolerate objects whose constructors fail before callback registration completes. This preserves the original constructor exception instead of masking it with an `AttributeError` from `AssetBase.__del__`.

This occurs when spawning a volume deformable without the optional tetrahedralization dependencies: the actionable dependency error is raised before callback registration, so callback cleanup must finish gracefully without obscuring it.

Fixes NVBug 6591313.

## Type of change

## Screenshots

Not applicable.

## Validation

- Manual partial-initialization reproduction (failed with the reported `AttributeError` before the fix and passed after the fix)

## Checklist

- [x] Documentation changes are not required
- [ ] I have added tests that prove the fix is effective

### DIFF
--- a/source/isaaclab/isaaclab/assets/asset_base.py
+++ b/source/isaaclab/isaaclab/assets/asset_base.py
@@ -501,13 +501,13 @@ class AssetBase(ABC):
 
     def _clear_callbacks(self) -> None:
         """Clears all registered callbacks."""
-        if self._initialize_handle is not None:
+        if getattr(self, "_initialize_handle", None) is not None:
             self._initialize_handle.deregister()
             self._initialize_handle = None
-        if self._invalidate_initialize_handle is not None:
+        if getattr(self, "_invalidate_initialize_handle", None) is not None:
             self._invalidate_initialize_handle.deregister()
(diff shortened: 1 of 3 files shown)


source/isaaclab/test/assets/test_asset_selector_cache.py
AntoineRichard, line 1: Please remove the test.

source/isaaclab/isaaclab/assets/asset_base.py
AntoineRichard, line 1: I'm not convinced this is needed. If this fails the code is cooked anyway...

mmichelis: So this was raised when tetrahedralization is not installed for deformables, which then raises an error by itself. The clear callbacks fails as well after that, so the user sees two errors of which one is irrelevant. It would be nice if clear callbacks finishes successfully.

AntoineRichard (approved):

kellyguo11 (approved):

PR #7047, opened 2026-08-12
Make KaminoSolverCfgBase public

rubengrandia:
# Description

Short amendment to previous PR https://github.com/isaac-sim/IsaacLab/pull/6916, making the CfgBase public is slightly more scalable and turns:
`isinstance(solver_cfg, (KaminoPADMMSolverCfg, KaminoDVISolverCfg))` into `isinstance(solver_cfg, KaminoSolverCfgBase)`. The latter would remain valid if we ever add a third backend.


source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
```diff
@@ -64,7 +63,7 @@ def _requires_external_contacts(solver_cfg: NewtonSolverCfg) -> bool:
         """
         if isinstance(solver_cfg, MJWarpSolverCfg):
             return not solver_cfg.use_mujoco_contacts
-        if isinstance(solver_cfg, (KaminoPADMMSolverCfg, KaminoDVISolverCfg)):
+        if isinstance(solver_cfg, KaminoSolverCfgBase):
```
hujc7, line 66: Agent review: this collapse is the point of the PR, and one call site still carries the old form.

`source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py:1003` is now the **only** `(KaminoPADMMSolverCfg, KaminoDVISolverCfg)` pair left in the tree:

```python
if isinstance(solver_cfg, (KaminoPADMMSolverCfg, KaminoDVISolverCfg)) and solver_cfg.use_collision_detector:
    builder.add_shape_sphere(body=body, radius=0.05)
    builder.add_ground_plane()
```

It is also the line a third backend would quietly break — the test would stop adding collision geometry for it and fail somewhere unrelated.

- [ ] no `(KaminoPADMMSolverCfg, KaminoDVISolverCfg)` pairs remain

source/isaaclab_newton/isaaclab_newton/physics/__init__.pyi
```diff
@@ -15,6 +15,7 @@ __all__ = [
     "KaminoMaterialsCfg",
     "KaminoPADMMCfg",
     "KaminoPADMMSolverCfg",
+    "KaminoSolverCfgBase",
```
hujc7, line 18: Agent review: the class is exported but absent from the API reference.

| | count |
|---|---|
| names in `__all__` | 25 |
| names in `docs/source/api/lab_newton/isaaclab_newton.physics.rst` | 24 |
| missing | `KaminoSolverCfgBase` |

It needs the autosummary line plus an `autoclass::` block — the same two edits every other Kamino class got in #6916. The new `kamino-solver-cfg-base.minor.rst` fragment announces it as "the public base configuration class for Kamino solvers", so the API reference is exactly where a reader will go looking and not find it.

While there: `kamino-solver.rst:182-185` opens the shared-field tables with "shared by `KaminoPADMMSolverCfg` and `KaminoDVISolverCfg`" — that sentence now has a single name to use.

- [ ] `KaminoSolverCfgBase` present in the API reference

source/isaaclab_newton/isaaclab_newton/physics/kamino_manager_cfg.py
```diff
@@ -223,7 +223,7 @@ class KaminoMaterialsCfg:
 
 
 @configclass
-class _KaminoSolverCfgBase(NewtonSolverCfg):
+class KaminoSolverCfgBase(NewtonSolverCfg):
```
hujc7, line 226: Agent found bug: publishing this class also makes an incomplete config publicly constructible, and the failure it produces is unlabeled.

```
KaminoSolverCfgBase()          -> constructs, 17 fields
NewtonCfg(solver_cfg=cfg)      -> accepted; class_type -> NewtonKaminoManager
cfg.to_solver_config()         -> NotImplementedError        (no message)
```

Nothing signals the class is not usable directly. `class_type` points at the **concrete** manager, so the config survives all the way into solver bring-up before raising from a private method the caller never invoked.

<details><summary>Contrast with the sibling base one level up</summary>

```
NewtonSolverCfg()              -> class_type = ...:NewtonManager   (the abstract manager)
  has to_solver_config?        -> False
  KaminoSolverCfgBase has it?  -> True
```

`NewtonSolverCfg` is a pure data holder dispatching to an abstract manager, so instantiating it is harmless. `KaminoSolverCfgBase` carries a working `to_solver_config()` with one hole in it and points at a real manager.
</details>

A message naming the concrete classes is enough:

```python
        raise NotImplementedError(
            f"{type(self).__name__} is a base configuration. "
            "Use KaminoPADMMSolverCfg or KaminoDVISolverCfg."
        )
```

Worth settling together with the API-reference entry above: documenting the base is what makes it reachable in the first place.

- [ ] picking the base by mistake produces an actionable error

hujc7 (commented):
Agent review: the rename and the two `coupler.py` collapses are exactly right — no comments on those. Three follow-throughs on the newly public surface: one leftover call site, the missing API-reference entry, and an unlabeled failure if the base is used directly.

PR #7048, opened 2026-08-12
hardcode thread limit for leapp tests

frlai:
# Description

There is a race condition in the current usd version that causes some environment setup to fail. this is especially prevalent in the leapp tests becuase leapp performs multiple environment setups per run. This PR sets the CPU thread count to 1. This provides a temporary fix by mitigating the issue.

## Checklist

### DIFF
--- a/source/isaaclab_rl/test/export/leapp_initialized_checkpoints.py
+++ b/source/isaaclab_rl/test/export/leapp_initialized_checkpoints.py
@@ -344,7 +344,9 @@ def main(argv: Sequence[str] | None = None) -> None:
 
     from isaaclab.app import AppLauncher
 
-    app_launcher = AppLauncher(headless=True)
+    # TODO: Remove once usd-core>=26.5 is the minimum. Earlier OpenUSD releases
+    # can corrupt the heap while parsing the Newton Franka payload concurrently.
+    app_launcher = AppLauncher(headless=True, limit_cpu_threads=1)
     simulation_app = app_launcher.app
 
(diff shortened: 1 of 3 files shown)


source/isaaclab_rl/test/export/leapp_initialized_checkpoints.py
```diff
@@ -344,7 +344,9 @@ def main(argv: Sequence[str] | None = None) -> None:
 
     from isaaclab.app import AppLauncher
 
-    app_launcher = AppLauncher(headless=True)
+    # TODO: Remove once usd-core>=26.5 is the minimum. Earlier OpenUSD releases
+    # can corrupt the heap while parsing the Newton Franka payload concurrently.
+    app_launcher = AppLauncher(headless=True, limit_cpu_threads=1)
```
kellyguo11, line 349: I thought this issue was addressed in kit already. this test still launches isaac sim right?

frlai: yes it should launch isaac sim. I'm not sure which version we are using actually. but the bug reported seems to strongly point to this being the issue.

I think an alternative can be to run the tests using a different physics backend instead actually. would that be a more desirable fix?

kellyguo11: the version we have in kit atm supposedly has a patch that includes the USD fix. I think a different physics backend could still run into the same issue if it is still indeed the USD bug. let me maybe monitor the CI for a bit more and see if this keeps happening.

PR #6446, opened 2026-07-10
Relocates RTX determinism settings to isaaclab_physx backend

nvsekkin:
# Description

- Removes `AppLauncher.apply_rtx_determinism_settings()` and the `/rtx/...` carb writes from core `isaaclab.app`.
- `AppLauncher` now publishes the backend-agnostic `/isaaclab/render/deterministic` setting from `--deterministic`.
- `IsaacRtxRenderer` consumes that setting on init and applies RTPT reproducibility settings via the new `apply_isaac_rtx_determinism_settings()` in `isaaclab_physx`.

## Type of change

- Refactor

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7049, opened 2026-08-12
Re-enable the Kit visualizer pause checks

fatimaanes:
Temporary workaround for NVBUG 6570125 while we wait for the upstream renderer fix.

## Problem

RTX responsiveDenoising continues refining the image after physics stops, so fresh renders captured during a pause are no longer stable.
The affected Kit visualizer checks are currently green for the wrong reasons:
Tiled camera: #6658 changed the paused captures to reuse the cached sensor frame, so the assertion effectively compares the same frame against itself and can't detect renderer instability.
Viewport: still performs a real comparison, but the known denoiser noise can push it over the existing threshold and the failure is absorbed by retries.

## What this changes

This PR restores fresh renders for the tiled pause check and adjusts the per-channel difference threshold only for the affected Kit RTX pause paths.
The existing 100-pixel gate stays unchanged. We're changing how large a per-channel difference must be before a pixel is counted, rather than simply allowing more differing pixels.

| site | @50 | @80 | @100 | @120 | @150 | @160 | outcome |
| Kit viewport (Newton) | 67 | **12** | 8 | 7 | 3 | — | 80 |
| Kit viewport (PhysX) | 14 | 4 | 3 | 1 | 0 | — | 80 |
| Kit tiled (Newton) | 999 | 604 | 226 | 120 | 59 | **49** | 160 |
| Kit tiled (PhysX) | 103 | 4 | 4 | 2 | 0 | — | 80 |
| Newton ViewerGL | 0 | 0 | 0 | 0 | 0 | — | 50 |

Values are the number of differing pixels at each per-channel threshold.
For the margin-critical case (Kit tiled camera, Newton backend), threshold 160 leaves at most 49 px of paused denoiser noise (worst of n=36) against at least 160 px with real motion (weakest of n=72), both measured with the denoiser enabled. With the existing 100-pixel gate the known renderer noise passes at 2.0× margin while real motion is still detected at 1.6×.

## Changes

`_assert_frames_remain_stable` now accepts a per-call `channel_diff_threshold`; the default remains 50.
Kit RTX pause checks use 80 for the viewport, 160 for the tiled camera on Newton, and 80 for the tiled camera on PhysX — three separate constants, since the tiled PhysX cell measures 103 px at the default threshold and must not silently track a later change to the viewport value. Newton ViewerGL remains at the strict default since it doesn't exercise the affected DLSS path.
`max_differing_pixels` remains 100 everywhere.
The tiled pause check is restored to fresh renders instead of cached frames, so the assertion is actually exercising renderer stability again.

## Validation

With retries disabled, the workaround passed 40/40 runs across all four affected configurations.

As a control, restoring the original threshold of 50 reproduced the failure 6/6 times at 766–949 differing pixels. Those six failures are all the **tiled** assertion: the control ran under `pytest -x` and the viewport test is defined first in the same file, so the run aborted at the tiled test and the viewport site passed at 50 in those reps. The control therefore demonstrates that the tiled workaround filters real denoiser noise rather than simply making the test pass; it is not evidence for the viewport site.

The viewport raise from 50 to 80 rests instead on production CI, where the assertion has really been failing (`107 pixels differed, expected at most 100`, with a historical band reaching 161 px), plus n=5 bench samples showing 67 px at threshold 50 falling to 12 px at 80.

The same change was validated on three independent L40S CI runners through #7055, with all four visualizer integration tests passing first-pass, no retries. Restoring fresh renders also had no meaningful runtime impact (~27s median).

This is a temporary workaround for NVBUG 6570125 and should be removed or re-measured once the upstream fix is available. That revert is load-bearing rather than cosmetic: with the denoiser disabled — the world after the renderer fix — the weakest real motion in the governing cell measures 80 px at threshold 160, which is *below* the 100-pixel gate. Leaving 160 in place after the fix ships would make the tiled check blind to the weakest motion it exists to catch, so the removal comment on the constants is the thing guarding that.


source/isaaclab_visualizers/changelog.d/fanes-nvbug-6570125-adjust-threshold.rst
```diff
@@ -0,0 +1,18 @@
+Changed
```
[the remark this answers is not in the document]
fatimaanes: Done — renamed to `fanes-nvbug-6570125-adjust-threshold.skip` in a3aca0724f.

Agreed: the change is confined to the visualizer integration test helper and has no user-visible package behavior. `docs/source/refs/contributing.rst:172` reserves `.skip` for "CI / docs / test-only PRs", and every prior test-only fragment in this package used it (#6446, #6632, #6686, #6418).

The NVBUG 6570125 rationale stays in the code comment on the threshold constants, as you suggested. I kept a two-line internal note in the fragment to match the shape of the existing `.skip` fragments here (e.g. `esekkin-relocate-rtx-determinism.skip`); the fragment publishes nothing either way.

source/isaaclab_visualizers/test/visualizer_integration_utils.py
```diff
@@ -175,6 +175,20 @@
 _TILED_CAMERA_MOTION_MIN_DIFFERING_PIXELS = 25
 """Minimum differing pixels for tiled camera motion checks."""
 
+# NVBUG 6570125 — Remove these overrides once it ships the fix and paused frames are stable again.
+_KIT_PAUSED_VIEWPORT_CHANNEL_DIFF_THRESHOLD = 80
+"""Per-channel threshold for paused Kit viewport comparisons (0–255 space)."""
+
+_KIT_PAUSED_TILED_CAMERA_NEWTON_CHANNEL_DIFF_THRESHOLD = 160
```
matthewtrepte, line 182: hm more of a note than a request: 160 out of 256 is a large part of the range; its not rejecting many cases. not sure if there's a cleaner way like loosening also the pixel count threshold so that this threshold doesn't need to be as high

but given that its temporary and well documented that is so and what the bug is, i think its fine.

fatimaanes: Yeah, I had the same concern. 160 felt high to me too. The measured separation from real motion is what made me comfortable with it as a temporary WAR, and I wouldn't want this to quietly become permanent.

matthewtrepte (approved):

PR #7055, opened 2026-08-12
[CI probe, do not merge] Run the Kit visualizer pause checks on CI hardware (NVBUG 6570125)

fatimaanes:
**CI probe — not for merge. The PR of record is #7049.**

This branch carries the exact same commits as #7049. It exists only to get the test suite to run.

## Why this duplicate exists

#7049 comes from a fork, and fork PRs receive no repository secrets. Without `NGC_API_KEY` the `Build Base Docker Image` job cannot log in to `nvcr.io` and fails at the Isaac Sim base image pull, and because every test job carries `needs: [build, config]`, all ~26 of them report SKIPPED — including `test-isaaclab-visualizers`, the job that runs the very test #7049 re-enables. That PR's central claim is therefore unverifiable by its own pipeline.

Opening the same change from an in-repo branch restores secrets, so the test matrix actually runs.

## What this run is meant to measure

The paused-frame thresholds in #7049 (viewport 80, Kit tiled camera 160) were calibrated on an **L40** (driver 595.58.03, Isaac Sim 6.1.0-alpha.47). CI runs on `g6e.2xlarge`, i.e. an **L40S** — same Ada AD102 family, different SKU.

There is one cell where the two can be compared directly, and they disagree. On the Kit viewport pause site at a channel threshold of 50, CI has been observed at **107 px** (job 93935796248, `expected at most 100`), with a historical failing band of 105–161 px. The same cell measured on the L40 bench peaked at **67 px** over 5 samples. CI runs roughly 1.6× hotter there.

The Kit **tiled** cell — the one that governs the 160 constant — has never produced a real number on CI hardware at all, because #6658 set both paused captures to `force_recompute=False` and left the assertion comparing a cached frame against itself. Its measured headroom (49 px against a gate of 100) is entirely L40-derived.

This run produces that missing number. The assertion prints its pixel count on failure, so a red result is as useful as a green one.

## Disposition

Closing once the run completes. Any threshold change it motivates lands on #7049.


fatimaanes:
CI probe complete. All four visualizer integration tests passed on L40S (g6e.2xlarge), two independent runners, first pass, no retries. The L40→L40S transfer question is answered. #7049 is the PR of record.

fatimaanes:
Closing — this served its purpose.

This PR was never intended to merge.

Review continues on #7049.

PR #6980, opened 2026-08-07
[Worflow] Make newton and rsl_rl default across core tasks

StafaH:
# Description

> [!IMPORTANT]
> Confirm the pull request base before submitting. Target `develop` for all
> contributions. The `release/3.0.0-beta2` branch is a frozen stable landing
> snapshot and is not used for ongoing maintenance.

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7065, opened 2026-08-13
[Workflow] Default manipulation core tasks to Newton

StafaH:
Splits the Cabinet and Lift Newton defaults from #6980.\n\n- Defaults Cabinet, Lift, and Kuka Allegro task configurations to Newton MJWarp.\n- Includes the shared three-backend core smoke-test matrix and teardown reliability cleanup.\n- Removes obsolete Play/MJWarp-discovery test utility logic.\n\nThe shared test commit is intentionally identical to the other environment shards; rebase/drop it after the first shard merges.\n\nSupersedes the manipulation portion of #6980.


AntoineRichard (approved):

source/isaaclab_tasks/test/core/test_environments.py
```diff
@@ -22,16 +22,18 @@
 from env_test_utils import _run_environments, setup_environment  # isort: skip
 
 
+@pytest.mark.parametrize("physics_preset_name", ["newton_mjwarp", "physx", "isaacsim_physx"])
```
AntoineRichard: This parametrizes every core task with all three global preset names, but preset resolution falls back to `default` when a requested preset is unavailable. Consequently, a case labeled `physx` can actually run the newly-default Newton backend (or another task default), so the matrix does not guarantee the backend it claims to cover. Please generate only valid `(task, physics preset)` pairs, or use an explicit strict physics selector that errors when the backend is unavailable.

This also removes `include_play=False`, substantially expanding collection on top of the three-way parameterization. The third task shard took about 1h46m. Please restore the Play filter unless testing every Play variant here is intentional.

source/isaaclab_tasks/changelog.d/default-newton-manipulation.minor.rst
```diff
@@ -0,0 +1,5 @@
+Changed
+^^^^^^^
+
+* **Breaking:** Changed the Cabinet and Lift task families to use Newton MJWarp
```
AntoineRichard, line 4: This entry explicitly classifies the default-backend switch as breaking, but I could not find a prior deprecation for the Cabinet or Lift default behavior. Please reconcile this with the repository policy that breaking changes require a deprecation first, or clarify why this workflow-default change is exempt.

AntoineRichard:
I traced the two current red checks at `ca111ab`; neither currently points to a regression in the manipulation default changes:

- **Docker Dependency Licenses Check:** the PR image was found and pulled from ECR, but the workflow later directly pulled the private `nvcr.io/.../internal/isaac-sim:latest-develop@sha256:...` base image. Fork PRs do not receive `NGC_API_KEY`, so this failed with `Access Denied` before the dependency-license scan ran. This is a workflow/credential issue.
- **rendering-correctness-kitless (legacy):** five of six files passed, including both Kuka Lift rendering suites. Shadow Hand passed its first RGB case, then stalled in the albedo case until the 1000-second per-file watchdog fired. There was no OOM or assertion failure. On sibling PR #7064, Shadow Hand passed while Franka cloth instead exited 139, which strongly suggests flaky GPU rendering/resource contention in this parallel job rather than a deterministic change from this PR.

I would rerun or isolate those jobs rather than modify the Newton manipulation configurations based on these failures.

source/isaaclab_tasks/test/core/test_environments.py
StafaH: Fixed in 237d2f3ef: the matrix now emits only task/backend pairs that explicitly declare the requested PhysicsCfg preset. The Play filter remains removed because no registered core task IDs contain Play variants.

source/isaaclab_tasks/changelog.d/default-newton-manipulation.minor.rst
StafaH: The default switch is intentional for the 3.0 release. Per the release direction, we are documenting the direct breaking migration in the changelog rather than adding a deprecation alias or transitional behavior.

kellyguo11 (approved):

PR #7066, opened 2026-08-13
[Workflow] Default classic core tasks to Newton

StafaH:
Splits the Cartpole and shared-renderer defaults from #6980.\n\n- Defaults Cartpole to Newton MJWarp.\n- Defaults `MultiBackendRendererCfg` to the Newton renderer while retaining explicit Isaac RTX selection.\n- Consolidates core environment smoke tests into one three-backend matrix (`newton_mjwarp`, `physx`, `isaacsim_physx`).\n- Removes obsolete Play/MJWarp-discovery test utility logic and the redundant Newton test module.\n\nValidation: `uv run python -m pytest source/isaaclab_tasks/test/core/test_preset_kit_decision.py source/isaaclab_tasks/test/core/test_runtime_compatibility.py -q` (51 passed).\n\nSupersedes the classic-controls portion of #6980.


AntoineRichard (approved):

source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
```diff
@@ -22,11 +22,11 @@ class _AutoRtxRendererCfg(RendererCfg):
 
 @configclass
 class MultiBackendRendererCfg(PresetCfg):
-    default: IsaacRtxRendererCfg = IsaacRtxRendererCfg()
+    default: NewtonWarpRendererCfg = NewtonWarpRendererCfg()
```
AntoineRichard: This changes the renderer default for every task that uses `MultiBackendRendererCfg`, not only the Cartpole family. The failures show that several existing consumers cannot use that default:

- Mimic and contrib GR1T2 PickPlace now instantiate `NewtonWarpRenderer` and fail with `ValueError: Failed to load texture: ...GR1T2_albedo.<UDIM>.png`, so dataset generation never starts.
- The XR-camera and Shadow Hand preset tests resolve `default` to `NewtonWarpRendererCfg` instead of their required `IsaacRtxRendererCfg`.
- Registered-task rendering loses the Kit `RenderProduct`, and Newton returns no output for the MDL shading data types those tasks request.

Please keep the shared default unchanged and introduce a Cartpole-specific renderer preset/default, or migrate and explicitly pin every affected consumer while providing the missing Newton renderer support. The current shared change is a runtime regression outside the PR title scope.

source/isaaclab_tasks/test/core/test_environments.py
```diff
@@ -22,16 +22,18 @@
 from env_test_utils import _run_environments, setup_environment  # isort: skip
 
 
+@pytest.mark.parametrize("physics_preset_name", ["newton_mjwarp", "physx", "isaacsim_physx"])
```
AntoineRichard: This matrix does not guarantee the backend named in the test ID. Global preset resolution falls back to `default` when a task does not expose the requested name. The CI log demonstrates this directly: all three `newton_mjwarp`, `physx`, and `isaacsim_physx` Fourbar cases pass even though Fourbar only declares `default`/`newton_kamino`; the labels are therefore not evidence that those backends ran.

Please generate only supported `(task, physics preset)` pairs, or use a strict physics selector that errors when unavailable. Also restore `include_play=False` unless every Play variant is intentional: this change collected 252 cases here and 42 stage-in-memory cases, substantially increasing runtime and repeated teardown pressure.

source/isaaclab_tasks/changelog.d/default-newton-classic.minor.rst
```diff
@@ -0,0 +1,6 @@
+Changed
+^^^^^^^
+
+* **Breaking:** Changed the Cartpole task family and the default renderer preset
```
AntoineRichard: This entry explicitly marks both default changes as breaking, but I could not find a prior deprecation for either the Cartpole backend default or the shared renderer default. Please reconcile this with the repository policy that breaking changes require a deprecation first, or document why these default changes are exempt.

AntoineRichard:
I traced all seven red jobs at `543ee34`:

1. **The shared renderer-default change causes deterministic cross-task regressions.** `isaaclab_mimic` and the contrib PickPlace smoke test both fail while Newton loads the GR1T2 UDIM texture; the XR/Shadow preset assertions resolve Newton instead of Isaac RTX; and the two RL summary expectations still describe the former concrete defaults.
2. **The default-rendering baselines were not migrated.** `test_rendering_registered_tasks` has seven failures: the Cartpole stage no longer has the Kit `RenderProduct`/PhysX prims, RGB differs by 94.01%, albedo by 2.47%, and the Newton renderer produces zero outputs for the three MDL shading variants. Shadow Hand also loses its Kit `RenderProduct`. The visualizer job separately reports a stable 6.33% pixel difference for `newton-cartpole-kit-tiled`; that helper leaves the renderer implicit, so the shared default change affects a test labeled Kit.
3. **Two task files terminate natively.** `test_environments_with_stage_in_memory.py` exits 139 while starting `Isaac-Lift-Soft-Franka-Camera`; `test_environments.py` exits 139 during `Isaac-Lift-Cloth-Franka-Camera` with `newton_mjwarp`. Neither produces a Python traceback or JUnit report. These should be isolated after the renderer scope is corrected.

The assertion/image/texture failures are deterministic and will not be fixed by a rerun. I would first make the Newton renderer default Cartpole-specific (or complete the migration of every shared consumer), then update or explicitly pin the Cartpole/Kit golden tests and rerun the two crash-only cases.

source/isaaclab_tasks/test/core/test_environments.py
```diff
@@ -21,17 +21,19 @@
 # Local imports should be imported last
 from env_test_utils import _run_environments, setup_environment  # isort: skip
 
+@pytest.mark.parametrize("physics_preset_name", ["newton_mjwarp", "physx", "isaacsim_physx"])
```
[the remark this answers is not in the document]
StafaH: Fixed in 6c98c4fd6: core smoke parameters are now generated only for explicitly declared PhysicsCfg presets, so Fourbar is not mislabeled as running unsupported backends.

source/isaaclab_tasks/isaaclab_tasks/utils/presets.py
StafaH: Fixed in 6c98c4fd6: MultiBackendRendererCfg retains its Isaac RTX default. Newton renderer defaults are now scoped to the Cartpole camera configurations, preserving contrib, Mimic, XR, and Shadow Hand behavior.

source/isaaclab_tasks/test/core/test_environments.py
StafaH: Fixed in 6c98c4fd6: the matrix now includes only task/backend pairs whose raw config explicitly exposes the requested physics preset. Play filtering remains removed because the current registry has no Play task variants.

source/isaaclab_tasks/changelog.d/default-newton-classic.minor.rst
StafaH: This is an intentional 3.0 breaking default change. The fragment now accurately scopes it to Cartpole and provides the explicit-preset migration path; we are not adding deprecation artifacts for this release.

StafaH:
Addressed the deterministic regressions in 6c98c4fd6 and 1e76a2997: the generic renderer default is restored to Isaac RTX, while Newton is the default renderer only for Cartpole camera configs; Cartpole visualizer goldens explicitly pin Isaac RTX; and the smoke matrix filters to supported PhysicsCfg presets. The existing teardown release/GC fix remains in place for the native cleanup crashes. The default switch is intentional for 3.0 and documented without deprecation artifacts.

kellyguo11 (approved):

ooctipus (approved):

PR #7067, opened 2026-08-13
[Workflow] Default core tasks to RSL-RL

StafaH:
Splits the default RL-library change from #6980.\n\n- Resolves `default_agent` from task registration when `--rl_library` is omitted.\n- Registers RSL-RL as the default for core tasks, including a Pendulum RSL-RL config.\n- Adds dispatch coverage and a breaking-change fragment.\n\nValidation: `uv run python -m pytest source/isaaclab_rl/test/test_entrypoints.py -q` (11 passed, 2 skipped).\n\nSupersedes the RSL-RL portion of #6980.


source/isaaclab_tasks/isaaclab_tasks/core/pendulum/agents/rsl_rl_ppo_cfg.py
AntoineRichard, line 1: Pendulum is MARL only, will that work with RSL_RL?

source/isaaclab_tasks/isaaclab_tasks/core/pendulum/__init__.py
AntoineRichard, line 1: Not sure if we should use RSL_RL

AntoineRichard (approved):
One concern on pendulum, otherwise LGTM

ooctipus (approved):

PR #7075, opened 2026-08-13
feat: Adds an opt-in asynchronous render path (ovrtx only)

pv-nvidia:
# Description

Adds an opt-in asynchronous render path for the OVRTX renderer. Rendering then overlaps the next step's simulation and Python work. In exchange, camera outputs describe the simulation state from one step earlier.

- `OVRTXRendererCfg.async_rendering: bool` enables it. The default is `False` and renders synchronously, unchanged. There is no environment variable, the config is the single source of truth.
- Only the OVRTX legacy stage path implements it. The OVRTX ovstage path warns and renders synchronously. Async on ovstage showed no benchmark gain and moves to a follow-up PR. Its write path is untouched here.
- Asynchronous rendering supports one camera per renderer. When a second camera registers on a shared renderer, it warns and falls back to synchronous rendering.
- Measured at 2048 envs (Cartpole, Lift-KukaAllegro, Reorient-Cube-Shadow): RL iteration throughput +16-27% and `env.step()` throughput +37-82%. Disabled, throughput is unchanged within noise.

Internally, frame execution goes through a render strategy selected at construction. The synchronous strategy steps inline. The asynchronous strategy keeps one `step_async` in flight with double-buffered transform staging. The first frame of a scene is drained immediately, so the first camera read is never blank.

Testing: kitless unit tests (strategy selection, staging, teardown) and async golden-image variants of the Cartpole rendering tests in post-merge CI. The async lane is graded against the same goldens with `[sync, async]` tolerance pairs. Includes a benchmarks-guide section and a changelog fragment for `isaaclab_ov`, the only touched package.

## Type of change

## Checklist


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -899,12 +917,13 @@ def _update_transforms_legacy(self) -> None:
         if body_q is None:
             return
 
-        with self._object_xform_binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
-            ovrtx_transforms = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)
+        # The strategy yields a staging buffer and publishes it to the binding on context exit.
+        num_objects = len(self._object_newton_indices)
+        with self._strategy.stage_object_transforms(self._object_xform_binding, num_objects) as object_transforms:
             wp.launch(
```
pv-nvidia, line 1031: No stream is specified in wp.launch

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1007,26 +1026,27 @@ def _update_camera_legacy(
...
+            camera_transforms,
+        ):
+            convert_camera_frame_orientation_convention_wp(
+                src=orientations.warp,
+                dst=converted_wp,
+                origin="world",
+                target="opengl",
+                device=self._device,
+            )
+            wp.launch(
```
pv-nvidia, line 1138: No stream is specified in wp.launch

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1007,26 +1026,27 @@ def _update_camera_legacy(
...
-        if self._camera_xform_binding is not None:
-            with self._camera_xform_binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
-                wp_transforms_view = wp.from_dlpack(attr_mapping.tensor, dtype=wp.mat44d)
-                wp.copy(wp_transforms_view, camera_transforms)
+        # The strategy yields staging buffers and publishes the transforms to the binding on context exit.
+        with self._strategy.stage_camera_transforms(self._camera_xform_binding, num_envs) as (
+            converted_wp,
+            camera_transforms,
+        ):
+            convert_camera_frame_orientation_convention_wp(
```
pv-nvidia, line 1131: No stream here?

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer_strategies.py
```diff
@@ -0,0 +1,421 @@
...
+    @contextmanager
+    def stage_camera_transforms(self, binding: Any, num_rows: int) -> Iterator[tuple[wp.array, wp.array]]:
+        """Yield freshly allocated staging buffers and copy the result into ``binding`` on exit.
+
+        See :meth:`_RenderStrategy.stage_camera_transforms`.
+        """
+        camera_quats = wp.empty(num_rows, dtype=wp.quatf, device=self._device)
+        camera_transforms = wp.zeros(num_rows, dtype=wp.mat44d, device=self._device)
+        yield camera_quats, camera_transforms
+        with binding.map(device=Device.CUDA, device_id=self._device_id) as attr_mapping:
```
pv-nvidia: No need to pass a stream here?

source/isaaclab_tasks/test/core/test_rendering_tolerance_lookup.py
pv-nvidia, line 1: I think this file is major overkill, this should just be a local LLM skill during the development of this PR.

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -43,25 +44,50 @@
...
+# Entries are ``[synchronous, asynchronous]``: pipelined rendering captures a slightly different render
+# state, so it carries its own tolerance. Read via :func:`max_different_pixels_percentage_for`.
 MAX_DIFFERENT_PIXELS_PERCENTAGE_BY_ENV_NAME = {
     # RTX anti-aliasing along the ground-plane edges varies slightly across GPU and driver environments.
-    "cartpole": 1.5,
+    "cartpole": [1.5, 2.5],
     # Aliasing artifacts of shadow on the table.
-    "franka_cloth": 8.0,
-    "franka_soft": 8.0,
+    "franka_cloth": [8.0, 8.0],
```
pv-nvidia: Instead of converting all these entries to arrays, it is better to keep the literals for those that only have sync tests, and modify the MAX_DIFFERENT_PIXELS_PERCENTAGE_BY_ENV_NAME accessors to deal with array vs value

source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -43,25 +44,50 @@
...
+def max_different_pixels_percentage_for(env_name: str) -> float:
+    """Return ``env_name``'s pixel-diff tolerance for the render path this run uses.
+
+    Args:
+        env_name: Key into :data:`MAX_DIFFERENT_PIXELS_PERCENTAGE_BY_ENV_NAME`.
+
+    Returns:
+        The synchronous or asynchronous tolerance, per :func:`_async_rendering_enabled`.
+    """
+    synchronous, asynchronous = MAX_DIFFERENT_PIXELS_PERCENTAGE_BY_ENV_NAME[env_name]
```
pv-nvidia: Here we need to differentiate between array and value when accessing MAX_DIFFERENT_PIXELS_PERCENTAGE_BY_ENV_NAME[env_name]

source/isaaclab/isaaclab/renderers/renderer_cfg.py
```diff
@@ -23,3 +35,67 @@ class RendererCfg:
...
     renderer_type: str = "default"
+
+    async_rendering: bool = False
+    """Trade one frame of camera latency for pipelined rendering. Defaults to False (synchronous).
+
+    When enabled, rendering overlaps the next step's simulation and Python work, and camera
+    outputs describe the simulation state from one step earlier. Only the OVRTX renderer
+    implements this. Other renderers warn and render synchronously.
+    :data:`ASYNC_RENDERING_ENV_VAR` overrides this value.
+    """
```
huidongc, line 46: The flag is ovrtx specific, why not add it to OVRTXRendererCfg instead?

docs/source/overview/core-concepts/renderers.rst
```diff
@@ -187,6 +187,38 @@ Or install the public ``ovrtx`` package directly from PyPI:
...
    The :class:`~isaaclab.renderers.BaseRenderer` class is under active development and may change without notice.
 
+.. _renderers-async-data-flow:
+
+Asynchronous Rendering
+----------------------
+
+:attr:`~isaaclab.renderers.RendererCfg.async_rendering` trades one frame of camera latency for
+throughput. Only the OVRTX renderer implements it. This section shows how states, observations,
+actions, and rendered frames line up in both modes.
```
huidongc, line 197: The camera image is one step behind state-based observations, which seems a bit concerning from training perspective. Could this cause any training issues? I guess this is why async rendering is opt-in?

I was also wondering instead of pipelining ovrtx render to overlap with physics and training, did you experiment with a more contained approach - all write and render ops in ovrtx renderer are async, but we wait to get the current render frame to use it as observation of the current step? Was the gain of this approach so small that you abandoned it?

source/isaaclab/isaaclab/renderers/renderer_cfg.py
ooctipus: agree IsaacLab should avoid using environment variable ASYNC_RENDERING_ENV_VAR, and declare everything in config

pv-nvidia: > The flag is ovrtx specific, why not add it to OVRTXRendererCfg instead?

That's what I first did, but async rendering is a concept which could be implementated by any renderer no? It just happens we just do it first for ovrtx? But I will move it back to `OVRTXRendererCfg`, it was an LLM code review that convinced me to make it more general :wink:  

> agree IsaacLab should avoid using environment variable ASYNC_RENDERING_ENV_VAR, and declare everything in config

Doesn't this make it hard to run benchmarks with http://omniperf-dashboard.nvidia.com? @r-schmitt for vis

But I agree, we can't have an environment variable for each config, unless it is auto generated and follows some pattern, I will remove it

docs/source/overview/core-concepts/renderers.rst
pv-nvidia: Yes, indeed, training is different. That being said, many (most?) real robots have a latency between what they see and what is processed?

No I didn't try the latter, but do we have a lot of code running between these steps? The gain is already rather small with one frame latency, so I won't expect much here. But I will give a try, who knows

pv-nvidia: Without adding latency, the gain is really tiny (less than 1ms), but this still could be a good option if we want to squeeze every bit of performance. But I would like to postpone this until for a future PR, this one needs to land before next Wednesday

PR #7080, opened 2026-08-13
Restore remote asset URLs in stage dumps

pbarejko:
# Description

When `ISAAC_LAB_SAVE_STAGES` is used, then assets aren't resolved properly 

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7083, opened 2026-08-13
Restore remote asset URLs in stage dumps

mataylor-nvidia:
# Description

When `ISAAC_LAB_SAVE_STAGES` is used, assets aren't resolved properly: remote assets are handed to callers as a local cache copy, so a stage built from one records an absolute path that resolves only on the machine holding the cache. Flattening for the dump therefore writes texture and material paths that are meaningless anywhere else.

This records the source URL of each cached copy **as the copy is located**, and adds `unmirror_file_path` to look it up, so a dump can name the asset it was built from.

Recording the pair is exact where recovering it from the path is not. A directory named after a URL scheme is an ordinary local layout — `Omniverse` is where Omniverse puts user projects by default — and inferring from the path rewrites it to a URL that was never fetched:

| local path | recovered from path | this PR |
| --- | --- | --- |
| `C:/Users/user/Omniverse/MyProject/scene.usd` | `omniverse://MyProject/scene.usd` | `""` (left alone) |
| `/data/omniverse/assets/robot.usd` | `omniverse://assets/robot.usd` | `""` (left alone) |
| `/home/user/projects/https/site/logo.png` | `https://site/logo.png` | `""` (left alone) |

The map is populated even when nothing is downloaded, because retrieval walks the whole dependency tree before consulting the cache — so a warm cache still names its sources. Verified against the real DexCube asset on a fully warm cache in a fresh process: the texture is recovered as its S3 URL despite zero downloads and only the root USD being requested.

This supersedes the string-parsing approach in #7080. It is one line shorter in `assets.py` (+31/-1 vs +32/-0), adds no module-level constants, and drops the scheme allowlist, the `':'`↔`'_'` port round-trip, and the coupling to the `_mirror_path` layout.

## Commits

1. **Restore remote asset URLs in stage dumps** — the map, `unmirror_file_path`, and the rendering-test consumer.
2. **Cover Windows path handling in cached asset URL lookup** — a drive letter parses as a URL scheme, so `C:/...` must not be recorded as a cached copy; also covers USD reporting cached copies with forward slashes on Windows.
3. **Treat a Windows drive letter as a local git asset path** — pre-existing bug found while running the suite on Windows. `_is_git_remote_path` used `bool(urlparse(git_path).scheme)`, and `urlparse("C:\assets").scheme == 'c'`, so `retrieve_git_asset_path` took a local checkout for a remote repository and tried to clone it into the cache. Independent of the rest; drop this commit if you'd rather it went separately.

## Type of change

## Testing

`source/isaaclab/test/utils/test_assets.py`: **45 passed**, run per-commit in a clean worktree.

Commits 1 and 2 each carry one failure — the pre-existing Windows `retrieve_git_asset_path` bug, confirmed failing identically on pristine `develop` — which commit 3 fixes.

The drive-letter regression test was verified to fail without its guard and pass with it.

## Checklist

### DIFF
--- a/source/isaaclab_tasks/test/rendering_test_utils.py
+++ b/source/isaaclab_tasks/test/rendering_test_utils.py
@@ -22,6 +22,8 @@ from isaaclab.utils.images import make_camera_output_grid, normalize_camera_outp
 from isaaclab.utils.warp import ProxyArray
 
 if TYPE_CHECKING:
+    from pxr import Sdf
+
     from isaaclab.sensors.camera import CameraData
 
 logger = logging.getLogger(__name__)
@@ -510,6 +512,20 @@ def _sanitize_golden_stage_text(text: str) -> str:
     return text.rstrip("\n") + "\n"
 
 
+def _restore_remote_asset_paths(layer: "Sdf.Layer") -> None:
+    """Point cached asset paths in ``layer`` back at the URLs they were downloaded from.
+
+    Remote USD assets are referenced through a local cache copy, so flattening resolves the
+    textures and materials they carry into absolute cache paths that exist only on the machine
+    that ran the test. Locally authored paths are left untouched.
+    """
+    from pxr import UsdUtils  # noqa: PLC0415
+
+    from isaaclab.utils.assets import unmirror_file_path  # noqa: PLC0415
+
+    UsdUtils.ModifyAssetPaths(layer, lambda asset_path: unmirror_file_path(asset_path) or asset_path)
+
+
 def maybe_save_stage(
     test_name: str,
     physics_backend: str,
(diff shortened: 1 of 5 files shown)


source/isaaclab_tasks/test/rendering_test_utils.py
```diff
@@ -510,6 +512,20 @@ def _sanitize_golden_stage_text(text: str) -> str:
...
+def _restore_remote_asset_paths(layer: "Sdf.Layer") -> None:
+    """Point cached asset paths in ``layer`` back at the URLs they were downloaded from.
+
+    Remote USD assets are referenced through a local cache copy, so flattening resolves the
+    textures and materials they carry into absolute cache paths that exist only on the machine
+    that ran the test. Locally authored paths are left untouched.
+    """
+    from pxr import UsdUtils  # noqa: PLC0415
+
+    from isaaclab.utils.assets import unmirror_file_path  # noqa: PLC0415
```
mataylor-nvidia, line 524: should this be moved to the top?

pbarejko: We already have imports at the top of isaaclab.utils, I would assume yes? i wonder why would we want to delay this import?

pbarejko (approved):

nvsekkin (approved):

StafaH (approved):

PR #7090, opened 2026-08-14
Use clone-plan positions for OVRTX roots

ooctipus:
# Description

Make the published `ClonePlan` authoritative for OVRTX environment-root placement.

- Require the renderer to consume a published plan with positions.
- Clone prototype rows before creating destination queries.
- Author identity root transforms with translations from `ClonePlan.positions` after cloning.
- Remove homogeneous fallback-plan resolution and USD-stage transform capture/cache machinery.

This keeps placement ownership in the clone plan and allows OVRTX destinations to be absent until the renderer clones them.

## Type of change

## Screenshots

Not applicable.

## Validation

- [x] OVRTX clone-plan and renderer contract tests: 34 passed
- [x] New regression slice against unfixed `develop`: 4 failed as expected
- [x] `./isaaclab.sh -f`

## Checklist

- [x] Documentation changes are not required for this internal behavior fix
- [x] I have added a changelog fragment for `isaaclab_ov`

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -24,7 +24,6 @@ import logging
 import math
 import os
 import re
-from itertools import compress
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NoReturn, cast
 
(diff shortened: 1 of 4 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1752,7 +1636,10 @@ def _clone_sources_ovstage(self):
 
         logger.info("Cloned %d sources successfully in OVRTX", num_cloned_sources)
 
-        # Restore the pre-clone xforms.
+        env_root_xforms = np.tile(np.eye(4, dtype=np.float64), (num_envs, 1, 1))
```
rilei-nvidia, line 1639: @ooctipus - I was working on a similar PR but never got around to finishing it - I'm glad you are pushing this forward.  

What do you think about adding something like:
https://github.com/isaac-sim/IsaacLab/pull/6748/files#diff-0c7aa8d7655a343a77ae7c94ec81db7bd647d651dd27906b2180bd878cf45e3cR64

So that if we do introduce quaternions on the ClonePlan, that logic for building the env root xforms is abstracted by the ClonePlan itself?

ooctipus: I'd reserve adding quaternion to the day we really need it, I think in general clone plan shouldn't really need to remember any transformation. Each individual world start at origin

StafaH (approved):

PR #7112, opened 2026-08-15
Replace scene-data requirement objects with type tables

ooctipus:
## Summary

`SceneDataRequirement` was a dataclass carrying two booleans, wrapped in `resolve_scene_data_requirements` / `aggregate_requirements` helpers plus a pair of `SimulationContext` accessors. Every producer had to build an object, merge it, compare it, and push it back just to say "I need a USD stage" or "I need a Newton model".

- Requirements are now two public flags, `SimulationContext.requires_usd_stage` and `requires_newton_model`, that producers OR into directly.
- The per-type data lives in index-aligned `TYPES` / `REQUIRES_STAGE` / `REQUIRES_MODEL` tables in `isaaclab.scene_data.scene_data_provider`, next to the provider that consumes them, so the lookup works without the optional backend packages installed.
- `isaaclab.physics.scene_data_requirements` is deleted, along with `get_scene_data_requirements()` and `update_scene_data_requirements()`.

Net effect is 233 lines removed against 60 added, with the same observable behaviour.

### Breaking change

Read a requirement with `sim.requires_newton_model` instead of `sim.get_scene_data_requirements().requires_newton_model`, and publish one by OR-ing the flags:

```python
i = TYPES.index("newton_warp")
sim.requires_usd_stage |= REQUIRES_STAGE[i]
sim.requires_newton_model |= REQUIRES_MODEL[i]
```

Migration guidance is in the changelog fragment.

## Test plan

- [x] `ruff check` clean over `source/`
- [x] `pytest source/isaaclab/test/scene/test_interactive_scene.py source/isaaclab/test/scene_data/ source/isaaclab/test/sim/test_simulation_context_visualizers.py` — 67 passed
- [x] Changelog fragment renders via `tools/changelog/cli.py compile --dry-run`
- [ ] CI


StafaH (approved):

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
```diff
@@ -22,6 +22,13 @@
 if TYPE_CHECKING:
     from pxr import Usd
 
+# Whether each visualizer and renderer type needs a USD stage or a Newton model, by type index.
+# fmt: off
+TYPES          = ["kit", "newton_gl", "newton", "newton_rtx", "rerun", "viser", "isaac_rtx", "newton_warp", "ovrtx"]
+REQUIRES_STAGE = [True,  False,       False,    False,        False,   False,   True,        False,         True   ]
+REQUIRES_MODEL = [False, True,        True,     True,         True,    True,    False,       True,          True   ]
```
StafaH: Why don't we just make a dict with key: type, value: tuple(requires_stage, requires_model)? That would also remove the indexing issue as well?

StafaH (approved):

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
ooctipus: its doable, but the math operation could be more involved with dict  . and the table looking is quite nice I think

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
AntoineRichard, line 1: Should we ask the different backend to explicitely state if they need these? I feel like it would be better than hard-coding it here? It's annoying because we don't want to import these, but I find it odd to have these definition in the core. I'd be cleaner if they could live in the modules. But the PR is an improvement over the previous form for sure!

AntoineRichard (approved):
LGTM

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
matthewtrepte: hm yea i think the list here is long enough to where the separate boolean lists will be better off as dicts

source/isaaclab/isaaclab/scene_data/scene_data_provider.py
ooctipus: that will be another round of refactor, but I agree with the direction

matthewtrepte (approved):

PR #7118, opened 2026-08-15
Fix backend propagation in nested dictionary conversion

sylvesterkaczmarek:
# Description

`convert_dict_to_backend()` recursively processed nested dictionaries without forwarding `backend` or `array_types`. Nested values therefore silently reverted to the default NumPy backend.

This forwards the caller's conversion settings through recursion, adds regression coverage, and includes a changelog fragment.

## Type of change

- Bug fix


AntoineRichard (commented):
Hi @sylvesterkaczmarek thanks a lot for taking the time to open all these PRs. Could you kindly take the time to look into the following: 

"the new test lives in a standalone file (`test_nested_backend_conversion.py`) that duplicates the scope of the existing `source/isaaclab/test/utils/test_dict.py`, which already covers `convert_dict_to_backend`. Consider folding this as a test case into that existing file instead, for discoverability and to avoid splitting coverage for the same function across files."

LGTM otherwise!

AntoineRichard (approved):
Thanks a lot for the changes (and all the PRs) LGTM! @kellyguo11 @StafaH can I get another approval?

hujc7 (approved):

ooctipus (approved):

PR #7119, opened 2026-08-15
Fix non-finite depth display normalization

sylvesterkaczmarek:
# Description

`normalize_camera_output_for_display()` normalized depth images using the raw tensor maximum. Depth camera outputs may contain `inf` for no-hit pixels and can contain other non-finite values, so the maximum became non-finite; dividing by it could produce `NaN` pixels and suppress finite depth contrast.

This change zeroes non-finite depth values before computing the display scale, preserving finite depth normalization while keeping no-hit pixels black.

## Validation

Adds unit coverage for:
- mixed finite, `inf`, and `NaN` depth values across supported depth display types;
- all-non-finite depth input.

## Type of change

- Bug fix


AntoineRichard (approved):
Thanks a lot for the PR, and contributing to IsaacLab! @kellyguo11 @hujc7 can I have another review?

source/isaaclab/isaaclab/utils/images.py
```diff
@@ -101,6 +101,7 @@ def normalize_camera_output_for_display(tensor: torch.Tensor, data_type: str) ->
     normalized = tensor.float()
 
     if data_type in ["depth", "distance_to_camera", "distance_to_image_plane"]:
+        normalized = torch.where(torch.isfinite(normalized), normalized, torch.zeros_like(normalized))
```
hujc7: consider using  torch.nan_to_num

sylvesterkaczmarek: Thanks, good suggestion. Updated this to torch.nan_to_num(..., nan=0.0, posinf=0.0, neginf=0.0) so all non-finite depth values are explicitly zeroed before display normalization.

ooctipus (approved):

PR #7120, opened 2026-08-15
Fix add-new-robot tutorial default joint-state mutation

sylvesterkaczmarek:
# Description

The `add_new_robot.py` tutorial builds the Dofbot wave command from `default_joint_pos.torch` and then writes into that tensor. `ProxyArray.torch` is a zero-copy view, so those writes also mutate the articulation's stored default joint positions. A later reset can therefore restore the latest wave command instead of the configured initial pose.

This change clones the default joint-position tensor before applying the wave command.

## Validation

Adds a unit source guard ensuring the tutorial keeps the command buffer detached from the stored default state.

## Type of change

- Bug fix


AntoineRichard (approved):
Thanks for the fix!

hujc7 (approved):

PR #7121, opened 2026-08-15
Fix backend factory fallback before simulator initialization

sylvesterkaczmarek:
# Description

`FactoryBase._get_backend()` can be called before a `SimulationContext` has been created, but it previously dereferenced `SimulationContext.instance().physics_manager` unconditionally. `SimulationContext.instance()` returns `None` before initialization, so backend resolution could fail before any fallback was applied.

Following maintainer confirmation, the pre-context fallback is now `newton`. Initialized backend resolution remains unchanged for Newton, PhysX, and OV PhysX contexts.

A separate follow-up should update `SimulationCfg(physics=None)` to default to Newton as well; that broader configuration-default change is intentionally outside this PR.

## Validation

Adds unit coverage for backend resolution with `SimulationContext._instance` unset and verifies the `newton` fallback.

## Type of change

- Bug fix


source/isaaclab/isaaclab/utils/backend_utils.py
```diff
@@ -79,7 +79,11 @@ def _get_backend(cls, *args, **kwargs) -> str:
         # Import lazily to avoid import cycles at module load time.
         from isaaclab.sim.simulation_context import SimulationContext
 
-        manager_name = SimulationContext.instance().physics_manager.__name__.lower()
+        sim_context = SimulationContext.instance()
+        if sim_context is None:
+            return "physx"
```
AntoineRichard: should that be newton? I think our code defaults to newton now no? @StafaH for viz.

sylvesterkaczmarek: At the moment SimulationCfg.physics=None still defaults to PhysxCfg() on develop, so I kept the pre-context fallback as physx to match the current default. If the intended default is moving to Newton, happy to update this accordingly.

AntoineRichard: Checked with the team, would be great if you could update that to newton! For a follow-up `SimulationCfg(physics=None)` should default to newton too.

sylvesterkaczmarek:
Updated per the team confirmation: the pre-context backend fallback is now `newton`, with the regression test and changelog updated accordingly. I’ve kept the `SimulationCfg(physics=None)` default change separate as the requested follow-up.

ooctipus (approved):

ooctipus:
Hi @sylvesterkaczmarek  thanks for the pr, we have a test file called source/isaaclab/test/utils/test_backend_utils.py, can you put your test there

sylvesterkaczmarek:
> Hi @sylvesterkaczmarek thanks for the pr, we have a test file called source/isaaclab/test/utils/test_backend_utils.py, can you put your test there

Sure. Thanks for your guidance.

sylvesterkaczmarek:
Done. The fallback test is now in test_backend_utils.py, and the standalone test file has been removed.

PR #7122, opened 2026-08-15
Fix YAML dumps to the current working directory

sylvesterkaczmarek:
# Description

`dump_yaml()` always attempted to create `os.path.dirname(filename)`. For a filename such as `config.yaml`, the directory component is an empty string, causing `os.makedirs("")` to fail before the file can be written.

This change only creates a parent directory when the output path actually contains one. Existing nested-path behavior is unchanged.

## Validation

Adds unit coverage for:
- writing `config.yaml` directly in the current working directory;
- automatic `.yaml` extension handling without a directory component.

## Type of change

- Bug fix


AntoineRichard (approved):
Thansk for the PR!

AntoineRichard:
@StafaH @hujc7 could you review?

hujc7 (approved):

PR #7124, opened 2026-08-15
Fix dotted callable reference resolution

sylvesterkaczmarek:
# Description

`ResolvableString` supports callable references with nested attribute paths, such as `module:Outer.InnerCallable`, and exposes the dotted path through its callable metadata. However, `string_to_callable()` previously passed the entire dotted attribute path to a single `getattr()` call, so those references could not actually be resolved.

This change resolves each attribute component in sequence, allowing nested callable references to round-trip through the existing serialization and lazy-resolution APIs.

## Validation

Adds unit regression coverage for:
- serializing and resolving a nested callable;
- resolving the same reference through `ResolvableString`;
- reporting a clear `ValueError` for a missing nested attribute.

## Type of change

- Bug fix

### DIFF
--- a/docs/source/refs/contributing.rst
+++ b/docs/source/refs/contributing.rst
@@ -81,6 +81,10 @@ for maintaining the documentation.
 Sending a pull request for the documentation is the same as sending a pull request for the codebase.
 Please follow the steps mentioned in the `Contributing Code`_ section.
 
+For documentation media, only small ``.jpg`` files should be committed directly to the Isaac Lab repository.
+Upload larger images, videos, animations, and other media to an external hosting location, then link to or embed
+the externally hosted file from the documentation.
+
 .. caution::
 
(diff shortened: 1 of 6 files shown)


AntoineRichard (commented):

**Assessment:** Dotted traversal is correct, but callable_to_string adopts __qualname__ without verifying that the resulting module path resolves.

**Findings:** string.py can regress exported aliases: module:name resolves while module:Hidden.name does not; public dotted syntax is also undocumented.

**Requested changes:** Resolve the qualified candidate through the module and fall back to __name__; consolidate tests; document one dotted example.

**Test scope:** Add one exported-alias round-trip regression and consolidate the standalone cases into existing string/configclass tests.

**Documentation / AI skills:** Documentation update required: add one dotted callable example to contributing.rst. No AI-skill update is required.

PR #7125, opened 2026-08-15
Fail fast on TorchScript model load errors

sylvesterkaczmarek:
# Description

`load_torchscript_model()` currently catches every exception raised by `torch.jit.load()`, prints an error, and returns `None`. Callers then continue initialization and typically fail later with an unrelated error such as `NoneType` having no `forward` method.

This change makes model-loading failures propagate at the actual load site, while preserving the existing explicit `FileNotFoundError` for missing paths. The function therefore continues to return a valid evaluation-mode TorchScript module on success and no longer returns `None` on corrupt or incompatible models.

## Validation

Adds unit regression coverage for:
- successful TorchScript load and evaluation mode;
- missing model paths;
- invalid TorchScript files raising the underlying load error instead of returning `None`.

## Type of change

- Bug fix

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-fix-torchscript-load-errors.major.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-fix-torchscript-load-errors.major.rst
@@ -1,0 +1,6 @@
+Changed
+^^^^^^^
+
+* **Breaking:** Changed ``load_torchscript_model`` to raise ``RuntimeError`` when TorchScript loading fails instead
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** Loader raises a contextual chained RuntimeError, preserves missing-path handling, and keeps valid models in eval mode.

**Findings:** No blocking issues found.

**Test scope:** Updated real-invalid-file cases plus one exception-chaining test are distinct and sufficient.

**Documentation / AI skills:** No additional documentation update is required: public docstring and major changelog fragment document migration. No AI-skill update is required.

PR #7127, opened 2026-08-15
Fix additive bias broadcasting for multidimensional observations

sylvesterkaczmarek:
# Description

`NoiseModelWithAdditiveBias` initializes its bias as `(num_envs, 1)` and, when `sample_bias_per_component` is enabled, expands it using only the last dimension of the first input.

That works for 2D observations such as `(num_envs, features)`, but for multidimensional observations such as images or structured features, the resulting 2D bias does not broadcast against the full input shape correctly.

This change constructs the per-component bias with the full non-batch shape of the first input, so each observation component receives an independent bias and broadcasting works for arbitrary input rank.

## Validation

Adds unit regression coverage for multidimensional observation tensors, including the expected bias shape and output shape, while preserving the existing 2D behavior.

## Type of change

- Bug fix


AntoineRichard (commented):

**Assessment:** Rank-aware reshape/repeat correctly supports multidimensional observations and preserves 2D behavior.

**Findings:** Both Python files lack final newlines; new rank-1 branch is untested.

**Requested changes:** Add final newlines and one rank-1 regression assertion.

**Test scope:** Add one focused rank-1 assertion; existing 4D coverage is otherwise sufficient.

**Documentation / AI skills:** No additional documentation update is required: no public contract changed; changelog is sufficient. No AI-skill update is required.

sylvesterkaczmarek:
Addressed the latest review: added focused rank-1 regression coverage and final newlines to both Python files. The earlier multidimensional per-component concern is also addressed by the current full non-batch shape expansion.

PR #7128, opened 2026-08-16
Fix lambda serialization for comma-containing expressions

sylvesterkaczmarek:
# Description

`callable_to_string()` currently extracts lambda source by splitting the source line after `lambda` and then truncating at the first comma. That breaks valid lambda expressions whose body or defaults contain commas, for example `lambda x: (x, x + 1)` or `lambda x: [x, x + 1, x + 2]`, producing an invalid serialized callable string.

This change replaces delimiter-based extraction with AST source extraction using `inspect.getsource()`, `textwrap.dedent()`, `ast.parse()`, and `ast.get_source_segment()`. The complete lambda expression is preserved without including enclosing syntax or trailing comments.

The existing `string_to_callable()` safety validation is unchanged. Regression coverage also verifies that an unsafe serialized lambda such as attribute traversal is still rejected by the existing forbidden-node checks.

## Validation

Added `pytest.mark.unit` regression coverage for:
- tuple-valued lambdas containing commas;
- list-valued lambdas containing multiple comma-separated items;
- serialization followed by `string_to_callable()` round-trip execution;
- preservation of the existing unsafe-lambda rejection path.

Focused Python validation of the extraction and round-trip behavior passed for the covered cases.

## Type of change

- Bug fix

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-fix-lambda-serialization-commas.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-fix-lambda-serialization-commas.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed ``callable_to_string`` truncating lambda expressions at the first comma, allowing tuple and collection
+  expressions to serialize and round-trip correctly.
--- a/source/isaaclab/isaaclab/utils/string.py
+++ b/source/isaaclab/isaaclab/utils/string.py
@@ -10,6 +10,7 @@ import functools
 import importlib
 import inspect
 import re
+import textwrap
 from collections.abc import Callable, Sequence
 from typing import Any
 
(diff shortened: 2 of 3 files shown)


AntoineRichard (commented):

**Assessment:** AST extraction works for complete statements but fails when inspect.getsource returns a source fragment such as a lambda-valued dictionary entry.

**Findings:** string.py ast.parse raises SyntaxError for existing valid multiline-dictionary lambdas; the documented ValueError contract is bypassed.

**Requested changes:** Guard ast.parse and retry the fragment as an expression (or fall back to the previous extraction), normalize failures to the documented ValueError, and add one focused dictionary-literal regression in test_string.py.

**Test scope:** Replace the standalone test file with one focused regression in test_string.py covering a comma-containing lambda inside a multiline dictionary.

**Documentation / AI skills:** No additional documentation update is required: no documentation surface changed. No AI-skill update is required.

PR #7129, opened 2026-08-16
Remove stale sections from cloning guide

ooctipus:
## Summary

- remove migration guidance for the retired template-cloning API
- remove the stale `copy_from_source` note and describe `replicate_physics=False` directly
- retain the PhysX collision-filtering guidance required by direct cloning workflows

## Testing

- `uv run pre-commit run --files docs/source/how-to/cloning.rst`
- `uv run isaaclab -f`
- `uv run --isolated --with-requirements docs/requirements.txt -- make -C docs current-docs`

### DIFF
--- a/docs/source/how-to/cloning.rst
+++ b/docs/source/how-to/cloning.rst
@@ -127,8 +127,9 @@ and ``quaternions`` buffers. If ``positions`` is provided, it authors
 Clone Plans
 -----------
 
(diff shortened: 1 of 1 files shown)


docs/source/how-to/cloning.rst
AntoineRichard: From what you said this is no longer true?

AntoineRichard (commented):
Should we add the bit on the replicate_physics?

AntoineRichard (approved):

PR #7131, opened 2026-08-16
Fix CircularBuffer.buffer access before first append

sylvesterkaczmarek:
# Description

`CircularBuffer` allocates its internal storage lazily on the first `append()`. The indexing API already detects the uninitialized state and raises a clear `RuntimeError`, and the `stacked` property also handles access before initialization explicitly.

The `buffer` property did not perform the same check. Accessing `buffer.buffer` before the first append passed `None` into `torch.transpose()`, exposing an implementation-level PyTorch exception instead of a useful CircularBuffer API error.

This change adds the missing initialization guard and reuses the existing empty-buffer error message from `__getitem__()`. Behavior after the first append is unchanged for both the normal and stacked storage layouts.

## Validation

Added `pytest.mark.unit` regression coverage that verifies:
- accessing `.buffer` before the first append raises a clear `RuntimeError`;
- after the first append, `.buffer` retains the existing `[batch_size, max_length, ...]` layout and first-frame warmup behavior.

A focused behavior check of both cases passes.

## Type of change

- Bug fix


sylvesterkaczmarek:
Reopened after rebuilding the branch on current develop. The PR now contains only the intended CircularBuffer guard, focused unit tests, and changelog fragment; the unrelated changes have been removed.

AntoineRichard (commented):

**Assessment:** Guard converts opaque pre-append TypeError into the established RuntimeError without affecting initialized layouts.

**Findings:** Regression is in a new one-off test file rather than existing test_circular_buffer.py.

**Requested changes:** Consolidate into test_circular_buffer.py and run the narrow unit test.

**Test scope:** Move one focused pre-append case into the existing suite; fold post-append success into the same case if retained.

**Documentation / AI skills:** No additional documentation update is required: changelog is sufficient. No AI-skill update is required.

PR #7137, opened 2026-08-17
Angehu/perf smoke integration

angehu-nv:
# Description

> [!IMPORTANT]
> Confirm the pull request base before submitting. Target `develop` for all
> contributions. The `release/3.0.0-beta2` branch is a frozen stable landing
> snapshot and is not used for ongoing maintenance.

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


kellyguo11:
run-ci

.github/workflows/build.yaml
```diff
@@ -928,6 +928,254 @@ jobs:
...
+            -e CUDA_CACHE_PATH=/tmp/jit-cache/nv \
+            -v "${GITHUB_WORKSPACE}/jit-cache:/tmp/jit-cache" \
+            --entrypoint bash "$CI_IMAGE_TAG" -lc \
+            "set -euo pipefail
+            mkdir -p /tmp/benchmark-output
+            ./isaaclab.sh benchmark runtime \
+              --task '$BENCHMARK_TASK' \
+              --num_envs '$NUM_ENVS' \
+              --num_steps 200 \
+              --warmup_steps 100 \
```
hujc7, line 1064: Nice we have reduced steps. Any reason that we have increased warmup steps?

.github/workflows/build.yaml
```diff
@@ -928,6 +928,254 @@ jobs:
...
+        - name: shadow-camera-newton-rtx
+          task: Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct
+          num_envs: 512
+          timeout_minutes: 20
+          args: physics=newton_mjwarp renderer=isaacsim_rtx
+        - name: shadow-camera-newton-renderer
+          task: Isaac-Reorient-Cube-Shadow-Camera-Benchmark-Direct
+          num_envs: 512
+          timeout_minutes: 20
+          args: physics=newton_mjwarp renderer=newton_renderer
```
hujc7, line 988: one thing I learnt is that sometimes perf is more impl related, e.g. shadow hand use joint limit which is a gpu property for isaacsim physx but cpu property for ovphysx. So potentially, having a wider coverage might be more helpful? Just something to consider. Testing on this is also fine to help find general perf regressions.

Might also consider testing on manager envs as direct is more self-contained.

tools/perf_smoke/store.py
```diff
@@ -0,0 +1,428 @@
...
+            "contract": self.contract,
+            "contract_hash": self.contract_hash,
+            "metrics": dict(self.metrics),
+            "commit": self.commit,
+            "timestamp": self.timestamp,
+            "run_id": self.run_id,
+        }
+
+
+def parse_row(data: Any, source: str) -> BaselineRow:
```
hujc7, line 124: In general, my personal preference is to make impl more object oriented so there's explicit modeling to help with understanding the code. e.g. This function could be fold into BaselineRow.from_data(...)

.github/workflows/build.yaml
```diff
@@ -928,6 +928,254 @@ jobs:
         container-name: "isaac-lab-rendering-correctness-kitless-ovstage-test"
         omni-github-test-type: "rendering-correctness-kitless-ovstage"
 
+  performance-smoke:
+    name: "performance-smoke (${{ matrix.name }})"
+    runs-on: [self-hosted, gpu]
+    # Budget for two attempts plus image pull and cache restore.
+    # Each attempt is bounded individually by `timeout` to prevent starving retry.
+    timeout-minutes: ${{ matrix.timeout_minutes * 2 + 10 }}
```
hujc7: `[P0] Agent found bug:` arithmetic is not a GitHub Actions expression operator, so this file does not parse — the whole `Docker + Tests` workflow is dead on this branch.

Also at `:1030`, `${{ matrix.timeout_minutes * 60 }}`. actionlint on this head: `936:49: parser did not reach end of input after parsing the expression. 2 remaining token(s) in the input: "*", "INTEGER"`; the same lint on `develop` reports zero. GitHub agrees — the run for this file at `3d958c62` is [34427315986](https://github.com/isaac-sim/IsaacLab/actions/runs/34427315986): **0 jobs**, *"This run likely failed because of a workflow file issue"*.

So the `ci:run-docker` label on 09-08 produced no Docker or GPU coverage at all, and nothing in this PR has ever executed.

Fix: precompute the products per matrix row as `job_timeout_minutes` / `bench_timeout_s`.

- [ ] `actionlint` reports no `[expression]` errors, and a `ci:run-docker` label yields a `Docker + Tests` run with jobs in it.

tools/perf_smoke/contract.py
```diff
@@ -0,0 +1,216 @@
...
+#: apply. Keys are the values a bundle reports in ``run.config``, which are NOT the
+#: matrix names. An unrecognised value raises.
+#: ``version`` is a manual invalidation switch. It is part of the hashed contract, so
+#: raising it retires every baseline recorded so far, and the gate reports SKIP until
+#: history rebuilds.
+RUNTIME_COMPATIBILITY: dict[str, Any] = {
+    "version": 3,
+    "always": ("torch", "warp"),
+    "by_physics_backend": {
+        "physx": ("ovphysx", "isaacsim"),
```
hujc7: `[P0] Agent found bug:` `versions.ovphysx` is required for every `physx` run, but no bundle the CI image produces carries it.

`ovphysx` ships only in the `ov` / `ovphysx` extras — `docker/Dockerfile.base:133` `IMAGE_EXTRAS` omits both, and `isaaclab_ov` depends on `ovstage`, not `ovphysx`. `recorders/record_version_info.py:121` records it `nullable=True`, so the key arrives as `null` and `contract.py:174` raises.

All five PhysX legs then report ERROR with no metrics, permanently. Worse on `develop`: `_cmd_write` has no such guard, the exception reaches `main` → exit 2, and under `set -euo pipefail` the `Record baseline` step reddens — so no PhysX baseline is ever recorded and that half of the gate can never warm up.

```suggestion
        "physx": ("isaacsim",),
```

`isaacsim` already pins Kit's PhysX, and `schema.py:32` makes `ovphysx` its own `PhysicsBackend` literal.

Probably worth treating as a class rather than one entry: the map reads as built from `schema.Versions`' field list rather than from a bundle produced on the CI image. The review bot's `"newton": ()` finding below is the same mistake from the other side.

tools/perf_smoke/compare.py
```diff
@@ -0,0 +1,381 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+# All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Compare one runtime benchmark against the recent history of comparable runs.
+
+Pure functions only. Everything this module needs is passed in.
```
hujc7, line 8: `[P1] Agent review:` 1549 lines land with no tests, in a package documented right here as "Pure functions only".

Nothing in `compare.py`, `contract.py`, `metrics.py` or `report.py` needs a GPU, a network, or Isaac Sim. And unlike most components, a regression here does not fail CI — it stops the gate from gating, and everything stays green.

A fixture test of `contract.build` over one recorded bundle per matrix combination would have caught the `ovphysx` issue before review. Beyond that: `compare.compare` across FAIL/WARN/PASS/SKIP, `resolve_thresholds` overrides and hard-floor disabling, `store.parse_row` tolerance, and `store.read` ordering across a month boundary.

`tools/changelog/test/` is the in-repo precedent for a stdlib-only tool under `tools/`.

- [ ] `pytest tools/perf_smoke` runs in CI.

.github/workflows/build.yaml
```diff
@@ -928,6 +928,254 @@ jobs:
...
+          echo "::error::Could not resolve the warp-lang pin from pyproject.toml; the JIT cache key would stop tracking Warp"
+          exit 1
+        fi
+        echo "value=$version" >> "$GITHUB_OUTPUT"
+
+    - name: Restore warm JIT cache
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}/jit-cache
+        key: perf-jit-warp${{ steps.warp_version.outputs.value }}-${{ matrix.name }}-${{ github.run_id }}
```
hujc7, line 1028: `[P1] Agent found bug:` `github.run_id` in the key means the exact key can never hit, so `actions/cache` saves on every run — nine new PR-scoped entries per run of this workflow, which nothing will ever restore.

PR-scoped caches are readable only by that PR, so this is quota burn against the 10 GB repository limit, and eviction is LRU across the whole repo — including the shared Warp cache the test jobs restore. `warm-warp-cache`, 160 lines below in this file, exists to avoid exactly this and documents it: *"only publishes when its contents change to avoid filling the repository quota with duplicate snapshots"*, gated on `PUBLISHES_WARP_CACHE`.

Fix: switch this step to `actions/cache/restore@v4` (keys unchanged), and add an `actions/cache/save@v4` step gated on `github.event_name == 'push' && github.ref == 'refs/heads/develop'`.

tools/perf_smoke/cli.py
```diff
@@ -0,0 +1,181 @@
...
+    summary = report_mod.render_aggregate(reports)
+    print(summary, end="")
+    if args.output_markdown:
+        args.output_markdown.parent.mkdir(parents=True, exist_ok=True)
+        args.output_markdown.write_text(summary, encoding="utf-8")
+
+    if not reports:
+        # Passes: unconfigured store produces SKIP rows, non-function gate produces zero rows.
+        print("::error::perf-smoke: no comparison artifacts were produced", file=sys.stderr)
+        return 1
```
hujc7: `[P2] Agent found bug:` this makes a gate fault fail the PR, which is the one thing the design says must never happen.

The summary step runs `aggregate` under `set -euo pipefail` with no `continue-on-error`, so exit 1 reddens `performance-smoke`. But the `download-artifact` step directly above it *does* carry `continue-on-error: true` — so a transient artifact-service failure yields an empty directory and fails the PR on infrastructure rather than on performance.

```suggestion
        return 0
```

The per-leg job already fails visibly when its benchmark crashed, so the case is still reported. The comment on the line above states the opposite of what the code does; worth correcting or deleting either way.

tools/perf_smoke/metrics.py
```diff
@@ -0,0 +1,159 @@
...
+        PerfSmokeError: If any metric is missing, non-numeric, or negative.
+    """
+    values: dict[str, float] = {}
+    for metric in METRICS:
+        if metric.path is None:
+            values[metric.name] = total_startup_time_s(bundle)
+            continue
+        value = nested_number(bundle, metric.path)
+        if value is None:
+            raise PerfSmokeError(f"Benchmark result does not contain a measured value for {metric.name}")
```
hujc7, line 155: `[P2] Agent suggestion:` `extract` raises on advisory metrics too, so one advisory field can take the gating metric down with it.

`cli.py:52` runs `extract` in the block whose handler produces `errored(...)` — a report with no metrics and no contract. If a schema change ever drops or nulls `resources.ram_gb.peak`, `total_fps` stops being compared even though it measured fine.

`store.parse_row:146-149` already states the right rule for the same situation on the read path: *"A malformed advisory value must not cost us the gating one recorded beside it, which cannot be re-recorded."* Worth applying symmetrically.

Two ways in, so a fix needs both: an absent key returns `None` here, while a key present with `null` raises inside `nested_number` → `number()`. Wrapping the per-metric extraction in a `try` and re-raising only when `metric.gating` covers both.

Latent today — `capture.py:351` defaults both memory peaks to `0.0` — so this is hardening, not a live break.

tools/perf_smoke/metrics.py
```diff
@@ -0,0 +1,159 @@
...
+    higher_is_worse: bool
+    gating: bool
+
+
+#: Every metric the gate measures. Only ``total_fps`` gates today: startup time
+#: was measured varying up to 16x on JIT/shader cache warmth alone, and the memory
+#: metrics have no established noise model yet. All four are recorded so the
+#: evidence to promote them accumulates before any threshold is trusted.
+METRICS: tuple[Metric, ...] = (
+    Metric("total_fps", "Total FPS", ("runtime", "total_fps", "mean"), higher_is_worse=False, gating=True),
```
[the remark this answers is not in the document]
hujc7: `[P1] Agent review:` measuring the false-FAIL rate, since the thread is otherwise a judgement call.

Monte Carlo replicating `_evaluate`, with the measurement and the history drawn from the same distribution so every FAIL is a false positive (200k trials per cell):

| true CV | n=3 | n=5 | n=10 | n=20 |
|---|---:|---:|---:|---:|
| 1.3% — GearMesh, measured | 0.000% | 0.000% | 0.000% | 0.000% |
| 13.0% — Cartpole, measured | **15.80%** | 11.30% | 6.96% | 4.68% |

At `MIN_BASELINE_SAMPLES = 3` and Cartpole's own 13% figure from `perf_smoke_thresholds.json:7`, about one clean PR in six gets a FAIL. The `>= 2σ` guard does not absorb it: `_robust_spread` is a raw MAD scaled by the asymptotic 1.4826 with no small-sample correction, and at n=3 the MAD is biased low enough that ordinary noise clears 2σ.

GearMesh is clean at every sample count, so the machinery works. The gap is that `Metric.gating` is a property of the metric, global across tasks, while the noise that decides whether gating is safe is a property of the task × runner.

.github/workflows/build.yaml
```diff
@@ -928,6 +928,254 @@ jobs:
...
+    - name: Compare against baseline history
+      # The exit code carries the verdict to the summary job through the uploaded
+      # artifact, so it must not also redden this job: job status answers "did the
+      # benchmark run", the check answers "did performance regress".
+      continue-on-error: true
+      if: always()
+      timeout-minutes: 5
+      env:
+        # The SAS URL is strictly only read from the environment for security.
+        ISAACLAB_BLOB_URL: ${{ secrets.ISAACLAB_BLOB_URL }}
```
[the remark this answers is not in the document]
hujc7: `[P1] Agent review:` one qualification, which narrows the fix rather than the severity.

`build.yaml` triggers on `pull_request`, not `pull_request_target`, and per GitHub's docs *"with the exception of `GITHUB_TOKEN`, secrets are not passed to the runner when a workflow is triggered from a forked repository"*. So on a fork PR `secrets.ISAACLAB_BLOB_URL` resolves empty, `store.is_configured()` returns `False`, and the compare step reports SKIP without touching the network.

The exposure is same-repo branch PRs — contributors who already have push access. That still justifies a separate read-only SAS for the compare step, since the current credential can create and clobber baseline rows, but it is not fork-reachable and does not need the job relocated to base-branch code.

hujc7 (commented):
Overall I think the impl is clean but would need some iteration with CI runs. Left some comments + AI reviews.

hujc7:
run-ci

tools/perf_smoke/store.py
StafaH: I don't think class is necessary here, the functional approach seems appropriate and very readable

.github/workflows/build.yaml
angehu-nv: Good point that implementation path affects performance a lot and definitely something open to in the future. We originally picked these as stable, broadly representative tasks from the revised cotributor/core task structure. I think this is sufficient scope for an initial roll-out though and breadth/depth of coverage can be re-evaluated once this is in CI to see if there's bandwidth or use in adding more tasks.

.github/workflows/build.yaml
angehu-nv: We found that there are some amortized effects from JIT, scene, etc. that happen early on but not specific frame periods besides the known first two frame spike and decided it would be cleaner to simply discard this period. It is actually not a large cost in time to discard these since the majority of the gate run time comes from environment set up.

fatimaanes:
run-ci

StafaH (approved):

StafaH:
run-ci

fatimaanes (approved):

StafaH:
run-ci

PR #7142, opened 2026-08-17
[Fix] Automatically work around OVPhysX multi-GPU device selection

hujc7:
## 1. Summary

- Multi-GPU training that selects OVPhysX hangs on any `--num_gpus >= 2` when an RTX renderer
  shares the process. The launchers now work around it **automatically** — no flag to discover.
- Workaround only. The real fix is tracked as nvbug 6573426 and lands in **ovphysx 0.5.11**.
  This repo pins `ovphysx==0.5.10` and no published build carries the fix yet.
- Self-removing: once the pin moves past 0.5.10, delete the detection.

## 2. Root cause (external)

OVPhysX's direct OVStage attach path does not apply the requested `active_cuda_gpus` before PhysX
creates its CUDA context. With an RTX renderer in the same process a CUDA context is already
current, so PhysX auto-selects that device. Every rank above local rank 0 then gets a device-0
binding while its tensors live on its own device:

```
RuntimeError: Failed to read tensor binding:
  read_tensor_binding: device mismatch: binding expects GPU (device=0), tensor is GPU (device=2)
```

Those ranks die and rank 0 blocks in the NCCL bootstrap. Only `local_rank > 0` is affected, which
is why single-GPU runs are fine and why this long presented as an NCCL or topology problem.

## 3. Fix

When the forwarded presets select OVPhysX, the launcher passes torchrun's `--virtual_local_rank`.
That sets `LOCAL_RANK=0` for every worker and rewrites `CUDA_VISIBLE_DEVICES` so each sees only its
assigned GPU as device 0, making PhysX's default correct on every rank.

**Detection matches the physics backend alone, not an OVPhysX/OVRTX pair.** The renderer has no
stable name at launcher time: `renderer=rtx` is a placeholder the worker resolves to OVRTX only
once it knows no Kit runtime is needed. A pair match reads `{ovphysx, rtx}`, misses, and lets that
configuration hang — arm 6 below is exactly that case. The cost is engaging the workaround for
OVPhysX with a renderer that does not need it, where it is inert.

No device-selection code changed: `AppLauncher` already derives the device from `LOCAL_RANK`
*bounded by the visible device count*, which is the invariant virtual mode preserves. The
`--local_ranks_filter 0` default still works, because torchrun applies it agent-side against the
real local rank.

**There is no flag.** A temporary workaround should not add public CLI surface, because removing
that surface later needs a deprecation and migration path — for an option whose only correct value
is the one the launcher already infers. With none, retiring this is a deletion. Both launcher
preamble messages are flushed so they precede the workers' inherited-fd output in a redirected log.

## 4. Limits

- One worker per GPU.
- Every worker reports `LOCAL_RANK=0`; use the global rank to name per-rank files or logs.
  `rsl_rl` keys `init_process_group` and its process identity on `RANK`; `skrl` does not read
  `LOCAL_RANK`. The benchmark path only range-checks `0 <= local_rank < local_world_size`.
- Multi-node was not validated.
- Applied automatically whenever OVPhysX is selected; there is no per-run opt-out.

## 5. Validation

An 8-GPU node, 8x L40. IsaacLab `0469d0e5cee` + this change. ovphysx 0.5.10,
ovrtx 0.4.1.364340, ovstage 0.1.1.355824, torch 2.11.0+cu128. `Isaac-Cartpole-Camera-Direct`,
64 envs, 3 iterations. Topology: GPU0-3 one PCIe switch (NUMA 0), GPU4-7 the other (NUMA 1).

| Arm | GPUs | `CUDA_VISIBLE_DEVICES` | Link | rc | Training time | Traceback | Device mismatch |
| control, workaround disabled [^1] | 4 | 0,1,2,3 | PIX | **1** | **none** | 4 | **3** |
| auto | 4 | 0,1,2,3 | PIX | 0 | 15.61 s | 0 | 0 |
| auto, reversed | 4 | 7,6,5,4 | PIX | 0 | 14.84 s | 0 | 0 |
| auto, cross-switch | 4 | 0,1,4,5 | SYS | 0 | 9.53 s | 0 | 0 |
| auto, cross-switch reversed | 4 | 5,4,1,0 | SYS | 0 | 9.85 s | 0 | 0 |
| `physics=ovphysx renderer=rtx` | 4 | 0,1,2,3 | PIX | 0 | 8.86 s | 0 | 0 |
| auto | 8 | all | both | 0 | 125.06 s | 0 | 0 |

The control failed with three `device mismatch` errors on exactly local_rank 1, 2 and 3, and used
the same renderer as the passing arms — the only variable between them is the workaround. Arms 2-7
pass no virtual-local-rank flag; detection fired on its own each time. Arm 6's banner reports
`Renderer rtx (ovrtx)` with zero Isaac Sim in the process, confirming the pair-match gap is real.

A sampled re-run confirmed physical placement rather than inferring it: `7,6,5,4` occupied
GPUs 4-7 and `5,4,1,0` occupied GPUs 0,1,4,5.

[^1]: The control was produced with an opt-out flag that this PR's final revision removes, so it is
not reproducible from the shipped CLI. The failure it demonstrates reproduces on any pre-change build.

## 6. Test

`test_virtual_local_rank_follows_the_ovphysx_preset` is parametrized over the affected and
unaffected preset combinations, including `physics=ovphysx renderer=rtx`, and asserts the preset
tokens still reach the worker verbatim. Companion tests cover the bare-flag forwarding to torchrun
and confirm the launcher exposes no opt-in option of its own.


source/isaaclab/isaaclab/cli/multigpu.py
```diff
@@ -280,6 +290,28 @@ def _is_skrl_jax_launcher(args_cli: argparse.Namespace, worker_args: list[str],
     return _forwarded_arg_value(worker_args, "--ml_framework") == "jax"
 
 
+def _use_virtual_local_rank(worker_args: list[str]) -> bool:
+    """Return whether the presets select OVPhysX, which needs one visible GPU per worker.
+
+    Works around nvbug 6573426 in ovphysx <= 0.5.10; delete with the pin. Matches the physics backend
+    only, because ``renderer=rtx`` resolves to OVRTX inside the worker, and matches the raw name, so a
+    future ``ovphysx`` alias would bypass it.
+    """
```
AntoineRichard, line 299:
```suggestion
#TODO delete when OvPhysX's pin changes
def _use_virtual_local_rank(worker_args: list[str]) -> bool:
    """Return whether the presets select OVPhysX, which needs one visible GPU per worker.

    Works around nvbug 6573426 in ovphysx <= 0.5.10; delete with the pin. Matches the physics backend
    only, because ``renderer=rtx`` resolves to OVRTX inside the worker, and matches the raw name, so a
    future ``ovphysx`` alias would bypass it.
    """
```

AntoineRichard (approved):

marcodiiga (commented):
Before merging these changes, please let's coordinate on next steps.

AntoineRichard (changes requested):
Not merging, OvPhysX 0.5.11 has the fixes for this.

hujc7:
ovphysx 5.11 released

PR #7144, opened 2026-08-18
[CI] Retry transient Python package downloads

StafaH:
## Description

Retries CI Python package installs up to three times with a three-second delay to tolerate transient PyPI connectivity failures.

## Validation

- `uv run isaaclab -f`
- Shell syntax and GitHub workflow YAML validation


nvsekkin:
overall the logic is clean and the separate file makes it easy to update if we needed to.

.github/actions/_lib/retry-command.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -u
+
+if [ "$#" -eq 0 ]; then
+  echo "Usage: $0 <command> [args...]" >&2
+  exit 2
+fi
+
+readonly max_attempts=3
+readonly retry_delay_seconds=3
```
nvsekkin: @nv-apoddubny had mentioned adding a delay, if `readonly retry_delay_seconds=3` is good with him I do not have any objections

nvsekkin:
should we add the same retry to build.sh @StafaH ? https://github.com/isaac-sim/IsaacLab/actions/runs/32083873111/job/95552207335?pr=7144#step:5:77 - I see that this guy failed with a similar problem

nvsekkin (approved):

hujc7 (approved):

AntoineRichard (dismissed):
## AI Review (Codex)

This review was generated by OpenAI Codex and verified against PR head 1f36f71601de54bb2a5f625ce9b93060f81861fb.

### Important findings

1. **The CI helper does not retry failed install commands.**

   .github/actions/_lib/with-python-package-retries.sh:14-20 only sets PIP_RETRIES and UV_HTTP_RETRIES before executing the command once. This does not match the PR description of up to three command attempts. The current isaaclab_newton and rendering-correctness jobs demonstrate the gap: each pip process exhausted all 12 internal HTTP retries on repeated files.pythonhosted.org 502 responses, then the job failed without another command attempt.

   Please define a bounded total retry budget. One option is a command-level loop with fewer internal HTTP retries; simply nesting three attempts around 12 HTTP retries could make one package failure take roughly 24 minutes. A CI wheel cache or wheelhouse would be more robust for these bootstrap dependencies.

2. **The PR removes PhysX launch coverage for two Franka-soft tasks.**

   source/isaaclab_tasks/test/core/test_environments_isaacsim_physx.py:28-29 adds Isaac-Lift-Soft-Franka and Isaac-Lift-Soft-Franka-Camera to _COVERED_TASKS. However, the referenced rendering path constructs only FrankaSoftCameraEnvCfg and explicitly skips physics_backend == physx. The non-camera task has no replacement test, so neither exclusion is actually covered for this PhysX path.

   Please remove these unrelated exclusions, or add targeted replacement coverage. If a known teardown issue requires suppression, use an explicit skip with the bug reference rather than marking the tasks as covered.

### CI diagnosis

- isaaclab_newton: pytest-mock metadata download failed after persistent HTTP 502 responses; tests never started.
- rendering-correctness: pytetwild metadata download failed after persistent HTTP 502 responses; tests never started.
- record-video: four tests passed; test_multiple_recorders_simultaneous narrowly missed the motion threshold (0.096 < 0.1). This test is unchanged by the PR and appears flaky/unrelated.

### Verification

- Focused CLI tests: 78 passed, 1 skipped.
- ./isaaclab.sh -f: all hooks passed.
- Changed shell scripts passed bash syntax validation.
- git diff --check passed.

GPU and rendering tests were not reproduced locally.

kellyguo11 (approved):

PR #7146, opened 2026-08-18
[Workflow] Use Newton 1.5.1 PyPI release

StafaH:
## Description

Updates the authoritative Newton pin from the `release-1.5` Git branch to the published `newton[sim]==1.5.1` PyPI release.

This also regenerates the wheel-builder and install-CI overrides and `uv.lock`, and updates the install CLI to detect and report the pinned package release instead of a Git commit.

## Type of change

- Workflow/dependency update (non-breaking)

## Release backport

## Screenshots

Not applicable.

## Checklist

- [x] Documentation is not affected; a changelog fragment is included
- [x] I have run the targeted install and pin-consistency tests
- [x] Contributor metadata is already represented in the repository revision history

## Validation

- `uv run python -m pytest source/isaaclab/test/cli/test_install.py -k EnsureNewton source/isaaclab/test/cli/test_uv_run_pyproject.py -k "EnsureNewton or version_single_source"` (4 passed)
- `uv run python tools/changelog/cli.py check develop --include-worktree`
- `uv run isaaclab -f`
- `uv lock --check`


AntoineRichard (approved):

ooctipus (approved):

pyproject.toml
```diff
@@ -214,7 +214,7 @@ torchaudio = "2.11.0"
 ovphysx = "0.5.10"
 ovrtx = "0.4.1.364340"
 ovstage = "0.1.1.355824"
-newton = "release-1.5"
+newton = "1.5.0"
```
hujc7: I think that's one of my confusions. Is there a way to define it only once?

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -350,35 +350,35 @@ def _ensure_cuda_torch() -> None:
...
     pip_cmd = get_pip_command(python_exe)
     using_uv = pip_cmd[0] == "uv"
 
-    # git installs record the commit in freeze output; skip if it is already present.
+    # Exact index installs use ``name==version`` in freeze output; skip if it is already present.
     frozen = run_command(pip_cmd + ["freeze"], capture_output=True, text=True, check=False)
     if frozen.returncode == 0 and any(
-        _requirement_name(line) == "newton" and commit in line for line in frozen.stdout.splitlines()
+        line.strip().lower() == f"newton=={version}" for line in frozen.stdout.splitlines()
     ):
```
hujc7, line 378: [P2] Agent suggestion: consider making this skip-check tolerant of a git/SHA pin — it is correct for `==1.5.0`, but silently degrades for any other pin shape.

Worth saying first that this line *fixes* a real pre-existing bug: with the old `release-1.5` branch pin the check could never match, because `pip freeze` records the resolved commit, never the ref name — so `isaaclab -i` uninstalled and reinstalled Newton from git on every run.

| freeze line | pin | before | this PR |
|---|---|---|---|
| `newton @ git+…@cca3bb8…` | `release-1.5` | reinstall | reinstall |
| `newton @ git+…@cca3bb8…` | `cca3bb8…` (full SHA) | skip | **reinstall** |
| `newton==1.5.0` | `1.5.0` | skip | skip |

The narrowing is row 2: a full-SHA pin used to be handled and now is not. If Newton ever needs a SHA pin again — a hotfix picked up between releases — `version` becomes a SHA, `newton==<sha>` matches no freeze line, and the CLI falls back to uninstall+reinstall on every invocation. No error, just churn, which is the failure mode hardest to notice.

Accepting both shapes keeps the anchored range the only thing that changes:

```suggestion
    if frozen.returncode == 0 and any(
        line.strip().lower() == f"newton=={version}" or line.strip().lower().endswith(f"@{version}")
        for line in frozen.stdout.splitlines()
        if _requirement_name(line) == "newton"
    ):
```

<details><summary>Behaviour of the suggested predicate across six freeze/pin combinations</summary>

| installed | pin | result |
|---|---|---|
| `newton==1.5.0` | `1.5.0` | skip |
| `newton @ git+…@<sha>` | full SHA | skip |
| `newton==1.2.0` (Isaac Sim's bundled) | `1.5.0` | install |
| `newton @ git+…@<sha>` | `1.5.0` | install |
| `newton==11.5.0` | `1.5.0` | install (no substring false-positive) |
| `newton-usd-schemas==1.5.0` only | `1.5.0` | install (wrong package ignored) |

</details>

- [ ] Either accept both pin shapes, or state explicitly that a non-registry pin is unsupported.

source/isaaclab/test/cli/test_uv_run_pyproject.py
```diff
@@ -173,10 +173,10 @@ def spec(package: str) -> str:
     for package in ("torch", "torchvision", "torchaudio"):
         assert f"{package}=={versions[package]}" in overrides
 
-    # Newton is pinned to a git ref (branch/tag/commit) via a uv override; warp-lang is a
-    # core dependency whose table value may be an exact pin ("1.2.3" -> ``==``) or a range
+    # Newton is pinned to an exact PyPI release via a uv override; warp-lang is a core
+    # dependency whose table value may be an exact pin ("1.2.3" -> ``==``) or a range
     # (">=1.2.3" -> mirrored verbatim).
-    assert any(dep.endswith(f"newton.git@{versions['newton']}") for dep in overrides)
+    assert f"newton[sim]=={versions['newton']}" in overrides
```
hujc7: [P3] Agent suggestion: this assertion hard-codes the registry pin shape, while the `warp-lang` assertion two lines below deliberately accepts either. Consider giving Newton the same treatment.

As written, returning Newton to a git pin fails this test, so a pin change and a test edit have to land together. The adjacent lines already show the preferred pattern:

```python
warp_spec = f"warp-lang=={warp_value}" if warp_value[0].isdigit() else f"warp-lang{warp_value}"
```

This assertion is also what keeps `[tool.isaaclab.versions].newton` and the `[tool.uv]` override from drifting — `_ensure_newton` installs the override string but compares against the versions-table value, so if the two ever disagree the CLI installs one version and compares against another, reinstalling forever. That makes it load-bearing, and worth having it survive a pin-shape change rather than block one.

- [ ] Accept either `newton[sim]==<version>` or `newton[sim] @ git+…@<ref>`, keyed off the table value the way `warp_spec` is.

hujc7 (approved):

StafaH:
run-ci

kellyguo11:
run-ci

PR #7149, opened 2026-08-18
Fix CUDA 719 crash from unpumped Kit marker backend on non-Kit visualizers

AntoineRichard:
# Description

Fixes a crash when using the Newton GL visualizer on a PhysX-backed task with debug-vis markers enabled: `env.step()` crashes on the very first call with `CUDA error: unspecified launch failure` (719) from PhysX's `GpuArticulationView`, preceded by `FabricManager::initializePointInstancer mismatched prototypes` warnings for the active debug-vis markers.

## Root cause

`VisualizationMarkers._ensure_backends_initialized()` stood up a Kit/USD marker backend whenever `sim.is_rendering` was `True` — which is also true for non-Kit visualizers like `newton_gl`, since that flag is set by any active visualizer, not just Kit. With no real Kit visualizer instance to pump `app.update()` every frame, the Kit backend's raw USD marker writes (position/prototype-index attrs) were never digested by Fabric, desyncing its point-instancer prototype table. That desync corrupts shared Fabric/GPU-tensor interop state, and the very next PhysX GPU articulation step then crashes with CUDA error 719.

Reproduced on an independent machine/GPU (Linux, RTX 5000 Ada / sm_89) — different architecture than the originally reported RTX 5090 (sm_120) — with the identical crash signature and warning sequence, confirming this is architecture-independent. Confirmed causally: disabling the task's debug-vis markers eliminated the crash entirely.

## Fix

`_ensure_backends_initialized()` now checks the concrete conditions that actually need a Kit-pumped marker backend (native GUI, RTX sensor rendering, XR, offscreen capture) plus the existing Kit-pumping-visualizer check, instead of the broader `sim.is_rendering`.

Fixes #(issue)

## Type of change

## Checklist


matthewtrepte (approved):
approved, thanks!

kellyguo11 (approved):

PR #7005, opened 2026-08-10
Test crash journal

mataylor-nvidia:
# Description

Currently when a test crashes unexpectedly all that is surfaced is 

`setup::copy_failed`  or `test_execution`

> note: `setup::copy_failed` maybe infra problem

and we lose context of what actually passed, failed and which test causes the crash

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7152, opened 2026-08-18
Dump thread stacks before killing a hung test

mataylor-nvidia:
# Description

A test that **crashes** reports a traceback, because `PYTHONFAULTHANDLER=1` installs `faulthandler` for `SIGSEGV` and friends. A test that **hangs** reports nothing.

Hang *detection* already works — `tools/conftest.py` catches three kinds:

| Kind | Trigger | Constant |
|---|---|---|
| `startup_hang` | no `AppLauncher initialization complete` / `collected ` marker | `STARTUP_DEADLINE = 120` s |
| `timeout` | wall clock exceeds the per-file budget | `DEFAULT_TIMEOUT = 1000` s |
| `shutdown_hang` | JUnit report written, process still alive | `SHUTDOWN_GRACE_PERIOD = 30` s |

The problem is what happens next: all three escalate straight to `os.killpg(pgid, SIGKILL)`. `SIGKILL` cannot be caught, so nothing gets a chance to dump. The report carries `nvidia-smi`, `ps auxf`, `dmesg` and the last `-v` test name — nothing pointing at the hung code.

The repo already records the symptom in a skip reason: *"Native hang: the per-file CI runner kills the suite after 1000s with no pytest outcome"* (`source/isaaclab_tasks/test/rendering_test_utils.py`).

This complements the crash journal (#7005): that recovers **which tests** had passed when a process died; this reports **where the process is stuck**.

## Change

The runner asks the process where it is stuck before killing it.

- **`tools/hang_dump.py`** (new) — pytest plugin registering `SIGUSR1` via `faulthandler.register()`, writing to a file named by `ISAACLAB_HANG_DUMP`, which the runner sets per test file and clears per attempt, mirroring the crash journal's `ISAACLAB_TEST_JOURNAL`. No-ops where the signal does not exist.

  The dump has to go to a **file**, not stderr. pytest captures at the file-descriptor level, so it has already redirected fd 2 by the time the plugin loads; a dump written there is discarded when the process is `SIGKILL`ed — the only case it is ever written in. The first revision of this PR wrote to `sys.__stderr__` and produced no dump in CI at all, which #7153 caught. `tools/ovrtx_log.py` keeps the renderer log in a file for the same reason.
- **`conftest.py`** — loads it via `pytest_plugins`, covering every suite.
- **`tools/conftest.py`** — `_dump_hung_process_stacks()` signals the process and drains its output before the existing `SIGKILL`, prepending the result to `pre_kill_diag`. The fd-drain block was extracted into `_drain_ready_output()` so the watchdog loop and the dump path share one implementation. `pre_kill_diag` is now also threaded into `_make_missing_report_result`, so a fresh-process retry that hangs reports its stack too.

The dump is taken twice — identical stacks seconds apart are what distinguish a wedged process from a slow one.

Report plumbing is otherwise unchanged: `pre_kill_diag` already flows into the `startup_hang` and `timeout` reports and the retry warnings, and the drain echoes to stdout/stderr, so the stack also streams live to the job log. Prepending rather than appending matters — `_get_diagnostics` truncates with `diag[:10000]`, so the stack survives and the system tables get trimmed instead.

## Artifact

The dump reached CI only through `pre_kill_diag`, which `_get_diagnostics` truncates to 10 000 characters — and a Kit process has enough threads to exceed that, so the part worth reading was the part being cut. The dumps are now written to `tests/hang-dumps/` and collected as a `hang-dumps-<container>` artifact, alongside the existing `comparison-images` and `ovrtx-logs` ones. The reports and the job log still carry the (truncated) dump; the artifact is the whole thing. It is absent unless something hung, which `if-no-files-found: ignore` already covers.

## Why `SIGUSR1`

`SIGTERM` and `SIGABRT` are unusable here. `AppLauncher` binds both to `_on_abort_signal`, which calls `SimulationApp.close()` — itself what a shutdown hang is stuck inside — so either would re-enter the hang. Binding `SIGABRT` also displaces `faulthandler`'s own handler.

A Python-level `signal` handler would not run regardless: those execute between bytecodes, and a thread wedged in a native Kit, CUDA, or renderer call never returns to the interpreter loop. `isaaclab.cli.multigpu` documents the same constraint when reaping stragglers.

`faulthandler.register()` installs a C-level handler that walks every thread from inside the signal handler, so it reports a process whose GIL will never be released. `SIGUSR1` is unused anywhere in `source/`, `tools/`, `scripts/`, `.github/`.

## Sample output

Against a process blocked in `threading.Event().wait()`:

```
=== HANG STACK DUMP (all threads) ===
----- dump 1 of 2 -----
Current thread 0x000073f05cc49080 (most recent call first):
  File "/usr/lib/python3.12/threading.py", line 355 in wait
  File "<string>", line 8 in wedged_call
```

## Type of change

## Release backport

## Screenshots

Not applicable.

## Checklist

## Testing

One regression test added to `test_test_orchestrator_result_handling.py`, exercising the **real** `capture_test_output_with_timeout` against a genuinely hung child. It asserts every property of the dump — the stack names the hung call, more than one dump is taken, and the stack precedes the system tables — because each is a facet of the same `pre_kill_diag` and each costs a real timeout to reproduce separately. Confirmed it fails without the change on the meaningful assertion (`assert 'HANG STACK DUMP' in ''`), not on a missing constant.

It skips on Windows, where the orchestrator's process handling (`select` on pipes, `os.killpg`, `start_new_session`) is unavailable.

### CI cost

The three hang tests each waited out a 15 s timeout, costing the Isaac Sim suite about a minute for coverage that needs no GPU. `test_test_orchestrator_result_handling.py` is now listed in `TESTS_TO_SKIP`, so the whole file is out of CI; the skip entry records the command to run it on demand when changing the orchestrator.

Note this also takes the file's pre-existing tests (crash-journal blaming, retry semantics, renderer-log bounds, result summary) out of CI.

The hang tests are also folded into one, so a local run is not a minute either. The remaining test still has to sit through a real timeout, so that timeout is sized to what the child actually needs: measured ~1.4 s idle and ~2.6 s with eight spawned at once, against a budget of 8 s. Whole file measured at 15 passed, 1 skipped in 2.9 s on Windows.

#7153 is a throwaway probe branched off this one, wedging a rendering correctness test so CI exercises the dump against a live Kit process with all its threads running.

## Scope

Deliberately excluded:

- **No native C++ frames** (no `py-spy`/`gdb`). Python stacks stop at the C boundary, so a Kit shutdown hang reads as `_close_app` → `SimulationApp.close()` without naming the RTX/PhysX call. Still localizes the hang to a test and a call site.
- **No change to pass/fail classification** — `passed (shutdown hanged)` stays a pass, so nothing goes red as a side effect.
- **Still uncovered:** `docker wait` in `run_tests.sh` has no deadline, so a container hanging *above* pytest is caught only by the 180-minute job timeout.


AntoineRichard:

The mechanism is empirically proven — the #7153 probe's second run (head `774826be`) shows the wedged Kit process answering SIGUSR1 with both dump passes, naming the hung call site through the full pytest/flaky/pluggy chain — and the failure paths are well-bounded (the dump loop is hard-capped at 6 s, a dead child breaks out, the pipe is drained so the handler can't wedge, and the group-level SIGKILL still runs unconditionally).

Asks before merge:
1. **Trivial conflict:** one hunk in root `conftest.py` against #7305 (the `try: import warp` block landed just above the `pytest_plugins` line this PR edits).
2. **The 10,000-char report cap eats the payload** — confirmed in the probe log itself: `_get_diagnostics` truncates at 10 k (plus two inline copies), so the 64 KiB `HANG_DUMP_LIMIT_BYTES` is effectively dead code; the real run truncated mid-dump-2 and dropped `nvidia-smi`/`ps`/`dmesg` from the report entirely. Budget the dump or raise the downstream cap. (Note the PR description's claim that the drain echoes the dump live to the job log is stale since the fd→file fix — the truncated `logger.info` is now the only path it reaches a human.) An alternative: upload `tests/test-hangdump-*.log` as an artifact (natural fit with #7228's plumbing), which preserves the full dump out-of-band.
3. **`shutdown_hang` silently discards the dump** — when the report file exists, `_run_one_pass` neither logs nor attaches `pre_kill_diag`, so the 6 s is spent and thrown away on the PR's own motivating case (`SimulationApp.close()` re-entrancy). A `logger.info(pre_kill_diag)` on that branch closes it.

Two small notes: the probe showed only Python threads in the dump (faulthandler can't walk Kit's native threads), so the 'all threads' docstring over-promises — 'all Python threads' is accurate; and please preserve the probe's run-2 log excerpt here before #7153 is closed, since the expiring CI log is currently the only record of the successful validation.

pbarejko (approved):

source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
```diff
@@ -9,17 +9,31 @@
```
StafaH, line 9: Looks like these tests will add a minute to the CI time? 15 second timeouts, multiple hanging tests? This might be too expensive. A typical pytest that has a small amount of code runs in ms

.github/actions/run-tests/run_tests.sh
```diff
@@ -470,6 +470,17 @@ run_tests() {
   elif docker cp "$container_name:/workspace/isaaclab/tests/ovrtx-logs" "$ovrtx_dir" 2>/dev/null; then
     echo "🟢 OVRTX renderer logs copied from container to $ovrtx_dir"
   fi
+
+  # Copy per-file thread stack dumps (written by tools/hang_dump.py when the runner kills a hung test).
+  # The reports carry these truncated to 10 000 chars; the whole dump is only here. Absent unless
+  # something hung, which is the normal case.
+  local hang_dir="$reports_dir/hang-dumps"
+  if [ -n "$volume_mount_source" ] && [ -d "${volume_mount_source}/tests/hang-dumps" ]; then
+    cp -r "${volume_mount_source}/tests/hang-dumps" "$hang_dir"
```
mataylor-nvidia, line 479: Upload the hang dump. This way it is not cut off

mataylor-nvidia:
run-ci

StafaH (approved):

source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
mataylor-nvidia: I have left these to be run locally only. In the future we can re-visit this

kellyguo11 (approved):

mataylor-nvidia:
run-ci

PR #7153, opened 2026-08-18
[DO NOT MERGE] Probe hang stack dump against a live Kit process

mataylor-nvidia:
# Description

**Do not merge.** Throwaway CI probe for #7152, branched off it. Revert by deleting the marked block and the `_probe_wedge()` call.

## What this proves

#7152 adds three regression tests that spawn a genuinely hung child and assert the stack dump appears. Those cover the *mechanism*, but the child is a toy `python` process with one thread.

What they cannot cover is the part that actually matters here: **does a real Isaac Sim process answer `SIGUSR1`?** Kit runs dozens of native threads, installs its own signal handling, and parks the main thread inside C++ calls that never return to the interpreter loop. If Kit blocked or claimed `SIGUSR1`, the feature would pass CI and still be useless on the hangs it was built for.

This probe wedges `test_rendering_cartpole.py` *after* its first case has really rendered — so Kit is fully initialized, has produced actual frames, and all its threads are live at the moment the signal arrives.

Note the injection is a **hang**, not a kill. Killing Kit produces a crash, and crashes already printed stacks before #7152 — `PYTHONFAULTHANDLER=1` covers `SIGSEGV`/`SIGABRT`. The gap being closed is specifically hangs.

## Expected result

The `rendering-correctness` job should show `test_rendering_cartpole.py` killed for `timeout`, with both the job log and the JUnit report carrying:

```
=== HANG STACK DUMP (all threads) ===
----- dump 1 of 2 -----
Current thread 0x... (most recent call first):
  File ".../threading.py", line 355 in wait
  File ".../test_rendering_cartpole.py", line ... in _probe_wedge
  File ".../test_rendering_cartpole.py", line ... in test_rendering_cartpole
  ...
Thread 0x... (most recent call first):
  <Kit's own threads>
```

Two things to check:

1. **`_probe_wedge` is named**, and the frames below it are the real pytest/Kit call chain — the diagnostic that did not exist before.
2. **Kit's other threads appear** as additional `Thread 0x...` blocks. That is the evidence `faulthandler.register` walks a live Kit process, not just the main thread.

The two dumps should be identical, which is the intended "wedged, not slow" signal.

## Cost and blast radius

`test_rendering_cartpole.py` carries no `PER_TEST_TIMEOUTS` entry, so it runs on `DEFAULT_TIMEOUT` (1000 s) plus the 700 s cold-cache buffer it earns as the first `enable_cameras=True` file — roughly **28 minutes** before the dump fires, against the job's 120-minute budget. Left at the defaults deliberately so the probe exercises the real configuration.

The orchestrator runs each file in its own process, so only `test_rendering_cartpole.py` is affected; the other six rendering files in the job run normally afterward. `test-rendering-correctness` is `continue-on-error` on pull requests, so a red result here does not block.

## Before

The same hang on `develop` produces:

```
=== SYSTEM DIAGNOSTICS ===
--- nvidia-smi ---
...
--- process tree (ps auxf) ---
```

No stack, anywhere. That is the gap #7152 closes.

## Type of change

## Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -19,7 +19,8 @@ and registered in the repo-root ``pyproject.toml``. Select them with the standar
 e.g. ``pytest -m unit source/isaaclab/test`` or ``pytest -m "not unit" source/isaaclab/test``.
 
 Also loads ``tools/ovrtx_log.py``, which replays the OVRTX renderer log per test, so every suite that
-builds a renderer reports what it logged the same way.
+builds a renderer reports what it logged the same way, and ``tools/hang_dump.py``, which lets the CI
+runner ask this process for a stack dump before it kills it for hanging.
 """
 
 from __future__ import annotations
@@ -27,7 +28,7 @@ from __future__ import annotations
 import json
 import os
 
-pytest_plugins = ["tools.ovrtx_log"]
+pytest_plugins = ["tools.ovrtx_log", "tools.hang_dump"]
 
 JOURNAL_ENV_VAR = "ISAACLAB_TEST_JOURNAL"
 """Environment variable naming the crash-journal file. Unset (the default) disables journaling."""
--- a/source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
+++ b/source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
@@ -9,6 +9,8 @@ from __future__ import annotations
 
 import importlib.util
 import json
+import os
+import signal
 import sys
 import xml.etree.ElementTree as ElementTree
 from pathlib import Path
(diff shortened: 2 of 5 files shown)


mataylor-nvidia:
## Run 1 result: the probe worked, and it found a real bug

The wedge landed exactly where intended, and the hang was detected on schedule — but **no stack dump was produced**:

```
22:41:24  collected 33 items
22:41:48  test_rendering_cartpole[physx-isaacsim_rtx-rgb] PASSED
23:05:02  test_rendering_cartpole[physx-isaacsim_rtx-albedo]      <- wedged mid-case
          Test .../test_rendering_cartpole.py timed out after 1700 seconds...
          --- nvidia-smi ---
```

`nvidia-smi` follows the timeout line directly, so `pre_kill_diag` was empty.

### Root cause

pytest captures at the **file-descriptor** level. It `dup2`s fd 2 to a temp file, so
`faulthandler.register(file=sys.__stderr__)` — which stores `fileno()` = 2 — wrote the dump into pytest's
capture buffer, which is discarded when the process is `SIGKILL`ed. That is the only case the dump is ever
written in.

The decisive evidence is this probe's own `[CI PROBE] wedging` marker: a plain `print(..., file=sys.__stderr__)`
with no signal involved, and it never reached the log either. So fd 2 is captured regardless of the dump path.

The regression tests in #7152 missed this because they hung a bare `python script.py` child, which has no
capture, so the dump reached stderr and they passed.

### Fix

#7152 now writes the dump to a file named by `ISAACLAB_HANG_DUMP`, set per test file and cleared per attempt,
mirroring the crash journal's `ISAACLAB_TEST_JOURNAL` — and for the same reason `tools/ovrtx_log.py` keeps the
renderer log in a file. Its tests now hang a real `python -m pytest` child; against the previous
implementation all three fail on `assert 'HANG STACK DUMP' in ''`.

Also added the `isaaclab_tasks` changelog fragment this branch was missing.

Re-running to confirm the dump appears.

### Worth noting

Two things did work on run 1: the timeout fired at exactly 1700 s (1000 `DEFAULT_TIMEOUT` + 700 cold-cache
buffer) as predicted, and the crash journal recovered all 33 verdicts from the killed process
(`recovered 33 test result(s) from the crash journal`).

PR #7157, opened 2026-08-19
Remove ovstage host copies

rilei-nvidia:
# Description

ovstage 0.1.1 fixes nvbug 6490020 so the OVRTX ovstage path no longer needs its per-frame host round-trip

- Write transforms and points straight from their Warp GPU buffers; make_dltensor now folds a producer's trailing axes into the lane count omni:xform and points expect.
- Order those writes with write_attribute(cuda_stream=...) instead of blocking the host on wp.synchronize_device, as the legacy binding path already does.
- Drop the _OVSTAGE_AVAILABLE guard: ovstage is an unconditional dependency of isaaclab_ov, so the fallback was unreachable.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
@@ -33,21 +33,11 @@ from typing import TYPE_CHECKING, Any, NoReturn, cast
 logger = logging.getLogger(__name__)
 
 import numpy as np
+import ovstage
 import torch
 import warp as wp
 
 import isaaclab.utils.warp  # noqa: F401  # initializes Warp runtime
-
-# ovstage is optional: when present the renderer uses the split-ownership model
-# (ovstage owns scene data, ovrtx owns rendering). When absent it falls back to
-# the legacy renderer-owned scene APIs (deprecated in ovrtx 0.4).
-_OVSTAGE_AVAILABLE = False
-try:
-    import ovstage
-
-    _OVSTAGE_AVAILABLE = True
-except ModuleNotFoundError:
-    pass
 
 # The ovrtx C library links to its own version of the USD libraries. Having
 # the pxr Python package available can cause the C library to load an
@@ -80,6 +70,13 @@ from isaaclab.cloner import ClonePlan
 from isaaclab.renderers import BaseRenderer, RenderBufferKind, RenderBufferSpec
 from isaaclab.sim import SimulationContext
 from isaaclab.utils.warp.warp_math import convert_camera_frame_orientation_convention_wp
+
+from isaaclab_ov.stage import (
+    create_ovstage,
+    points_tensor_from_warp,
+    xform_tensor_from_numpy,
+    xform_tensor_from_warp,
+)
 
 from .ovrtx_annotator_utils import (
     build_instance_id_to_labels_and_semantics,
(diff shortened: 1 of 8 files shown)


source/isaaclab_ov/isaaclab_ov/stage.py
```diff
@@ -0,0 +1,106 @@
...
+
+    The caller owns the data: the returned tensor must stay alive until the consuming write
+    completes, and that write must be ordered against the kernels that produced
+    :paramref:`points` — pass their Warp stream as ``write_attribute(cuda_stream=...)``.
+
+    Args:
+        points: Warp array of shape ``[N]`` and dtype :class:`warp.vec3f`.
+
+    Returns:
+        A :class:`ovstage.DLTensor` with shape ``[N]`` and ``lanes=3``.
```
pbarejko, line 101: nit: return types `Any` do not match the documentation.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1935,17 +1888,18 @@ def _update_transforms_ovstage(self) -> None:
...
+        # host-side ``wp.synchronize_device()`` that would stall the CPU.
         self._stage.write_attribute(
             self._object_xform_query,
             "omni:xform",
             ordinal=self._current_ordinal,
-            tensors=_xform_tensor_from_numpy(object_transforms.numpy().reshape(-1, 4, 4)),
+            tensors=xform_tensor_from_warp(object_transforms),
             is_array=False,
             semantic=ovstage.AttributeSemantic.MATRIX,
+            cuda_stream=wp.get_stream(self._device).cuda_stream,
```
pbarejko, line 2166: Can we review situation with streams in subsequent PR please? Some helpers would be useful, Instead of `wp.get_stream()` we store it as a filed and we do `cuda_stream=self._stream`? 

Additionally we have synchronization issues :

* Rigid xforms (binding.map + wp.launch) - mplicit: write into mapped buffer on Warp’s current stream. No explicit cuda_stream to OVRTX.
* Particles (binding.write(..., ASYNC)) - Explicit: passes wp.get_stream(...).cuda_stream so OVRTX can GPU-wait.
* Cameras (wp.launch then wp.copy into map) - Same as xforms — assume mapped memory is consumed later somehow.

Additionally @pv-nvidia for vis.

pbarejko (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
pv-nvidia: I will address this in a tiny separate PR that can be merged into the release.

Note that because of a particular implementation detail in ovrtx 0.4 (_copy-back uses the default NULL stream which syncs against all blocking streams, which includes the warp stream_), it seems the current code is currently working fine without changes though

pv-nvidia: https://github.com/isaac-sim/IsaacLab/pull/7166

source/isaaclab_ov/isaaclab_ov/stage.py
rilei-nvidia: Fixed

kellyguo11 (approved):

source/isaaclab_ov/isaaclab_ov/stage.py
```diff
@@ -0,0 +1,104 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: Please remove the _ from the filename

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -32,22 +32,16 @@
 logger = logging.getLogger(__name__)
 
 import numpy as np
+
+# ovstage is a hard dependency of isaaclab_ov. Opting into it selects the split-ownership model
```
StafaH: Do we need the comment for import?

StafaH (approved):

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: I will remove this

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1966,59 +1920,55 @@ def _update_geometries_ovstage(self) -> None:
...
-                ``[total_particles, 3]``.
-            particle_offsets: Start index of each prim's slice into :paramref:`particle_np`.
+            particle_q: Flat world-space particle positions [m], shape ``[total_particles]``,
+                dtype ``wp.vec3f``. Slices are passed zero-copy as CUDA DLTensors.
+            particle_offsets: Start index of each prim's slice into :paramref:`particle_q`.
             particle_counts: Number of particles in each prim's slice.
         """
         particle_slices = [
-            _points_tensor_from_numpy(particle_np[particle_offset : particle_offset + particle_count])
+            points_tensor_from_warp(particle_q[particle_offset : particle_offset + particle_count])
```
marcodiiga, line 2221: This rebuilds a Warp slice and consumes a DLPack descriptor for every prim on every rendered frame. On an RTX A6000 with the repository-pinned OVStage 0.1.1.355824 and Warp 1.16.0, a real `write_attribute(...).wait()` benchmark with 128 points per prim measured 0.853 ms versus 0.265 ms for the previous host-staging path at 16 prims, 3.170 ms versus 0.603 ms at 64, and 12.255 ms versus 2.557 ms at 256. Reusing the descriptors reduced those measurements to 0.128, 0.329, and 1.126 ms. Please cache the per-query slices/descriptors and rebuild them only when the source pointer/device or offset/count layout changes, clearing the cache during teardown. A multi-prim regression benchmark would also cover this per-frame cost.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
```diff
@@ -1935,17 +1888,18 @@ def _update_transforms_ovstage(self) -> None:
             inputs=[object_transforms, self._object_newton_indices, body_q],
             device=self._device,
         )
-        # Synchronize then copy to CPU numpy: ovstage's make_dltensor only accepts the lanes=16
-        # dtype override on numpy arrays, not DLPack producers. wp.mat44d exports as (N,4,4) lanes=1
-        # via DLPack, which conflicts with the lanes=16 omni:xform column created at population time.
-        wp.synchronize_device(self._device)
+        # The tensor is handed over zero-copy, so ovstage reads ``object_transforms`` in place and
+        # must not do so until the kernel above has landed. ``cuda_stream`` gives it the Warp stream
+        # that kernel was enqueued on, so it inserts a GPU-side wait instead of us forcing a
```
marcodiiga: `cuda_stream` provides producer ordering by draining work already queued on that stream before OVStage accesses the tensor; it does not promise insertion of a nonblocking GPU-side stream wait. Because this call immediately `.wait()`s for the write operation, the calling thread can still block; the improvement here is eliminating the host copy and replacing device-wide synchronization with stream-scoped producer ordering. Please update this comment, the analogous particle comment, and the changelog/PR description accordingly. If a nonblocking GPU-side handoff is required, use the event-based API and verify that path end to end.

marcodiiga (approved):
Reviewed at `5a8972a9`. The GPU lane folding, buffer lifetime, `CPU_INCREMENTAL` transform behavior, and non-default-stream round trips are sound. I found no P0 correctness issue, so I’m approving. I left two actionable follow-ups: a measured per-prim points-update performance regression and synchronization wording that currently overstates nonblocking behavior.

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: Thanks, addressed!

source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py
rilei-nvidia: I will ticket this up as a follow-up

PR #7161, opened 2026-08-19
[Task Clean-up][Assets] Serve the Shadow Hand from one asset with native PhysX tendons

hujc7:
# Description

The Shadow Hand shipped as two assets, one per engine, with the PhysX tendons synthesised at
spawn — a per-environment patch that never completed scene construction at 4096 envs. One asset
now serves both, its `Physics` USD variant selecting the engine, with the four tendons authored
natively. `set_fixed_tendon_position_target_index` / `_mask` on `BaseArticulation` let a task
command a tendon without naming a backend.

The spawn-time tendon writers are not used here: the asset carries its four tendons natively and
runtime control goes through the tensor API. Their namespace fix and per-instance selection live in
#7527.

Both engines now spawn the hand at the asset's own pose. PhysX previously overrode `init_state`
with an identity rotation, so its hand sat differently from Newton's; removing that override
leaves one placement for both. The cube starts 98 mm from the fingertips instead of 132 mm, which
is a task change, not a rendering one — **existing PhysX checkpoints are not comparable across
this PR.** The two engines' variants are aligned to 0.00 mm on palm, ffdistal, thdistal, forearm
and wrist.

> [!NOTE]
> **No dependency change — this runs on the pinned Newton 1.5.1 and `uv.lock` is untouched.**
> [newton#4017](https://github.com/newton-physics/newton/pull/4017) makes `ArticulationView`
> generic over custom frequencies, but it was deliberately not backported to the 1.5 patch series,
> so it arrives with 1.6. `mjc_view_compat.py` back-fills that surface on 1.5, which keeps the
> tendon adapter written once against the 1.6 API instead of carrying two code paths. It is deleted by
> **#7468**, which is held until the pinned Newton reaches 1.6.

Fixes # (issue)

## Type of change

## Release backport

## Validation

Trained at each task's default environment count and iteration budget on Newton 1.5.1, through the
compatibility shim:

| task | success (mean last 20) |
|---|---|
| `Isaac-Shadow-Reorient` | 0.9617 |
| `Isaac-Shadow-Reorient-Direct` | 0.9704 |
| `Isaac-Shadow-Handover-Direct` | 1.0000 |
| `Isaac-Shadow-Handover` | 0.8729 |

On PhysX, the handover tasks completed ten of ten seeds at their default 2048 environments with no
non-finite values.

This PR redefines `success_rate` to test distance at episode end rather than latching on first
contact, so these are not comparable with values recorded under the previous definition. Training
budgets cut to the measured plateau: reorient 10000 → 3000, handover 5000 → 3500.

Re-run at the shipping configuration on the current head, one arm per engine, task defaults:

| task | engine | iters | success |
|---|---|---|---|
| `Isaac-Reorient-Cube-Shadow` | `newton_mjwarp` | 3000 | 0.9665 |
| `Isaac-Reorient-Cube-Shadow` | `isaacsim_physx` | 3000 | 0.9083 |

Suites across every touched package, all green: `isaaclab_physx` articulation 171, `isaaclab_newton`
articulation 198, `isaaclab_ov` articulation 87 (cpu) + 84 (cuda), newton
tendon control and reorient utils 21. The two fixed-tendon mask fixes each fail on the parent commit and pass on this one.

### Asset warnings measured against nvbug 6701561

That bug reports the Shadow Hand emitting 183 `material:binding` "outside the scope of the
reference" warnings and 1225 `Failed to find rigid body ... /joints/robot0_forearm` warnings, one
per environment. Running its own repro on `develop` reproduces both exactly; on this branch the
rigid-body warnings are gone and the binding warnings drop to 90:

| | `develop` | this PR |
|---|---|---|
| `outside the scope of the reference` | 183 (visual bindings) | 90 (physics bindings) |
| `Failed to find rigid body .../robot0_forearm` | 1225 | 0 |

The remaining 90 are a converter defect in the new asset, not a regression: the prototypes in
`instances.usda` bind a stage-absolute physics material that USD cannot remap on reference, so it
discards them. They are redundant — `base.usda` binds the same material at composed scope, and all
32 colliders resolve to it either way — so they are cosmetic. Removed from the published asset
separately; that edit does not touch this PR.

## Rendering references

The unified asset renders the hand in its own materials, so every baseline it feeds moved with it:
the `isaaclab_visualizers` goldens, the `isaaclab_tasks` rendering goldens for `shadow_hand` and
`registered_tasks/Isaac-Reorient-Cube-Shadow-Camera-Direct`, and that task's golden USD stage.

Pixel references were taken from the rendering suites' own `comparison-images` artifacts rather
than regenerated locally: they do not reproduce across GPU models, and a local L40 leaves the
shaded outputs about nine percent different from the L40S the runners use while every geometric
output matches exactly.

## Checklist


source/isaaclab_ov/isaaclab_ov/assets/articulation/articulation.py
```diff
@@ -3146,6 +3153,83 @@ def set_fixed_tendon_rest_length_mask(
...
+
+        Args:
+            target: Target tendon length [m or rad, depending on the spanned joints' type].
+                Shape is (len(env_ids), len(fixed_tendon_ids)).
+            fixed_tendon_ids: The tendon indices to command. Defaults to None (all fixed tendons).
+            env_ids: The environment indices to command. Defaults to None (all environments).
+        """
+        if isinstance(target, torch.Tensor):
+            offset = -target
+        elif isinstance(target, (int, float)):
```
ooctipus: does this ever happen?

source/isaaclab_tasks/isaaclab_tasks/core/handover/handover_env_cfg.py
```diff
@@ -37,51 +35,58 @@ def _shadow_hand_cfg(
...
-        },
-    )
-    ovphysx_cfg = SHADOW_HAND_CFG.replace(
-        prim_path=prim_path,
-        # OVPhysX does not expose the fixed-tendon runtime API, so spawn without tendon overrides.
-        spawn=SHADOW_HAND_CFG.spawn.replace(fixed_tendons_props=None),
-        init_state=SHADOW_HAND_CFG.init_state.replace(pos=init_pos, rot=init_rot),
-    )
+
+    def _for_engine(physics: str):
```
ooctipus: please check here

ooctipus: please expand the preset and don't return preset

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/__init__.py
ooctipus: if doesn't exceed the length we can just inline it

hujc7:
run-ci

source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation.py
```diff
@@ -2937,6 +2981,29 @@ def set_fixed_tendon_offset_index(
         """
         raise NotImplementedError()
 
+    def set_fixed_tendon_position_target_mask(
```
hujc7: why this is not implemented?

hujc7:
@ooctipus for reivew

ooctipus:
I think the namespace correction is real, but after tracing the USD model and the public call path, I do not think the current `_usd_multi_apply_schema` shape is the right public contract.

The important distinction is the identity of an applied schema:

```text
single-apply:   (prim, schema type)
multiple-apply: (prim, schema type, instance name)
```

For example, the complete address of one fixed-tendon root is:

```text
(/World/Hand/Joints/FFJ1, PhysxTendonAxisRootAPI, index_finger)
```

The current fragment path supplies the prim through the path expression and the schema type through `_usd_multi_apply_schema`, but it omits the instance-name coordinate and silently substitutes “all existing instances.” Consequently, given one joint with:

```usda
prepend apiSchemas = [
    "PhysxTendonAxisRootAPI:index_finger",
    "PhysxTendonAxisRootAPI:shared_coupling",
]
```

`PhysxFixedTendonCfg(stiffness=10.0)` changes both instances. Narrowing the prim-path expression cannot solve this because both instances live on the same prim.

`PhysicsDriveAPI` does not establish broadcast as our production convention. `apply_drive()` derives exactly one instance from the typed joint (`angular` for a revolute joint and `linear` for a prismatic joint), constructs that typed `UsdPhysics.DriveAPI` instance, and writes through its schema accessors. The new test that applies both `angular` and `linear` to one revolute joint and changes both demonstrates name-template resolution, but it establishes semantics opposite to the production drive writer.

If we want a real public multiple-apply fragment API, I think the instance must be modeled explicitly. A possible surface is:

```python
PhysxTendonAxisRootCfg(
    instance_names="index_finger",
    stiffness=10.0,
    damping=0.1,
)

PhysxTendonAxisCfg(
    instance_names="index_finger",
    gearing=[-0.5],
    force_coefficient=[1.0],
)
```

with deliberate selection semantics:

```text
"index_finger"                  -> that exact instance
["index_finger", "middle"]     -> exactly those instances
None                            -> all existing instances (explicit broadcast)
[]                              -> configuration error
```

I would prefer `instance_names` to be required (`MISSING`) so omission does not accidentally mean broadcast; a caller can pass `None` explicitly when it really wants all instances. This is schema identity metadata, not a USD property, so it must be removed from/excluded from the property-writing loop intentionally. The typed backend writer can then use `PhysxTendonAxisRootAPI.Get(prim, name)` for selection and `GetAll(prim)` only for the explicit `None` case, letting the generated schema form `physxTendon:<instance>:<property>`. This also avoids recreating public parsing/name-construction helpers around `Usd.SchemaRegistry`.

I also think root and axis ownership should remain separate:

- `PhysxTendonAxisRootCfg`: whole-tendon properties (`stiffness`, `damping`, limits, offset, rest length, enabled).
- `PhysxTendonAxisCfg`: per-joint-axis properties (`gearing`, `forceCoefficient`, `jointAxis`).

`PhysxTendonAxisRootAPI` automatically includes the regular axis API for the same instance, so the root joint may receive both fragments, while child joints receive only the axis fragment. Keeping the fields separate avoids duplicate ownership. Today `PhysxFixedTendonCfg` exposes only a subset of RootAPI properties, no AxisAPI properties, and `apply_fixed_tendon_properties()` filters for RootAPI prims only, so an AxisAPI-only child joint cannot be targeted. That is a reasonable boundary if this API is intentionally only for tuning whole-tendon dynamics, but it is not general tendon-schema authoring.

The architecture and tests would then express the actual contract:

- two root instances on one prim: selecting one leaves the other unchanged;
- `instance_names=None`: both change, proving broadcast is explicit;
- a root fragment cannot own `gearing`;
- an axis fragment cannot own `stiffness`;
- an axis fragment can target both the root joint's included axis and an AxisAPI-only child joint;
- generated properties use `physxTendon:<instance>:*`, never `PhysxTendonAxisRootAPI:<instance>:*`.

For this PR specifically, the new Shadow Hand asset already has native-authored tendon topology/properties and the asset config does not provide `fixed_tendons_props`. Runtime control selects tendon names explicitly and writes through the articulation tensor API. Therefore the migration does not require a new generic multiple-apply fragment convention.

My suggested scope is therefore one of:

1. Keep a private, explicitly broadcast-oriented PhysX tendon tuner and land the namespace correction as a focused schema bugfix; or
2. Introduce complete selective multiple-apply semantics, with explicit instance identity and separate RootAPI/AxisAPI ownership, as a deliberately designed public API.

The current middle state—public generic multiple-apply support whose only selection policy is implicit broadcast, while exposing only part of the tendon schema—is the part I am concerned about.

hujc7:
run-ci

ooctipus:
I split the tendon schema-addressing work into #7534. It fixes the PhysX property namespace, adds instance selection to the concrete PhysX tendon configs, and introduces a separate fixed-tendon AxisAPI cfg while keeping generic multiple-apply machinery out of core and retaining the existing public appliers.

Because this Shadow Hand migration uses native-authored tendon topology and writes runtime actions through the tensor API, those spawn-time schema writers are not a functional prerequisite for the migration itself. Please rebase #7161 on top of #7534 (or onto develop after #7534 merges) and keep the schema/utility ownership in #7534. That leaves this PR focused on the asset, tendon action, and task migration.

source/isaaclab/isaaclab/envs/mdp/actions/tendon_actions.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def IO_descriptor(self) -> GenericActionIODescriptor:
+        """The IO descriptor of the action term.
+
+        Adds the tendon names, scale, offset and clip to the base descriptor.
+
+        Returns:
+            The IO descriptor of the action term.
+        """
+        super().IO_descriptor
```
ooctipus: do we still need io descriptor?

source/isaaclab_assets/isaaclab_assets/robots/shadow_hand.py
```diff
@@ -40,182 +83,79 @@
...
     ),
     init_state=ArticulationCfg.InitialStateCfg(
         pos=(0.0, 0.0, 0.5),
-        rot=(0.0, -0.7071, 0.7071, 0.0),
+        # Newton's importer cancels the root xform during FK, so the orientation is re-applied here for both engines
+        rot=(0.0, 0.0, -0.70710678118, 0.70710678118),
         joint_pos={".*": 0.0},
     ),
+    # backends enumerate joints and bodies differently; Newton's order is the public one
+    joint_ordering="mjwarp",
```
ooctipus, line 102: do we need default ordering? I think we leave it auto determined  by physics engine instead

source/isaaclab_newton/isaaclab_newton/assets/articulation/mjc_tendon_control.py
```diff
@@ -0,0 +1,252 @@
...
+if TYPE_CHECKING:
+    from newton import Control, Model
+    from newton.selection import ArticulationView
+
+    from .articulation import Articulation
+
+logger = logging.getLogger(__name__)
+
+
+class MjcTendonControl:
```
ooctipus: Oof this file might not be a part of articulation.... could it be actuator.... or deserving a new folder called tendon

source/isaaclab_newton/isaaclab_newton/physics/_mjwarp_view_compat.py
ooctipus, line 1: this file only introduce a custom view for mjc tendon, not a custom view for all mujoco frequencies, maybe collapse this into mjc_tendon_control.py

source/isaaclab_visualizers/test/golden_images/shadow_hand/newton-kit-tiled.png
ooctipus, line 1: golden seems to be reverted

source/isaaclab/isaaclab/envs/mdp/actions/tendon_actions.py
ooctipus: Follow-up after tracing ActionManager and the sibling action terms: please keep the IO descriptor. JointAction and the other action terms expose shape, type, and name mappings through it; tendon_names is necessary to interpret this action vector. My earlier question is not a requested removal. A small clarity cleanup could bind `descriptor = super().IO_descriptor`, but no architectural change is needed here.

source/isaaclab_assets/isaaclab_assets/robots/shadow_hand.py
ooctipus: Follow-up after tracing the ordering paths: I would not automatically remove canonical ordering. Direct observations expose whole joint vectors, and the backends enumerate them differently, so `None` changes policy semantics across backends. `robot_schema` would be a more backend-neutral convention if those relationships are complete and stable; otherwise keeping `mjwarp` as the explicitly chosen public convention is defensible. Please document that decision, but my earlier comment is not a request to revert it.

source/isaaclab_newton/isaaclab_newton/assets/articulation/mjc_tendon_control.py
ooctipus: After tracing ownership, this is actuator control rather than a tendon domain model. I would not add a top-level `tendon` package. Fold the mapping, target buffering, and `mujoco.ctrl` submission into the existing `NewtonActuatorControl` in `actuator_control.py`, so the articulation has one backend-control owner. If this area later grows into several independent implementations, a plural `isaaclab_newton/actuators/` package would be the convention-aligned extraction point.

source/isaaclab_newton/isaaclab_newton/physics/_mjwarp_view_compat.py
ooctipus: Follow-up: merely collapsing this into tendon control would move the problem without fixing it. A Newton-version view shim is not an asset concern, and the equal-block actuator-row inference is not a generic view implementation. Preferred fix: require the Newton version with native custom-frequency view support and delete this file and its tests. If a 1.5 shim absolutely has to remain temporarily, a private module under `isaaclab_newton/sim/views/` is a more honest location, but it should not be presented as a general mapping.

ooctipus:
I reviewed the current head (`c6134d2`) again after the schema work moved to #7527. The unified-asset direction makes sense, but I think the remaining non-schema change still needs a clearer runtime/control boundary before merge.

## 1. Runtime tendon targets should not put all tendon properties on the normal step path

Both PhysX implementations currently do this whenever an articulation has any fixed tendon:

```python
if self.num_fixed_tendons > 0:
    self.write_fixed_tendon_properties_to_sim_*()
```

On PhysX that submits stiffness, damping, limit stiffness, position limits, rest length, and offset as a complete row. On OVPhysX it issues up to six separate attribute writes, including CPU-only bindings. This also changes the established static-property lifecycle: a property buffered for a later explicit write is now silently resubmitted on every step.

The target setter should mark only the runtime target dirty. The static property setters should retain their explicit-write contract:

```python
# initialized/reset with the articulation
self._fixed_tendon_target_dirty = False

# after a target setter updates the offset buffer
self._fixed_tendon_target_dirty = True

# write_data_to_sim
if self._fixed_tendon_target_dirty:
    self.write_fixed_tendon_properties_to_sim_index()  # PhysX batch API requires the complete row
    self._fixed_tendon_target_dirty = False
```

OVPhysX can avoid the complete property writer and submit only the dynamic offset:

```python
if self._fixed_tendon_target_dirty:
    self._root_view.set_attribute(TT.FIXED_TENDON_OFFSET, self._data._fixed_tendon_offset.data)
    self._fixed_tendon_target_dirty = False
```

The production implementation should preserve CUDA-graph capture behavior, but the important ownership rule is: setting a runtime target schedules a target write; merely having an authored tendon does not schedule repeated static-property writes.

## 2. Manager and Direct action semantics currently disagree

The manager term hard-clamps every raw action to `[-1, 1]`:

```python
processed = raw.clamp(-1.0, 1.0) * scale + offset
```

The Direct tasks use `unscale_transform()` without saturating the result. For tendon limits `[0, pi]`, a raw action of `2` therefore becomes `pi` in Manager and `1.5*pi` in Direct.

The generic action term should not embed the assumption that every caller supplies normalized policy output. Bounds already have an explicit API in `cfg.clip`:

```python
self._processed_actions[:] = self._raw_actions * self._scale + self._offset
if self.cfg.clip is not None:
    self._processed_actions.clamp_(min=self._clip[:, :, 0], max=self._clip[:, :, 1])
```

The Shadow task should state its own contract:

```python
mdp.FixedTendonPositionActionCfg(
    asset_name="robot",
    tendon_names=TENDON_NAMES,
    preserve_order=True,
    scale=0.5 * (TENDON_POSITION_LIMITS[1] - TENDON_POSITION_LIMITS[0]),
    offset=0.5 * (TENDON_POSITION_LIMITS[0] + TENDON_POSITION_LIMITS[1]),
    clip={".*": TENDON_POSITION_LIMITS},
)
```

and Direct should implement the same target-space bound:

```python
target = unscale_transform(actions, self.tendon_lower_limits, self.tendon_upper_limits)
target = saturate(target, self.tendon_lower_limits, self.tendon_upper_limits)
```

I would keep the generic `preserve_order=False` default to match the joint-action convention, but use `True` wherever a configured name list defines checkpoint action order. If this is intended as a general peer of `JointActionCfg`, `scale` and `offset` should eventually accept the same scalar-or-regex-dict forms through shared parsing rather than another copied parser.

## 4. Newton 1.5 compatibility guesses actuator ownership, while the native view still needs an explicit target mapping

There are three different relationships here, and only the first is one-to-one:

- an actuator has one transmission target;
- one fixed tendon can contain contributions from many joints;
- one tendon can be targeted by zero, one, or several actuators.

For example, the Shadow Hand path is conceptually:

```text
ctrl[rh_A_FFJ0]
  -> position actuator
  -> tendon rh_FFJ0
  -> joint rh_FFJ2 (coefficient 1)
  -> joint rh_FFJ1 (coefficient 1)
```

The `1:N` tendon-to-joint relation is handled by the MuJoCo tendon Jacobian. It does not require IsaacLab to pair one actuator with one joint, and it does not prevent the actuator row from having an unambiguous articulation owner.

The Newton 1.5 compatibility code does not have that owner metadata, so `_actuator_rows_per_instance()` guesses it by equally partitioning all actuator rows in a world:

```python
rows = np.flatnonzero(actuator_worlds == world)
articulations = np.count_nonzero(tendon_worlds == world) // tendon_count
width = rows.size // articulations
```

That happens to fit the measured homogeneous Shadow scenes, but it is not a valid generic join. If a world contains one Shadow Hand with 20 actuators and four tendons plus a tendon-free Franka with nine actuators, the calculation sees `4 / 4 == 1` tendon-bearing articulation and assigns all 29 actuator rows to the hand. Different tendon counts across articulations are ambiguous for the same reason.

[newton-physics/newton#4017](https://github.com/newton-physics/newton/pull/4017), merged for Newton 1.6, supplies the missing relation explicitly as `mujoco:actuator_articulation[row]`. Newton resolves that owner through actuator → transmission target → tendon joints → articulation. The resulting native `ArticulationView` can therefore expose 20 actuator columns and four tendon columns for one Shadow Hand without assuming that actuator row `i` corresponds to tendon row `i`.

There is one more mismatch to fix before the 1.6 native view can replace the compatibility wrapper. This proposed call is the right direction but is not sufficient with the resolver as currently written:

```python
view = articulation.root_view
columns = resolve_fixed_tendon_actuator_columns(view, model)
```

The fallback wrapper changes the meaning of the actuator-frequency labels:

```python
custom_frequency_labels["mujoco:actuator"] = actuator_target_label
```

so the current resolver sees tendon target names such as `rh_FFJ0`. The native Newton view derives those labels from `mujoco:actuator_label`, which contains actuator names such as `rh_A_FFJ0`. The resolver currently interprets both as target labels. Its mock tests encode the fallback meaning, so they do not catch the native-view failure.

The durable interface should expose an explicit actuator-column-to-tendon mapping rather than overload display labels. It should permit repeated tendon ids because several compatible actuators may target the same tendon:

```python
actuator_columns = [9, 10, 14]
target_tendon_ids = [0, 0, 1]

commands[:, actuator_columns] = tendon_targets[:, target_tendon_ids]
```

If multiple position actuators per tendon are intentionally unsupported, validate that restriction and report it clearly instead of relying on a one-to-one assumption.

The position-control contract also needs a trustworthy source. Requiring `ctrl_source == CTRL_DIRECT` and position semantics is correct, but Newton USD `MjcActuator` rows currently default to `GENERAL`; the solver later recognizes position actuators from gain and bias parameters. IsaacLab should not duplicate that solver heuristic. Newton should expose or correct the actuator classification used by this mapping.

With that metadata available, the ownership boundary should be:

- Newton owns actuator-to-articulation ownership and actuator-to-transmission-target metadata.
- `NewtonActuatorControl` resolves that mapping once, owns one complete `mujoco.ctrl` buffer, collects every direct-control contribution, and flushes it exactly once.
- `ArticulationData` owns the public fixed-tendon target buffer, and `Articulation` delegates target submission to `NewtonActuatorControl`.
- MuJoCo distributes each tendon actuator force across the tendon contributing joints.

This avoids the current second writer of `mujoco.ctrl`:

```python
commands.zero_()
scatter_fixed_tendon_position_targets(..., outputs=[commands])
view.set_attribute("mujoco.ctrl", control, commands)
```

which clears and rewrites the complete actuator-width view and can erase another `CTRL_DIRECT` contribution on the same articulation.

The intended final tree remains:

```text
isaaclab_newton/assets/articulation/
├── articulation.py          # public backend implementation and delegation
├── articulation_data.py
├── actuator_control.py      # joint and fixed-tendon control ownership
└── kernels.py
```

There should be no standalone `mjc_tendon_control.py`, no top-level tendon package, and no `mjc_view_compat.py` after the Newton dependency update. If Newton 1.5 support must remain temporarily, name and validate it as a narrow compatibility path for the supported homogeneous scene shape rather than presenting the inferred row partition as a generic view mapping.

hujc7:
run-ci

ooctipus (approved):
Thanks for addressing all the concerns approved

hujc7:
run-ci

source/isaaclab_tasks/isaaclab_tasks/core/handover/mdp/commands_cfg.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: Can merge this with commands.py?

source/isaaclab_newton/isaaclab_newton/physics/_mjwarp_view_compat.py
```diff
@@ -0,0 +1,272 @@
+# Copyright (c) 2022-2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
```
StafaH, line 1: Let's not use _ file name. Also @ooctipus has comment, if you follow that this file might not be needed either

StafaH (approved):

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

kellyguo11:
@hujc7 are you planning to address the comments around [_mjwarp_view_compat.py](https://github.com/isaac-sim/IsaacLab/pull/7161/files/ecd4ab5cb575d42a9ae3e463060486161dbb679b#diff-9cf01803304427aa38b5b3a7d5225fac69e43134017e52ea7e391b45bfa3192e)?

hujc7:
> @hujc7 are you planning to address the comments around [_mjwarp_view_compat.py](https://github.com/isaac-sim/IsaacLab/pull/7161/files/ecd4ab5cb575d42a9ae3e463060486161dbb679b#diff-9cf01803304427aa38b5b3a7d5225fac69e43134017e52ea7e391b45bfa3192e)?

All the comments should be addressed. For the shim, it will stay as a WAR for newton 1.5.1/0, the cleanup is ready for develop branch once this gets cherry-picked

PR #7164, opened 2026-08-19
Fix Newton self-collisions being uncontrolled via the deprecated PhysX cfg

AntoineRichard:
## Summary
- `ArticulationRootPropertiesCfg`/`PhysxArticulationRootPropertiesCfg.enabled_self_collisions` only authored `physxArticulation:enabledSelfCollisions`. Newton's schema resolver checks `newton:selfCollisionEnabled` first and only falls back to the PhysX attribute when it's unauthored, so the deprecated cfg's setting silently never reached Newton simulations.
- `modify_articulation_root_properties` now mirrors `enabled_self_collisions` onto `newton:selfCollisionEnabled` (applying `NewtonArticulationRootAPI`) whenever it's set, so the deprecated cfg controls self-collisions on both backends.
- Migrated `ALLEGRO_HAND_CFG`, `SHADOW_HAND_CFG`, `SHADOW_HAND_NEWTON_CFG`, and `KUKA_ALLEGRO_CFG` off the deprecated cfg to explicit `PhysxArticulationCfg` + `NewtonArticulationCfg` fragments. Several of the reorientation/handover tasks built on these hands (`Isaac-Reorient-Cube-Allegro*`, `Isaac-Reorient-Cube-Shadow-Direct*`, `Isaac-Shadow-Handover-Direct`) default to the `newton_mjwarp` preset, so this was a real training-time gap, not just a hypothetical one.
- Adds `isaaclab_physx`/`isaaclab_newton` as declared dependencies of `isaaclab_assets` (needed for the fragment imports above).

## Test plan
- [x] `test_physx_articulation_root_writes_self_collisions` extended to assert `newton:selfCollisionEnabled` / `NewtonArticulationRootAPI` are authored; confirmed it fails without the fix and passes with it.
- [x] `source/isaaclab/test/sim/test_schemas.py` (43 tests) and `test_articulation_fragments.py` (26 tests) pass.
- [x] `source/isaaclab_assets` test suite passes (`test_asset_configs` loads all three migrated robot configs).
- [x] `test_initialization_hand_with_tendons` (Shadow Hand, PhysX backend) passes with the fragment-based `articulation_props`.
- [x] `uv run isaaclab -f` passes (formatting, changelog fragments).


source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -525,6 +525,19 @@ def modify_articulation_root_properties(
     # apply per-field exceptions + main-namespace writes
     _apply_namespaced_schemas(articulation_prim, cfg, cfg_dict)
 
+    # backward-compat shim: the deprecated PhysX-only 'enabled_self_collisions' field (present only
+    # on ArticulationRootPropertiesCfg / PhysxArticulationRootPropertiesCfg) only authors
+    # 'physxArticulation:enabledSelfCollisions'. Newton's schema resolver checks the native
+    # 'newton:selfCollisionEnabled' attribute first and only falls back to the PhysX attribute when
+    # the newton-native one is unauthored, so mirror the value there too.
```
AntoineRichard: Shorten

source/isaaclab/test/sim/test_schemas.py
```diff
@@ -507,7 +507,10 @@ def test_articulation_root_base_no_physx_schema_when_only_fix_root_link_set(setu
 @pytest.mark.isaacsim_ci
 def test_physx_articulation_root_writes_self_collisions(setup_simulation):
     """Setting ``enabled_self_collisions`` on ``PhysxArticulationRootPropertiesCfg`` must
-    author ``physxArticulation:enabledSelfCollisions`` AND apply ``PhysxArticulationAPI``."""
+    author ``physxArticulation:enabledSelfCollisions`` AND apply ``PhysxArticulationAPI``, and must
+    mirror the value onto ``newton:selfCollisionEnabled`` / ``NewtonArticulationRootAPI`` so the
+    deprecated PhysX-only cfg also takes effect under Newton (which checks its native attribute
+    first and only falls back to the PhysX one when unauthored)."""
```
AntoineRichard: shorten

kellyguo11 (approved):

StafaH (approved):

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -525,6 +525,16 @@ def modify_articulation_root_properties(
     # apply per-field exceptions + main-namespace writes
     _apply_namespaced_schemas(articulation_prim, cfg, cfg_dict)
 
+    # backward-compat shim: mirror the deprecated PhysX-only 'enabled_self_collisions' onto
+    # 'newton:selfCollisionEnabled', which Newton checks before falling back to the PhysX attribute.
+    enabled_self_collisions = cfg_dict.get("enabled_self_collisions")
```
ooctipus: I wonder would this not work if just delete it, the physx resolver should pick it up?

AntoineRichard: PhysX picks it up, but newton does not, the issue is that if there is a newton authored property, then it takes precendence over the PhysX one.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -582,10 +582,41 @@ def modify_articulation_root_properties(
...
+                newton_properties = []
+                if schema_definition is not None:
+                    for property_name in schema_definition.GetPropertyNames():
+                        prop = articulation_prim.GetProperty(property_name)
+                        if prop and prop.IsAuthored():
+                            newton_properties.append(prop)
+                for prop in newton_properties:
+                    if not prop.FlattenTo(parent_prim):
+                        raise RuntimeError(f"Failed to move '{prop.GetPath()}' to '{parent_prim.GetPath()}'.")
+                for prop in newton_properties:
```
marcodiiga, line 600: [P1] Avoid removing referenced property specs during relocation — With a standard `UsdFileCfg` reference, these Newton attributes are authored in the referenced layer. `FlattenTo(parent_prim)` succeeds, but `articulation_prim.RemoveProperty(...)` edits only the current target and returns false, so this block raises and aborts spawning. I reproduced this with public `usd-core==25.11` and `newton-usd-schemas==0.4.1`. The shared `PhysicsManager._relocate_articulation_root` removes the applied schema but deliberately does not remove individual properties. Please reuse that helper or omit these removals and add a referenced-layer regression. Also account for raw `apiSchemas` metadata: before codeless registration, `GetAppliedSchemas()` hides the Newton token, which can leave it on the child and produce two roots when the stage is reopened with the schema registered.

source/isaaclab/isaaclab/sim/schemas/schemas.py
```diff
@@ -582,10 +582,41 @@ def modify_articulation_root_properties(
...
+                    raise RuntimeError(f"Failed to remove '{newton_root_schema}' from '{articulation_prim.GetPath()}'.")
 
             # remove api from root
             articulation_prim.RemoveAppliedSchema("PhysxArticulationAPI")
             articulation_prim.RemoveAPI(UsdPhysics.ArticulationRootAPI)
+            articulation_prim = parent_prim
+
+    # Mirror after any root relocation so the Newton API does not recreate an articulation root
+    # on the former root link.
+    enabled_self_collisions = cfg_dict.get("enabled_self_collisions")
```
marcodiiga, line 613: [P1] Do not overwrite an explicit Newton value from the supported PhysX cfg — This writer receives the non-deprecated `PhysxArticulationRootPropertiesCfg` as well as the deprecated alias. If an asset authors `newton:selfCollisionEnabled=False`, setting the PhysX field to true now silently replaces that Newton choice, although the class documents one cfg/namespace and `NewtonArticulationRootPropertiesCfg` owns the Newton side. Since the affected assets are already migrated to an explicit PhysX + Newton fragment pair, please either scope this compatibility mirror to the deprecated alias or document the supported PhysX cfg as dual-authoring and add a precedence regression.

source/isaaclab_assets/pyproject.toml
```diff
@@ -20,11 +20,15 @@ requires-python = ">=3.12"
 dependencies = [
     "isaaclab",
     "isaaclab_contrib",
+    "isaaclab_physx",
```
marcodiiga, line 23: [P2] Regenerate the lockfile for these package edges — The committed `isaaclab-assets` block in `uv.lock` still lists only `isaaclab` and `isaaclab-contrib`, so package-specific frozen sync/export metadata does not record these two new requirements. The branch and current `develop` both need a lock refresh, but these missing dependency edges are introduced here; please regenerate `uv.lock` (or coordinate the refresh with the current base).

source/isaaclab_assets/changelog.d/fix-hand-self-collisions-newton.rst
```diff
@@ -0,0 +1,12 @@
+Fixed
+^^^^^
+
+* Fixed self-collisions being uncontrolled under Newton for
+  :data:`~isaaclab_assets.robots.allegro.ALLEGRO_HAND_CFG`,
+  :data:`~isaaclab_assets.robots.shadow_hand.SHADOW_HAND_CFG`,
+  :data:`~isaaclab_assets.robots.shadow_hand.SHADOW_HAND_NEWTON_CFG`, and
+  :data:`~isaaclab_assets.robots.kuka_allegro.KUKA_ALLEGRO_CFG`. Their ``articulation_props`` used
```
marcodiiga, line 8: [P2] Document the public config-shape migration — These exported hand configs change `spawn.articulation_props` from a cfg object to a list. Downstream code that customizes them through `cfg.spawn.articulation_props.enabled_self_collisions = ...` now raises `AttributeError`. The fragment pair is the intended abstraction, but the changelog should mark the shape change and show how to update or replace the appropriate fragment.

marcodiiga (approved):
Reviewed current head 01257e8. No P0 or critical release blocker found, so approving. The PhysX + Newton fragment pair is the repository’s intended extensible abstraction: subclasses are accepted and each fragment writes an isolated namespace. The Newton schema is supplied by the existing Apache-2.0 codeless `newton-usd-schemas` package; this PR embeds no schema code or data. I left actionable comments for referenced/unregistered-schema relocation, mirror scope, lock metadata, and migration guidance. The tests verify USD authoring and root placement; an end-to-end Newton collision assertion remains worthwhile.

PR #7165, opened 2026-08-19
Drop the unreachable Newton preset from the Digit tasks

AntoineRichard:
## Problem

`LocomotionVelocityRoughEnvCfg` declares `events.base_com` as a preset whose
`newton_mjwarp` branch disables the centre-of-mass randomization, because Newton does
not support it:

```python
base_com = preset(
    default=EventTerm(func=mdp.randomize_rigid_body_com, ...),
    newton_mjwarp=None,
)
```

On tasks that offer Newton, that branch is a *companion*: it rides along with the
`newton_mjwarp` already on the physics axis and is never selectable on its own.

Digit is PhysX-only — `DigitPhysicsCfg` declares only `isaacsim_physx`, `physx` and
`default`, and its own docstring says so. The inherited branch therefore names no
reachable backend:

| token | result |
|---|---|
| `physics=newton_mjwarp` | **rejected** — the task has no Newton physics variant |
| `presets=newton_mjwarp` | **resolves**, stays on `PhysxCfg`, and silently drops the randomization |

So the three Digit tasks advertised a `presets=newton_mjwarp` that was never a backend
switch. Selecting it gave you PhysX physics with a Newton-only workaround applied.

Affects `IsaacContrib-Velocity-Flat-Digit`, `IsaacContrib-Velocity-Rough-Digit` and
`IsaacContrib-Tracking-LocoManip-Digit` — all three derive from `DigitRoughEnvCfg`.

## Fix

Collapse the preset to its default in `DigitRoughEnvCfg.__post_init__`, so the task
offers only what it can run. One edit covers all three tasks.

## Verification

- All three tasks now report `presets: []` (was `['newton_mjwarp']`); `base_com`
  survives with `body_names="torso_base"`.
- `AnymalCRoughEnvCfg` still carries `['default', 'newton_mjwarp']` — the shared
  config and every task that does offer Newton are untouched.
- New `test_digit_physics_presets.py` follows the existing
  `test_dr_legs_physics_presets.py` pattern (offline, no sim). Confirmed the guard
  fails without the fix and passes with it:

  ```
  1 failed, 2 passed   # fix reverted
  3 passed             # fix applied
  ```

## Notes

Found while building a task-discovery API that enumerates each task's runnable backend
combinations — this surfaced as a task declaring a preset it cannot resolve.

Sibling velocity tasks were checked: `Velocity-Flat-AnymalC` has the same `base_com`
preset *and* `newton_mjwarp` on its physics axis, so the companion is correctly paired
there; `G1` and `Spot` override `base_com` away entirely. Digit was the only one that
inherited the companion without the backend.


source/isaaclab_tasks/isaaclab_tasks/contrib/velocity/config/digit/rough_env_cfg.py
```diff
@@ -261,5 +261,12 @@ def __post_init__(self):
         # events
         self.events.add_base_mass.params["asset_cfg"].body_names = "torso_base"
         self.events.base_external_force_torque.params["asset_cfg"].body_names = "torso_base"
-        self.events.base_com.default.params["asset_cfg"].body_names = "torso_base"
+        # ``LocomotionVelocityRoughEnvCfg`` declares ``base_com`` as a preset whose
+        # ``newton_mjwarp`` branch disables the randomization, because Newton does not
+        # support it. Digit is PhysX-only (see :class:`DigitPhysicsCfg`), so that branch
+        # names no reachable backend: it survives as a bare ``presets=newton_mjwarp``
+        # token that would drop this randomization from a PhysX run. Collapse the preset
+        # to its default so the task offers only what it can actually run.
```
AntoineRichard: Shorten

source/isaaclab_tasks/test/core/test_digit_physics_presets.py
AntoineRichard: Remove

kellyguo11 (approved):

PR #7166, opened 2026-08-19
fix: Order OVRTX mapped-attribute commits after the producing Warp stream

pv-nvidia:
## Description

Filling a CUDA-mapped OVRTX attribute buffer requires the unmap to name the Warp stream that produced the data — an unmap without a CUDA sync performs no synchronization, and the `with binding.map(...)` form cannot carry one (its `__exit__` takes no arguments). Until now the ordering held only by accident: OVRTX's commit kernel launches on the legacy default CUDA stream, which happens to serialize behind Warp's blocking-type streams. That is an implementation detail its API contract does not promise.

This adds `isaaclab_ov.renderers.map_attribute_for_warp_writes`, a reusable context manager that maps a binding for CUDA writes and unmaps it with the producing Warp stream, and routes the renderer's object and camera transform writes — the only GPU attribute mappings in the codebase — through it. Render-var maps need no changes: CUDA maps already order reads via `_map_render_var_to_dlpack`, and CPU maps are synchronized by OVRTX before `map()` returns.

No behavior change on current OVRTX builds; this turns an accidental ordering into a contractual one and gives future call sites one safe pattern.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_mapping.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_mapping.py
@@ -22,19 +22,22 @@ from typing import Any
 import warp as wp
 
 
-def get_warp_device_id(device: str) -> int:
-    """CUDA device index of the Warp device ``device`` resolves to.
+def _cuda_device_id(device: str) -> int:
+    """CUDA device index parsed from a Warp device string, e.g. ``"cuda:1"`` -> ``1``.
 
-    Resolving through Warp keeps the mapped device consistent with where Warp enqueues the fill
-    work: a bare ``"cuda"`` means Warp's current CUDA device, not index ``0``.
+    TODO: A bare ``"cuda"`` parses to ``0`` while Warp enqueues fill work on its *current* CUDA
+    device, so the mapping and the fill can target different GPUs on multi-GPU processes. The
+    split predates this helper and is kept here to avoid a behavior change; a follow-up caches
+    the resolved Warp device on the renderer instead of re-deriving it from strings.
 
     Args:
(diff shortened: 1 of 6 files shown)


source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_mapping.py
```diff
@@ -0,0 +1,71 @@
...
+    # ovrtx without the guarded environment ``ovrtx_renderer`` establishes (OVRTX_SKIP_USD_CHECK,
+    # actionable install error). Any real ``binding`` was created through that path, so ovrtx is
+    # already imported by the time this runs.
+    from ovrtx import Device  # noqa: PLC0415
+
+    attr_mapping = binding.map(device=Device.CUDA, device_id=_cuda_device_id(device))
+    try:
+        yield wp.from_dlpack(attr_mapping.tensor, dtype=dtype)
+    finally:
+        attr_mapping.unmap(stream=wp.get_stream(device).cuda_stream)
```
pbarejko, line 71: Can we have the stream defined once and reused for all writes? `_write_particle_q_slices` uses it too!

pv-nvidia: Could do that, but my upcoming PR #7169 goes a bit further and caches the warp device itself to work around a bug where passing device id `"cuda"` would pick up whatever the current device is instead of using the device at the time the ovrtx renderer was created; a potential bug in mgpu scenario

And then we can just do ` self._warp_device.stream.cuda_stream` 

Whatever you prefer, I could also do it in this PR but it felt a different task

rilei-nvidia (approved):

PR #7167, opened 2026-08-19
[Tasks] Fix composite physics preset selection

StafaH:
# Description

Fix typed `physics=NAME` selection for presets whose alternatives bundle a complete `SimulationCfg`.

The cabinet tasks use complete simulation presets because switching physics backends also changes the simulation timestep. Preset discovery and runtime validation previously classified alternatives only by their direct type, so these valid bundles appeared only under `presets=` and `physics=newton_mjwarp` was rejected.

This change centralizes typed preset classification in `PresetTarget.matches()`, recognizes a `SimulationCfg` carrying a `PhysicsCfg` as a physics preset, uses that classification consistently for help discovery and runtime validation, and documents complete simulation physics presets.

## Type of change

- Documentation update

## Validation

- `uv run isaaclab train --task Isaac-Open-Drawer-Franka-Direct --rl_library rsl_rl --max_iterations 5 physics=newton_mjwarp`
  - Completed 5 iterations with 4,096 environments and reported `Physics: newton_mjwarp`.
- Resolved all five physics choices for all four affected registrations and verified the existing backend-specific timestep and decimation values.
- `uv run --no-sync python -m pytest source/isaaclab_tasks/test/core/test_preset_cli.py source/isaaclab_tasks/test/core/test_hydra.py -q`
  - `112 passed`.
- `uv run --no-sync isaaclab -f`
  - Passed.

## Checklist

- [x] I have run the pre-commit checks with `isaaclab -f`

### DIFF
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
@@ -367,9 +367,9 @@ def _enumerate_agents(task_name: str, agent_library: str) -> tuple[list[str], di
 def _bucket_variants_by_target(walked: dict) -> dict[PresetTarget, set[str]]:
     """Convert :func:`collect_presets` output into ``{target: set[name]}``.
 
-    Routes each ``(name, cfg)`` by ``isinstance(cfg, target.base_classes)``;
-    cfgs matching no typed target fall into ``DOMAIN``. The implicit
-    ``default`` field is filtered -- it's the fallback, not a selectable name.
+    Routes each ``(name, cfg)`` through :meth:`PresetTarget.matches`; cfgs
+    matching no typed target fall into ``DOMAIN``. The implicit ``default``
+    field is filtered -- it's the fallback, not a selectable name.
 
     Routing by class hierarchy means new backends subclassing
     :class:`~isaaclab.physics.PhysicsCfg` /
@@ -383,7 +383,7 @@ def _bucket_variants_by_target(walked: dict) -> dict[PresetTarget, set[str]]:
             if name == "default":
                 continue
             matched = next(
-                (t for t in typed_targets if isinstance(cfg, t.base_classes)),
+                (target for target in typed_targets if target.matches(cfg)),
                 PresetTarget.DOMAIN,
             )
             result[matched].add(name)
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
@@ -259,9 +259,9 @@ class _DescriptionBuilder:
 
     @staticmethod
     def _description(target: PresetTarget) -> str:
-        """One-line description; for typed targets includes the cfg base class name."""
+        """One-line description of a selector's semantic target."""
         if target.base_classes:
-            return f"(typed) selects a {target.base_classes[0].__name__} variant"
+            return f"(typed) selects a {target.value} backend"
         return "broadcast: applied to every matching PresetCfg"
 
 
@@ -371,10 +371,11 @@ def _bucket_variants_by_target(walked: dict) -> dict[PresetTarget, set[str]]:
     matching no typed target fall into ``DOMAIN``. The implicit ``default``
     field is filtered -- it's the fallback, not a selectable name.
 
-    Routing by class hierarchy means new backends subclassing
+    Direct routing by class hierarchy means new backends subclassing
     :class:`~isaaclab.physics.PhysicsCfg` /
     :class:`~isaaclab.renderers.renderer_cfg.RendererCfg` bucket automatically
-    regardless of what name the env_cfg gives the field.
+    regardless of what name the env_cfg gives the field. Targets may also
+    recognize documented container shapes through :meth:`PresetTarget.matches`.
     """
     typed_targets = [t for t in PresetTarget if t.base_classes]
     result: dict[PresetTarget, set[str]] = {target: set() for target in PresetTarget}
--- a/source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
+++ b/source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
@@ -12,7 +12,7 @@ needs to know about that category in one place:
   ``physics=NAME``) and ``self.value``.
 * ``base_classes`` -- the cfg base classes whose subclass instances belong to
   this bucket. Help-time bucketing in :mod:`isaaclab_tasks.utils.preset_cli`
-  routes variants by ``isinstance`` against these. Empty for
+  routes variants through :meth:`PresetTarget.matches`. Empty for
   :attr:`PresetTarget.DOMAIN`, which is the catch-all whose membership is
   "no typed target matched".
 * ``legacy_aliases`` -- deprecated-name to canonical-name table for this
@@ -29,6 +29,7 @@ import functools
 
 from isaaclab.physics import PhysicsCfg
 from isaaclab.renderers.renderer_cfg import RendererCfg
+from isaaclab.sim import SimulationCfg
 
 
 class PresetTarget(enum.Enum):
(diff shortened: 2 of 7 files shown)


AntoineRichard (approved):

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -118,6 +121,22 @@ def __new__(
...
+
+        Args:
+            cfg: Preset alternative to classify.
+
+        Returns:
+            Whether the alternative belongs to this target.
+        """
+        if isinstance(cfg, self.base_classes):
+            return True
+        return self is PresetTarget.PHYSICS and isinstance(cfg, SimulationCfg) and isinstance(cfg.physics, PhysicsCfg)
```
hujc7, line 139: **[P2] Agent review:** this line widens the classification contract, but the test that documents that contract wasn't updated — so green CI here doesn't tell you the new branch is covered.

`test_bucket_variants_routes_by_base_class_isinstance` (`source/isaaclab_tasks/test/core/test_preset_cli.py:243`) still states the old rule:

> Variants bucket by ``isinstance`` against ``PresetTarget.base_classes``.
> […] and everything else falls into DOMAIN.

That is no longer true — a `SimulationCfg` carrying a `PhysicsCfg` is "everything else" by base-class `isinstance`, yet now routes to `PHYSICS`. The test keeps passing only because it never constructs a `SimulationCfg`; its `_PhysWrapper` case is a `PhysicsCfg` **subclass**, a different shape from the bundle this PR adds support for.

Suggest extending that test rather than adding a file:

```diff
     walked = {
         ...
+        "sim": {  # complete SimulationCfg bundle -> PHYSICS via the new rule
+            "default": SimulationCfg(dt=1 / 60, physics=_PhysVariant()),
+            "physx": SimulationCfg(dt=1 / 60, physics=_PhysVariant()),
+        },
+        "sim_no_backend": {  # carries no PhysicsCfg -> must stay DOMAIN
+            "plain": SimulationCfg(dt=1 / 240),
+        },
     }
     result = _bucket_variants_by_target(walked)
+    assert "physx" in result[PresetTarget.PHYSICS]
+    assert "plain" in result[PresetTarget.DOMAIN]
```

The second assertion earns its place as much as the first: it pins the `isinstance(cfg.physics, PhysicsCfg)` half of the predicate, so a later refactor that drops it fails here instead of silently letting `physics=` accept any complete simulation config. The name and docstring want updating alongside — `routes_by_base_class_isinstance` no longer describes the rule.

Checked locally: the same `_bucket_variants_by_target` call buckets a `SimulationCfg(physics=...)` into `DOMAIN` on `develop` and into `PHYSICS` at this head.

- [ ] `test_bucket_variants_*` covers both bundle-with-backend and bundle-without-backend

I see the description notes no tests were added by request; if that's settled, the name/docstring half still stands on its own.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
```diff
@@ -12,7 +12,7 @@
   ``physics=NAME``) and ``self.value``.
 * ``base_classes`` -- the cfg base classes whose subclass instances belong to
   this bucket. Help-time bucketing in :mod:`isaaclab_tasks.utils.preset_cli`
-  routes variants by ``isinstance`` against these. Empty for
+  routes variants through :meth:`PresetTarget.matches`. Empty for
```
hujc7: **[P3] Agent review:** this line was updated to point at `matches()`, but three sibling statements of the same rule were not, so the module now describes two different contracts.

- `:13` — "the cfg base classes whose subclass instances belong to this bucket"; membership is no longer subclass-only.
- `:109` — `__new__`'s `base_classes` arg: "route to this target via :func:`isinstance`".
- `:21` / `:55` — "Adding a new typed target = appending one enum member". A future container-shaped target would also need a branch in `matches()`, so this no longer holds.

The Isaac Lab Review Bot raised the `:109` one in its summary ("A minor residual inconsistency remains in the unchanged `__new__` parameter documentation, which still describes routing solely via `isinstance`") without opening a thread — repeating it here so it isn't lost.

Non-blocking. Either refresh the three lines to name `matches()` and its container clause, or make the rule data instead of a branch — an optional per-member `(container_cls, attr)` so `PHYSICS` carries `(SimulationCfg, "physics")` and `matches()` stays generic, which would keep the "appending one enum member" promise literally true.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
```diff
@@ -367,9 +367,9 @@ def _enumerate_agents(task_name: str, agent_library: str) -> tuple[list[str], di
 def _bucket_variants_by_target(walked: dict) -> dict[PresetTarget, set[str]]:
     """Convert :func:`collect_presets` output into ``{target: set[name]}``.
 
-    Routes each ``(name, cfg)`` by ``isinstance(cfg, target.base_classes)``;
-    cfgs matching no typed target fall into ``DOMAIN``. The implicit
-    ``default`` field is filtered -- it's the fallback, not a selectable name.
+    Routes each ``(name, cfg)`` through :meth:`PresetTarget.matches`; cfgs
```
hujc7, line 370: **[P3] Agent review:** bucketing now routes through `matches()`, but the help text labelling the resulting bucket still describes the old, narrower rule.

`_description` at `source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py:264` builds the row from `base_classes[0].__name__`:

```python
return f"(typed) selects a {target.base_classes[0].__name__} variant"
```

So for the four cabinet registrations `--help` now prints `physics=NAME  (typed) selects a PhysicsCfg variant` above a list whose variants are `SimulationCfg` bundles. Minor, but it is the one place a user learns what `physics=` accepts.

Non-blocking — wording like "selects a physics backend" would cover both shapes without per-target strings.

hujc7 (approved):
Approving — three non-blocking comments inline.

Since the typed selector exists to *reject* `physics=` names that never land on a physics config, I checked that guard still fires. Ran the new `matches()` and `_pick_alternative` through `_validate_typed_presets` (which this PR leaves byte-identical):

| case | result |
|---|---|
| `physics=` → `SimulationCfg` bundle carrying a backend | passes |
| `physics=` → `SimulationCfg` with `physics=None` | raises |
| `physics=` → plain scalar | raises |
| `physics=` → `RendererCfg` only | raises |
| `renderer=` → physics bundle | raises |

Row 2 is the load-bearing one: the widening didn't collapse into "any `SimulationCfg` counts".

Two notes for later, not for this PR — the guard asserts type *membership*, never "the backend actually changed", which isn't a new hole (`fast: PhysxCfg(...)` / `slow: PhysxCfg(...)` passes today for the same reason) but is now reachable by more shapes; and the added `from isaaclab.sim import SimulationCfg` costs ~3 ms with no `pxr`/`omni` pulled in, since `isaaclab.sim` already loads transitively on this path.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
StafaH: Addressed in 4010d8a16. Extended the closest existing bucketing test with a `SimulationCfg` carrying `PhysicsCfg` (PHYSICS) and one without a backend (DOMAIN), and renamed its test/docstring to describe `PresetTarget.matches`. The focused CLI/Hydra suite passes: 112 tests.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_target.py
StafaH: Addressed in 4010d8a16. Updated the module and enum docs, `DOMAIN` documentation, and `__new__` parameter documentation so they consistently define routing through `matches()`, including the documented container-shape extension point.

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
StafaH: Addressed in 4010d8a16. Typed help now says `selects a physics backend` or `selects a renderer backend`, which accurately covers both direct cfg variants and complete simulation bundles. The updated help-text assertions are included in the passing 112-test focused suite.

PR #7175, opened 2026-08-19
Raise on unsupported camera renderer outputs

AntoineRichard:
## Summary
- Raise clear errors when a camera requests unknown data types or outputs unsupported by its active renderer.
- Raise instead of silently dropping Isaac RTX albedo and simple-shading outputs on Isaac Sim versions before 6.0.
- Replace the warning regression test with error coverage for renderer-capability and typo paths.

## Breaking change
Camera configurations that previously ran while silently omitting unsupported data types now fail at camera initialization. Remove unsupported types from `CameraCfg.data_types` or select a compatible renderer.

## Validation
- `uv run --extra test --frozen isaaclab -f`
- Runtime-free smoke test of `Camera._create_buffers` for both error paths.
- The Isaac Sim integration test could not run in this environment because the full Isaac Sim runtime is not installed.

## In-tree configuration finding
Cartpole simple-shading presets default to `newton_renderer`, which cannot produce those outputs and will now raise. Kuka Allegro already rejects the equivalent Newton combinations in its existing task-level compatibility validation.

## Out of scope
Teaching task discovery about renderer capabilities remains follow-up work because renderer capability is currently per-instance.

### DIFF
--- a/source/isaaclab/isaaclab/sensors/camera/camera.py
+++ b/source/isaaclab/isaaclab/sensors/camera/camera.py
@@ -658,15 +658,12 @@ class Camera(SensorBase):
                 unknown.append(name)
         errors = []
         if unknown:
-            errors.append(
-                f"Unknown camera data types: {unknown}.\n\tHint: Check CameraCfg.data_types for spelling errors."
-            )
+            errors.append(f"Unknown camera data types: {unknown}.")
         if unsupported:
             errors.append(
                 f"Renderer {type(self._renderer).__name__} does not support the following requested data types:"
                 f" {unsupported}."
                 f"\n\tSupported data types: {sorted(str(kind) for kind in specs)}"
-                "\n\tHint: Remove the unsupported data types or select a renderer that supports them."
             )
         if errors:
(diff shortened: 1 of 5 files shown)


source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -654,13 +655,21 @@ def _create_buffers(self):
...
         if unsupported:
-            logger.warning(
-                "Renderer %s does not support the following requested data types and will not produce them: %s",
-                type(self._renderer).__name__,
-                unsupported,
+            errors.append(
+                f"Renderer {type(self._renderer).__name__} does not support the following requested data types:"
+                f" {unsupported}."
+                f"\n\tSupported data types: {sorted(str(kind) for kind in specs)}"
+                "\n\tHint: Remove the unsupported data types or select a renderer that supports them."
```
nvsekkin: this hint can be removed, I think we can leave it as `Supported data types:`

nvsekkin (approved):

StafaH (approved):

PR #7177, opened 2026-08-19
Fix default ground plane size in TerrainImporter

aneangel:
## Summary
- Reduce the default `size` in `TerrainImporter.import_ground_plane` from `(2.0e6, 2.0e6)` to `(2.0e3, 2.0e3)` to resolve the reported issue when viewing a policy trained in Isaac Lab with the Newton Visualizer.

## Context
This was originally reported against Newton (https://github.com/newton-physics/newton/issues/3977), but after investigation the issue is isolated to Isaac Lab's default ground plane size, not Newton itself.

## Test plan
- [x] Verify `TerrainImporter.import_ground_plane` creates a ground plane with the new default size
- [x] Confirm no numerical precision issues in downstream simulations that rely on the default ground plane


matthewtrepte (approved):

AntoineRichard (approved):

matthewtrepte (approved):

ooctipus (changes requested):
Thanks for the diagnosis and PR, but changing the default terrain size to fix a visualizer bug feels like a wrong work around. But thanks for bring up the issue

I will dig into this  a bit to figure out whats wrong

Also your golden seems missing the table.

ooctipus:
https://github.com/newton-physics/newton/pull/4006
https://github.com/newton-physics/newton/pull/4007

could potentially solve the issue

PR #7179, opened 2026-08-19
Sensor destructor shutdown

pbarejko:
# Description

When a run aborts with a camera still alive, `Camera.__del__` → `SensorBase._clear_callbacks` touches `isaaclab.sim` through the lazy loader. During interpreter shutdown `sys.meta_path` is already `None`, so that import raises `ImportError: sys.meta_path is None, Python is likely shutting down`. That exception is printed last and hides the real failure (in NVBug 6605866, an `OvstageError` ~70 lines earlier). Same class of defect as NVBug 6591313 (`AssetBase.__del__`), different destructor.

Skip sensor/camera destructor cleanup when `sys.is_finalizing()` is true or `sys.meta_path` is `None`, using the same `_sys=sys` default-argument idiom as the env destructors. Bind `sys` at definition time so the check still works after module globals are cleared.

Fixes NVBug 6605866

Fixes # (issue)

## Type of change

- Documentation update

## Screenshots

## Checklist


source/isaaclab/isaaclab/sensors/camera/camera.py
```diff
@@ -226,17 +227,24 @@ def __init__(self, cfg: CameraCfg):
...
+        """
+        if _sys.is_finalizing() or _sys.meta_path is None:
+            return
         # unsubscribe callbacks
         super().__del__()
-        # release the frame view's backend state
-        if self._view is not None:
+
+        # release the frame view's backend state, getattr covers partial initialization
+        if getattr(self, "_view", None) is not None:
```
StafaH, line 242: Can we do this without getattr here and below? Is there a possibility of _view and _renderer not existing in the class?

StafaH (approved):

kellyguo11 (approved):

source/isaaclab/isaaclab/sensors/camera/camera.py
pbarejko: Using `getattr` is intentional here. `__del__` can be called on a partially initialized object if `__init__` raises before all fields are set. In that case, directly accessing `self.foo` could raise an `AttributeError` during destruction, whereas `getattr(self, "foo", None)` lets cleanup safely handle the “not initialized” case.

PR #7190, opened 2026-08-19
Point the LEAPP semantics link at its current page

mataylor-nvidia:
## Summary

- point the LEAPP semantic annotation link at `semantics/usage.html`, the page it lives on now

## Why

The LEAPP docs moved the guide, so `https://nvidia-isaac.github.io/leapp/guides/semantics.html` returns 404. The link checker runs on any pull request that touches a Markdown or reStructuredText file and checks every doc, so this one stale link fails `Check for Broken Links` on changes that have nothing to do with the tutorial -- for example isaac-sim/IsaacLab#7059, which only edits `.github/`.

`develop` already carries the corrected URL; this ports just that line, without the rest of the tutorial rewrite on that branch.

## Impact

Documentation only. No behavior change.

## Validation

- `uv run isaaclab -f`
- confirmed the old URL returns 404 and the new one returns 200

### DIFF
--- a/docs/source/tutorials/06_exporting/exporting_direct_workflow_policies_with_leapp.rst
+++ b/docs/source/tutorials/06_exporting/exporting_direct_workflow_policies_with_leapp.rst
@@ -240,6 +240,6 @@ into deployment systems.
 .. note::
 
    Refer to the `LEAPP semantic annotation guide
-   <https://nvidia-isaac.github.io/leapp/guides/semantics.html>`_
+   <https://nvidia-isaac.github.io/leapp/semantics/usage.html>`_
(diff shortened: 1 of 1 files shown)


AntoineRichard:
@frlai for viz

AntoineRichard:
@mataylor-nvidia did you mean to target develop?

frlai:
@AntoineRichard This was already updated on develop. I think the goal is actually to target the release branch. there are actually 2 approaches here.

I think I changed the link to suddenly and I'm willing to also publish and update to the leapp docs where the old URL will actually redirect you to the new URL. If we are against updating the release branch I can also do that. But if this change targeting the release branch is ok this would be much easier.

PR #7192, opened 2026-08-19
[Bump] usd-exchange 3.0.0 and drop the OpenUSD work-thread-limit workaround

hujc7:
# Description

Bumps the pinned standalone OpenUSD provider from `usd-exchange` 2.3.0 to **3.0.0**.

`usd-exchange` 3.0.0 makes OpenUSD 26.08 the default flavor ([usd-exchange!257](https://gitlab-master.nvidia.com/omniverse/usd-exchange/usd-exchange/-/merge_requests/257)). The PyPI wheel ships the default flavor, so this moves `pxr` in **kit-less** installs from OpenUSD 25.05 to 26.08. Kit is unaffected — it serves its own OpenUSD and its own usdex from its extension roots.

**Isaac Sim 6.1 pins `usd-exchange` exactly, so this now needs a uv override.** Develop moved `importers` to `isaacsim-asset-isolated==6.1.0.0`, which requires `usd-exchange==2.3.0` — 6.0.1.0 required `>=2.2.2`. `isaacsim-robot-schema` 6.1.0.0 pins the same, and the new transitive `usd-optimize` 1.2.1 caps at `>=2.2,<3`. Without an override `uv lock` is unsatisfiable.

The two schema packages ship **zero** compiled extensions and bind usdex's Python API at runtime, so 3.0.0 satisfies them with no ABI coupling. `usd-optimize` does not — its cap is real, and the consequence is scoped in §Known limitation.

## Type of change

Kit-less installs move to OpenUSD 26.08. Anything pinning Isaac Lab against 25.05 outside Kit must re-resolve, and `merge_mesh` on both asset converters silently stops merging (§Known limitation).

## Changes

**Provider bump**

| File | Change |
| --- | --- |
| `pyproject.toml` | `usd-exchange==3.0.0`; `[tool.isaaclab.versions] usd_exchange = "3.0.0"`; new `[tool.uv] override-dependencies` entry forcing 3.0.0 past the Isaac Sim 6.1 pins |
| `tools/wheel_builder/uv-overrides.txt`, `source/isaaclab/test/install_ci/uv_pip/uv-overrides.txt` | regenerated via `tools/wheel_builder/gen_uv_overrides.py` to mirror the root override list |
| `urdf_converter_cfg.py`, `mjcf_converter_cfg.py` | `.. caution::` on `merge_mesh` |
| `test_source_package_metadata.py` | literal provider assertion |
| `uv.lock` | regenerated |

**Drops the OpenUSD work-thread-limit workaround, repo-wide.** Every site was gated on
`TODO: Remove once usd-core>=26.5`; 3.0.0 supplies 26.08 to kit-less installs.

| File | Removed |
| --- | --- |
| `test_leapp_export_flow.py` | `_LEAPP_TEST_CPU_THREAD_LIMIT`, subprocess env injection, `--limit_cpu_threads` args, **and `test_openusd_thread_limit_is_set_before_subprocess_startup`** |
| `leapp_initialized_checkpoints.py` | `limit_cpu_threads=1` |
| `test_environments_newton.py` | `os.environ["PXR_WORK_THREAD_LIMIT"]` + orphaned `os` import |
| `test_environments_isaacsim_physx.py` | `limit_cpu_threads=1` |
| `rendering_test_utils.py` | `monkeypatch.setenv` + unused `monkeypatch` arg |
| `export_utils.py` | hidden `--limit_cpu_threads` CLI flag |

:warning: **Reviewer note:** the deleted assertion test was added to develop only days ago,
alongside converting this workaround to the env-var form. Removing it is a deliberate
reversal of that change, not an oversight — it asserts the limit reaches LEAPP subprocesses
and cannot pass once the limit is gone.

## Known limitation

**`merge_mesh` silently stops merging in kit-less installs. This is entirely upstream — no Isaac Lab code is involved.** `isaacsim-asset-isolated` 6.1 routes `UrdfConverterCfg.merge_mesh` / `MjcfConverterCfg.merge_mesh` through `usd-optimize` (`isaacsim/asset/importer/utils/impl/merge_mesh_utils.py` → `core.executeOperation(...)`). Two defects compound, both inside those wheels:

1. `usd-optimize` bundles its own OpenUSD 25.05 (`usd_optimize.libs`, 24 `libusd_*`, `pxrInternal_v0_25_5`, reached via `RPATH $ORIGIN/../../../usd_optimize.libs`) while its Python half imports the installed 26.08 `pxr`. One package spans two USD ABIs, so a stage-cache id minted by one is invisible to the other.
2. `merge_mesh_utils` catches only `ImportError` and **discards** `executeOperation`'s `(ok, msg, _)`. The error never leaves the wheel, so Isaac Lab cannot surface it.

Confirmed by calling the wheel's own `merge_mesh_utils._create_optimizer_context(stage)` with its exact `meshCleanup` config and reading the discarded return — no Isaac Lab code in the path: `ok=True` against develop's lock, `ok=False` against this branch's.

Isaac Lab's only role is forwarding `merge_mesh=True`, so there is no Lab-side fix; this PR documents the limitation instead.

**Ownership and upstream status.** `usd-optimize` is NVIDIA Omniverse ([NVIDIA-Omniverse/usd-optimize](https://github.com/NVIDIA-Omniverse/usd-optimize)); `isaacsim-asset-isolated` and `isaacsim-robot-schema` are the Isaac Sim team. [usd-optimize#2](https://github.com/NVIDIA-Omniverse/usd-optimize/issues/2) is this exact break — opened 2026-09-01, closed 2026-09-08. The `>=2.2,<3` cap is not a stale bound: it was added deliberately in 1.2.1 in response to usd-exchange 3.0.0, and the maintainer states there that **usd-optimize v1.3.0 will move to OpenUSD 26.08, "likely next month."** The override in this PR removes that guard.

Isaac Sim's `usd-exchange==2.3.0` pin is therefore not over-strict — it is what keeps the importer stack ABI-consistent while `usd-optimize` is on 25.05. The separable Isaac Sim defect is `merge_mesh_utils` discarding `executeOperation`'s return, which is what makes this fail silently instead of loudly.

Measured through `UrdfConverter` / `UrdfConverterCfg` on a URDF with two OBJ meshes on each of two links, counting `UsdGeom.Mesh` prims in the generated `payloads/`. Both environments come from `uv sync --frozen` against their own `uv.lock` — no `pip`, no hand-installed wheels — and carry identical package sets (`usd-optimize` 1.2.1, `isaacsim-asset-isolated` 6.1.0.0, `urdf-usd-converter` 0.3.2), with `usd-exchange` the only variable. 3 runs per arm, deterministic:

| env | `merge_mesh=False` | `merge_mesh=True` |
| --- | --- | --- |
| `develop`'s lock — usd-exchange 2.3.0, `pxr` 25.05 | 4 meshes / 16 points / 16 faces | **2 meshes** — merged |
| this branch's lock — usd-exchange 3.0.0, `pxr` 26.08 | 4 meshes / 16 points / 16 faces | **4 meshes** — unchanged, no error |

The `usd-optimize` call underneath returns `ok=True` 3/3 at 2.3.0 and `ok=False` 3/3 at 3.0.0. Symbols confirm the split: `usd_optimize.libs/libusd_usd-*.so` and `_usd_optimize_impl_core*.so` carry `pxrInternal_v0_25_5` and resolve through `RPATH $ORIGIN/../../../usd_optimize.libs`, while the installed `usd_exchange.libs/libusd_usd-*.so` carries `pxrInternal_v0_26_8`. `usd-optimize`'s Python half imports the installed 26.08 `pxr`, so one package spans two USD ABIs and the stage-cache id minted by one is invisible to the other.

The failure message is `usd-optimize`'s own shipped diagnostic, verified present in `libusd_optimize-2145aa6e.core.so`:

> Stage id … not found in UsdUtilsStageCache. The Python pxr bindings and the C++ core may be linked against different libusd builds — check that the pxr you import resolves to the same libusd_*.so the Usd Optimize core was built against.

**No existing test catches this.** `test_merge_mesh` (`source/isaaclab/test/sim/test_mjcf_converter.py:270`) asserts only that the USD file exists and the spawned prim is valid, never that anything merged, so a silent no-op passes it; it is also `@pytest.mark.isaacsim_ci`, selected only when `ISAACSIM_CI_SHORT=true` (`tools/conftest.py:1587`) — the Kit lane, where merging still works. There is no URDF `merge_mesh` test at all.

Documented as a `.. caution::` on both config fields and as a **Known limitation** bullet in the `isaaclab` changelog fragment. It clears when `usd-optimize` ships an OpenUSD 26.08 build; the override comment says to drop the override then. Only `usd-optimize` 1.1.0 and 1.2.1 exist on the index today, and only `isaacsim-asset-isolated` 6.0.1.0 and 6.1.0.0.

## Validation

**Which OpenUSD is live** (measured, same process before/after `AppLauncher`):

| Context | OpenUSD | Source |
| --- | --- | --- |
| Kit-less | **26.08** | the wheel |
| Inside Kit | **25.11** | Kit's `omni.usd.libs`; `usdex .so` mapped = **0** |

Kit shadows the wheel's USD at startup, so the pin only moves the kit-less `pxr`.

**Training smoke — 18/18 green.** 3 iters / 16 envs, `Isaac-Cartpole-Direct` (physics-only) and `Isaac-Cartpole-Camera-Direct` (renderer arms), asserted on fresh checkpoints. Both Isaac Sim lanes: source build at develop `e34059611a`, and Docker `latest-develop`.

| physics | renderer | src build | docker |
| --- | --- | --- | --- |
| isaacsim_physx | none / rtx / newton_renderer | ✅ ✅ ✅ | ✅ ✅ ✅ |
| newton_mjwarp | none / rtx / newton_renderer | ✅ ✅ ✅ | ✅ ✅ ✅ |
| ovphysx | none / ovrtx_renderer / newton_renderer | ✅ ✅ ✅ | ✅ ✅ ✅ |

**Converters — 85 tests + both CLIs.** Kit-less path (35) and Kit path (23 urdf + 12 mjcf + 15 mesh); `convert_urdf.py` and `convert_mjcf.py` verified by recomposing output (27 prims / 5 joints / 6 bodies from a repo URDF).

**Dependency audit.** Single `pxr` provider; only `usd-exchange` claims `pxr/` files; `usd-core` absent. `usd-exchange` now carries a uv override (above); `uv lock --check` clean, and the resolved-content delta against develop is exactly `usd-exchange 2.3.0 → 3.0.0` plus eight workspace-member versions develop's own lock was lagging — no third-party drift, no package added or removed.

**Not re-run after the develop merge.** The 18-arm training smoke and the 85 converter tests above predate the Isaac Sim 6.1 sync; `test_merge_mesh` in `source/isaaclab/test/sim/test_mjcf_converter.py` asserts only that the run succeeds and yields a spawnable USD, so it is expected to stay green against a silent no-op. Re-run under CI.

**Known, pre-existing — not introduced here.** `import usdex.core` *inside* a running Kit process aborts (two USD ABIs in one process). Identical at both versions, so the bump only changes the namespace in the message:

```
3.0.0 → SIGABRT  pxrInternal_v0_26_8__pxrReserved__::pxr_boost::python::error_already_set
2.3.0 → SIGABRT  pxrInternal_v0_25_5__pxrReserved__::pxr_boost::python::error_already_set
```

Nothing in the training or converter paths does that import, which is why all 18 arms pass.

**Follow-up for the importer owners (not this PR).** `urdf_usd_converter` and `mujoco_usd_converter` still call `addDiffuseTextureToPreviewMaterial`, deprecated in usdex 3.0 and still present in the newest wheels (0.3.3 / 0.5.0). It works, but renames the authored shader prim — same URDF, provider the only variable:

| usd-exchange | Shader prim | Interface input |
| --- | --- | --- |
| 2.3.0 | `DiffuseTexture` | `inputs:DiffuseTexture` |
| 3.0.0 | `ColorTexture` | `inputs:ColorTexture` |

They should move to `addColorTextureToPreviewMaterial` before usdex removes the deprecated call.

## Checklist


AntoineRichard (commented):
Thread-limit audit: the PR removes all of the OpenUSD race-workaround limits that existed at its current merge base. The remaining `PXR_WORK_THREAD_LIMIT` in `AppLauncher` is the separate distributed/multi-GPU resource-partitioning path, so keeping it is correct.

One rebase note: `develop` merged #7229 after this branch, adding an earlier startup-time `PXR_WORK_THREAD_LIMIT=1` in `source/isaaclab_rl/test/export/test_leapp_export_flow.py::_run_checked`, together with its constant/TODO and a dedicated assertion. Since this PR is currently behind and conflicts with `develop`, please handle that site explicitly during conflict resolution if the intent remains complete retirement.

Also, the dependency bump only upgrades kit-less OpenUSD; Kit-backed tests still use the Kit-provided OpenUSD. Thus removal of their limits rests on the reported test evidence rather than on the 26.08 provider bump. This is non-blocking, but worth keeping in mind when reviewing CI for the probabilistic parser failure.

hujc7:
run-ci

hujc7:
@kellyguo11 local testing is working, it seems lab side is not constrained by sim side bump.

fatimaanes (approved):

mataylor-nvidia:
Looks like `usd-core` was dropped from the pyproject.toml in favor of `usd-exchange` as long as `usd-exchange` is able to select the later package with the fix it should be the correct behavior

mataylor-nvidia (approved):

hujc7:
run-ci

hujc7:
**Blocked by `usd-optimize` under Isaac Sim 6.1**

The develop sync to `isaacsim-asset-isolated` 6.1.0.0 invalidated this PR's premise. 6.1 newly depends on `usd-optimize`, the only compiled OpenUSD consumer in the importer tree — every other package there is pure Python and version-tolerant.

`usd-optimize` 1.2.1 is built against OpenUSD 25.05: its core has 19 undefined `pxr` symbols, all `pxrInternal_v0_25_5`. `usd-exchange` 3.0.0 ships 26.08 and exports **0** symbols in that namespace, so the loader can no longer dedupe by SONAME (`libusd_ar-f98cb08e.so` vs `-983356b2.so`) and both USD runtimes load. `executeOperation` then fails on every call with a stage-cache miss.

Effect, measured through `UrdfConverter` on lock-derived envs, 3 runs per arm:

| `usd-exchange` | `merge_mesh=True` |
| --- | --- |
| 2.3.0 (25.05) | 2 meshes — merged |
| 3.0.0 (26.08) | 4 meshes — unchanged, **no error** |

It's silent because `merge_mesh_utils` guards only `ImportError` and discards `executeOperation`'s return.

Both upstream constraints (`usd-optimize`'s `usd-exchange>=2.2,<3`, Isaac Sim's `==2.3.0`) are correct; forcing past them is what breaks it. Handled here by excluding `usd-optimize` via an override, so `merge_mesh` logs the miss instead of silently skipping and only one OpenUSD loads. Kit-backed runs are unaffected, and kit-less `merge_mesh` has never worked in a released Isaac Lab — 6.0 called a Kit-only extension, and 6.1 landed on develop yesterday.

Tracked upstream at [usd-optimize#2](https://github.com/NVIDIA-Omniverse/usd-optimize/issues/2); the maintainer states v1.3.0 will move to 26.08. The override comes out then.

hujc7:
run-ci

PR #7216, opened 2026-08-20
Fix colorized semantic segmentation normalization

AntoineRichard:
# Description

Fixes `mdp.image(normalize=True)` returning colorized semantic-segmentation observations as `uint8`. The normalizer now recognizes the renderer contract for this output (RGBA `uint8`) and returns the same normalized `float32` representation used for RGB-like images. Raw `int32` semantic-ID maps remain unchanged.

No new dependencies.

## Type of change

## Screenshots

Not applicable.

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/normalize-colorized-semantic-segmentation.rst
+++ b/source/isaaclab/changelog.d/normalize-colorized-semantic-segmentation.rst
@@ -1,0 +1,5 @@
+Fixed
+^^^^^
+
+* Fixed :func:`isaaclab.envs.mdp.image` so colorized semantic-segmentation observations are
+  converted to normalized ``float32`` tensors when ``normalize=True``.
--- a/source/isaaclab/isaaclab/utils/images.py
+++ b/source/isaaclab/isaaclab/utils/images.py
@@ -52,11 +52,13 @@ def normalize_camera_image(
 
     Dispatch (in order of check):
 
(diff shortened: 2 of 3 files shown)


kellyguo11 (approved):

mmichelis (commented):
In practice the normalize might be called on warp arrays as well, could they? These changes currently only work for torch tensors, is that intended?

AntoineRichard:
> In practice the normalize might be called on warp arrays as well, could they? These changes currently only work for torch tensors, is that intended?

No that function clearly looks torch only for now:

```python
    if is_rgb_like(data_type):
    if is_rgb_like(data_type) or (data_type == "semantic_segmentation" and images.dtype == torch.uint8):
        if images.dtype == torch.uint8 and images.ndim == 4 and images.is_contiguous():
            return normalize_image_uint8(images, channel_dim=channel_dim, out=out)
        # PyTorch fallback for callers that pre-floated or pass a strided view.
        resolved_channel_dim = channel_dim + images.ndim if channel_dim < 0 else channel_dim
        spatial_dims = tuple(d for d in range(1, images.ndim) if d != resolved_channel_dim)
        images = images.float() / 255.0
        images -= torch.mean(images, dim=spatial_dims, keepdim=True)
        return images
    if is_depth_like(data_type):
        images[images == float("inf")] = 0
        return images
    if is_normals_like(data_type):
        return (images + 1.0) * 0.5
    return images
```

mmichelis (approved):

PR #7228, opened 2026-08-20
Save each test's OVRTX renderer output as a CI artifact

mataylor-nvidia:
# Description

The OVRTX renderer log reaches a diagnosis two ways today, and both are bounded, because what they report lands in the job log and in the JUnit XML: the autouse fixture in `tools/ovrtx_log.py` replays the byte range a test appended, and the orchestrator quotes a tail into the report of a run that died before it could replay anything. Both also cover the log and nothing else, so a dump a crashing renderer leaves beside it never leaves the container.

This saves each test's renderer output per test and uploads it as a job artifact.

- **`tools/ovrtx_log.py`** — a new `save_output()` writes the test's own byte range of the log into `<artifact dir>/<test name>.<attempt>/`, then copies the rest of the renderer's directory — the dump a crash leaves beside the log — whole. Subdirectories are skipped: the renderer logs into the system temp directory, which every other process on the machine also writes to. The attempt number keeps a retry from overwriting the attempt that failed.

  The log is sliced per test rather than copied whole because the renderer appends to one file for the lifetime of the process. Copying it whole would credit the second test with the first's output and every test after them with both — an artifact growing as the square of a run whose logs are already the large part. A test that adds nothing to that range saves nothing, which covers both a test that never builds a renderer and a skip that follows one that did.

- **`tools/conftest.py`** — names that directory per pass through `ISAACLAB_OVRTX_LOG_DIR`, as `tests/ovrtx-logs` under the workspace root. Absolute, for the journal's reason: the save happens in a fixture, so a test that changed directory would otherwise leave its artifact under the temporary cwd. Unset by default, so a plain local `pytest` run still writes nothing beyond the replay.

- **`tools/conftest.py`, abnormal termination** — the test a crash, hang, or timeout killed never reaches the teardown that saves, which leaves the one test in the run whose output anybody is going to read as the one test missing from the artifact. `_make_crash_pass_result` now saves on its behalf, named after the test the crash journal blames so it sits beside the tests that saved their own. Saved whole, since the per-test offsets went with the process, and `OSError`-suppressed, since a report matters more than an artifact.

- **`run_tests.sh` / `run-tests/action.yml`** — copies that tree out of the container alongside the JUnit reports and uploads it as `ovrtx-logs-<container-name>`.

The report cap goes from 64 KiB to 1 MiB while here. It exists to keep a multi-megabyte log out of the job log and the JUnit XML, and the rendering tests it was built for log a few hundred kilobytes each, so the old size cut a whole test's log down to its tail for no benefit. The saved copy covers the runs that outgrow even the new cap. The two existing tests that exercise the cap now count their filler from `LOG_LIMIT_BYTES` rather than a literal, so raising it again cannot quietly turn them into tests of a log that fits inside it.

## Type of change

## Screenshots

Not applicable.

## Testing

Six tests added.

`source/isaaclab/test/cli/test_ovrtx_log.py`, each run against a nested pytest session driving a fake renderer:

- each test's log is saved under a directory of its own name, the replay still happens alongside it, and one test's saved log holds none of another's output
- a dump written beside the log is saved with it, while a sibling subdirectory is left alone
- a test that never builds a renderer saves nothing
- a test that adds nothing to a log an earlier test left behind saves nothing

`source/isaaclab/test/cli/test_test_orchestrator_result_handling.py`:

- the pass hands its subprocess an absolute `ISAACLAB_OVRTX_LOG_DIR` under the reports directory
- a crash (`-11`) and a timeout (`-9`, `kill_reason="timeout"`) each save the blamed test's renderer log, named after that test — parametrized over both

Run locally on Windows: `10 passed, 1 skipped` for `test_ovrtx_log.py` (the skip is the pre-existing POSIX-only `/dev/stdout` case) and `14 passed` for `test_test_orchestrator_result_handling.py`.

## Checklist


tools/ovrtx_log.py
```diff
@@ -35,13 +42,24 @@
 loading this module does not import ``isaaclab_ov``.
 """
 
-LOG_LIMIT_BYTES = 64 * 1024
+LOG_LIMIT_BYTES = 1024 * 1024
```
mataylor-nvidia, line 48: bumped length of ovrtx logs printed to stdout

tools/ovrtx_log.py
```diff
@@ -97,13 +140,32 @@ def pytest_configure(config):
...
     start = log_size(LOG_PATH)
     yield
-    if section := format_log_section(LOG_PATH, request.node.name, start=start):
+    # The log exists only once the renderer has written to it, while the env var is set for every test in
+    # the run, so a test that never built one has nothing to save.
+    if (directory := os.environ.get(LOG_DIR_ENV_VAR)) and os.path.exists(LOG_PATH):
+        destination = _unused_dir(directory, _slugify(label))
+        # The renderer writes more than its log -- a crash leaves a dump beside it -- so everything in its
+        # directory goes, bar the subdirectories a shared temp directory collects from other processes.
+        for entry in os.scandir(os.path.dirname(LOG_PATH)):
```
mataylor-nvidia: scan the directory for all traces and dumps, copying them after each test

nvsekkin (approved):

tools/conftest.py
```diff
@@ -336,6 +345,14 @@ def _make_crash_pass_result(
             f" blamed {culprit or 'session shutdown (all tests completed teardown)'}"
         )
 
+    # The blamed test died before the fixture that saves its renderer output could run, leaving the one
+    # test in the run whose output is worth reading as the one missing from the artifact. Named after the
+    # test alone, as the fixture names its own, so the two sit together. Saved whole because the per-test
+    # offsets went with the process, and suppressed because a report matters more than an artifact.
+    blamed = culprit.rpartition("::")[2] if culprit else pass_file_label
+    with contextlib.suppress(OSError):
+        ovrtx_log.save_output(os.path.abspath(OVRTX_LOG_DIR), blamed)
```
nvsekkin, line 354: I believe `OVRTX_LOG_DIR` resolves to `/tmp`, which may have other unrelated files. we may want to filter what we are actually saving here (eg. *.nv-gpudmp, *.nvdbg, *.dmp.zip, *.dmp.zip.toml, *.nvcudmp, and crash_*.txt)

AntoineRichard:
🤖 **AI-assisted review** (batch triage of open CI/infra PRs, with automated code-review/silent-failure/test-coverage passes on a checkout of this branch)

The mechanism is sound — per-test `start` offsets prevent cross-test misattribution, `_unused_dir` keeps retries from overwriting failed attempts, and the orchestrator's save-on-behalf covers the process that died before its fixture ran. The failing `rendering-correctness-kitless (legacy)` check is the known franka_soft flake (this branch passed 7/7 one commit earlier with the feature present), and the uploaded artifact was 432 KB compressed, so size/retention is empirically fine.

Blocking asks:
1. **nvsekkin's post-approval request is still open:** filter what gets copied out of `/tmp` — `tools/ovrtx_log.py:162-165` currently takes *every* top-level regular file in the system temp dir (any process's), unfiltered and uncapped, into each per-test directory, and `entry.is_file()` follows symlinks.
2. **Guard the fixture save:** `tools/ovrtx_log.py:196-197` calls `save_output` unguarded while the identical call at `tools/conftest.py:353-354` is wrapped in `contextlib.suppress(OSError)`. As an autouse fixture this runs for every test in the repo, so an ENOSPC/EACCES would error otherwise-passing teardowns — and since the save precedes the replay `print` at :199, it also suppresses the diagnostic the report shows.

Non-blocking findings worth a look:
- Artifact dirs are keyed on the bare test name while `tools/conftest.py:822-828` deliberately slugs full paths because concurrent shards share `/workspace/isaaclab` — same-basename tests in sibling shards collide, making `.N` ambiguous between retry and other-shard (with a FileExistsError TOCTOU in `_unused_dir`).
- `tools/conftest.py:352`: when the journal rebuilds cleanly and `culprit is None`, the fallback still saves the whole process log at `start=0` under `<file>.py.0` — the concatenated, misattributed artifact this design otherwise eliminates, duplicating what the fixtures already saved.
- The 64 KiB→1 MiB cap feeds `format_log_section` into JUnit XML `details` that get uploaded and forwarded to omni-github — a multi-crash run embeds ~1 MB per crashed pass.
- Untested paths: attempt numbering past `.0`, `_slugify` on parametrized names, and the `size < start` re-open branch. Also the PR description says node ID but the code uses `request.node.name`.

tools/conftest.py
mataylor-nvidia: Each test invocation creates a new `/tmp...` directory. My thought is that if the artifacts ovrtx change name etc we will still pick up all the artifacts it produces

nvsekkin (approved):

pbarejko (approved):

kellyguo11 (approved):

PR #7235, opened 2026-08-21
Adds distinct exit codes for crashed, hung and failing test runs

myurasov-nv:
# Description

Every failure mode in the individual-test orchestrator exits `1`, so a crashed process, a hung test, and a failing assertion cannot be told apart from the exit code. The orchestrator already separates them internally: it counts `Failing`, `Crashed`, `Startup Hang`, and `Timeout`, and prints each in the result summary. This carries that distinction into the exit code and prints the resolved code with a label.

`1` still means failing assertions. A crashed process is `20`, a timeout `21`, and a startup hang `22`. When a run hits more than one, the code reports the outcome that proved the least, so a crash outranks a timeout, which outranks a startup hang. `0` is unchanged, and callers that only check for a non-zero code are unaffected.

| Before | After |
| ------ | ----- |
| A lost GPU, a 1000 second hang, and a pixel mismatch all end in `Process completed with exit code 1` | The same 3 runs exit `20`, `21`, and `1`, each printed as `Exit Code: N (label)` in the summary |

## Reproduction

1. Run a suite through the orchestrator and read the exit code:

   ```
   ./isaaclab.sh -p -m pytest --ignore=tools/conftest.py tools -v --junitxml=tests/report.xml; echo "exit=$?"
   ```

2. Make one file reach its hard timeout, by lowering the timeout for that file or running a test that sleeps past it.

   Observed before this change: the summary reports `Timeout: 1` while the run reports `exit=1`, the same code a failing assertion produces.

3. Repeat with this change.

   Observed: the summary reports `Exit Code: 21 (timeout)` and the run reports `exit=21`, while a run whose only failures are assertions still reports `exit=1`.

## Type of change

## Screenshots

Not applicable; the change has no rendered output.

## Checklist

- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`: not run; no dev setup on this machine. The diff was checked by hand for line length and whitespace, and `python3 -m py_compile tools/conftest.py` passes
- [ ] I have made corresponding changes to the documentation: not applicable; each code is documented where it is defined and printed in the run summary
- [ ] I have added tests that prove my fix is effective or that my feature works: not added; the counters are computed inside `pytest_sessionstart`, and `tools/conftest.py` takes over any pytest session that loads it, so a test in the same directory cannot drive this path in-process. `resolve_exit_code` was instead checked over all 16 combinations of the 4 counters, confirming that each mode maps to its own code and that zero versus non-zero matches the previous expression exactly
- [ ] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that): not applicable; no package under `source/` is touched

### DIFF
--- a/tools/conftest.py
+++ b/tools/conftest.py
@@ -80,6 +80,48 @@ kill it.  The test results are taken from the report file (pass/fail), not
 from the kill.
 """
 
+EXIT_TESTS_FAILED = 1
+"""Exit code for a run that completed with at least one failing assertion."""
+
+EXIT_CRASHED = 20
+"""Exit code for a run where a test process died without writing its report."""
+
+EXIT_TIMEOUT = 21
+"""Exit code for a run where a test reached its hard timeout."""
+
+EXIT_STARTUP_HANG = 22
+"""Exit code for a run where a test never finished starting up."""
+
+EXIT_CODE_LABELS = {
+    0: "all tests passed",
+    EXIT_TESTS_FAILED: "test failures",
+    EXIT_CRASHED: "crashed process",
+    EXIT_TIMEOUT: "timeout",
+    EXIT_STARTUP_HANG: "startup hang",
+}
+"""Label for each exit code, printed with the result summary."""
+
+
+def resolve_exit_code(num_failing, num_timeout, num_crashed, num_startup_hang):
+    """Return the exit code for a finished run, based on how it failed.
+
+    A crashed process, a hang and a failing assertion have different owners, so
+    each reports its own code rather than a shared ``1``; the code alone is then
+    enough to route a red job.  When a run hits more than one, the code reports
+    the outcome that proved the least: a process that died never reached the
+    assertion a failing test did.  ``1`` still means failing assertions, so
+    callers that only check for a non-zero code are unaffected.
+    """
+    if num_crashed:
+        return EXIT_CRASHED
+    if num_timeout:
+        return EXIT_TIMEOUT
+    if num_startup_hang:
+        return EXIT_STARTUP_HANG
+    if num_failing:
+        return EXIT_TESTS_FAILED
+    return 0
+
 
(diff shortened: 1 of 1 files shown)


myurasov-nv:
Dear Isaac Lab Review Bot!

`resolve_exit_code` now carries `int` annotations on its 4 counters and on its return value, plus an Args block describing each counter and a Returns line stating the precedence, in commit 30f2613. That matches the annotated helpers already in this file, such as `_merge_pass_status` and `_format_test_file_results`.

On the contract change: it is intended, and it is limited to the 3 modes that previously could not be told apart. The consumers I checked do not match on exactly 1, since `run_tests.sh` compares the container status against 0 and returns it unchanged, and the composite action fails on any non-zero step. I did not audit consumers outside this repository.

myurasov-nv:
Existing CI failures:

| Failure | Status | Likely Reason |
| --- | --- | --- |
| `rendering-correctness-kitless (legacy)` | Not from this PR | 37 golden-image comparison failures on `franka_cable`, `franka_cloth` and `franka_soft`. The same check is currently failing on open PRs 7233, 7228 and 7220. |

The JUnit report for the failing run (32447613867) holds 314 cases with 37 failures, 0 errors, and no timeouts or crashes. Every failure is a pixel-percentage or SSIM threshold on a golden image. This change cannot affect that: it alters the process exit code and one line of the run summary, nothing in the render path.

Worth noting for this PR specifically: those 37 failures are ordinary assertion failures, so they resolve to exit code 1, the value the previous expression also returned.

`develop` is green (run 32449678669), but the `push` path runs the reduced `.github/test-subsets/postmerge-rendering.toml` subset and so does not exercise these files. Nothing was changed on this branch in response.

PR #7241, opened 2026-08-21
[CI] Report cumulative retry wall time

StafaH:
# Description

Include every subprocess attempt in the per-file wall-clock metric. Startup-hang, timeout, and fresh-process retries previously replaced the prior attempt duration, which hid substantial CI time from the final summary.

This changes only test-runner reporting; test selection, execution, and coverage are unchanged.

## Type of change

## Validation

- `uv run python -m pytest source/isaaclab/test/cli/test_test_orchestrator_result_handling.py`
- `uv run isaaclab -f`

## Checklist

### DIFF
--- a/source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
+++ b/source/isaaclab/test/cli/test_test_orchestrator_result_handling.py
@@ -341,6 +341,44 @@ def test_shutdown_hang_after_report_is_not_a_failure(monkeypatch, tmp_path: Path
     assert not was_failure
 
 
+def test_startup_retry_wall_time_includes_every_attempt(monkeypatch, tmp_path: Path) -> None:
+    """The reported wall time must include startup attempts discarded by a successful retry."""
+    orchestrator = _load_orchestrator_module()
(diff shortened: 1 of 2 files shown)


AntoineRichard (approved):
LGTM.

AI comment: 
```
  - Moderate — Preserve the final-attempt duration in shutdown-hang diagnostics. At .worktrees/pr-7241-review/tools/conftest.py:855, wall_time is replaced with the
    cumulative duration. The existing warning at line 932 then reports that cumulative value as when the final subprocess was “killed after.” For example, an 8-second
    startup-hang retry followed by a 2-second shutdown hang is logged as “killed after 10s,” although that subprocess was killed after 2 seconds. The new startup-retry
    test ends normally, while the existing shutdown-hang test has no retry, so neither covers this combination. Keep separate attempt_wall_time and total_wall_time
    values: use the former for attempt-specific diagnostics and the latter in status/summary reporting.
```

mataylor-nvidia (approved):

pbarejko (approved):

PR #7243, opened 2026-08-21
[CI] Run lightweight task tests concurrently

StafaH:
# Description

Run an explicit allowlist of lightweight `isaaclab_tasks` test files concurrently with four workers. Every file still runs in its own pytest subprocess, preserving the existing process-isolation boundary and the same collected tests.

Safety boundaries:

- only explicitly listed task files are eligible
- allowlisted files are checked for cold-camera and device-split requirements
- AppLauncher, camera startup, device-split, queue-driven, and unclassified tests remain sequential
- setting the worker count to one retains the original execution order

Using the wall times from PR #6299, the eligible files account for about 37–85 seconds per shard and should complete in roughly the duration of the slowest eligible file. The expected critical-path reduction is approximately 30–74 seconds per shard.

No existing test content, test selection, assertions, or coverage was changed.

## Type of change

## Validation

- `uv run python -m pytest source/isaaclab/test/cli/test_test_orchestrator_result_handling.py` (12 passed)
- ran all 27 allowlisted files with four independent pytest processes
- retried `test_xr_camera_feed_presets.py` with `uv run --extra teleop ...` (5 passed)
- `bash -n .github/actions/run-tests/run_tests.sh`
- `uv run isaaclab -f`

## Checklist

- [x] I have added tests for concurrency, exactly-once execution, and allowlist safety

### DIFF
--- a/tools/conftest.py
+++ b/tools/conftest.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+import concurrent.futures
 import contextlib
 import logging
 import os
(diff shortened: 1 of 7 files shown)


AntoineRichard (approved):
I don't have a blocking opinion, but I could see a risk around memory. Maybe a bit concerned about RAM usage?

tools/test_settings.py
```diff
@@ -124,6 +124,41 @@
 ]
 """A list of tests to skip in CI (see conftest.py)."""
```
mataylor-nvidia, line 126: Is there a way to verify these are "safe" to run in parallel?

tools/conftest.py
```diff
@@ -1088,87 +1185,64 @@ def run_individual_tests(test_files, workspace_root, ci_marker, test_node_ids_by
...
+                )
+                for test_file in parallel_files
+            ]
+            for test_file, future in zip(parallel_files, futures):
+                record_result(test_file, future.result())
+    for test_file in sequential_files:
+        test_content = _read_test_content(test_file)
         is_cold_cache_test = not cold_cache_applied and "enable_cameras=True" in test_content
-        if is_cold_cache_test:
-            timeout += COLD_CACHE_BUFFER
```
mataylor-nvidia, line 1125: Is there no longer timeout

tools/test_settings.py
mataylor-nvidia: Any way we can re organize the tests better? We use the `conftest` as a harness for running all the tests is there any better way to clean this up. Maybe have different conftests  and split the tests across different directories based on the testing setup and requirements?

PR #7259, opened 2026-08-21
Fix JointPositionToLimitsAction IO descriptor

sylvesterkaczmarek:
# Description

Fixes `JointPositionToLimitsAction.IO_descriptor` raising `AttributeError`.

The descriptor implementation references `self._offset`, copied from the generic joint-action descriptor path. `JointPositionToLimitsAction` does not define an offset and its configuration has no offset field, so descriptor export can fail even though the action term itself is valid.

The descriptor now reports `offset = 0.0`, matching the action term's actual processing (`processed = raw * scale` before optional clipping and limit rescaling).

## Type of change

- Bug fix

## Validation

- Added a unit regression test that constructs the action term and reads its IO descriptor.
- Verifies the descriptor reports zero offset, resolved joint names, and action shape.
- Before the fix, descriptor access raises `AttributeError`.

## Checklist

- [x] No new dependencies

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-joint-limits-io-offset.patch.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-joint-limits-io-offset.patch.rst
@@ -1,4 +1,0 @@
-Fixed
------
-
-* Fixed ``JointPositionToLimitsAction.IO_descriptor`` raising when reporting its implicit zero offset.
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** Zero offset is the correct implicit descriptor value and preserves processing and EMA behavior.

**Requested changes:** Rename the changelog fragment from .patch.rst to .rst.

**Test scope:** One focused regression passed and fails against the base bug.

**Documentation / AI skills:** No additional documentation update is required: existing descriptor/API docs remain accurate. No AI-skill update is required.

PR #7260, opened 2026-08-21
Fix NonHolonomicAction IO descriptor without clipping

sylvesterkaczmarek:
# Description

Fixes `NonHolonomicAction.IO_descriptor` for the default `clip=None` configuration.

`_clip` is only allocated when a clip dictionary is configured, but the descriptor previously accessed `self._clip` unconditionally. As a result, exporting or inspecting the action descriptor could raise `AttributeError` for an otherwise valid default action configuration.

The descriptor now reports `clip = None` when clipping is disabled and retains the existing tensor value when clipping is configured.

## Type of change

- Bug fix

## Validation

- Added a unit regression test using the default no-clip configuration.
- Verifies descriptor shape, resolved joint names, and `clip is None`.
- Before the fix, descriptor access raises `AttributeError`.

## Checklist

- [x] No new dependencies

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-nonholonomic-io-clip.patch.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-nonholonomic-io-clip.patch.rst
@@ -1,4 +1,0 @@
-Fixed
------
-
-* Fixed ``NonHolonomicAction.IO_descriptor`` when clipping is left at its default ``None`` value.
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** Guard correctly avoids uninitialized _clip access when clip=None and preserves configured clipping.

**Requested changes:** Rename fragment to .rst and run focused CI.

**Test scope:** Focused regression is sufficient; run it in dependency-complete CI.

**Documentation / AI skills:** No additional documentation update is required: existing clip docstring remains accurate. No AI-skill update is required.

PR #7261, opened 2026-08-21
Fix multi-joint clipping for binary joint actions

sylvesterkaczmarek:
# Description

Fixes per-joint clipping for binary joint actions when one binary input controls multiple joints.

`BinaryJointAction` has a one-dimensional raw action, but maps that value to one processed command per resolved joint. Clip ranges are resolved against those joint names and applied to `processed_actions`, yet the clip tensor was allocated with `action_dim == 1`. A multi-joint gripper therefore fails when a clip entry resolves to joint index 1 or later.

The clip tensor now uses the processed joint count (`_num_joints`), matching the command it clamps.

## Type of change

- Bug fix

## Validation

- Added a unit regression test with a two-joint gripper and separate clip ranges.
- The test verifies both joints are clipped correctly.
- Before the fix, assigning the second joint's clip range fails.

## Checklist

- [x] No new dependencies

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-binary-joint-clip.patch.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-binary-joint-clip.patch.rst
@@ -1,0 +1,4 @@
+Fixed
+-----
+
+* Fixed per-joint clipping for binary joint actions when one binary input controls multiple joints.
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** The clip tensor must match resolved joints rather than the one-dimensional policy input; the one-line fix is correct.

**Findings:** Invalid .patch.rst changelog filename; focused test has no hosted unit-test result.

**Requested changes:** Rename fragment and run the focused unit test.

**Test scope:** Two-joint regression with distinct bounds is minimal and sufficient.

**Documentation / AI skills:** No additional documentation update is required: no public configuration contract changed. No AI-skill update is required.

PR #7262, opened 2026-08-21
Allow IO descriptors on functions without docstrings

sylvesterkaczmarek:
# Description

Allows `@generic_io_descriptor` to decorate functions that do not have a docstring when no explicit descriptor description is supplied.

The decorator currently tries to populate a missing description with `" ".join(func.__doc__.split())`. For a valid Python function without a docstring, `func.__doc__` is `None`, so decoration raises `AttributeError` before the function can be used.

The descriptor schema already permits `description=None`. The decorator now copies the function docstring only when one exists, otherwise leaving the description unset.

## Type of change

- Bug fix

## Validation

- Added a `pytest.mark.unit` regression test using a function with no docstring.
- Verifies decoration succeeds, `description` remains `None`, and the wrapped function still executes normally.
- Existing behavior for functions with docstrings or explicit descriptions is unchanged.

## Checklist

- [x] No new dependencies

### DIFF
--- a/docs/source/policy_deployment/01_io_descriptors/io_descriptors_101.rst
+++ b/docs/source/policy_deployment/01_io_descriptors/io_descriptors_101.rst
@@ -141,7 +141,8 @@ We are also attaching an IO descriptor to this term. The IO descriptor is define
 
 The ``@generic_io_descriptor`` decorator is a special decorator that is used to attach an IO descriptor to a custom observation term.
 It takes arbitrary arguments that are used to describe the observation term, in this case we provide extra information that could be
-useful for the end user:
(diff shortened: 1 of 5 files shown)


AntoineRichard (commented):

**Assessment:** Guard correctly preserves explicit descriptions and leaves description=None without a function docstring.

**Findings:** Invalid .patch.rst filename/content convention; public decorator docs omit the new fallback.

**Requested changes:** Rename/format fragment and add one sentence to the decorator docs.

**Test scope:** One focused decoration/execution test is sufficient.

**Documentation / AI skills:** Documentation update required: state that missing docstrings leave description=None. No AI-skill update is required.

PR #7264, opened 2026-08-21
Remove stray ActionManager descriptor output

sylvesterkaczmarek:
# Description

Removes stray debug output from `ActionManager.get_IO_descriptors`.

Successful descriptor export currently executes `print(item["export"])` for every action term, so normal export writes unrelated `True` / `False` lines to stdout. The export flag is already consumed immediately afterwards and does not need to be printed.

This removes only that debug print. The existing error message for descriptors that actually fail to build is unchanged.

## Type of change

- Bug fix

## Validation

- Added a `pytest.mark.unit` regression test using a valid action descriptor.
- The test captures stdout and verifies successful descriptor export is silent.
- The exported descriptor content is still checked.

## Checklist

- [x] No new dependencies

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-action-descriptor-stdout.patch.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-action-descriptor-stdout.patch.rst
@@ -1,0 +1,3 @@
+Bug Fixes
+---------
+* Removed stray debug output from action IO descriptor export.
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** Minimal deletion removes unintended stdout without affecting export filtering or formatting.

**Requested changes:** Rename the changelog fragment to .rst.

**Test scope:** One captured-stdout regression is sufficient.

**Documentation / AI skills:** No additional documentation update is required: no public API or usage change. No AI-skill update is required.

PR #7268, opened 2026-08-21
Fix camera aperture offset warning for negative offsets

sylvesterkaczmarek:
# Description

Fixes the unsupported camera aperture-offset warning in `convert_camera_intrinsics_to_usd()`.

The current condition applies `abs()` to the result of a boolean expression rather than to the signed principal-point offsets themselves. As a result, positive deviations of `c_x` / `c_y` from image center trigger the warning, while equally valid negative deviations silently do not.

This change applies the absolute value to each offset before comparing it with the tolerance, so unsupported principal-point offsets are detected symmetrically on either side of image center.

## Type of change

- Bug fix

## Validation

- Added `pytest.mark.unit` regression coverage for negative horizontal and vertical principal-point offsets.
- Added a centered-principal-point case verifying that no warning is emitted.
- Source change is one condition only.
- Full Isaac Lab test suite was not run locally.

## Checklist

- [x] No new dependencies

### DIFF
--- a/source/isaaclab/changelog.d/sylvesterkaczmarek-camera-aperture-offset-warning.patch.rst
+++ b/source/isaaclab/changelog.d/sylvesterkaczmarek-camera-aperture-offset-warning.patch.rst
@@ -1,1 +1,0 @@
-Fixed ``convert_camera_intrinsics_to_usd`` so unsupported camera aperture offsets are detected on either side of the image center.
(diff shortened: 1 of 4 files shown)


AntoineRichard (commented):

**Assessment:** Predicate fix is correct and tests cover negative and centered offsets.

**Findings:** Invalid .patch.rst filename and fragment content lacks a recognized Fixed heading/bullet.

**Requested changes:** Rename to .rst; format as Fixed/carets/bullet; run changelog gate.

**Test scope:** Focused predicate tests are sufficient; do not expand them.

**Documentation / AI skills:** No additional documentation update is required: changelog is the appropriate release-note surface. No AI-skill update is required.

PR #7053, opened 2026-08-12
feat: Enable scene partitioning by default and Use Isaac Sim alpha.50

nvsekkin:
# Description

- Enable per-environment Isaac RTX scene partitioning by default through `IsaacRtxRendererCfg`.
- Enable the all-environment spectator view before RTX startup only for an active viewport, Kit visualization, recording, livestreaming, or XR.
- Keep regular headless camera-sensor training isolated.
- Add per-instance partition ownership to environment-owned visualization markers and built-in task markers.
- Restore partition-isolation coverage and refresh affected Kit-tiled goldens.
- Use Isaac Sim alpha.50, whose Kit 360924 pin contains the validated NVBug 6566677 fix

## Scope

Marker ownership covers generic MDP commands and core task markers, including lift point clouds and experimental reorientation. Sensor debug markers, teleop visualizers, contrib commands, and terrain importer markers remain out of scope.

## Type of change

- Dependency update
- Rendering behavior change
- Test update

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/)
- [x] I have made corresponding documentation and changelog changes
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [x] I have added my name to `CONTRIBUTORS.md` or my name already exists there

PR #7270, opened 2026-08-21
Keep partitioned environments visible in viewport and XR views

rwiltz:
#Description
  
  `AppLauncher` enables the all-environment spectator view by passing
  `--/rtx/scenePartitioning/showAllPartitionsByDefault=true` whenever a Kit visualizer,
  viewport, recording, livestream, or XR is requested. The Kit builds shipped with the
  pinned Isaac Sim (6.0.1.0, Kit `110.1.2+production.326809`) do not implement that
  setting — it appears nowhere in the `omni.hydra.rtx` binaries.

  Since per-environment scene partitioning became the default in #7053, RTX therefore
  culls every prim carrying `omni:scenePartition` from cameras without a matching token.
  The Kit viewport and the XR view carry no token, so partitioned environments vanish from
  both while environment-scoped camera sensors still render them. RTX honors the token
  only for non-instanced geometry, so the failure looks selective: on
  `IsaacContrib-PickPlace-Locomanipulation-G1-Abs` the robot and packing table (USD
  instance proxies) render, while the steering wheel and bin (plain meshes) appear only in
  the XR picture-in-picture feed.

  `KitVisualizer._apply_viewport_camera_scene_partition()` has a working fallback that
  tags the viewport camera with `env_0`, but it is skipped because the ineffective setting
  reads back as `true`. Nothing tags the XR camera at all.

  This defaults per-environment scene partitioning off for launches that render a view
  without a partition token, in the same `AppLauncher` branch that already decides visual
  output intent. Partitioning only isolates per-environment camera renders, which a
  spectator launch does not need. The Kit argument is still passed, so relaxing the gate
  is a one-line change once the Isaac Sim bump #7053 assumed (alpha.50 / Kit 360924) lands.

  Precedence is unchanged: an explicit `ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION`
  or a task-set `IsaacRtxRendererCfg.enable_scene_partitioning` still wins.

  Reproducer:

      uv run --extra teleop isaaclab teleop run \
        --task IsaacContrib-PickPlace-Locomanipulation-G1-Abs --visualizer kit --xr

  ## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/app/app_launcher.py
+++ b/source/isaaclab/isaaclab/app/app_launcher.py
@@ -37,7 +37,10 @@ from isaaclab.app.loading_screen import report_activity
 from isaaclab.app.logging_utils import apply_python_logging_level, resolve_python_logging_level
 from isaaclab.app.settings_manager import get_settings_manager, initialize_carb_settings
 from isaaclab.utils._device import set_cuda_device
-from isaaclab.utils.renderers import ISAAC_RTX_SHOW_ALL_PARTITIONS_BY_DEFAULT_SETTING
+from isaaclab.utils.renderers import (
+    ISAAC_RTX_SHOW_ALL_PARTITIONS_BY_DEFAULT_SETTING,
+    set_isaac_rtx_per_env_scene_partition_default,
+)
 
 # import logger
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


source/isaaclab/isaaclab/app/app_launcher.py
```diff
@@ -1288,6 +1312,7 @@ def _resolve_kit_args(self, launcher_args: dict):
             setting = argument.partition("=")[0]
             if not any(arg.partition("=")[0] == setting for arg in sys.argv + self._kit_args):
                 self._kit_args.append(argument)
+            self._disable_per_env_scene_partitioning_default()
```
nvsekkin, line 1315: flipping this here would revert the default behavior we flipped. once the public package includes the fix we won't need this. maybe we can mark this with a temporary workaround or we can condition on isaac sim version

PR #7280, opened 2026-08-21
Preserve Blender conversion paths and failures

sylvesterkaczmarek:
# Description

Fixes Blender subprocess invocation in `scripts/tools/process_meshes_to_obj.py`.

The helper previously constructed the command as one formatted string and then called `.split(" ")`. Any valid Blender, input, output, or script path containing whitespace was therefore split into multiple command-line arguments.

The subprocess was also launched without `check=True`, so a failed Blender conversion could be silently treated as success by the wrapper.

The helper now builds the subprocess argument list directly, preserves each path as one argument, requests `check=True`, and raises a clear error when Blender cannot be found on `PATH`.

## Type of change

- Bug fix

## Validation

- Added a unit regression test with Blender, input, and output paths containing spaces.
- The test verifies each path remains one subprocess argument and `check=True` is used.
- Added coverage for the missing-Blender error path.
- No Blender installation is required by the tests.
- No package changelog fragment is required because this changes a repository script, not a `source/<pkg>/` package.

## Checklist

- [x] No new dependencies

### DIFF
--- a/scripts/tools/process_meshes_to_obj.py
+++ b/scripts/tools/process_meshes_to_obj.py
@@ -42,15 +42,16 @@ def run_blender_convert2obj(in_file: str, out_file: str):
         in_file: Input mesh file.
         out_file: Output obj file.
     """
+    if BLENDER_EXE_PATH is None:
+        raise FileNotFoundError("Unable to find the Blender executable on PATH.")
+
     # resolve for python file
     tools_dirname = os.path.dirname(os.path.abspath(__file__))
     script_file = os.path.join(tools_dirname, "blender_obj.py")
-    # complete command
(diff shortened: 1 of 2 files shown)


AntoineRichard (commented):

**Assessment:** Argument-list invocation preserves whitespace and check=True correctly propagates conversion failures.

**Findings:** Managed isaaclab tests are touched without a required changelog fragment.

**Requested changes:** Add source/isaaclab/changelog.d/<slug>.skip and run the changelog gate.

**Test scope:** Coverage is focused and sufficient.

**Documentation / AI skills:** No additional documentation update is required: no public CLI syntax or documented workflow changed. No AI-skill update is required.

PR #7281, opened 2026-08-21
Fix Blender OBJ output path handling

sylvesterkaczmarek:
# Description

Fixes two path-handling bugs in `scripts/tools/blender_obj.py`.

When the output is a basename such as `robot.obj`, `os.path.dirname()` returns an empty string. The converter previously tried to create that empty path with `os.makedirs("")`, so writing directly to the current working directory failed.

The optional USD export also used `out_file.replace("obj", "usd")`, which replaces every matching substring in the full path. A valid path such as `objects/robot.obj` could therefore be changed to `usdects/robot.usd`.

The converter now creates a parent directory only when one is present and derives the USD path by replacing only the file extension.

## Type of change

- Bug fix

## Validation

- Added a unit regression test for an OBJ basename in the current working directory.
- Added a regression test verifying `objects/robot.obj` maps to `objects/robot.usd` without changing directory names.
- Blender is mocked, so the tests do not require a Blender installation.
- Added the required `isaaclab` `.skip` changelog marker because managed package tests are touched.

## Checklist

- [x] No new dependencies

### DIFF
--- a/scripts/tools/blender_obj.py
+++ b/scripts/tools/blender_obj.py
@@ -67,8 +67,9 @@ def convert_to_obj(in_file: str, out_file: str, save_usd: bool = False):
     if not out_file.endswith(".obj"):
         out_file += ".obj"
     # create directory if it doesn't exist for destination file
-    if not os.path.exists(os.path.dirname(out_file)):
-        os.makedirs(os.path.dirname(out_file), exist_ok=True)
+    out_dir = os.path.dirname(out_file)
+    if out_dir and not os.path.exists(out_dir):
+        os.makedirs(out_dir, exist_ok=True)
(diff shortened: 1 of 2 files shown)


AntoineRichard (commented):

**Assessment:** Both basename-directory handling and final-extension replacement are correct.

**Findings:** Managed isaaclab tests are touched without a required changelog fragment.

**Requested changes:** Add source/isaaclab/changelog.d/<slug>.skip and rerun the changelog gate.

**Test scope:** Two mocked tests cover two distinct regressions and pass.

**Documentation / AI skills:** No additional documentation update is required: no public API or documented behavior surface changed. No AI-skill update is required.

PR #7287, opened 2026-08-21
Cache pip and uv downloads in GitHub Actions

mataylor-nvidia:
# Description

Python packages were re-downloaded from PyPI on every CI run. Only `license-check.yaml` enabled a package cache; every other workflow that installs Python dependencies did so uncached.

**`astral-sh/setup-uv` — `enable-cache: true`**

| Location | Covers |
|---|---|
| `docs.yaml` (both jobs) | `uv sync --extra test` |
| `kitless-docker.yml` | `uv run --with pytest --with pyyaml` |
| `.github/actions/install-ci-run` | `uv venv --seed` + managed 3.12 |

The composite action sets this explicitly rather than relying on the default `auto`, which enables the cache only on GitHub-hosted runners — that action also runs on the self-hosted arm64 runner via `arm-ci.yml`.

**`actions/setup-python` — `cache: pip`**

Added to `changelog-check.yml`, `skills-check.yml`, `tools-tests.yml`, `wheel.yml`. Each pins `cache-dependency-path: pyproject.toml`; the action's `**/requirements.txt` default would otherwise key off `tools/template/requirements.txt`, which is unrelated to what these jobs install.

**Deliberately left uncached**

- `resolve-ov-pins` and `nightly-changelog` install nothing with pip (stdlib `tomllib` / `tools/changelog/cli.py` only), so a cache round-trip would be pure overhead.
- `test-multi-gpu.yaml` runs `./isaaclab.sh --install` on a self-hosted runner. Its pip cache would be multi-GB (torch, Isaac Sim wheels) and would compete with the Warp kernel cache for the 10 GB per-repository quota — the same quota `build.yaml` notes was exhausted once before.

Note that the Docker-side `--mount=type=cache` on pip in `docker/Dockerfile.base` does not persist across runs: `docker-build` creates a per-job `docker-container` builder and removes it in a trap, and BuildKit does not export `type=cache` mount contents through `--cache-to`. That is unchanged here; the deps-hash image short-circuit already covers that path.

## Type of change

## Checklist

<sub>CI-only change: no source package touched, so no changelog fragment is required (`tools/changelog/cli.py check` passes clean).</sub>

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -150,6 +150,10 @@ jobs:
         uses: astral-sh/setup-uv@v6
         with:
           enable-cache: true
+          # Neither this job nor kitless-docker pins python-version, so both land
+          # on the same `-unknown-` cache key and only one can save. This job
+          # caches the wheel's full extras resolution; keep it separate.
+          cache-suffix: wheel-extras
 
       # Compose Docker-image-style metadata for the artifact. The artifact
       # name is what QA sees in `gh run download`, so we make it scannable:
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -136,10 +136,20 @@ jobs:
           python-version: "3.12"
           architecture: x64
           cache: pip
-          # Pin the cache key to the dependency source of truth. Without this the
-          # action keys off its `**/requirements.txt` default, which here matches
-          # only the unrelated tools/template/requirements.txt.
-          cache-dependency-path: pyproject.toml
+          # Keyed on this workflow file, not pyproject.toml: the pip cache key is
+          # otherwise identical across every 3.12 ubuntu job here, so they collide
+          # and all but the first fail to save. Covers the `build` and `wheel`
+          # packages that tools/wheel_builder/build.sh installs with pip.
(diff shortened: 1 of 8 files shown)


.github/workflows/wheel.yml
```diff
@@ -135,6 +135,25 @@ jobs:
         with:
           python-version: "3.12"
           architecture: x64
+          cache: pip
+          # Keyed on this workflow file, not pyproject.toml: the pip cache key is
+          # otherwise identical across every 3.12 ubuntu job here, so they collide
+          # and all but the first fail to save. Covers the `build` and `wheel`
+          # packages that tools/wheel_builder/build.sh installs with pip.
+          cache-dependency-path: .github/workflows/wheel.yml
```
nvsekkin, line 143: you mentioned the build package here but did not include it in the dependency paths, should we also include `tools/wheel_builder/build.sh` here?

nvsekkin: I guess we don't really edit this file directly so this may not be too important to include

nvsekkin (approved):

StafaH (approved):

kellyguo11 (approved):

PR #7292, opened 2026-08-22
[Cloner] Scope Newton global imports with clone plans

ooctipus:
# Description

Newton replication currently calls `ModelBuilder.add_usd()` from the stage root and relies on ignore paths for replicated environments. This change moves global ownership to the scene composition root and carries it in `ClonePlan.global_paths`.

- `InteractiveScene` returns env-scoped clone configs and an ordered `tuple[str, ...]` of shared prim roots.
- `ClonePlan` carries that tuple without representing globals as clone rows.
- Hand-built scenes pass non-empty globals explicitly; the ordinary no-global case defaults to `()`.
- The replication pipeline passes the plan declaration directly to backend contexts.
- Newton imports only the physics scene and each declared global root with `root_path=...`.
- Terrain-heightfield discovery scans those same roots.

There is no stage-root guessing, global-path registry, or full-stage fallback in clone-plan replication.

## Data flow

The scene owns classification:

```python
clone_cfgs, global_paths = self._collect_asset_cfgs()
with ReplicateSession(clone_cfgs, ..., global_paths=global_paths, stage=stage):
    self._add_entities_from_cfg()
```

The plan carries the declaration:

```python
ctx = BackendContext(stage, global_paths=plan.global_paths)
```

Newton performs narrow imports. For a plan declaring ground and light:

```python
builder.add_usd(stage, root_path="/physicsScene")
builder.add_usd(stage, root_path="/World/Ground")
builder.add_usd(stage, root_path="/World/Light")
```

The physics-scene result remains the returned stage metadata; global-import result dictionaries are not merged. Globals stay outside `sources`, `destinations`, `clone_mask`, and `cfg_rows`, so they create no clone work or per-environment solver/sensor structures.

PhysX collision filtering remains a separate post-clone concern based on `collision_group == -1`; global import ownership does not change that policy.

## Direct scenes

Only scenes with shared roots need an explicit argument:

```python
global_paths = ("/World/ground",)
plan = clone_plan_from_env_0(src, dest, num_envs, device, positions, global_paths)
```

Scenes without shared roots keep the compact default:

```python
plan = clone_plan_from_env_0(src, dest, num_envs, device, positions)
```

## Startup benchmark

RTX 5090, 4,096 environments, seed 42, Newton MJWarp, three post-warm-up runs per commit in alternating order. The baseline is `bc8b7bdf005`, immediately before scoped global import. Values are medians; positive deltas mean the PR is faster.

| Workload | Metric | Baseline | PR | PR delta |
|---|---|---:|---:|---:|
| `Isaac-Velocity-Rough-AnymalD`, kitless | Scene creation | 4.932 s | 4.934 s | -0.05% |
|  | Environment creation | 13.511 s | 13.569 s | -0.43% |
|  | Total startup | 15.420 s | 15.520 s | -0.65% |
| `Isaac-Lift-KukaAllegro`, kitless | Scene creation | 14.248 s | 14.202 s | +0.32% |
|  | Environment creation | 26.222 s | 26.184 s | +0.15% |
|  | Total startup | 28.875 s | 28.830 s | +0.16% |
| `Isaac-Lift-KukaAllegro-Camera`, OVRTX RGB64 | Scene creation | 8.312 s | 8.232 s | +0.96% |
|  | Environment creation | 35.743 s | 35.782 s | -0.11% |
|  | Total startup | 39.936 s | 39.976 s | -0.10% |

All differences are below 1%, so these workloads show no measurable gain. That is expected for the two kitless tasks because their USD stages do not contain materialized replicated environment trees. The warmed OVRTX workload also does not reproduce the earlier single-run estimate.

A full Kit-stage comparison could not be run because Isaac Sim/Kit is not installed on the benchmark machine. That is the case where avoiding traversal of a materialized replicated USD tree should matter, and it remains unmeasured here.

## Testing

- Clone-plan, replication-session, and Newton global-world coverage: `90 passed` on the final API.
- Newton contact-sensor module after moving its ground into the scene declaration: `85 passed, 8 xpassed`.
- Earlier full focused suite: `151 passed`.
- Finalized-model coverage verifies the declared ground collider has Newton `shape_world == -1`; the declared light correctly creates no Newton physics entity.
- 16-env Cartpole, Anymal-D, and Newton + OVRTX camera startup/reset/first-step checks passed.
- Three 4,096-environment benchmark runs per workload and commit passed.
- Repository formatting and pre-commit checks passed.

The contact suite exposed the ownership boundary correctly: its ground had previously been spawned by `build_simulation_context()` outside the scene config and therefore was absent from the clone plan. The fix declares the ground in `ContactSensorTestSceneCfg`; the single intentionally groundless test sets `terrain = None`. No discovery fallback was added.

## Type of change

- Performance/architecture improvement
- Documentation update

## Checklist

- [x] I have run the pre-commit checks.
- [x] I have made corresponding changes to the documentation.
- [x] I have added tests that prove the feature works.
- [x] I have added changelog fragments for changed packages.


source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -384,35 +389,38 @@ def clone_plan_from_env_0(
...
-    }
+    cfg_rows = {id(cfg): (0,) for cfg in REPLICATION_QUEUE if match(cfg.prim_path, destination) is not None}
     return ClonePlan(
         sources=(source,),
         destinations=(destination,),
         clone_mask=torch.ones((1, num_clones), dtype=torch.bool, device=device),
         env_ids=torch.arange(num_clones, dtype=torch.long, device=device),
         positions=positions,
         cfg_rows=cfg_rows,
+        global_paths=tuple(dict.fromkeys(global_paths)),
```
StafaH: Can we use a different input type then Iterable[str], so that we can avoid the conversions?

source/isaaclab/isaaclab/cloner/usd.py
```diff
@@ -30,7 +30,7 @@ class UsdReplicateContext:
 
     replicate_priority = 100
 
-    def __init__(self, stage: Usd.Stage):
+    def __init__(self, stage: Usd.Stage, global_paths: Sequence[str] = ()):
```
StafaH: type here is Sequence now

source/isaaclab/isaaclab/scene/interactive_scene.py
```diff
@@ -278,7 +278,7 @@ def _collect_asset_cfgs(self) -> list[Any]:
             )
         else:
             self._clone_valid_set = None
-        return cfgs
+        return cfgs, tuple(dict.fromkeys(global_paths))
```
StafaH: the main thing to try to improve is to avoide the tuple(dict.fromkeys( if we can with something cleaner.

StafaH (commented):
Nice! I have 2 comments:

The type for global paths looks inconsistent, and also the conversion looks odd. Can we enfore a type that doesn't require conversion? If its tuple, then we know everywhere that user needs to pass tuple.

Second question is on global world. Let's verify (either in a test or manually), that the generated newton model has the global prim inside newton's global world (found at index -1).

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Updated all clone/import `global_paths` annotations to `tuple[str, ...]` and now pass the tuple through unchanged. The normalization conversions in both clone-plan constructors are removed.

source/isaaclab/isaaclab/cloner/usd.py
ooctipus: Updated this and the other replication contexts to use `tuple[str, ...]` consistently.

source/isaaclab/isaaclab/scene/interactive_scene.py
ooctipus: Removed `tuple(dict.fromkeys(...))`. `InteractiveScene` now constructs the ordered unique tuple directly and returns it unchanged.

ooctipus:
Addressed the global-world verification in `test_newton_builder_world_hook.py`: the test now builds and finalizes a real Newton model from declared `("/World/Ground", "/World/Light")` roots, verifies all scoped `add_usd` calls, and asserts the ground shape has `shape_world == -1` with two local worlds. The light root is imported too; as expected, a USD light creates no Newton physics body or shape.

ooctipus:
Re-ran the startup comparison with three alternating post-warm-up runs per commit at 4,096 environments (seed 42). The medians show no measurable end-to-end improvement: Anymal-D Rough kitless -0.65%, KukaAllegro kitless +0.16%, and KukaAllegro Camera/OVRTX -0.10% total startup. Scene-creation deltas were -0.05%, +0.32%, and +0.96%, respectively, all within run-to-run spread. I replaced the earlier 6–9% claim in the PR body because it was based on only one baseline and was sensitive to warm-up variance. This matches the expectation that kitless stages do not materialize the replicated USD env tree. A direct Kit/full-stage benchmark is still needed; Isaac Sim/Kit is not installed on this machine.

StafaH (approved):

PR #7298, opened 2026-08-22
Fix MJWarp USD friction loss import

### LINKED ISSUE #6829: [Bug Report] Newton USD import drops mjc:frictionloss.
Title: [bug] Newton USD import drops mjc:frictionloss.

Describe the bug

On the Newton backend, joint Coulomb friction authored on a USD as mjc:frictionloss never reaches the solver — Model.joint_friction / dof_frictionloss stay 0 even though mjc:damping works. Loading a mjcf directly loads everything properly.

Steps to reproduce

Newton core handles it correctly with the mjc resolver, which pinpoints the gap. Minimal (pure Newton):

# revolute joint authored with mjc:frictionloss=0.11 on a USD stage
b = newton.ModelBuilder()
solvers.SolverMuJoCo.register_custom_attributes(b)
b.add_usd(stage, schema_resolvers=[SchemaResolverNewton(), SchemaResolverPhysx()])           # Isaac Lab's list
print(b.finalize().joint_friction)   # -> [... 0.0]  (dropped)

b2 = newton.ModelBuilder(); solvers.SolverMuJoCo.register_custom_attributes(b2)
b2.add_usd(stage, schema_resolvers=[SchemaResolverNewton(), SchemaResolverPhysx(), SchemaResolverMjc()])
print(b2.finalize().joint_friction)  # -> [... 0.11]  (correct)
End-to-end: any Newton USD robot with mjc:frictionloss authored → dof_frictionloss == 0 in sim.

System Info

- Commit: f703bea7e
 - Isaac Sim Version: 6.0.0-rc.22 (release.33481) · Newton 1.0.0
 - OS: Ubuntu 22.04
- GPU: RTX 3090
- CUDA: 13.2 (Warp toolkit 12.9)
- GPU Driver: 595.84

Checklist

- [x] I have checked that there is no similar issue in the repo (required)

Acceptance Criteria

- [ ] A USD joint with mjc:frictionloss authored yields matching Model.joint_frice Newton backend

NeoZng:
# Description

Fixes #6829.

Preserve MuJoCo-authored joint friction loss when Isaac Lab imports USD stages for the Newton MJWarp backend.

Before this change, both Newton production import paths passed only `SchemaResolverNewton` and `SchemaResolverPhysx` to Newton:

* vectorized clone replication; and
* standalone stage import.

Consequently, a joint authored with `mjc:frictionloss=0.11` finalized with `Model.joint_friction=0.0`, even though Newton supports the attribute through `SchemaResolverMjc`.

This PR makes USD resolver selection an active-manager policy:

* `NewtonManager` defaults to Newton then PhysX resolvers;
* `NewtonMJWarpManager` appends the MuJoCo resolver;
* clone and standalone imports consume the same manager-owned resolver list; and
* resolver order remains Newton → PhysX → MuJoCo, so MJC values are fallbacks and do not override higher-priority authored values.

MuJoCo custom attributes continue to be registered through the existing `_builder_attribute_solvers = (SolverMuJoCo,)` mechanism. The follow-up removes the redundant MJWarp registration override instead of duplicating that base-class path.

## Scope and alternative considered

PR #7386 fixes the immediate `frictionloss` symptom by appending the complete `SchemaResolverMjc` to every Newton physics import. That resolver also interprets additional joint, shape, contact, and scene attributes. Applying it unconditionally would therefore expand MJC semantics to Featherstone, XPBD, VBD, Kamino, and MPM rather than changing only MJWarp.

This PR intentionally keeps the immediate fix solver-scoped: the manager that registers and consumes MuJoCo-specific attributes also owns the MuJoCo resolver. Regression coverage verifies that MJWarp imports `mjc:frictionloss` and `mjc:damping`, while Featherstone preserves its previous behavior.

A broader cross-backend solution should separately classify portable MJC core semantics from MuJoCo-specific extensions, ideally by splitting those resolver responsibilities in Newton upstream. That larger architecture change is outside the scope of this bug fix.

No dependency or public API is added.

## Type of change

* Bug fix (non-breaking change which fixes an issue)

## Release backport

* [x]  Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

Not applicable; this fixes a non-visual USD import path.

## Verification

Latest follow-up commit:

* focused production-path regression matrix: 5 passed;

  * MJWarp × clone/standalone preserves `frictionloss=0.11` and `damping=0.23`;
  * Featherstone × clone/standalone leaves both MJC-authored values unconsumed;
  * the existing explicit-global clone import test still passes;
* full pre-commit suite passed;
* changelog validation passed;
* Python bytecode compilation passed; and
* `git diff --check` passed.

Previous PR head verification in the matching isolated environment:

* Python 3.12.13;
* `isaaclab-newton==5.4.0`;
* `newton==1.5.0`;
* `warp-lang==1.16.0`;
* `mujoco-warp==3.11.0`; and
* full Newton manager abstraction suite: 157 passed on a CUDA host.

## Checklist

* [x] I have read and understood the [[contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
* [x] I have run the pre-commit checks
* [x] I have made corresponding changes to the documentation (not applicable: no public API or user workflow changed)
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective
* [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
* [x] I have added my name to `CONTRIBUTORS.md`

### DIFF
--- a/source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
+++ b/source/isaaclab_contrib/isaaclab_contrib/coupling/coupler.py
@@ -23,6 +23,7 @@ from isaaclab_newton.physics.mpm_manager import NewtonMPMManager
 from isaaclab_newton.physics.newton_manager import NewtonManager
 from isaaclab_newton.physics.vbd_manager import NewtonVBDManager
 from newton import CollisionPipeline, Model, ModelBuilder, ShapeFlags
+from newton.solvers import SolverBase
 from newton.solvers.experimental.coupled import SolverCoupled, SolverCoupledADMM, SolverCoupledProxy
 
 from isaaclab.physics import PhysicsManager
@@ -190,6 +191,14 @@ class NewtonCouplerManager(NewtonVBDManager):
         super()._register_builder_attributes(builder)
         for entry in PhysicsManager._cfg.solver_cfg.entries:
             entry.solver_cfg.class_type._register_builder_attributes(builder)
+
+    @classmethod
+    def _registers_builder_attributes_from_solver(cls, solver_cls: type[SolverBase]) -> bool:
+        """Return whether the manager or a configured nested entry registers ``solver_cls`` attributes."""
+        return super()._registers_builder_attributes_from_solver(solver_cls) or any(
+            entry.solver_cfg.class_type._registers_builder_attributes_from_solver(solver_cls)
+            for entry in PhysicsManager._cfg.solver_cfg.entries
+        )
 
     @classmethod
     def _prepare_builder_for_finalize(cls, builder: ModelBuilder) -> None:
--- a/source/isaaclab_contrib/test/coupling/test_coupler.py
+++ b/source/isaaclab_contrib/test/coupling/test_coupler.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 from dataclasses import dataclass, field
 from types import SimpleNamespace
 
+import isaaclab_newton.physics.newton_manager as newton_manager_module
 import numpy as np
 import pytest
 from isaaclab_newton.physics import (
@@ -24,6 +25,7 @@ from isaaclab_newton.physics import (
     KaminoPADMMSolverCfg,
     MJWarpSolverCfg,
     MPMSolverCfg,
+    NewtonCfg,
     NewtonCollisionPipelineCfg,
     NewtonVBDManager,
     VBDSolverCfg,
@@ -32,6 +34,8 @@ from isaaclab_newton.physics import (
 from isaaclab_newton.physics.newton_manager import NewtonManager
 from newton import ModelBuilder, ShapeFlags
 from newton.solvers.experimental.coupled import SolverCoupledADMM, SolverCoupledProxy
+
+from pxr import Sdf, Usd, UsdGeom, UsdPhysics
 
 from isaaclab_contrib.coupling import (
     CouplerAdmmCfg,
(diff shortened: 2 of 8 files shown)


AntoineRichard (commented):
> [!NOTE]
> **AI-generated review.** This review was produced by Claude (Claude Code) at @AntoineRichard's request, and reviewed by him before posting. Treat the findings as a starting point; the code references and behavioral claims below were verified against `newton==1.5.0` sources, but a human maintainer owns the approval decision.

Thanks for this @NeoZng — I compared this PR against #7386, which fixes the same issue (#6829) by appending `SchemaResolverMjc` unconditionally on every Newton import path. **I prefer this PR's approach** and would like to converge on it, with two changes requested below.

## Why this approach wins

I checked the claim that scoping matters, rather than taking it on faith:

- `newton/_src/usd/schemas.py` — `SchemaResolverMjc` is not joint-only. Beyond `mjc:frictionloss` and `mjc:armature`, it maps `PrimType.SCENE` (`mjc:option:timestep` → `time_steps_per_second`, `mjc:option:iterations`, `mjc:flag:gravity`), `PrimType.SHAPE` (`mjc:margin`, `mjc:gap`, `mjc:solref` → contact `ke`/`kd`, `mjc:shellinertia` → mass model), `PrimType.MATERIAL` (torsional/rolling friction, `mjc:solmix`, `mjc:priority`) and `PrimType.ACTUATOR`. Applying it unconditionally really would push MuJoCo contact and limit semantics into Featherstone, XPBD, VBD, Kamino and MPM for any MJCF-derived asset — so the scoping argument in your PR description holds up.
- `newton/_src/usd/schema_resolver.py::SchemaResolverManager.get_value_with_resolver` — precedence is first-authored-value-wins in resolver order, then the caller default, then the first non-`None` mapping default. Your Newton → PhysX → MuJoCo ordering therefore makes MJC a strict fallback, as documented. I also checked the mapping-default path for a possible leak (Mjc supplying defaults such as `mu_torsional=0.005` on assets with no `mjc:*` authored): all three resolvers declare the same key sets and Newton is first, so Mjc's defaults are unreachable. No issue.

Secondary wins over the alternative: the `_get_usd_import_schema_resolvers()` hook matches the existing `_get_usd_import_ignore_paths` / `_builder_attribute_solvers` extension pattern; `replicate.py` loses its private `newton._src.usd.schemas` import in favour of the public `newton.usd`; and the parametrized clone × standalone × MJWarp/Featherstone test runs on CPU and includes a negative control.

## Requested changes

**1. Gate on the registered solvers, not on the manager class.**

`NewtonCoupledMJWarpVBDManager` (`source/isaaclab_contrib/isaaclab_contrib/custom_coupling/coupled_mjwarp_vbd_manager.py:22`) extends `NewtonVBDManager`, not `NewtonMJWarpManager`, yet sets `_builder_attribute_solvers = (SolverMuJoCo,)`. It runs a MuJoCo solver, so it wants the MJC resolver, but the current class-hook override does not reach it — the bug stays unfixed there.

Deriving the resolver list from the attributes the manager already registers keeps the same MJWarp-only scoping, covers the coupled manager, and picks up any future MuJoCo-based manager automatically. Roughly, in `NewtonManager`:

```python
@classmethod
def _get_usd_import_schema_resolvers(cls) -> list[SchemaResolver]:
    resolvers: list[SchemaResolver] = [SchemaResolverNewton(), SchemaResolverPhysx()]
    if SolverMuJoCo in cls._builder_attribute_solvers:
        resolvers.append(SchemaResolverMjc())
    return resolvers
```

That would let the `NewtonMJWarpManager` override go away entirely. If you would rather keep the explicit override, an override on `NewtonCoupledMJWarpVBDManager` is the minimum, but the derived form is less likely to drift.

**2. Note the call sites the hook deliberately does not cover.**

`visualization_builder.py:98` and `isaaclab/assets/articulation/ordering_resolvers.py:481` still hardcode `[SchemaResolverNewton(), SchemaResolverPhysx()]`. That is correct — resolvers only select which attributes get parsed, so they cannot affect rendering or joint ordering — but after this PR there are two conventions in the tree. A one-line comment at each site (or at the hook) saying those paths intentionally stay on the fixed pair would keep the next reader from "fixing" the inconsistency.

## Note on #7386

For the record, that PR's "no-op-safe when a solver has not registered MuJoCo custom attributes" claim is only half right, and the reason is worth knowing here too: `SchemaResolverMjc.validate_custom_attributes` raises when no `mujoco`-namespace custom attribute is registered, and `newton/_src/utils/import_usd.py` calls it for every resolver unconditionally — but a bare `newton.ModelBuilder()` already carries 17 `mujoco`-namespace attributes, so the guard never fires. It is safe from raising, and it does change behavior for non-MJWarp solvers. Your gating is what makes that a non-question.

Once the two points above are addressed I am happy to see this land, and #7386 closed in favour of it.

NeoZng:
@AntoineRichard  Thanks for the detailed review and the comparison with #7386. I have pushed the requested follow-up on top of the refreshed develop base as commit 924e407159.

The implementation now:

- derives SchemaResolverMjc selection from registered SolverMuJoCo builder attributes rather than manager inheritance;

- reuses that policy for both standalone stage import and clone replication;

- covers the coupled MJWarp/VBD manager automatically and runtime-configured nested solvers through the generic coupler's capability query;

- documents why visualization and articulation-ordering paths intentionally retain the fixed Newton/PhysX resolver pair; and

- adds distinct-valued Newton/PhysX/MJC precedence coverage plus solver positive/negative controls.


This keeps the fix narrower than #7386: MJC schema interpretation is enabled only for managers that actually register SolverMuJoCo, so it does not leak MuJoCo semantics into plain Featherstone, XPBD, VBD, Kamino, or MPM managers.

Local validation:

- full repository pre-commit passed;

- Newton manager: 181 passed;

- Newton cloner: 43 passed;

- custom coupled manager: 12 passed;

- generic coupler: 57 passed;

- MJCF converter: 12 passed;

ooctipus (approved):

kellyguo11:
run-ci

PR #7301, opened 2026-08-22
Centralize shared physics preset ownership

ooctipus:
# Description

Centralize backend physics presets at task-family composition roots so robot-specific files own only genuinely robot-specific scalar configuration. Standardize Newton MJWarp velocity environments on two physics substeps at the shared family boundary.

- Move the Kuka Allegro OvPhysX capacity settings into the shared lift `PhysicsCfg` and remove `KukaAllegroPhysicsCfg`.
- Move the SO-101 articulation contact ordering setting into the shared stack `PhysicsCfg` and remove `SO101StackPhysicsCfg`.
- Make `RoughPhysicsCfg` own the shared PhysX, MJWarp, and Kamino alternatives once.
- Set the shared velocity MJWarp configuration to `num_substeps=2` and remove every robot-specific substep assignment.
- Remove the nine flat-locomotion `PhysicsCfg(RoughPhysicsCfg)` wrappers. Their environment configs now mutate only robot-specific MJWarp values on the inherited `RoughPhysicsCfg` instance.
- Preserve shared Newton contact, collision-pipeline, and shape settings instead of replacing the whole shared alternative in robot-specific files.
- Add an `isaaclab_tasks` major changelog fragment for the shared two-substep behavior change.

No additional dependencies are required.

## Type of change

- Refactor: centralize physics-preset ownership.
- Breaking behavior change: Newton MJWarp velocity environments now use two physics substeps.

## Validation

- Existing focused lift, stack, velocity, and camera preset tests: 15 passed before the substep follow-up.
- Focused velocity Newton config test after the substep follow-up: 1 passed.
- Registered task preset-selection check: passed across the task registry.
- Runtime configuration compatibility checks: 31 passed.
- Structural audit confirms there is no `PhysicsCfg(SomeOtherPhysicsCfg)` inheritance under `isaaclab_tasks`.
- Structural audit confirms no core/contrib robot-specific velocity config assigns `.num_substeps`.
- Direct config audit confirms all ten affected concrete core/contrib velocity configs resolve MJWarp with two substeps.
- `uv run --frozen isaaclab -f`: Ruff, formatting, and applicable source checks passed; the repository-wide changelog gate still reports pre-existing missing or modified fragments in untouched packages.
- Additional preset CLI run: 26 passed; one could not complete because Isaac Sim requested interactive NVIDIA EULA acceptance, which was not accepted automatically.

## Screenshots

Not applicable.

## Checklist

- [x] Physics preset ownership is centralized without a compatibility layer.
- [x] The shared two-substep behavior change is recorded explicitly.
- [x] Existing tests cover preset construction and selection.
- [x] No test files were added or changed.
- [x] An `isaaclab_tasks` major changelog fragment is included.
- [x] No additional documentation changes are needed.


kellyguo11 (approved):

source/isaaclab_tasks/isaaclab_tasks/core/velocity/velocity_env_cfg.py
```diff
@@ -64,6 +70,7 @@ class RoughPhysicsCfg(PresetCfg):
         debug_mode=False,
```
StafaH, line 70: Some comments for newton_mjwarp defaults I noticed: 

njmax is really high. 
num_substeps=1 should maybe be 2? 
friction type pyramidal might be alot better than cone. 

Also we can probably set defaults that don't need to be overrides, it makes sense that each robot has different value but I would expect the main difference is only njmax

source/isaaclab_tasks/isaaclab_tasks/core/velocity/config/g1/flat_env_cfg.py
```diff
@@ -3,47 +3,23 @@
...
 @configclass
 class G1FlatEnvCfg(G1RoughEnvCfg):
-    sim: SimulationCfg = SimulationCfg(physics=PhysicsCfg())
-
     def __post_init__(self):
         super().__post_init__()
 
+        # physics
+        newton_mjwarp = self.sim.physics.newton_mjwarp
+        newton_mjwarp.solver_cfg.njmax = 95
```
StafaH, line 19: I would have expected njmax 192, and nconmax 48. nconmax 10 is very low.

StafaH (approved):

source/isaaclab_tasks/isaaclab_tasks/core/velocity/velocity_env_cfg.py
ooctipus: Updated in 55815deb530. All Newton MJWarp velocity environments now use two physics substeps. `RoughPhysicsCfg` owns `num_substeps=2` once, and every core/contrib robot-specific `.num_substeps` assignment has been removed. Spot has an independent velocity physics composition, so its MJWarp value is also set to 2. The shared config already uses `cone="pyramidal"`; robot-specific constraint capacities remain separate for now.

source/isaaclab_tasks/isaaclab_tasks/core/velocity/config/g1/flat_env_cfg.py
ooctipus: `95/10` are the existing Isaac Lab G1-flat values; this refactor copies them onto the inherited shared preset rather than introducing new tuning. I also checked current MJLab main: its G1-flat config uses [`njmax=300` and `nconmax=None`](https://github.com/mujocolab/mjlab/blob/0fb8a681136be94ffc636a3dd423cabb97d91f10/src/mjlab/tasks/velocity/config/g1/env_cfgs.py#L188-L195), while G1 rough sets `nconmax=70`. Since `192/48` is neither the prior Isaac Lab behavior nor the current MJLab value, I am preserving parity here. I agree `nconmax=10` is worth a separate capacity/stability benchmark and tuning PR.

PR #7303, opened 2026-08-22
Remove post-Hydra preset resolution

ooctipus:
## Summary

- make resolve_task_config and parse_env_cfg the task configuration composition boundary, with explicit programmatic overrides
- remove late preset fallback handling from environments, simulation, launch scanning, camera validation, benchmarks, and RL summaries
- require runtime consumers to receive concrete physics, renderer, and camera configurations
- route scripts, tools, integrations, and relevant tests through registered task composition

This is a separate follow-up to #7301; it does not mix the earlier preset ownership cleanup into this PR.

## Validation

- 129 Hydra and Shadow Hand camera tests passed
- 31 benchmark capture and RL entrypoint tests passed
- 23 experimental frontend tests passed, 1 skipped
- 5 custom-coupling tests passed
- representative composition audit confirmed six task trees contain no remaining PresetCfg nodes
- Python compile checks passed for all changed runtime and script paths
- ruff and ruff-format passed

The full format command passes every hook except the changelog comparison, which uses the stale local origin/develop ref and flags six pre-existing upstream fragments that this branch does not modify. This PR includes all required package fragments.

### DIFF
--- a/source/isaaclab/isaaclab/benchmark/capture.py
+++ b/source/isaaclab/isaaclab/benchmark/capture.py
@@ -233,7 +233,7 @@ def _backends_from_env_cfg(env_cfg: object) -> tuple[str | None, str | None]:
     physics_cfg = getattr(getattr(env_cfg, "sim", None), "physics", None)
     physics_type = type(physics_cfg)
     physics_descriptor = (
-        "physx"
+        "newton_mjwarp"
         if physics_cfg is None
         else f"{physics_type.__module__}.{physics_type.__name__} {getattr(physics_cfg, 'class_type', '')}".lower()
     )
--- a/source/isaaclab/isaaclab/benchmark/capture.py
+++ b/source/isaaclab/isaaclab/benchmark/capture.py
@@ -12,24 +12,21 @@ raw measurements and metadata to the typed schema dataclasses
 :class:`~isaaclab.benchmark.schema.Resources`.
 
 Import-time dependencies stay light: no torch, isaacsim, or RL-library
-imports. Preset metadata is imported lazily when run_config_from_presets()
-needs it. The benchmark object is accepted at call time; its recorder classes
+imports. The benchmark object is accepted at call time; its recorder classes
 are never imported here.
 """
 
 from __future__ import annotations
 
 import socket
-from collections.abc import Sequence
 from datetime import datetime, timezone
-from typing import Any, get_args
+from typing import Any
 
 from isaaclab.benchmark.schema import (
     GpuDeviceInfo,
     GpuResources,
     Hardware,
     MeanStd,
-    PhysicsBackend,
     Resources,
     RunConfig,
     Versions,
@@ -97,7 +94,7 @@ def synth_run_id(
     Args:
         framework: RL framework name, or ``None`` for non-learning runs
             (substituted with ``"runtime"``).
-        physics_backend: Physics backend preset string.
+        physics_backend: Physics backend name.
         task: Gym task id.
         seed: Environment/agent seed.
         stamp: Timestamp string (e.g. ``"20260612-150000"``).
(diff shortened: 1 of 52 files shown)


source/isaaclab/isaaclab/sim/simulation_context.py
```diff
@@ -46,15 +46,14 @@
 _VISUALIZER_ALIASES = {"newton": "newton_gl"}
 
 
-def _resolve_physics_cfg(physics_cfg: Any, use_isaac_sim: bool) -> PhysicsCfg:
+def _resolve_physics_cfg(physics_cfg: PhysicsCfg | None, use_isaac_sim: bool) -> PhysicsCfg:
     """Resolve a simulation physics config to a concrete backend."""
     if physics_cfg is None:
         from isaaclab_physx.physics import PhysxCfg
```
StafaH, line 52: If no physics confic, maybe we can resolve newton now?

source/isaaclab_tasks/isaaclab_tasks/core/cartpole/cartpole_direct_camera_env.py
```diff
@@ -28,16 +27,12 @@ class CartpoleCameraEnv(CartpoleEnv):
...
-        # observation space. Use the default camera preset only for that
-        # derivation; leave full config resolution to DirectRLEnv.
-        camera_cfg = cfg.tiled_camera.default if isinstance(cfg.tiled_camera, PresetCfg) else cfg.tiled_camera
         cfg.frame_stack = max(1, cfg.frame_stack)
         if isinstance(cfg.observation_space, list):
             cfg.observation_space = [
                 int(cfg.observation_space[0]) * cfg.frame_stack,
-                int(camera_cfg.height),
-                int(camera_cfg.width),
+                int(cfg.tiled_camera.height),
```
StafaH, line 34: Originally this code was modified to be this way since overriding the camera width and height was not correctly updating observation space. One issue was that the preset had not resolved yet. Please double check that from the command line it correctly picks up that observation space is updated and there is no attribute error

StafaH (approved):

source/isaaclab/isaaclab/sim/simulation_context.py
ooctipus: Corrected in 8a501ac07fe after confirming the intended default. When SimulationCfg.physics is None, SimulationContext now resolves NewtonCfg() with MJWarp. Passing an explicit PhysxCfg retains Isaac Sim PhysX. The configuration docs, benchmark metadata, changelog, and focused regression coverage now match that contract.

source/isaaclab_tasks/isaaclab_tasks/core/cartpole/cartpole_direct_camera_env.py
ooctipus: Verified through the command-line composition path and added a focused regression test in ac557d7a552. The test passes env.tiled_camera.height=45, env.tiled_camera.width=80, and env.frame_stack=1 through sys.argv/Hydra, then exercises CartpoleCameraEnv initialization far enough to assert observation_space == [3, 45, 80]. The camera config is concrete at that point and there is no attribute error.

source/isaaclab/isaaclab/utils/__init__.pyi
```diff
@@ -107,4 +106,4 @@ from .string import (
 )
 from .types import ArticulationActions
 from .version import has_kit, get_isaac_sim_version, compare_versions
-from .configclass import checked_apply, configclass, resolve_cfg_presets
+from .configclass import checked_apply, configclass
```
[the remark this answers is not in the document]
ooctipus: This removal is intentional and documented as breaking in a major fragment. A forwarding wrapper cannot direct callers to task composition because resolve_cfg_presets accepted an arbitrary already-constructed object, while the replacement boundary requires a registered task ID plus Hydra composition. Retaining that wrapper would preserve the exact late-resolution path this change removes. The migration is resolve_task_config or parse_env_cfg before environment construction.

source/isaaclab/isaaclab/benchmark/capture.py
```diff
@@ -308,91 +275,26 @@ def _backend_defaults_from_env_cfg(env_cfg: object) -> tuple[str | None, str | N
...
-
-    if env_cfg is not None:
-        active_physics, active_rendering = _backend_defaults_from_env_cfg(env_cfg)
-        physics = active_physics or physics
-        rendering = active_rendering or rendering
+    Raises:
+        ValueError: If the config does not contain a supported concrete physics backend.
+    """
+    physics, rendering = _backends_from_env_cfg(env_cfg)
+    if physics is None:
```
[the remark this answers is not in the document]
ooctipus: Updated in 8a501ac07fe: benchmark capture now maps sim.physics=None to newton_mjwarp, matching the corrected SimulationCfg and SimulationContext default. It still raises for an unrecognized concrete physics object, and the focused capture suite passes.

kellyguo11 (approved):

PR #7324, opened 2026-08-24
Add IsaacLab environment capture bundle for reproducing bug reports

mataylor-nvidia:
# Description

Most setup failures are a difference between two machines rather than a defect in the code, and the difference is rarely in the lockfile. This adds `tools/capture_env.py`, which records the parts a lockfile does not — the GPU and driver, the installed distributions as they exist on disk, the environment variables Isaac Lab reads, and the symlinks and `.pth` files that decide which code actually gets imported — renders `REPRODUCE.md` with the steps to rebuild that environment, and diffs two captures against each other.

```bash
python3 tools/capture_env.py capture --command "<the command that failed>"
python3 tools/capture_env.py diff customer-bundle.zip
```

The reproduction steps are derived from the capture rather than written in advance, because a generic recipe is wrong in three ways that matter:

- **Extras.** `uv.lock` is parsed into a dependency graph and the extras whose requirements are all installed are selected. A bare `uv sync` installs the default dependency set and *removes* everything else — on this checkout that is 143 packages including `isaacsim`, `ovrtx`, `rl-games`, and `skrl`, so it dismantles the environment rather than rebuilding it.
- **Packages outside the lockfile.** Anything installed but outside the sync closure is listed for explicit reinstallation. `INSTALLER` cannot reveal these, because `uv pip install` records the same `uv` that `uv sync` does; only a comparison against the lockfile finds them.
- **Isaac Sim.** It gets a step matched to how it was obtained: the `isaacsim` wheel, a downloaded package at the recorded version, or a local build at the revision parsed out of its `VERSION` string.

The script imports nothing outside the standard library and never imports `isaaclab`, so it runs on an installation too broken to import anything, and reads the environment from disk rather than through the interpreter that runs it. Each bundle carries a copy of the script, so `diff` runs on a checkout too old to contain it or a machine with no checkout at all.

The process environment is captured by an exact, closed allowlist of the names this project reads. Everything else is counted and never named, so a bundle cannot carry credentials or internal hostnames off the machine that produced it. Uncommitted source changes are excluded for the same reason unless `--include_diff` is passed.

## Verification against uv

The derived sync command and the residue list were checked against `uv`'s own answer on a working checkout:

| | |
|---|---|
| Derived command | `uv sync --locked --extra all --extra test` |
| `uv sync --locked --dry-run` (what the generic recipe would do) | uninstalls 143 packages |
| `uv sync --locked --extra all --extra test --dry-run` | 0 installs, 7 removals |
| Packages the tool reports as outside the lockfile | the same 7, exactly |

# Type of change

# Checklist

- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file — n/a, `tools/` is not a source package and the changelog fragment check passes


docs/source/refs/troubleshooting.rst
```diff
@@ -16,6 +16,87 @@ Tricks and Troubleshooting
...
+-----------------------------------------
+
+Most setup failures are a difference between two machines rather than a defect in the code, and the
+difference is rarely in the lockfile. ``tools/capture_env.py`` records the parts a lockfile does not:
+the GPU and driver, the installed packages as they exist on disk, the environment variables Isaac Lab
+reads, and the symlinks and ``.pth`` files that decide which code actually gets imported.
+
+.. code:: bash
+
+    python3 tools/capture_env.py capture --command "<the command that failed>"
```
mataylor-nvidia, line 32: The reason to have a zip and not docker image is to minimize on the size of the artifact produced.

image would be 500+ mb

zip should only be few mb

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

mataylor-nvidia:
@greptile review

docs/source/refs/troubleshooting.rst
```diff
@@ -19,6 +19,106 @@ Tricks and Troubleshooting
...
```
mataylor-nvidia, line 32: Note on `python3` here rather than `uv run`, since the rest of the tooling is uv-first: this is deliberate.

The script exists to diagnose environments that are broken, and the uv-managed venv is often the thing that is broken. `uv run` resolves and syncs the project environment before executing, so on exactly the machines this is meant to capture it would either refuse to start or mutate the environment before it could be recorded. The script uses only the standard library, never imports Isaac Lab, and reads the virtual environment from disk rather than from the interpreter running it, so any `python3` produces the same capture.

It also has to run where there is no checkout and no uv at all: each bundle carries a copy of the script so `diff` works on a machine that has nothing but Python.

mataylor-nvidia:
run-ci

PR #7328, opened 2026-08-25
Fix the cable being culled by its scene partition in the Franka cable env

mataylor-nvidia:
# Description

Kit RTX computes the bounding box of a `UsdGeom.BasisCurves` prim once and never refreshes it while the curve deforms (**OMPE-105749**). A per-environment scene partition is sized from the union of the bounding boxes of the prims it contains, so it only covers the cable's *spawn* extent — and the cable is culled from the camera images once it deforms outside it.

This applies the workaround recommended by the rendering team ([slack thread](https://nvidia.slack.com/archives/C0ADR7ELXAB/p1786729788310189?thread_ts=1786657405.744359&cid=C0ADR7ELXAB), tracked by [OMPE-105975](https://jirasw.nvidia.com/browse/OMPE-105975)): place tiny geometry at opposite corners of a virtual box enveloping each environment so the partition's bounding-box union covers the whole workspace.

## Measured behaviour

Sweeping the cable away from its spawn pose in 0.1 m steps, on Kit 110.1.2 / Isaac Sim 6.1, RTX PRO 3000 Blackwell, counting cable pixels in `instance_id_segmentation_fast`:

| offset in y | partitioning off (control) | markers **absent** | markers **present** |
| --- | --- | --- | --- |
| 0.50 m | 414 px | 414 px | 414 px |
| **0.60 m** | 265 px | **0 px** | 265 px |
| **0.70 m** | 151 px | **0 px** | 151 px |
| **0.80 m** | 56 px | **0 px** | 56 px |
| 0.90 m | 0 px | 0 px | 0 px (cable leaves the frustum) |

Without the markers the cable is culled from 0.6 m onward. With them, visibility matches the unpartitioned control exactly at every offset.

The same effect is visible in a standalone partitioned stage containing only animated curves and per-env cameras: the curve tracks the unpartitioned control pixel-for-pixel out to 3.8 m of deformation and then drops to zero at 4.0 m — a hard cliff, not a fade. The threshold differs between the two because it is the distance from the curve's spawn extent to the edge of *its* partition's bounds, which depends on what else shares the partition.

## Changes

- `FrankaCableSceneCfg` gains `partition_bounds_marker_min` / `partition_bounds_marker_max`: two 1 mm static visual cubes at diagonally opposite corners of a box enveloping the cable's reachable space (the `cable_out_of_bounds` termination box padded by 0.25 m per face). No colliders, so they do not reach the physics model.
- `FrankaCableEnvCfg.__post_init__` sets both to `None` when scene partitioning is off, using `isaac_rtx_per_env_scene_partition_enabled()` — the same helper `IsaacRtxRendererCfg.enable_scene_partitioning` derives its default from.
- New `test_rendering_franka_cable_partition_visibility`: settles the cable, moves it 0.7 m in y, and asserts it still contributes pixels to `instance_id_segmentation_fast`. 0.7 m is clear of both the 0.6 m cull threshold and the 0.9 m framing limit; the camera clip range and resolution are widened for the probe so a zero count means culling, not framing. Counting segmentation pixels keeps the check exact and golden-image-free.
- Known Issues entry documenting the bug, the measured thresholds, and both workarounds; cross-referenced from the renderers guide and the cable limitations list.

## Type of change

- Documentation update

## Release backport

## Verification

- `test_rendering_franka_cable_partition_visibility` **fails without the markers** (`Cable vanished after moving 0.7 m in y (2197 px at spawn, 0 px after the move)`) and **passes with them**.
- Config gate checked in all three env-var states (`=1`, `=0`, unset) for both env cfgs.
- Stage inspection confirms both marker prims spawn under each `/World/envs/env_<i>` root, visible, at the intended world positions, and that they expand the env's bbox union from the cable's spawn extent to the pinned box.
- Partitioning confirmed genuinely active throughout: pointing a camera at a partition it does not belong to renders 0 px for every frame.
- `uv run isaaclab -f`

## Checklist


docs/source/refs/issues.rst
```diff
@@ -119,6 +119,49 @@ Until the upstream fix lands in Newton, please use the ``physx`` preset for
...
+  stays inside it wherever it moves. ``Isaac-Lift-Cable-Franka`` and
+  ``Isaac-Lift-Cable-Franka-Camera`` ship this workaround as the
+  ``partition_bounds_marker_min`` and ``partition_bounds_marker_max`` scene entries; copy
+  the pattern into custom cable environments. With the markers in place, cable visibility matches
+  an unpartitioned render exactly. The markers are static visual prims without colliders, so they
+  do not participate in physics, and they are dropped automatically when scene partitioning is off.
+
+* **Disable scene partitioning.** Set
+  :attr:`~isaaclab_physx.renderers.IsaacRtxRendererCfg.enable_scene_partitioning` to
+  ``False`` (or export ``ISAAC_LAB_ENABLE_ISAAC_RTX_PER_ENV_SCENE_PARTITION=0``) to opt
```
pbarejko: I consider the environment variable as internal detail. This has been used for testing and can be removed at any time. Can you remove it please and can we recommend to use `enable_scene_partitioning` only?

docs/source/refs/issues.rst
```diff
@@ -119,6 +119,49 @@ Until the upstream fix lands in Newton, please use the ``physx`` preset for
...
+  out of partitioning entirely, at the cost of the per-environment culling speedup.
+
+Both workarounds can be dropped once Kit updates bounding boxes for animated curves.
```
pbarejko: Let us do not make those promises. Can you remove this please?

source/isaaclab_tasks/isaaclab_tasks/core/lift/config/franka_soft/franka_cable_env_cfg.py
```diff
@@ -117,6 +131,23 @@ class FrankaCableSceneCfg(_FrankaSoftSceneCfg):
         init_state=CableObjectCfg.InitialStateCfg(pos=(0.32, 0.0, 0.011)),
     )
 
+    # Workaround for OMPE-105749: Kit RTX never refreshes the bounding box of an animated
+    # ``BasisCurves`` prim, so a partition sized from the cable's initial extent clips the cable
+    # once it moves outside that extent. These two static cubes pin the partition to the full
+    # workspace volume. FrankaCableEnvCfg drops them when partitioning is off, and they can go
+    # altogether once Kit updates animated-curve bounding boxes.
+    partition_bounds_marker_min: AssetBaseCfg | None = AssetBaseCfg(
```
pbarejko, line 139: How can we tell that this is solving the problem visually? Do we have rendering correctness test with it?

mataylor-nvidia: yes correctness test is added which is makes sure the cable falls as expected without disappearing
image

mataylor-nvidia: The following test:

https://github.com/isaac-sim/IsaacLab/pull/7328/changes#diff-ffe63c9a59f22b07366d6a6bd9efda82566582903687293ea99be50dc0ef1292R60

fails without the work around because the rod disappears, but passes with it

pbarejko (approved):

kellyguo11 (approved):

PR #6930, opened 2026-08-06
Enable deterministic Newton physics

kellyguo11:
# Description

Add an explicit `NewtonCfg.deterministic_mode` setting for Newton physics. The `not_guaranteed`, `run_to_run`, and `gpu_to_gpu` values map to Warp's corresponding deterministic modes.

Apply the selected mode to supported Newton solver kernels and deterministic contact ordering in the collision pipeline, including matching the pipeline's deterministic sort capacity to solver contact capacity. MJWarp on the GPU, XPBD, and Featherstone are supported. Unsupported solver paths (Kamino, implicit MPM, and MJWarp's MuJoCo CPU backend) are rejected only when a Newton determinism guarantee is explicitly configured.

The existing AppLauncher `--deterministic` RTX setting and behavior remain unchanged. Supported custom solver-configuration subclasses retain deterministic support.

No new dependencies are required. The pinned Newton revision already provides the required solver and collision-pipeline determinism APIs.

## Type of change

- New feature (non-breaking change which adds functionality)
- Documentation update

## Screenshots

Not applicable; this is a physics-configuration change.

## Validation

- `uv run --frozen python -m pytest source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py -k deterministic -q` — 8 passed, 84 deselected
- Subclass regression test verified to fail before the fix and pass after it
- `uv run --isolated --extra test -- sphinx-build -W --keep-going -j auto docs docs/_build/current` — build succeeded
- `uv run --frozen isaaclab -f` — all hooks passed before commit and before push

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `uv run --frozen isaaclab -f`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7334, opened 2026-08-25
[Fix] Make --deterministic reproduce training runs by configuring the physics backend

hujc7:
# Description

`--deterministic` configured PyTorch and the Isaac RTX renderer but never reached the physics solver, so training on Newton backends was not reproducible even with the flag passed. Two defaults moved out from under the flag between 3.0beta2 and GA:

| Default | v3.0.0-beta2 | GA |
|---|---|---|
| `CartpolePhysicsCfg.default` | `PhysxCfg()` | `NewtonCfg` MJWarp, `deterministic_mode="not_guaranteed"` |
| `MultiBackendRendererCfg.default` | `IsaacRtxRendererCfg()` | `NewtonWarpRendererCfg()` |

The physics switch landed in `0caae64dc7c` (#7066), absent from all three `v3.0.0-beta*` tags. At beta2 the flag worked because PhysX is run-to-run deterministic for rigid bodies **and** Isaac RTX was the default renderer; both premises were removed without re-wiring the flag.

`--deterministic` now sets `PhysicsCfg.deterministic` on the resolved physics config, in `apply_env_overrides()` — the existing CLI-to-cfg seam, after `scan()` resolves the backend and before the solver is built. That field is the backend-agnostic request; each physics manager translates it when the simulation starts:

- **Newton** derives `deterministic_mode="run_to_run"`, applies the MJWarp `disable_sensors` prerequisite on the GPU path, and leaves MuJoCo-CPU alone. An explicitly set `deterministic_mode` wins.
- **PhysX / OvPhysX** enable `enable_enhanced_determinism`. OvPhysX is best-effort and not verified end to end.

Validation stays with the backend: `NewtonManager._validate_deterministic_solver_cfg()` rejects an unsupported solver at solver initialization, so there is one policy and one set of error messages rather than a copy in the RL layer. Adding a backend no longer means editing `isaaclab_rl`.

**A determinism request that would starve a sensor is now refused.** Disabling MuJoCo Warp's sensors also skips its `rne_postconstraint` stage, which fills Newton's `body_qdd` / `body_parent_f`. The IMU, PVA and joint-wrench sensors read that state, so `Isaac-Ant`, `Isaac-Humanoid` and `Isaac-Repose-Cube-Shadow` — all defaulting to `newton_mjwarp` and feeding `joint_wrench` into their policy observations — would have trained on values that are never refreshed, with no error. `NewtonManager` raises at solver initialization, the only point where both the solver config and the registered sensors are visible.

The guard restates `{"body_qdd", "body_parent_f"}`, which Newton also states internally (`solver_mujoco.py:4360` as a set, `:5179` as an equivalent `or`-chain). That duplication is tracked upstream in https://github.com/newton-physics/newton/issues/4109, which asks for two things: Newton refusing the combination at the source, and exporting the field set as a public constant. Either lets this guard shrink — the constant and its sensor-tracking delete entirely once Isaac Lab pins a Newton that raises. Until then the guard is what prevents the silent case, so it stays.

Fixes NVBug 6658578 (P0, Isaac Lab 3.0 GA).

## Type of change

## Validation

Three `--deterministic` runs of `Isaac-Cartpole-Camera` (50 epochs, task defaults) on one L40, each in its own container:

```
det1 vs det2   55/55 checkpoint tensors bitwise identical
det1 vs det3   55/55 checkpoint tensors bitwise identical
ctrl1 vs ctrl2 41/55 differ          <- same task, no flag
```

The unflagged controls diverge, so the agreement above is the flag's doing rather than a task that is trivially reproducible. `Isaac-Ant --deterministic` fails at startup with the sensor message. MuJoCo-CPU could not be exercised: `SolverMuJoCo.get_max_contact_count()` raises `NotImplementedError` on that path, so it is unreachable in Isaac Lab today.

8 unit tests added; 200 pass across `test_entrypoints_common.py` and `test_newton_manager_abstraction.py`.

## Release backport

## Checklist


source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -475,13 +476,58 @@ def import_local_module(module_name: str, module_path: Path) -> ModuleType:
...
+    solver_cfg = getattr(physics_cfg, "solver_cfg", None)
+    solver_name = type(solver_cfg).__name__
+    if solver_name not in _DETERMINISTIC_NEWTON_SOLVERS:
+        raise ValueError(
+            f"--deterministic is not supported by {solver_name}. Use MJWarp on the GPU, XPBD, or"
+            " Featherstone, or drop --deterministic."
+        )
+    if getattr(solver_cfg, "use_mujoco_cpu", False):
+        raise ValueError(
+            "--deterministic is not supported by the MuJoCo CPU backend. Set"
```
kellyguo11: is CPU actually not deterministic? I would have thought CPU could be deterministic by default

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -475,13 +476,58 @@ def import_local_module(module_name: str, module_path: Path) -> ModuleType:
...
+            " MJWarpSolverCfg.use_mujoco_cpu=False, or drop --deterministic."
+        )
+    if solver_name == "MJWarpSolverCfg":
+        # MJWarp's internal sensor kernels mix atomic reduction families, which Warp's
+        # deterministic code generation rejects. Isaac Lab sensors read Newton state directly.
+        solver_cfg.disable_sensors = True
```
kellyguo11: would this cause any issues in tasks?

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -475,13 +476,58 @@ def import_local_module(module_name: str, module_path: Path) -> ModuleType:
...
+    if class_name != "NewtonCfg":
+        return
+
+    solver_cfg = getattr(physics_cfg, "solver_cfg", None)
+    solver_name = type(solver_cfg).__name__
+    if solver_name not in _DETERMINISTIC_NEWTON_SOLVERS:
+        raise ValueError(
+            f"--deterministic is not supported by {solver_name}. Use MJWarp on the GPU, XPBD, or"
+            " Featherstone, or drop --deterministic."
+        )
```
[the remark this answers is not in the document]
kellyguo11: Agreed. More fundamentally, this validation duplicates `NewtonManager._validate_deterministic_solver_cfg()`, which already uses `isinstance` and owns the backend-specific error messages. Please remove the solver whitelist/rejection logic from the RL layer: set the requested mode on the resolved config and let `NewtonManager` validate it during solver initialization. That preserves supported subclasses and prevents the two validation policies from drifting.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
kellyguo11: Confirmed: the pure MuJoCo CPU pipeline is documented as entirely deterministic and reproducible: https://mujoco.readthedocs.io/en/stable/computation/index.html#reproducibility. `use_mujoco_cpu=True` therefore should not be rejected merely because `--deterministic` was passed. I suggest leaving `NewtonCfg.deterministic_mode` unchanged for that path while still applying the Torch/rendering parts of the flag. The current Newton validation only establishes that Warp’s `DeterministicMode` guarantee is not applicable to the CPU solver; it is not evidence that CPU MuJoCo is nondeterministic.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
kellyguo11: For maintained Isaac Lab tasks this should be safe: `disable_sensors` disables MuJoCo Warp’s internal sensor computation, while Isaac Lab sensors read Newton state directly. It can still change behavior for custom/direct integrations that consume native MJWarp sensor outputs. Please scope this automatic mutation to the MJWarp GPU path for which the deterministic mode requires it, and document that native sensor-output limitation.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
```diff
@@ -475,13 +476,58 @@ def import_local_module(module_name: str, module_path: Path) -> ModuleType:
...
+
+    Args:
+        physics_cfg: Concrete physics config, as resolved by :func:`~isaaclab.app.scan`.
+
+    Raises:
+        ValueError: If the resolved backend cannot provide a determinism guarantee.
+    """
+    class_name = type(physics_cfg).__name__
+    if class_name in ("PhysxCfg", "OvPhysxCfg"):
+        physics_cfg.enable_enhanced_determinism = True
```
kellyguo11: This treats `OvPhysxCfg.enable_enhanced_determinism=True` as if it satisfies the same reproducibility guarantee. Our OvPhysX 0.5.10 Ant tests still produced three distinct 100-iteration reward hashes with `--deterministic`, enhanced determinism, `PXR_WORK_THREAD_LIMIT=1`, and single-threaded USDRT population. Please either exclude OvPhysX from the guaranteed path, document this as best-effort, or add an end-to-end OvPhysX determinism test before claiming the guarantee.

kellyguo11:
I checked this against current `develop` and #6930. The core CLI-to-physics connection is still needed: #6930 added `NewtonCfg.deterministic_mode` but deliberately left `AppLauncher --deterministic` unchanged; today Newton still defaults to `not_guaranteed`, Cartpole resolves to Newton MJWarp by default, and `apply_env_overrides()` does not propagate the flag into physics. Thus the reported training command configures Torch/rendering but not Newton physics.

I support a narrower version of this PR: keep the post-preset-resolution wiring, but let each physics backend own validation, special-case pure MuJoCo CPU as already deterministic, scope the MJWarp sensor mutation/limitation precisely, and document the physics behavior as RL-entrypoint behavior rather than universal `AppLauncher` behavior. Also, do not promise OvPhysX bitwise reproducibility without an end-to-end test; our 0.5.10 results still diverged with all available controls enabled.

source/isaaclab_rl/isaaclab_rl/entrypoints/common.py
kellyguo11: ovphysx determinism issues are addressed in ovphysx 0.5.11

kellyguo11:
run-ci

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -121,6 +121,10 @@ def _paused_gc():
     from isaaclab_newton.physics.newton_collision_cfg import NewtonCollisionPipelineCfg
 
 
+_SENSOR_STAGE_STATE_ATTRIBUTES = frozenset({"body_qdd", "body_parent_f"})
```
kellyguo11: is there a cleaner way to handle this? or perhaps this is more of a newton issue we should raise? this and the changes below could probably have a cleaner design?

hujc7: Yes, after diving deeper into this, I think it's a newton side bug. Filed issue and following up to see if updates can also happen in mjwarp

kellyguo11 (approved):

kellyguo11:
run-ci

PR #7339, opened 2026-08-25
[CLI] Report the required extra when an optional dependency is missing

hujc7:
# Description

Running an `isaaclab` command without the extra that provides its backend failed with a bare `ModuleNotFoundError` naming no remedy:

```
ModuleNotFoundError: No module named 'rl_games'
```

The extras are documented in the quickstart and installation guides, but nothing connected the failure to them, so every doc example had to repeat the flag to stay runnable. This maps the missing top-level module back to the extra that installs it:

```
rl_games is not installed. It is provided by the 'rl-games' extra:
  uv run --extra rl-games isaaclab train --rl_library rl_games --task Isaac-Cartpole
```

Covers every optional extra, not just RL frameworks — `isaacsim`, `ovphysx`, `ovrtx`, `viser`, `rerun`, `moviepy`, `robomimic`, `leapp`, `pytetwild`. The suggested command matches the environment: `uv run --extra ...` under uv, `pip install "isaaclab-dev[...]"` otherwise.

**Where the catch lives.** The RL backends import their framework inside `run()` deliberately, so `--help` works without it installed; the failure therefore surfaces in the dispatcher, not at module import. `isaaclab train` / `play` also run their script in a subprocess via `run_python_command`, so a handler in `cli()` alone never sees the child's exception. The catch is in `isaaclab_rl.entrypoints.dispatch` (inside the child) plus `isaaclab.cli` (for in-process subcommands).

**Deliberate omissions.** `rsl_rl` is unmapped because `rsl-rl-lib` is a base dependency, so it is never missing due to a skipped extra. `rlinf` is unmapped because its extra installs that framework's dependencies rather than the framework itself, so the hint would name the wrong remedy. An import error raised from inside an installed package is re-raised unchanged.

**Why the map is static.** No installed dist-info carries `Provides-Extra` (the extras live on the root `isaaclab-dev` project, which has no dist-info), and `packages_distributions()` only sees installed packages — precisely what a missing one is not. Reading `pyproject.toml` at error time would work only from a source checkout, not a wheel. The drift risk therefore moves into the test, which derives the expected mapping from `pyproject.toml`.

Relates to NVBug 6657697.

## Type of change

## Validation

Verified end to end in an environment without the extras:

| Command | Result |
|---|---|
| `isaaclab train --rl_library rl_games` | names `rl-games` |
| `isaaclab train --rl_library skrl` | names `skrl` |
| `isaaclab train --rl_library sb3` | names `sb3` (maps `stable_baselines3`) |
| `isaaclab train --rl_library rsl_rl` | no hint, training starts (base dependency) |
| the suggested `uv run --extra rl-games ...` | `Training time: 0.67 seconds`, exit 0 |

The suggested remedy was executed, not just inspected. 7 unit tests pass; both drift guards were verified non-vacuous by poisoning the map (an undeclared extra, and a wrong `torch` -> `sb3` mapping) — each fails the suite.

## Release backport

## Checklist

- [x] I have made corresponding changes to the documentation (the extras are already documented in the quickstart; this connects the failure to them, so no doc change was needed)

### DIFF
--- a/source/isaaclab/changelog.d/cli-missing-extra-hint.rst
+++ b/source/isaaclab/changelog.d/cli-missing-extra-hint.rst
@@ -1,0 +1,8 @@
+Added
+^^^^^
+
+* Added :func:`~isaaclab.utils.extras.missing_extra_hint`, which turns a missing optional
+  dependency into a message naming the extra that provides it and the command to install it.
+  ``isaaclab`` commands now report ``rl_games is not installed. It is provided by the
+  'rl-games' extra`` instead of a bare ``ModuleNotFoundError``. The suggested command matches
+  the environment: ``uv run --extra ...`` under uv, ``pip install "isaaclab-dev[...]"`` otherwise.
--- a/source/isaaclab/isaaclab/cli/__init__.py
+++ b/source/isaaclab/isaaclab/cli/__init__.py
@@ -67,6 +67,7 @@ def random_agent(args: list[str] | None = None) -> None:
         args = sys.argv[1:]
     run_python_command(ISAACLAB_ROOT / "scripts" / "environments" / "random_agent.py", args, check=True)
 
+
 def teleop(args: list[str] | None = None) -> None:
     """Run a live teleoperation, demonstration recording, or demonstration replay workflow.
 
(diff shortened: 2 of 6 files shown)


StafaH:
Thanks for the PR @hujc7! Closing as we don't want to over-engineer the missing dependency handling. The documentation is clear about optional dependencies and extras.

PR #6071, opened 2026-06-09
Fixed `usd_replicate()` authoring environment grid positions on nested prims (e.g. cameras), overwriting their local transforms.

huidongc:
# Description

Fixed `usd_replicate()` authoring environment grid positions on nested prims (e.g. cameras), overwriting their local transforms.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7342, opened 2026-08-25
[Backport release/3.0.0-beta2] FrameView local poses (#5677) + view-scoped Fabric selections (#6805)

pv-nvidia:
# Description

Backports two already-merged FrameView PRs to `release/3.0.0-beta2`, cherry-picked in merge order:

- #5677 — GPU-accelerated FrameView local poses + `xform_space_writer` API (upstream `69888c34e471`)
- #6805 — scale `FabricFrameView` selections to the view, not the stage (upstream `480370cdc5b0`, nvbug 6535498)

Two support commits, because this branch predates some prerequisites:

- `Shim needed for Arena compatibility.` (@amillane) adds `split_clone_template`, backported from #6071. Ordered first so no commit imports it before it exists.
- `fix: use the OVPhysX tensor binding API…` restores this branch's `physx.create_tensor_binding()` / `binding.read()`. The #5677 pick referenced `OvPhysxView`, which arrives with #6224 and is absent here, so `OvPhysxFrameView` init raised `ModuleNotFoundError`.

Conflict resolutions converge each touched file on its post-#6805 state from `develop`, so the backported code is byte-identical to what has been running and tested there. Two consequences of that choice:

- `get_scales()` now returns `ProxyArray` in all backends, as on `develop`. On this branch it previously returned `wp.array`, so callers that pass the result straight to Warp APIs need `.warp` (Torch users: `.torch`). This is the one API change the backport carries.
- The Fabric-hierarchy import stays unconditional (the optional import is #6499, not backported).

The only deviation from `develop`'s state is test-infra: device parametrization uses literal `["cpu", "cuda:0"]` because `test_devices()` does not exist on this branch.

## Type of change

- New feature (non-breaking change which adds functionality) — `FrameView.close()`, writer-scope API
- Breaking change (relative to this release branch only): `get_scales()` returns `ProxyArray` instead of `wp.array`, matching `develop`

## Tests

- `test_views_xform_prim_ovphysx.py` — 49 passed (26 `cpu` + 24 `cuda:0`; needs one process per device, `ovphysx<=0.3.7` binds device mode in C++ on first `PhysX(...)`)
- `test_views_xform_prim_newton.py` — 58 passed

### DIFF
--- a/docs/source/api/lab/isaaclab.utils.rst
+++ b/docs/source/api/lab/isaaclab.utils.rst
@@ -188,3 +188,16 @@ Warp operations
    :members:
    :imported-members:
    :show-inheritance:
+
+Warp Fabric kernels
+^^^^^^^^^^^^^^^^^^^
+
+Warp kernels for reading and writing Fabric ``Matrix4d`` attributes
+(``omni:fabric:worldMatrix`` / ``omni:fabric:localMatrix``) via
+:class:`wp.fabricarray` and :class:`wp.indexedfabricarray`. Used by
+:class:`~isaaclab_physx.sim.views.FabricFrameView` to keep child world and
+local matrices consistent without round-tripping through USD.
+
+.. automodule:: isaaclab.utils.warp.fabric
+   :members:
+   :show-inheritance:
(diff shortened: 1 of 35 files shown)


AntoineRichard (approved):
I don't think we should be doing this, but if they need it.

kellyguo11 (approved):

kellyguo11:
run-ci

PR #7343, opened 2026-08-25
Fix missed rays in base height reward

### LINKED ISSUE #1928: [Bug Report] base_height_l2 Bug
If you are submitting a bug report, please fill in the following details and use the tag [bug].

### Describe the bug

Using base_height_l2 in a manager-based reinforcement learning environment and using a raycaster sensor to adjust the target height based to account for the terrain, I found that with this setup, the reward value for this item would become '-inf' after about 80 iterations during training.

### Steps to reproduce

The settings for this item are: 
pnealize_base_hight = RewTerm(func = mdp.base_height_l2, weight = -0.9, params={"target_height":0.35, "sensor_cfg": SceneEntityCfg(name="raycaster")})

### System Info

- Commit: [e.g. 8f3b9ca]
- Isaac Sim Version: 4.5
- OS: Ubuntu 22.04
- GPU: RTX 4060Ti 16G
- CUDA: 12.1
- GPU Driver: [e.g. 470.82.01, this can be seen by using `nvidia-smi` command.]

### Checklist

### Acceptance Criteria

- [ ] Criteria 1
- [ ] Criteria 2

AntoineRichard:
# Description

RayCaster uses non-finite hit positions when rays miss the configured terrain mesh. The base height reward previously averaged those values directly, causing a single missed ray to produce an infinite penalty and destabilize training.

This change ignores non-finite ray hits when estimating terrain height and returns a neutral penalty when no valid terrain height is available. The implementation does not mutate the shared sensor buffer and preserves the existing behavior for finite scans and flat terrain.

Fixes #1928

## Type of change

## Release backport

- [x] Backport this pull request to the active release branch after it merges into develop

## Screenshots

Not applicable.

## Testing

- Focused MDP reward and environment unit tests: 5 passed
- Full pre-commit suite: passed
- Verified both new regression tests fail against the previous implementation
- Benchmarked equivalent GPU reductions and selected the fastest measured form

## Checklist

- [x] I have run the pre-commit checks with ./isaaclab.sh --format
- [x] My name already exists in CONTRIBUTORS.md

### DIFF
--- a/source/isaaclab/isaaclab/envs/mdp/rewards.py
+++ b/source/isaaclab/isaaclab/envs/mdp/rewards.py
@@ -105,18 +105,29 @@ def base_height_l2(
     asset_cfg: SceneEntityCfg = SceneEntityCfg("robot"),
     sensor_cfg: SceneEntityCfg | None = None,
 ) -> torch.Tensor:
-    """Penalize asset height from its target using L2 squared kernel.
+    """Penalize the asset height error using a squared L2 kernel.
 
-    Note:
-        For flat terrain, target height is in the world frame. For rough terrain,
-        sensor readings can adjust the target height to account for the terrain.
+    When :paramref:`sensor_cfg` is provided, the target height is relative to the mean height of the valid ray hits.
+    Non-finite ray hits are ignored. If all rays miss the terrain, the penalty is zero for that environment.
+
+    Args:
+        env: The environment.
+        target_height: Target asset height [m] relative to the terrain, or to the world origin when
+            :paramref:`sensor_cfg` is not provided.
+        asset_cfg: The asset whose height is penalized.
+        sensor_cfg: The ray-caster sensor used to estimate the terrain height.
+
+    Returns:
+        The squared height error [m²], shape (num_envs,).
     """
     # extract the used quantities (to enable type-hinting)
     asset: RigidObject = env.scene[asset_cfg.name]
     if sensor_cfg is not None:
         sensor: RayCaster = env.scene[sensor_cfg.name]
-        # Adjust the target height using the sensor data
-        adjusted_target_height = target_height + torch.mean(sensor.data.ray_hits_w.torch[..., 2], dim=1)
+        ray_hits_z = sensor.data.ray_hits_w.torch[..., 2]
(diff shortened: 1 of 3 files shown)


kellyguo11:
@StafaH @ooctipus would be good to get your review on this

source/isaaclab/isaaclab/envs/mdp/rewards.py
```diff
@@ -105,18 +105,29 @@ def base_height_l2(
...
+    Returns:
+        The squared height error [m²], shape (num_envs,).
     """
     # extract the used quantities (to enable type-hinting)
     asset: RigidObject = env.scene[asset_cfg.name]
     if sensor_cfg is not None:
         sensor: RayCaster = env.scene[sensor_cfg.name]
-        # Adjust the target height using the sensor data
-        adjusted_target_height = target_height + torch.mean(sensor.data.ray_hits_w.torch[..., 2], dim=1)
+        ray_hits_z = sensor.data.ray_hits_w.torch[..., 2]
```
ooctipus, line 127: instead of doing the magic in the code, why not expose a nan handling argument in the input arguments, maybe like `convert_inf: float = 0.0`

source/isaaclab/isaaclab/envs/mdp/rewards.py
```diff
@@ -105,18 +105,29 @@ def base_height_l2(
...
+        terrain_height = torch.nanmean(ray_hits_z.masked_fill(torch.isinf(ray_hits_z), torch.nan), dim=1)
```
ooctipus, line 128: do we really need convert inf to nan, the nan to 0.0... also 0.0 for missing terrain might not be idea, I'd imagine you want to bound inf to something like 10m or -inf to -10m instead of 0

AntoineRichard:
Taking a bit more time to read through all this, I agree with you @ooctipus. I think the issue, and this PR are taking the wrong route. Having -inf is expected if the object is not ontop of the terrain. The main question is should it be -inf, or something very large so that the network do not get NaNs. But maybe the user should use a different MDP in that case. I'm closing this PR. and I'll reply accordingly and close the attached issue.

PR #7347, opened 2026-08-25
Pin the OVRTX render product to the renderer's CUDA device

mataylor-nvidia:
# Description

IsaacLab creates its OVRTX `RenderProduct` without authoring `deviceIds`. OVRTX therefore picks the render device itself, places the render product on CUDA 0, and correctly returns a CUDA-0 DLPack buffer from `render_var.map(device=Device.CUDA)`. The camera tile extraction kernels, however, launch with `device=self._device` — the simulation device. On a multi-GPU machine running with `--device cuda:1`, `extract_all_tiles_kernel` then reads a CUDA-0 buffer from CUDA 1. On a machine without peer access between the two GPUs that is an illegal memory access and the run aborts with CUDA error 700; where peer access is available it silently reads across the link instead of failing.

This authors `uint[] deviceIds = [<cuda index>]` on the render product, resolving the index with the same helper the attribute mappings use (`ovrtx_mapping.cuda_device_id`, promoted from `_cuda_device_id` so both callers share one parser). Both initialization paths — legacy and ovstage — pass it, so every OVRTX buffer the renderer reads is allocated on the device its kernels launch on. The generated prim matches the form OVRTX documents in its own `doc-pin-render-product-to-gpu-0-usda` snippet:

```
def RenderProduct "RenderProduct" ( ... ) {
    rel camera = [</World/envs/env_0/Camera>]
    uint[] deviceIds = [1]
    ...
```

`device_id=None` keeps OVRTX's automatic assignment, so the USD builders behave exactly as before for any caller that does not pass a device.

This also resolves the `FIXME` in `_prepare_ppisp_hdr_source`, whose stated precondition ("until render var mapping can be constrained like transform bindings") is now met. Its `wp.clone` stays as a fallback for the case OVRTX logs as *"deviceIds ... not in the active device set"* and reverts to automatic assignment; the comment is updated to say so.

Fixes # (issue)

## Type of change

## Release backport

## Testing

```
uv run python -m pytest source/isaaclab_ov/test/test_ovrtx_usd.py \
  source/isaaclab_ov/test/test_ovrtx_renderer_kernels.py \
  source/isaaclab_ov/test/test_ovrtx_renderer_contract.py \
  source/isaaclab_ov/test/test_ovrtx_clone_plan.py \
  source/isaaclab_ov/test/test_ovrtx_deformable_bindings.py
# 107 passed

uv run python -m pytest source/isaaclab/test/renderers/test_renderer_factory.py
# 4 passed
```

Three tests added, each verified to fail on the unfixed sources and pass with the fix:

* `test_render_product_pins_device_ids_to_the_requested_cuda_device` — parses the authored snippet through `Sdf.Layer` and asserts `deviceIds` is a `UIntArray` equal to `[1]`.
* `test_render_product_omits_device_ids_when_no_device_is_given` — nothing authored when no device is given.
* `test_create_render_data_pins_the_render_product_to_the_spec_device` — covers the wiring: a spec on `cuda:1` produces a render product pinned to device 1.

`_make_ovrtx_renderer_without_backend` bypasses `__init__`, so it now sets `_device` to the same default `__init__` uses.

Not verified here: the multi-GPU, no-peer-access machine from the report. The change is validated at the USD authoring and renderer wiring level only.

## Checklist

### DIFF
--- a/source/isaaclab_ov/changelog.d/ovrtx-render-product-device-ids.rst
+++ b/source/isaaclab_ov/changelog.d/ovrtx-render-product-device-ids.rst
@@ -1,0 +1,9 @@
+Fixed
+^^^^^
+
+* Fixed an illegal memory access (CUDA error 700) when rendering with OVRTX on a device other than
+  ``cuda:0``. The OVRTX render product is now pinned to the renderer's CUDA device through its
+  ``deviceIds`` attribute, so its render var buffers are allocated on the same device as the Warp
+  kernels that extract camera tiles from them. Previously OVRTX chose the device itself, which on a
+  multi-GPU machine placed the buffers on ``cuda:0`` while the extraction kernels ran on the
+  simulation device.
--- a/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_mapping.py
+++ b/source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_mapping.py
@@ -22,8 +22,11 @@ from typing import Any
 import warp as wp
 
 
-def _cuda_device_id(device: str) -> int:
+def cuda_device_id(device: str) -> int:
     """CUDA device index parsed from a Warp device string, e.g. ``"cuda:1"`` -> ``1``.
+
+    Shared by the attribute mappings below and by the render product's ``deviceIds``, so both
+    resolve the renderer's device the same way.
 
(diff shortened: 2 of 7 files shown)


pbarejko (approved):

kellyguo11 (approved):

mataylor-nvidia:
I have put in the PR description to backport this to` release/3.0.0`

PR #7366, opened 2026-08-26
Run Docker CI from PR comments on develop

mataylor-nvidia:
## Summary

Port #7059 to `develop`. Cherry-pick of 367e498138c233e56896f3deb818aaed8a094dd8 with the two `build.yaml` conflicts resolved against develop's copy.

@kellyguo11 — **draft, because merging this blocks every open `develop` PR until its author asks for CI.** Details under Impact. Timing and announcement are your call; I'll mark it ready when you say.

## Why

#7059 merged into `release/3.0.0-beta2` only. That branch is the repository default, so the `run-ci` command itself is already live everywhere: `issue_comment` workflows always run from the default branch. `build.yaml` is not, because `pull_request` workflows are read from the PR's merge ref, so develop's own copy governs develop PRs.

Two consequences today:

1. Docker and GPU CI still starts automatically on every `develop` push. 192 of the 198 open PRs target `develop`, so the change currently saves nothing on ~97% of PR traffic.
2. Worse: commenting `run-ci` on a `develop` PR **silently succeeds**. The command workflow authorizes, applies and removes `ci:run-docker`, and writes "Requested Docker CI for PR #N" to the step summary — but develop's `build.yaml` has no `labeled` trigger, so nothing starts. The requester gets a green check and no CI.

`release/3.0.0` is in the same state (2 open PRs). `main` carries a different, legacy workflow set and is unaffected.

## Conflict resolution

`develop`'s `build.yaml` has diverged substantially from `release/3.0.0-beta2`'s — develop has the `push:` post-merge trigger and the `detect-changes` composite action, beta2 has neither. The `types: [labeled]` change auto-merged; two hunks did not:

- **Trigger comment.** Kept develop's post-merge `push` path, which `release/3.0.0-beta2` does not have, folded into the new gating explanation.
- **`cancel-in-progress`.** Took the label expression over develop's `${{ github.event_name == 'pull_request' }}`. It is a strict superset here: `push` and `workflow_dispatch` carry no `github.event.label`, so it evaluates `false` for them and preserves exactly the post-merge protection the old expression provided. Noted inline so the next reader does not have to re-derive it.

`run-docker-ci.yml` is inert on `develop` — it only ever executes from the default branch — but is carried here anyway so the next release branch cut from `develop` keeps the command. Without it, `run-ci` dies repo-wide at the next release cut.

`detect-changes`'s `triggered-jobs-pr` text is step-summary-only and remains accurate, so it is unchanged.

## Impact

The `develop` ruleset (`refs/heads/develop` and `refs/heads/release/*`, enforcement active) requires `Build Base Docker Image`, `Build cuRobo Docker Image`, `test-curobo`, `rendering-correctness-kitless (legacy)`, `isaaclab (core) [1-3/3]` and the rest of the Docker-produced contexts. Every one of those is produced by `build.yaml`, so this port covers the required-checks surface completely.

Once this merges, those contexts are produced only by a labeled or dispatched run, so **every open `develop` PR sits on "Expected — waiting for status" until its author comments `run-ci`.** That is the intended design, not a regression, but it needs an announcement rather than a quiet merge.

## Out of scope — `build.yaml` is not develop's whole GPU surface

`develop` carries self-hosted workflows that `release/3.0.0-beta2` does not, and they still start automatically on `pull_request: [opened, synchronize, reopened]`:

| Workflow | Runner | Gate today | On beta2? |
|---|---|---|---|
| `kitless-docker.yml` | self-hosted | `detect-changes` only | no |
| `arm-ci.yml` | self-hosted | `detect-changes` only | no |
| `test-multi-gpu-pytest.yaml` | self-hosted | path filter | no |
| `test-multi-gpu.yaml` | self-hosted | path filter | yes |

None of these produce a required check, so this is purely a cost question rather than anything blocking. Left alone deliberately: extending `run-ci` to them is a separate decision with its own fallout, not something to fold into a port. Happy to do it in a follow-up if you want the gate to cover the whole GPU surface on `develop`.

## Validation

- `pre-commit` passes on all three changed files.
- Both workflows parse as YAML.
- **This PR tested the mechanism on itself, and the gate holds.** Because `pull_request` workflows load from the merge ref, and this PR's merge ref already contains `types: [labeled]`, `Docker + Tests` did not start when the PR opened — confirmed, no `Build Base Docker Image` / `test-curobo` / `isaaclab (core)` contexts are present, while the ungated `Kit-less Docker Image` and `Installation Tests` workflows did start. The required contexts stay missing until someone comments `run-ci` here, which exercises the full path end to end against `develop`: the command workflow runs from the default branch, authorizes, labels, and the merge ref's `build.yaml` picks the label up. If CI turns green on this PR after a `run-ci` comment, the port is proven.

Worth noting separately: #7059's command workflow has fired exactly once since merging (run 32997166039) and that run skipped at the `if`. The App-token path is still unproven in production on either branch, so a `run-ci` here doubles as its first real exercise.

## Release backport

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -46,6 +46,9 @@ To upload images to a PR -- simply drag and drop an image while in edit mode and
 
 ## Checklist
 
+Docker and GPU tests run on demand. Push the commits you want tested, then
+comment `run-ci` on the pull request.
+
 - [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
 - [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
 - [ ] I have made corresponding changes to the documentation
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,8 +4,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 # region help
-# Every test job runs on every PR and merge queue candidate. Pushes to protected branches run the
-# post-merge integration tests.
+# Docker builds and their dependent tests run only on a labeled PR event, a push to a protected
+# branch, or a manual workflow dispatch. Do not filter the label with a job-level ``if``:
+# GitHub treats skipped required jobs as successful status checks.
+#
+# GitHub cannot filter label names at trigger time, so adding any label to a PR
+# starts this workflow. Only a ``ci:run-docker`` event cancels an in-flight run;
+# other label events queue behind it so that a triage label cannot cancel a
+# build whose required checks have already been reported.
+#
+# Pushes to protected branches still run the post-merge integration tests.
 #
 # =============================================================================
 # CI DEBUGGING TIPS
@@ -53,7 +61,7 @@ on:
       - develop
       - 'release/**'
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [labeled]
     branches:
       - main
       - develop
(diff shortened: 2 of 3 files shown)


mataylor-nvidia:
run-ci

kellyguo11 (approved):

PR #7370, opened 2026-08-26
Cover MPM particle rendering in unit and golden-image tests

mataylor-nvidia:
# Description

`create_mpm_particle_visualization` builds the `UsdGeom.Points` prims that render MPM particle clouds on the USD stage. Its two unit tests passed, but they only exercised the material-binding branch — the fixture used a single environment of all-zero positions and uniform widths, so nothing observed the geometry the renderer actually consumes.

Mutating the spawner confirmed the gap. All but one of these injected bugs left the suite green:

| Injected bug | Rendering symptom | Old suite | New suite |
| --- | --- | --- | --- |
| `positions[env_idx]` → `positions[0]` | every environment renders env 0's particles | passed | fails |
| widths array zeroed | particles render invisible | passed | fails |
| `SetResetXformStack(True)` dropped | world-frame points inherit the per-env translation, so every cloud is double-offset | passed | fails |
| binding skipped when a material exists | material silently ignored | fails | fails |
| binding applied when no material prim was authored | dangling `material:binding` instead of the display-color fallback | passed | fails |
| path/environment count guard removed | environments silently dropped | passed | fails |
| guard moved after the prims are defined | mismatch leaves a half-built stage | passed | fails |

The all-zero single-environment fixture was the root cause: with one environment and no distinct values, per-environment indexing and widths are simply unobservable.

## Changes

- The fixture now spans two environments with distinct positions (`np.arange(...)`) and per-particle widths, at environment-namespaced paths matching the `{ENV_REGEX_NS}/<asset>/Particles` layout `MPMObject` uses.
- `test_each_environment_renders_its_own_particle_slice` asserts each prim's points against its own slice, plus widths, display color, and the returned paths.
- `test_particle_clouds_ignore_the_inherited_environment_transform` covers the world-frame contract — points are authored in world space, so the prims must reset the environment's xform stack.
- `test_prim_path_count_must_match_environment_count` covers the documented `ValueError`, and asserts no prim is authored when it fires.
- `test_missing_visual_material_prim_falls_back_to_display_color` is strengthened: the old `GetDirectBindingRel().GetTargets() == []` also passes on an *invalid* relationship, so it now asserts the direct binding's material path is empty.
- `test_visual_material_is_bound_to_every_particle_cloud` replaces `test_particle_visualization_uses_standard_material_binding_helper`. The real `bind_visual_material` works on an in-memory stage, so mocking it was unnecessary; the test now asserts the binding authored on the stage rather than that a helper was called.

No source behaviour changes — `visualization.py` is untouched and was verified correct on every point above.

## Golden-image rendering coverage

The unit tests above assert the USD scene description handed to the renderer; they do not render. Nothing anywhere rendered MPM particles, so a cloud that came out invisible or in the wrong place passed every existing test. There is no MPM arm in any `test_rendering_*` suite today.

`test_rendering_mpm_particles` adds one, capturing the UR10 particle-push pile through the Isaac RTX renderer at 4 envs.

- **Newton only.** MPM runs on Newton's coupled MPM/MJWarp solver, so there is no PhysX or OVPhysX arm. `UR10ParticlePushEnvCfg` pins `sim.physics` directly rather than exposing physics presets, so only the renderer is selected through Hydra.
- **RTX only.** The Warp rasterizer draws particles straight from Newton state as synthetic hits, which is a separate code path from the `UsdGeom.Points` clouds under test.
- **Test-local camera.** Attached through a scene subclass in `rendering_test_utils.py`, so the contrib task keeps its state-only public config.
- **Settled pile.** The capture runs 30 environment steps first (half a second at the task's 2x decimation over a 1/120 s step) so the pile slumps out of its reset lattice into a granular mound, and the points prims re-sync on every step rather than once. Settling cost some stability - the granular surface left the `simple_shading` AOVs at SSIM 0.9827-0.9850 against the 0.985 default, passing only on retry - so the suite uses a measured 0.97 SSIM threshold that two consecutive fresh processes clear with no retries.
- **Pinned resets.** Every random draw in `randomize_push_scene` is pinned (`reset_cycle`, zeroed pile yaw and particle jitter, a single curriculum level). Seeding alone is not enough: the pose bank is built once per process, so each run otherwise renders a different robot pose and pile.

### Is the golden actually sensitive to particles?

Zeroing the particle widths so the cloud renders invisible:

| | normal | particles invisible |
| --- | --- | --- |
| differing pixels | 0.00% | 45.57 - 65.64% |
| SSIM | pass | 0.6772 - 0.6850 |

Well outside both the 8% pixel gate and the 0.97 SSIM gate, so the goldens track the particle path rather than the surrounding workcell, and the loosened SSIM threshold retains a wide margin over the regression it exists to catch.

### Golden provenance - these need regenerating on CI

The goldens were captured on an **NVIDIA A40**, and this machine does not agree with the committed goldens. An untouched `test_rendering_cartpole[newton-isaacsim_rtx_renderer-static]` fails here at **4.15% differing pixels against a 1.50% threshold** (SSIM 0.987), reproducibly across all three flaky retries. Whatever the cause, goldens authored here encode this machine's output, so please regenerate these 13 PNGs from a CI runner's artifacts before merging.

One related finding worth recording: the ~37% rgb instability I first hit was **not** test flakiness. It comes from the flaky retry constructing a *second* env in the same process, which renders RTX colour differently; geometry AOVs are unaffected. A fresh process matches a fresh-process golden exactly. So a retry cannot rescue this suite - if attempt 1 fails, attempts 2 and 3 fail systematically.

## Type of change

- Test-only change (no runtime behaviour change, no public API change)

## Release backport

## Verification

- `uv run python -m pytest source/isaaclab_newton/test/sim/test_mpm_visualization.py` — 5 passed.
- Full MPM suite (`test_mpm_visualization.py` + `test_mpm_spawners.py`) — 15 passed.
- Each of the seven mutations above was applied to `visualization.py` in isolation and re-run: each is caught by exactly one test, so the coverage is non-overlapping rather than redundant.
- `uv run python -m pytest source/isaaclab_tasks/test/core/test_rendering_mpm_particles.py` - 4 passed, 1 skipped (`motion_vectors` skips on Newton) in **two consecutive fresh processes**, with no failed attempts and no retries.
- `pre-commit run ruff` / `ruff-format` on the changed files, and `tools/changelog/cli.py check` against `develop`.

## Checklist

- [ ] I have made corresponding changes to the documentation — not applicable, no public API or documented behaviour changed


maxkra15:
great catch, thanks!

maxkra15 (approved):

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

kellyguo11 (approved):

mataylor-nvidia:
run-ci

PR #7371, opened 2026-08-26
Fix the unconverged first frame in the quickstart media

mataylor-nvidia:
# Description

The first frame of every quickstart clip was visibly unconverged — dark blotches across the floor that clear by the second frame.

The cause is a render setting, not the media pipeline. `ViewerRTX` hard-codes `omni:rtx:quality` to `0` and `omni:rtx:waitForEvents` to `[]` to minimize `step()` latency for an interactive viewer, which disables the RTX path tracer's quality convergence loop, and exposes no way to change it: `_init_ovrtx` authors the render product's attributes, exports the stage to a temporary USD, then hands that file to the renderer — so anything written to the stage afterwards is ignored.

This PR adds a config hook that authors attributes in that window, then uses it to re-enable the convergence loop for the quickstart capture.

### 1. `NewtonRTXVisualizerCfg.render_settings`

Takes `{name: (usd_type_name, value)}` entries and authors them onto the render product from `_add_camera_lights_and_render_product`, the only point where they reach the renderer. Defaults to empty, so nothing changes unless a caller opts in.

```python
NewtonRTXVisualizerCfg(render_settings={"omni:rtx:quality": ("Int", 100)})
```

Two deliberate choices:

- The type is a **string**, not an `Sdf.ValueTypeNames` object, because `configclass` deep-copies every field and those objects raise `RuntimeError: Pickling of "pxr.Sdf.ValueTypeName" instances is not enabled`.
- It is given **per entry rather than inferred**, since inference cannot distinguish `Int` from `UInt` and would misread a two-token array such as `["AllLoadingFinished", "OnlyOnFirstRequest"]` as a type-and-value pair.

### 2. Quickstart capture

Sets `omni:rtx:quality` back to `100` — Kit's own default for `RenderProduct` prims (`RenderProductQualityPrimHandler`) — and regenerates the media. An earlier revision of this PR raised `omni:rtx:quality:minSpp` to `32` instead; re-enabling the full convergence loop matches or beats that on every axis measured below, without a magic sample-count constant that has to be re-tuned per scene.

Measured against the pixels that stay static across the rest of each clip (own script, consistent across all three variants — absolute numbers are not directly comparable to an externally-computed metric):

| | f0 defective px | f0 mean deviation | flicker | bytes |
| --- | --- | --- | --- | --- |
| `agent-comparison` before | 24,480 | 4.419 | 0.0004 | 1,105,959 |
| `agent-comparison` minSpp=32 | 683 | 0.616 | 0.0001 | 849,531 |
| **`agent-comparison` quality=100** | **339** | **0.518** | **0.0001** | **855,557** |
| `task-sampler` before | 17,530 | 4.572 | 0.0004 | 651,499 |
| `task-sampler` minSpp=32 | 522 | 0.493 | 0.0001 | 453,339 |
| **`task-sampler` quality=100** | **364** | **0.443** | **0.0001** | **456,433** |

`quality=100` cuts first-frame defective pixels roughly in half again versus `minSpp=32`, at effectively the same file size (+0.7-1.4%). Flicker is unchanged within this metric's resolution for both fixes.

Also fixes the capture config entry points, which named a module `capture` while the file is `capture_quickstart.py`. `importlib` raised `ModuleNotFoundError`, so the four clips routed through `play.py` (`play`, `cartpole`, `g1`, `kuka_allegro`) **could not be generated at all** — only the two that invoke the script directly worked.

### Settings ruled out

Measured and rejected: `pt:samplesPerPixel`, `pt:samplesPerIteration`, `pt:denoising:enabled`, `indirectDiffuse:denoiser:enabled`, `reflections:denoiser:enabled`, `dlss:frameGeneration`, `post:aa:limitedOps`, `ghostFactor`, and camera-side `autoExposure` are all no-ops for this artifact. Setting `omni:rtx:waitForEvents` per frame (instead of, or in addition to, `omni:rtx:quality`) resets temporal accumulation and produces several times the flicker; `omni:rtx:quality=100` alone converges the headless capture without touching `waitForEvents`, so it is left at `ViewerRTX`'s own default (`[]`).

Fixes # (no issue filed)

## Type of change

## Release backport

## Screenshots

`task-sampler` frame 0, G1 panel at 2× — note the dark blotches on the floor before:

| Before | After |
| ------ | ----- |
| before | after |
| image |  image |

The "before" link resolves against `develop` and will break once this merges. The frame-0 stills above were captured against the `minSpp=32` revision; the gif links themselves reflect the current `quality=100` revision.

## Checklist

Verified by regenerating the media end to end (`generate_quickstart.sh`, exit 0, all six clips) and measuring the result as above, rather than by unit test. The local `check-changelog-fragments` hook reports failures for packages this PR does not touch — it diffs against `origin/develop`, which on this checkout is a fork behind `upstream/develop`. `isaaclab_visualizers` is satisfied by the included fragment.


tools/docs/media/generate_quickstart.sh
```diff
@@ -30,6 +90,18 @@ cd "${REPO_ROOT}"
 UV_EXTRAS="ovrtx,ovphysx,video"
 CAPTURE_STEPS=300
 
+# capture_quickstart.py renders 640x480 per clip. The quickstart figures are ``width:100%``
+# in a ~750px content column, so each of the three stacked panels lands on ~250 CSS px —
+# about 500 device px on a HiDPI display. The previous 320-wide capture was upscaled there,
```
StafaH: AI comment is talking about old logic and how it changed it, we can remove in favor of current facts about the code, or no comment at all. Also applies elsewhere

StafaH (approved):

tools/docs/media/generate_quickstart.sh
mataylor-nvidia: I removed the old ai generated code

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

kellyguo11 (approved):

kellyguo11:
run-ci

mataylor-nvidia:
## FPS impact if `render_settings` were applied globally

Ran a before/after benchmark to check whether forcing the path tracer's full quality-convergence loop (`omni:rtx:quality=100`, as this PR re-enables for quickstart capture) on **every** step has a training-throughput cost, if it were made the global default instead of opt-in.

### Renderer

Yes — this exercises OVRTX specifically. `NewtonRTXVisualizerCfg` / `NewtonViewerRTX` wrap Newton's `ViewerRTX`, which is backed by the `ovrtx` (Omniverse RTX) path-tracer module (`libovrtx`), the same renderer this PR's `_add_camera_lights_and_render_product()` hook authors `omni:rtx:quality` onto. It's a distinct code path from `NewtonGLVisualizerCfg` (OpenGL rasterizer), so these numbers say nothing about the GL viewer or about environments not using an RTX visualizer at all (the vast majority of training runs, which don't attach a visualizer).

### Setup

- Task: `Isaac-Cartpole-Direct`, `--num_envs 4096`, physics backend Newton (MJWarp).
- Launched via `uv run isaaclab benchmark runtime --task Isaac-Cartpole-Direct --num_envs 4096 --num_steps 1000 --warmup_steps 50 --visualizer newton_rtx`, i.e. `NewtonViewerRTX`/OVRTX stepped every simulation frame (`update_frequency=1`).
- No display attached (`DISPLAY` unset) → the viewer initializes OVRTX headless via EGL and renders offscreen each step; no window is opened.
- Hardware: single NVIDIA L40, driver 550.127.05, CUDA 12.8.
- "Before" = current code, `omni:rtx:quality` hard-coded to `0` by `ViewerRTX` (this PR's baseline).
- "After" = this PR's `NewtonRTXVisualizerCfg.render_settings` mechanism, but with the *default* changed to `{"omni:rtx:quality": ("Int", 100)}` so every `newton_rtx` visualizer picks it up unconditionally. This is **not** how the PR ships it — as merged, `render_settings` defaults to `{}` and is only set by the quickstart capture config — I patched the default locally purely to measure the worst case of "what if this were on for everyone."
- Same binary/config otherwise; only `render_settings`'s default value differed between the two runs. Reverted after measuring.

### Results

| Run | `omni:rtx:quality` | Total FPS (mean) | Wall time (1000 steps) |
| --- | --- | --- | --- |
| Before | 0 (hard-coded) | 827,582 | 4.95 s |
| After (global) | 100 | 799,286 | 5.12 s |

~3.5% FPS regression at 4096 envs if the convergence loop were forced on globally. Consistent across two independent runs (300-step and 1000-step) despite fairly high per-iteration variance (std ~38% of mean in both configs — a property of the async OVRTX render dispatch itself, not of this setting).

**Not a concern for this PR as written** — `render_settings` defaults to `{}`, so no existing training/env path is affected; this only matters if someone opts into `quality=100` (or similar) for their own visualizer config, or if a future change makes it the default.

PR #2332, opened 2025-04-17
Adds time-based mdp (observation) functions.

TheIndoorDad:
# Description

As discussed in #2284, I was writing an observation function to pass the time remaining in an episode (in seconds) to an observation term in a Manager-based environment, and found this could not be done without modifying `ManagerBasedRLEnv` to initialize `episode_length_buf` before managers are loaded. Here is a summary of changes made:

* Added initialization of `episode_length_buf` in :meth:`load_managers()` of :class:`~isaaclab.envs.ManagerBasedRLEnv` to make it available for use in mdp functions. Note: existing initialization in :meth:`__init__` left in place in case it is needed for other use cases. Potentially redundant? Assess.
* Added :attr:`~isaaclab.envs.ManagerBasedRLEnv.curr_episode_length` to :class:`~isaaclab.envs.ManagerBasedRLEnv` which returns reshaped ``episode_length_buf`` so it is visible as an attribute in the documentation.
* Added time observation functions to `~isaaclab.envs.mdp.observations` module, :func:`~isaaclab.envs.mdp.observations.current_time_s` and :func:`~isaaclab.envs.mdp.observations.remaining_time_s`.

I'm not certain whether the documentation will be updated automatically or if there are further steps I need to take. When I build the documentation on my machine it is updated, but the outputs are ignored by git. Please let me know if there's anything else I need to do.

I could also use some advice on tests (apologies in advance for my lack of experience here, my background is not in software dev). Locally I modified the `Isaac-Velocity-Rough-Anymal-C-v0` task to add the two new observation functions, and began to train a policy in rsl_rl using the provided `scripts/reinforcement_learning/rsl_rl/train.py` script, and both were available to be viewed and appeared to be working correctly. I tried to run the existing suite of unit tests but it gave me an error I don't understand (see below). I also started to create a new script similar to [`isaaclab/test/envs/check_manager_based_env_anymal_locomotion.py`](https://github.com/isaac-sim/IsaacLab/blob/7de6d6fef9424c95fc68dc767af67ffbe0da6832/source/isaaclab/test/envs/check_manager_based_env_anymal_locomotion.py) but that would have required a policy trained using the new observation functions (which I can produce, but wasn't sure if that would be worthwhile since it wouldn't be available to others). 

Output when running `./isaaclab.sh -t`:
```
[INFO] Warm starting the simulation app before running tests.                                                                                                                                                                                                                                                                        
[...]

PR #7373, opened 2026-08-26
[Fix] Surface the real error when environment construction fails, instead of an AttributeError from __del__

hujc7:
# Description

When environment construction fails partway through `__init__`, the destructor emits a *second*,
misleading traceback ahead of the real failure:

```
Exception ignored in: <function ManagerBasedEnv.__del__ at 0x...>
AttributeError: 'ManagerBasedRLEnv' object has no attribute '_is_closed'
```

`ManagerBasedEnv.__del__` (and the `DirectRLEnv` / `DirectMARLEnv` equivalents) read `self._is_closed`
unconditionally. That attribute is assigned by `ManagerBasedEnv.__init__`, but a **subclass may run
fallible setup before delegating to it**: `ManagerBasedRLEnv.__init__` allocates `episode_length_buf`
and `reset_buf` on `cfg.sim.device` before calling `super().__init__()`. A CUDA OOM, an invalid device,
or a malformed config raises there, leaving `_is_closed` unset.

The original exception still propagates to the caller — it is not swallowed. But Python cannot raise
out of `__del__`, so it routes the `AttributeError` to `sys.unraisablehook`, which prints it on
unbuffered stderr. The spurious traceback therefore lands *first*, naming a different exception type
at a different location, which is what makes the real failure hard to find.

This defaults the lookup to `True`, so an environment that never finished initializing is treated as
already closed and needs no teardown. That matches the existing contract: `_is_closed = True` is
already the first statement of `ManagerBasedEnv.__init__` for exactly this reason, and genuine
partial-init cleanup is handled explicitly by the `try/except` around `_init_sim()` that releases the
`SimulationContext` singleton. Applied to all three env classes, which share an identical destructor.

**Why not move the allocation after `super().__init__()`?** Manager construction calls every
observation term once to measure its output shape (`ManagerBase.__init__` -> `_prepare_terms` ->
`term_cfg.func(env)`), and terms such as `mdp.observations.current_time_s` read
`env.episode_length_buf`. The early allocation was introduced deliberately in #2332 for that reason.
Guarding the destructor is the change that does not perturb construction order.

Fixes NVBugs 6604738. Related: 6591313 (`AssetBase.__del__`) and 6311065 (`direct_rl_env.__del__`) —
both fixed, neither covers this path; `AssetBase` uses a different attribute (`_is_initialized`).

## Type of change

## Release backport

## Validation

`test_env_destructors.py` gains one parametrized case covering a partially constructed env. Verified it
fails before the fix and passes after:

| | Result |
| --- | --- |
| Without the fix | `3 failed` — `AttributeError` at `manager_based_env.py:248` |
| With the fix | `14 passed` (11 existing + 3 new) |

```bash
uv run --extra test python -m pytest source/isaaclab/test/envs/test_env_destructors.py -q
```

## Known gap

This guards `__del__` only. `_is_closed` is also read unguarded in `close()` at
`manager_based_env.py:605`, `direct_rl_env.py:591`, `direct_marl_env.py:593`,
`manager_based_rl_env.py:358` and `leapp_deployment_env.py:427`, so
`try: env = ManagerBasedRLEnv(cfg) finally: env.close()` still raises. A follow-up declaring
`_is_closed: bool = True` as a class attribute would cover every read site at once.

## Checklist


hujc7:
run-ci

PR #7387, opened 2026-08-27
Fix/newton site shape classification

NeoZng:
# Description

"NewtonSiteFrameView" currently rejects any frame expression that matches "model.shape_label" after Newton model finalization. This is incorrect because Newton's shape table contains more than collision geometry: MJCF "<site>" elements and visual-only shapes can also appear there without participating in collisions.

This causes a lifecycle-dependent bug for MJCF-imported sites. A site can resolve before "sim.reset()", but after model finalization the same path is rejected simply because it exists in "model.shape_label".

This change keeps the fix intentionally minimal. Instead of treating every Newton shape record as a collider, the post-finalization path now rejects only shapes carrying "ShapeFlags.COLLIDE_SHAPES" or "ShapeFlags.COLLIDE_PARTICLES". Non-colliding shape records, including MJCF sites and visual-only shapes, remain valid FrameView targets.

The pre-finalization path continues to enforce the same non-physics-frame contract by rejecting prims with "UsdPhysics.CollisionAPI". No changes are made to the broader FrameView lifecycle, site injection behavior, transform ownership, shape scaling, or the relationship between "FrameView" and "FrameTransformer".

Regression coverage uses the existing Newton FrameView test suite and verifies the concrete failure mode after finalization: an MJCF-imported site remains resolvable, and a visual-only shape is not rejected by the collision guard. Existing collision-shape tests continue to cover rejection of actual colliders.

# Type of change

* Bug fix (non-breaking change which fixes an issue)

# Release backport

# Checklist

Docker and GPU tests run on demand. Push the commits you want tested, then comment "run-ci" on the pull request.

* [x] I have read and understood the contribution guidelines
* [x] I have run the pre-commit checks with "./isaaclab.sh --format"
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective
* [x] I have added a changelog fragment under "source/<pkg>/changelog.d/"
* [x] I have added my name to "CONTRIBUTORS.md" or my name already exists there


AntoineRichard:
Hi @NeoZng thanks for PR and reporting this bug. @camevor will look into it!

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
```diff
@@ -32,13 +33,20 @@
...
+_SITE_FLAG = int(ShapeFlags.SITE)
+_COLLISION_FLAGS = int(ShapeFlags.COLLIDE_SHAPES | ShapeFlags.COLLIDE_PARTICLES)
 
 
 def _has_regex_tokens(pattern: str) -> bool:
     """Return whether ``pattern`` contains regex metacharacters (i.e. is not a literal path)."""
     return any(token in _REGEX_TOKENS for token in pattern)
 
 
+def _is_non_colliding_site(flags: int) -> bool:
```
camevor: This adds sites to the allowlist, but still excludes e.g. visual shapes. Simply blocking on collision shapes would be preferable.

camevor (commented):
Thanks for your PR, @NeoZng! The disparity in validation between the pre-finalization and post-finalization path is indeed unwanted. Your observation does raise a few related questions (e.g. why are both paths needed), so I recommend keeping this fix minimal. However, visual shapes should also be allowed at the same time.

NeoZng:
thanks, agreed. I changed the finalized-model check to only block shapes carrying collision flags, so visual-only shapes are allowed as well as frame sites. I also kept the scope minimal and reduced the regression test to just the post-finalization MJCF site case plus a visual-only shape check.
Need your further checks @camevor .

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
NeoZng: sure

camevor (commented):
Thanks for updating the PR @NeoZng.
With the scope reduced and visual shapes enabled, this looks good to me.
Could you please fix the docs test, which is currently failing; inlining `_COLLISION_FLAGS` at its use site should fix it.

NeoZng:
sure, done. @camevor 
llm tends to use a private flag even if the flag only used once lol

camevor:
Thanks @NeoZng, looks like your last commit introduced a `pre-commit` failure (`ruff format`). Could you address that one, too?

NeoZng:
fix pre-commit and pytest on gpu, the ci use kit-less import so i ask codex to re-generate a usd example for test to omit need for isaacsim @camevor

camevor (approved):

AntoineRichard (approved):

AntoineRichard:
run-ci

PR #7395, opened 2026-08-27
[Docs] Document Newton coupled solvers

maxkra15:
# Description

Adds a top-level **Coupled Solvers** page to the Concepts documentation for Newton-based simulations.

The page:

- introduces shared-model partitioning and named solver-entry ownership;
- compares proxy and ADMM coupling, including when to choose each approach;
- shows the current Isaac Lab configuration patterns for both coupling methods;
- links to Newton's authoritative implementation documentation, the generated Isaac Lab coupling API, and the maintained Franka proxy example.

It also adds the page to the top-level Concepts navigation and links the existing VBD proxy guide back to the high-level comparison.

This change is intentionally documentation-only. It does not modify runtime behavior, public APIs, dependencies, or source packages.

## Type of change

- [x] Documentation update

## Release backport

## Screenshots

Not applicable; this change adds and cross-links a text-based concept page.

## Validation

- [x] Complete warnings-as-errors Sphinx build passed.
- [x] Relevant formatting, spelling, RST, and Git LFS checks passed for all three changed files.
- [x] Proxy and ADMM configuration examples instantiate against the current `develop` configuration classes.
- [x] Branch diff contains documentation files only.

The repository-wide changelog hook uses `origin/develop` as its local comparison ref. In this fork that ref is older than `upstream/develop`, so the local all-files run reports pre-existing upstream fragments; this docs-only diff touches no source package and requires no changelog fragment.

## Checklist

- [x] Code tests are not applicable because this change is documentation-only
- [x] A package changelog fragment is not applicable because no source package is touched

### DIFF
--- a/docs/source/concepts/coupled_solvers.rst
+++ b/docs/source/concepts/coupled_solvers.rst
@@ -1,0 +1,186 @@
+.. Copyright (c) 2026, The Isaac Lab Project Developers (https://github.com/isaac-sim/IsaacLab/blob/main/CONTRIBUTORS.md).
+.. All rights reserved.
+..
+.. SPDX-License-Identifier: BSD-3-Clause
+
+.. _newton-coupled-solvers:
+
+Coupled Solvers
+===============
+
+Newton can partition one model between multiple solvers and exchange state and
+forces between them during each simulation step. This lets a task combine
+solver families that target different physics, such as an MJWarp rigid robot
+interacting with VBD cloth or an MPM material.
+
+Isaac Lab exposes this experimental framework through
+:mod:`isaaclab_contrib.coupling`. The adapter turns configuration selectors
+into named Newton solver entries, constructs the selected coupling algorithm,
+and integrates it with the normal :class:`~isaaclab_newton.physics.NewtonCfg`
+lifecycle. For the shared-model architecture, iteration algorithms, supported
+constraint rows, and solver-specific implementation details, see Newton's
+`Coupled Solvers concept page
+<https://newton-physics.github.io/newton/stable/concepts/coupling.html>`_.
+
+
+The coupling model
+------------------
+
+A coupled simulation starts from one Newton model. Instead of giving the whole
+model to one solver, Isaac Lab partitions it into named entries. Each entry
+selects a solver and owns a disjoint part of the model:
+
+.. code-block:: text
+
+   Shared Newton model
+   ├── "rigid" entry → MJWarp owns robot bodies, joints, and shapes
+   ├── "soft" entry  → VBD owns deformable particles and static shapes
+   └── interface     → proxy or ADMM coupling connects the entries
+
+Each solver receives a view of the shared model and advances only the elements
+owned by its entry. An element can belong to at most one entry; unassigned
+elements remain outside the nested solvers. Keep an articulation in one entry
+unless a supported ADMM constraint intentionally connects bodies owned by
+different entries.
+
+Isaac Lab resolves ownership selectors, constructs the Newton entry views, and
+runs the coupled solver through the normal Newton backend lifecycle. Newton
+owns the coupling algorithms and the exchange of poses, forces, and constraint
+information between entries.
+
+
+Choose proxy or ADMM coupling
+-----------------------------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 40 40
+
+   * - Approach
+     - How it works
+     - When to use it
+   * - Proxy
+     - A source-owned body or particle appears as a virtual endpoint in a
+       destination solver. The destination returns feedback on a later pass or
+       iteration.
+     - Use when the interaction is naturally directional, such as a rigid
+       collider inside a deformable or particle solve. Proxy coupling can
+       reuse the destination solver's contact path and is the established path
+       for Isaac Lab's coupled MJWarp--VBD and rigid--MPM tasks.
+   * - ADMM
+     - The coupler creates interface constraints between entries, iterates the
+       sub-solvers, and applies equal and opposite interface forces.
+     - Use when the interface should be symmetric, especially for supported
+       cross-entry joints, body--particle attachments, or frictional contacts.
+       ADMM has more tuning parameters and supports a narrower set of
+       constraint rows.
+
+Proxy coupling is usually the simpler starting point for collider-style
+rigid--deformable interaction. Use ``mode="lagged"`` first; the
+``"staggered"`` mode uses a newer source state but is more sensitive to the
+timestep and ordering. Increase coupling iterations only after each entry is
+stable on its own.
+
+ADMM is a better fit when assigning a source and destination would make the
+physical interface artificially one-way. Its fixed iteration count and
+penalty, proximal, and stabilization parameters are part of the coupled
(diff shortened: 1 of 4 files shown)


docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+
+Newton can partition one model between multiple solvers and exchange state and
+forces between them during each simulation step. This lets a task combine
+solver families that target different physics, such as an MJWarp rigid robot
+interacting with VBD cloth or an MPM material.
+
+Isaac Lab exposes this experimental framework through
+:mod:`isaaclab_contrib.coupling`. The adapter turns configuration selectors
+into named Newton solver entries, constructs the selected coupling algorithm,
+and integrates it with the normal :class:`~isaaclab_newton.physics.NewtonCfg`
```
AntoineRichard:
```suggestion
and integrates it with  :class:`~isaaclab_newton.physics.NewtonCfg`
```

docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+unless a supported ADMM constraint intentionally connects bodies owned by
+different entries.
+
+Isaac Lab resolves ownership selectors, constructs the Newton entry views, and
+runs the coupled solver through the normal Newton backend lifecycle. Newton
+owns the coupling algorithms and the exchange of poses, forces, and constraint
+information between entries.
+
+
+Choose proxy or ADMM coupling
```
AntoineRichard:
```suggestion
Choose Proxy or ADMM coupling
```

docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+``"staggered"`` mode uses a newer source state but is more sensitive to the
+timestep and ordering. Increase coupling iterations only after each entry is
+stable on its own.
+
+ADMM is a better fit when assigning a source and destination would make the
+physical interface artificially one-way. Its fixed iteration count and
+penalty, proximal, and stabilization parameters are part of the coupled
+constraint solve, so tune them together with the timestep and the participating
+solvers. Newton's concept page is the source of truth for the currently
+supported joints, contacts, and limitations.
```
AntoineRichard, line 94: Could we also some discussion around the performance? Maybe the accuracy as well?

AntoineRichard (changes requested):
LGTM! Couple of small comments.

docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+
+
+Configure a coupled solver
+--------------------------
+
+In Isaac Lab, :class:`~isaaclab_contrib.coupling.CouplerEntryCfg` defines each
```
mmichelis, line 108: Small thing I noticed while building this locally: the three :class: refs to the coupling config classes don't actually resolve, so they render as plain grey text instead of links. 

The reasoning according to my agent:

The reason is that docs/source/api/lab_contrib/isaaclab_contrib.coupling.rst documents them with the full submodule path (.. autoclass:: isaaclab_contrib.coupling.coupler_cfg.CouplerEntryCfg) and no .. currentmodule::, so the anchors Sphinx registers are isaaclab_contrib.coupling.coupler_cfg.CouplerEntryCfg. The shortened path in the prose doesn't match any of them. It fails silently because nitpicky is still commented out in docs/conf.py, which is why the warnings-as-errors build stays green. I confirmed it in the generated HTML: NewtonCfg comes out as a real <a href=...> link, while the three above come out as bare <code>.

To be clear this isn't something you introduced, using-vbd-solver.rst has the exact same dead refs today, so feel free to leave it. But since this page leans on those three classes pretty heavily it'd be nice for them to be clickable. Cheapest fix is adding a .. currentmodule:: isaaclab_contrib.coupling to the API page and dropping the module prefix from the autoclass lines, which would fix the existing VBD page refs at the same time. Writing out the full coupler_cfg path here would also work if you'd rather keep the change contained to this PR.

docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+
+   Shared Newton model
+   ├── "rigid" entry → MJWarp owns robot bodies, joints, and shapes
+   ├── "soft" entry  → VBD owns deformable particles and static shapes
+   └── interface     → proxy or ADMM coupling connects the entries
+
+Each solver receives a view of the shared model and advances only the elements
+owned by its entry. An element can belong to at most one entry; unassigned
+elements remain outside the nested solvers. Keep an articulation in one entry
+unless a supported ADMM constraint intentionally connects bodies owned by
```
mmichelis: this admm to split up an articulation feels very specific, I don't think we have an application for this in isaaclab yet?

docs/source/concepts/coupled_solvers.rst
```diff
@@ -0,0 +1,186 @@
...
+.. code-block:: python
+
+   from isaaclab_contrib.coupling import CouplerAdmmCfg
+
+   physics = NewtonCfg(
+       solver_cfg=CouplerAdmmCfg(
+           entries=entries,
+           contact_pairs=[("rigid", "soft")],
+           iterations=5,
+           rho=1.0,
```
mmichelis, line 172: would it make sense to also have a section explaining all the possible parameters (rho, baumgarte, mass scale, ...)

mmichelis (commented):
Thanks for the great docs! General question: Since coupling is still in contrib, should we more advertise this as an experimental feature, or directly as a core concept?

mmichelis (approved):

docs/source/concepts/coupled_solvers.rst
maxkra15: Fixed this at the API source by adding `currentmodule` and documenting the public exports under their short names, also verified the generated HTML links

docs/source/concepts/coupled_solvers.rst
maxkra15: Added a brief Tune Coupling section that explains the high-level effects of entry substeps, the main proxy controls, and `iterations`, `rho`, `gamma`, and `baumgarte` for ADMM. We should extend this later with a more detailed guide

docs/source/concepts/coupled_solvers.rst
maxkra15: removed

docs/source/concepts/coupled_solvers.rst
maxkra15: Added a compact performance and accuracy discussion covering proxy overhead and ordering sensitivity, ADMM per-iteration solver cost, and the runtime-versus-convergence tradeoff.

docs/source/concepts/coupled_solvers.rst
maxkra15: Updated as suggested

docs/source/concepts/coupled_solvers.rst
maxkra15: Updated the heading capitalization.

kellyguo11:
run-ci

PR #7397, opened 2026-08-27
[Fix] Pin Newton to the commit carrying newton#4017 and stop the installer dangling its prebundle

hujc7:
# Description

Reading a MuJoCo actuator attribute through an articulation view — `mujoco.actuator_trntype`, which
the fixed-tendon adapter needs — fails on `release-1.5` with *"has custom frequency
'mujoco:actuator' which is not supported by ArticulationView"*.
[newton#4017](https://github.com/newton-physics/newton/pull/4017) makes `ArticulationView` generic
over custom frequencies. It merged to `main` and was not backported, so the pin is the merge commit;
`main` carries one further commit, so this takes 4017 and nothing else.

The bump also exposes a latent installer bug. `_ensure_newton` uninstalled Newton before
reinstalling it, but Isaac Sim's `isaacsim.pip.newton` prebundle symlinks into the installed tree
(`pip_prebundle/newton -> …/site-packages/newton`). Removing the distribution left every link
dangling and the install failed its own prebundle check (nvbugs 6343978) with 200 dangling symlinks.
It only ever fired when the pin changed, because the freeze check returns early while the pinned
commit is already installed — so the bug has been latent since the guard was added. The pinned build
is now force-reinstalled over the existing tree.

> [!NOTE]
> `uv.lock` is edited in place rather than regenerated. Newton's own dependencies are identical
> across the two revisions (`warp-lang>=1.16.0`), so nothing downstream re-resolves, and a
> regenerate rewrites 1087 unrelated lines: uv 0.12.6 simplifies away a marker equal to the union of
> the three declared `[tool.uv] environments`, which the committed lock still spells out. Verified
> with `uv sync --frozen`, which accepts the lock without re-resolving.

Fixes # (issue)

## Type of change

## Release backport

## Validation

The pin lives in four places, cross-checked by `test_wheel_builder_metadata.py` and
`test_uv_run_pyproject.py`: `[tool.isaaclab.versions]`, `[tool.uv] override-dependencies`,
`tools/wheel_builder/uv-overrides.txt` and `source/isaaclab/test/install_ci/uv_pip/uv-overrides.txt`.

* `uv sync --frozen` resolves newton 1.6.0.dev0 without re-resolving the lock.
* Spawning the Shadow Hand on `newton_mjwarp` reports 4 fixed tendons in the expected order, where
  `release-1.5` raised at articulation start-up.
* `source/isaaclab/test/cli/` — 246 passed.

`test_installs_pinned_git_build_when_absent` now asserts the installer does **not** uninstall and
does pass `--force-reinstall`, so the dangling-symlink cause is covered rather than the old
uninstall-first contract.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -211,7 +211,10 @@ torchaudio = "2.11.0"
 ovphysx = "0.5.10"
 ovrtx = "0.4.1.364340"
 ovstage = "0.1.1.355824"
-newton = "release-1.5"
+# Pinned past release-1.5 for newton#4017 (ArticulationView generic over custom frequencies),
+# which the MuJoCo tendon adapter needs to read mujoco.actuator_trntype. Merged to main and
+# not backported, so this is the merge commit rather than a branch.
+newton = "4055a1594d91ae8c1a8688b0932cd10f75d30603"
 warp = "1.16.0"
 
 [tool.ruff]
@@ -394,7 +397,7 @@ override-dependencies = [
     "numpy>=2",
     "mujoco~=3.11.0",
     "mujoco-warp~=3.11.0",
-    "newton[sim] @ git+https://github.com/newton-physics/newton.git@release-1.5",
+    "newton[sim] @ git+https://github.com/newton-physics/newton.git@4055a1594d91ae8c1a8688b0932cd10f75d30603",
     # Force the Newton-matched schemas over isaacsim's ==0.2.0 pin.
     "newton-usd-schemas>=0.4.1",
     "torch==2.11.0",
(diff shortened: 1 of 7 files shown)


hujc7:
Superseded by #7445, which rebases this bump onto #7405.

Two things changed after verification:

- The installer change here is wrong. `pip install --force-reinstall` was measured in the Isaac Sim image to behave identically to the `pip uninstall` it replaced — both delete `pip_prebundle/newton/` and leave 792 dangling symlinks, and the force-reinstall log itself prints `Uninstalling newton`. #7445 drops the `install.py` and `test_install.py` changes and the corresponding changelog section.
- On top of #7405 the bump is much smaller: the pin lives in three mirrored sites plus `uv.lock`, and regenerating the lock is a 9-line diff instead of 1087. No installer change is needed, because #7405's `_ensure_newton` already parses both the `@<sha>` and `==<version>` pin forms.

The pin itself is unchanged: `4055a1594d91ae8c1a8688b0932cd10f75d30603`, the merge commit of newton#4017.

PR #7402, opened 2026-08-28
Report resolved camera resolutions in benchmark KPIs

yts-nv:
# Description

OmniPerf camera benchmark jobs currently record the requested resolution in runner metadata, but the Isaac Lab KPI payload does not record the resolved camera configuration. This makes it impossible for downstream dashboards to distinguish a verified 64x64 run from a run that retained the task default.

This change:

- traverses the concrete environment config after Hydra and command-line overrides are applied;
- records each CameraCfg and RayCasterCameraCfg path with its resolved image width and height;
- emits the data under benchmark_info.camera_resolutions for runtime, startup, training, and play benchmarks across RSL-RL, RL-Games, SKRL, and SB3;
- omits the field for tasks without a configured camera resolution.

Example KPI metadata:

```json
{
  "camera_resolutions": {
    "env.scene.tiled_camera": {
      "width": 64,
      "height": 64
    }
  }
}
```

## Type of change

## Release backport

- [ ] Backport this pull request to the active release branch after it merges into develop

## Screenshots

Not applicable.

## Validation

- 21 focused benchmark capture and KPI serialization tests passed.
- Real Hydra composition verified Isaac-Cartpole-Camera reports env.scene.tiled_camera at 64x64.
- Real Hydra composition verified Isaac-Cartpole-Camera-Direct reports env.tiled_camera at 64x64.
- The complete uv run isaaclab -f gate passed, including Ruff, formatting, codespell, changelog, and Git LFS checks.

## Checklist

- [ ] I have made corresponding changes to the documentation; no public documentation page requires an update for this KPI metadata addition
- [x] I have added tests that prove the feature works
- [x] I have added a changelog fragment under source/isaaclab/changelog.d/
- [ ] I have added my name to CONTRIBUTORS.md or my name already exists there


AntoineRichard (changes requested):
AI-generated review (Codex).

Two camera-discovery edge cases should be addressed before merge:

1. The global visited-object check runs before camera recording. If the same camera config object is referenced from two config paths, only the first path is emitted. The current test codifies this by assigning shared_camera to tiled_camera and duplicate but expecting only tiled_camera. Because the payload is path-based and the PR says it records each camera config path, camera detection should happen before container deduplication so every path is preserved.

2. Camera identity is matched only through MRO class names CameraCfg and RayCasterCameraCfg. This covers their current subclasses, but an unrelated task class with the same simple name and integer width/height fields is falsely reported, while a future independent camera-config family would be missed. Please use concrete types through a function-local import, or at least match module plus qualified class name.

For clarity: no-camera configurations are handled safely—the helper returns no metadata, so benchmark_info.camera_resolutions is omitted. Two distinct camera config objects are both reported and sorted by path; only aliased references lose the second path.

Suggested focused tests: no camera, two distinct cameras, one camera object referenced at two paths, and an unrelated same-named CameraCfg class.

yts-nv:
Thanks @AntoineRichard , updated for the feedback.

AntoineRichard (approved):
Thanks @yts-nv! Should we add this to the schema?

yts-nv:
> Thanks @yts-nv! Should we add this to the schema?

Sure! added.

fatimaanes (approved):

PR #7404, opened 2026-08-28
Fix: script-default visualizers not cleared in headless mode (--headless and HEADLESS=1)

### LINKED ISSUE #7403: Script-default visualizers incorrectly flagged as explicit, causing crash with --headless and HEADLESS=1
**Describe the bug**

When a task script registers a default visualizer via `parser.set_defaults(visualizer=["kit"])`,
two related bugs in `AppLauncher` cause a `RuntimeError` in headless mode:

**Bug 1 — `_resolve_visualizer_settings` promotes defaults to explicit intent**

`_cli_visualizer_explicit` is set to `True` even when the user never typed `--viz` or
`--visualizer` on the command line. The value came from `set_defaults()`, not from the user.
This means `SimulationContext._resolve_visualizer_cfgs` sees an "explicitly requested"
visualizer and raises:

    RuntimeError: Explicitly requested visualizer(s) ['kit'] could not be configured.

**Bug 2 — Neither `--headless` nor `HEADLESS=1` clear `launcher_args["visualizer"]`**

`_resolve_headless_settings` set `_cli_visualizer_disable_all = True` for the `--headless`
path but never cleared `launcher_args["visualizer"]`, leaving `["kit"]` in the args.
The `HEADLESS=1` env-var path had no clearing logic at all.

**Steps to reproduce**

Both of the following crash on any task that calls `parser.set_defaults(visualizer=["kit"])`:

    HEADLESS=1 python scripts/environments/zero_agent.py --task=<task> --num_envs=1

    python scripts/environments/zero_agent.py --task=<task> --num_envs=1 --headless

**Expected behavior**

Both headless invocation styles should suppress visualizers cleanly and start without error.

I have a fix ready — will link PR once this issue is open.

ChaitanyaParate:
## Summary

Fixes a runtime crash in `AppLauncher` when a task script registers a default visualizer via
`parser.set_defaults(visualizer=["kit"])` and the process is launched in headless mode.

Fixes #7403

### Root cause

Two bugs compounded:
1. `_resolve_visualizer_settings` incorrectly promoted `set_defaults()` values to explicit user
   intent (`_cli_visualizer_explicit = True`), even though the user never typed `--viz`.
2. `_resolve_headless_settings` never cleared `launcher_args["visualizer"]` for either the
   `--headless` CLI path or the `HEADLESS=1` env-var path.

### Fix

- Removed the lines that unconditionally set `_cli_visualizer_explicit = True` from a
  `set_defaults()` value in `_resolve_visualizer_settings`.
- Added clearing logic in `_resolve_headless_settings` for both headless paths:
  sets `_cli_visualizer_disable_all = True`, clears `_cli_visualizer_types` and
  `launcher_args["visualizer"]`.

### Testing

- Added assertions to `test_headless_resolution_with_script_default_visualizer`
  covering both the `HEADLESS=1` and `--headless` cases.
- Verified manually with a downstream task environment.

## Checklist

- [x] Pre-commit hooks run and passed (`./isaaclab.sh -f`)
- [x] Tests added / updated
- [x] `CHANGELOG.rst` updated (v6.1.18)
- [x] `extension.toml` version updated
- [x] `CONTRIBUTORS.md` updated
- [x] No new warnings introduced


AntoineRichard (commented):
## AI review (Codex)

@matthewtrepte, could you please review the visualizer-intent semantics here, particularly the programmatic `AppLauncher` contract and the livestream interaction?

The reported bug is real, and clearing `launcher_args["visualizer"]` is necessary with the current plumbing: `_set_visualizer_settings()` publishes the visualizer type list from `launcher_args`, while publishing `explicit` and `disable_all` from the normalized instance fields. Clearing only `_cli_visualizer_types` therefore leaves a stale `kit` request in Carb settings, which `SimulationContext` can later validate as an unresolved explicit request.

I recommend changes before merge because the current fix introduces two high-confidence regressions:

### Important findings

1. **`HEADLESS=1` / deprecated `--headless` can disable the Kit visualizer required by livestreaming.**

   The new disable-all branches run before the later livestream handling and do not exempt `self._livestream in {1, 2}`. A configuration such as `LIVESTREAM=1 HEADLESS=1`, or `--livestream 1 --headless`, can therefore publish `visualizer_disable_all=True` after the task launcher has injected the Kit visualizer required for the stream. `SimulationContext` then resolves no visualizers, so the Kit update/render pump needed by the remote stream may be lost.

2. **Removing presence-based inference globally breaks programmatic visualizer requests.**

   `AppLauncher` supports dictionaries and keyword arguments, so calls such as `AppLauncher(visualizer=["kit"])` and `AppLauncher({"visualizer": ["kit"]})` do not pass through `ExplicitAction` and have no `visualizer_explicit` marker. After this change they are treated as non-explicit; headless/config fallback can win and `SimulationContext` can ignore the supplied visualizer list. `visualizer=["none"]` likewise no longer sets disable-all and may fail to suppress configured visualizers.

### Suggested direction

- Make the normalized instance state the single source of truth when publishing settings: pass `self._cli_visualizer_types` to `sync_visualizer_cli_settings_to_carb()` instead of relying on the potentially stale `launcher_args["visualizer"]`. That removes the need for each later resolver to keep both representations synchronized manually.
- Preserve provenance instead of dropping inference for every caller. One approach is to ensure parser-created namespaces always contain `visualizer_explicit=False`, including the `parser.set_defaults(...)` case, then infer explicit intent only when the marker is absent for direct dict/kwargs callers.
- Preserve or explicitly replace the Kit update pump while livestreaming rather than applying unconditional disable-all semantics.
- Add regression coverage for:
  - `HEADLESS=1` plus livestream;
  - deprecated `--headless` plus livestream;
  - direct dict and kwargs inputs for `visualizer=["kit"]` and `visualizer=["none"]`;
  - an actual `argparse` path using `parser.set_defaults(visualizer=["kit"])`, rather than only a synthetic dictionary.

The focused test added here covers the reported state transition, but it currently encodes the assumption that a missing `visualizer_explicit` key always means a parser default. That is not true for the supported direct dict/kwargs paths.

_This review was generated by an AI coding agent (OpenAI Codex) and should be validated by a human maintainer._

AntoineRichard (commented):
## AI review update — please retarget to `develop`

@ChaitanyaParate, after comparing this patch with the latest `develop`, I recommend retargeting the PR to `develop` and revising the fix against that branch. A substantial part of the current diff and of my earlier review is specific to `release/3.0.0-beta2` and no longer applies directly.

This note supersedes the base-dependent portions of my earlier AI review:

- The deprecated `--headless` CLI path has been removed on `develop`, so that part of the patch should not be ported.
- `develop` intentionally supports `HEADLESS=1` together with the Kit visualizer after #7180 (`Declare the USDRT scene delegate in the headless app`). Unconditionally clearing visualizers under `HEADLESS=1` would conflict with that supported behavior and could also remove the Kit update pump required by livestreaming.
- `develop` still contains the presence-based fallback that protects direct dict/kwargs calls such as `AppLauncher(visualizer=["kit"])`. Removing it globally would still introduce the programmatic-input regression described in my earlier review.

The narrower issue that remains on `develop` is provenance: a script default such as `parser.set_defaults(visualizer=["kit"])` is still promoted to explicit user intent because the parsed namespace does not distinguish that default from a direct programmatic visualizer argument.

### Suggested `develop`-scoped fix

1. Ensure parser-generated namespaces carry `visualizer_explicit=False` when the user did not pass `--viz`.
2. Distinguish an explicit false marker from a missing marker: parser defaults remain non-explicit, while direct dict/kwargs visualizer values retain their existing explicit semantics.
3. Do not clear visualizers merely because `HEADLESS=1`; headless Kit is now supported on `develop`.
4. Consider publishing `self._cli_visualizer_types` as the normalized source of truth instead of separately synchronizing `launcher_args["visualizer"]`.
5. Test a real `argparse` `set_defaults()` path plus direct dict/kwargs inputs. Preserve coverage for `HEADLESS=1` + Kit and livestream.

Because retargeting will materially change the diff, the revised patch should receive a fresh review against `develop`.

_This update was generated by an AI coding agent (OpenAI Codex) after comparing PR #7404 with the latest remote `develop`; human maintainer validation is required._

ooctipus:
looks like this pr is created by mistake closing it now, 
feel free to reopen when cleaned up

PR #7405, opened 2026-08-28
[Docker] Install the Docker images from uv.lock

hujc7:
## Summary

Docker images stop mutating Isaac Sim's shipped Python environment: the previous build
replaced **41** of its packages, this one replaces **0**. All three images install from
`uv.lock`, so builds are reproducible and `[tool.uv].override-dependencies` apply.

# Description

The lock is the only path that applies `[tool.uv].override-dependencies`, which pip cannot
read. The change is contained to the Docker build; `isaaclab.sh --install` is unchanged.

- **Isaac Sim image** — `uv sync` resolves into a venv at `/opt/isaaclab-venv`, built on Kit's
  own interpreter. Isaac Sim is untouched, still reached through the existing `_isaac_sim`
  symlink, and `isaaclab.sh` sources `setup_python_env.sh` so `import isaacsim` keeps working.
- **Kit-less image** — `uv sync` replaces the pip-style install, retiring the duplicated
  `tools/wheel_builder/uv-overrides.txt`.
- **cuRobo image** — same migration. Because Isaac Sim's site-packages is no longer touched,
  the image stops deleting the prebundled torch, re-bootstrapping `pip` with `get-pip.py`, and
  uninstalling `quadprog`. cuRobo is still built from its pinned commit.
- **No new extras** — the images use the ones that already exist. `teleop` no longer bundles
  the `isaacsim` wheel, so environments that already provide Kit do not install a second,
  version-mismatched copy; installs that need it use `--extra teleop,isaacsim`, and the docs
  say so wherever XR teleop appears (agreed with the teleop owner). `test` is now the test
  suite alone and `dev` adds the documentation toolchain, so the images can install `test`
  without shipping Sphinx and its GPL-3.0-or-later `docutils`, which the Docker licence gate
  rejects.

- **Image invariant test** — the images install with `uv sync`, which never runs
  `command_install`'s dangling-symlink guard, so #6329's protection would have been lost. A
  post-build step asserts no prebundled package lost its `__init__.py`, bound to the digest
  just built. It has already caught a real regression on a dependent branch.
- **Removed** the `quadprog` uninstall (absent from the lock), the `toml` install
  (`install_deps.py` now uses stdlib `tomllib`), and the `imageio-ffmpeg` uninstall
  (excluded during the sync instead).

Newton resolves to 1.5.0 from the lock's pinned commit where the pip path floated to
`release-1.5` HEAD (1.5.1); regenerating `uv.lock` is the lever if a newer commit is needed.

`uv.lock` is regenerated with uv 0.9.25, the version the images pin. Newer uv renormalises
environment markers and rewrites ~2200 lines with no semantic change.

## Type of change

## Release backport

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works — a post-build image-invariant test, plus building all three images and diffing their package sets against the previous build


hujc7:
@StafaH when you got time.

hujc7:
run-ci

AntoineRichard:
**[AI-generated review]**

I am concerned that this Docker-focused change expands into the behavior of the general `isaaclab.sh --install` command. The Dockerfiles already control `VIRTUAL_ENV` and `UV_PROJECT_ENVIRONMENT`, so they can run the lock-based `uv sync` directly with the Docker-specific extras. I would expect the generic install path for Conda, arbitrary virtualenvs, and other non-Docker workflows to remain unchanged.

Could the lock-based installation be contained to the Docker build, retaining only the supporting `pyproject.toml`/`uv.lock` metadata changes that the images need? If changing the public install CLI is an intentional separate goal, it needs broader environment, working-directory, and selector-parity handling and tests. The current x86 installation CI failures demonstrate that this larger blast radius is already affecting non-Docker workflows.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,232 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+
+
+def _lock_install_available(pip_cmd: list[str]) -> bool:
+    """Whether the environment supports installing from ``uv.lock``.
+
+    ``uv sync`` manages a virtual environment, so it cannot target Isaac Sim's bundled
+    interpreter or a bare system Python; those keep the pip path. ``get_pip_command``
+    already resolves to uv only for a virtualenv target, so it is the gate here.
+    """
+    return pip_cmd[0] == "uv" and (ISAACLAB_ROOT / "uv.lock").is_file()
```
AntoineRichard: **[AI-generated review comment]**

This gate is broader than the Docker use case. `get_pip_command()` also returns uv for Conda and arbitrary virtualenvs when uv is on `PATH`, so they enter the new lock path even though the changelog says other environments retain pip. Current x86 CI shows the consequence: uv warns that the selected environment is ignored, and the Conda/venv tests subsequently cannot import `isaaclab` or `warp`.

Given the Docker-focused goal, could this path require an explicitly configured Docker/project environment, leaving the existing pip dispatch unchanged elsewhere?

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,232 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+
+
+def _install_from_lock(extras: list[str]) -> None:
+    """Install the project from ``uv.lock``.
+
+    ``--frozen`` keeps uv from re-resolving and silently defeating the lock; ``--inexact``
+    leaves packages the lock does not name (a seeded pip, a source-built Isaac Sim) in place
+    instead of removing them.
+    """
+    command = ["uv", "sync", "--frozen", "--inexact"]
```
AntoineRichard: **[AI-generated review comment]**

This command relies on the caller already being in `ISAACLAB_ROOT`. That is true in the Docker `RUN cd ...` step, but `isaaclab.sh` itself does not change directories. Calling the absolute launcher from another directory now fails because uv cannot find `pyproject.toml`.

Keeping `uv sync` inside the Dockerfile avoids changing that public CLI contract. If the generic lock path is retained, pass an explicit `--project`/`--directory` and explicitly target the intended environment.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,232 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+) -> list[str] | None:
+    """Resolve the lock extras to install, or None when the pip path must be used.
+
+    The lock is the only path that applies ``[tool.uv].override-dependencies``, so the torch
+    index pins and the Newton and Pink IK corrections :func:`_install_with_pip` performs
+    afterwards are already expressed in it.
+    """
+    if not _lock_install_available(get_pip_command(python_exe)):
+        return None
+    extras = _lock_install_extras([name for name, _ in extra_features], optional_submodules, install_isaacsim)
```
AntoineRichard: **[AI-generated review comment]**

Dropping the selectors here changes existing CLI semantics in uv environments: `contrib[rlinf]` resolves to no extra, while `rl[rsl-rl]`, `visualizer[rerun]`, and `ov[ovrtx]` expand to every mapped sibling. This generalized selector translation is another source of complexity that is not needed if lock-based syncing remains Docker-specific. If the generic path stays, the mapper needs selector parity with the pip path and should fall back when an exact mapping is unavailable.

hujc7:
Agreed — done in `d338e17a39a`. `install.py` is reverted to `develop`; the lock install is now contained to the Docker build. `Dockerfile.base` syncs directly via a new `kit-image` extra; the only non-`docker/` changes are the `pyproject.toml`/`uv.lock` metadata the images need and `install_deps.py` moving to stdlib `tomllib`.

On the blast radius: `utils.py:307` counts `CONDA_PREFIX` as a virtualenv, so `get_pip_command()` returned uv for Conda and the gate let it through — my docstring claimed otherwise. That caused the x86 Conda failures.

The CLI lock path is worth doing separately, with the environment/cwd/selector handling and `install_ci` coverage you list. Rebuilt image: 244 packages vs 205, no `isaacsim*` in the venv, and 0 packages replaced in Isaac Sim's own site-packages where the previous build replaced 41.

source/isaaclab/isaaclab/cli/commands/install.py
hujc7: Confirmed. `utils.py:307` includes `CONDA_PREFIX` in `in_venv`, so the gate caught Conda; the docstring saying otherwise was wrong. Reverted.

source/isaaclab/isaaclab/cli/commands/install.py
hujc7: Correct — the `RUN cd ${ISAACLAB_PATH}` was the tell. Reverted; the Dockerfile runs `uv sync` itself. `--project`/`--python` if the CLI path returns.

source/isaaclab/isaaclab/cli/commands/install.py
hujc7: Confirmed — the selector was dropped outright. Reverted; parity is a follow-up item.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,227 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+    """Resolve requested install tokens to lock extras, or None when unmappable."""
+    extras: dict[str, None] = {}
+    for name in requested_features:
+        if name not in LOCK_EXTRAS_BY_FEATURE:
+            return None
+        for extra in LOCK_EXTRAS_BY_FEATURE[name]:
+            extras.setdefault(extra)
+    for name in optional_submodules:
+        if name not in LOCK_EXTRAS_BY_FEATURE:
+            continue
```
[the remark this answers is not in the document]
hujc7: Fixed in `77136b4d201`, now moot — `install.py` is out of this PR.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,227 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+) -> list[str] | None:
+    """Resolve the lock extras to install, or None when the pip path must be used.
+
+    The lock is the only path that applies ``[tool.uv].override-dependencies``, so the torch
+    index pins and the Newton and Pink IK corrections :func:`_install_with_pip` performs
+    afterwards are already expressed in it.
+    """
+    if not _lock_install_available(get_pip_command(python_exe)):
+        return None
+    extras = _lock_install_extras([name for name, _ in extra_features], optional_submodules, install_isaacsim)
```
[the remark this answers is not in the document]
hujc7: Correct; moot — `install.py` is out of this PR. Selector parity tracked for the follow-up.

source/isaaclab/isaaclab/cli/commands/install.py
```diff
@@ -521,6 +521,227 @@ def _install_centralized_dependencies(pip_cmd: list[str], optional_submodules: l
...
+
+
+def _install_from_lock(extras: list[str]) -> None:
+    """Install the project from ``uv.lock``.
+
+    ``--frozen`` keeps uv from re-resolving and silently defeating the lock; ``--inexact``
+    leaves packages the lock does not name (a seeded pip, a source-built Isaac Sim) in place
+    instead of removing them.
+    """
+    command = ["uv", "sync", "--frozen", "--inexact"]
```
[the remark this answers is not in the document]
hujc7: Correct; moot — `install.py` is out of this PR, the Dockerfile runs `uv sync` directly.

source/isaaclab/isaaclab/cli/commands/install.py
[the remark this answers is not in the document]
hujc7: Correct; moot — `install.py` is out of this PR.

hujc7:
@greptileai review

hujc7:
run-ci

kellyguo11:
@myurasov-nv would be good to get your review on this as well

pyproject.toml
```diff
@@ -119,13 +119,18 @@ importers = [
     "tinyobjloader==2.0.0rc13",
```
StafaH, line 119: There's alot of undesireable changes to the pyproject here, and they are causing a huge change in the uv.lock

Instead of modifying the extras (which we have tuned very carefully and are used in the docs in alot of places), for the docker build we can avoid using the extra. An extra is the same as installing each dependency individually together, so we can just inline the installation of which dependencies we need. You could even programmatically grab the dependency list from this pyproject, and remove/modify the list as needed for certain installation that can't mix.

StafaH (changes requested):

pyproject.toml
hujc7: It was a uv version problem that generates the huge uv lock diff. An extra extra does help keep the rest of code clean. The alternative is weird and could deviate. Defining all in a single place is better I think. let me know if it could work.

docker/Dockerfile.base
```diff
@@ -113,41 +123,79 @@ RUN touch /bin/nvidia-smi && \
...
+
+# Fix the line endings for the shell scripts (Windows git may add \r), then restore the
+# launcher's mode; one layer because both walk the tree the COPY above just wrote. ``sed -i``
+# rewrites the file under the build umask, so an explicit 755 keeps it runnable by a
+# container started with a uid other than the image's own.
+RUN find ${ISAACLAB_PATH} -type f -name "*.sh" -exec sed -i 's/\r$//' {} + \
+ && chmod 755 ${ISAACLAB_PATH}/isaaclab.sh
+
+# Install apt dependencies for extensions that declare them in their extension.toml,
+RUN ${ISAACLAB_PATH}/isaaclab.sh -p ${ISAACLAB_PATH}/tools/install_deps.py apt ${ISAACLAB_PATH}/source && \
```
StafaH, line 153: Here and elsewhere, can we use uv run? Do we need isaaclab.sh? What would it take to remove isaaclab.sh -p completely everywhere?

hujc7: Split this out as #7503, stacked on this PR, so the lock migration stays reviewable on its own.

`uv run` works but is not the cleanest form here. Putting the environment on `PATH` makes the `isaaclab` console script directly callable, so the call sites become `isaaclab -p ...` — shorter than the current launcher path and with no wrapper. `uv run` needs `--frozen --no-sync --project` to behave the same: without them it re-resolves the project on every call and warns when invoked outside it (124 ms vs 110 ms per call).

That removes every `isaaclab.sh -p` invocation from the images, including the `python`/`python3` aliases. Two aliases drop out with it: `isaaclab` was shadowing the console script once it is on `PATH`, and `tensorboard` was an explicit path into the environment.

One ordering constraint: `install_deps.py apt` ran before the workspace install, where the console script cannot import its own package. Moving it below the workspace sync fixes that, and nothing between the two steps needs apt first.

`isaaclab.sh` still ships unchanged — this only changes how the images reach it.

StafaH (approved):

kellyguo11:
run-ci

ooctipus:
run-ci

ooctipus (approved):

AntoineRichard (approved):

PR #7410, opened 2026-08-28
Keep configclass usable as a decorator after submodule imports

AntoineRichard:
## Description

`from isaaclab.utils import configclass` returns the *submodule* instead of the decorator whenever any code has previously imported `isaaclab.utils.configclass` — which most of `isaaclab` itself does (`terrains.sub_terrain_cfg`, `controllers.*_cfg`, `renderers.renderer_cfg`, `utils.noise.noise_cfg`, ...). The user-facing symptom is an order-dependent `TypeError: 'module' object is not callable` on the most common import in Isaac Lab user code:

```python
from isaaclab.scene import InteractiveScene   # imports the cfg chain
from isaaclab.utils import configclass        # -> module, not function

@configclass                                   # TypeError: 'module' object is not callable
class MyCfg: ...
```

Root cause: the lazy-loader stub exports the function through `__getattr__`, but a submodule import makes the import system bind the module object as the `configclass` attribute on the package. `__getattr__` is only consulted when the attribute is absent, so the module permanently shadows the function (the classic module-name/symbol-name collision).

Fix: make the `configclass` submodule callable — its module class now forwards `__call__` to the decorator. Both import spellings work regardless of import order, nothing is renamed, and no call sites change. Normalizing internal imports alone would not fix it, since any *user* submodule import re-triggers the shadowing.

Adds a subprocess-based regression test (fresh interpreter, controlled import order) that fails on `develop` and passes here.

## Type of change

## Testing

- `test_configclass.py::test_configclass_import_survives_submodule_shadowing` — fails on `develop`, passes here
- `source/isaaclab/test/utils/test_configclass.py` — 45 passed

## Checklist

### DIFF
--- a/source/isaaclab/changelog.d/antoiner-configclass-shadowing.rst
+++ b/source/isaaclab/changelog.d/antoiner-configclass-shadowing.rst
@@ -1,0 +1,7 @@
+Fixed
+^^^^^
+
+* Fixed ``from isaaclab.utils import configclass`` returning the submodule instead of the decorator
+  after any import of ``isaaclab.utils.configclass`` (order-dependent
+  ``TypeError: 'module' object is not callable``). The submodule is now callable and forwards to
+  the decorator, so both import spellings work regardless of import order.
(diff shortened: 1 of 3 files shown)


ooctipus:
maybe `from isaaclab.utils.configclass import configclass` will just work

If we want to take this path, maybe also update all 
`from isaaclab.utils.configclass import configclass`  to 
`from isaaclab.utils import configclass`

PR #7411, opened 2026-08-28
Restructure and correct the troubleshooting and known issues references

AntoineRichard:
# Description

The `refs/troubleshooting.rst` and `refs/issues.rst` pages had drifted from the code. This reorganizes both by backend and fixes the claims that no longer hold. Every claim on these pages was checked against the current source or reproduced against a running simulator; the entries that could not be settled are called out below rather than silently kept.

## Structure

- `troubleshooting.rst` is grouped under **Installation**, **PhysX backends**, **Newton backends**, **Renderers and visualizers**, **Distributed training**, and **Debugging and diagnostics**, so backend-specific advice is no longer interleaved with generic advice.
- The seven `ModuleNotFoundError` headings are collapsed into one section with a lookup table. The raw error strings made the local table of contents unreadable, and five of them shared a single answer. The literal error strings stay in the body so docs search still finds them.
- Every `issues.rst` entry is grouped by backend and carries an **Affects:** line, since several are renderer-specific rather than physics-specific.

## Corrections

`troubleshooting.rst`:

- The PhysX GPU buffer section was wrong four ways at once: it referenced a nonexistent `SimulationConfig` class, passed a `sim_params` keyword that `SimulationContext` does not accept, pointed at `isaaclab.sim.PhysxCfg` instead of `isaaclab_physx.physics.PhysxCfg` (so both `:class:` xrefs were broken), and told the reader to raise the buffer "to 4096" when the default is `2**21`. Rewritten against the real config path, with a `PresetCfg` variant matching how tasks set it today.
- The `isaaclab_assets` entry told users to include an `assets` token in their install command. No such token exists — `isaaclab_assets` is in `CORE_ISAACLAB_SUBMODULES` and cannot be deselected. Same confusion applied to `isaaclab_physx`, `isaaclab_ov` and `isaaclab_tasks`.
- viser's URL example used port 8012; `ViserVisualizerCfg.port` defaults to 8080.
- The "install `imgui-bundle`" advice is obsolete — it is a base dependency, annotated in `pyproject.toml` as "in base, no extra required".
- Backend selection used a `--physics` flag; the correct form is the `physics=` Hydra override, with the quickstart as the canonical selector list.
- Removed the `No module named 'pip'` entry, whose own body said it no longer occurs, and refreshed the Isaac Sim 2022/2023-era log and crash-trace examples.

`issues.rst`:

- **Digit / closed-loop articulations** — the stated root cause is fixed upstream. Newton now deduplicates the link axis (`arti_link_ids = sorted(set(arti_link_ids))`), so `link_count` is no longer inflated. Separately, `DigitPhysicsCfg` declares only `isaacsim_physx`, so the recommended workaround ("use the `physx` preset") no longer describes a choice the user has. Rewritten.
- **Stale values after reset** — no longer true for articulation link poses. Reproduced by commanding a cartpole slider `+0.5 m` and reading `body_state_w` with no intervening step: it reflected exactly the commanded delta. The entry is narrowed to RTX sensors, where the behavior does still hold, rather than deleted.
- **Blank initial camera frames** — did not reproduce on Isaac Sim 6.0.1.0 using the maintained `run_usd_camera.py` scene (frame 0 already had content, all 20 frames non-blank). Since the stated mechanism is texture loading and that run had a warm asset cache, the entry is scoped to cold caches and texture-heavy scenes rather than removed.
- Dropped the Isaac Lab 1.2 and Isaac Sim 5.0 / 5.1 version references throughout.

## Entries deliberately left on trust

These could not be settled here and their guidance is unchanged: the `Ctrl+C` `omni.physx` message (nondeterministic), instanceable assets for markers, and the conda `GLIBCXX` error (no conda environment available; conda remains a documented install path). The URDF importer entry is also unchanged — a minimal repro on `URDF USD Converter v0.1.3` merged the fixed-joint child despite mass and inertia and emitted no `ReportError`, contradicting the entry's premise, but a single synthetic asset is not enough to rewrite it.

## Type of change

## Release backport

## Checklist

### DIFF
--- a/docs/source/refs/issues.rst
+++ b/docs/source/refs/issues.rst
@@ -5,81 +5,47 @@ Known Issues
 
     Please also refer to the `Omniverse Isaac Sim documentation`_ for known issues and workarounds.
 
-Stale values after resetting the environment
---------------------------------------------
+Each entry below names the backends it affects. An issue listed under one backend does not
+apply to the others unless it says so.
 
-When resetting the environment, some of the data fields of assets and sensors are not updated.
-These include the poses of links in a kinematic chain, the camera images, the contact sensor readings,
-and the lidar point clouds. This is a known issue which has to do with the way the PhysX and
-rendering engines work in Omniverse.
+.. contents::
+   :local:
+   :depth: 2
+
+
+PhysX backends
+--------------
+
+Sensor readings are stale immediately after a reset
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Affects:** ``physics=isaacsim_physx``, ``physics=ovphysx``, and any RTX-based renderer.
 
 Many physics engines do a simulation step as a two-level call: ``forward()`` and ``simulate()``,
-where the kinematic and dynamic states are updated, respectively. Unfortunately, PhysX has only a
-single ``step()`` call where the two operations are combined. Due to computations through GPU
-kernels, it is not so straightforward for them to split these operations. Thus, at the moment,
-it is not possible to set the root and/or joint states and do a forward call to update the
-kinematic states of links. This affects both initialization as well as episodic resets.
+where the kinematic and dynamic states are updated respectively. PhysX has only a single
+``step()`` call where the two operations are combined. Because of computations through GPU
+kernels, it is not straightforward to split them. As a result, writing a root or joint state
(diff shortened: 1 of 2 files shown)


kellyguo11 (approved):

ooctipus (approved):

ooctipus:
run-ci

Issue #7398, opened 2026-08-27
3.0.0b2: ~20% Humanoid-Direct throughput regression vs v2.3.2 stack (Shadow-Cube unaffected)


ViktorM:
Migrating from the Isaac Lab v2.3.2 stack to 3.0.0b2 costs ~20% of Humanoid-Direct training throughput at scale, while Shadow-Cube gets slightly faster. All numbers below are measured with the two stacks equalized down to the RL code: same torch 2.11.0 with identical nvidia-* libs on both sides, the same rl_games code forced onto both, identical task configs (dt, decimation, PhysX solver and GPU buffer settings verified field by field; zero PhysX buffer overflows in any run), exclusive machine, back-to-back runs.

Setup: RTX PRO 6000 Blackwell Max-Q (driver 595.84), Threadripper 9970X, Ubuntu 24.04. Old = Isaac Lab v2.3.2 + Isaac Sim 5.1.0. New = isaaclab 3.0.0b2.post1 + Isaac Sim 6.0.1. rl_games PPO, 300-iteration runs, single GPU.

RL collection throughput (step FPS, rollout phase only, no optimizer time):

| task | envs | 2.3.2 | 3.0.0b2 | delta |
|---|---|---|---|---|
| Humanoid-Direct | 16384 | 750k | 599k | **-20%** |
| Humanoid-Direct | 8192 | 622k | 551k | -11% |
| Humanoid-Direct | 4096 | 516k | 445k | -14% |
| Shadow-Cube (Repose) | 16384 | 311k | 319k | +2% |
| Shadow-Cube (Repose) | 8192 | 249k | 257k | +3% |
| Shadow-Cube (Repose) | 4096 | 183k | 193k | +6% |

Pure sim without RL (stepping DirectRLEnv with random actions, 1000 timed steps): Humanoid 1267k -> 1185k (-6%) at 16K, 971k -> 925k (-5%) at 8K; Shadow 528k -> 517k (-2%), 394k -> 384k (-3%).

[...]

PR #7362, opened 2026-08-26
[Performance] Cache unchanged local wrench composition

AntoineRichard:
# Description

Avoids repeated WrenchComposer body-frame composition kernel launches when inputs have not changed and all wrenches are local-frame. A conservative pose-dependency flag keeps global and mixed wrench paths recomposing on every explicit call.

Addresses isaac-sim/IsaacLab-Internal#911.

No new dependencies.

## Performance

A writer-level spike with 4,096 environments measured:

- Newton: about 40–45% lower external-wrench writer time
- PhysX: about 60% lower external-wrench writer time

These measurements used mocked asset writers and do not claim an end-to-end task speedup. The regression test verifies one composition launch for repeated unchanged local wrenches and two launches for global wrenches.

## Type of change

- Performance improvement (non-breaking change)

## Release backport

- [x] Backport this pull request to the active release branch after it merges into develop

## Screenshots

Not applicable.

## Validation

- [x] WrenchComposer unit suite: 376 passed
- [x] Repository formatting and pre-commit checks passed
- [x] Changelog fragment validation passed

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] I have updated the affected method documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove the optimization is effective
- [x] I have added an isaaclab changelog fragment
- [x] My name is already present in CONTRIBUTORS.md

PR #7431, opened 2026-08-29
Optimize PhysX global external wrench submission

NeoZng:
# Description

This PR avoids unnecessary body-transform reads when PhysX can submit an external wrench directly in the global frame at the body's center of mass.

`WrenchComposer.all_wrenches_global_at_com` tracks a conservative aggregate invariant across all `add_*`, `set_*`, and `add_raw_buffers_from` inputs. The three PhysX writers use the raw global-at-CoM force and global torque buffers only while that invariant holds. Any local-frame or explicitly positioned force contribution, including one from another instantaneous or permanent composer, retains the existing body-frame composition path. Partial resets also retain the conservative fallback state.

This keeps the optimization independent of individual MDP terms and covers multiple external-wrench contributors without changing their existing add/set semantics. It is scoped to PhysX `Articulation`, `RigidObject`, and `RigidObjectCollection`; other backends and assist-force MDP code are unchanged. No new dependencies are required.

Related context: #7398 addressed cached COM-property access. This PR does not reopen that issue; it optimizes the remaining body-pose composition path for eligible external wrenches.

## Validation

- `uv run isaaclab -f` and `./isaaclab.sh --format`: all pre-commit hooks passed.
- WrenchComposer CPU/CUDA suite: 374 passed.
- PhysX writer regression using the existing kitless runtime fixtures: 18 passed on CPU/CUDA. It covers direct global-at-CoM submission and the local-frame/positioned-force fallbacks for all three affected asset types.
- Focused cross-backend interface/body-ordering suite: 6 passed, 18 skipped because the corresponding optional backends were unavailable, and 2293 deselected.
- The direct-submission regression fails on the unpatched develop writer and passes with this change because the baseline reads body poses before submission.

### PhysX performance profile

The workload was a custom articulated-robot recovery setup. A reset event installed a persistent world-frame upward assist force at a body CoM, and an interval event refreshed it every control step through:

```python
asset.permanent_wrench_composer.set_forces_and_torques_index(
    forces=forces,
    body_ids=[body_id],
    env_ids=env_ids,
    is_global=True,
)
```

The composer submitted the buffered force on every physics step. The same assist-force event code was used for the develop baseline and the patch.

A same-process A/B used 4096 environments, a 27-body articulation, 100 warmup steps, and 2000 alternating paired samples per path:

| Scope | Develop | Patch | Change |
|---|---:|---:|---:|
| External-wrench writer mean | 0.313958 ms | 0.188137 ms | -0.125821 ms (-40.08%) |
| Full physics-cycle mean | 10.971542 ms | 10.856342 ms | -0.115200 ms (-1.05%) |

The reported 20-iteration PhysX training profiles used the same seed and configuration. The task had no Eval or policy export. RSL-RL measures `collect_time` around the rollout and logs both collection and learning time before saving a checkpoint, so checkpoint writing, startup, shutdown, and process wall time are excluded. The means below discard iterations 0-4 and use the remaining 15 samples per run. Positive improvements indicate that the patch was faster.

Collection time:

| Environments | Develop | Patch | Improvement | Paired saving (95% interval) |
|---:|---:|---:|---:|---:|
| 4096 | 1.752667 s | 1.728533 s | +1.38% | 24.13 ms [13.85, 34.42] |
| 16384 | 3.170067 s | 3.131667 s | +1.21% | 38.40 ms [24.85, 51.95] |

Iteration time:

| Environments | Develop | Patch | Improvement |
|---:|---:|---:|---:|
| 4096 | 1.770000 s | 1.745333 s | +1.39% |
| 16384 | 3.198667 s | 3.160667 s | +1.19% |

All reported runs completed 20/20 iterations and show a small collection-time benefit. The targeted 2000-pair writer A/B remains the direct evidence for the optimized path, rather than a claim of a universal training speedup.

## Type of change

## Release backport

## Screenshots

Not applicable; this change has no visual output.

## Checklist

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -219,6 +219,7 @@ Guidelines for modifications:
 * Yun Liu
 * YuTeh Shen
 * Zehao Wang
+* Zeng Qingcheng
 * Zijian Li
 * Ziqi Fan
 * Zoe McCarthy
--- a/source/isaaclab/changelog.d/external-wrench-global-submission.minor.rst
+++ b/source/isaaclab/changelog.d/external-wrench-global-submission.minor.rst
@@ -1,0 +1,4 @@
+Changed
+^^^^^^^
+
+* Exposed whether all buffered external wrenches are global-frame and applied at the bodies' centers of mass.
--- a/source/isaaclab/isaaclab/utils/wrench_composer.py
+++ b/source/isaaclab/isaaclab/utils/wrench_composer.py
@@ -66,6 +66,9 @@ class WrenchComposer:
         self._asset = asset
         self._active = False
         self._dirty = False
+        # Track whether all buffered external-wrench contributions remain directly representable in the global frame
+        # at each body's CoM. Partial resets conservatively preserve false.
+        self._all_wrenches_global_at_com = True
         if hasattr(self._asset.data, "body_com_pos_w"):
             self._get_com_pos_fn = lambda a=self._asset: a.data.body_com_pos_w.warp
         else:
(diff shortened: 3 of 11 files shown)


AntoineRichard:
@NeoZng Thanks for the PR, I don't think this is the right form. I had pushed something addressing the other side of the problem here: #7362, but I think I can do both at the same time while keeping everything in the composer. The assets should not do weird things with the composer outputs, the composer should infer the most efficient path on its own, it has everything it needs.

NeoZng:
ur right. maintaining a state machine inside wrenchcomposer is much better than having the asset side do it separately and repeatedly (also, creating unnecessary cross-layer coupling). maybe we need an "aggregate mode" enum to tackle different fast paths. looking forward to your patch.
pr closed.

PR #7434, opened 2026-08-29
Stops cubes_stacked from reporting success for a cube still in flight

2047767028-lang:
# Description

`cubes_stacked` decides the stack tasks' success from an instantaneous geometric configuration — each cube within `xy_threshold` of the one below it and one cube-height above it — with no requirement that the cubes be supported or stationary. A cube released above its target satisfies exactly that description for a frame or two on the way down.

That matters more than it sounds, because the Mimic data generator accumulates success across the whole episode:

```python
generated_success = generated_success or exec_success
```

so a single frame of a cube passing through the right height marks the entire episode a success, and it is written into the generated dataset as a demonstration of the task being completed.

### Evidence

Replaying the criterion offline over two generated datasets (the reproduction was validated first by confirming it passes 100% of the demos the generator accepted):

| dataset | accepted demos | stack broken at the final frame |
|---|---|---|
| A | 380 | **28 (7.4%)** |
| B | 358 | **19 (5.3%)** |

In those demos the top cube ends a median 6.3 cm away in xy and at the *same height* as the cube it should be sitting on — it is on the table, not on the stack. The discriminator is what the mechanism predicts: cube_3's speed at the qualifying frame is a median 0.028–0.032 m/s in the broken demos versus **0.0000 m/s** in the rest.

One timeline, taken from the case most favourable to any other explanation — the one whose jaws came closest to fully open while the criterion held:

```
frame  stacked  cube3-cube2 dz  cube3 z   jaw error
  191            7.91 cm        14.62 cm   18.465 mm   <- held above the target
  193            5.60 cm        12.33 cm    3.124 mm   <- jaws opening, cube falling
  195     *      5.01 cm        11.76 cm    0.424 mm   <- criterion fires, mid-fall
  196     *      4.30 cm        11.04 cm    0.135 mm
  199            0.33 cm         4.59 cm    0.012 mm   <- cube reaches the table
  205            0.00 cm         2.05 cm    0.007 mm   <- 6.8 cm from cube_2
```

The gripper releases 7.5 cm above `cube_2`; the cube free-falls through the height the criterion is looking for and lands on the table.

### The fix

Require the cubes to have settled. `max_lin_vel` defaults to `0.01` m/s — comfortably above the noise floor of a cube at rest and well below anything falling — and can be set to `None` to skip the check.

The default is not guessed. Replaying 100 generated demos through the recorded `root_velocity` — the same signal the check reads — shows cubes resting on the stack are **not** at zero: the contact solver leaves a median 0.0132 m/s, peaking at 0.0341 m/s over the 91 sound demos, against a median 0.104 m/s for a cube caught mid-fall.

| max_lin_vel | keeps sound demos | keeps defective ones |
|---|---|---|
| 0.010 | 15/91 | 0/9 |
| 0.020 | 77/91 | 0/9 |
| **0.050** | **91/91** | **2/9** |
| 0.100 | 91/91 | 4/9 |

It is deliberately not a perfect separator: the distributions overlap at the bottom because a cube near the apex of a bounce is briefly slow, and tightening past 0.034 m/s starts discarding real demonstrations. An angular-velocity term buys one more defect at the cost of four sound demos, so it was left out.

Measured end to end on the stock `Isaac-Stack-Cube-Franka-IK-Rel-Mimic-v0`, generating 100 accepted demos each way:

| criterion | accepted | ends broken | attempts | success rate |
|---|---|---|---|---|
| before | 100 | **9 (9.0%)** | 279 | 35.8% |
| after | 100 | **2 (2.0%)** | 300 | 33.3% |

### Scope and caveats

- All six call sites of `cubes_stacked` are `DoneTerm` success terminations. The Mimic subtask-boundary signals use `object_stacked` (observations side), so annotation boundaries are unaffected. `object_stacked` has the same one-sided geometry and can presumably fire in flight too; it is consumed edge-triggered by annotation and is left out of scope here.
- The 0.05 m/s default was measured on the stack task's default backend (PhysX, `stack_env_cfg.py: default = isaacsim_physx`). The Newton solver config offered alongside it, and the SO-101 stack task which shares this termination, were not measured separately; if either shows a higher resting jitter the threshold should be raised for it rather than the check disabled.

### Tests

`source/isaaclab_tasks/test/contrib/stack/test_cubes_stacked_at_rest.py` — ten sim-free cases with hand-built scene objects: a resting stack passes; a cube passing through the stacked configuration mid-fall does not (the defect); `max_lin_vel=None` restores the position-only check; every cube in the checked set is subject to the test; the 0.030 / 0.049 / 0.051 m/s edges; the two-cube variant ignores the third scene object; the Franka variants' remapped cube roles are honoured; batched verdicts stay independent; the gripper check still gates.

### A note on the gripper tolerance, which looks like the opposite bug

`atol`/`rtol` of `1e-4` against `gripper_open_val` work out to 0.104 mm on a 40 mm jaw travel, and 11% of failed attempts had reached the stacked geometry but missed that tolerance — which reads like a false-negative worth relaxing. It is not. Every one of those 76 episodes is a drop: none has its stack intact at the final frame, and the top cube has moved a median 10.8 cm by then. Relaxing the tolerance to 3 mm would have admitted 54 of them into the dataset. The tight tolerance is masking this defect by accident; it is not causing one, and it is left alone here.

## Type of change

This lowers reported success rates for the stack tasks slightly, in both Mimic generation and policy evaluation, because episodes that end with the cube on the table stop counting. `max_lin_vel=None` in the termination term's `params` restores the previous behaviour.

## Checklist

- [ ] I have run the `pre-commit` checks with `./isaaclab.sh --format`

https://claude.ai/code/session_01JFz91VMgmrjS4XR6f9Q9Z2


ooctipus:
can you remove the test file? I think this pr is then good to go

source/isaaclab_tasks/isaaclab_tasks/contrib/stack/mdp/terminations.py
```diff
@@ -34,7 +34,21 @@ def cubes_stacked(
     height_diff: float = 0.0468,
     atol: float = 0.0001,
     rtol: float = 0.0001,
+    max_lin_vel: float | None = 0.05,
 ) -> torch.Tensor:
+    """Whether the cubes are stacked, released, and at rest.
+
+    Args:
+        max_lin_vel: Speed, in m/s, below which a cube counts as at rest. The position checks below
```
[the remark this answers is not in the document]
2047767028-lang: Done in e005587 — `Speed [m/s] below which a cube counts as at rest`, and the measured values in the paragraph below now carry `[m/s]` too.

2047767028-lang:
Happy to — just want to check which part you would rather not have, because I suspect the objection is the weight rather than the coverage.

The file is a plain unit test (hand-built scene objects, no `gym.make`, no USD, no physics), but I gave it the `AppLauncher(headless=True)` preamble that the `isaaclab_mimic` tests use, which pulls up Kit for what is really pure tensor logic. `test_so101_pose_ik.py`, right next to it in `test/contrib/stack/`, is sim-free and imports nothing from `isaaclab.app` — that is the shape this should probably have had.

So, whichever you prefer:

1. **Drop it entirely** — say the word and I will remove it in the next push.
2. **Keep it, without Kit** — drop the `AppLauncher` preamble so it matches `test_so101_pose_ik.py` and runs in milliseconds.

I have no attachment to it; the fix stands on its own either way. Option 2 only appeals to me because the failure it guards is quiet — the criterion returns `True` for a cube in flight, so a regression shows up as slightly inflated success rates rather than as anything failing.

kellyguo11:
if we can keep the test running fast without kit, that should work as well.

2047767028-lang:
Done in 282d9cc — dropped the `AppLauncher` preamble, so the file now has the same shape as `test_so101_pose_ik.py` next to it. Nothing else changed.

Measured on a `develop` checkout with no Kit available: 12 passed in 0.68 s, and after the import `sys.modules` holds no `omni` / `carb` / `pxr` / `isaacsim` entries. `ruff` and `codespell` are clean.

source/isaaclab_tasks/test/contrib/stack/test_cubes_stacked_at_rest.py
AntoineRichard, line 1: THis would need to be reduced to what is strictly the minimum set of test that'd be need to validate.

source/isaaclab_tasks/isaaclab_tasks/contrib/stack/mdp/terminations.py
```diff
@@ -34,7 +34,21 @@ def cubes_stacked(
...
+    Args:
+        max_lin_vel: Speed [m/s] below which a cube counts as at rest. The position checks below
+            describe an instantaneous configuration, and a cube dropped above its target passes
+            through that configuration on the way down -- without this, the drop is scored as a
+            successful stack. Pass None to skip the check.
+
+            The default leaves room for the contact solver's residual jitter, which keeps a settled
+            cube from ever reading exactly zero: over a sample of 91 sound generated demos, cubes
+            resting on the stack read a median 0.013 [m/s] and peaked at 0.034 [m/s], while a cube
+            caught mid-fall read a median 0.10 [m/s].
```
AntoineRichard: Keep comments as concise as possible.

AntoineRichard (changes requested):
@2047767028-lang thanks, the fix looks reasonable. @maxkra15  can you double check? Please make sure to make the code as dense as possible!

2047767028-lang:
Done in db144c9. The test is down to five cases in 41 lines (resting stack, each cube still moving, `max_lin_vel=None` opt-out), the docstring is two sentences, the inline comment one line, and the gate a single loop. The empirical justification for the 0.05 m/s default stays in the PR description. Changelog fragment shortened to match. Ruff and codespell clean; 5 passed in 0.7 s without Kit.

source/isaaclab_tasks/test/contrib/stack/test_cubes_stacked_at_rest.py
```diff
@@ -0,0 +1,41 @@
...
+"""Tests that :func:`cubes_stacked` does not report success for a cube still in flight."""
+
+from types import SimpleNamespace
+
+import pytest
+import torch
+
+from isaaclab_tasks.contrib.stack.mdp.terminations import cubes_stacked
+
+CUBE = 0.0468
```
AntoineRichard: why hardcode this here?

AntoineRichard:
Agreed with @ooctipus — removal pushed in 91857e4.

@kellyguo11 the Kit-free version did run fast (0.7 s), so speed turned out not to be the deciding factor. What settled it for me is that the test hand-built the scene objects it then read from, so it re-implemented the `ProxyArray` accessors rather than exercising them: a change to `root_lin_vel_w` or `.torch` would break the termination in a real env with the test still green. It also never pinned a speed near the 0.05 m/s default it was meant to defend — the falling cubes were at 0.1 m/s, so any threshold in (0, 0.1) passes it. That left it guarding a three-line gate against deletion, in exchange for a fake scene to keep in sync.

The measurements behind the 0.05 m/s default stay in the PR description. If we want real coverage here later, the version worth writing boots the stack env and writes cube root states directly to place one mid-fall — that exercises the actual data path, at the cost of a sim boot.

kellyguo11:
run-ci

2047767028-lang:
@AntoineRichard the change you asked for is already in — you pushed it yourself in 91857e4 (test file removed, agreeing with @ooctipus). The review is still formally `CHANGES_REQUESTED` though, which blocks the PR, so could you re-review when you get a moment? The diff is now just `terminations.py` +12, a changelog fragment and CONTRIBUTORS.

Also flagging @maxkra15 — you were mentioned in the review body but never added as a reviewer, so there is no review request on your side and this may not have shown up in your queue.

Both red checks are stale and unrelated to this diff:

- `isaaclab_ov`: `test_external_force_on_multiple_bodies_at_position[cuda:0-1]` fails at `source/isaaclab_ov/test/assets/test_articulation.py:2348` with `assert 0.004370182752609253 > 0.1` — the Z-only assertion that #7504 replaced with `torch.linalg.vector_norm(...)`. That job checked out `Merge 91857e4724ac6c2e7aecbcf0706037e8e327caac into 13296b112817ba89c0ba5eec1671fcf3fb07afe6` (2026-09-02 05:14 UTC), which predates #7504's merge commit 541de9a7a (2026-09-03 03:20 UTC), so the fix was not in the tree. This PR only touches `isaaclab_tasks`.
- `Check for Broken Links`: an `opensource.org` 403. The same check went red on two intermediate heads of #7433 and green on its final one.

A fresh `run-ci` should clear both. Happy to merge `develop` into the branch instead if you would rather the retrigger came from a push.

2047767028-lang:
run-ci

AntoineRichard (approved):

Issue #4119, opened 2025-12-01
[Question] FabricManager::initializePointInstancer mismatched prototypes


hojae-io:
### Question

Every time I create a visualization marker as below, I get the FabricManager warnings.
Is there any way to remove these warnings?

```
# Create markers if necessary for the first time
if not hasattr(self, "_lft_lin_vel_visualizer") or self._lft_lin_vel_visualizer is None:
    # Left fingertip linear velocity visualizer (red)
    marker_cfg = RED_ARROW_X_MARKER_CFG.copy()
    marker_cfg.prim_path = "/Visuals/Action/left_fingertip_linear_velocity"
    marker_cfg.markers["arrow"].scale = (0.1, 0.1, 0.1)
    self._lft_lin_vel_visualizer = VisualizationMarkers(marker_cfg)
    # Right fingertip linear velocity visualizer (red)
    marker_cfg = RED_ARROW_X_MARKER_CFG.copy()
    marker_cfg.prim_path = "/Visuals/Action/right_fingertip_linear_velocity"
    marker_cfg.markers["arrow"].scale = (0.1, 0.1, 0.1)
    self._rft_lin_vel_visualizer = VisualizationMarkers(marker_cfg)
    # Left fingertip angular velocity visualizer (green) - only if angular velocity is enabled
    if self._include_angular_velocity:
        marker_cfg = GREEN_ARROW_X_MARKER_CFG.copy()
        marker_cfg.prim_path = "/Visuals/Action/left_fingertip_angular_velocity"
        marker_cfg.markers["arrow"].scale = (0.1, 0.1, 0.1)
        self._lft_ang_vel_visualizer = VisualizationMarkers(marker_cfg)
        # Right fingertip angular velocity visualizer (green)
        marker_cfg = GREEN_ARROW_X_MARKER_CFG.copy()
        marker_cfg.prim_path = "/Visuals/Action/right_fingertip_angular_velocity"
        marker_cfg.markers["arrow"].scale = (0.1, 0.1, 0.1)
        self._rft_ang_vel_visualizer = VisualizationMarkers(marker_cfg)
# Set their visibility
self._lft_lin_vel_visualizer.set_visibility(True)
self._rft_lin_vel_visualizer.set_visibility(True)
if self._include_angular_velocity:
    self._lft_ang_vel_visualizer.set_visibility(True)
    self._rft_ang_vel_visualizer.set_visibility(True)
```

```
[...]

PR #7445, opened 2026-08-31
[Bump] Bump Newton to 1.6.0rc1 for newton#4017

hujc7:
## Summary

- Bumps Newton from `1.5.1` to `1.6.0rc1`, the first release carrying [newton#4017](https://github.com/newton-physics/newton/pull/4017), unblocking the MuJoCo tendon adapter that PR #7161 needs.
- `warp-lang` and `mujoco`/`mujoco-warp` move with it, because `1.6.0rc1` requires them.
- Pins `newton[sim]` in `[project].dependencies` as well, so a wheel built from this metadata resolves the fix rather than falling back to the newest stable.
- Updates test literals in `isaaclab` and `isaaclab_newton` for Newton's new actuator class names, and asks for a determinism guarantee in the manipulation determinism test.

# Description

`resolve_fixed_tendon_actuator_columns` reads `mujoco.actuator_trntype` through an
`ArticulationView`. Every Newton 1.5.x release rejects that with *"has custom frequency
'mujoco:actuator' which is not supported by ArticulationView"*; newton#4017 makes
`ArticulationView` generic over custom frequencies.

`1.6.0rc1` (published 2026-09-03) is the first release containing that fix — verified two ways:
the merge commit `4055a159` is an ancestor of `v1.6.0rc1`, and `custom_frequency_label_attributes`,
the symbol the fix introduces in `newton/_src/sim/model.py`, is present at `v1.6.0rc1` and absent
at `v1.5.1`. No `1.6.0` final exists yet, and no 1.5.x backport carries the fix, so `1.6.0rc1` is
the only released option.

The release requires `warp-lang>=1.17.0` and `mujoco-warp~=3.12.0`, so those pins move too:

| Package | Before | After |
| --- | --- | --- |
| `newton[sim]` | `1.5.1` | `1.6.0rc1` |
| `warp-lang` | `1.16.0` | `1.17.0` |
| `mujoco` | `3.11.0` | `3.12.0` |
| `mujoco-warp` | `3.11.0` | `3.12.0` |

`warp-lang` is pinned to the `1.17.0` release rather than left as `>=1.17.0`: Newton's prerelease
enables prerelease resolution, and an unbounded bound then resolves to a nightly
(`1.18.0.dev20260903`).

## Wheel installs get the pin too

`[tool.uv].override-dependencies` only reaches `uv sync`, so it does not constrain a wheel built
from this metadata. `1.6.0rc1` is a prerelease, which a default resolve skips, so a loose bound
resolved `newton 1.5.1` — without newton#4017, the fix this bump exists to deliver. Since
`wheel.yml` builds on every push to `develop`, that would have shipped.

`[project].dependencies` now pins `newton[sim]==1.6.0rc1`, matching the override. A default
resolve of the new metadata returns `1.6.0rc1` under both `uv pip compile` and `pip download`,
with no `--pre` flag. `isaacsim-core==6.0.1.0` requires `newton[sim]==1.2.1`, so installing the
`isaacsim` extra still needs an override — exactly as on `develop`, which carries an equally
exact pin in the same place.

## Newton actuator class rename

`1.6.0rc1` introduces `DrivePD` / `DriveNeuralMLP` / `DriveNeuralLSTM` alongside the existing
`ControllerPD` / `ControllerNeural*`. Both names remain importable, so this is **not** a breaking
API change and no production code changes. Tests in `isaaclab` and `isaaclab_newton` assert on the
class name Newton *constructs with*, which switched to the `Drive*` form, so those literals are
updated.

## Determinism

`test_manipulation_env_determinism[cuda-Isaac-Open-Drawer-Franka]` failed on this branch in CI.
Two separate questions: whether the bump caused it, and why it failed at all.

**Not caused by the bump.** Each arm was run 5x on the same machine, because a single run of an
intermittent test samples a coin rather than measuring it:

| GPU | develop pins (`newton 1.5.1` / `warp 1.16.0` / `mujoco 3.11.0`) | this PR (`newton 1.6.0rc1` / `warp 1.17.0` / `mujoco 3.12.0`) |
| --- | --- | --- |
| RTX PRO 6000 Blackwell | 3 / 5 failed | 3 / 5 failed |
| L40 | 3 / 3 passed | 4 / 4 passed |

Both pin sets fail at the same rate on Blackwell and both are clean on L40, so the failure is
hardware-dependent, not version-dependent. Intermediate Newton revisions (the commits before and
at [newton#4119](https://github.com/newton-physics/newton/pull/4119), the MuJoCo 3.12 bump) show
the same behaviour, so neither Warp 1.17, MuJoCo 3.12, nor #4119 is implicated.

**Why it failed.** The test asserted bit-reproducible rewards without ever asking for a
determinism guarantee. Newton defaults to `wp.DeterministicMode.NOT_GUARANTEED`, under which
Warp's atomics may accumulate in any order, so two runs of the same environment are not required
to agree — the assertion was stricter than the configuration it ran under, and whether it held
came down to GPU scheduling.

The Newton-backed case now passes `deterministic_mode="run_to_run"`, which is exactly the
guarantee this test asserts: one device, one process, run twice. `deterministic_mode` is a Newton
setting, so the PhysX-backed `IsaacContrib-Lift-Cube-Franka` case leaves it unset and the helper
keeps asserting that the mode is only applied to a `NewtonCfg` backend. This is a
test-configuration fix, not a masking of the intermittency: the guarantee the assertion depends
on is now actually requested.

Related: [#7561](https://github.com/isaac-sim/IsaacLab/pull/7561) addresses the adjacent layer and
does not overlap this PR. Newton's solvers take a per-module `deterministic` argument, which is
what `deterministic_mode` reaches and what this PR sets; Newton's sensor and geometry kernels take
no such argument and fall back to the `warp.config.deterministic` global, which #7561 raises from
the same config field in the RL entrypoint.

## Type of change

## Release backport

## Validation

- `uv lock --check` — clean; the relock moves exactly the four packages above, and the
  `[project].dependencies` pin moves exactly one further lockfile line
- Default resolve of the published metadata returns `newton==1.6.0rc1` under both
  `uv pip compile` and `pip download`, with no `--pre` flag
- `uv run --frozen python -m pytest source/isaaclab/test/cli/test_wheel_builder_metadata.py source/isaaclab/test/cli/test_uv_run_pyproject.py` — 21 passed, which enforces that the two generated `uv-overrides.txt` files match the override table and that `[tool.isaaclab.versions]` mirrors the dependency pins
- `uv run isaaclab -f` — passed
- Full Docker + Tests suite — **48 passed / 0 failed** on `b184335e80b`


hujc7:
run-ci

kellyguo11 (approved):

PR #7428, opened 2026-08-29
[Docs] Consolidate and clean migration guide

StafaH:
# Description

Consolidates the migration documentation into one user-facing guide instead of separate nested pages.

- Adds a starting-point chooser and in-page table of contents, with the Isaac Gym migration link first.
- Includes the Isaac Gym, simulation comparison, and deformables content directly in the main migration guide.
- Keeps the former subpage sources as include-only fragments so Sphinx does not build standalone pages.
- Adds a concise recommended workflow for Isaac Gym migrations.
- Updates the Isaac Gym migration skill to point to the consolidated guide.

## Type of change

- Documentation update

## Release backport

- [x] Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

Not applicable.

## Validation

- `uv run --isolated --extra test -- make -C docs current-docs`
- `uv run --no-project python tools/skills/cli.py check`
- Formatting, RST, codespell, and Git LFS pre-commit checks

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the formatting and pre-commit checks
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] My name already exists in `CONTRIBUTORS.md`
- [x] Tests are not applicable to this documentation-only change
- [x] Changelog fragments are not applicable because no `source/<pkg>/` package changed

PR #7448, opened 2026-08-31
Condense the deformables migration fragment

mmichelis:
# Description

#7428 consolidated the migration guide and moved `migrating_deformables.rst` into `include/deformables.rst` with its content unchanged. That leaves the deformables content rendered twice on the same page: the `Deformable Object API Changes` summary in the assets section, and the full fragment appended at the end. This PR condenses the fragment and collapses the duplicated summary into a pointer to it.

Most of the fragment duplicated content that is already maintained elsewhere:

- The "Added Properties" and "Material Changes" property lists restate docstrings that autodoc already renders from `schemas_cfg.py` and `physics_materials_cfg.py`. The added-property list had in fact drifted from source, omitting `settling_threshold`, `sleep_threshold`, `max_depenetration_velocity`, `self_collision` and `self_collision_filter_distance`.
- The surface-deformable and USD-file code examples repeat `scripts/demos/deformables.py` verbatim (`deformables.py:157` for the cloth, `:165` for the teddy bear), and they are not migration diffs since surface deformables are new.
- The `root_physx_view` to `root_view` section repeats the guide's own asset-view section.

The fragment goes from 240 lines to roughly 100. What was kept is the content that cannot be recovered from source:

- The surface versus volume distinction and which material cfg selects which type.
- A trimmed import table, now using the short `isaaclab_newton.sim` paths so the Newton rows are consistent with the PhysX rows. Both packages use the same `lazy_export` machinery, so the deeper submodule paths were unnecessary.
- The removed-properties table with replacements, now class-qualified because the entries span `PhysxDeformableBodyPropertiesCfg`, `PhysxDeformableBodyMaterialCfg` and the deleted `PhysxDeformableCollisionPropertiesCfg`. Autodoc can never render a class that no longer exists, so a migration note is the only place this can live.
- The volume-only kinematic target restriction. This is enforced at runtime in `deformable_object.py` but has no `Raises:` entry in the method docstring, so the migration guide is currently its only documentation.

The volume before/after example is kept. It is the one thing a migration guide uniquely provides: the old call shape next to the new one, in context. `sim_utils.DeformableBodyPropertiesCfg` and `sim_utils.DeformableBodyMaterialCfg` appear nowhere else in `docs/` or `scripts/`, so a reader cannot reconstruct the old form from any current-API source. Its `emphasize-lines` are corrected to point at the lines that actually differ; they previously highlighted the first two import lines, which are identical in both blocks.

The surface and USD examples are replaced with a pointer to `scripts/demos/deformables.py` and the deformable object tutorial, which are runnable and cover those cases.

The added-property list is replaced with a pointer rather than a shorter list. A hand-maintained enumeration is what drifted in the first place, so the fragment now enumerates only removed fields, which cannot drift.

One drive-by correction: the collision-offset example referred to `PhysxCollisionCfg` in a section that otherwise imports from `isaaclab_physx.sim`, but that class is only exported from `isaaclab_physx.sim.schemas`. It is now qualified explicitly.

## Type of change

- Documentation update

## Release backport

## Checklist

Notes on the two unchecked items: this is a docs-only change with no source package touched, so there is nothing to test and no package needs a changelog fragment.

Verification: a full Sphinx build succeeds with 15 warnings, all of which are pre-existing `duplicate object description` autodoc warnings unrelated to this PR. There are no prose or cross-reference warnings. `./isaaclab.sh --format` passes. Every technical claim in the rewritten fragment was checked against source, including live imports of the `isaaclab_newton.sim` paths and `dataclasses.fields()` introspection of each removed and added property.

### DIFF
--- a/docs/source/migration/include/deformables.rst
+++ b/docs/source/migration/include/deformables.rst
@@ -5,143 +5,51 @@ Migration of Deformables
 
 .. currentmodule:: isaaclab
 
-In the newer versions of Omni Physics (107.0 and later), the old deformable body functionality has become deprecated.
-The following sections describe the changes to migrate to the new Omni Physics API, specifically moving away from
-Soft Bodies and towards Surface and Volume Deformables. The deformable object asset classes remain in
-``isaaclab.assets``. Schema define/modify functions remain unified in ``isaaclab.sim.schemas``, and deformable
-material spawning remains unified in ``isaaclab.sim.spawners.materials``. Deformable property and material
-configuration classes are backend-specific: PhysX configurations live in ``isaaclab_physx.sim`` and Newton
-configurations live in ``isaaclab_newton.sim``.
+Isaac Lab 3.0 updates the deformable body API to align with Omni Physics 110.0. The old soft body
+API is deprecated and replaced by two distinct deformable types:
 
-.. note::
+- **Volume deformables**: 3D objects simulated with a tetrahedral FEM mesh (soft cubes, teddy
+  bears). They support kinematic targets on individual vertices.
+- **Surface deformables**: 2D surfaces simulated directly on a triangle mesh (cloth, membranes).
+  They add stretch, shear, and bend stiffness, but do not support kinematic vertex targets.
 
-  The following changes are with respect to Isaac Lab v3.0.0 and Omni Physics v110.0. Please refer to the
-  `release notes`_ for any changes in the future releases.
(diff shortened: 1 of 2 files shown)


StafaH (approved):

AntoineRichard (approved):

AntoineRichard:
run-ci

PR #7453, opened 2026-08-31
[Newton] Delegate homogeneous world prefix generation to `ModelBuilder.replicate()`

camevor:
# Description

This depends on newton-physics/newton#4012 and pins the Newton 1.6 development build that first provides `ModelBuilder.replicate(..., label_prefixes=...)`.

The homogeneous Newton cloning path previously replicated the model and then walked every replicated entity label to replace the prototype environment with its destination environment. It now rebases only the retained prototype labels, restores them after the call, and lets `ModelBuilder.replicate()` assign each copy's destination prefix while it is created. String-valued custom attributes still use the existing targeted rewrite because Newton cannot identify which strings are entity paths.

The destination itself is the source of identity: prefixes are derived from the queued source/destination mapping. No clone-plan environment metadata, scene discovery, fallback clone path, or retained prototype mutation is introduced. Labels exactly at a directly cloned asset root, including Newton's `_free_joint` and `_articulation` companion labels, are expressed relative to the nearest templated destination ancestor and remain on the same fast path.

The Newton commit pin is mirrored in workspace and wheel installation inputs so every supported install receives the API this path requires.

The final PR is `+195/-218`, net **-23 LOC**.

## Performance

GPU 0, 4096 worlds, two samples per PR head. The targeted measurement uses one 48-body/48-shape Newton prototype and times replication plus all required label publication:

| Head | Targeted path median | Versus develop |
|---|---:|---:|
| develop base (`3a31389a9`) | 1.428 s | — |
| original PR (`cbb17ab6e`) | 1.160 s | -18.8% |
| cleaned PR (`844f4c4db`) | 1.171 s | **-18.0%** |

The cleanup is within 1.0% (11 ms) of the original PR result and preserves the removed-pass gain. The 257 ms reduction from develop is consistent with the original 215 ms observation on `Isaac-Velocity-Flat-G1`.

Whole-workload check used:

```bash
CUDA_VISIBLE_DEVICES=0 uv run --no-sync isaaclab zero_agent \
  --task Isaac-Lift-KukaAllegro --num_envs 4096 --max_steps 1 \
  --device cuda:0 --visualizer none physics=newton_mjwarp
```

| Head | Warm wall median | Scene creation median |
|---|---:|---:|
| original PR | 21.30 s | 6.324 s |
| cleaned PR | 21.39 s | 6.345 s |

That end-to-end difference is within startup noise; the cleanup introduces no measured workload regression.

## Type of change

- Startup performance enhancement

## Validation

[...]

PR #7462, opened 2026-08-31
Route clone backends through ClonePlan

ooctipus:
# Description

This is a narrow follow-up to merged #7453. It keeps that Newton fast path and makes the existing `ClonePlan` the single mapping consumed by USD, Newton, PhysX, and OvPhysX replication.

Planning records `context_rows`, mapping each simulation-owned clone-context type to the plan rows it consumes. `cloner.replicate(plan)` retrieves those already-registered contexts from `SimulationContext`, orders them by replication priority, and passes each the same plan. Backends no longer rebuild the mapping through local `queue(...)` or `queue_mapping(...)` state.

Clone planning is a NumPy host-side control plane. `ClonePlan`, its constructors, strategies, queries, and raw clone APIs use NumPy arrays; backend code no longer performs Torch device synchronization or `detach().cpu().tolist()` round trips. Runtime owners convert once when they need a device tensor. `InteractiveScene` derives its environment roots and device-side origins from the same plan instead of calculating a second grid.

The public structure stays small:

- `ClonePlan.global_paths` remains the explicit declaration for shared scene prims.
- `ClonePlan.env_ids` and `positions` remain optional for query-only plans; execution validates what it needs.
- Per-asset `cloning_contexts`, `cfg_rows`, `ReplicateSession`, and the existing scene-construction lifecycle remain.
- Standalone tooling keeps the raw USD, Newton, PhysX, and OvPhysX replication functions.
- OVRTX camera/export/SDP architecture remains deferred to the renderer cutover; this PR only removes Torch round trips from its mechanical plan-array consumption.
- No stage discovery, fallback context construction, registry-wide duck routing, or second clone mapping is introduced.

## Migration

Custom clone contexts now implement `replicate(plan)` and must be registered with `SimulationContext.get_or_create_backend(...)` before dispatch. Remove context-local `queue(...)` / `queue_mapping(...)` calls and the former backend `PHYSICS_CONTEXT` aliases; raw standalone replication functions remain available.

The high-level `stage=` argument is removed from `cloner.replicate(...)` and `ReplicateSession`; each simulation-owned context already owns its stage.

Clone arrays and raw clone API arrays are now NumPy. The unused public `device` arguments are removed from `CloneCfg`, plan constructors, `grid_transforms`, `ReplicateSession`, clone strategies, and raw backend replication functions. Custom clone strategies now implement `(combinations: np.ndarray, num_clones: int) -> np.ndarray`.

## Size

| Added | Deleted | Net |
| ---: | ---: | ---: |
| 1,267 | 1,700 | **-433** |

These counts are for this PR alone against current `develop`; merged #7453 code and dependency changes are no longer duplicated in the diff.

## Performance

The targeted benchmark warms with 128 worlds, then times replication and required label publication for one 48-body/48-shape Newton prototype at 4,096 worlds. Measurements taken while #7453 was under review used physical GPU 0 and the same Newton revision:

| Revision | Targeted median | Versus pre-#7453 develop |
| --- | ---: | ---: |
| develop before #7453 | 1.428 s | — |
| #7453 | 1.171 s | -18.0% |
| this follow-up before restacking | 1.181 s | **-17.3%** |

The restack removed the duplicated #7453 implementation rather than reimplementing it. A file-by-file audit confirms this PR no longer changes #7453's label-prefix fast-path files, and the one-source/all-true Newton route still delegates to `ModelBuilder.replicate(...)`.

## Type of change

- Breaking clone-context and NumPy API simplification
- Documentation update

## Validation

- `84 passed` — clone-plan algebra
- `7 passed` — replicate-session lifecycle
- `15 passed` — focused Newton label-prefix/root-name coverage from #7453
- Repository formatting, lint, RST, spelling, license, changelog, LFS, and hygiene hooks passed against current `upstream/develop`
- `git diff --check` passed
- Static audit found no Torch import or CPU/list materialization in the core, Newton, PhysX, or OvPhysX clone paths

## Release backport

## Checklist

- [x] I have run the relevant pre-commit checks
- [x] I have added focused contract, routing, and lifecycle tests
- [x] I have added a changelog fragment for every touched source package


ooctipus:
PR 3 has now been split at the execution/lifecycle boundary.

This PR contains only the existing `ClonePlan` execution contract and backend routing. The declarative cfg inventory, pre-construction lifecycle, queue deletion, camera/raycaster declaration changes, startup-profiler migration, and broad workflow call-site changes are in the follow-up PR 3B branch.

Accordingly, the earlier automated inline findings on `_plan_cfgs`, camera spawning, `RayCasterCfg.spawn`, and the renamed startup-profiler target refer to the previous full head and are no longer present in this diff. The preserved full tip is available on `ooctipus:codex/flattened-clone-plan-full`.

ooctipus:
run-ci

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -219,6 +224,46 @@ def make_valid_clone_combinations(
...
 
+def _context_rows(
+    cfgs: tuple[Any, ...], cfg_rows: dict[int, tuple[int, ...]], populated_rows: set[int]
+) -> dict[type[object], tuple[int, ...]]:
+    """Route plan rows to the clone contexts registered for this simulation."""
+    sim = sim_utils.SimulationContext.instance()
+    if sim is None:
+        return {}
+
+    usd_context = string_to_callable("isaaclab.cloner.usd:UsdReplicateContext")
```
StafaH: Is there a reason to do this instead of importing the function?

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -219,6 +224,46 @@ def make_valid_clone_combinations(
...
+
+    for cfg in cfgs:
+        rows = cfg_rows.get(id(cfg))
+        if rows is None:
+            continue
+        references = cfg.cloning_contexts
+        contexts = (
+            (() if physics_context is None else (physics_context,))
+            if references is None
+            else tuple(string_to_callable(value) if isinstance(value, str) else value for value in references)
```
StafaH: Optional: I'm not a fan of the usage of string_to_callable, you can create a strictly types dictionary instead that maps the string to specific functions.

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -219,6 +224,46 @@ def make_valid_clone_combinations(
...
+    """Route plan rows to the clone contexts registered for this simulation."""
+    sim = sim_utils.SimulationContext.instance()
+    if sim is None:
+        return {}
+
+    usd_context = string_to_callable("isaaclab.cloner.usd:UsdReplicateContext")
+    physics_context = sim.physics_manager.clone_context_type
+    if isinstance(physics_context, str):
+        physics_context = string_to_callable(physics_context)
+    routed: dict[type[object], set[int]] = {}
```
StafaH: The type info here is not helping, maybe a comment on what this dict will be used for?

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -300,13 +347,15 @@ def set_spawn_paths(spawn_cfg: Any, paths: list[str | None]) -> None:
         for cfg, spawn_cfg, destination, _ in groups:
             set_spawn_paths(spawn_cfg, [destination.format(0)])
         cfg_rows = {id(cfg): (0,) for cfg, _, _, _ in groups}
+        clone_mask = torch.ones((1, num_clones), dtype=torch.bool, device=device)
```
StafaH: Broader question, does the clone plan need to use torch at all? Since it's a one-time initialization, maybe numpy and bound to CPU would keep things fast and simple?

source/isaaclab/isaaclab/cloner/usd.py
```diff
@@ -14,96 +15,62 @@
...
+    from .clone_plan import ClonePlan
 
-def _select_env_ids(env_ids: torch.Tensor, mask: torch.Tensor | None, row: int) -> torch.Tensor:
-    """Return the environment ids selected by a replication row."""
+
+def _select_columns(env_ids: torch.Tensor, mask: torch.Tensor | None, row: int) -> torch.Tensor:
+    """Return the mask columns selected by a replication row."""
     if mask is None:
-        return env_ids
+        return torch.arange(len(env_ids), device=env_ids.device)
```
StafaH: Same comment is before, what would it take to drop torch from cloning? Newton replication is all on CPU, so GPU arrays wouldn't be useful there

source/isaaclab/isaaclab/cloner/usd.py
```diff
@@ -14,96 +15,62 @@
...
         # no-op outside a live Kit application.
         with disabled_fabric_change_notifies(self.stage):
-            self._apply_queue()
+            self._apply(items)
 
-    def _apply_queue(self) -> None:
-        """Author the queued copy specs into the stage's root layer."""
+    def _apply(
+        self,
+        items: list[tuple[str, str, torch.Tensor, torch.Tensor | None, torch.Tensor | None]],
```
StafaH: Can we use something more descriptive than "items"

source/isaaclab/test/install_ci/uv_pip/uv-overrides.txt
```diff
@@ -1,7 +1,7 @@
 numpy>=2
 mujoco~=3.11.0
 mujoco-warp~=3.11.0
-newton[sim]==1.5.1
+newton[sim] @ git+https://github.com/newton-physics/newton.git@24bd863528d6b91137408930d0fbe8fa216ad962
```
StafaH, line 4: This PR is mixing a newton bump as well?

ooctipus:
run-ci

ooctipus:
run-ci

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
```diff
@@ -234,14 +269,24 @@ def replicate_builder_mapping(
         stride = source_builder.shape_count
         source_xform_inv = _invert_xform(xforms_np[0])
         xforms = _compose_world_xforms(positions_np, quaternions_np, source_xform_inv)
-        builder.replicate(source_builder, num_worlds, xforms=xforms)
+
+        label_groups = _label_groups(source_builder)
+        original_labels = {name: list(labels) for name, labels in label_groups.items()}
+        try:
```
StafaH, line 286: why try, finally? Are we expecting possible failure?

source/isaaclab/isaaclab/cloner/clone_plan.py
```diff
@@ -219,6 +224,46 @@ def make_valid_clone_combinations(
...
+        physics_context = string_to_callable(physics_context)
+    routed: dict[type[object], set[int]] = {}
+
+    for cfg in cfgs:
+        rows = cfg_rows.get(id(cfg))
+        if rows is None:
+            continue
+        references = cfg.cloning_contexts
+        contexts = (
+            (() if physics_context is None else (physics_context,))
```
StafaH: This is very unreadable. Here and in a few other places, you can make much more readable code like:
```
  if references is None:
      contexts = () if physics_context is None else (physics_context,)
  else:
      contexts = tuple(string_to_callable(value) if isinstance(value, str) else value for value in references)
```

source/isaaclab_newton/isaaclab_newton/cloner/replicate.py
```diff
@@ -189,153 +182,77 @@ def _renderer_wants_visual_shapes() -> bool:
...
-"""Physics-only replication context for Newton assets.  USD replication is added automatically
-by :func:`~isaaclab.cloner.replicate` when the asset has a spawner and Kit is available."""
+    def replicate(self, plan: ClonePlan) -> tuple[ModelBuilder, object, dict]:
+        """Build and publish a Newton model from this context's plan rows."""
+        if plan.env_ids is None:
+            raise ValueError("ClonePlan.env_ids is required for replication.")
+        rows = plan.context_rows[type(self)]
+        cfg = self._sim.cfg.physics
+        load_visual_shapes = cfg.load_visual_shapes if isinstance(cfg, NewtonCfg) else None
+        return _replicate_newton(
```
StafaH, line 248: It's more cleaner to do something like:

```
  return _replicate_newton(
      stage=self._sim.stage,
      sources=tuple(plan.sources[row] for row in rows),
      destinations=tuple(plan.destinations[row] for row in rows),
      env_ids=plan.env_ids,
      mapping=plan.clone_mask[list(rows)],
      positions=plan.positions,
      quaternions=None,
      up_axis=self.up_axis,
      load_visual_shapes=...,
      global_paths=plan.global_paths,
  )
```

Similarly in a couple other places

source/isaaclab_ov/isaaclab_ov/cloner/replicate.py
```diff
@@ -57,158 +41,110 @@ def _matrix_to_clone_transform(matrix: Gf.Matrix4d) -> CloneTransform:
...
+    if positions is not None and positions.shape != (len(env_ids), 3):
+        raise ValueError(f"positions must have shape [num_envs, 3], got {list(positions.shape)}.")
+    if quaternions is not None and quaternions.shape != (len(env_ids), 4):
+        raise ValueError(f"quaternions must have shape [num_envs, 4], got {list(quaternions.shape)}.")
+
+    columns_by_row = [[] for _ in range(mapping.shape[0])]
+    for row, column in mapping.detach().to(dtype=torch.bool).nonzero(as_tuple=False).cpu().tolist():
+        columns_by_row[row].append(column)
+    env_ids_cpu = env_ids.detach().cpu().tolist()
+    positions_cpu = None if positions is None else positions.detach().cpu().tolist()
```
StafaH: The following code always makes me raise my eyebrows
`positions.detach().cpu().tolist()`
 We can probably write this better (and more optimally). I believe we did a similar cleanup in newton side once?

StafaH (approved):

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Done. `UsdReplicateContext` is now imported directly. Its `pxr` imports are deferred to the execution method so importing the cloner before Kit remains safe.

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Built-in USD and physics contexts are concrete types now. I kept `string_to_callable` only for the documented user-configurable `AssetBaseCfg.cloning_contexts` string boundary; a closed core dictionary would make core enumerate backend packages and would exclude custom context classes.

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Done. This is now named `rows_by_context`, and both the helper and `ClonePlan.context_rows` document that it maps clone-context types to their consumed plan rows.

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Done end-to-end. `ClonePlan`, plan construction, strategies, and queries now use CPU NumPy arrays, and the public `device` arguments (including `CloneCfg.device`) are removed.

source/isaaclab/isaaclab/cloner/usd.py
ooctipus: Done. USD cloning now accepts and selects NumPy arrays directly; there is no Torch import or device/CPU/list conversion in the USD clone path.

source/isaaclab/isaaclab/cloner/usd.py
ooctipus: Done. The collection is now `replication_rows`, grouped in `rows_by_depth`.

source/isaaclab/test/install_ci/uv_pip/uv-overrides.txt
ooctipus: This is temporary stack overlap with #7453, which is still open. That predecessor uses the new `ModelBuilder.replicate(..., label_prefixes=...)` API and therefore carries the Newton 1.6 development pin. The pin and fast-path files disappear from this PR after #7453 merges and this branch is restacked.

source/isaaclab_newton/isaaclab_newton/cloner/newton_clone_utils.py
ooctipus: `_rebase_labels` temporarily mutates the retained source builder and can itself fail after a partial mutation; `builder.replicate` can also fail. The `finally` restores the snapshot on both paths. Restoration also matters on success because these source builders are retained in `NewtonManager._cl_protos` for later consumers.

source/isaaclab/isaaclab/cloner/clone_plan.py
ooctipus: Done. The nested conditional is replaced with this explicit `if references is None` / `else` form.

source/isaaclab_newton/isaaclab_newton/cloner/replicate.py
ooctipus: Done. The Newton calls use named arguments now, as do the corresponding OvPhysX and PhysX calls.

source/isaaclab_ov/isaaclab_ov/cloner/replicate.py
ooctipus: Done. Clone planning and all raw backend clone boundaries are now NumPy, so OvPhysX indexes the arrays directly with no `detach().cpu().tolist()` chain.

ooctipus:
run-ci

ooctipus:
run-ci

PR #7480, opened 2026-09-02
[Workflow] Fix template generator in uv environments

StafaH:
# Description

The template generator attempted to install its dependencies with `python -m pip` every time it started. Fresh uv environments do not include the pip module, so the install failed and the generator then crashed while importing its prompt dependency.

This change:

- declares Jinja2 in the root dependency metadata so source and wheel installs provide the renderer
- removes runtime dependency installation from `isaaclab --new`
- replaces InquirerPy with the existing Rich dependency and native numbered multi-selection validation
- removes the duplicate template requirements file
- makes generated projects use the kit-less Newton backend by default instead of installing Isaac Sim
- forwards `isaacsim`, `ov`, `ovphysx`, and `ovrtx` as optional extras in generated projects
- documents backend selection in the generated README
- adds prompt, launcher, generated-metadata, and wheel-metadata regression coverage

No new prompt dependency is introduced: Jinja2 was already present transitively, while removing InquirerPy also removes its otherwise-unused `pfzy` dependency.

## Type of change

## Release backport

## Screenshots

Not applicable.

## Test plan

- `uv run python -m pytest --confcutdir=tools/template tools/template/test_cli.py -q` (4 passed)
- `uv run python -m pytest source/isaaclab/test/cli/test_misc_commands.py source/isaaclab/test/cli/test_wheel_builder_metadata.py -q` (18 passed)
- verified the generated-metadata regression fails against the pre-fix template because `isaacsim` is a base dependency
- `uv run python tools/changelog/cli.py check develop`
- `uv run isaaclab -f`
- generated an external direct single-agent project with RSL-RL in a clean uv environment without InquirerPy or pfzy, then compiled its generated Python and parsed both generated TOML files

## Checklist

- [x] I updated the generated project documentation
- [x] I have added a changelog fragment for the touched `isaaclab` package


kellyguo11 (approved):

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
...
+==============================
+
+The template generator creates a working Cartpole task, its selected agent
+configurations, and the packaging needed for the Isaac Lab CLI to discover it.
+The quickest path is an external project: choose a name and workflow, run
+``uv sync``, and start the generated task with the Newton backend.
+
+The generator runs entirely in the active Isaac Lab environment. It does not
+invoke ``pip`` or install a second set of template dependencies, so it works in
+the pip-less virtual environments created by ``uv``.
```
AntoineRichard: By reading this, it's not that clear to me what the template generator is used for.
I feel like it's missing the "why?".

docs/source/developer-tools/template_generator.rst
AntoineRichard, line 1: This reads like it's missing some context at the begining to explain why we have this.

AntoineRichard: This did a good job at it!

```
Create new project or task
==========================

The template generator creates external Isaac Lab projects and internal tasks.

The template generator enables you to create an:

* **External project** (recommended): An independent repository that can also load as an Isaac Sim extension.

  .. hint::

    The generator initializes a Git repository in the selected directory.

* **Internal task**: A task added directly to the Isaac Lab repository for contribution upstream.

  .. warning::

    Pip installations of Isaac Lab do not support *Internal* templates.
    If ``isaaclab`` is loaded from ``site-packages`` or ``dist-packages``, the *Internal* option is disabled
    and the *External* template will be used instead.
```

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
+.. _own-project:
+.. _template-generator:
+
+Build your own project or task
+==============================
+
+The template generator creates a working Cartpole task, its selected agent
```
AntoineRichard: Leading with a "Cartpole task" feels weird too, this is a detail.

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
...
+configurations, and the packaging needed for the Isaac Lab CLI to discover it.
+The quickest path is an external project: choose a name and workflow, run
+``uv sync``, and start the generated task with the Newton backend.
+
+The generator runs entirely in the active Isaac Lab environment. It does not
```
AntoineRichard: This reads weird, does it? If we create an external project this is not the case?

AntoineRichard (commented):
Should this be accompanied with a skill?

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
...
+   uv run isaaclab random_agent --task <TASK_NAME> --num_envs 16 --viz newton
+
+If the environment launches and the cart moves, the project is ready to edit.
+You can then train and play a policy with the same command surface used by
+Isaac Lab itself:
+
+.. code-block:: bash
+
+   uv run isaaclab train --rl_library rsl_rl --task <TASK_NAME>
+   uv run isaaclab play --rl_library rsl_rl --task <TASK_NAME> --checkpoint latest --viz newton
```
AntoineRichard: I don't like this split. I think the original doc did it better. Here we tell the user what to do without telling them why they are doing it. Leading with a: "Let's with creating an external project." would go be good. Even more if we explain the two options before!

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
...
+
+.. code-block:: bash
+
+   uv run --extra ovrtx isaaclab random_agent --task <TASK_NAME> \
+      physics=newton_mjwarp renderer=ovrtx
+
+Place ``--extra`` before ``isaaclab``. Keep it on every command that needs the
+optional runtime; this lets ``uv`` reproduce the intended environment without a
+separate installation step. See :ref:`isaac-lab-quickstart` for all physics,
+renderer, and visualizer combinations.
```
AntoineRichard: Does it make sense to have this here? Can't we just link the users to the different backends doc directly with a short note?

docs/source/developer-tools/template_generator.rst
```diff
@@ -0,0 +1,282 @@
...
+
+Choose **Internal** only when working from an Isaac Lab source checkout. The
+generator writes the new task into ``source/isaaclab_tasks`` instead of creating
+a separate project. From the Isaac Lab repository root, list and test it with:
+
+.. code-block:: bash
+
+   uv run python scripts/environments/list_envs.py --show_presets
+   uv run isaaclab random_agent --task <TASK_NAME> --num_envs 16
+   uv run isaaclab train --rl_library <RL_LIBRARY> --task <TASK_NAME>
```
AntoineRichard, line 262: Based on all the above, it's weird to have this here, could be later.

AntoineRichard:
The ASCII tree in "Understand the generated project" is abbreviated in a way that hides the module layer, so a reader can't match it against what the generator actually writes.

The doc shows:

```text
└── source/
    └── my_cartpole/
        ├── pyproject.toml
        ├── config/
        │   └── extension.toml
        └── my_cartpole/
            └── tasks/
```

I ran `generate()` for an external, manager-based single-agent project with rsl_rl/PPO (`name = "microduck_lab"`) and it wrote:

```text
source/microduck_lab/docs/CHANGELOG.rst
source/microduck_lab/microduck_lab/ui_extension_example.py
source/microduck_lab/microduck_lab/tasks/__init__.py
source/microduck_lab/microduck_lab/tasks/microduck_lab/mdp/
source/microduck_lab/microduck_lab/tasks/microduck_lab/config/cartpole/microduck_lab_env_cfg.py
source/microduck_lab/microduck_lab/tasks/microduck_lab/config/cartpole/agents/rsl_rl_ppo_cfg.py
```

Two things the tree stops short of: there's a task-family directory named after the project *inside* `tasks/`, and there's a `ui_extension_example.py` stub at the module root (plus the second `[[python.module]]` entry it needs in `extension.toml`), neither of which is mentioned anywhere in the doc. I only found out where to put things by reading `generator.py`.

Suggest expanding the tree two levels so it matches the generator's output, and adding a line about the UI stub — it's the file people will most want to delete first.

AntoineRichard:
Small wording suggestion rather than a restructure: the Task layer currently reads as prescriptive —

> The **Task** defines an environment and its agent configurations. Shared logic lives in the task-family directory, while robot-specific environment and agent configurations live under `config/<robot>`.

That's a good description of one common shape, but it reads as the required one. I hit two cases where it doesn't fit while porting a 14-task locomotion family (MicroDuck) into a generated project:

- The variation axis is behaviour, not robot: 11 behaviour families (velocity, standup, sitstand, roulade, ballkick, velstand, groundpick, rollercrouch, rollerstandup, spin, rollerslope) across 3 robot variants, with the robot picked per family. Following the letter would have produced eleven single-child `config/microduck/` directories.
- There's one `mdp/` package shared by all 11 families (actions, commands, curriculums, events, observations, rewards, symmetry, terminations — the substantial part of the port). The template's `mdp/` is per task family, and the four-layer model has no slot for cross-family shared code.

The mechanism handles both fine — I put the shared package at `tasks/mdp/` and the template's own `_BLACKLIST_PKGS = ["utils", ".mdp"]` already keeps it out of registration. It's only the prose that doesn't describe it.

Suggest saying `config/<robot>` is one common shape rather than the shape, plus one sentence on where cross-family shared code goes. Two sentences would have saved me a ruling.

AntoineRichard:
The doc and the generator are both silent on tests, and generated projects ship no pytest configuration. Two things bit me carrying an existing suite (19 files, 598 tests) into a generated project:

- Every marked test warns: `PytestUnknownMarkWarning: Unknown pytest.mark.unit`. Isaac Lab's marker vocabulary is declared in the monorepo root `pyproject.toml` and isn't part of any distribution, so it doesn't travel. I added a `[tool.pytest.ini_options]` block with `unit`, `integration`, `smoke`, `kitless` by hand.
- All 14 environment smoke tests died on `ModuleNotFoundError: No module named 'env_test_utils'`. `_run_environments` lives in `source/isaaclab_tasks/test/env_test_utils.py`, which isn't packaged and isn't importable from an installed `isaaclab`. I vendored a copy — it only depends on `isaaclab` and `isaaclab_tasks`, so it works, but it's now a fork that will drift.

Suggest the template gain a `[tool.pytest.ini_options]` marker block (cheap, and it makes the markers discoverable), and that the doc say a sentence about tests either way. The `env_test_utils` one is a bigger call — either promote it to a real module like `isaaclab_tasks.test_utils`, or say plainly that external projects are expected to write their own environment harness.

Unrelated to tests but in the same file, one more per-file-ignore would help: the template sets `select = ["E", "F", "I", "UP", "W"]` and ignores `**/__init__.py` but not `**/__init__.pyi`, so Isaac Lab's own lazy-export idiom (`lazy_export()` + a `.pyi` listing `__all__`) fails `F403` in a generated project out of the box.

AntoineRichard:
Splitting this out of my testing comment above, because it's the one thing in that bundle the doc can't fix on its own.

External projects can't import the environment-test helpers. `source/isaaclab_tasks/test/` is a sibling of the package directory, and packaging includes only `["isaaclab_tasks", "isaaclab_tasks.*"]` — correct for test files, nobody wants the suite in the wheel, but it takes the reusable helpers down with it. In-tree the gap is bridged by a `sys.path.insert` in `test/conftest.py`, which of course doesn't travel.

Lived consequence in the port: all 14 environment smoke tests in my generated project failed at collection with `ModuleNotFoundError: No module named 'env_test_utils'`. I vendored a copy of the 429-line file — it only needs `isaaclab` and `isaaclab_tasks`, so it works — and that copy is now a fork that will drift.

I've opened #7488 proposing the helpers move to an installed path (e.g. `isaaclab_tasks.testing`), with test files themselves staying unshipped.

For this PR, either resolution seems fine: point the doc at a packaged home if #7488 lands, or say plainly that projects are expected to vendor `env_test_utils.py` and own the drift. Both are a sentence — what's costly is the current silence, which leaves people to discover it at collection time.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed in 73db28d19. The introduction now leads with why the generator exists: it bootstraps package structure, task registration, and agent configurations for either a standalone project or an upstream task.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed in 73db28d19. The page now opens with the generator purpose and explains External versus Internal before walking through either flow.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed in 73db28d19. Cartpole is now introduced only after the general purpose, as the replaceable working example included by both generator modes.

docs/source/developer-tools/template_generator.rst
StafaH: Clarified in 73db28d19. This now sits next to the command and explicitly says the generator uses dependencies from the active Isaac Lab environment; it applies while creating an external project as well.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed in 73db28d19. The External/Internal comparison and workflow choices now come first, followed by an explicitly named external-project walkthrough.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed in 73db28d19. I replaced the backend matrix with a short note about the generated extras, one example, and links to Backends and Presets plus the quickstart combinations.

docs/source/developer-tools/template_generator.rst
StafaH: Addressed by keeping the internal-task procedure later in the page, after the external-project structure and development workflow. The mode comparison itself is now at the beginning.

StafaH:
Thanks for testing this downstream. I addressed the follow-up findings:

- [Generated tree and UI stub](https://github.com/isaac-sim/IsaacLab/pull/7480#issuecomment-5511218421): the tree now includes the extension docs, Python module, UI stub, task-family directory, MDP package, configuration, and agents. The guide and generated README explain how to remove the optional UI stub and its second `[[python.module]]` entry.
- [Task organization](https://github.com/isaac-sim/IsaacLab/pull/7480#issuecomment-5511218740): `config/cartpole` is now described as one common shape rather than a required robot hierarchy, behavior-based families are supported explicitly, and cross-family code can live under `tasks/mdp` without being treated as a registration package.
- [Testing and Ruff defaults](https://github.com/isaac-sim/IsaacLab/pull/7480#issuecomment-5511218992): generated projects now include pytest in the dev group, register `unit`, `integration`, `smoke`, and `kitless`, and exempt `**/__init__.pyi` from `F401`/`F403` like `__init__.py`.
- [Environment test helpers](https://github.com/isaac-sim/IsaacLab/pull/7480#issuecomment-5511332082): the guide and generated README now state that `source/isaaclab_tasks/test` is not packaged. External projects should own local fixtures built from public APIs; copying `env_test_utils.py` is vendoring and requires tracking upstream until #7488 provides a packaged API.

These changes are in ac6aa7a00, with the remaining guide flow feedback refined in 73db28d19. I considered the skill suggestion but left it out: the generator is already the interactive workflow, and keeping its behavior in the Developer Tools guide avoids a second source of truth.

I also fixed the Python dependency-license failure in e63055503. The workflow still passed the deleted `tools/template/requirements.txt` to `uv pip install`; the root `uv sync` already installs the declared template dependencies, so the stale requirements argument is now removed.

StafaH:
Confirmed the CI repair on the latest head: [Python `license-check` passed](https://github.com/isaac-sim/IsaacLab/actions/runs/33805155517/job/100814021447). The removed template requirements file is no longer referenced anywhere in the repository.

StafaH:
run-ci

StafaH:
Updated the template generator so our flagship choices are always option 1 when applicable:

- `Manager-based | single-agent` is the first workflow.
- `rsl_rl` is the first single-agent RL library.
- `PPO` is the first algorithm for every library that supports it.
- The workflow and RL-library comparison tables now use the same ordering as the prompts.
- Added an interactive-flow regression test that selects each first option and verifies all three choices.

Validated with `uv run python -m pytest --confcutdir=tools/template tools/template/test_cli.py -q` (5 passed) and the full formatting/lint suite (with only the known stale-baseline changelog hook skipped). The branch also includes the latest `develop` (`d2a630628`).

Implemented in `52e1a434e`.

StafaH:
Aligned the external project template with `IsaacLabTutorial` as the canonical downstream uv project in `a6706d31d`:

- Replaced the nested uv workspace/setuptools package with one root `uv_build` project and `src/<package>` layout.
- Separated project, task-family, and robot/config names. Generated task code now lives under `tasks/<task-family>/config/<robot>` with concise `env_cfg.py` / `env.py` filenames.
- Made the package root passive and kept task discovery on the `isaaclab.tasks` entry point.
- Added a generated `tests/test_registration.py` covering task IDs, environment entry points, and the default agent.
- Made the Isaac Sim UI extension opt-in (default: no UI), while retaining a working `config/extension.toml` and UI module when selected.
- Updated the generated README and Developer Tools guide to document the actual output and removed the obsolete workspace diagram/templates.
- Kept Manager-based, RSL-RL, and PPO as option 1.

Validation:

- `uv run python -m pytest --confcutdir=tools/template tools/template/test_cli.py -q` — 7 passed
- Generated manager-based and direct/UI projects both passed their registration tests and built wheels successfully.
- Generated projects passed Ruff check and format validation.
- `uv run --isolated --extra dev -- make -C docs current-docs` completed successfully.
- `SKIP=check-changelog-fragments uv run isaaclab -f` passed; the skipped local hook compares against the fork's stale `origin/develop` baseline.

The branch is also merged with the latest upstream `develop` at `8fbdc3852` (merge commit `1263fefee`).

StafaH:
run-ci

AntoineRichard (approved):
Thanks I think it's great!

AntoineRichard:
@StafaH do we want to backport this one? I would say yes. But it won't be QA tested?

AntoineRichard:
run-ci

StafaH:
run-ci

kellyguo11:
run-ci

hujc7:
The uv lock seems to be geneated by an older version. @StafaH

PR #7485, opened 2026-09-02
[Fix] Publish and fetch the checkpoints a task's components declare

hujc7:
## Summary

`play --checkpoint pretrained` crashed on the Shadow Hand camera tasks: the vision CNN the policy was
trained with was never published, so playback loaded a policy whose feature extractor did not exist.

The component that writes such a file now declares it on its own configuration. Publication, retrieval
and playback all read that one declaration, which the tooling finds by walking the resolved environment
config — so a task lists nothing.

# Description

## 1. Change

- `FeatureExtractorCfg` declares `checkpoint_name` and `checkpoint_glob`, and the extractor prefers the
  published copy in its log directory, falling back to the file a training run wrote.
- `get_declared_checkpoints(env_cfg)` returns every declaration reachable from a resolved config,
  matching on the declared field names rather than `hasattr` so an unrelated lazy config value is never
  resolved. `get_declared_checkpoint_path(...)` gives the path beside a policy checkpoint.
- The publish script collects each match beside the policy as `<policy stem>_<name><extension>` and
  refuses to publish a job whose declared checkpoint was not collected. A re-collect clears the previous
  copy first, so a run that writes no checkpoint cannot leave a stale one beside a fresh policy.
- The fetch downloads each file and records it on the configuration that declared it
  (`FeatureExtractorCfg.checkpoint_path`), so the component loads it without depending on the log
  directory its RL workflow derives — `rl_games` and `skrl` derive the policy's grandparent, a level
  above where the download lands. All four `play` entry points now make the same call. A policy
  published without a checkpoint its components declare raises at the fetch instead of warning and
  failing later inside the component.

No new public type: `isaaclab` is untouched.

## 2. Verification

Live QA on 4gpu in the CI image, against the checkpoints published on the asset server:

- Fetch: for both camera tasks the resolved config yields `{'feature_extractor': 'cnn_*.pth'}`, the
  policy downloads, and the declared file lands beside it (`exists=True`).
- Playback, the original repro:
  `isaaclab play --rl_library rsl_rl --task <task> --checkpoint pretrained physics=newton_mjwarp renderer=newton_renderer`
  Both tasks log `[INFO]: Loading feature extractor checkpoint from <cache>/..._feature_extractor.pth`,
  with no `Traceback` or `FileNotFoundError`.
- 132 unit tests across the touched packages, including regression tests for the hand-over, the stale
  re-collect and the incomplete pair, each verified to fail without its fix; `pre-commit` clean.

## Type of change

## Release backport

## Checklist

- [ ] Documentation needs no change: the declaration is component configuration, described in the changelog fragments


source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
```diff
@@ -219,16 +230,57 @@ def get_published_pretrained_checkpoint_path(
...
+    Returns:
+        The policy path with its workflow extension replaced by ``_<name>.pth``.
+
+    Raises:
+        ValueError: If the workflow is invalid.
+    """
+    if workflow not in WORKFLOW_PRETRAINED_CHECKPOINT_EXTENSIONS:
+        raise ValueError(f"Unsupported workflow: {workflow!r}")
+    extension = WORKFLOW_PRETRAINED_CHECKPOINT_EXTENSIONS[workflow]
+    return checkpoint_path.removesuffix(extension) + f"_{name}{AUXILIARY_CHECKPOINT_EXTENSION}"
```
hujc7: AUXILIARY_CHECKPOINT_EXTENSION is extention different? feel weird to define something general .pth as a global var rather than hard coding it. Or let define site just define it.
also the get_auxiliary_checkpoint_path fuction logic seems general? can it reuse pretain path logic?

source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
```diff
@@ -257,9 +315,39 @@ def get_published_pretrained_checkpoint(
             return None
     else:
         print("Using pre-fetched pre-trained checkpoint")
+    for name in auxiliary_names:
+        _retrieve_published_auxiliary_checkpoint(ov_path, workflow, name, download_dir, os.path.dirname(resume_path))
     return resume_path
 
 
+def _retrieve_published_auxiliary_checkpoint(
```
hujc7: at define site, it's auxiliary checkpoint, at retrieval, does it still need to differentiated?

source/isaaclab_tasks/isaaclab_tasks/core/reorient/config/shadow_hand/feature_extractor.py
```diff
@@ -200,9 +207,16 @@ def __init__(
             os.makedirs(self.log_dir)
 
         if self.cfg.load_checkpoint:
-            list_of_files = glob.glob(self.log_dir + "/*.pth")
-            latest_file = max(list_of_files, key=os.path.getctime)
-            checkpoint = os.path.join(self.log_dir, latest_file)
+            # a published checkpoint carries the suffix; a training run writes cfg.checkpoint_glob
+            list_of_files = glob.glob(os.path.join(self.log_dir, f"*_{self.cfg.checkpoint_name}.pth")) or glob.glob(
+                os.path.join(self.log_dir, self.cfg.checkpoint_glob)
```
hujc7: not clear to me. so there's pretrain and local?
for model weight, is the checkpoint specially renmaed? or it has native name? and when both are included, which wins the max later?

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
@StafaH @ooctipus @AntoineRichard for review.

hujc7:
run-ci

source/isaaclab/isaaclab/utils/checkpoints.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def __post_init__(self) -> None:
+        if (self.run_glob is None) == (self.url is None):
+            raise ValueError(
+                f"The {self.name!r} checkpoint must declare exactly one of run_glob and url,"
+                f" got run_glob={self.run_glob!r} and url={self.url!r}."
+            )
+
+    @property
+    def is_run_artifact(self) -> bool:
```
ooctipus: lets not add properties and method into config class, config class should state data and __post_init__ only

ooctipus:
Thanks for addressing the missing companion checkpoint problem. The behavioral goal is right: a published policy should include every file required to play it.

After looking at this together with #7630, #7594, Antoine's #7535 and the remote neural-actuator fix in #6839, I think we can simplify the architecture and avoid introducing `Checkpoint` or `CheckpointCfg`.

The common abstraction across policies, actuator networks, feature extractors, VAEs, local files, W&B, and pretrained S3 assets is resolution behavior, not additional config data. Existing component-specific string fields such as `network_file`, `model_path`, or `checkpoint` can already hold a local path or source reference. A one-field config wrapper would not add useful information.

In the current class, the fields actually describe three different lifecycles:

- `run_glob` is producer/output-discovery metadata.
- `url` is one kind of consumer source.
- `local_path` is mutable runtime resolution state.

Those are orthogonal rather than mutually exclusive. For example, a component can resume from W&B while producing a new local checkpoint. Mutating `local_path` on the resolved environment config also makes the config depend on where it happened to be materialized.

I suggest this ownership instead:

1. Components keep their normal string checkpoint fields.
2. One external resolver accepts a string reference. Local, Nucleus, HTTP, and S3 references delegate to the existing `retrieve_file_path()`; W&B resolves an exact artifact/file; a pretrained reference resolves a member through the published bundle manifest.
3. Resolution returns a local path without mutating the environment config.
4. Training records the exact files it produced in the run manifest, for example `{"policy": "model_1000.pt", "feature_extractor": "cnn_50000.pth"}`, instead of publication rediscovering outputs by recursively finding config objects and globbing.
5. Publishing consumes that manifest and makes the files visible as one complete bundle, with a manifest or commit marker published last.

That gives the related work one owner each: #7594/#7630 choose the preset-qualified policy bundle, #7532 chooses the compatible agent config, #7535 supplies the fetch error contract, #6839 establishes reuse of the shared asset cache, and this PR adds exact companion membership and bundle publication. It avoids creating multiple checkpoint representations for the same underlying problem.

There are also a few correctness cases worth addressing while reshaping this:

- `_build_core_jobs` reads declarations from the default env config before applying physics, renderer, and domain presets, so the bundle may not describe the config actually trained.
- `collect_pretrained_checkpoint` includes `preset_names` when selecting the policy but omits them when selecting the run for companions, which can pair files from different runs.
- A missing declared companion currently only warns and still returns the policy, even though that bundle cannot be played.
- Publication overwrites the policy before its companions. A later upload failure can expose a new policy with missing or stale members.

I would add architecture tests proving that an actuator network and a feature extractor can use the same external resolution path with plain string fields, that configs remain unchanged after resolution, that the exact preset-resolved job config is used, and that duplicate or missing bundle members fail before a bundle becomes visible.

The stacked #7509 still preserves `Checkpoint(run_glob, url, local_path)`, so I would prefer to settle this ownership boundary here before doing that refactor.

source/isaaclab/isaaclab/utils/checkpoints.py
```diff
@@ -0,0 +1,100 @@
...
+import glob
+import os
+from dataclasses import MISSING
+
+from isaaclab.utils.assets import retrieve_file_path
+from isaaclab.utils.configclass import configclass
+
+
+@configclass
+class Checkpoint:
```
ooctipus: this checkpoint class might have overfit to the pretrained checkpoint structure,

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

source/isaaclab_rl/isaaclab_rl/entrypoints/backends/play_rl_games.py
```diff
@@ -122,8 +121,7 @@ def main():
             log_root_path = os.path.abspath(log_root_path)
             print(f"[INFO] Loading experiment from directory: {log_root_path}")
             if args_cli.checkpoint == "pretrained":
-                backend_names = get_pretrained_checkpoint_backend_names(env_cfg)
-                resume_path = get_published_pretrained_checkpoint("rl_games", train_task_name, *backend_names)
+                resume_path = get_published_pretrained_checkpoint("rl_games", train_task_name, env_cfg=env_cfg)
```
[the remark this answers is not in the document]
hujc7: Fixed in 84876b82eca. The fetch now records each companion's downloaded path in `Checkpoint.local_path`, and `Checkpoint.resolve` returns it before searching a directory, so no entrypoint derives a directory for a companion any more. The rl_games and skrl `dirname(dirname(resume_path))` shape is unchanged but is no longer reachable for companion lookup.

source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
```diff
@@ -250,51 +301,69 @@ def get_published_pretrained_checkpoint(
...
-    download_dir = os.path.join(".pretrained_checkpoints", workflow)
-    if physics_backend is None:
-        download_dir = os.path.join(download_dir, task_name)
-    resume_path = os.path.join(download_dir, filename)
-
-    if not os.path.exists(resume_path):
-        print(f"Fetching pre-trained checkpoint : {ov_path}")
+    # one cache directory per published checkpoint: play treats it as the run log directory and
+    # writes videos, exported policies, and additional checkpoints into it
+    download_dir = os.path.join(
```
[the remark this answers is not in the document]
hujc7: Same root cause as the RL-Games comment, fixed in 84876b82eca. Rather than mirroring each workflow's nesting, the fetch hands the component its file directly: each declaration records its own `local_path`, which `resolve` prefers. The docstring that promised a log-directory lookup was reworded accordingly.

source/isaaclab_rl/isaaclab_rl/utils/pretrained_checkpoint.py
```diff
@@ -250,51 +301,69 @@ def get_published_pretrained_checkpoint(
...
-            )
-            print(
-                "A pre-trained checkpoint is currently unavailable for this task.\n"
-                f"  The asset server does not provide '{ov_path}'.\n"
-                f"  Either no checkpoint is published for task '{task_name}'{backends}, or the asset"
-                " server could not be reached.\n"
-                "  Train the task, or pass --checkpoint <path> to use a checkpoint of your own."
-            )
-            return None
+            print(f"[WARNING]: The asset server does not provide the {checkpoint.name} checkpoint '{declared_path}'.")
```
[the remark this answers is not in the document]
hujc7: Fixed on the publication side in 84876b82eca: `publish_pretrained_checkpoint` now fails the job when a declared checkpoint was not collected, instead of publishing the policy alone, so an unusable bundle is never advertised. The fetch still warns rather than failing, since a bundle that was already published incomplete is not something the fetching side can repair; the component's own `resolve` raises with the directory it searched.

source/isaaclab/isaaclab/utils/checkpoints.py
```diff
@@ -0,0 +1,91 @@
...
+    run_glob: str | None = None
+    """Glob, relative to the training run directory, matching the file this run writes."""
+
+    url: str | None = None
+    """Published location of pre-existing weights."""
+
+    @property
+    def is_run_artifact(self) -> bool:
+        """Whether this run produces the file, as opposed to fetching a published one."""
+        return self.run_glob is not None
```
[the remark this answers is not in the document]
hujc7: Fixed in 84876b82eca. `__post_init__` raises `ValueError` unless exactly one of `run_glob` and `url` is set, naming both values, with a parametrized test covering neither and both.

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

StafaH (commented):
Thanks for fixing this! The separate cache directory and clearer error are nice improvements.

The config discovery feels a bit heavy for this one use case. Could we just explicitly fetch the CNN alongside the policy and pass its path to the extractor? I’d rather keep this simple and generalize when we have another use case. The config seems over-engineered for this specific case.

Also, looks like recollection could leave an old CNN beside a new policy, and downloading a missing CNN only warns. Can we make sure the pair is complete before publishing or loading?

hujc7:
run-ci

source/isaaclab_rl/isaaclab_rl/entrypoints/backends/play_rsl_rl.py
```diff
@@ -140,8 +139,7 @@ def main(env_cfg: ManagerBasedRLEnvCfg | DirectRLEnvCfg | DirectMARLEnvCfg, agen
             log_root_path = os.path.abspath(log_root_path)
             print(f"[INFO] Loading experiment from directory: {log_root_path}")
             if args_cli.checkpoint == "pretrained":
-                backend_names = get_pretrained_checkpoint_backend_names(env_cfg)
-                resume_path = get_published_pretrained_checkpoint("rsl_rl", train_task_name, *backend_names)
+                resume_path = get_published_pretrained_checkpoint("rsl_rl", train_task_name, env_cfg=env_cfg)
```
kellyguo11, line 142: why was this only needed for RSL RL?

hujc7:
run-ci

StafaH (approved):

StafaH:
run-ci

StafaH:
Closing as the issue is resolved by #7695

PR #7500, opened 2026-09-02
Bug Fixes for Mimic-Cosmos Workflows

shauryadNv:
# Description

This PR bundles four fixes for issues found:

  1. 6684416: **RTX renderer: fix crash on not-yet-warmed-up annotator buffers.** `IsaacRtxRenderer.render()` unconditionally trimmed the tiled buffer to 2/3 channels for `motion_vectors`/`normals`/`SIMPLE_SHADING_MODES`/`RGB_HDR`. Immediately after an annotator is attached (e.g. at env creation, before the RTX renderer has pumped a frame), Replicator can momentarily return a buffer whose channel dimension is 0. Warp's array slicing rejects trimming an already-empty dimension (unlike NumPy, which allows it), raising `RuntimeError: Invalid indexing in slice: 0:0:1`. Guard the trim and skip writing that data type for the frame instead of crashing.
  2. 6683697: **`robust_eval.py`/`play.py`: pass `enable_cameras=True` explicitly.** `release/3.0.0` removed the `--enable_cameras` CLI flag and the `ENABLE_CAMERAS` env-var fallback from `AppLauncher` in favor of `launch_simulation()`'s scene-scan auto-detection, but these two robomimic scripts still use the legacy `AppLauncher(args_cli)` pattern and were never migrated. Without cameras enabled, and with the old explicit "pass --enable_cameras" guard in `IsaacRtxRenderer.__init__` also removed in this release, the failure now surfaces as an opaque `ValueError: Invalid object in Py_Graph in getWrappedGraphFromNode` deep inside OmniGraph/SyntheticData. Pass `enable_cameras=True` explicitly, matching the pattern already used in `generate_dataset.py`.
  3. 6683610: **Docs: add `uv run` examples for the HDF5/MP4 conversion and merge tools.** The Augmented Imitation Learning doc recommends `uv` as the primary workflow throughout (dataset generation, training, eval all show a "uv (Recommended)" tab), but the `hdf5_to_mp4.py`, `mp4_to_hdf5.py`, and `merge_hdf5_datasets.py` examples only showed a bare `python ...` invocation with no indication of which environment/extras to use. Added matching `uv (Recommended)` / `isaaclab.sh` tab-sets (`--extra mimic`, since these tools only need `h5py`/`opencv`/`numpy`, no Isaac Sim import). 
  4. 6702683: **RTX renderer: fail fast on oversized tiled camera buffers.** `IsaacRtxRenderer.render()` flattens every environment's camera tile into one Warp array per data type. Warp requires every array dimension to fit in a signed 32-bit int, so a large enough `num_envs * resolution` combination overflows that limit and crashes deep inside `render()`'s `.flatten()` call with `ValueError: Array shapes must not exceed the maximum representable value of a signed 32-bit integer, got 2621440000 in dimension 0` and no indication of what to change. Compute the worst-case tiled buffer size upfront in `create_render_data()` and raise a clear error naming the offending env count/grid/resolution and suggesting to reduce `--num_envs` or camera resolution. This does not lift the underlying Warp/RTX limit — it only replaces the opaque failure with an actionable one.
  
## Type of change

- Documentation update

## Release backport

## Checklist

### DIFF
--- a/docs/source/overview/imitation-learning/augmented_imitation.rst
+++ b/docs/source/overview/imitation-learning/augmented_imitation.rst
@@ -93,11 +93,23 @@ The ``hdf5_to_mp4.py`` script converts camera frames stored in HDF5 demonstratio
 
 Example usage for the cube stacking task:
 
-.. code:: bash
+.. tab-set::
 
-    python scripts/tools/hdf5_to_mp4.py \
-    --input_file datasets/mimic_dataset_1k.hdf5 \
-    --output_dir datasets/mimic_dataset_1k_mp4
+   .. tab-item:: uv (Recommended)
+
+      .. code:: bash
+
+          uv run --extra mimic python scripts/tools/hdf5_to_mp4.py \
+          --input_file datasets/mimic_dataset_1k.hdf5 \
+          --output_dir datasets/mimic_dataset_1k_mp4
+
+   .. tab-item:: isaaclab.sh / isaaclab.bat
+
+      .. code:: bash
+
+          ./isaaclab.sh -p scripts/tools/hdf5_to_mp4.py \
+          --input_file datasets/mimic_dataset_1k.hdf5 \
+          --output_dir datasets/mimic_dataset_1k_mp4
 
 .. _running-cosmos:
 
@@ -251,12 +263,25 @@ The ``mp4_to_hdf5.py`` script converts the visually augmented MP4 videos back to
 
 Example usage for the cube stacking task:
 
-.. code:: bash
+.. tab-set::
 
-    python scripts/tools/mp4_to_hdf5.py \
-    --input_file datasets/mimic_dataset_1k.hdf5 \
-    --videos_dir datasets/cosmos_dataset_1k_mp4 \
-    --output_file datasets/cosmos_dataset_1k.hdf5
+   .. tab-item:: uv (Recommended)
+
+      .. code:: bash
+
+          uv run --extra mimic python scripts/tools/mp4_to_hdf5.py \
+          --input_file datasets/mimic_dataset_1k.hdf5 \
+          --videos_dir datasets/cosmos_dataset_1k_mp4 \
+          --output_file datasets/cosmos_dataset_1k.hdf5
+
+   .. tab-item:: isaaclab.sh / isaaclab.bat
+
+      .. code:: bash
+
+          ./isaaclab.sh -p scripts/tools/mp4_to_hdf5.py \
+          --input_file datasets/mimic_dataset_1k.hdf5 \
+          --videos_dir datasets/cosmos_dataset_1k_mp4 \
+          --output_file datasets/cosmos_dataset_1k.hdf5
 
 Pre-generated Dataset
 ^^^^^^^^^^^^^^^^^^^^^
@@ -291,11 +316,23 @@ The ``merge_hdf5_datasets.py`` script combines multiple HDF5 datasets into a sin
 
 Example usage for the cube stacking task:
 
-.. code:: bash
+.. tab-set::
 
-    python scripts/tools/merge_hdf5_datasets.py \
-    --input_files datasets/mimic_dataset_1k.hdf5 datasets/cosmos_dataset_1k.hdf5 \
-    --output_file datasets/mimic_cosmos_dataset.hdf5
+   .. tab-item:: uv (Recommended)
+
+      .. code:: bash
+
+          uv run --extra mimic python scripts/tools/merge_hdf5_datasets.py \
+          --input_files datasets/mimic_dataset_1k.hdf5 datasets/cosmos_dataset_1k.hdf5 \
+          --output_file datasets/mimic_cosmos_dataset.hdf5
+
+   .. tab-item:: isaaclab.sh / isaaclab.bat
+
+      .. code:: bash
+
+          ./isaaclab.sh -p scripts/tools/merge_hdf5_datasets.py \
+          --input_files datasets/mimic_dataset_1k.hdf5 datasets/cosmos_dataset_1k.hdf5 \
+          --output_file datasets/mimic_cosmos_dataset.hdf5
 
 Model Training and Evaluation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/overview/imitation-learning/augmented_imitation.rst
+++ b/docs/source/overview/imitation-learning/augmented_imitation.rst
@@ -54,6 +54,27 @@ requires.
 Cosmos Augmentation
 ~~~~~~~~~~~~~~~~~~~
 
+.. important::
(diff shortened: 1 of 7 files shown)


source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -83,6 +83,32 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
...
 SIMPLE_SHADING_MODE_SETTING = "/rtx/minimal/mode"
 
+# Warp arrays require every dimension to fit in a signed 32-bit int (see
+# ``warp.types.array.__getitem__``/``check_array_shape``). ``render()`` flattens the whole
+# tiled buffer (all environments' camera tiles concatenated) into one Warp array per data
+# type, so this bounds how large ``num_envs * tile pixels`` can get before that call raises.
+_WARP_MAX_ARRAY_DIM = 2_147_483_647
+
+
+def _validate_tiled_buffer_size(spec: CameraRenderSpec) -> None:
```
huidongc: This early validation doesn't feel like Isaac RTX specific. If I understand it correctly the limitation is from the max number of entries in the warp array, so it seems we should have this validation in a common places?

Also the assumption of 4 channels per pixel is not correct, which will make the validation over conservative for other cases.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
huidongc: If the flatten() call is the culprit, we should better seek to refactor the `reshape_tiled_image` kernel so that it takes the 3d arrays `(tile_res_w * tile_num_w, tile_res_h * tile_num_h, C)` and outputs the 4d arrays `(N, H, W, C)`.

See the suggestion here: https://nvidia.github.io/warp/stable/user_guide/limitations.html#arrays

This way we don't have to reduce the number of envs and camera resolutions.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -83,6 +83,32 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
 }
 SIMPLE_SHADING_MODE_SETTING = "/rtx/minimal/mode"
 
+# Warp arrays require every dimension to fit in a signed 32-bit int (see
+# ``warp.types.array.__getitem__``/``check_array_shape``). ``render()`` flattens the whole
+# tiled buffer (all environments' camera tiles concatenated) into one Warp array per data
+# type, so this bounds how large ``num_envs * tile pixels`` can get before that call raises.
+_WARP_MAX_ARRAY_DIM = 2_147_483_647
```
huidongc: This is not a contract but an implementation detail in Warp that may probably get changed in future version of warp and cause Isaac Lab to drift from the actual limitation.

We should try to tackle this problem from another angle, see my comment at L624.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
```diff
@@ -83,6 +83,32 @@ def _raise_missing_ppisp_error(exc: ModuleNotFoundError) -> NoReturn:
...
+    """Raise a clear error if this camera's tiled buffer would overflow Warp's array shape limit.
+
+    Worst case is 4 elements per pixel: RGBA/HDR annotators return 4 channels directly, and
+    colorized segmentation annotators reinterpret a uint32 id as 4 uint8 channels.
+    """
+    num_tile_cols = math.ceil(math.sqrt(spec.view_count))
+    num_tile_rows = math.ceil(spec.view_count / num_tile_cols)
+    tiled_pixels = (num_tile_cols * spec.cfg.width) * (num_tile_rows * spec.cfg.height)
+    max_elements_per_pixel = 4
+    if tiled_pixels * max_elements_per_pixel > _WARP_MAX_ARRAY_DIM:
```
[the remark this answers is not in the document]
shauryadNv: Removed check, irrelevant now.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
shauryadNv: Fixed.

source/isaaclab_physx/isaaclab_physx/renderers/isaac_rtx_renderer.py
shauryadNv: Fixed.

kellyguo11:
run-ci

kellyguo11:
run-ci

kellyguo11:
run-ci

scripts/imitation_learning/robomimic/robust_eval.py
```diff
@@ -64,7 +64,9 @@
 args_cli = parser.parse_args()
 
 # launch omniverse app
-app_launcher = AppLauncher(args_cli)
+# Evaluation rollouts run camera-observation tasks and require an RTX renderer. Request
+# rendering support here so callers do not need a legacy CLI flag.
+app_launcher = AppLauncher(args_cli, enable_cameras=True)
```
kellyguo11: is it always required to enable rendering?

docs/source/overview/imitation-learning/augmented_imitation.rst
```diff
@@ -93,11 +93,23 @@ The ``hdf5_to_mp4.py`` script converts camera frames stored in HDF5 demonstratio
...
+
+          uv run --extra mimic python scripts/tools/hdf5_to_mp4.py \
+          --input_file datasets/mimic_dataset_1k.hdf5 \
+          --output_dir datasets/mimic_dataset_1k_mp4
+
+   .. tab-item:: isaaclab.sh / isaaclab.bat
+
+      .. code:: bash
+
+          ./isaaclab.sh -p scripts/tools/hdf5_to_mp4.py \
```
kellyguo11, line 131: latest isaac sim removed ffmpeg in the opencv build it ships with, we might need to add a note for users to explicitly install the full opencv package first before depending on ffmpeg

scripts/imitation_learning/robomimic/robust_eval.py
shauryadNv: Good catch, fixed it now.

docs/source/overview/imitation-learning/augmented_imitation.rst
shauryadNv: Thanks for pointing that out, have updated the docs.

shauryadNv:
run-ci

shauryadNv:
run-ci

shauryadNv:
run-ci

PR #7505, opened 2026-09-02
Clear self-collision filter pairs before finalizing shadow Newton model

matthewtrepte:
## Summary
- The PhysX-backend shadow Newton visualization model never runs collision detection, but USD-authored self-collision filter pairs (`physxArticulation:enabledSelfCollisions`) were still imported into it and replicated across every cloned env.
- At real training env counts, this filter-pair set could reach billions of entries, causing `ModelBuilder.finalize()` to run out of memory.
- Clears `builder.shape_collision_filter_pairs` before finalizing the shadow model since it has no use for them.

## Test plan
- [x] `uv run python -m pytest source/isaaclab/test/sim/test_newton_manager_visualization_state.py` — 23 passed, including new regression test `test_ensure_visualization_model_clears_shape_collision_filter_pairs_before_finalize`
- [x] `uv run python -m pytest source/isaaclab/test/sim/test_simulation_context_visualizers.py source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py` — 200 passed
- [x] `uv run isaaclab -f`

## Release backport


matthewtrepte:
run-ci

ooctipus:
Thanks for fixing this. I agree with the invariant: a Newton model used only for visualization/scene queries should not carry solver-only collision data. I think the current late clear is incomplete, though, because Newton generates contact pairs during the same finalization.

I reproduced this against both Newton 1.5.1 and the exact 1.6 development commit pinned by #7453 (`24bd863528d6b91137408930d0fbe8fa216ad962`):

```text
10 replicated filtered pairs, unchanged:       10 filters,  0 contacts
builder.shape_collision_filter_pairs = []:      0 filters, 10 contacts
clear filters + collision group 0:              0 filters,  0 contacts
```

At 1,024 worlds × 48 shapes, clearing alone generated 1,155,072 contact pairs. `ModelBuilder.finalize()` currently always calls `_find_shape_contact_pairs()`, so deleting an exclusion makes that pair eligible for the contact array. The new test mocks `finalize()`, which hides this transfer.

I suggest moving the policy into `build_visualization_builder_from_stage_envs()`, where we know every produced builder is scene-only, and applying it before composition/replication:

```python
for visual_builder in (global_builder, *source_builders.values()):
    visual_builder.shape_collision_filter_pairs = []
    visual_builder.shape_collision_group[:] = [0] * visual_builder.shape_count

builder = ModelBuilder(up_axis=up_axis)
builder.add_builder(global_builder)
replicate_builder_mapping(builder, sources, mapping, positions, quaternions, source_builders)
```

The standalone branch should apply the same two assignments before returning its builder. The fresh assembled builder is important: assigning `[]` replaces Newton's compact filter representation with a plain list; adding already-stripped component builders to a fresh builder keeps the destination compact-empty and preserves homogeneous contact-template finalization. Please retain assignment rather than `.clear()`, because reading the property can materialize the compact replicated filters.

Then `_ensure_visualization_model()` can return to simply finalizing the already-correct builder; the late clear and its comment go away. This keeps the generic cloner unchanged and adds no cfg field, helper, fallback, or public API.

For the regression, I would replace the mocked-ordering test with a small real CPU `ModelBuilder.finalize()` test covering replicated filtered shapes and asserting both:

```python
assert len(model.shape_collision_filter_pairs) == 0
assert model.shape_contact_pair_count == 0
```

Collision group 0 leaves shape flags and the BVH intact; I also verified the same Newton raycast hit and distance with groups 1 and 0. The existing raycaster suite should still be run, but a second large test is unnecessary.

This also ports cleanly to the planned ownership model: the same normalization belongs on source/global builders in the visualization-only `NewtonReplicateContext` branch before plan replication, while Newton physics resources retain collision data.

ooctipus:
Implemented the construction-boundary fix in `4716126cc`.

- Collision filters and groups are stripped from visualization-only component builders before composition/replication.
- A fresh destination builder preserves Newton's compact empty-filter representation.
- The manager-level late clear and mocked-order test are removed.
- A real in-memory USD/Newton finalization test now checks both standalone and replicated models have geometry but no filters or contact pairs.

Validation:
- 45 Newton cloner tests passed.
- 22 Newton manager/visualization-state tests passed.
- The regression test fails without the fix and passes on Newton 1.5.1 and the exact Newton 1.6 commit from #7453.

kellyguo11:
run-ci

AntoineRichard (approved):
LGTM thanks Matthew!

ooctipus (approved):

ooctipus:
run-ci

PR #7506, opened 2026-09-02
Fix Newton GL Pause Rendering desync and document Rerun Play/Pause limitation

matthewtrepte:
## Summary
- Fixed the Newton GL visualizer's "Pause Rendering" button not reflecting the paused state after pressing `Space`. Both controls now read/write the same underlying `_paused` flag, so the button label and `is_rendering_paused()` stay in sync regardless of whether rendering was paused via the button or the keyboard shortcut. Also clarified the on-screen control hint from "Space - Pause/Resume" to "Space - Pause/Resume Rendering".
- Documented that the Rerun visualizer's native Play/Pause timeline controls do not work while visualizing a live simulation/training run (they only scrub already-logged `.rrd` recordings). Isaac Lab's own **Pause Rendering** / **Reset Episode** ImGui controls should be used instead for the live view.

## Release backport

## Test plan
- [x] `uv run isaaclab -f` (ruff, ruff format, rst checks, changelog fragment check) — all passed
- [x] `uv run python tools/changelog/cli.py check develop` — passed
- [x] `uv run --extra test python -m pytest source/isaaclab_visualizers/test/test_newton_adapter.py -v` — 51/51 passed, including new regression test `test_newton_gl_viewer_rendering_pause_state_stays_in_sync_with_space_key` (verified it fails without the fix, passes with it)
- [x] Manually verified live in the Newton GL visualizer window: Space now flips the Pause/Resume Rendering button label immediately, matching a direct click

### DIFF
--- a/docs/source/overview/core-concepts/visualization.rst
+++ b/docs/source/overview/core-concepts/visualization.rst
@@ -758,6 +758,15 @@ the num of environments can be overwritten and decreased using ``--num_envs``:
 The FPS control in the Rerun visualizer UI may not affect the visualization frame rate in all configurations.
 
 
+**Rerun Visualizer Play/Pause Timeline Controls**
+
+The native Play/Pause timeline controls in the Rerun visualizer UI do not work while visualizing a live
+simulation or training run. They are hidden by default, but Rerun's dock panel UI can still be used to reveal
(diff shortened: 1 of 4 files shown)


kellyguo11 (approved):

kellyguo11:
run-ci

PR #7513, opened 2026-09-03
Resolve nested callable references in manager term configs

### LINKED ISSUE #6067: [Bug Report] V3.0.0-Beta: Observation ModifierCfg func is not resolved from ResolvableString before signature validation

ooctipus:
## Description

After a Hydra / `from_dict()` round-trip, callable references nested outside `ManagerTermBaseCfg.params` can remain lazy string values. Observation modifiers are stored in `ObservationTermCfg.modifiers`, so the existing parameter recursion never reaches `ModifierCfg.func` before `ObservationManager` performs class detection and signature validation.

This change extends the existing `ManagerBase._resolve_param_value` recursion instead of adding another resolution path:

- `_process_term_cfg_at_play` walks every manager-term field rather than only `params`
- `_resolve_param_value` traverses nested configuration dataclasses in addition to its existing dict, list, tuple, and nested-term cases
- string-valued `func` fields are resolved at that single point
- the root term uses the same resolver before its own signature validation
- production code does not inspect or depend on the `ResolvableString` type
- `ObservationManager`, `ModifierCfg`, and existing `params` forwarding behavior remain unchanged

The modifier documentation now reflects the existing contract: function modifiers use `params` for call-time keyword arguments, while stateful class modifiers keep constructor settings on dedicated `ModifierCfg` subclasses.

No second recursive walker, `class_type`, wrapper configuration, modifier dispatch branch, or callable-proxy introspection behavior is introduced.

## Release backport

- [x]  This PR already targets the active release branch; do not backport it again.

## Tests

- Added an end-to-end observation modifier round-trip regression using a dedicated stateful modifier config and empty call-time `params`
- Added a manager-level regression proving runtime traversal reaches term fields outside `params`
- 92 focused manager, modifier, config, dictionary, and string tests passed
- Formatting, lint, security, and RST checks passed
- Changelog validation passed against the current upstream `develop`

Fixes #6067


ooctipus:
run-ci

source/isaaclab/test/managers/test_observation_manager_unit.py
```diff
@@ -62,6 +86,32 @@ def __post_init__(self):
     policy: PolicyCfg = PolicyCfg()
 
 
+def test_stateful_modifier_cfg_roundtrip_preserves_func():
+    """A stateful modifier remains usable after its function becomes a lazy string."""
+    cfg = HistoryObservationsCfg()
+    cfg.policy.history_length = None
+    cfg.policy.dummy.modifiers = [StatefulBiasModifierCfg(value=2.0)]
```
StafaH: AI Review:

This test avoids the configuration that triggered #6067. The reported usage is ModifierCfg(func=MyModifier, params={...}), where params are read by the modifier constructor.

After this PR resolves and instantiates the class, mod_cfg.params remains populated. compute_group() then invokes the instance as:

obs = modifier.func(obs, **modifier.params)

For the reported __call__(self, data) implementation, this raises TypeError on the first observation computation. The dedicated config subclass and empty params used here hide that failure.

Please make the regression reproduce #6067 with a non-empty ModifierCfg.params and ensure constructor parameters are not forwarded to the constructed modifier’s __call__.

source/isaaclab/isaaclab/managers/manager_base.py
```diff
@@ -411,7 +411,11 @@ def _resolve_param_value(self, term_name: str, key: str | int, value: Any):
             self._process_term_cfg_at_play(f"{term_name}.{key}", value)
         elif isinstance(value, dict):
             for sub_key, sub_value in value.items():
-                self._resolve_param_value(f"{term_name}.{key}", sub_key, sub_value)
+                value[sub_key] = self._resolve_param_value(f"{term_name}.{key}", sub_key, sub_value)
         elif isinstance(value, (list, tuple)):
             for i, item in enumerate(value):
                 self._resolve_param_value(f"{term_name}.{key}", i, item)
+        elif not isinstance(value, type) and hasattr(value, "__dataclass_fields__") and hasattr(value, "__dict__"):
```
StafaH: Can we use explicit dataclass/config-class check? 

AI comment: The current duck-typing is difficult to read, skips slotted dataclasses, and attempts setattr() on frozen dataclasses

My comment: This type of hasattr setattr behaviour loses typing and adds misdirection that becomes unreadable

source/isaaclab/isaaclab/managers/manager_base.py
```diff
@@ -391,17 +391,17 @@ def _process_term_cfg_at_play(self, term_name: str, term_cfg: ManagerTermBaseCfg
...
+        # initialize class-based terms
         if inspect.isclass(term_cfg.func):
             term_cfg.func = term_cfg.func(cfg=term_cfg, env=self._env)
 
-    def _resolve_param_value(self, term_name: str, key: str | int, value: Any):
-        """Recursively resolve a single param value (SceneEntityCfg, nested term cfgs, dicts, lists)."""
+    def _resolve_param_value(self, term_name: str, key: str | int, value: Any) -> Any:
+        """Recursively resolve a value in a manager term configuration."""
+        if key == "func" and isinstance(value, str):
+            return string_to_callable(value)
```
StafaH: Could we avoid calling `string_to_callable()` directly from `ManagerBase`? Since `ResolvableString` already represents a lazy callable reference, it would be cleaner for it to expose a `resolve()` method and own the conversion logic.

That would let the manager resolve only explicit callable fields on known config types such as `ManagerTermBaseCfg` and `ModifierCfg`, instead of interpreting every string under a dictionary key named `"func"` as an import path. The latter can incorrectly affect arbitrary dictionaries in `params`, for example `{"func": "linear"}`.

Raw strings could still be supported on explicit `func` fields for backward compatibility. This keeps the recursive traversal generic while moving callable resolution to the abstraction that owns it.

StafaH (approved):

source/isaaclab/isaaclab/managers/manager_base.py
```diff
@@ -411,7 +411,11 @@ def _resolve_param_value(self, term_name: str, key: str | int, value: Any):
             self._process_term_cfg_at_play(f"{term_name}.{key}", value)
         elif isinstance(value, dict):
             for sub_key, sub_value in value.items():
-                self._resolve_param_value(f"{term_name}.{key}", sub_key, sub_value)
+                value[sub_key] = self._resolve_param_value(f"{term_name}.{key}", sub_key, sub_value)
         elif isinstance(value, (list, tuple)):
             for i, item in enumerate(value):
                 self._resolve_param_value(f"{term_name}.{key}", i, item)
```
StafaH: The recursive calls for lists and tuples discard the returned value, so this only works when children happen to be mutated in place. Any resolver that replaces a value—such as resolving a ResolvableString into a callable—will not update the sequence.

Could we assign resolved list elements back and rebuild tuples? This would make _resolve_param_value() a consistent recursive transformation instead of relying on mutation as a side effect:

```
elif isinstance(value, list):
    for i, item in enumerate(value):
        value[i] = self._resolve_param_value(f"{term_name}.{key}", i, item)
elif isinstance(value, tuple):
    value = tuple(
        self._resolve_param_value(f"{term_name}.{key}", i, item)
        for i, item in enumerate(value)
    )
```

source/isaaclab/isaaclab/managers/manager_base.py
StafaH: can be replaced with dataclasses.is_dataclass() and iterating dataclasses.fields(). That avoids traversing unrelated runtime attributes stored in __dict__, but the discarded sequence results are the more worthwhile comment.

StafaH (approved):

StafaH (commented):
Left some comments, but overall LGTM

ooctipus:
run-ci

ooctipus:
run-ci

PR #7466, opened 2026-09-01
[Docs/Workflow] Fix installation bugs/issues for conda/uv/venv

StafaH:
# Description

Downloaded Isaac Sim packages linked through `_isaac_sim` were still allowed to select conda, uv, or venv Python environments even though that workflow is no longer supported. The CLI warnings and launcher behavior also contradicted the updated installation documentation.

This PR:

- rejects downloaded Isaac Sim packages combined with conda, uv, or venv environments in the CLI and platform launchers;
- marks live Isaac Sim source builds so their supported uv workflow remains available;
- removes the nine orphaned installation include pages and cleans up stale references;
- documents the breaking workflow change in the 3.0 release notes and package changelog;
- synchronizes the installation skill guidance; and
- adds regression coverage for downloaded-package and source-build behavior.

No additional dependencies are required.

## Type of change

- Bug fix
- Breaking change
- Documentation update

## Release backport

- [x]  Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

Not applicable.

## Validation

- `uv run --frozen python -m pytest --confcutdir=source/isaaclab/test/cli source/isaaclab/test/cli/test_env_commands.py source/isaaclab/test/cli/test_install.py source/isaaclab/test/cli/test_misc_commands.py` — 34 passed
- `uv run --isolated --extra test -- make -C docs current-docs` — passed without warnings
- `uv run isaaclab -f` — passed
- `uv run --no-project python tools/skills/cli.py check` — passed
- `uv run --frozen python tools/changelog/cli.py check upstream-develop --include-worktree` — passed

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove the fix is effective
- [x] I have added a changelog fragment for the touched package
- [x] My name already exists in `CONTRIBUTORS.md`

PR #7521, opened 2026-09-03
[Fix] Allow virtual environments built on Isaac Sim's own Python

hujc7:
# Description

#7466 stopped downloaded Isaac Sim packages from running with conda, `uv`, or `venv` environments.
The restriction is right for the breakages it targets, but it tests whether *any* virtual environment
is active, while the property that actually matters is *which interpreter runs Kit's extension modules*.

A virtual environment does not introduce an interpreter. It adds a `site-packages` directory and
reuses the interpreter it was created from, so one created on the package's own Python runs that exact
binary. The failures the restriction exists for come from environments that ship their own interpreter
and native libraries — the conda workarounds already in the tree are the evidence: preloading torch's
`libgomp` and conda's `libstdc++` for `CXXABI_1.3.15` in `commands/envs.py`, and `_sanitized_conda_env()`
guarding against conda's "SRE mismatch" from a mismatched Python. None of those apply to a virtual
environment built on Isaac Sim's own interpreter.

This PR compares the interpreter against the Isaac Sim directory instead of testing for an active
environment, in both `isaaclab.sh` and the CLI. The check resolves the interpreter that will be
launched, falling back to `pyvenv.cfg`'s `home` when `bin/python` is a copy rather than a symlink.
Anything it cannot resolve stays rejected, so conda environments, foreign-interpreter virtual
environments, and environments with no readable `pyvenv.cfg` behave exactly as they do today.

`isaaclab.bat` keeps the broader check. Narrowing it needs junction resolution and case-insensitive
prefix matching that I cannot exercise on this platform, so I would rather leave it to someone who can
test it than ship untested path handling in a launcher.

This unblocks `uv venv --python _isaac_sim/kit/python/bin/python3`, and the container images, which are
exactly this shape: Kit's tree plus a virtual environment created on its interpreter.

## Type of change

## Release backport

## Screenshots

Not applicable.

## Validation

- `uv run --frozen python -m pytest source/isaaclab/test/cli/test_install.py source/isaaclab/test/cli/test_env_commands.py` — 32 passed, 1 skipped
- Full `source/isaaclab/test/cli/` — 272 passed, 1 skipped
- Both new accepting tests fail without the change and pass with it; the paired rejecting tests pass either way
- Every test added by #7466 still passes, including the launcher rejection that relies on a `VIRTUAL_ENV` with no `pyvenv.cfg`
- Built `Dockerfile.base`, `Dockerfile.curobo` and `Dockerfile.kitless` against this branch; `isaaclab.sh -p`, `isaaclab -p` and `import isaaclab` all resolve to the image's virtual environment
- `uv run isaaclab -f` — passed

## Checklist

### DIFF
--- a/isaaclab.sh
+++ b/isaaclab.sh
@@ -21,12 +21,29 @@ for arg in "$@"; do
     fi
 done
 
+# A virtual environment reuses the interpreter it was created from, so one created on the
+# package's own Python runs that exact binary and loads Kit's extensions unchanged. Only
+# environments that supply their own interpreter, such as conda, are rejected. Anything
+# unreadable counts as foreign so the check fails closed.
+venv_on_bundled_python=false
+if [ -n "$VIRTUAL_ENV" ] && [ -z "$CONDA_PREFIX" ] && [ -f "$VIRTUAL_ENV/pyvenv.cfg" ]; then
+    venv_home=$(sed -n 's/^[[:space:]]*home[[:space:]]*=[[:space:]]*//p' "$VIRTUAL_ENV/pyvenv.cfg" | head -1)
+    sim_root=$(readlink -f "$ISAACLAB_PATH/_isaac_sim" 2>/dev/null)
+    venv_home=$(readlink -f "$venv_home" 2>/dev/null)
+    if [ -n "$sim_root" ] && [ -n "$venv_home" ]; then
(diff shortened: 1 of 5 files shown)


hujc7:
Folding this into #7405, which already carries the same commit.

PR #7516, opened 2026-09-03
[Sensor] Fix ray caster path matching regressions

StafaH:
# Description

This PR fixes the two remaining ray-caster path-matching failures:

1. The ray-caster camera tutorial counted four sensors for two camera prims because `/World/Origin_.*/CameraSensor` allowed `.*` to cross path separators under whole-path regular-expression matching.
2. Legacy Newton multi-mesh ray casters could fail to associate tracked-target sites when registration used a regular-expression path but initialization resolved it to concrete prim paths.

The stale post-reset Newton sensor result reported in #7236 is now fixed at the sensor state boundary by #7523, which this branch inherits from `develop`. The earlier environment-loop synchronization workaround and its tests/changelog fragments have therefore been removed from this PR.

## Changes

- Restrict the tutorial environment wildcard to one path segment with `/World/Origin_[^/]+/CameraSensor`.
- Associate legacy Newton tracked-target site labels by configuration order instead of path-expression strings.
- Remove the obsolete `--enable_cameras` argument from the tutorial smoke case.
- Add regression coverage for regular-expression target resolution.

Fixes #6572

Related: #7236, resolved by #7523.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Release backport

- [x] Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

Not applicable.

## Validation

- `uv run python -m pytest source/isaaclab_newton/test/sensors/test_newton_raycast_sensor.py -vv` (18 passed, including the #7523 FK regressions in eager and CUDA-graph modes)
- `uv run python -m pytest source/isaaclab/test/app/test_standalone_scripts.py::test_commands_respect_script_launcher_capabilities -vv` (1 passed)
- Verified the ad-hoc tracked-target regression fails with the old string-key association and passes with this fix in eager and CUDA-graph modes.
- Verified the exact clone-plan topology from #6572 passes on the merged branch in eager and CUDA-graph modes.
- `ISAACLAB_CHANGELOG_BASE_REF=upstream-develop uv run isaaclab -f`

The full standalone Newton multi-object demo smoke could not launch in this checkout because the newly merged installer guard rejects combining the downloaded Isaac Sim package with the active Python virtual environment; it exited before application startup.

## Checklist

- [x] I have read and understood the contribution guidelines
- [x] I have run the pre-commit checks
- [x] I have made corresponding changes to the tutorial and smoke configuration
- [x] My changes generate no new warnings
- [x] I have added tests that prove the remaining fix is effective
[...]

PR #7523, opened 2026-09-03
[Newton] Refresh FK before ray-cast sensor reads

### LINKED ISSUE #7236: [Bug Report] In-step resets compute observations before forward kinematics: Newton ray-cast sensors report the pre-reset pose
### Describe the bug

On the Newton backend, observations computed after an **in-step reset** (timeout / termination / manual reset handled inside `ManagerBasedRLEnv.step()`) are built from **pre-reset forward kinematics**. Any ray-cast sensor attached to an articulation body (`RayCaster`, `RayCasterCamera`, and subclasses) publishes its pose — and therefore casts its rays / renders its depth image — from the body transform of the *previous* episode's final pose. The first observation of every episode is sampled metres away from where the robot actually is.

The direct `reset()` / `reset_to()` paths are fine; only the two reset branches inside `step()` are affected. `DirectRLEnv.step()` has the same gap.

Mechanism (commit 3c0694c):

- `ManagerBasedRLEnv.step()` (`source/isaaclab/isaaclab/envs/manager_based_rl_env.py`, ~L270 and ~L282): after `self._reset_idx(reset_env_ids)` and after `self._reset_idx(manual_reset_ids)` it goes straight to `command_manager.compute()` → `observation_manager.compute()`. There is no `scene.write_data_to_sim()` / `sim.forward()` in between.
- `ManagerBasedEnv.reset()` and `reset_to()` (`source/isaaclab/isaaclab/envs/manager_based_env.py`) *do* call `self.scene.write_data_to_sim(); self.sim.forward()` right after `_reset_idx`, with the comment `# update articulation kinematics` — so the requirement is known, and `step()` just misses it.
- The asset data layer protects itself: `ArticulationData._ensure_fk_fresh()` (`source/isaaclab_newton/isaaclab_newton/assets/articulation/articulation_data.py`, ~L131) lazily calls `SimulationManager.forward()` when a pose-dependent property is read after a manual write. That is why `robot.data.*` looks fresh after the reset.
- The sensor pose path does not: `_NewtonRayCasterPoseMixin.get_world_poses()` (`source/isaaclab_newton/isaaclab_newton/sensors/ray_caster/newton_raycast_sensor.py`, ~L168) reads the body transforms straight out of `NewtonManager.get_state_0()` with a warp kernel, no lazy-FK guard. `NewtonManager.forward()` (`source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py`, ~L577) documents that body poses are only recomputed from joint coordinates by `forward()` or the next `step`.

So within the reset step the articulation data and the sensors disagree about where the robot is, and the observation the policy gets as the first observation of the new episode mixes post-reset proprioception with a pre-reset height scan / depth image.

### Steps to reproduce

Stock task, stock sensor, no Isaac Sim needed. A 1 s episode forces a timeout reset inside `step()` every 50 control steps; the reset event jitters the spawn by ±1 m so the pre- and post-reset poses are far apart. After each reset step we compare `height_scanner.data.pos_w` (xy) with the robot base (the scanner's offset is z-only, so the normal value is 0).

<details>
<summary><code>repro_stale_reset.py</code></summary>

```python
"""Minimal repro: in-step resets leave ray-cast sensor poses at the PRE-reset body transforms.

Stock task, Newton backend, no Isaac Sim. A 1 s episode forces a timeout reset inside
`ManagerBasedRLEnv.step()` every 50 control steps; we compare the height_scanner's
published pose against the robot base right after each in-step reset.

Run:  python repro_stale_reset.py            (stock behaviour -> hundreds of mm stale)
      python repro_stale_reset.py --patched  (forward() after _reset_idx -> fixed)
"""
import argparse
import sys

import numpy as np


def main() -> None:
    from isaaclab.app import add_launcher_args
    from isaaclab_tasks.utils import setup_preset_cli

    parser.add_argument("--patched", action="store_true")
    parser.add_argument("--steps", type=int, default=160)
    add_launcher_args(parser)
    args, remaining = setup_preset_cli(parser, agent_library="rsl_rl")
    sys.argv = [sys.argv[0]] + remaining

    import gymnasium as gym
    import torch
    from isaaclab.app import launch_simulation
    from isaaclab.envs import ManagerBasedRLEnv
    from isaaclab_newton.physics import NewtonCfg
    from isaaclab_tasks.utils import resolve_task_config

    task_id = "Isaac-Velocity-Rough-G1"
    env_cfg, _ = resolve_task_config(task_id, "rsl_rl_cfg_entry_point", play_mode=True)
    if env_cfg.sim.physics is None:
        env_cfg.sim.physics = NewtonCfg()
    env_cfg.scene.num_envs = 1
    env_cfg.curriculum = None
    env_cfg.episode_length_s = 1.0  # timeout reset inside step() every 50 control steps
    env_cfg.terminations.base_contact = None
    # Large reset jitter so the pre- and post-reset base poses are far apart.
    env_cfg.events.reset_base.params["pose_range"]["x"] = (-1.0, 1.0)
    env_cfg.events.reset_base.params["pose_range"]["y"] = (-1.0, 1.0)

    if args.patched:
        orig = ManagerBasedRLEnv._reset_idx

        def _reset_idx(self, env_ids):
            orig(self, env_ids)
            # What reset()/reset_to() already do after _reset_idx, and step() does not.
            self.sim.forward()

        ManagerBasedRLEnv._reset_idx = _reset_idx

    with launch_simulation(env_cfg, args):
        env = gym.make(task_id, cfg=env_cfg).unwrapped
        robot = env.scene["robot"]
        scanner = env.scene["height_scanner"]
        env.reset()
        t = lambda v: v.torch if hasattr(v, "torch") else v
        action = torch.zeros((1, env.action_space.shape[-1]), device=env.device)
        with torch.inference_mode():
            for k in range(args.steps):
                _, _, terminated, truncated, _ = env.step(action)
                if bool((terminated | truncated).any()):
                    base = t(robot.data.root_link_pos_w)[0, :2].cpu().numpy()
                    scan = t(scanner.data.pos_w)[0, :2].cpu().numpy()
                    print(f"step {k + 1:>4} (in-step reset): |height_scanner.pos_w - base| = "
                          f"{np.linalg.norm(scan - base) * 1000:8.1f} mm", flush=True)
        env.close()


if __name__ == "__main__":
    main()
```
</details>

```
$ python repro_stale_reset.py
step   50 (in-step reset): |height_scanner.pos_w - base| =   1320.9 mm
step  100 (in-step reset): |height_scanner.pos_w - base| =   2244.5 mm
step  150 (in-step reset): |height_scanner.pos_w - base| =   1647.2 mm

$ python repro_stale_reset.py --patched     # forward() after _reset_idx, as reset() does
step   50 (in-step reset): |height_scanner.pos_w - base| =      0.0 mm
step  100 (in-step reset): |height_scanner.pos_w - base| =      0.0 mm
step  150 (in-step reset): |height_scanner.pos_w - base| =      0.0 mm
```

The `--patched` variant monkeypatches `ManagerBasedRLEnv._reset_idx` to append `self.scene.write_data_to_sim(); self.sim.forward()` — i.e. exactly what `reset()` already does.

### System Info

- Commit: 3c0694c (2026-08-16)
- Isaac Sim Version: not installed (kit-less Newton backend; `newton` 1.5.0, `mujoco-warp` 3.11.0, `warp-lang` 1.16.0)
- OS: Ubuntu 22.04.5, kernel 6.8.0
- GPU: NVIDIA GeForce RTX 5090 D
- CUDA: 13.0 (driver 580.142); torch 2.11.0+cu128; Python 3.12

### Additional context

- Impact on the stock tasks: every `core/velocity` task attaches `height_scanner = RayCasterCfg(prim_path="{ENV_REGEX_NS}/Robot/base", ...)`, so on Newton the first height scan of every training episode is taken around the previous episode's final position. For a per-step-updating `RayCaster` only that one observation is wrong; for a ray-cast camera with a refresh schedule and a history/backfill buffer the stale frame is copied into the whole history stack and persists until the next scheduled refresh (in our 10 Hz depth-camera setup that is up to ~0.8 s of policy input per episode).
- How we found it: the same policy walked down stairs markedly worse in our Isaac Lab port than in the mjlab original, and the divergence traced to episode starts. mjlab's `step()` calls `sim.forward()` (and re-senses) between `_reset_idx` and the observation compute, which is the behaviour the patch above restores.
- Related but distinct: #6394 reports stale *RTX camera images* from `reset()` with fabric enabled (render-side transforms). This report is about the Newton backend, ray-cast sensors, and the **in-step** reset path; `reset()` itself is correct here.
- Suggested fix (either): call `self.scene.write_data_to_sim(); self.sim.forward()` after both `_reset_idx` calls in `ManagerBasedRLEnv.step()` (and `DirectRLEnv.step()`), mirroring `reset()`; or give `_NewtonRayCasterPoseMixin.get_world_poses()` the same lazy-FK guard the asset data layer uses.

### Checklist

### Acceptance Criteria

- [ ] After an in-step reset, ray-cast sensor poses (`RayCaster.data.pos_w`, `RayCasterCamera.data.pos_w`) match the post-reset body transforms in the observation returned by that `step()` call (the repro prints 0.0 mm without the monkeypatch).
- [ ] Same for `DirectRLEnv.step()`.
- [ ] A regression test covering "sensor pose after in-step reset" on the Newton backend.

ooctipus:
# Description

Fixes #7236.

Newton ray-cast tasks read `NewtonManager.get_state_0().body_q` inside their graph-capturable query pipeline. After an in-step joint or root-state write, the reset masks are current but the derived `body_q` remains stale until forward kinematics runs. This made the first sensor observation after a reset use the previous pose.

This fixes the stale read at its ownership boundary instead of adding eager simulator synchronization to the RL environment loops:

- `NewtonManager._update_sensor_tasks()` obtains state through the guarded `get_state()` accessor before BVH refit and sensor graph capture/replay.
- The graph-captured raycast callback keeps using raw `get_state_0()` state, so `forward()` is never captured.
- The ray-caster's direct `get_world_poses()` accessor applies the same lazy-FK rule.
- The renderer's now-redundant state refresh is removed because the sensor-task scheduler owns freshness for both renderer and raycast consumers.

This is an alternative to the in-step synchronization part of #7516. Its regex and legacy tracked-target fixes are independent of this PR.

## Architecture and performance

The state access boundary owns FK freshness; environment stepping remains unaware of individual sensor requirements. The guard runs only when a sensor or renderer update is requested. Rendering does not gain an additional refresh because its existing call moved into the shared scheduler, and Newton's device-resident reset masks limit the actual FK work to invalidated worlds/articulations.

## Type of change

## Release backport

- [x]  This PR already targets the active release branch; do not backport it again.

## Screenshots

Not applicable.

## Validation

- Added observable regressions for sensor-data reads and direct pose-getter reads immediately after a carrier root-pose write, with no intervening simulation step or FK-sensitive asset getter. Both tests failed before the fix by exactly the authored displacement in eager and CUDA-graph modes.
- `uv run --frozen --extra test python -m pytest source/isaaclab_newton/test/sensors/test_newton_raycast_sensor.py -vv` (16 passed)
- `uv run --frozen --extra test python -m pytest source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py source/isaaclab/test/sim/test_newton_manager_visualization_state.py -q` (194 passed)
- `uv run --frozen isaaclab -f` (all checks passed)

## Checklist

- [x] No standalone documentation change is required; the ownership rule is documented at the access boundaries

### DIFF
--- a/source/isaaclab_newton/changelog.d/fix-raycaster-fk-sensitive-read.rst
+++ b/source/isaaclab_newton/changelog.d/fix-raycaster-fk-sensitive-read.rst
@@ -1,0 +1,4 @@
+Fixed
+^^^^^
+
+* Fixed Newton ray-caster updates reading stale carrier poses after joint or root state writes.
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -2658,10 +2658,10 @@ class NewtonManager(PhysicsManager):
 
     @classmethod
     def get_state(cls, scene_data_provider: SceneDataProvider | None = None) -> State:
-        """Get the current Newton state for visualization.
+        """Get the current Newton state with derived transforms refreshed.
 
-        Use this method from visualizers/renderers/video recorders that need a
-        backend-agnostic Newton ``State``. When the sim backend is PhysX this
+        Use this method from sensors, visualizers, renderers, and video recorders that need
+        a backend-agnostic Newton ``State``. When the sim backend is PhysX this
         refreshes the shadow ``_state_0.body_q`` from the live PhysX scene via
         :meth:`update_visualization_state` before returning, so callers never
         observe stale transforms. Under the Newton sim backend, pending
@@ -2708,12 +2708,13 @@ class NewtonManager(PhysicsManager):
 
     @classmethod
     def _update_sensor_tasks(cls, *names: str) -> None:
-        """Refit the shape and particle BVHs and run the requested scene-query tasks."""
+        """Refresh derived state, refit the BVHs, and run the requested scene-query tasks."""
         for name in names:
             if name not in cls._sensor_tasks:
                 raise KeyError(f"Newton sensor task '{name}' is not registered.")
 
-        state = cls.get_state_0()
+        # Resolve pending FK before entering the graph-capturable sensor pipeline.
+        state = cls.get_state()
         if state is not cls._sensor_state:
             cls._sensor_state = state
             cls._sensor_state_dirty = True
--- a/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
+++ b/source/isaaclab_newton/isaaclab_newton/renderers/newton_warp_renderer.py
@@ -551,8 +551,6 @@ class NewtonWarpRenderer(BaseRenderer):
     def render(self, render_data: RenderData):
         """Render and write to output buffers. See :meth:`~isaaclab.renderers.base_renderer.BaseRenderer.render`."""
 
-        # Refresh the shadow state under PhysX before the manager refits the BVH.
-        NewtonManager.get_state()
         if render_data.sensor_task_name is None:
             render_data.sensor_task_name = f"newton_warp_render:{id(render_data)}"
             NewtonManager._register_sensor_task(render_data.sensor_task_name, lambda: self._launch_render(render_data))
--- a/source/isaaclab_newton/isaaclab_newton/sensors/ray_caster/newton_raycast_sensor.py
+++ b/source/isaaclab_newton/isaaclab_newton/sensors/ray_caster/newton_raycast_sensor.py
@@ -166,7 +166,8 @@ class _NewtonRayCasterPoseMixin:
         )
 
     def get_world_poses(self: Any, indices=None) -> tuple[ProxyArray, ProxyArray]:
-        """Return world poses for legacy camera helpers."""
+        """Return current world poses after resolving pending FK."""
+        NewtonManager.get_state()
         self._update_newton_site_transforms(
             self._sensor_site_indices, self._newton_pose_w, self._newton_pos_w.warp, self._newton_quat_w.warp
         )
@@ -192,7 +193,7 @@ class _NewtonRayCasterPoseMixin:
         pos_buf: wp.array,
         quat_buf: wp.array,
     ) -> None:
-        """Update site transforms using the manager-bound model and state."""
+        """Update site transforms from manager state already refreshed by the caller."""
         model = NewtonManager.get_model()
         state = NewtonManager.get_state_0()
         wp.launch(
--- a/source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py
+++ b/source/isaaclab_newton/test/physics/test_newton_manager_abstraction.py
@@ -293,7 +293,7 @@ def test_sensor_task_builds_and_refits_bvhs_before_rendering(monkeypatch):
     """Shape and particle BVHs are built and refit before a render task runs."""
 
     state = object()
-    status = {"shape_refit": False, "particle_refit": False, "rendered": False}
+    status = {"state_refreshed": False, "shape_refit": False, "particle_refit": False, "rendered": False}
 
     class FakeModel:
         shape_count = 1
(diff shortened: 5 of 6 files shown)


StafaH (approved):

kellyguo11 (approved):

kellyguo11:
run-ci

ooctipus:
run-ci

PR #7524, opened 2026-09-03
Render the table in Franka cable golden images

mmichelis:
# Description

The Franka cable rendering test never made its table visible, so the Warp rasterizer goldens showed the robot and cable floating over the ground plane with no work surface. The table spawns with `visible=False` because the pose command's success visualizer normally draws it in its place, but the golden captures hide that visualizer to keep frames deterministic. Cloth and soft already compensate through `_configure_franka_camera_test_env_cfg`, but cable did not use that helper because it names its command and reset event `cable_pose` and `reset_cable` rather than `deformable_pose` and `reset_deformable`.

This PR:

- Generalizes `_configure_franka_camera_test_env_cfg` with `command_name` and `reset_event_name` parameters and routes `rendering_test_franka_cable` through it, replacing the duplicated reset-range block and the separate `_apply_franka_camera_golden_scene_overrides` call.
- Makes the table visible by setting `visual_material` and `visible` on the existing spawn instead of replacing the spawn with the failure marker config. The marker is built from the base `TABLE_SPAWN_CFG` and carries no `physics_material`, so assigning it wholesale would discard the cable table's `static_friction=0.01, dynamic_friction=0.01` override and silently change the physics the goldens capture. The in-place update is byte-identical to the previous behavior for cloth and soft.
- Regenerates the 32 `franka_cable` golden images.

The cable test deliberately does **not** hide the pose command's success visualizer, unlike cloth and soft. That marker turns out not to disturb the capture once the table itself is visible: all 32 goldens reproduce unchanged with and without hiding it.

The cable table's friction override is preserved. Comparing the resolved config before and after the helper runs, only `visible` changes:

| | `visible` | friction (static, dynamic) |
| --- | --- | --- |
| baseline (before this change) | `False` | `(0.01, 0.01)` |
| after this change | `True` | `(0.01, 0.01)` |
| wholesale marker copy (rejected approach) | `True` | `None` |

Validation, each test id in its own process on an RTX 5090:

- All 9 `franka_cable` ids covering `newton-isaacsim_rtx_renderer`, `newton-ovrtx_renderer` and `newton-newton_renderer` pass against the regenerated goldens.
- Cloth and soft were re-run against their **unchanged** goldens to confirm the shared helper change is inert for them: `franka_cloth` kit-based RTX, `franka_cloth` kitless Warp and `franka_soft` kitless Warp all pass, and no cloth or soft golden files are modified.

## Type of change

## Release backport

## Screenshots

The before/after is visible directly in the diff for `source/isaaclab_tasks/test/golden_images/franka_cable/newton-newton_renderer-rgb.png`, which GitHub renders as an image diff.

Previously that golden showed the robot and the blue cable over a bare ground plane, with the green and blue debug-vis axis markers drawn. It now shows the pink table under the cable, matching the corresponding `franka_cloth` golden, and the axis markers are gone because the shared helper sets `debug_vis=False`.

The `isaacsim_rtx_renderer` and `ovrtx_renderer` goldens change only marginally, since the success visualizer marker was already standing in for the table in those views.

## Checklist


nvsekkin (approved):

nvsekkin:
thanks for the PR, looks good to me. we probably do not need to re-baseline the ones that already has the table (for example: we don't need to re-baseline `source/isaaclab_tasks/test/golden_images/franka_cable/newton-isaacsim_rtx_renderer-simple_shading_full_mdl.png`)

kellyguo11:
run-ci

PR #7530, opened 2026-09-03
[Odin] Reject Newton Warp + simple shading presets on the cartpole camera tasks

AntoineRichard:
# Description

`Isaac-Cartpole-Camera` and `Isaac-Cartpole-Camera-Direct` expose the renderer and the camera
data type as two independent `presets=` selectors, so `newton_renderer` can be combined with
`simple_shading_constant_diffuse`, `simple_shading_diffuse_mdl`, or `simple_shading_full_mdl`.
The Newton Warp renderer does not publish those data types, so the combination is accepted at
config resolution and only fails later, in `Camera._create_buffers`, once the simulator has
already started:

```
ValueError: Renderer NewtonWarpRenderer does not support the following requested
data types: ['simple_shading_constant_diffuse']
```

This showed up as **72 failed rows in a benchmark sweep** (dispatch `20260901-153531`, image
built from `release/3.0.0` at `f88dbc59c82`): every failure was one of the two cartpole camera
tasks with `renderer=newton_renderer` and one of the three `simple_shading_*` presets, each
burning roughly 400 s of startup before raising. The same data-type presets completed normally
on `isaacsim_rtx` and `ovrtx` (107 completed rows), so the gap is specific to the Newton Warp
backend. `simple_shading_*` is an Isaac RTX shading-mode setting that OVRTX aliases to
`LdrColor`; Newton has no analogue.

This PR adds a `validate_config` hook to both cartpole camera env configs, mirroring the guards
`ShadowHandCameraEnvCfg` and `ReorientEnvCfg` (in `lift_env_cfg.py`) already have.
`configclass`'s `validate()` invokes it, so the combination now fails immediately at config
resolution with an actionable message instead of after the simulator has come up. It also means
task-discovery tooling that calls `env_cfg.validate()` (e.g. our benchmark harness) drops these
combinations from the matrix automatically, rather than scheduling 72 jobs that are known to fail.

Whether the Newton Warp renderer should gain `simple_shading_*` support is a rendering-team
decision; this PR does not prejudge it.

One note for reviewers: the allow-list here was re-derived from
`NewtonWarpRenderer.supported_output_types`, which publishes 10 kinds, rather than copied from
Shadow Hand. **Shadow Hand's and Lift's literal sets list only 7**, omitting `rgba`, `rgb_hdr`,
and `albedo`, so they currently false-reject e.g. `presets=newton_renderer,albedo`. This PR uses
the renderer's actual 10 so cartpole's `albedo` preset keeps working on Newton, which matches the
sweep (no cartpole `albedo` rows failed). Fixing the two stale sets is left out of scope here.

## Type of change

Behaviour change: `presets=newton_renderer,simple_shading_*` on the two cartpole camera tasks now
raises at config resolution instead of at env construction. It never produced a working run, so
nothing that used to train stops training. Migration guidance is in the changelog fragment: use
`presets=newton_renderer,rgb`, or keep the shading data types on an RTX backend with
`presets=isaacsim_rtx,simple_shading_full_mdl`.

The generated environment browser lists renderer presets and data-type presets as flat,
independent lists and does not encode combinations (`Isaac-Reorient-Cube-Shadow-Camera-Direct`,
which already has this guard, still lists both), so no regeneration is needed.

## Screenshots

N/A

## Checklist

- [x] I have made corresponding changes to the documentation (changelog fragment)
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file (changelog fragment; version files not edited per repo policy)

## Test evidence

New test: `source/isaaclab_tasks/test/core/test_cartpole_camera_presets.py` (12 cases). It
resolves the real configs via `resolve_presets` and calls `cfg.validate()`; no simulator is
launched. Both tasks, both directions: Newton + each `simple_shading_*` must raise;
`isaacsim_rtx`/`ovrtx` + `simple_shading_full_mdl` must still resolve; `newton_renderer,rgb` must
still resolve.

With the fix:

```
$ uv run --frozen --extra test python -m pytest \
    source/isaaclab_tasks/test/core/test_cartpole_camera_presets.py -q
12 passed in 3.23s
```

With the two config files reverted to their `develop` content (guard removed, test unchanged):

```
6 failed, 6 passed in 3.46s
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_constant_diffuse-direct]
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_constant_diffuse-manager]
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_diffuse_mdl-direct]
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_diffuse_mdl-manager]
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_full_mdl-direct]
FAILED ...::test_newton_renderer_rejects_simple_shading[simple_shading_full_mdl-manager]
E       Failed: DID NOT RAISE ValueError
```

Neighbouring suites (Shadow Hand presets, env-cfg import rules) plus this one: `102 passed`.
`uv run --frozen isaaclab -f` passes.

The GPU/Isaac Sim rendering suites (`test_rendering_cartpole*.py`,
`test_rendering_registered_tasks.py`) were not run here. Checked statically instead: their
`newton_renderer` parametrizations draw from `_NEWTON_WARP_DATA_TYPES` in
`rendering_test_utils.py`, which is a strict subset of the allow-list, and their
`simple_shading_*` cases pin `isaacsim_rtx`/`ovrtx`, so none of them hits the new guard.


kellyguo11:
The early failure is useful, but the renderer capability contract should not be copied into each task config. `NewtonWarpRenderer.supported_output_types()` is already the source of truth, and the existing Shadow Hand/Lift lists have demonstrably drifted. Could we expose a config-time capability query through the resolved renderer class and make camera/renderer compatibility validation generic? The runtime `Camera._create_buffers()` check should remain as a defensive guard.

_This comment was AI-generated._

ooctipus (approved):

ooctipus:
im fine with validate_config dsolution, but kelly has a good point

AntoineRichard:
run-ci

StafaH (approved):
Let's move the assert into the camera to avoid duplication at the task level! otherwise LGTM

AntoineRichard:
run-ci

StafaH:
run-ci

StafaH:
run-ci

PR #7440, opened 2026-08-30
Optimize shared camera image observations

StafaH:
## Summary

- move Kuka Allegro camera observations onto the shared image MDP term
- fuse image normalization with channel-layout conversion and infer single camera output types
- remove redundant camera, observation-manager, and clone work across camera tasks

## Performance

Measured with 4096 `Isaac-Lift-KukaAllegro-Camera` environments using Newton physics, the Newton renderer, and 128x128 RGB:

- shared image MDP: 1.84 ms -> 1.39 ms (~24% faster)
- full environment: ~215 ms/step before and after (renderer-bound; difference is within noise)

## Validation

- 160 targeted tests passed
- 2-environment Kuka Allegro Newton/newton-renderer smoke test passed
- formatting and code/style hooks passed

The repository-wide changelog hook remains blocked by a pre-existing modified immutable `isaaclab_ov` fragment and unrelated missing package fragments.

PR #7531, opened 2026-09-03
[Odin] Fix semantic_segmentation camera observations reaching the policy as integers

AntoineRichard:
## Description

Every `semantic_segmentation` camera row in benchmark dispatch `20260901-153531` (image built from `origin/release/3.0.0` at `f88dbc59c82`, `rsl_rl`, all core tasks) failed at the first training step — **60 failed rows** total, with:

```
RuntimeError: Input type (unsigned char) and bias type (float) should be the same
```

The failure is renderer-independent: `isaacsim_rtx` 18 rows, `ovrtx` 18, `newton_renderer` 24. The segmentation output reaches the feature extractor's first convolution still as an integer tensor.

### Root cause

Both Cartpole camera observation paths normalize only RGB-like and `depth` data types, so `semantic_segmentation` falls through unconverted:

- `source/isaaclab_tasks/isaaclab_tasks/core/cartpole/mdp/observations.py` (`CameraImageStack.__call__`)
- `source/isaaclab_tasks/isaaclab_tasks/core/cartpole/cartpole_direct_camera_env.py` (`CartpoleCameraEnv._get_observations`)

Both files are byte-identical between `release/3.0.0` and `develop`, so the bug is present on `develop` as-is.

### Why the fix belongs in the observation term

The renderer is producing exactly what its published contract says it should — `NewtonWarpRenderer.supported_output_types` deliberately emits RGBA `uint8` when colorized and a single `int32` id channel when not, *"matching the Isaac RTX / OVRTX contract so backend-independent consumers see the same dtype"*. Segmentation label ids are integers; making a renderer emit floats would break every consumer that reads ids (visualization, semantic id lookup, dataset export).

Nor does it belong in the feature extractor: the extractor's contract is "float32 image in", and the observation pipeline already owns dtype/layout normalization for every other camera data type via `normalize_camera_image`. Making the CNN defensively cast would paper over the same gap for every future task and duplicate logic that already exists.

So the fix goes where the gap is: the observation term, routed through the shared helper.

### The int32-vs-uint8 subtlety

Segmentation has **two** dtypes depending on `colorize_semantic_segmentation`:

- `colorize=True` (the `CameraCfg` default, and what the failing sweep used): RGBA `uint8`, 4 channels
- `colorize=False`: a single `int32` label-id channel

`normalize_camera_image` already handled the colorized `uint8` case correctly — it was simply never called for this data type. The non-colorized `int32` case was **not** handled: it fell through every branch and was returned unchanged, so a fix that only wired up the existing call would still break `colorize=False`.

The `int32` half is not demonstrated by the sweep (see the caveat below), but it is not opportunistic scope creep: the caller now dispatches on the data type alone, so the helper is the single place that decides what segmentation means, and leaving it to silently return `int32` unchanged would ship a fix that reads as complete while still crashing under `colorize=False`.

Handled by keying on the tensor dtype rather than on the `colorize` flag or a hardcoded uint8 assumption:

- In `normalize_camera_image`, non-`uint8` segmentation is cast to `float32`. Label ids carry no meaningful scale, so they are cast and **not** rescaled — applying `(x / 255) - mean` to label ids would be inventing semantics.
- In both Cartpole callers, the uint8 deferred-normalize fast path (which keeps the frame-stack ring buffer in `uint8` for cheaper per-step copies) is gated on `camera_data.dtype == torch.uint8`, so colorized segmentation rides it and `int32` label maps are normalized before entering the ring.

## Type of change

## Test evidence

Construction-only reproduction, no simulator needed. The bug fires when the extractor first sees an observation, so exercising the observation term directly on stub camera output is sufficient and much faster.

Extended `source/isaaclab/test/utils/test_images.py` with the `int32` case next to the existing colorized case, and dropped `semantic_segmentation` from the "unknown type passthrough" parametrization — that class asserts `out is src` under the heading "Unknown data_types return the input unchanged", and segmentation is no longer unknown. It would still pass by the accident that `.float()` on a float32 tensor returns self, so leaving it would have left the suite documenting the opposite of the new behaviour.

Added `source/isaaclab_tasks/test/core/test_cartpole_camera_observations.py` for the observation term itself; no sim-free test of it existed (the `test_rendering_cartpole*.py` neighbours need a renderer and golden images, and the `*_camera_presets.py` files only resolve configs). Two tests, each parametrized over `frame_stack` `[1, 2]` so both the immediate and the deferred-normalize branch are covered, and each asserting exact values rather than just dtype.

Both dtypes are covered: colorized `uint8` RGBA and non-colorized `int32`.

**Without the fix** (the three source files reverted to `origin/develop`, tests kept):

```
$ uv run --frozen --extra dev python -m pytest source/isaaclab/test/utils/test_images.py \
      source/isaaclab_tasks/test/core/test_cartpole_camera_observations.py -q
FAILED test_images.py::TestNormalizeCameraImageSegmentation::test_non_colorized_semantic_segmentation_is_cast_to_float[cpu]
FAILED test_images.py::TestNormalizeCameraImageSegmentation::test_non_colorized_semantic_segmentation_is_cast_to_float[cuda:0]
FAILED test_cartpole_camera_observations.py::test_colorized_segmentation_is_normalized_like_rgb[cpu-1]
FAILED test_cartpole_camera_observations.py::test_colorized_segmentation_is_normalized_like_rgb[cpu-2]
FAILED test_cartpole_camera_observations.py::test_colorized_segmentation_is_normalized_like_rgb[cuda:0-1]
FAILED test_cartpole_camera_observations.py::test_colorized_segmentation_is_normalized_like_rgb[cuda:0-2]
FAILED test_cartpole_camera_observations.py::test_non_colorized_segmentation_is_cast_to_float[cpu-1]
FAILED test_cartpole_camera_observations.py::test_non_colorized_segmentation_is_cast_to_float[cpu-2]
FAILED test_cartpole_camera_observations.py::test_non_colorized_segmentation_is_cast_to_float[cuda:0-1]
FAILED test_cartpole_camera_observations.py::test_non_colorized_segmentation_is_cast_to_float[cuda:0-2]
10 failed, 62 passed in 8.56s
```

Note that the pre-existing colorized-`uint8` helper test passes on `develop`: `normalize_camera_image` always handled that case correctly, and the crash came from the Cartpole callers never invoking it.

**With the fix:**

```
$ uv run --frozen --extra dev python -m pytest source/isaaclab/test/utils/test_images.py \
      source/isaaclab_tasks/test/core/test_cartpole_camera_observations.py -q
72 passed in 3.53s
```

`uv run --frozen isaaclab -f` passes clean.

### Caveats a reviewer should know

- The sweep exercised the **colorized `uint8` path**: `CameraCfg.colorize_semantic_segmentation` defaults to `True` and the Cartpole config declares `observation_space=[4, 96, 96]` (4 channels = RGBA). The `int32` path is reachable only with `colorize=False`; it was genuinely broken (the helper returned it unchanged) but the 60 rows do not prove it.
- The **direct-environment edit is not covered by a test**. `CartpoleCameraEnv._get_observations` calls `super()._get_observations()`, which needs a constructed environment, so it cannot be exercised sim-free. The edit is line-for-line identical to the manager-term edit, which is tested.

## Relationship to #7440

#7440 touches `isaaclab/utils/images.py` and the shared `isaaclab/envs/mdp/observations.py::image` term, but **neither Cartpole file**, so it does not fix this. Its `images.py` work is a fused normalize+layout-conversion perf change that adds an `output_channel_dim` parameter; it leaves the segmentation dispatch condition semantically unchanged and does not add `int32` handling. This PR adds an early-return branch above that condition and leaves #7440's line untouched, so the two should merge cleanly in either order.

## Checklist

- [x] I have made corresponding changes to the documentation (docstrings for `normalize_camera_image`)
- [x] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file (changelog fragments; `extension.toml` is generated)

## Release backport

### DIFF
--- a/source/isaaclab/changelog.d/fix-semantic-segmentation-dtype.rst
+++ b/source/isaaclab/changelog.d/fix-semantic-segmentation-dtype.rst
@@ -1,0 +1,8 @@
+Fixed
+^^^^^
+
+* Fixed :func:`isaaclab.utils.images.normalize_camera_image` returning non-colorized
+  ``"semantic_segmentation"`` unchanged. Such output is an ``int32`` label map on every renderer,
+  and feeding it to a convolution raised ``Input type (int) and bias type (float) should be the
+  same``. It is now cast to ``float32``; label ids carry no scale, so they are not rescaled.
+  Colorized (``uint8`` RGBA) segmentation keeps its existing ``(x / 255) - mean`` normalization.
--- a/source/isaaclab/isaaclab/utils/images.py
+++ b/source/isaaclab/isaaclab/utils/images.py
@@ -52,6 +52,8 @@ def normalize_camera_image(
 
     Dispatch (in order of check):
 
+    - non-colorized ``"semantic_segmentation"`` (any non-``uint8`` dtype, ``int32`` label ids in
+      practice): cast to ``float32``. Label ids carry no scale, so no further normalization.
     - :func:`is_rgb_like` or colorized ``"semantic_segmentation"`` (``uint8``) and contiguous
       4D: routes to the
       fused Warp kernel via :func:`~isaaclab.utils.warp.ops.normalize_image_uint8`. ``out`` and
@@ -79,10 +81,15 @@ def normalize_camera_image(
 
     Returns:
         The normalized tensor. For RGB-like and colorized semantic-segmentation input this is a
-        fresh (or pre-allocated) float32 tensor; for depth-like input it is ``images`` itself
-        (mutated in place); for normals-like input it is a new tensor; for anything else,
-        ``images`` unchanged.
+        fresh (or pre-allocated) float32 tensor; for non-colorized semantic segmentation it is a
+        float32 view or copy of ``images``; for depth-like input it is ``images`` itself (mutated
+        in place); for normals-like input it is a new tensor; for anything else, ``images``
+        unchanged.
     """
+    if data_type == "semantic_segmentation" and images.dtype != torch.uint8:
+        # Non-colorized segmentation is an integer label map (``int32`` for every renderer). Label
+        # ids have no meaningful scale, so only cast to float32 so downstream convolutions accept it.
+        return images.float()
     if is_rgb_like(data_type) or (data_type == "semantic_segmentation" and images.dtype == torch.uint8):
         if images.dtype == torch.uint8 and images.ndim == 4 and images.is_contiguous():
             return normalize_image_uint8(images, channel_dim=channel_dim, out=out)
--- a/source/isaaclab/isaaclab/utils/images.py
+++ b/source/isaaclab/isaaclab/utils/images.py
@@ -81,14 +81,14 @@ def normalize_camera_image(
 
     Returns:
         The normalized tensor. For RGB-like and colorized semantic-segmentation input this is a
-        fresh (or pre-allocated) float32 tensor; for non-colorized semantic segmentation it is a
-        float32 view or copy of ``images``; for depth-like input it is ``images`` itself (mutated
-        in place); for normals-like input it is a new tensor; for anything else, ``images``
+        fresh (or pre-allocated) float32 tensor; for non-colorized semantic segmentation it is
+        ``images`` cast to float32; for depth-like input it is ``images`` itself (mutated in
+        place); for normals-like input it is a new tensor; for anything else, ``images``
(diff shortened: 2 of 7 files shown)


kellyguo11 (approved):

kellyguo11:
run-ci

PR #7491, opened 2026-09-02
Select pretrained checkpoints from resolved task configs

maxkra15:
# Description

Select pretrained checkpoints from the final resolved environment configuration instead of retaining preset-selection history in Hydra. Physics and renderer backends remain shared dimensions, while a task-owned `variant_resolver` identifies only the policy-relevant contract (observations, actions, and shapes); presets are retained solely as reproducible training recipes.

Runtime playback, benchmark playback, and LEAPP export all consume the same checkpoint declaration as the training/publication matrix. Unsupported configurations fail closed before downloading a shape-incompatible checkpoint. For camera Cartpole this distinguishes raw RGB from depth and rejects RGB feature-extractor policies or changed image shapes rather than silently falling back to the raw-RGB model.

Camera Cartpole declares the existing RSL-RL RGB policy plus RL-Games RGB and depth policies. The default variant preserves existing unsuffixed artifact names; depth uses the stable `_depth` suffix. Backend discovery is reused during a fetch, so the resolved config's renderer tree is traversed only once.

This addresses the reported `A pre-trained checkpoint is currently unavailable for this task` failure for `Isaac-Cartpole-Camera-Direct` with `presets=rgb` and `presets=depth`.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Documentation update

## Release backport

- [x]  Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

Not applicable.

## Validation

- `uv run --frozen --extra test python -m pytest scripts/tools/test/test_train_and_publish_checkpoints.py source/isaaclab/test/benchmark/test_api.py source/isaaclab_rl/test/test_pretrained_checkpoint.py source/isaaclab_rl/test/export/test_pretrained_checkpoint_lookup.py source/isaaclab_tasks/test/core/test_hydra.py source/isaaclab_tasks/test/core/test_preset_cli.py -q` — 180 passed
- Regression tests cover implicit/default RGB, explicit RGB, final scalar overrides to depth, ResNet18/Theia rejection, changed image-shape rejection, and a single backend traversal per fetch
- `SKIP=check-git-lfs-pointers uv run --frozen isaaclab -f` — all applicable hooks passed; Git LFS pointer check skipped because this is a source-only change
- `uv run --frozen python tools/changelog/cli.py check develop` — passed
- Publication matrix listing produces the existing RGB artifact and distinct depth artifact with unchanged names
[...]

PR #6440, opened 2026-07-09
fix(rsl_rl): default obs_groups for new runners

saime428:
Summary:
- Populate a default `obs_groups` mapping for `OnPolicyRunner` configs when running with rsl-rl >= 4.0.0.
- Preserve any explicitly configured `obs_groups` values.
- Add regression coverage for missing and existing `obs_groups` in the rsl-rl deprecation helper.

Why:
- rsl-rl 4.x expects `obs_groups` in runner configs. Existing `play.py` already routes configs through `handle_deprecated_rsl_rl_cfg()` before calling `OnPolicyRunner`, so filling the missing default there keeps play/eval paths compatible without changing each script separately.

Validation:
- [x] `python -m compileall source\isaaclab_rl\isaaclab_rl\rsl_rl\utils.py`
- [x] `python -m compileall source\isaaclab_rl\test\test_rsl_rl_cfg_deprecation.py`
- [x] `python -m ruff check source\isaaclab_rl\isaaclab_rl\rsl_rl\utils.py source\isaaclab_rl\test\test_rsl_rl_cfg_deprecation.py`
- [x] lightweight `handle_deprecated_rsl_rl_cfg()` dummy-runner check for default and preserved `obs_groups`
- [ ] `python -m pytest source\isaaclab_rl\test\test_rsl_rl_cfg_deprecation.py -q` (collection is blocked in this local environment by missing Isaac Sim / Warp packages)

Notes for reviewers:
- Closes #5363.
- This does not change public APIs or training logic; it only fills a compatibility default when the runner config omits `obs_groups`.

PR #7532, opened 2026-09-03
[Odin] Fix preset-based agent selection for rsl_rl, rl_games and sb3

AntoineRichard:
# Description

Preset-based `--agent` auto-selection is dead code for `rsl_rl`, `rl_games` and `sb3`. Two defects sit **in series** on the same code path:

1. **The selection guard cannot see past the CLI default.** `_auto_select_agent` in `source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py` was only reached when `args.agent is None`. `add_common_train_args` registers `--agent` with `default=agent_default`, and every backend except `skrl` passes a non-`None` `agent_default` (`rsl_rl_cfg_entry_point`, `rl_games_cfg_entry_point`, `sb3_cfg_entry_point`). The parsed value is therefore never `None`, so the branch never runs.
2. **The benchmark entrypoints never asked for it.** `benchmark_train_{rsl_rl,rl_games,sb3}.py` and `benchmark_play_{rsl_rl,rl_games,sb3}.py` called `setup_preset_cli(parser, argv)` without `agent_library`, so auto-selection was not attempted at all (and the registered-agent help listing was missing too). Only the `skrl` benchmark entrypoints wired it.

## Provenance and symptom

Benchmark sweep dispatch `20260901-153531` (image built from `release/3.0.0` at `f88dbc59c82`, `rsl_rl`) produced **60 failed rows**: `resnet18` (30) and `theia_tiny` (30), failing 100% across all three renderers and all physics backends. Every one died before the first training step:

```
ValueError: Observation 'critic' in observation set 'critic' not found in the observations
from the environment. Available observations from the environment: ['policy']
```

Call chain: `OnPolicyRunner.__init__` → `ppo.py construct_algorithm` → `rsl_rl/utils/utils.py:233 resolve_obs_groups`.

## Why this is an agent-entrypoint bug, not a missing observation group

`Isaac-Cartpole-Camera` already declares the correct pairing:

```python
"agent_preset_compatibility": {
    "rsl_rl_cfg_entry_point": _RAW_CAMERA_PRESETS,
    "rsl_rl_feature_cfg_entry_point": ("resnet18", "theia_tiny"),
    ...
}
```

and `CartpoleCameraFeaturePPORunnerCfg` sets `obs_groups` with `critic: ["policy"]`. The feature entry point exists, is registered, is correct — it was simply unreachable, so `resnet18`/`theia_tiny` ran against the raw-camera `CartpoleCameraPPORunnerCfg`, whose `obs_groups` asks for a `critic` group the env does not expose. Nothing needs to be added to the environment.

## Why both defects are in one PR

Neither half fixes the observed failure alone; this was measured, not assumed. Reverting either half of the change and running the regression test:

| state | result |
| --- | --- |
| defect 2 fixed only (`agent_library` wired, guard unchanged) | 4 failed — `--agent` still parses to the non-`None` default, so the guard rejects |
| defect 1 fixed only (guard fixed, `agent_library` not wired) | 4 failed — `if agent_library and ...` is `False`, auto-selection never attempted |
| both fixed | passes |

They are two links in one chain, so splitting them yields a PR that fixes nothing observable and a PR that cannot be tested end-to-end.

## Scope: train **and** play, benchmark **and** non-benchmark

To pre-empt the obvious question — this is not a play-only or a train-only fix.

| path | affected by | fixed by | proven by |
| --- | --- | --- | --- |
| `benchmark_train_{rsl_rl,rl_games,sb3}` | defects 1 + 2 | `preset_cli.py` + `agent_library=` wiring | `test_training_request_selects_preset_compatible_agent` — **this is the observed 60-row failure** |
| `benchmark_play_{rsl_rl,rl_games,sb3}` | defects 1 + 2 | `preset_cli.py` + `agent_library=` wiring | `test_play_request_selects_preset_compatible_agent` |
| `isaaclab_rl` `train_*`/`play_*` (non-benchmark) | defect 1 only — they already pass `agent_library` | `preset_cli.py` alone; **no file in this PR touches them** | `test_setup_preset_cli_auto_selects_agent_over_non_none_default` |

Each half of the benchmark wiring was reverted independently and re-tested:

* revert the three **train** wirings → the 4 training selection cases fail. Load-bearing for the reported failure.
* revert the three **play** wirings → the 2 playback selection cases fail.

There is **no train/play asymmetry** that would justify wiring only one of them: both register `--agent` with the same non-`None` default via the same helper, both call `setup_preset_cli`, and both feed `args.agent` into the same `resolve_task_config(...)` and then into the same `OnPolicyRunner(...)` construction. The reason the sweep only surfaced the training failure is ordering, not asymmetry — a row that dies at `train_rc=1` never reaches playback. Measured on the playback path before the wiring was added:

```
$ BenchmarkPlayRequest(backend="rsl_rl", task="Isaac-Cartpole-Camera", presets=("resnet18",))
args.agent : rsl_rl_cfg_entry_point          # raw-camera config, wrong
# with the wiring:
args.agent : rsl_rl_feature_cfg_entry_point  # correct
```

Playback would therefore have loaded a feature-trained checkpoint into the raw-camera policy architecture.

## Fix chosen

Detect an explicitly typed `--agent` by re-parsing the same argv into a namespace pre-seeded with a sentinel: argparse only applies a default for a destination the namespace does not already carry, so the sentinel survives unless the user actually typed the flag. Auto-selection runs only when it does survive; an explicit `--agent` still wins. The six benchmark entrypoints now pass `agent_library`.

The probe uses the same parser on the same argv as the real parse, so its verdict *is* argparse's verdict. Verified across every spelling — `--agent V`, `--agent=V`, the abbreviation `--age V`, repeated flags, and `--agent` after a `--` separator (correctly not explicit: argparse does not set it in the real parse either). A literal argv scan for `--agent` would get the abbreviation wrong and silently override a user's explicit choice, which is why the probe is preferred; the repo's existing `ExplicitAction` idiom would work too but requires touching all ten `--agent` registration sites across three packages, and a missed site fails the same silent way.

### Rejected: make `--agent` default to `None` everywhere

This is the obvious fix and it is not safe. `skrl` can default to `None` because `train_skrl.py` reconstructs the entry point from `--algorithm` when it is `None`. The other three pass `args_cli.agent` straight into `resolve_task_config(...)`, which has no such fallback — `hydra.py:619` sets `agent_cfg = load_cfg_from_registry(...) if agent_entry else None`, and the entrypoints then dereference `agent_cfg.max_iterations`. A `None` default would break every plain `--task=X` run that relies on the canonical entry point, i.e. the overwhelmingly common case. Fixing that would mean adding a fallback to each of the six benchmark entrypoints plus the four train/play entrypoints: a much wider blast radius than the bug.

### Blast radius

Behavior changes only where auto-selection actually fires, and only when the user did **not** type `--agent`:

* **Rule 1 (preset-based)** fires only for tasks that opted in with `agent_preset_compatibility` — today `Isaac-Cartpole-Camera` and the two cartpole-showcase tasks. That is the declared contract finally being honored. Anyone who wants the previous (broken) pairing can still pass `--agent rsl_rl_cfg_entry_point` explicitly; covered by a test.
* **Rule 2 (default-absent)** fires only when `<library>_cfg_entry_point` is *not* registered and exactly one other entry point is. That path previously resolved an unregistered entry point and crashed, so this is strictly a repair.
* Everything else — no preset pairing declared, or the canonical default is registered — keeps the exact default it had; covered by a test for all three libraries.
* `skrl` is unaffected: it already passed `agent_default=None`, and the explicit/implicit distinction collapses to the old `is None` check for it.

`rl_games` and `sb3` share the defect and are fixed by the same change. **`sb3` has no task declaring `agent_preset_compatibility`, so there is no positive selection test for it** — its wiring is covered by the shared code path and by a default-preservation test only.

## Type of change

## Screenshots

Not applicable.

## Validation

Regression tests were confirmed to fail on `develop` without the fix and pass with it.

**Before the fix** (production changes reverted to `develop`, tests in place):

```
$ uv run --frozen --extra test python -m pytest source/isaaclab_tasks/test/core/test_preset_cli.py \
    source/isaaclab/test/benchmark/test_api.py -q -p no:warnings \
    -k "preset_compatible_agent or keeps_backend_default_agent or auto_selects_agent_over_non_none or keeps_explicit_agent"

FAILED test_preset_cli.py::test_setup_preset_cli_auto_selects_agent_over_non_none_default
FAILED test_api.py::test_training_request_selects_preset_compatible_agent[resnet18-rsl_rl]
FAILED test_api.py::test_training_request_selects_preset_compatible_agent[resnet18-rl_games]
FAILED test_api.py::test_training_request_selects_preset_compatible_agent[theia_tiny-rsl_rl]
FAILED test_api.py::test_training_request_selects_preset_compatible_agent[theia_tiny-rl_games]
5 failed, 6 passed, 53 deselected

E  AssertionError: assert 'rl_games_cfg_entry_point' == 'rl_games_feature_cfg_entry_point'
```

**After the fix:**

```
$ uv run --frozen --extra test python -m pytest source/isaaclab_tasks/test/core/test_preset_cli.py \
    source/isaaclab_tasks/test/core/test_hydra.py source/isaaclab/test/benchmark \
    source/isaaclab_rl/test/test_entrypoints_common.py -q -p no:warnings
492 passed, 1 skipped in 30.89s

$ uv run --frozen --extra test --extra skrl --extra sb3 --extra rl-games --extra rsl-rl \
    python -m pytest source/isaaclab_rl/test/test_typed_preset_cli_train_play.py -q -p no:warnings
8 passed in 33.27s
```

(The last suite fails on a bare `--extra test` environment on `develop` too — `ModuleNotFoundError: No module named 'skrl'` etc. — so it was rerun with the RL extras.)

```
$ uv run --frozen python tools/changelog/cli.py check develop
✓ All modified packages have valid changelog fragments.

$ uv run --frozen isaaclab -f
all hooks passed
```

Tests added:

* `source/isaaclab_tasks/test/core/test_preset_cli.py` — auto-selection over a non-`None` default; explicit `--agent` wins over the preset, parametrized over the three spellings argparse accepts. The two pre-existing auto-selection tests only covered `skrl` with `agent_default=None`, which is exactly why this bug went unnoticed.
* `source/isaaclab/test/benchmark/test_api.py` — drives the real benchmark entrypoints via `BenchmarkTrainingRequest`/`BenchmarkPlayRequest` and `_parse_args`, asserting the feature entry point is selected for `resnet18`/`theia_tiny` × `rsl_rl`/`rl_games` on the train path (mirroring the failing rows), the same on the play path, and that the canonical default survives otherwise (`rsl_rl`, `rl_games`, `sb3`).

Both assert the *selected agent config*, not merely absence of an exception.

Not run: a real training job on the failing rows (requires GPU sim). The failure is a construction-time config selection, fully reproduced at CLI level.

## Related PRs — checked, no overlap

* **#7491** (Select pretrained checkpoints from resolved task configs) touches `isaaclab_rl/entrypoints/common.py`, the four `play_*` entrypoints, the four `benchmark_play_*` entrypoints, `cartpole/__init__.py` and `test_preset_cli.py` — adjacent, but changes neither `agent_default` nor the selection guard. Textual conflicts are possible in `test_preset_cli.py` and the `benchmark_play_*` files (one-line `setup_preset_cli(...)` call); no semantic conflict.
* **#6440** (`fix(rsl_rl): default obs_groups for new runners`) adds an `obs_groups` default in `isaaclab_rl/rsl_rl/utils.py`. It would mask this symptom for runners that omit `obs_groups`, but `CartpoleCameraPPORunnerCfg` sets `obs_groups` explicitly, so it does not fix these 60 rows — and it would not make the correct feature config get selected either. Complementary, not duplicate.
* `gh pr list --search "agent preset"` returned nothing touching this code.

## Release backport


hujc7 (commented):
Superseded — re-posted as inline comments anchored to the relevant lines, so each finding sits on the code it concerns. See the inline thread(s) on this PR.

hujc7 (approved):

hujc7:
run-ci

source/isaaclab_tasks/isaaclab_tasks/utils/preset_cli.py
```diff
@@ -136,7 +137,7 @@ def setup_preset_cli(
     args, remaining = parser.parse_known_args(args_to_parse)
 
     task_name = getattr(args, "task", None) or argv_helper.task_name
-    if agent_library and getattr(args, "agent", None) is None and task_name:
+    if agent_library and task_name and hasattr(args, "agent") and not _agent_passed_explicitly(parser, args_to_parse):
```
hujc7, line 140: **[P2] Agent review:** `agent_preset_compatibility` drives selection here but nothing validates an explicit `--agent` against it, so an incompatible one is honoured silently and still reaches the opaque runner-construction error this PR removes for the auto-selected path.

Raising as a follow-up rather than a change here, since *"Narrow the preset agent-selection fix to its minimum"* deliberately kept this out of scope.

hujc7 (commented):
**[P2] Agent review:** both bot assessments are pinned to `61a2957` and predate the last two commits. The review-bot's approval specifically praises the `_auto_select_agent` return-value shape that `d29e344d7f2` reverted, so it no longer describes the code. Worth re-running them before merge.

kellyguo11 (approved):

PR #7536, opened 2026-09-03
Move external wrench frame selection into WrenchComposer

AntoineRichard:
# Description

`WrenchComposer.compose_to_body_frame()` was called unconditionally by every backend writer on every physics step. It reads `body_com_pos_w` and `body_link_quat_w`, which on PhysX chains to `_ensure_fk_fresh()` → `update_articulations_kinematic()` (a scene-global articulation FK pass) plus `root_view.get_link_transforms()`, then launches the composition kernel.

That work is unnecessary whenever the buffered wrench is already in a frame the consumer accepts. This PR moves the frame decision into the composer, which hands each writer the cheapest valid representation:

- **all-local content** → the local buffers, untouched. An all-local wrench composes to exactly itself, because the pose data is multiplied by zero.
- **all-global-at-CoM content** → the global buffers, untouched, submitted in the world frame by backends that accept one.
- **anything else** → composed as before.

Writers declare what they can consume once at construction (`supports_world_at_com`) and call `resolve_submission()` per step. All frame logic stays inside the composer; no writer touches a raw buffer or knows a frame rule.

## New public API

```python
class WrenchComposer:
    class Frame(IntEnum):
        BODY = 0          # body frame, force at the body's CoM
        WORLD_AT_COM = 1  # world frame, force at the body's CoM

    def __init__(self, asset, *, supports_world_at_com: bool = False) -> None: ...
    def resolve_submission(self) -> tuple[wp.array, wp.array, Frame]: ...
```

Additive only. `compose_to_body_frame()`, `out_force_b`, and `out_torque_b` remain public and behaviorally unchanged, so no deprecation cycle is needed. `supports_world_at_com` defaults to `False`, which is exactly the previous behavior.

Eligibility is tracked with a private mask of which input buffer families hold contributions (`LOCAL`, `GLOBAL_AT_COM`, `GLOBAL_POSITIONED`), classified from the same routing the `add_forces_to_dual_buffers_*` kernels use. The mask is sticky across partial resets and cleared only by a full `reset()` — proving a partial reset removed every contribution of a family would require scanning the buffers. That can forgo the fast path but never changes the resulting wrench.

## Equivalence

All three branches are exact, from `compose_wrench_to_body_frame`:

- **local only**: the three global buffers are zero, so both `quat_rotate_inv` terms vanish and the output is identically `(local_force_b, local_torque_b)`. Note a *positioned* local force also qualifies — the kernel folds `cross(P_b, F_b)` straight into `local_torque_b`.
- **global-at-CoM only**: `global_force_w` and the local buffers are zero, so `corrected_torque_w == global_torque_w`. Submitting the world buffers is the same wrench, provided the consumer applies force at the CoM — verified against the vendored `omni.physics.tensors` API, where `position_data=None` means "at the link transform" for **both** `is_global` values, so the flag only reinterprets the vectors' frame and never moves the application point.
- **anything else**: unchanged.

## Backend coverage

| backend | `supports_world_at_com` | change |
|---|---|---|
| PhysX | `True` | writer maps the returned frame onto `is_global` |
| OvPhysX | `True` | writer maps it onto a new `wrench_is_world` kernel flag that skips the inline rotate |
| Newton | `False` (default) | writer swaps to `resolve_submission()`; no constructor change |

OvPhysX is the largest win. Its wrench binding wants a world-frame wrench, so a global wrench previously round-tripped: the composer rotated world→body, then the packing kernel rotated body→world. Measured on this branch with a body rotated 90° about +z, a local force `(1,0,0)` emits `(0,1,0)` while a global force `(1,0,0)` emits `(1,0,0)` — an exact round trip, so skipping both rotations is output-preserving. The packed `[6:9]` link position is still written unconditionally on both paths.

Newton keeps the body frame because it binds a body-frame array to the solver. Worth noting for a follow-up: its writer kernel already performs the same body→world rotation OvPhysX's does, so Newton could plausibly take the same flag and win the global-at-CoM path too. That is deliberately out of scope here.

## Validation

Environment: `OMNI_KIT_ACCEPT_EULA=YES uv run --extra isaacsim --extra test --extra ovphysx`, which resolves `BACKENDS == ['physx', 'newton', 'ovphysx']` with CUDA. Note `EXP_PATH` must be exported or `_iface_test_boot.py` takes its kitless path and `physx` silently drops out, making backend tests skip green.

- `test_wrench_composer.py` plus `test/assets/`: **6760 passed**, 65 skipped, 68 xfailed.
- 5 failures in `test_articulation_ordering.py` during full-suite runs are **pre-existing and unrelated**: the identical 5 fail at the branch base with none of this branch present, and pass in isolation. Test counts reconcile — 6350 passing at base, 6370 at head for the same selection, and the +20 is exactly this branch's new tests.
- `uv run isaaclab -f` clean.

New coverage: composer unit tests for every content case and the mask lifecycle; per-backend submission tests asserting the chosen frame and that no body-pose read occurs on a fast path; an OvPhysX articulation test driving the ordered kernel's world path under a **non-identity body ordering**; and a cross-backend equivalence test over all three backends × four content combinations, comparing the submitted wrench against the composed one.

Two regressions were confirmed to be caught rather than assumed: injecting the eligibility bug the design guards against (`_content & LOCAL` instead of `==`) fails the mixed-content test, and perturbing the expected world-frame torque fails exactly the two `WORLD_AT_COM` cases and no others.

## Performance

Real PhysX, RTX 5090, ANYMAL-C (17 bodies) × 4096 envs = 69,632 links. Same-process paired A/B with the arm order alternated per iteration, 400–500 pairs, 95% CI on paired differences. Baseline is `compose_to_body_frame()` — what the writers previously called unconditionally.

| scope | content | baseline | patched | paired delta | 95% CI |
|---|---|---:|---:|---:|---|
| composition path only | global-at-CoM | 0.2194 ms | 0.0191 ms | **−0.2003 ms** (−91.3%) | [0.1915, 0.2091] |
| composition path only | local | 0.1926 ms | 0.0159 ms | **−0.1767 ms** (−91.8%) | [0.1712, 0.1822] |
| full `write_data_to_sim` | global-at-CoM | 1.0799 ms | 0.9897 ms | **−0.0902 ms** (−8.4%) | [0.0797, 0.1007] |
| full `write_data_to_sim` | local | 1.0072 ms | 0.9305 ms | **−0.0767 ms** (−7.6%) | [0.0719, 0.0815] |

Read the end-to-end row as the honest headline: **~0.08 ms per asset per physics step**. Roughly half the isolated saving does not reach the caller, because other work in `write_data_to_sim` touches body poses anyway and pays part of that cost regardless. A cuboid rigid-object scene at the same env count saved only ~6.5 µs, as expected — there is no articulation FK pass to skip.

Scope caveat: no in-tree task sets `is_global=True` today (both shipped `apply_external_force_torque` events use the local default), so the local path is what delivers value now, and it is the one that reaches every backend.

## Supersedes

This replaces two open PRs, both of which found one half of this problem:

- **#7431** (@NeoZng) identified the global-at-CoM case and the cost of the pose read. Its mechanism is correct and its writer-level measurement is corroborated here. It is superseded on placement rather than substance: it put the frame decision in three PhysX writers behind a public eligibility flag, leaving Newton and OvPhysX unserved. The credit for the world-at-CoM half of this mechanism is theirs.
- **#7362** cached composition for unchanged local wrenches. Superseded because not composing at all is strictly stronger — the cache does nothing when the wrench is rewritten every step, which is the common case for a per-step randomized push.

## Type of change

## Checklist

### DIFF
--- a/source/isaaclab/isaaclab/utils/wrench_composer.py
+++ b/source/isaaclab/isaaclab/utils/wrench_composer.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import warnings
 from collections.abc import Sequence
+from enum import IntEnum, IntFlag, auto
 from typing import TYPE_CHECKING
 
 import numpy as np
@@ -30,7 +31,29 @@ if TYPE_CHECKING:
 
 
 class WrenchComposer:
-    def __init__(self, asset: BaseArticulation | BaseRigidObject | BaseRigidObjectCollection) -> None:
+    class Frame(IntEnum):
+        """Representation of a submitted external wrench."""
+
+        BODY = 0
+        """Body frame, force applied at the body's center of mass."""
+
+        WORLD_AT_COM = 1
+        """World frame, force applied at the body's center of mass."""
+
+    class _Content(IntFlag):
+        """Which input buffer families currently hold contributions."""
+
+        NONE = 0
+        LOCAL = auto()
+        GLOBAL_AT_COM = auto()
+        GLOBAL_POSITIONED = auto()
+
+    def __init__(
+        self,
+        asset: BaseArticulation | BaseRigidObject | BaseRigidObjectCollection,
+        *,
+        supports_world_at_com: bool = False,
+    ) -> None:
         """Wrench composer with dual-buffer architecture.
 
         This class composes forces and torques applied to rigid bodies. Forces and torques can be
@@ -55,6 +78,9 @@ class WrenchComposer:
 
         Args:
             asset: Asset to use.
+            supports_world_at_com: Whether the consumer can apply a world-frame wrench at the body's
+                center of mass. When False, :meth:`resolve_submission` always returns a body-frame
+                wrench. Defaults to False.
         """
         self.num_envs = asset.num_instances
         # Avoid isinstance to prevent circular import issues; check by attribute presence instead.
@@ -66,6 +92,8 @@ class WrenchComposer:
         self._asset = asset
         self._active = False
         self._dirty = False
+        self._supports_world_at_com = supports_world_at_com
+        self._content = self._Content.NONE
         if hasattr(self._asset.data, "body_com_pos_w"):
             self._get_com_pos_fn = lambda a=self._asset: a.data.body_com_pos_w.warp
         else:
@@ -265,6 +293,7 @@ class WrenchComposer:
 
         self._active = True
         self._dirty = True
+        self._content |= self._classify(forces, torques, positions, is_global)
 
         wp.launch(
             add_forces_to_dual_buffers_index_kernel(env_ids, body_ids),
@@ -328,6 +357,7 @@ class WrenchComposer:
 
         self._active = True
         self._dirty = True
+        self._content |= self._classify(forces, torques, positions, is_global)
 
         wp.launch(
             set_forces_to_dual_buffers_index_kernel(env_ids, body_ids),
@@ -388,6 +418,7 @@ class WrenchComposer:
 
         self._active = True
         self._dirty = True
+        self._content |= self._classify(forces, torques, positions, is_global)
 
         wp.launch(
             add_forces_to_dual_buffers_mask,
@@ -453,6 +484,7 @@ class WrenchComposer:
 
         self._active = True
         self._dirty = True
+        self._content |= self._classify(forces, torques, positions, is_global)
 
         wp.launch(
             set_forces_to_dual_buffers_mask,
@@ -493,6 +525,7 @@ class WrenchComposer:
 
         self._active = True
         self._dirty = True
+        self._content |= other._content
 
         wp.launch(
             add_raw_wrench_buffers,
@@ -542,6 +575,30 @@ class WrenchComposer:
         )
         self._dirty = False
 
+    def resolve_submission(self) -> tuple[wp.array, wp.array, Frame]:
+        """Return the cheapest representation of the buffered wrench this consumer can accept.
+
+        Composition into the body frame reads the body poses, which on some backends forces a
+        kinematics update. That work is unnecessary when the buffered wrench is already in a frame
+        the consumer accepts: an all-local wrench is its own body-frame composition, and an
+        all-global-at-CoM wrench is directly submittable by a consumer that accepts a world frame.
+
+        The eligibility state is conservative after a partial :meth:`reset`: it may keep composing
+        even when the selected reset removed every contribution that made composition necessary. A
+        full :meth:`reset` restores it. This never changes the resulting wrench, only the cost.
+
+        Returns:
+            Force [N], torque [N·m], and the frame both are expressed in. Shapes are
+            ``(num_envs, num_bodies)`` with dtype ``wp.vec3f``. The arrays are owned by the composer
+            and stay valid until the next mutating call.
+        """
+        if self._content == self._Content.LOCAL:
+            return self._local_force_b, self._local_torque_b, self.Frame.BODY
+        if self._supports_world_at_com and self._content == self._Content.GLOBAL_AT_COM:
+            return self._global_force_at_com_w, self._global_torque_w, self.Frame.WORLD_AT_COM
+        self._ensure_composed()
+        return self._out_force_b, self._out_torque_b, self.Frame.BODY
+
     def reset(
         self,
         env_ids: wp.array | torch.Tensor | Sequence[int] | slice | None = None,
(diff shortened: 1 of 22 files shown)


source/isaaclab/isaaclab/utils/wrench_composer.py
```diff
@@ -30,7 +31,29 @@
...
+        WORLD_AT_COM = 1
+        """World frame, force applied at the body's center of mass."""
+
+    class _Content(IntFlag):
+        """Which input buffer families currently hold contributions."""
+
+        NONE = 0
+        LOCAL = auto()
+        GLOBAL_AT_COM = auto()
+        GLOBAL_POSITIONED = auto()
```
AntoineRichard: Why auto()?

AntoineRichard:
@NeoZng could you see if this works for you. The bulk of the changes are tests.

NeoZng:
Thanks @AntoineRichard — I tested the current patch cleanly on the latest `develop`. It works for our recovery-assist use case, and the architecture is cleaner.

Our reset-mode MDP event writes a persistent `[0, 0, 300] N` world-frame force at the torso CoM using `permanent_wrench_composer.set_forces_and_torques_index(..., is_global=True)` without `positions`; the wrench is then submitted on every physics step. Startup CoM randomization was up to +/-3 cm.

On Isaac Sim 6.0.1 with an RTX 5090, 388 WrenchComposer tests and 14 focused PhysX/Newton interface tests passed. A learner-free profile of the actual MDP term and full `ManagerBasedRLEnv.step()` path (5 physics substeps, 200 alternating pairs) gave:

| Environments | Full MDP step | Paired saving (95% CI) |
| ---: | ---: | ---: |
| 4096 | 67.1557 -> 66.6808 ms (-0.71%) | 0.4749 ms [0.3446, 0.6052] |
| 16384 | 123.2129 -> 122.3709 ms (-0.68%) | 0.8420 ms [0.6322, 1.0519] |

The patch is correct and ready for our use case.

maxkra15 (approved):
Full-coverage `set_*` operations retain stale content flags

With the default `env_ids=None`, `set_forces_and_torques_index()` first resolves it to `_ALL_ENV_INDICES`. It then calls `reset(env_ids=env_ids)`, which takes the partial-reset path and does not clear `_content`, even though every environment has been reset.

As a result, replacing all buffered content does not replace its classification. For example:

1. Set a positioned global wrench → `_content == GLOBAL_POSITIONED`.
2. Set an all-local wrench with the default `env_ids=None`.
3. `_content` becomes `GLOBAL_POSITIONED | LOCAL`.

The buffers now contain only local content, but `resolve_submission()` treats them as mixed and continues composing through the body poses. The same issue exists in `set_forces_and_torques_mask()` when its default all-true environment mask is used.

This is correctness-safe, but it can disable the fast path indefinitely—the behavior this PR is intended to optimize.

Could we preserve whether the original selection covered all environments and replace `_content` with the new classification in that case, while retaining the conservative OR behavior for genuinely partial updates?

marcodiiga:
run-ci

source/isaaclab_physx/isaaclab_physx/assets/articulation/articulation.py
```diff
@@ -251,16 +251,17 @@ def write_data_to_sim(self):
                 composer.add_raw_buffers_from(self._permanent_wrench_composer)
             else:
                 composer = self._permanent_wrench_composer
-            composer.compose_to_body_frame()
+            force_user, torque_user, frame = composer.resolve_submission()
```
marcodiiga, line 254: Non-blocking correctness issue: this local-frame fast path skips the `body_link_quat_w` read that previously called `_ensure_fk_fresh()`. After a root or joint state write, PhysX requires `update_articulations_kinematic()` before link transforms reflect the new state. Because `apply_forces_and_torques_at_position(..., is_global=False)` rotates the wrench using those link transforms, the first local wrench after a reset can be rotated using the pre-reset orientation. Please call `self.data._ensure_fk_fresh()` before submission when `frame is WrenchComposer.Frame.BODY`; the timestamp guard makes this a no-op in steady state.

marcodiiga (approved):
Reviewed the exact head. The frame-selection math and backend mappings are sound, and I found no blocking issue. I left one non-blocking PhysX articulation reset/FK edge case; the existing content-mask and flattened-view notes are performance follow-ups.

ClemensSchwarke (approved):

ooctipus (approved):

AntoineRichard:
run-ci

ooctipus:
run-ci

StafaH (approved):

StafaH:
run-ci

PR #7538, opened 2026-09-03
Fix stale gravity in OvPhysxManager.get_gravity and in IMU/PVA sensors

AntoineRichard:
# Description

Two related gravity defects, both surfaced by a QA report against the OVPhysX backend.

**1. `OvPhysxManager.get_gravity()` returned a stale value.** `set_gravity()` authors the new gravity into OvStage and applies the ordinal to the running simulation, but `get_gravity()` read `SimulationCfg.gravity`, which the setter never touched. Every call after a live update returned the construction-time value, silently and permanently:

```
get_gravity before set : (0.0, 0.0, -9.81)
set_gravity called with: (0.0, 0.0, -3.72)
get_gravity after  set : (0.0, 0.0, -9.81)
```

The fix tracks the applied gravity on the manager rather than writing back to `cfg.gravity`. Writing back would have been wrong: both `_call_physx` and `_call_ovphysx` in `randomize_physics_scene_gravity` resample from `env.sim.cfg.gravity` as the pristine base, so mutating it would make successive `"add"`/`"scale"` randomizations compound and skew the distribution.

**2. IMU and PVA snapshotted gravity at initialization.** Both sensors read gravity once in `_initialize_impl` and baked it into a static Warp buffer, so gravity randomized at runtime never reached the accelerometer bias or the projected gravity direction. This affected **both** PhysX and OVPhysX — it was not specific to defect 1, and fixing `get_gravity()` alone would not have changed sensor behavior.

Both sensors now refresh the buffer at the top of `_update_buffers_impl`, skipping the refill when the scene gravity is unchanged (a tuple compare, so steady-state cost is one `get_gravity()` call per sensor per update and no GPU work). The buffer is filled **in place** rather than reallocated: the PhysX sensors consume it through a recorded `wp.Launch` that captures the array pointer, and PVA additionally exposes it as the public `GRAVITY_VEC_W` proxy. PVA's normalization mirrors `math_utils.normalize`'s eps-clamp, so zero gravity yields a zero direction rather than NaNs.

Newton's `Pva` already passes the live `model.gravity` array into its kernel each update and needed no change.

## Known remaining gaps (not addressed here)

- `GRAVITY_VEC_W` is still snapshotted at construction in the **asset** data classes (`rigid_object_data.py`, `rigid_object_collection_data.py`, `articulation_data.py`) for both `isaaclab_physx` and `isaaclab_ov`. Newton already fixed its three by binding `ProxyArray(model.gravity[:world_count])` zero-copy. Extending that here needs an API decision first: Newton's `GRAVITY_VEC_W` carries the actual m/s² vector with consumers normalizing on read, whereas PhysX/OV carry a pre-normalized unit direction. Converging on Newton's convention would change a public attribute's units and touch consumers in `isaaclab_experimental` and `isaaclab_tasks_experimental`.
- Newton's `Imu` applies no gravity bias at all, unlike the PhysX and OVPhysX IMUs which bias the accelerometer by `-g`. This looks like a cross-backend behavioral gap rather than a staleness bug.

## Type of change

## Screenshots

None.

## Checklist

## Testing

Regression coverage, each verified to fail without the corresponding fix and pass with it:

- `test_set_gravity_writes_and_releases_ovstage_control_resources` — extended to assert `get_gravity()` round-trips after a successful `set_gravity()`, still reports the previously applied value when the OvStage write fails, and that `cfg.gravity` is left untouched as the randomization base. Without the fix it fails with `assert (0.0, 0.0, -9.81) == approx((0.0, 0.0, -3.72))`, reproducing the reported symptom exactly.
- `test_sensor_tracks_runtime_gravity_changes` (new) — a mid-run gravity change must reach a *replayed* recorded launch, with the buffer pointer unchanged.
- `test_sensor_skips_gravity_refill_when_unchanged` (new) — an unchanged gravity must not clobber the buffer every update.

Suites run locally: PhysX IMU 11 passed, PhysX PVA 11 passed, OV IMU 12 passed / 14 skipped, OV PVA 13 passed / 15 skipped, `isaaclab_ov/test/physics/` 53 passed (includes the real-backend gravity test). `pre-commit run --all-files` clean.

### DIFF
--- a/source/isaaclab_ov/changelog.d/fix-ovphysx-get-gravity-stale.rst
+++ b/source/isaaclab_ov/changelog.d/fix-ovphysx-get-gravity-stale.rst
@@ -5,3 +5,7 @@ Fixed
   gravity after :meth:`~isaaclab_ov.physics.OvPhysxManager.set_gravity` changed the running scene.
   The manager now tracks the applied gravity vector, while ``SimulationCfg.gravity`` stays the
   nominal value that randomization terms resample from.
+* Fixed :class:`~isaaclab_ov.sensors.Imu` and :class:`~isaaclab_ov.sensors.Pva` reporting gravity
+  captured at sensor initialization. Both sensors now re-read the scene gravity on every update, so
+  runtime randomization through :func:`~isaaclab.envs.mdp.events.randomize_physics_scene_gravity`
+  is reflected in the accelerometer bias and the projected gravity direction.
--- a/source/isaaclab_ov/changelog.d/fix-ovphysx-get-gravity-stale.rst
+++ b/source/isaaclab_ov/changelog.d/fix-ovphysx-get-gravity-stale.rst
@@ -1,0 +1,7 @@
+Fixed
+^^^^^
+
+* Fixed :meth:`~isaaclab_ov.physics.OvPhysxManager.get_gravity` returning the construction-time
+  gravity after :meth:`~isaaclab_ov.physics.OvPhysxManager.set_gravity` changed the running scene.
+  The manager now tracks the applied gravity vector, while ``SimulationCfg.gravity`` stays the
+  nominal value that randomization terms resample from.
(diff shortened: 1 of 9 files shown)


AntoineRichard:
Superseded by #7416. The gravity-staleness fixes here touch the same IMU/PVA kernels as the solver-acceleration work in #7416, so the two could not be reviewed or merged independently. All three commits from this branch are now merged into #7416, along with the PVA solver-acceleration change and a scene-wide-vector refactor of the gravity parameters.

Closing in favor of #7416.

PR #7543, opened 2026-09-03
[CI] Link _isaac_sim only where Kit is installed in the test container

hujc7:
## Summary

`Multi-GPU training smoke (kit-less)` fails on every PR into `release/3.0.0`, exiting before pytest
runs. Two lines in the shared test-container script fix it.

# Description

## 1. Cause

`run_tests.sh` bind-mounts nine writable runtime directories under `/isaac-sim/…` into every
container, so `/isaac-sim` exists as a directory in the kit-less image as well — which ships no Kit.
The container script then plants the link unconditionally:

```bash
rm _isaac_sim || true
ln -s /isaac-sim _isaac_sim
```

`isaaclab.sh` treats an `_isaac_sim` directory without a `.isaaclab_source_build` marker as a
downloaded Isaac Sim and refuses to combine it with an active virtual environment (#7466, on this
branch as 69d5d1fbf94). `Dockerfile.kitless` sets `VIRTUAL_ENV=/opt/isaaclab-venv` (#6355), so every
kit-less run stops at:

```
[ERROR] Downloaded Isaac Sim packages cannot be combined with a Python virtual environment.
🔴 Failed to copy specific result file: Could not find the file /workspace/isaaclab/tests/multi-gpu-smoke-kitless-report.xml
```

## 2. Change

Plant the link only where Kit actually lives at that path. Kit images are unaffected — they have
`/isaac-sim/python.sh` and get the link exactly as before; the kit-less image resolves its
interpreter through `VIRTUAL_ENV`, as it did before #7466.

The same script serves ~29 CI jobs through `run-tests` and `run-package-tests`, all but one of which
pass a Kit image, so testing the container rather than threading a "kit-less" flag through every
caller keeps the fix to one place.

## 3. Verification

- The regression test added beside the existing kit-less invariant in
  `docker/test/test_container_profiles.py` fails on the unmodified script
  (`AssertionError: ['ln -s /isaac-sim _isaac_sim']`) and passes with it. 14 pass in that file.
- `bash -n` on the script; pre-commit on both files.
- The same two-line change on `develop` turned the job green end to end: `8 passed, 7 deselected in
  790.74s`, with zero occurrences of the venv error in the job log.

## 4. Relationship to the `develop` fix

The develop-side PR pairs this with a second, unrelated fix for the `Build Base Docker Image` job,
which fails there because #7405 added an image-invariants step that a deps-cache hit cannot satisfy.
**#7405 is not on this branch** — its `Dockerfile.base` still installs with
`_isaac_sim/python.sh -m pip`, there is no invariants step, and its author left the backport box
unticked. That half would be dead config here, so this PR carries only the part that applies.

## Type of change

## Checklist

- [ ] Documentation needs no change: CI-internal script
- [ ] Changelog: no source package changed

### DIFF
--- a/.github/actions/run-tests/run_tests.sh
+++ b/.github/actions/run-tests/run_tests.sh
@@ -320,8 +320,11 @@ run_tests() {
       set -e
       cd /workspace/isaaclab
       mkdir -p tests
-      rm _isaac_sim || true
-      ln -s /isaac-sim _isaac_sim
+      # The runtime mounts above create /isaac-sim in every image. Link it only where Kit
+      # lives there: in the kit-less image the link would read as a downloaded Isaac Sim,
+      # which isaaclab.sh refuses to combine with the image's VIRTUAL_ENV.
+      rm -f _isaac_sim
+      if [ -x /isaac-sim/python.sh ]; then ln -s /isaac-sim _isaac_sim; fi
       if [ -n \"\${WARP_CACHE_PATH:-}\" ]; then
         ./isaaclab.sh -p tools/verify_warp_cache.py
       fi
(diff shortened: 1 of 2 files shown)


hujc7:
@greptileai review

PR #7561, opened 2026-09-04
[Fix] Set Warp's global determinism mode under --deterministic

hujc7:
# Description

`--deterministic` did not make camera observations reproducible. Training `Isaac-Cartpole-Camera --deterministic` on an RTX PRO 6000 Blackwell splits into two bitwise-identical outcome clusters; on an L40 it does not. Reported as NVBug 6658578 against Isaac Lab 3.0 GA.

Newton's solvers accept a `deterministic` argument and apply it as a per-module option, so PR #7334 already covered the physics kernels. Its **sensor and geometry kernels take no such argument** and fall back to `warp.config.deterministic`, which stayed at `NOT_GUARANTEED` — Isaac Lab has never set it (`git log --all -S "config.deterministic"` is empty).

The concrete path: `newton._src.geometry.bvh.compute_enabled_shapes` claims output slots with `wp.atomic_add`, so the enabled-shape order — and with it the primitive order the scene BVH is built over — varies between processes. Ray queries then break ties differently, and a tiled camera renders a few pixels differently from bit-identical simulation state. That is enough to make an image-observation policy diverge.

`apply_env_overrides` now also requests `wp.config.deterministic = RUN_TO_RUN`. Warp reads the setting at module build time and the BVH is built during `ModelBuilder.finalize()`, so the request must land before the environment is created — `_apply_deterministic_request` at solver init is too late. An explicitly chosen mode such as `GPU_TO_GPU` is left alone.

## Type of change

## Validation

Where the divergence starts, from the real training path with every renderer input fingerprinted per frame (6 runs, Blackwell). All geometric inputs are bit-identical at the frame where the output already differs:

```
cam_xform / cam_rays / shape_transform / shape_body : never diverge
body_q, BVH bounds                                  : render 390  (after)
rendered colour (output)                            : render 389  (first)
```

`Isaac-Cartpole-Camera --deterministic`, epoch-25 reward, unique run dirs, 8 runs per arm:

| Branch | Configuration | Runs | Distinct outcomes |
|---|---|---|---|
| develop | before this change | 8 | 2 |
| develop | `wp.config.deterministic` set externally | 8 | 1 |
| develop | **this change**, plain `--deterministic` | 8 | **1** |
| release/3.0.0 | before | 8 | 2 |
| release/3.0.0 | `wp.config` set | 8 | 1 |

Architecture control on L40 (4gpu), 12 runs, unique run dirs: all identical — the defect does not reproduce on Ada, which is why PR #7334's L40-only validation missed it.

Unit tests: 31 pass in `test_entrypoints_common.py`, including three added here covering the request, an explicit mode being preserved, and no change without the flag.

Fixes NVBug 6658578.

## Follow-up (not in this PR)

Newton-side: sensor and geometry modules should accept a `deterministic` option like the solvers do, rather than depending on a process global; and `compute_enabled_shapes` produces order-dependent output regardless of that setting.

## Checklist


hujc7:
run-ci

source/isaaclab_rl/test/test_entrypoints_common.py
AntoineRichard, line 1: Are all these tests really required?

AntoineRichard (approved):
I'd double check if we really need all these tests. Otherwise looks good.

kellyguo11 (approved):

kellyguo11:
run-ci

PR #7563, opened 2026-09-04
Fix OVPhysX device routing for CPU-resident tensor types

AntoineRichard:
# Description

`_CPU_ONLY_TYPES` in `isaaclab_ov` is the single constant deciding whether an `OvPhysxView` tensor buffer is allocated on, and validated against, host or device memory. It omitted eight tensor types that are CPU-resident even on a GPU simulation:

| tensor type | alias | shape |
|---|---|---|
| `ARTICULATION_CONTACT_OFFSET` | `CONTACT_OFFSET` | `[N, S]` |
| `ARTICULATION_REST_OFFSET` | `REST_OFFSET` | `[N, S]` |
| `RIGID_BODY_CONTACT_OFFSET` | `RIGID_BODY_CONTACT_OFFSET` | `[N, S]` |
| `RIGID_BODY_REST_OFFSET` | `RIGID_BODY_REST_OFFSET` | `[N, S]` |
| `ARTICULATION_BODY_DISABLE_GRAVITY` | `BODY_DISABLE_GRAVITY` | `[N, L]` |
| `RIGID_BODY_DISABLE_GRAVITY` | `RIGID_BODY_DISABLE_GRAVITY` | `[N]` |
| `ARTICULATION_DOF_DRIVE_TYPE` | `DOF_DRIVE_TYPE` | `[N, D]` |
| `ARTICULATION_DOF_DRIVE_MODEL` | `DOF_DRIVE_MODEL` | `[N, D, 3]` |

Because `_native_device` consults this set, it returned the simulation device for host-resident data. Two consequences:

1. A hidden per-call host-to-device staging copy on every access, on a path the view believed was device-native. Silent, and not attributable to Isaac Lab in a profile. Values are still correct — ovphysx stages the mismatch transparently, so this is not a data-integrity bug.
2. `_check_device` rejected a **correctly** placed buffer: allocating one of these on the host, where the data actually lives, raised `OvPhysxView.DeviceMismatch`, whose message advises moving the buffer yourself — precisely the advice that fails here.

No shipped Isaac Lab workflow reads these eight through the tensor path today, so this is latent rather than a live outage. It is not an obscure corner: Isaac Lab already authors these same properties through USD schema writes at scene construction, and the tensor path exists to change such values per-env at runtime.

Each of the eight is also exposed as a public alias carrying its shape, dtype and units, taken from the wheel's own type table. This module is the alias layer over the ovphysx enum, so referencing them raw would have left the candidate tuple a mix of aliases and raw enum members with no stated reason why these eight differed. Naming follows the existing convention: the `ARTICULATION_` prefix is dropped for articulation types, as with `DOF_STIFFNESS` and `BODY_MASS`, and retained for rigid-body ones, as with `RIGID_BODY_MASS`.

A second, orthogonal gap surfaced during the audit: ovphysx documents `ARTICULATION_DOF_DRIVE_TYPE` as read-only, but it was missing from `_READ_ONLY_NAMES`, so `set_attribute` accepted writes and forwarded them. An audit of the full classification against the wheel's type table found this to be the only such gap — of the 21 types ovphysx documents read-only, 20 were already classified, with no misclassified members.

## Type of change

- **Breaking change** — writes to `articulation_dof_drive_type` now raise `OvPhysxView.ReadOnlyAttribute` instead of being silently forwarded. Migration: remove the write; drive type is authored through the USD drive schema, not the tensor path.

## Release backport

## Screenshots

Not applicable.

## Checklist

## How it was tested

ovphysx exposes no residency query (`ovphysx_tensor_spec_t` carries dtype/ndim/shape and no device field), so residency was **measured**, not assumed: CUDA memcpys were counted via CUPTI around each binding read into a host buffer and into a device buffer, under the DirectGPU configuration `isaaclab_ov` itself sets (`"/physics/suppressReadback": True`). A host read showing `DtoH` means GPU-resident; no memcpy on the host read plus `HtoD` on the device read means CPU-resident.

Measured across every GPU sample scene in the ovphysx 0.5.11 wheel:

| scene | before | after |
|---|---|---|
| `two_articulations_gpu` | `agree=40 MISMATCH=8` | `agree=48 MISMATCH=0` |
| `links_chain_sample_gpu` | `agree=40 MISMATCH=8` | `agree=48 MISMATCH=0` |
| `mixed_base_articulations_gpu` | `agree=36 MISMATCH=8` | `agree=44 MISMATCH=0` |
| `boxes_falling_on_groundplane_gpu` | `agree=9 MISMATCH=3` | `agree=12 MISMATCH=0` |
| `volume_deformable_multi` | `MISMATCH=0` | `MISMATCH=0` |
| `surface_deformable_material` | `MISMATCH=0` | `MISMATCH=0` |

Every previously classified member measured CPU-resident as declared in all scenes, so the set was incomplete rather than wrong — there were no false positives to correct.

The guarding test was a tautology: it compared the view's derived set against the canonical set it is derived from, so it could not fail for any contents. It now asserts against an independent inventory of measured residency, mirroring the existing `_EXPECTED_READ_ONLY_NAMES` pattern. Both new assertions were confirmed to fail before the fix and pass after it; the read-only case fails with `DID NOT RAISE ReadOnlyAttribute` without the fix.

- `pytest source/isaaclab_ov/test/sim/test_ovphysx_view.py` — 81 passed
- `test_articulation_helpers`, `test_rigid_object_helpers`, `test_deformable_object_helpers`, `test_ovphysx_scene_data_backend`, `test_randomize_rigid_body_material_mdp` — all pass
- `uv run isaaclab -f` — passes
- `tools/changelog/cli.py check develop` — passes

`test_views_xform_prim_ovphysx.py` reports 23 failed / 28 passed both with and without this change; every failure is the process-global device lock covered by the `device_split` marker, which requires per-device re-invocation. Baselined on a clean tree to confirm it is unrelated.

Environment: ovphysx 0.5.11, Linux x86_64, Python 3.12, NVIDIA RTX 5000 Ada Generation Laptop GPU.


kellyguo11 (approved):

kellyguo11:
run-ci

PR #7569, opened 2026-09-04
Support OVPhysX in randomize_rigid_body_collider_offsets MDP event

AntoineRichard:
# Description

> **Stacked on #7563.** This branch needs the `REST_OFFSET` / `CONTACT_OFFSET` / `RIGID_BODY_*_OFFSET` aliases and the CPU-only routing for those tensor types that #7563 adds. Review/merge that first; this diff collapses to a single commit once it lands, at which point the base should be retargeted to `develop`.

`randomize_rigid_body_collider_offsets` had no OVPhysX implementation. Its backend dispatch tested only for `"newton"` and fell through a bare `else` to `_RandomizeRigidBodyColliderOffsetsPhysx`, which calls `get_rest_offsets` / `get_contact_offsets` / `set_rest_offsets` / `set_contact_offsets` on `asset.root_view`. On OVPhysX that view is an `OvPhysxView`, which defines none of them, so adding the term to a config raised `AttributeError` during `OvPhysxManager.reset()` (via `PHYSICS_READY`) and the environment never built. The error surfaced from inside the PhysX implementation, so it did not name the unsupported backend either.

Found by the OVQA agent harness on Isaac-Cartpole (release/3.0.0, ovphysx 0.5.11). No shipped task uses the term today, so this is latent rather than a live regression.

**Fix**

- New `_RandomizeRigidBodyColliderOffsetsOvPhysx`, mirroring the PhysX variant. OVPhysX runs the PhysX solver, so rest/contact offsets are written directly, per collision shape, through the asset's `OvPhysxView`: `REST_OFFSET` / `CONTACT_OFFSET` for articulations, `RIGID_BODY_REST_OFFSET` / `RIGID_BODY_CONTACT_OFFSET` for rigid objects. Both bindings are CPU-resident `[N, S]` buffers, so the full tensor is read-modify-written on the host with the selected envs as write indices.
- Dispatch now matches `randomize_rigid_body_material` in the same file: `ovphysxmanager` first (it contains the substring `physx`), then Newton, then PhysX, and a `ValueError` naming the manager for anything else instead of silently selecting PhysX.
- Docstring lists OVPhysX as a supported backend.

**Note on write order**

PhysX enforces `restOffset < contactOffset` and on OVPhysX a violating `setRestOffset` is only logged, not raised. The term writes rest before contact, same as the PhysX variant, so raising both above a small authored contact offset can silently skip the rest write. Kept for PhysX parity; flagging in case we want to reorder in both backends.

**Verification**

New `source/isaaclab_ov/test/test_randomize_rigid_body_collider_offsets_mdp.py` drives the public term (stubbed `cfg` / `env` / `asset_cfg`) against a real OVPhysX `RigidObject` and `Articulation`, so the dispatch itself is covered. It reproduced the reported `AttributeError` before the fix. Assertions: selected envs land in the sampled range, unselected envs are untouched, and omitting one distribution leaves that offset alone. Passes on `cuda:0` and `cpu` (separate processes, ovphysx device lock). PhysX and Newton paths are unchanged.

Dependencies: #7563.

Fixes # (no tracking issue; reported by the OVQA agent harness)

## Type of change

## Release backport

## Screenshots

Not applicable: no visual change.

## Checklist


kellyguo11:
run-ci

PR #7571, opened 2026-09-04
Fix teleop recording success term

maxkra15:
## Summary

- keep an inert success termination registered while record_demos evaluates the original condition manually
- preserve reward-manager references without allowing automatic success resets during demonstration capture
- add a Franka Reach diffik startup regression test and changelog fragment

## Root cause

record_demos removed the success termination before gym.make. The Franka Reach reward configuration now references that term through is_terminated_term, so reward-manager initialization could no longer resolve the success key and raised the regular-expression mismatch.

## Testing

- verified the new regression test fails with the original behavior and the reported success: [] error
- verified the regression test passes with the fix
- uv run --frozen --extra teleop --with pytest python -m pytest source/isaaclab/test/cli/test_teleop_entrypoints.py -q
- ./isaaclab.sh -f

### DIFF
--- a/scripts/tools/record_demos.py
+++ b/scripts/tools/record_demos.py
@@ -200,6 +200,11 @@ logger = logging.getLogger(__name__)
 _CLOUDXR_ENV_SHORTHANDS: dict[str, str] = {}
 
 
+def _never_terminate(env: gym.Env) -> torch.Tensor:
+    """Return a false termination signal for every environment."""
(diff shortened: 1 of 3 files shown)


kellyguo11 (approved):

AntoineRichard (approved):

AntoineRichard:
run-ci

hougantc-nvda (approved):

PR #4716, opened 2026-02-25
Adds newton contact sensor

camevor:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context.
List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (existing functionality will not work without user modification)
- Documentation update

## Screenshots

Please attach before and after screenshots of the change if applicable.

## Checklist

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [ ] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the changelog and the corresponding version in the extension's `config/extension.toml` file
- [ ] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7574, opened 2026-09-04
Fix compute_first_contact/air missing transitions as the sensor clock ages

### LINKED ISSUE #7283: [Bug Report] compute_first_contact / compute_first_air miss most transitions: default abs_tol is far below the float32 timer error

AntoineRichard:
# Description

`ContactSensor.compute_first_contact(dt)` / `compute_first_air(dt)` silently missed most touchdowns and lift-offs once the simulation had run for a few seconds. On a transition step the contact (resp. air) timer is exactly one sensor update interval, so the check degenerated to `dt < dt + 1e-8`. The sensor clock is a float32 accumulator whose rounding error reaches ~1e-6 after a few seconds of simulated time, roughly 100x the default tolerance, so the events vanished depending on the magnitude of the sim clock. The reporter's CPU reproducer shows 352/500 touchdowns and lift-offs missed at the default tolerance; it reproduces exactly on `develop`. All three backends (Newton, PhysX, OVPhysX) share the same logic, and the 2.x torch sensor had the same defect, so this is long-standing rather than a Newton regression.

**Fix.** `abs_tol` now defaults to `None` and is resolved at call time by a single helper in `BaseContactSensor` to **half the sensor update interval**, i.e. `0.5 * max(cfg.update_period, physics_dt)`. Valid timer values are integer multiples of that interval, so half an interval is the midpoint between "one update ago" and "two updates ago". This stays robust to float32 clock drift for hours of simulated time and works for both `history_length == 0` (lazy, once-per-policy-step refresh) and `history_length > 0` (every physics substep). The Warp kernels are unchanged. Passing an explicit `abs_tol` still works; callers who want the previous behaviour can pass `abs_tol=1e-8`.

**Tests.** Public-API regression tests in the Newton, PhysX and OVPhysX contact sensor suites: settle a body in contact, age the sensor clock to 2.5 / 10 / 30 s, lift and land the body, and poll `compute_first_contact` / `compute_first_air` with default arguments. These fail on the unfixed code (e.g. `compute_first_air missed the lift-off at clock 2.833s: reported [], expected [0]`) and pass with the fix. The Newton variant covers both `history_length` modes.

**Refresh before query.** Both methods now call the sensor's outdated-buffer refresh before comparing, restoring what the 2.x torch sensor did by reading through `self.data`. The Warp port (#4707 / #4716) read the private timer buffer directly, so a `history_length == 0` caller that polled before touching `data` saw the previous step's timers. The aged-clock tests now poll before reading `data`; the Newton lazy-mode cases fail without the refresh (transition reported one policy step late) and pass with it. Adopted from the approach in #7294.

**Out of scope, noted for follow-up.**
- The transition interval is counted in both the ending and the starting phase, so `last_air_time` / `last_contact_time` overestimate each phase by one update interval (also raised in the issue). That bias is present since 2.x and baked into tuned `feet_air_time` thresholds, so changing it would rescale rewards for trained velocity policies and deserves its own PR.
- On `cuda:0` with `history_length == 0`, the PhysX and OVPhysX sensors serve stale net forces when buffers refresh only on data access, independently of this change. That is why the PhysX/OVPhysX public-API tests cover only the substep cadence.

Fixes #7283

## Type of change

## Release backport

## Checklist


aidaodedjl:
Thanks for the quick turnaround and for the archaeology on the 2.x torch sensor — good to know this predates the Newton backend entirely.

The half-interval default is the right minimal fix, and the reasoning for rejecting option (2) is airtight. One pushback on the reason for rejecting option (1), though, since #7294 implements exactly that path: the policy-step-visibility concern is solvable, and that PR solves it. The latches are cleared by the reset kernels and — the part that matters for substep refresh — `compute_first_contact()` / `compute_first_air()` refresh outdated buffers before reading, so a lazy caller never sees a latch from a previous policy step, and an every-substep caller never sees a stale one. The flags never need to cross a boundary the kernel cannot see; the query pulls them across.

Not arguing to swap fixes — the tolerance change is smaller, and the calibrated-thresholds argument for keeping the `last_*_time` bias out of this PR is the correct call. The question is whether the latching design is still of interest as a follow-up: beyond not resting on a comparison margin, it makes `compute_first_*` event-accurate rather than timer-inferred, which matters if anyone ever queries with a `dt` smaller than the update interval, and it composes with per-transition event ages (useful once the `last_*_time` bias gets revisited, since the correction will need exactly that distinction). If the answer is "the tolerance fix is all we want", I will close #7294 against this one with no hard feelings — it did its job getting the issue confirmed.

ClemensSchwarke (approved):

kellyguo11:
run-ci

PR #7588, opened 2026-09-04
Fix streaming_gt_types docstring and visualization docs polish

matthewtrepte:
## Summary

Fixes the `streaming_gt_types` docstring (it was missing the already-supported `"normals"` value and had a British spelling), which was the last open item from #7054's review. While in there, also:

- Gates Newton RTX against Kit-based physics backends (`physx`/`ovphysx`) with a clear error at startup instead of a process hang from a native crash, and documents the restriction.
- Reworks the 5 hero videos on the visualization docs page so the robot appears at a consistent size and position across all of them, and restacks the streaming-view videos vertically.
- Adds a troubleshooting section for the `NVST_R_BUSY` livestreaming/port-49100 conflict, with Linux and Windows recovery steps.

## Release backport

## Test plan
- [x] `uv run isaaclab -f` — lint, formatting, changelog checks all pass
- [x] `uv run --extra test python -m pytest source/isaaclab_visualizers/test/test_newton_adapter.py -v` — 54/54 pass, including a new regression test for the RTX gating
- [x] `uv run --no-project python tools/skills/cli.py check` — pass
- [x] Built the docs locally and visually checked the visualization and troubleshooting pages

### DIFF
--- a/docs/source/concepts/visualization.rst
+++ b/docs/source/concepts/visualization.rst
@@ -29,38 +29,39 @@ This page covers:
    .viz-grid-fit { justify-content:center; }
    .viz-grid-fit > div { flex:0 0 auto; }
    .viz-grid-fit video, .viz-grid-fit img { width:auto; height:488px; }
-   .viz-grid-natural { justify-content:center; }
-   .viz-grid-natural > div { flex:0 0 auto; }
-   .viz-grid-natural img { width:auto; height:260px; }
-   /* video (unlike img above) sizes by flexing to share the row instead of a fixed height:
-      these clips are much wider (near 16:9) than the marker screenshots img was tuned for, so
-      a fixed height risks overflowing the page width when 2 sit side by side. */
-   .viz-grid-natural > div:has(video) { flex:1 1 0; min-width:0; }
-   .viz-grid-natural video { width:100%; height:auto; max-height:260px; }
    .viz-grid-stretch video.viz-no-crop { object-fit:contain; background:#000; }
    .viz-grid-stretch video.viz-crop-bottom { object-position:center bottom; }
    .viz-grid-stretch video.viz-crop-kit-bottom { object-position:center 76%; }
    .viz-grid-stretch video.viz-crop-x8 { width:calc(100% + 16px); margin-left:-8px; }
    .viz-grid-stretch img.viz-crop-top { object-position:center 25%; }
    .viz-hero-wrap.viz-hero-newton-gl { aspect-ratio:960/397; }
-   .viz-hero-wrap.viz-hero-newton-gl video.viz-crop-newton-hero { width:calc(100% + 2px); height:calc(100% + 55px);
-                margin-left:-1px; margin-top:-40px; object-fit:cover; object-position:center 55.3%; }
+   .viz-hero-wrap.viz-hero-newton-gl video.viz-crop-newton-hero { width:calc(100% + 2px); height:calc(100% + 70px);
+                margin-left:-1px; margin-top:-55px; object-fit:cover; object-position:center 55.3%; }
    .viz-label.viz-label-raise { bottom:10px; }
    /* Trims pixels off each hero tile on top of whatever object-position crop is already
       applied, so the tile itself is shorter rather than just repositioning the existing crop.
-      Both classes crop to the same final 241px height so the 2x2 tile grid stays even, split
-      differently per tile: Kit/Rerun/Newton RTX crop 35px off the top and 10px off the bottom;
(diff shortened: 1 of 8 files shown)


matthewtrepte:
@isaaclab-review-bot review

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
```diff
@@ -1027,6 +1027,17 @@ def initialize(self, scene_data_provider: SceneDataProvider) -> None:
             return
 
         scene_data_provider = self._set_scene_data_provider(scene_data_provider)
+        if isinstance(self, NewtonRTXVisualizer) and self.physics_backend in ("physx", "ovphysx"):
```
kellyguo11: why ovphysx? shouldn't it be an issue with isaacsim_physx?

docs/source/concepts/visualization.rst
```diff
@@ -859,12 +861,33 @@ Limitations
...
 **Kit: incompatible with ovphysx / ovrtx presets**
 
 ``--viz kit`` cannot be used with ``presets=ovphysx`` or ``presets=ovrtx`` in the same process.
 Use ``--viz newton_gl``, ``--viz rerun``, or ``--viz viser`` with those presets, or omit
 ``--viz`` for headless execution.
 
+**Newton RTX: incompatible with Kit-based physics backends**
+
+``--viz newton_rtx`` raises a ``RuntimeError`` at startup if the active physics backend is
+``physx`` or ``ovphysx`` (e.g. ``presets=isaacsim_physx`` or ``presets=ovphysx``), since OVRTX is
```
kellyguo11: that's not correct - ovphysx should still be supported since it's kitless?

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
matthewtrepte: Good catch — you're right, ovphysx is itself kitless and doesn't require Kit. Only `physx` (i.e. `presets=isaacsim_physx`) actually conflicts with OVRTX. Fixed in cd7c6af4ff, along with the matching docs warning and a regression test asserting ovphysx does *not* trip the guard.

docs/source/concepts/visualization.rst
matthewtrepte: Fixed — narrowed this to just `presets=isaacsim_physx`; `presets=ovphysx` is kitless and remains supported with `--viz newton_rtx`. See cd7c6af4ff.

source/isaaclab_visualizers/isaaclab_visualizers/newton/newton_visualizer.py
kellyguo11: physx doesn't mean isaacsim_physx - physx does autoswitching, so we might need to be careful about that one. isaacsim_physx should be checked explicitly as well

matthewtrepte: Traced this through PhysxAutoCfg/_resolve_physx_auto_cfg and FactoryBase._get_backend(): both an explicit `physics=isaacsim_physx` and `physics=physx` (once the auto selector resolves to Kit) end up constructing the same `PhysxManager` class, so `physics_backend` currently reports `"physx"` for both at the point `initialize()` runs — the guard already covered the isaacsim_physx case via that string. Added `"isaacsim_physx"` to the check anyway per your request, as defense-in-depth in case another backend-name source ever reports the explicit selector string instead, and extended the regression test to cover both. See b68924c730.

docs/source/concepts/visualization.rst
```diff
@@ -859,12 +862,34 @@ Limitations
...
 **Kit: incompatible with ovphysx / ovrtx presets**
 
 ``--viz kit`` cannot be used with ``presets=ovphysx`` or ``presets=ovrtx`` in the same process.
 Use ``--viz newton_gl``, ``--viz rerun``, or ``--viz viser`` with those presets, or omit
 ``--viz`` for headless execution.
 
+**Newton RTX: incompatible with Kit**
+
+``--viz newton_rtx`` raises a ``RuntimeError`` at startup if the active physics backend is
+``physx`` (i.e. ``presets=isaacsim_physx``), since OVRTX is a kitless renderer and cannot share
```
kellyguo11:
```suggestion
``isaacsim_physx`` (i.e. ``presets=isaacsim_physx``), since OVRTX is a kitless renderer and cannot share
```

kellyguo11:
run-ci

kellyguo11:
run-ci

PR #7590, opened 2026-09-04
Close the Newton viewer before dropping the reference to it

fatimaanes:
# Description

`NewtonVisualizer` dropped its viewer without closing it in two paths:
- normal shutdown through `close()`
- the `step()` recovery path that disables the viewer after an initialization failure

This left OVRTX resources to be released by the garbage collector in an undefined order. If the renderer was destroyed before its active bindings and retained step results, shutdown could report:
```
Renderer destroyed with 1 active binding(s)
OV RTX: Leaking step result outputs
```
Both paths now use a shared `_release_viewer()` helper that calls `viewer.close()` before clearing the reference.

During normal shutdown, teardown failures continue to propagate after the remaining cleanup completes. If viewer cleanup also fails in the `step()` recovery path, the error is logged without interrupting training.

## Type of change
- Bug fix (non-breaking)

## Release backport

The same issue is present on `release/3.0.0`.

### Controlled teardown reproduction
The controlled test reproduced the reported shutdown signature:

| Version | Active-binding warnings | Leaked-step-result errors |
| --- | --- | --- |
| Without fix | 1 | 2 |
| With fix | 0 | 0 |

### Windows — RTX PRO 6000
Ran each of the two reported workloads three times, for six runs per batch:

| Version | Completed | Active-binding warnings | Leaked-step-result errors | New warnings |
| --- | --- | --- | --- | --- |
| Without fix | 6/6 | 1/6 | 0/6 | — |
| With fix | 6/6 | 0/6 | 0/6 | 0 |

The OVRTX window failed to initialize on this machine, so every run exercised the `step()` recovery path. A pre-existing USD asset-loading crash occurred once in each batch before viewer initialization and is being tracked separately.

### Linux — L40
The nine new regression tests fail against the original implementation and pass with the fix:

| Version | Visualizer test results |
| --- | --- |
| Without fix | 58 passed, 9 failed |
| With fix | 67 passed, 0 failed |

Seven tests requiring the full Isaac Sim runtime were excluded because of a pre-existing collection error.

## Checklist
- [ ] I have made corresponding changes to the documentation — not applicable
- [x] I have added the required changelog fragment under `source/<pkg>/changelog.d/`
- [x] My name is already included in `CONTRIBUTORS.md`

### DIFF
--- a/source/isaaclab_visualizers/changelog.d/fix-newton-visualizer-viewer-teardown.rst
+++ b/source/isaaclab_visualizers/changelog.d/fix-newton-visualizer-viewer-teardown.rst
@@ -1,0 +1,6 @@
+Fixed
+^^^^^
+
+* Fixed :class:`~isaaclab_visualizers.newton.newton_visualizer.NewtonVisualizer` releasing its viewer
+  without calling the viewer's :meth:`close`, which left the RTX backend's ordered GPU teardown to the
+  garbage collector and intermittently leaked render step results and attribute bindings on shutdown.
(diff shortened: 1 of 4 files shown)


kellyguo11:
run-ci

kellyguo11 (approved):

kellyguo11:
run-ci

PR #7599, opened 2026-09-05
Fix OvPhysX scene data joint name collisions

kellyguo11:
# Description

Fix repeated `Failed to find rigid body` warnings from the OVPhysX scene-data backend when rigid-body prims share leaf names with USD joint prims. This affects Ant's leg and foot bodies and produces one warning per colliding body per environment.

`OvPhysxSceneDataBackend.setup()` previously compressed every `RigidBodyAPI` prim into an environment-wildcard expression, including joint prims that carried `RigidBodyAPI`. It also did not account for same-named non-rigid joint prims when choosing wildcard expressions.

The backend now excludes joint prims from rigid-body candidates while retaining their names for collision detection. Same-named bodies use exact paths, while unrelated bodies retain wildcard compaction. Exact paths are fused into one OVPhysX binding per distinct environment-relative body path, so the number of native bindings remains independent of the environment count.

Related to #7358 and NVBug 6150853.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Release backport

## Testing

- Added a regression test covering joint-name collisions both with and without `RigidBodyAPI`, while verifying that an unrelated body still uses a wildcard and colliding bodies share one fused exact-path binding.
- Before the fix, the 128-environment Ant reproduction emitted 1,024 `Failed to find rigid body` warnings. After the fix, the same rollout completed with zero target warnings:

  ```bash
  uv run --isolated --frozen --extra ovphysx python scripts/environments/zero_agent.py \
    --task Isaac-Ant --num_envs 128 --max_steps 10 presets=physx --visualizer newton
  ```

- `uv run --isolated --frozen --extra ovphysx --extra test python -m pytest -q source/isaaclab_ov/test/physics/test_ovphysx_scene_data_backend.py` — 43 passed.
- `uv run isaaclab -f` passed all formatting and repository hooks against the current `develop` base.
- `uv run python tools/changelog/cli.py check develop --include-worktree` passed.

## Screenshots

Not applicable.

## Checklist

- [ ] I have made corresponding changes to the documentation (not applicable; no public API or workflow changed)

### DIFF
--- a/source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
+++ b/source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
@@ -91,11 +91,12 @@ class OvPhysxSceneDataBackend(SceneDataBackend):
     """Scene-data backend for the OVPhysX physics manager.
 
     Mirrors the contract of ``PhysxSceneDataBackend`` but adapts to the
-    ovphysx wheel's one-pattern-per-binding API: each distinct env-wildcard
-    rigid-body prim path produces its own ``TT.RIGID_BODY_POSE`` binding.
-    :attr:`transforms` reads each binding into its pre-allocated float32
-    staging buffer and concatenates them into a single ``wp.transformf``
-    array.
+    ovphysx wheel's binding API: each distinct env-wildcard rigid-body prim
+    path produces its own ``TT.RIGID_BODY_POSE`` binding. Bodies whose leaf
+    names collide with non-rigid prims use one fused exact-path binding per
+    distinct env-relative path. :attr:`transforms` reads each binding into
+    its pre-allocated float32 staging buffer and concatenates them into a
+    single ``wp.transformf`` array.
 
     The merged-buffer + staging-buffer separation is required because the
     wheel's ``TensorBinding.read(dst)`` writes into ``dst`` only when
@@ -143,7 +144,7 @@ class OvPhysxSceneDataBackend(SceneDataBackend):
         return paths
 
     def setup(self, physx, stage, device: str) -> None:
-        """Discover RigidBodyAPI prims, dedup by env-wildcard form, create one binding per pattern.
+        """Discover rigid bodies and create compact pose bindings without leaf-name collisions.
 
         Args:
             physx: Live ``ovphysx.PhysX`` instance (the wheel handle).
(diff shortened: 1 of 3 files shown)


kellyguo11:
run-ci

hujc7 (commented):
Superseded — re-posted as inline comments anchored to the relevant lines, so each finding sits on the code it concerns. See the inline thread(s) on this PR.

source/isaaclab_ov/isaaclab_ov/physics/ovphysx_manager.py
```diff
@@ -164,20 +165,44 @@ def setup(self, physx, stage, device: str) -> None:
...
+            # does not create one native binding per environment.
             total_count = 0
-            for pattern in sorted(patterns):
+            for pattern, exact_paths in binding_targets:
                 try:
-                    view = OvPhysxView(physx, pattern=pattern, device=device)
+                    if exact_paths is None:
+                        view = OvPhysxView(physx, pattern=pattern, device=device)
+                    else:
+                        view = OvPhysxView(physx, prim_paths=exact_paths, device=device)
```
hujc7, line 205: **[P0] Agent found bug:** this fused binding covers only `env_0`, so the PR removes the warnings by dropping the bindings that produced them.

Isaac-Ant at 128 envs — the configuration in the description — instrumented at head:

```
binding '/World/envs/env_*/Robot/torso'            row_count=128
binding '/World/envs/env_*/Robot/front_left_foot'  row_count=1
... 7 more colliding bodies, row_count=1 each

transform_count = 136        # expected 9 x 128 = 1152
```

**1016 of 1152 body poses are no longer bound**, and each colliding body renders frozen at its initial pose. The description reports 1,024 warnings removed — the warnings removed and the bindings lost are one for one.

**Cause.** In the kit-less lane `stage.Traverse()` above yields `env_0`'s bodies only: `env_1..N` roots exist with empty subtrees, because `clone_plan.py:227` gates `UsdReplicateContext` on `has_kit()`. So `exact_path_groups` collects a single path. The wheel is not implicated — a `prim_paths` list naming `env_1` alone binds 1 row, and a four-env list binds 4.

**Suggested direction.** Build the exact-path list by expanding the colliding body's env-relative suffix across the env root prims, rather than from the traversal. That keeps the design the description states: one fused binding per env-relative path, independent of the environment count.

- [ ] Isaac-Ant at 128 envs reports `transform_count == 1152` with zero warnings

The regression test cannot catch this as written — it authors `env_0` and `env_1` subtrees, which never both exist in this lane. Authoring only `env_0`'s subtree plus bare `env_1..N` roots would fail today and pass with the fix.

AntoineRichard:
@kellyguo11 this bug is on PhysX. They should not send this many warnings when joint and body names collide. Alternatively, we should also update the assets so that it's not the case. Letting us know that it ignored a body that happens to be named like a joint is cute but I'm not sure we care?

AntoineRichard:
I'm not sure we should be doing this PR. I told @marcodiiga about this, I think this needs to be addressed at the root rather than hacked in lab.

kellyguo11:
we will pick up the fix in ovphysx 0.6

PR #7610, opened 2026-09-06
[Tests] Consolidate installation CI environments

StafaH:
# Description

Streamlines the installation CI suite (`source/isaaclab/test/install_ci`) so that each expensive environment build carries every assertion for its install path, instead of rebuilding the same environment once per check. The x86 job currently spends 23 of its 25 minutes inside pytest, and six environment builds account for 80% of that.

**Changes**
- **System-Python smoke** (`cli/test_cli_install_core_in_globalenv_smoke.py`, renamed from `..._install_in_globalenv_smoke.py` per the directory naming rule): installs `-i core` instead of the bare `-i`, which resolves to `-i all` and pulled the full Isaac Sim stack into the system Python only to assert `import isaaclab`. The unique thing this test covers is the no-venv, PEP 668 install path; the optional features and Isaac Sim are already covered by the uv and wheel tests. This test was 469 s of the 1380 s pytest run.
- **uv `-i core`**: two files ran the identical `./isaaclab.sh -i core` install back to back (41 s each). `test_cli_install_in_uvenv_smoke.py::test_install_core_makes_assets_importable` now carries the union of both assertion sets (five core imports plus the unknown-install-token check); the copy in `test_cli_install_all_in_uvenv_training.py` is removed.
- **conda**: the conda `-i core` install (63 s) was a strict subset of the `-i newton,rl[rsl-rl]` install that followed it. The core import checks now run in the training env.
- **Cartpole probes**: the camera render and 2-iteration camera training probes ran twice on the same runner, after `-i all` and after `uv pip install <wheel>[all]`. They are about the runtime, not the installer, so they now run once on the wheel path; the `-i all` uv test keeps the shared state-training command (`_STATE_TRAIN_CMD` + `_assert_training_passed`, per the directory's AGENTS.md).
- **Wheel RL extras**: `uv_pip/test_uv_pip_install_isaaclab_rl_tasks_imports_rl_tasks.py` installed `<wheel>[sb3,skrl,rsl-rl]` into a second env to check three things the wheel-builder smoke (same extras) did not. Those three checks now live in `misc/test_wheel_builder_smoke.py` and the extra environment is gone.

Collected inventory goes from 62 to 60 tests; no assertion is dropped. Docstring `Setup:`/`Tests:` sections are updated to list exactly what runs.

## Validation

Ran the changed tests through the CI harness from a clean clone on an RTX PRO 6000 workstation (fresh pip and uv cache volumes, so absolute times are cold-cache):

```
tools/run_install_ci.py docker --gpu --build-wheel -- -sv \
  -k "Wheel_Builder or test_install_core_makes_assets_importable or test_install_all_trains_cartpole or Globalenv or Condaenv"
19 passed, 41 deselected in 734.97s
```

| Test | Time (cold cache) |
|---|---:|
| system-Python `-i core` + import | 358.9 s |
| conda `-i newton,rl[rsl-rl]` + core imports + train | 141.4 s |
| uv `-i all` + state train | 135.0 s |
| wheel-builder smoke (build, install, 15 checks) | 47.6 s class fixture |
| uv `-i core` (merged assertions) | 37.3 s |

`--collect-only` in docker mode collects 60 tests with the expected names. Full-suite before/after can only be measured on the CI runners (warm cache volumes), so the job total is the number to watch on this PR's Installation Tests run.

**Expected effect on the x86 job.** The conda and uv `-i core` removals and the dropped wheel env are direct savings (about 2 min). The camera probe removal saves the shader compile plus two camera iterations on one path. The system-Python change is a strict reduction in work but a smaller win than the payload size suggests: that path is slow because `./isaaclab.sh -i` drives plain `pip` through 12 sequential editable installs (the same `-i core` takes 30 s in a uv env), so most of its cost is per-package resolution rather than the Isaac Sim wheels. A follow-up could look at parallelizing the suite (the environments are already isolated) and at the 20 tests that currently skip because their guard imports `isaacsim` from the test runner's Python, which never has it inside the image.

## Type of change

- Test and CI performance improvement (non-breaking)

## Checklist


StafaH:
run-ci

StafaH:
run-ci

StafaH:
run-ci

ooctipus (approved):

PR #7614, opened 2026-09-07
Refresh PhysX and OVPhysX contact sensor buffers on every physics step

### LINKED ISSUE #7613: [Bug Report] PhysX GPU contact sensor keeps reporting the last in-contact force after contact is lost when history_length=0 and data is read lazily
### Describe the bug

On the PhysX backend on GPU, a `ContactSensor` with the default `history_length=0` and the default `InteractiveSceneCfg.lazy_sensor_update=True` keeps reporting the last in-contact force after the body has left contact, indefinitely, whenever its data is read less often than every physics step (for example once per policy step with `decimation=4`).

**Root cause (verified).** `RigidContactView.get_net_contact_forces()` on GPU zeroes a body's entry only on the exact physics step where that body's contact is lost. If the getter is not called on that step, the entry is never zeroed and keeps the last in-contact value until the next touchdown. With `history_length=0` the sensor only calls the getter when `.data` is accessed, so a policy-rate reader skips most contact-lost steps.

The Isaac Lab side is not at fault: the raw getter output, the cached Warp view over it, and the sensor's `net_forces_w` agree at every read, and disabling the CUDA-graph replay (`_use_graph = False`) changes nothing. CPU is not affected. `history_length > 0` is not affected because `BaseContactSensor.update()` then refreshes (and therefore calls the getter) on every physics step.

**Experiments** (cube settled on flat terrain, then held in the air; `sim_dt = 0.0025`, data read once per 4 physics steps, `cuda:0`):

| getter call cadence | force reported in the air |
|---|---|
| only on data access (every 4th step) | 1226 N, stuck for all subsequent steps |
| every physics step | 0.0 |
| once on the exact lift-off step, then every 4th step | 0.0 |
| once one step *after* lift-off, then every 4th step | 1226 N, stuck |

Teleporting the body is not the trigger: a single teleport into the air followed by free fall sticks as well.

### Steps to reproduce

Minimal pytest reproduction, built on the scaffolding of `source/isaaclab_physx/test/sensors/test_contact_sensor.py` (place it next to that file):

```python
import torch
import warp as wp
from test_contact_sensor import CUBE_CFG, FLAT_TERRAIN_CFG, ContactSensorSceneCfg, _perform_sim_step

from isaaclab.scene import InteractiveScene
from isaaclab.sensors import ContactSensorCfg
from isaaclab.sim import build_simulation_context

SIM_DT = 0.0025


def test_lazy_sensor_reports_contact_loss():
    with build_simulation_context(device="cuda:0", dt=SIM_DT, add_lighting=True) as sim:
        sim._app_control_on_stop_handle = None
        scene_cfg = ContactSensorSceneCfg(num_envs=1, env_spacing=1.0, lazy_sensor_update=True)
        scene_cfg.terrain = FLAT_TERRAIN_CFG
        scene_cfg.shape = CUBE_CFG
        scene_cfg.contact_sensor = ContactSensorCfg(
            prim_path=CUBE_CFG.prim_path, track_pose=True, update_period=0.0, track_air_time=True, history_length=0
        )
        scene = InteractiveScene(scene_cfg)
        sim.reset()
        sensor, shape = scene["contact_sensor"], scene["shape"]
        contact_pose = CUBE_CFG.contact_pose.to(device=shape.device).unsqueeze(0)
        air_pose = CUBE_CFG.non_contact_pose.to(device=shape.device).unsqueeze(0)

        trace = []
        for tag, pose in (("contact", contact_pose), ("air", air_pose)):
            for _ in range(6):
                for _ in range(4):  # policy step of 4 physics steps, data untouched in between
                    shape.write_root_pose_to_sim_index(root_pose=pose)
                    _perform_sim_step(sim, scene, SIM_DT)
                trace.append((tag, round(torch.norm(sensor.data.net_forces_w.torch, dim=-1).max().item(), 1)))
        print(trace)
        assert max(f for tag, f in trace[7:] if tag == "air") < 0.1, trace
```

Output on `cuda:0`:

```
[('contact', 613.7), ('contact', 1226.5), ('contact', 1226.3), ('contact', 1226.2), ('contact', 1226.2), ('contact', 1226.2),
 ('air', 1226.2), ('air', 1226.2), ('air', 1226.2), ('air', 1226.2), ('air', 1226.2), ('air', 1226.2)]
```

Passes on `cpu`, and passes on `cuda:0` with `history_length=1` or with the data read every physics step.

### System Info

- Commit: 3fbbd15dfcb (develop, 2026-09-04)
- Isaac Sim 6.0.1, PhysX backend (`isaaclab_physx`)
- OS: Ubuntu 22.04, RTX 5090 D, CUDA 12.9, driver 580

### Additional context

**Exposure.** Every shipped velocity task sets `history_length=3` and is unaffected. The Kuka Allegro and Franka lift tasks configure fingertip contact sensors with the default `history_length=0` and feed `normal_force_matrix_w` into observations, so on PhysX GPU those observations can hold a stale contact force after the finger releases the object. Any user config with the defaults is exposed.

**Fix direction.** The PhysX contact sensor should call its (cheap, eager, non-capturable) PhysX getters on every physics step inside `update()` regardless of `history_length`; the Warp kernels that consume the fetched buffers can stay lazy. OVPhysX wraps the same tensor API and probably needs the same treatment, but has not been checked. Newton is not affected.

Found while working on #7283 / #7574.

### Checklist

AntoineRichard:
# Description

On the PhysX and OVPhysX backends on GPU, a `ContactSensor` with the default `history_length=0` and the default `lazy_sensor_update=True` kept reporting the last in-contact force after the body had left contact, for as long as the body stayed in the air, whenever its data was read less often than every physics step (for example once per policy step with `decimation=4`).

**Root cause.** PhysX zeroes a body's entry in `get_net_contact_forces()` only on the exact physics step where that body's contact is lost. With `history_length=0` the sensor only called the getter when `.data` was accessed, so a policy-rate reader skipped most contact-lost steps and never saw the zero. The Isaac Lab kernels and the CUDA-graph replay are not involved: the raw getter output, the cached Warp view, and the sensor output agree at every read, and disabling the graph changes nothing. CPU is unaffected. `history_length > 0` is unaffected because the base class already refreshes every physics step. The experiment table and a standalone reproduction are in #7613; OVPhysX reproduces identically through its contact binding.

**Fix.** Both sensors override `update()` to fetch the physics-owned buffers on every physics step, skipping the fetch when the base class already refreshed on that step (history-bearing sensors, or `force_recompute`). The Warp kernels that turn the fetched buffers into sensor data still run lazily on data access, so the per-step cost is one getter call, which is exactly what history-bearing sensors already pay. The buffers are refreshed in place, so the cached views and the captured CUDA graph stay valid. On PhysX the outer-graph-capture guard moves into a helper shared by the eager fetch and the buffer update, so a capture is rejected on either path and the existing capture test keeps its "raise before touching PhysX" contract.

**Tests.** `test_lazy_sensor_reports_contact_loss` in both the PhysX and OVPhysX contact sensor suites: settle a cube for six policy steps of four physics steps each, reading the sensor once per policy step, then hold it in the air for six more, and require every air read after the first to be below 0.1 N. On the unfixed code it fails on `cuda:0` with the force stuck at the resting value for all six reads, and passes on `cpu`; with the fix it passes on both. The PhysX `test_contact_history_updates_at_sensor_period` row for `history_length=0` asserted the old zero-fetch cadence and was updated to the new one; its history-bearing rows are unchanged.

**Exposure.** Every shipped velocity task sets `history_length=3` and is unaffected. The Kuka Allegro and Franka lift tasks configure fingertip contact sensors with the default history length and feed the force matrix into observations, so on PhysX GPU those observations could hold a stale contact force after a finger released the object.

Found while working on #7283 / #7574.

Fixes #7613

## Type of change

## Release backport

## Checklist


ClemensSchwarke (commented):
Is the overhead of reading the pose every step significant? If so, maybe address the bot comment. Otherwise looks good.

kellyguo11:
run-ci

AntoineRichard:
@ClemensSchwarke Measured it on the RTX 5090 workstation rather than guessing. Short answer: not significant, and the pose read is now out of the per-step path anyway (21280d6).

**Micro-timing inside the live env** (`Isaac-Velocity-Flat-AnymalD`, PhysX GPU, 4096 envs x 17 bodies, `history_length=0`, decimation 4, device-synced, median of 200 calls):

| per physics step | no pose | `track_pose=True` (061d759) | `track_pose=True` (21280d6) |
|---|---|---|---|
| `get_net_contact_forces` | 26 us | 31 us | 26 us |
| `get_transforms` (pose) | n/a | 32 us | 32 us |
| full fetch helper (data access) | 28 us | 51 us | 51 us |
| `sensor.update()` (added per-step work) | 49 us | 73 us | 48 us |
| `sim.step` | 3.85 ms | 3.89 ms | 3.99 ms |
| `env.step` (4 physics steps) | 16.5 ms | 16.7 ms | 16.4 ms |

So the PR adds four extra fetches per env step: about 0.1 ms, or 0.7% of a 16.5 ms env step, for a `history_length=0` sensor. With `track_pose=True` that was 0.2 ms (1.2%) at 061d759; after 21280d6 the per-step refresh passes `include_pose=False` on both backends and it is back to 0.1 ms. Poses are still read on data access, as before. `track_pose` defaults to `False` and no shipped task enables it.

**End-to-end runtime benchmark**, same machine, `isaaclab benchmark runtime`, 1000 steps + 50 warmup, seed 42, `physics=isaacsim_physx`, base (57cc1cd) vs PR head (061d759), one fresh `uv` env per worktree, strictly sequential runs with alternating order:

| task | n per side | base FPS | PR FPS | delta (1 SE) |
|---|---|---|---|---|
| Flat-AnymalD, shipped (`history_length=3`) | 8 | 238.0k | 239.0k | +0.4% +/- 2.3% |
| Flat-AnymalD, `history_length=0` | 8 | 241.4k | 240.3k | -0.5% +/- 1.6% |
| Flat-AnymalD, `history_length=0`, `track_pose=True` | 8 | 242.8k | 237.4k | -2.2% +/- 1.5% |
| Rough-AnymalD, shipped (`history_length=3`) | 3 | 50.1k | 48.8k | -2.5% +/- 9.9% |
| Rough-AnymalD, `history_length=0` | 3 | 47.1k | 47.7k | +1.3% +/- 3.3% |

Every delta is within one standard error of zero. Run-to-run scatter is 3-4% on flat and ~10% on rough, so end-to-end can only bound the effect; the micro-timing above is the actual number. The shipped velocity tasks use `history_length=3` and already fetched every step before this PR, so the first row is the expected null result.

The `history_length=0` rows are variants of the shipped AnymalD configs registered by a small launcher around `isaaclab.benchmark.entrypoints.runtime.run` that only overrides `scene.contact_forces.history_length` / `track_pose`.

Both bot comments are addressed in 21280d6: pose skipped in the per-step refresh (PhysX and OVPhysX), and the new capture-time `RuntimeError` for lazy history-less PhysX sensors is documented in the `update()` docstring and the changelog fragment.

source/isaaclab_ov/isaaclab_ov/sensors/contact_sensor/contact_sensor.py
```diff
@@ -395,6 +405,26 @@ def _update_buffers_impl(self, env_mask: wp.array | None = None) -> None:
...
+
+        Args:
+            dt: Time elapsed since the previous sensor update [s].
+            force_recompute: Whether to recompute the sensor buffers regardless of their
+                configured update period. Defaults to False.
+        """
+        super().update(dt, force_recompute=force_recompute)
+        # Skip the read if the base class already refreshed the buffers on this step, which it
+        # does when the sensor carries a history buffer.
+        if self._is_initialized and self._data_generation != self._data_generation_last_update:
```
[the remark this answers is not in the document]
AntoineRichard: Done in 21280d6: the per-step refresh in `update()` now calls the fetch helper with `include_pose=False` on both the PhysX and OVPhysX sensors, so poses are only read on data access as before. Measured on an RTX 5090 at 4096 envs x 17 bodies: the pose read is 32 us per physics step, and `sensor.update()` with `track_pose=True` went from 73 us to 48 us, matching the no-pose case.

source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
```diff
@@ -190,6 +190,26 @@ def contact_view(self) -> physx.RigidContactView:
...
+        Args:
+            dt: Time elapsed since the previous sensor update [s].
+            force_recompute: Whether to recompute the sensor buffers regardless of their
+                configured update period. Defaults to False.
+        """
+        super().update(dt, force_recompute=force_recompute)
+        # Skip the fetch if the base class already refreshed the buffers on this step, which it
+        # does when the sensor carries a history buffer.
+        if self._is_initialized and self._data_generation != self._data_generation_last_update:
+            self._fetch_physx_buffers()
```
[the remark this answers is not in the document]
AntoineRichard: Done in 21280d6: `update()` now documents the `RuntimeError` raised under an outer CUDA graph capture, and the PhysX changelog fragment notes that lazy history-less sensors now reject a capture like history-bearing sensors already did, with the guidance to update the sensor outside the capture.

AntoineRichard:
run-ci

ClemensSchwarke (approved):

AntoineRichard:
run-ci

ooctipus (approved):

marcodiiga:
run-ci

source/isaaclab_physx/isaaclab_physx/sensors/contact_sensor/contact_sensor.py
```diff
@@ -190,6 +190,31 @@ def contact_view(self) -> physx.RigidContactView:
...
+
+        Raises:
+            RuntimeError: If an outer CUDA graph capture is active. The PhysX tensor reads
+                cannot be graph-captured, so the sensor must be updated outside the capture.
+        """
+        super().update(dt, force_recompute=force_recompute)
+        # Skip the fetch if the base class already refreshed the buffers on this step, which it
+        # does when the sensor carries a history buffer.
+        if self._is_initialized and self._data_generation != self._data_generation_last_update:
+            self._fetch_physx_buffers(include_pose=False)
```
marcodiiga, line 216: Non-blocking test follow-up: please extend `test_contact_sensor_refuses_update_inside_outer_graph_capture` to exercise public `update(dt)` with `history_length=0` and `force_recompute=False`. The existing test calls `_update_buffers_impl()` and replaces `_fetch_physx_buffers()`, so it does not protect the new eager path shown here. Keep the real fetch helper and assert that the capture raises before any backend contact getter is called.

source/isaaclab_physx/test/sensors/test_contact_sensor.py
```diff
@@ -809,6 +816,60 @@ def count_fetches() -> None:
...
+        # read once per policy step and left untouched in between.
+        decimation = 4
+        num_policy_steps = 6
+
+        def run_policy_step(root_pose: torch.Tensor) -> float:
+            """Holds the shape at ``root_pose`` for one policy step and reads the sensor once."""
+            for _ in range(decimation):
+                shape.write_root_pose_to_sim_index(root_pose=root_pose)
+                _perform_sim_step(sim, scene, sim_dt)
+            return torch.linalg.norm(sensor.data.net_normal_forces_w.torch, dim=-1).max().item()
```
marcodiiga, line 861: Non-blocking test follow-up for both backend suites: this regression has no partner filter and only checks `net_normal_forces_w`, so it does not verify filtered contact loss. Please extend the contact-loss coverage with a filtered body pair and assert that `normal_force_matrix_w` also returns to zero when read once per four physics steps, with a nonzero-contact assertion first.

An ovphysx GPU check with two 0.5 m cubes, no terrain, zero gravity, and 0.05 m overlap followed by 10 m separation cleared both outputs with this fix. Restoring the old update cadence left both outputs stuck around 4687 N. This is worth retaining as a regression alongside the ground-contact case.

source/isaaclab_physx/test/sensors/test_contact_sensor.py
```diff
@@ -800,7 +805,9 @@ def count_fetches() -> None:
         assert last_update == pytest.approx(expected_last_update)
 
         _ = contact_sensor.data
-        expected_fetches_after_read = 1 if history_length == 0 else expected_fetches
+        # A history-less sensor did not recompute during the steps, so the read triggers one
+        # more fetch; a history-bearing one is already up to date.
+        expected_fetches_after_read = expected_fetches + 1 if history_length == 0 else expected_fetches
```
marcodiiga, line 810: Non-blocking performance follow-up for both backends: the extra fetch asserted here means that a lazy, history-less sensor fetches contacts in `update()` and fetches them again when `.data` is read in the same generation. With decimation 1, this doubles contact-buffer retrievals per physics step. Please consider tracking which sensor generation already fetched contacts so data recomputation can reuse them, while still fetching poses lazily and invalidating that state on reset. The existing cadence test can then protect the single-fetch behavior.

marcodiiga (approved):
Reviewed at 2bd5e60d8bc51fda9d4ba87347d85f3fcbe780fd. The per-step contact refresh fixes the stale-force regression while preserving sensor output cadence and avoiding duplicate fetches after eager updates. I found no new correctness blocker.

Validation: 14 focused PhysX tests passed, including CPU/GPU contact loss, history cadence, reset, and CUDA-graph behavior. The ovphysx contact-loss regression passed on CPU and GPU. A GPU two-cube release check also cleared both net and filtered forces; restoring the old update cadence reproduced stale forces in both the ground and two-cube cases. The [previous Docker CI reports](https://github.com/isaac-sim/IsaacLab/actions/runs/34483763910) for this same head also show all four new contact-loss cases passing.

Approving with three non-blocking follow-ups inline (one redundant-fetch optimization and two regression-test additions); please address them despite the approval. The [fresh Docker CI run](https://github.com/isaac-sim/IsaacLab/actions/runs/34591681485) is still in progress. The [existing Multi-GPU failure](https://github.com/isaac-sim/IsaacLab/actions/runs/34483722816/job/102892828006) occurred during runner setup, before checkout or test execution.

PR #7621, opened 2026-09-07
[Backport develop] Fix Franka Reach relative IK teleoperation (#7619)

kellyguo11:
# Description

Backports #7619 from `release/3.0.0` to `develop`.

The change restores native keyboard, gamepad, and SpaceMouse teleoperation for both supported Franka Reach relative-IK configurations:

- `physics=isaacsim_physx presets=diffik`
- `physics=newton_mjwarp presets=newton_ik`

Both configurations accept six-dimensional relative pose actions, so their native device configurations disable the unsupported gripper command. The default joint-position and absolute-pose presets remain unchanged.

This is an exact patch replay of merged release commit `17e5f3d51620168806647feeaeab85c2688d2bc4`; no conflict resolution or develop-specific adaptation was required. No new dependencies are required.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Documentation update

## Release backport

- [ ]  This change originated on `release/3.0.0`; do not backport it again.

## Screenshots

Not applicable.

## Validation

- Stable patch ID matches the merged #7619 release commit.
- `uvx --from pre-commit==4.6.2 pre-commit run --all-files` — passed.
- `tools/changelog/cli.py check develop` — passed against the current develop base.
- `git diff --check upstream/develop..HEAD` — passed.
- Source PR validation for this exact patch: 23 Reach preset tests passed, the Newton random-agent smoke test passed, the documentation build passed without warnings, and `uv run isaaclab -f` passed.

The targeted test and documentation commands could not be rerun on this macOS host because the current lockfile supports Linux x86_64/aarch64 and Windows AMD64 only. Develop CI provides the supported-platform verification.

## Checklist

- [x] I have read and understood the contribution guidelines.
- [x] I have run the available pre-commit checks.
- [x] I have included the corresponding documentation change.
- [x] My changes generate no new warnings in the available checks.
- [x] The source PR's tests cover this exact patch.
- [x] I have included the `isaaclab_tasks` changelog fragment.
- [x] The original contributor already appears in `CONTRIBUTORS.md`.

PR #7624, opened 2026-09-07
Fix Franka Reach OSC velocity limit and normalize OSC pose targets

maxkra15:
# Description

Follow-up to #7033. The state-based benchmark run hit `ValueError: The rewards returned by the environment contain NaN values` on `Isaac-Reach-Franka-OSC` (ovphysx, `presets=diffik_abs`, one seed of three). Investigating the OSC task and controller turned up three defects that together make the task run far outside its intended envelope:

1. **The OSC arm ran without a solver joint velocity limit.** The zero-gain effort actuator introduced in #7033 copied the deprecated `velocity_limit_sim` alias from the Menagerie actuator config, but that config sets `joint_velocity_limit`. The copied value was `None`, and the Menagerie USD authors no joint velocity limit of its own, so the live limit resolved to `FLT_MAX` (plain `Isaac-Reach-Franka` keeps 20 / 25 rad/s). Instrumented training showed the arm reaching ~60 rad/s under saturated 100 N·m efforts with self-collisions enabled, which is the regime where PhysX can blow up into NaN. The actuator now forwards `joint_velocity_limit`; with the fix the peak joint speed in training clamps at 25 rad/s.
2. **`OperationalSpaceController` used `pose_abs` quaternions unnormalized.** The policy emits four raw values and the axis-angle error scales with their norm (a norm of 3.6 yields an 11 rad "orientation error"), so the commanded efforts scaled with the raw output too. The controller now normalizes the target and falls back to the current end-effector orientation (or identity) for degenerate quaternions, matching `DifferentialIKController`.
3. **`diffik_abs` leaked into the OSC task.** The OSC action term replaces the parent's arm-controller presets, so `presets=diffik_abs` on the OSC task only zeroed the action-magnitude reward weight, which is the only regularizer on the raw pose/quaternion outputs. This explains why only the `diffik_abs` rows were affected. The OSC task now resolves the default weight. `presets=diffik_abs` stays accepted as a deprecated no-op alias that emits a `FutureWarning` at config validation, so existing command lines keep working; the alias will be removed in a future release. Migration: drop `presets=diffik_abs` from `Isaac-Reach-Franka-OSC` commands.

While rebasing onto `develop`, #7621 had added a `teleop_devices` preset keyed on `diffik` / `newton_ik` to `Isaac-Reach-Franka`; it leaked into the OSC task the same way, so the OSC config now resolves that to the default (empty) device set as well.

The exact NaN event did not reproduce locally (two full seed-43 runs on ovphysx completed cleanly), so the physics blow-up is inferred from the runaway joint velocities rather than caught in the act.

## Type of change

- Deprecation of the `diffik_abs` preset on `Isaac-Reach-Franka-OSC` (no-op alias with `FutureWarning`)

## Release backport

## Validation

- `uv run --extra dev python -m pytest source/isaaclab/test/controllers/test_operational_space_features.py` — 4 passed (2 new; both fail without the controller fix)
- `uv run --extra dev python -m pytest source/isaaclab_tasks/test/core/test_reach_franka_presets.py` — 29 passed on top of `develop` (3 new + 3 new preset cases; the new tests fail without the config fix)
- `uv run --extra dev --extra isaacsim python -m pytest source/isaaclab/test/controllers/test_operational_space.py` — 18 passed
- `Isaac-Reach-Franka-OSC presets=diffik_abs --max_iterations 1 --num_envs 4` through the hydra CLI: `FutureWarning` emitted at env creation, training runs
- 40-iteration `Isaac-Reach-Franka-OSC physics=ovphysx` training: no NaN, peak arm joint speed 25.1 rad/s (was ~59 rad/s before)
- `uv run python tools/changelog/cli.py check develop` and `uv run isaaclab -f` pass

## Checklist


maxkra15:
run-ci

maxkra15:
run-ci

kellyguo11 (approved):

kellyguo11:
@hujc7 for viz, we should make sure the controller side changes are picked up when we move to Newton controllers

kellyguo11:
run-ci

PR #7628, opened 2026-09-08
[Docker] Point uv at the shipped environment in the kit-less image

hujc7:
# Description

`uv run` inside the kit-less container ignored the environment the image ships and built a
second one from scratch. Where the source tree is bind-mounted from the host — Compose,
CI, and the documented developer workflow all do this — the editable rebuild wrote
`egg-info` into files the image's uid cannot touch and the command died with
`Permission denied`. Reported as NVBug 6732972 against Isaac Sim 6.1.0 rc26 /
Isaac Lab 3.0.0 EA, on both x86_64 and aarch64.

## 1. Summary

* `uv run isaaclab train …` works inside the kit-less image again, instead of failing with
  `could not create '<package>.egg-info': Permission denied`.
* Removes a silent 188-package reinstall (plus a downloaded CPython) from every in-container
  `uv` invocation.

## 2. Cause

`docker/Dockerfile.kitless` is the only multi-stage image in the repo, and `ENV` does not
cross stages. Its `builder` stage sets `UV_PROJECT_ENVIRONMENT`, `UV_PYTHON` and
`UV_PYTHON_PREFERENCE`; the `runtime` stage sets neither. `Dockerfile.base` and
`Dockerfile.curobo` are single-stage, so their copies survive to runtime — which is why this
reproduces only on the kit-less image.

uv resolves the project environment from `UV_PROJECT_ENVIRONMENT` alone and does not honour
`VIRTUAL_ENV`, so every in-container `uv run` emitted

```
warning: `VIRTUAL_ENV=/opt/isaaclab-venv` does not match the project environment path `.venv`
         and will be ignored; use `--active` to target the active environment instead
```

and then built `/workspace/isaaclab/.venv` from the lock, rebuilding all 12 editable path
packages in place under `source/`.

Introduced by #6946, which moved the runtime venv from `${ISAACLAB_PATH}/.venv` — uv's own
default project path, which is why this worked before — to `/opt/isaaclab-venv`.

## 3. Fix

Four `ENV` lines in the runtime stage:

| Variable | Without it |
|---|---|
| `UV_PROJECT_ENVIRONMENT` | uv targets `${ISAACLAB_PATH}/.venv` and installs 188 packages |
| `UV_PYTHON`, `UV_PYTHON_PREFERENCE` | `pyproject` pins `python-preference = "only-managed"`, so uv downloads its own CPython and **replaces** the venv rather than reusing it |
| `UV_NO_SYNC` | `uv run` still rebuilds the 12 editable packages in place, so the `egg-info` write still fails on a mounted source tree |

All four are required: the first three alone leave the reported failure intact (§4).

## 4. Validation

Measured against a rebuilt image, with `source/` bind-mounted read-only from the host as CI
does. QA's verbatim command, on an 8x NVIDIA L40 node:

```
uv run --extra ov isaaclab train --rl_library rsl_rl --task Isaac-Cartpole-Camera-Direct \
    physics=ovphysx renderer=ovrtx presets=rgb --max_iterations 5
```

rc=0, 0 tracebacks, 0 `egg-info` errors, 0 uv project-environment warnings,
`Training time: 85.08 seconds`. This is QA's SMOKE TEST at their `--max_iterations 5`, kept
verbatim as the bug repro; its reward and success-rate numbers are artifacts of a 5-iteration
run and are not policy results.

Arm-by-arm, same image and mounts:

| Image | `uv run --extra ov …` |
|---|---|
| unfixed | fails — `egg-info … Permission denied` |
| + `UV_PROJECT_ENVIRONMENT`, `UV_PYTHON`, `UV_PYTHON_PREFERENCE` | fails — same error |
| + `UV_NO_SYNC` (this PR) | rc=0, `sys.prefix=/opt/isaaclab-venv`, no `.venv` created |

`pre-commit run --all-files` clean; `docker/test` 38 passed, 1 skipped.

## 5. Test added

Every step in `.github/actions/validate-kitless-image/action.yml` invoked `python` directly or
`uv pip install --python "$VIRTUAL_ENV/bin/python"`; nothing ran `uv run`, which is why CI
stayed green while the documented user command was broken. This PR adds a `uv run` step
asserting `sys.prefix == $VIRTUAL_ENV` and that no `${ISAACLAB_PATH}/.venv` appears. Verified
to fail on the unfixed image and pass on the fixed one.

`kitless-docker.yml`'s change gate already lists both `docker/Dockerfile.kitless` and
`.github/actions/validate-kitless-image/`, so this PR builds and validates the image with no
workflow changes.

## Type of change

## Release backport

## Checklist

- [x] No documentation change is required — the documented commands are unchanged; this restores them working inside the image
- [x] My changes generate no new warnings — it removes one uv emitted on every in-container run


hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

hujc7:
run-ci

docker/Dockerfile.kitless
```diff
@@ -3,128 +3,45 @@
...
-ENV LANG=C.UTF-8
-# Build arg like Dockerfile.base, so the image home and the Compose mounts
-# derived from DOCKER_USER_HOME cannot drift apart.
-ARG DOCKER_USER_HOME_ARG=/home/isaaclab
-ARG VENV_PATH_ARG=/opt/isaaclab-venv
-ENV DOCKER_USER_HOME=${DOCKER_USER_HOME_ARG}
-ENV HOME=${DOCKER_USER_HOME}
-ENV VIRTUAL_ENV=${VENV_PATH_ARG}
-ENV PATH=${VENV_PATH_ARG}/bin:/usr/local/bin:${PATH}
+ENV UV_NO_SYNC=1
```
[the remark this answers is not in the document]
hujc7: Fixed in `5174febe85c`. Confirmed the failure first — in the built image `uv run --extra leapp python -c "import leapp"` gave `ModuleNotFoundError: No module named 'leapp'`, and `leapp` is indeed not in `--extra all`.

`UV_NO_SYNC` is no longer set in the image. It moved to the only caller that needs it, the validate action, which is the one place that bind-mounts `source/` read-only:

```
docker run --rm -e UV_NO_SYNC=1 "${mount_args[@]}" "$IMAGE_TAG" ...
```

Checked the blast radius before moving it: `run_tests.sh` (multi-GPU smoke, arm-ci) drives the image through `./isaaclab.sh -p`, never `uv run`, so it is unaffected.

After the change, in the image: `uv run --extra leapp` installs 16 packages and imports cleanly, and with the source mounted read-only plus caller-side `UV_NO_SYNC=1`, `uv run` still resolves `/opt/isaaclab-venv` with no stray `.venv`.

.github/workflows/kitless-docker.yml
```diff
@@ -137,6 +143,7 @@ jobs:
         isaacsim-version: "24.04"
         dockerfile-path: docker/Dockerfile.kitless
         cache-tag: cache-kitless
+        verify-test-path: docker/test/test_image_invariants.py
```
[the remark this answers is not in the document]
hujc7: Fixed in `5174febe85c`. Added to the change patterns in `kitless-docker.yml`, alongside the two test files already selected there:

```
^docker/test/test_image_invariants\.py$ :: Built-image invariant tests
```

docker/Dockerfile.kitless
```diff
@@ -141,22 +58,64 @@ RUN apt-get update \
...
+
+# From the lock, not ``isaaclab.sh --install``: only the lock applies ``[tool.uv]
+# override-dependencies``, which pins ``packaging`` above ovphysx's ``<24``.
+# ``--frozen`` stops uv re-resolving a lock it judges stale; ``--inexact`` keeps the seeded pip.
+# Cache mount as in Dockerfile.base and Dockerfile.curobo, but under ${HOME}: this image sets
+# HOME globally, so uv caches there rather than in root's home, and an unmatched mount bakes the
+# whole cache into the layer. UV_CACHE_DIR pins it so the two cannot drift.
+# ovrtx writes shader caches inside its own site-packages, which a volume-mounted run reaches as
+# the runner's uid rather than the image's.
+RUN --mount=type=cache,target=/home/isaaclab/.cache/uv \
```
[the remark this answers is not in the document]
hujc7: Fixed in `5174febe85c`. Confirmed the mechanism in the built image before changing anything — `/home/isaaclab/.cache` was `root:root 755` while `/home/isaaclab` is `isaaclab:isaaclab 750`, exactly as described.

`${HOME}/.cache` is now pre-created in the runtime-user `RUN` so BuildKit finds the mount-point parent already present and owned:

```
&& mkdir -p "${HOME}/.cache" "${VIRTUAL_ENV}" "${ISAACLAB_PATH}" \
&& chown -R isaaclab:isaaclab "${HOME}" \
```

After the change it is `isaaclab:isaaclab 755`.

One correction to the impact: runtime `uv` was not actually broken before this, because `.cache` was mode 755 and `.cache/uv` itself was already `isaaclab`-owned — `uv pip install` returned rc=0. The defect was real but the consequence was a root-owned directory inside a 750 home rather than a write failure.

.github/workflows/kitless-docker.yml
```diff
@@ -104,6 +104,12 @@ jobs:
         fetch-depth: 1
         lfs: true
 
+    # The GPU runners have no uv on PATH; the invariant check below needs it.
+    - name: Set up uv
+      uses: astral-sh/setup-uv@v6
```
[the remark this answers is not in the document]
hujc7: Not changing this here, and flagging why rather than silently skipping it.

The concern is legitimate — a mutable tag on a self-hosted runner that later authenticates to ECR is real supply-chain exposure. But `astral-sh/setup-uv@v6` is unpinned in 8 places across 6 workflows today:

```
.github/workflows/build.yaml:180,206
.github/workflows/kitless-docker.yml:72,109
.github/workflows/docs.yaml:62,94
.github/workflows/wheel.yml:150
.github/workflows/license-check.yaml:53
```

Pinning only the line this PR happens to add would leave the same exposure on every other job while making this file inconsistent with its neighbours — including `build.yaml:206`, which is the step this one was copied from. Repository-wide action pinning is worth doing, but as its own change with a policy decision behind it, not as a side effect of a kit-less Docker fix.

The line itself was added because `verify-test-path` runs `uv run` on the runner and the GPU runners have no `uv` on `PATH` — `build.yaml:204` documents the same requirement.

hujc7:
@greptileai review

kellyguo11 (commented):
The root-cause analysis looks sound, but this has expanded beyond the NVBug and the current validation injects a condition absent from the normal user workflow. I left focused inline comments on scope and regression coverage.

docker/Dockerfile.kitless
```diff
@@ -3,128 +3,46 @@
...
 
-# Kit-less image for GPU-accelerated training without Isaac Sim or Kit:
-# Newton physics, OVRTX rendering, and every core RL framework.
+# Kit-less image: Newton physics, OVRTX rendering and the RL frameworks, without Isaac Sim or Kit.
 ARG BASE_IMAGE_ARG=ubuntu:24.04
 
 FROM ghcr.io/astral-sh/uv:0.12.9@sha256:8b940d3a9d65bed080436972241af2e21c84b5e8c9193f7014ed71479ee795ff AS uv
 
-FROM ${BASE_IMAGE_ARG} AS builder
+FROM ${BASE_IMAGE_ARG}
```
kellyguo11: Could we keep NVBug 6732972 scoped to propagating the runtime uv environment and split this single-stage rebuild into a follow-up? Converting the image architecture also changes toolchain retention, COPY ownership, caching, and aarch64 build behavior. Those changes are not needed to prove the reported `uv run` regression is fixed, and they substantially increase the review and backport surface.

docker/Dockerfile.kitless
```diff
@@ -3,128 +3,46 @@
...
+ENV HOME=${DOCKER_USER_HOME}
+ENV VIRTUAL_ENV=${VENV_PATH_ARG}
+ENV PATH=${VENV_PATH_ARG}/bin:/usr/local/bin:${PATH}
 ENV DEBIAN_FRONTEND=noninteractive
 ENV LANG=C.UTF-8
-ENV VIRTUAL_ENV=${VENV_PATH_ARG}
+
+# uv reads UV_PROJECT_ENVIRONMENT, never VIRTUAL_ENV. UV_PYTHON overrides pyproject's
+# ``python-preference = "only-managed"``, which would fetch a CPython and rebuild the venv.
+# UV_NO_SYNC is deliberately NOT set here: it would stop ``uv run --extra <name>`` installing an
```
kellyguo11: The PR description says `UV_NO_SYNC` is required and reports that the other three settings alone still reproduced the permission error, but the current image deliberately omits it. Please reconcile the implementation and description and demonstrate that the normal kit-less user path succeeds without an injected variable. Otherwise this revision may still allow `uv run` to rebuild editable packages in a bind-mounted source tree.

.github/actions/validate-kitless-image/action.yml
```diff
@@ -43,6 +43,23 @@ runs:
       print("core RL frameworks available:", ", ".join(modules))
       '
 
+      # Nothing else in this action runs ``uv run``, the documented entry point -- every other
+      # step calls ``python`` directly, which is how nvbugs 6732972 stayed green here.
+      # UV_NO_SYNC belongs to the caller that mounts the source read-only, not to the image: in
+      # the image it would also stop ``uv run --extra <name>`` installing an extra the image does
+      # not carry. Without it the editable rebuild fails on the read-only mount.
+      docker run --rm -e UV_NO_SYNC=1 "${mount_args[@]}" "$IMAGE_TAG" \
```
kellyguo11: Injecting `UV_NO_SYNC=1` here makes the check pass under a condition the image and Compose user path do not provide, so it can mask the original regression. Please exercise QA's exact `uv run --extra ov isaaclab train ... --max_iterations 5` command under the shipped environment without this override, or add the setting to the actual user path. The regression test should fail whenever the documented workflow would still attempt editable rebuilds.

hujc7:
@greptileai review

kellyguo11 (approved):

kellyguo11:
run-ci

docker/Dockerfile.kitless
hujc7: Done — reverted in `569bd8b4844`. Back to the two-stage image with the four `uv` settings and the `uv run` check: 3 files, +38 lines. Single stage, cache mount, arm64 deps and image invariants moved to #7671 (draft). Agreed on backport surface especially.

docker/Dockerfile.kitless
hujc7: Resolved by the revert: `ENV UV_NO_SYNC=1` is back at `docker/Dockerfile.kitless:121`, so "all four are required" is accurate again.

Omitting it was a real defect. Measured with no injected variable, source bind-mounted, container uid 1000 against a host checkout owned by another uid:

```
error: Cannot update time stamp of directory 'isaaclab_visualizers.egg-info'
```

`docker/docker-compose.yaml` bind-mounts `../source` for this service, so Compose users would have hit it. With the setting restored, the same call resolves `/opt/isaaclab-venv` and creates no `.venv`.

.github/actions/validate-kitless-image/action.yml
hujc7: Agreed — the injection was in the reverted commits, so the action runs `uv run --extra ov` under the shipped environment with no override, and fails if the image stops suppressing the rebuild.

Your QA-command point stands: the check asserts `sys.prefix` and the absence of `${ISAACLAB_PATH}/.venv`, not a full `isaaclab train ... --max_iterations 5`. Say if you want that here rather than #7671.

docker/Dockerfile.kitless
```diff
@@ -3,128 +3,46 @@
...
+# extra the image does not carry, such as the documented ``--extra leapp``. Callers that mount
+# the source read-only pass it themselves.
```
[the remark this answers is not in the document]
hujc7: Correct — confirmed `docker/docker-compose.yaml` bind-mounts `../source` for the kit-less service, and without `UV_NO_SYNC` a uid-1000 container on a foreign-owned checkout reproduces `Cannot update time stamp of directory 'isaaclab_visualizers.egg-info'`. `569bd8b4844` restores it in the image; the extras trade-off moves to #7671.

PR #7651, opened 2026-09-08
Install ovrtx-extensions from the wheelhouse and enable OVRTX crash upload

mataylor-nvidia:
# Description

Adds `ovrtx-extensions` as a second wheelhouse package next to `ovphysx`, and uses it to turn on Omniverse crash-report upload for the OVRTX runtime.

`ovrtx-extensions` is published to NVIDIA's internal Artifactory index ([package](https://artifactory.nvidia.com/ui/packages/pypi:%2F%2Fovrtx-extensions?name=ovrtx&type=packages)) rather than PyPI or `pypi.nvidia.com`, so the wheelhouse is the only path that can reach it. That path is already gated to `develop` (`USE_OVPHYSX_WHEELHOUSE`), which keeps this off pull requests from forks and off every release branch.

**What changed**

- `.github/workflows/build.yaml` — `wheelhouse-packages` becomes `ovphysx ovrtx-extensions` in the three jobs that already mount the wheelhouse: `isaaclab_ov`, `rendering-correctness-kitless (legacy)`, and `rendering-correctness-kitless (ovstage)`.
- `.github/actions/run-tests/run_tests.sh` — sets `ISAACLAB_OVRTX_CRASH_UPLOAD=1` on the test container when, and only when, the wheelhouse package list contains `ovrtx-extensions`.
- `source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_crash_reporter.py` — applies the `--/crashreporter/...` settings through `ovrtx_extensions.apply_settings`, gated on that variable.
- `source/isaaclab_ov/isaaclab_ov/renderers/ovrtx_renderer.py` — calls it from `OVRTXRenderer.__init__`, after `redirect_shader_cache` and before `Renderer(OVRTX_CONFIG)`.
- `.github/workflows/config.yaml` — comment recording what the wheelhouse image has to contain, since it is empty today and a name missing from it fails the job under `pip --no-index`.
- Tests in `source/isaaclab_ov/test/test_ovrtx_crash_reporter.py`, plus a changelog fragment for `isaaclab_ov`.

**Settings applied**

```
--/crashreporter/enabled=true
--/crashreporter/url="https://services.nvidia.com/submit"
--/crashreporter/product="Omniverse.ovrtx"
--/crashreporter/version="<importlib.metadata.version('ovrtx')>"
--/crashreporter/preserveDump=true
--/crashreporter/gatherUserStory=false
--/crashreporter/devOnlyOverridePrivacyAndForceUpload=true
--/crashreporter/alwaysUpload=true
```

**Why the settings are applied at renderer construction**

The first revision of this PR ran the shim from a setup step in `run_tests.sh`. That was wrong: carb settings are process-local, so they died with the short-lived helper, and the pytest process that actually builds the renderer — and actually crashes — was never configured. Fixed in `baee515dc5` by moving the call into `OVRTXRenderer.__init__`.

It takes the `RendererConfig` for the same reason `redirect_shader_cache` does. Applying a setting can be what first loads the ovrtx library, `ovrtx_initialize` runs exactly once per process with the config of whichever caller loads it first (`ovrtx._src.bindings._LibraryLoader.create_bindings`), and `ovrtx_extensions` has no config of its own to hand it. Loading the bindings with the renderer's real config first is what keeps the `log_file_path`, `log_level` and `keep_system_alive` that `tools/ovrtx_log.py` replays per test.

**Still inert until the wheelhouse image is set.** `ovphysx_wheelhouse_image` is `""` on `develop` today, so the wheelhouse path is disabled and the env var is never set. It also needs an image that ships an `ovrtx-extensions` wheel plus its dependencies — `pip --no-index` cannot fall back to an index.

## Type of change

## Release backport

## Checklist

`source/isaaclab_ov/test/test_ovrtx_crash_reporter.py` covers the settings contract, that every setting is applied, that the renderer's config reaches the applier, that an unset variable touches nothing, and that an explicit request without `ovrtx_extensions` raises. It runs with no GPU, no ovrtx runtime and no renderer, mirroring `test_ovrtx_shader_cache_redirect.py`.


mataylor-nvidia:
run-ci

PR #7653, opened 2026-09-08
Fix benign OmniRtx*API schema errors with newton_rtx + ovphysx

matthewtrepte:
## Summary

- Fixes benign `[omni.rtx] FindAppliedAPIPrimDefinition(...) returned nothing` errors when `newton_rtx` is used after another backend has initialized USD.
- Treats the current `newton_rtx` viewer and `OVRTXRendererCfg` as OVRTX runtime consumers during the existing config scan.
- Rejects incompatible Isaac Sim / Kit combinations before loading OVRTX, then calls OVRTX's own `register_schema_paths()` once before physics or user code can initialize USD.
- Removes the public Isaac Lab setup helper, duplicate path resolution, visualizer-to-`isaaclab_ov` dependency, and best-effort fallback. Registration failures now fail at the selected runtime boundary.
- Removes the obsolete `NewtonViewerRTX` environment-path workaround; pinned OVRTX 0.4.1 resolves its native library and schema paths itself.

The resulting PR is +21 LOC overall and -18 production LOC against its develop base.

## Root cause

USD's plugin registry is initialized once per process. `ovrtx.register_schema_paths()` must therefore run before a physics backend or renderer first accesses USD. Previously, OVPhysX could initialize USD before Newton constructed `ViewerRTX`, leaving the `OmniRtx*API` schemas undiscoverable.

OVRTX already owns native-library discovery and schema-path publication. Isaac Lab only needs to invoke that primitive at the launch composition root after validating that no incompatible Kit runtime was selected.

## Test plan

- `uv run --frozen --extra test python -m pytest source/isaaclab/test/app/test_launch_simulation_require_kit.py -q` — 9 passed.
- Related launcher and visualizer suites — 74 passed.
- Verified a config-declared `newton_rtx` registers OVRTX before storage setup and user code.
- Verified CLI `newton_rtx` with Isaac Sim PhysX is rejected before OVRTX is loaded.
- With `LD_LIBRARY_PATH` and `OMNI_USD_PLUGINS_BASE_PATH` unset, verified `ovrtx.register_schema_paths()` publishes the USD plugin path and `ovrtx.Renderer()` constructs successfully.
- Changed-file formatting, lint, spelling, RST, and changelog checks pass.

## Release backport


kellyguo11:
run-ci

source/isaaclab/isaaclab/app/sim_launcher.py
```diff
@@ -517,6 +529,22 @@ def launch_simulation(
...
+        # one-shot-per-process initialization. If that happens first, ovrtx's later `import ovrtx`
+        # (inside newton's ViewerRTX, once the newton_rtx visualizer is constructed) is too late to
+        # add its schema paths, and applying OmniRtx*API schemas on the render product logs benign
+        # "FindAppliedAPIPrimDefinition(...) returned nothing" errors. Best-effort: an install
+        # without isaaclab_ov (or without the optional ovrtx extra inside it) simply skips this,
+        # since it is a log-noise mitigation, not something a newton_rtx launch depends on.
+        try:
+            from isaaclab_ov.renderers import prepare_ovrtx_runtime
+        except ImportError:
+            pass
```
huidongc: Should we fail silently?

source/isaaclab/isaaclab/app/sim_launcher.py
```diff
@@ -517,6 +529,22 @@ def launch_simulation(
...
+        else:
+            prepare_ovrtx_runtime()
```
huidongc: I don't know a better solution but still feel this is probably an anti pattern because the core now tries to import a renderer module. 

It seems we need to enforce some sort of initialization order so that ovrtx gets warmed up before ovphysx.

huidongc: @ooctipus for your thoughts

matthewtrepte: @huidongc 

since sim_launcher.py does already import OvPhysxCfg/OVRTXRendererCfg from isaaclab_ov at the top, so the try/except around prepare_ovrtx_runtime wasn't really guarding anything, so removed it.

checked the rest of core too, this file is actually the only one that imports isaaclab_ov/isaaclab_physx/isaaclab_newton eagerly. everywhere else does it lazily inside functions to stay optional.

making it lazy too would be cleaner but that's a bigger change than this PR, might be worth a follow-up

huidongc (approved):

kellyguo11:
run-ci

kellyguo11 (approved):

matthewtrepte:
run-ci

PR #7658, opened 2026-09-09
Fix newton_gl visualizer CUDA illegal access with --device cpu

matthewtrepte:
## Summary

Fixes the CUDA illegal-memory access when a CPU simulation uses the `newton_gl` visualizer on a machine with CUDA present.

```bash
uv run --extra all python scripts/environments/zero_agent.py \
  --task IsaacContrib-Stack-Cube-UR10-Short-Suction-IK-Rel \
  --visualizer newton_gl physics=isaacsim_physx \
  --device cpu --num_envs 16
```

## Root cause

Several allocations implicitly used Warp's process-global default device. Warp defaults to `cuda:0` when CUDA is available, even when the simulation, its published scene data, or the consuming viewer is on CPU. This mixed CPU pointers with CUDA mappings and render buffers.

## Changes

- `SimulationContext` selects the configured PyTorch and Warp process device before constructing physics, rendering, or visualization backends.
- `SceneDataProvider` derives transform allocations and launches from the transform publication and geometry mappings from the points publication. Matching formats retain the zero-copy passthrough.
- Newton marker meshes and instances allocate on `viewer.device`; registration remains owned and cached by each viewer.
- Rebuilding a Newton shadow model invalidates its prior body mapping at the same lifecycle boundary.
- Removes device ownership from `PhysicsManager`, per-frame device guessing from the Newton mapping cache, redundant tests, and unrelated `uv.lock` changes.

The resulting PR is +31 net production lines versus +68 before this cleanup, and +112 total lines versus +369.

## Validation

- Exact reported workflow, bounded with `--max_steps 5`: exited successfully with no CUDA illegal-access errors.
- Affected scene-data, marker, Newton visualization-state, and Newton adapter suites: **117 passed**.
- `SimulationContext` device-ordering cases on CPU and `cuda:0`: **2 passed**; the assertion runs inside `PhysxManager.initialize`, proving selection happens before backend setup.
- Focused Ruff lint and format hooks: passed.

## Release backport


kellyguo11:
run-ci

PR #7673, opened 2026-09-09
fix(newton): resolve ArticulationData's per-instance axis for the flat builder

pascal-roth:
# Description

`ArticulationData._create_simulation_bindings` (in `isaaclab_newton`) indexed every per-instance
sim binding (root pose, joint positions/velocities/limits/targets, body mass/inertia/wrench,
tendon properties — about 20 sites) with a hardcoded `[:, 0]`. That only holds for the replicated
builder's `(num_worlds, 1, ...)` layout. The flat (non-replicated) builder used when
`replicate_physics=False` produces the opposite layout, `(1, num_instances, ...)`, so `[:, 0]`
silently collapsed every non-replicated scene with more than one environment down to a single
instance's data for every binding, including write paths used to actually drive the robot (joint
position/velocity targets, effort, etc).

The fix resolves the per-instance axis dynamically from the root transforms' shape instead of
hardcoding it, and reuses that axis for every binding in the method.

**Scope note / follow-up:** the same `[:, 0]` pattern also exists in `rigid_object_data.py`,
`cable_object_data.py`, and `rigid_object_collection.py`. Those are **not** fixed by this PR — I
don't have a scene available to exercise them, so I'm flagging it as a known follow-up rather than
changing unverified code.

**Repro / test plan:** reproduced against a custom multi-environment task with
`replicate_physics=False`. Before the fix, `root_quat_w` (and other root/body/joint properties)
returned shape `(1, ...)` instead of `(num_envs, ...)`; after the fix it correctly returns
`(num_envs, ...)`.

Fixes # (issue)
N/A, found and fixed during development — no linked issue.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Release backport

- [ ]  Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

N/A — backend indexing fix, no visual output.

## Checklist

Docker and GPU tests run on demand. Push the commits you want tested, then
comment `run-ci` on the pull request.

- [x] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
[...]

PR #7672, opened 2026-09-09
fix(newton): split flat-builder site injection per environment

pascal-roth:
# Description

`_cl_inject_sites_fallback` — the non-replicated/flat-builder path in `isaaclab_newton`'s
`NewtonManager`, used whenever a scene is built with `replicate_physics=False` — packed every
environment's matched site index into a single one-element outer list instead of one sublist per
environment. Any consumer indexing `per_world[env_index]` for `env_index > 0` (e.g.
`NewtonRaycastSensor._resolve_site_indices`) hit an `IndexError` as soon as a scene had more than
one environment with `replicate_physics=False`.

The fix groups matches by their `env_<N>` path segment instead, so the site map has one sublist
per environment, matching the convention already used by the replicated builder path.

Fixes # (issue)

**Test plan:** Reproduced against a custom multi-environment task with `replicate_physics=False`
and a robot-attached ray-caster sensor; confirmed the `IndexError` before the fix and clean
initialization after. Not covered by an automated CI test yet.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Release backport

- [ ]  Backport this pull request to the active release branch after it merges into `develop`

## Screenshots

N/A — backend indexing fix, no visual output.

## Checklist

Docker and GPU tests run on demand. Push the commits you want tested, then
comment `run-ci` on the pull request.

- [ ] I have read and understood the [contribution guidelines](https://isaac-sim.github.io/IsaacLab/main/source/refs/contributing.html)
- [x] I have run the [`pre-commit` checks](https://pre-commit.com/) with `./isaaclab.sh --format`
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added a changelog fragment under `source/<pkg>/changelog.d/` for every touched package (do **not** edit `CHANGELOG.rst` or bump `extension.toml` — CI handles that)
- [x] I have added my name to the `CONTRIBUTORS.md` or my name already exists there

PR #7675, opened 2026-09-09
fix(newton): resolve environment-vs-world index in the FK/reset-mask kernels

pascal-roth:
# Description

`NewtonManager.invalidate_fk` and `invalidate_body_state` (in `isaaclab_newton`) mark environments
as needing FK recomputation and solver reset by writing `world_mask`/`fk_mask` through four Warp
kernels (`_or_reset_masks_from_mask`, `_scatter_reset_masks_from_ids`,
`_or_world_reset_mask_from_mask`, `_scatter_world_reset_mask_from_ids`) that all conflated the
environment index with the world index.

That is correct for the replicated builder (`world_count == num_envs`, one world per environment)
but wrong for the flat (non-replicated) builder used when `replicate_physics=False`, where every
environment shares a single world 0 and the environment index is carried by the articulation
("arti") axis instead. Treating an environment index as a world index there wrote out of bounds
into `world_mask` (sized `world_count == 1`), producing an "illegal memory access" CUDA error on
every `env.reset()` call for any non-replicated scene with more than one environment.

The fix resolves the builder convention once at launch time (`world_count == 1`) and passes it
into each kernel so it indexes the correct axis.

This is the third bug in the same family (environment-index-vs-world-index confusion in Newton's
flat/non-replicated builder), alongside #7672 and #7673.

**Reproduction / verification:** reproduced against a custom multi-environment task with
`replicate_physics=False`: `env.reset()` crashed with a CUDA illegal-memory-access error before the
fix, confirmed synchronously with `CUDA_LAUNCH_BLOCKING=1`, which pinpointed the crash to
`_scatter_reset_masks_from_ids`. After the fix, the same task runs a full episode and records video
with no errors.

Fixes # (issue)

## Type of change

## Release backport

## Screenshots

N/A — backend kernel fix, no visual output.

## Checklist

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
+++ b/source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
@@ -276,12 +276,21 @@ class _ParticleVisualPrim:
 def _or_reset_masks_from_mask(
     env_mask: wp.array(dtype=wp.bool),
     articulation_ids: wp.array2d(dtype=int),
+    is_flat_builder: int,
     world_mask: wp.array(dtype=wp.bool),
     fk_mask: wp.array(dtype=wp.bool),
 ):
-    """OR env_mask into world_mask and set corresponding articulation bits in fk_mask."""
+    """OR env_mask into world_mask and set corresponding articulation bits in fk_mask.
+
+    ``articulation_ids`` maps ``(world, arti)`` to a model articulation index. The replicated
+    builder gives one world per environment (``world == env_id``, any ``arti`` count); the flat
+    (non-replicated) builder collapses every environment into a single world 0, with the
+    environment index carried by ``arti`` instead. ``is_flat_builder`` (``world_count == 1``,
+    resolved once at launch time) selects which axis is the environment index into ``env_mask``.
+    """
     world, arti = wp.tid()
-    if env_mask[world]:
+    env_id = wp.where(is_flat_builder, arti, world)
+    if env_mask[env_id]:
         world_mask[world] = True
         fk_mask[articulation_ids[world, arti]] = True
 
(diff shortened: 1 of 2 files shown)


source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1515,19 +1544,20 @@ def invalidate_body_state(
         cls._mark_transforms_dirty()
         if cls._world_reset_mask is None:
             return
+        is_flat_builder = int(cls._world_reset_mask.shape[0] == 1 and (cls._num_envs or 0) > 1)
```
hujc7, line 1547: **[P0] Agent found bug:** Always `0`, so the `invalidate_body_state` half is dead. `_world_reset_mask` is allocated `world_count + 1` long (`:1685`) — a single-world model has length 2, so `shape[0] == 1` is unreachable; and `_num_envs` is `None`/`1` on this path (measured), so the second conjunct fails independently. With `_num_envs` recorded, `world_count == 1 and num_envs > 1` evaluates `True` on the same scene.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -290,14 +299,20 @@ def _or_reset_masks_from_mask(
...
+
+    See :func:`_or_reset_masks_from_mask` for the flat-vs-replicated ``(world, arti)`` convention
+    that ``is_flat_builder`` resolves.
+    """
     i, arti = wp.tid()
-    world = wp.int32(env_ids[i])
-    world_mask[world] = True
+    env_id = wp.int32(env_ids[i])
+    world = wp.where(is_flat_builder, 0, env_id)
     fk_mask[articulation_ids[world, arti]] = True
```
hujc7, line 314: **[P0] Agent found bug:** Measured on a flat Cartpole × 4 envs with this PR: `invalidate_fk(env_ids=[1])` → `fk_reset_mask=[1, 1, 1, 1]` — every articulation flagged for one requested environment. The launch still spans `(len(env_ids), articulation_ids.shape[1])`, and in the flat layout `arti` is the environment axis, so each requested env marks the whole row. The mask path is correct (`env_mask=[0,1,0,0]` → `[0, 1, 0, 0]`). Setting the shared world's flag is right; the FK flags must select only the requested environments' articulations — either an explicit env→articulation mapping, or the narrower contract stated and enforced (one instance per env, in env order), in which case `arti == env_id` is the filter.

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -1480,18 +1507,20 @@ def invalidate_fk(
             return
 
         if articulation_ids is not None and env_mask is not None:
+            is_flat_builder = int(articulation_ids.shape[0] == 1 and articulation_ids.shape[1] > 1)
```
hujc7, line 1510: **[P1] Agent review:** From reading: `shape[0] == 1 and shape[1] > 1` cannot distinguish "N flat environments" (measured here: `articulation_ids.shape=(1, 4)`) from "one environment whose view selects M articulations" — the latter would read `env_mask[arti]` past a length-1 mask. Same root cause as above: the flat/replicated distinction needs to be recorded, not inferred from a shape.

hujc7 (changes requested):
**Agent review:** Mapping environments onto their world is necessary, but one of the four kernels over-marks in the flat layout and the guard for one pair can never fire. Details inline; please add regression tests for sparse `env_ids`, mask-based selection, and >1 selected articulation per environment.

Issue #7676, opened 2026-09-09
[Bug Report] Camera pose writes do not reach the renderer on the Newton backend


alexmillane:
### Describe the bug

`Camera.set_world_poses` moves the camera on PhysX but not on Newton. Under Newton, `camera.data.pos_w` reports the new pose while the rendered image stays at the old one.

### Steps to reproduce

Run the [regression test here](https://github.com/alexmillane/IsaacLab/pull/4/files):

```
ISAACLAB_TEST_SAVE_IMAGES=1 uv run python -m pytest \
  source/isaaclab/test/sensors/test_camera_pose_write_across_backends.py -v -s
```

It moves a downward-looking camera from 2 m to 8 m over a ground plane and checks both consequences of the write, on both backends:

```
test_camera_pose_write_moves_reported_pose[physx]  PASSED
test_camera_pose_write_moves_reported_pose[newton] PASSED
test_camera_pose_write_moves_render[physx]         PASSED
test_camera_pose_write_moves_render[newton]        FAILED
  Far depth (2.08 m) should be > 1.5x close depth (2.08 m); got 1.00x.
```

The test saves some images for inspection

**PhysX**

<img width="256" height="128" alt="Image" src="https://github.com/user-attachments/assets/a6b2aa4b-c4b0-40df-aa41-50f36111d78a" /> 

<img width="256" height="128" alt="Image" src="https://github.com/user-attachments/assets/149e6c32-1044-4e20-8595-0bd3ee8cf500" />

Images correctly change in response to moving the camera from 2m -> 8m

**Newton**

<img width="256" height="128" alt="Image" src="https://github.com/user-attachments/assets/a9371872-31f8-4ba2-a888-e876ed514bc4" />

<img width="256" height="128" alt="Image" src="https://github.com/user-attachments/assets/b7a0b7a4-f687-4cef-98d8-551546df38f7" />

No change from moving the camera from 2m -> 8m

### System Info

- Commit: `8c3bc5e8e`
- Isaac Sim Version: 6.0.1
- OS: Ubuntu 22.04
- GPU: RTX 5090
- CUDA: 12.2
- GPU Driver: 580.65.06

### Additional context

[...]

PR #7691, opened 2026-09-09
Mirror Newton frame-view pose writes onto Fabric transforms

mataylor-nvidia:
## Description

`Camera.set_world_poses` moves the camera on PhysX but not on Newton: `camera.data.pos_w` reports the new pose while the rendered image stays at the old one.

The write goes through the sensor's `FrameView`. Under PhysX that view is Fabric-backed, so the RTX renderer which reads `omni:fabric:worldMatrix` follows it. Under Newton the view is a `NewtonSiteFrameView`, and the write lands in Warp site state that nothing mirrors onto the prim.

`NewtonSiteFrameView` now retains each site's destination prim path and, when a transform writer scope exits, stamps the site world poses onto the prims' Fabric world **and** local matrices. The local matrix matters: Newton's body sync ends in a Fabric hierarchy pass that forward-propagates `parent * local` onto non-rigid-body prims, which would otherwise overwrite the world matrix on the next render for any frame parented under a body.

Frames parented to a Newton body already tracked that body through that same hierarchy pass. Only the frame's own write was missing. That behaviour is now pinned by a test.

Two constraints are load-bearing and documented at the point they apply:

- **No seeding of Newton-driven prims.** `SetWorldXformFromUsd()` resets a prim to its spawn pose, because Newton never writes poses back to USD. Only genuinely absent attributes are created and seeded, and parents never receive a local matrix.
- **Writes must go through the read-write selection.** Refreshing it is what notifies the renderer. Writing through the read-only one still lands the matrix in Fabric. Test that reads the matrix back passes, while the image keeps showing the old pose.

Coverage is partial by design: Newton clones physics without cloning USD, so a site can be real while its destination prim exists on no stage, and those have nothing to render.

### Refactor

The Fabric prim tagging, selections, and view-to-slot mapping this needs already existed inside `FabricFrameView`. They move to a new shared helper, `FabricXformSelection` in `isaaclab/sim/views/fabric_xform_selection.py`, used by both backends, with `seed_from_usd` distinguishing a view that owns Fabric from one that only pushes state onto it. The helper is deliberately not exported from `isaaclab.sim.views`: both backends import it by module path, so it stays an internal cross-backend helper rather than public API. `FabricFrameView` loses 259 lines and keeps its private surface intact via read-through properties.

## Type of change

- This change requires a documentation update: no

## Release backport

## Screenshots

`test_camera_pose_write_moves_render` from gh issue https://github.com/isaac-sim/IsaacLab/issues/7676, run with `ISAACLAB_TEST_SAVE_IMAGES=1`. A downward-looking camera is moved from 2 m to 8 m over the ground plane by `Camera.set_world_poses` now working for both backends. The orange marker sits under the camera; the blue and green ones are outside the frame at 2 m and only come into view at 8 m. Depth is greyscaled against a fixed 12 m scale, so a camera that did not move would render identically at both heights.

| | 2 m | 8 m |
| --- | --- | --- |
| PhysX RGB |  |  |
| Newton RGB |  |  |
| PhysX depth |  |  |
| Newton depth |  |  |

Both backends report mean visible depth 1.929 m at 2 m and 8.290 m at 8 m, a 4.30x ratio against the test's 1.5x threshold. The Newton and PhysX frames are byte-identical at both heights. Before this change the Newton renders stayed at the 2 m pose.

## Checklist

### DIFF
--- a/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
+++ b/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
@@ -294,16 +294,12 @@ class NewtonSiteFrameView(BaseFrameView):
                     self._site_label_scales.append(scale)
                     self._site_label_prim_paths.append(spec_paths)
                 else:
-                    # A lone pattern is the cloned ``.*`` form, whose one label covers every
-                    # environment; explicit per-environment patterns take one path each.
-                    single_pattern = len(body_patterns) == 1
-                    for index, body_pattern in enumerate(body_patterns):
+                    # Only reached before the model exists, where a spec always resolves to one
+                    # pattern -- the cloned ``.*`` form -- whose label covers every environment.
+                    for body_pattern in body_patterns:
                         self._site_labels.append(NewtonManager.cl_register_site(body_pattern, xform))
                         self._site_label_scales.append(scale)
-                        if spec_paths is None:
-                            self._site_label_prim_paths.append(None)
-                        else:
-                            self._site_label_prim_paths.append(spec_paths if single_pattern else (spec_paths[index],))
+                        self._site_label_prim_paths.append(spec_paths)
             self._physics_ready_handle = NewtonManager.register_callback(
                 self._on_physics_ready, PhysicsEvent.PHYSICS_READY, name=f"site_view_{self._prim_path}"
             )
--- a/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
+++ b/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
@@ -599,6 +599,11 @@ class NewtonSiteFrameView(BaseFrameView):
         if self._fabric_sel is None and not self._initialize_fabric_mirror():
             return
 
+        # Bodies sync to Fabric at render cadence, so after a ``render=False`` step the parent matrix
+        # the local derivation below reads still holds the last rendered pose. Syncing first keeps
+        # the two spaces consistent; it is a no-op when no transform changed since the last sync.
+        NewtonManager.sync_transforms_to_usd()
+
         count = self._fabric_sel.count
         pos_ta, quat_ta = self._get_world_poses_impl(None)
         wp.launch(
--- a/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
+++ b/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
@@ -294,8 +294,6 @@ class NewtonSiteFrameView(BaseFrameView):
                     self._site_label_scales.append(scale)
                     self._site_label_prim_paths.append(spec_paths)
                 else:
-                    # Only reached before the model exists, where a spec always resolves to one
-                    # pattern -- the cloned ``.*`` form -- whose label covers every environment.
                     for body_pattern in body_patterns:
                         self._site_labels.append(NewtonManager.cl_register_site(body_pattern, xform))
                         self._site_label_scales.append(scale)
@@ -421,9 +419,6 @@ class NewtonSiteFrameView(BaseFrameView):
                                 raise RuntimeError(
                                     f"FrameView destination root '{destination_root}' does not end with '{suffix}'."
                                 )
-                            # The ``.*`` pattern fans out to the cloned body in each environment, so
-                            # the per-environment destination paths still line up with the sites.
-                            # The count check in ``_create_buffers`` catches it if they do not.
                             return (
                                 (destination_root[: -len(suffix)],),
                                 wp.transform(pos, quat),
--- a/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
+++ b/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
@@ -599,9 +599,8 @@ class NewtonSiteFrameView(BaseFrameView):
         if self._fabric_sel is None and not self._initialize_fabric_mirror():
             return
 
-        # Bodies sync to Fabric at render cadence, so after a ``render=False`` step the parent matrix
-        # the local derivation below reads still holds the last rendered pose. Syncing first keeps
-        # the two spaces consistent; it is a no-op when no transform changed since the last sync.
+        # Bodies sync to Fabric only at render cadence, so after a ``render=False`` step the local
+        # derivation below would read a parent still at its last rendered pose. No-op when clean.
         NewtonManager.sync_transforms_to_usd()
 
         count = self._fabric_sel.count
@@ -637,14 +636,12 @@ class NewtonSiteFrameView(BaseFrameView):
         )
 
     def _initialize_fabric_mirror(self) -> bool:
-        """Build the Fabric selection backing :meth:`_mirror_to_fabric`.
+        """Build the Fabric selection backing :meth:`_mirror_to_fabric`, returning whether there is
+        anything to mirror (a ``False`` result is sticky).
 
-        Returns whether there is anything to mirror; a ``False`` result is sticky.
-
-        Selects with ``seed_from_usd=False`` because Newton never writes body poses back to USD, so
-        seeding a Newton-driven prim from it would reset the prim to its spawn pose.  Coverage can be
-        partial: Newton clones physics without cloning USD, so a site can be real while its
-        destination prim exists on no stage at all, and those have nothing to draw.
+        Seeding from USD is off because Newton never writes poses back, so it would reset the prim to
+        its spawn pose. Coverage can be partial: Newton clones physics without cloning USD, so a site
+        can be real while its destination prim exists on no stage, and those have nothing to draw.
         """
         if self._mirror_disabled or self._site_prim_paths is None or self._count == 0:
             self._mirror_disabled = True
@@ -658,8 +655,7 @@ class NewtonSiteFrameView(BaseFrameView):
                 skip_missing_prims=True,
             )
         except ImportError:
-            # No Fabric runtime (kitless run, no renderer): the site state is still correct, there is
-            # simply nothing consuming the mirrored transforms.
+            # No Fabric runtime (kitless run): the site state is still correct, nothing consumes it.
             self._mirror_disabled = True
             logger.info("Fabric runtime unavailable; Newton site poses will not be mirrored to prims.")
             return False
@@ -674,10 +670,8 @@ class NewtonSiteFrameView(BaseFrameView):
                 " the rest have no prim on the stage (physics-only clones) and nothing to render."
             )
 
-        # This view only ever writes through the selection.  Beyond access rights, refreshing the
-        # read-write selection is what notifies the renderer that the matrices changed -- writing
-        # through the read-only one still lands the values in Fabric, but the image keeps showing the
-        # old pose.
+        # Refreshing the read-write selection is what notifies the renderer: a write through the
+        # read-only one lands in Fabric, but the image keeps showing the old pose.
         selection.read_write = True
         self._fabric_sel = selection
         self._mirror_site_indices = wp.array(selection.kept_indices, dtype=wp.int32, device=self._device)
--- a/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
+++ b/source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
@@ -590,12 +590,10 @@ class NewtonSiteFrameView(BaseFrameView):
     def _mirror_to_fabric(self) -> None:
         """Stamp the current site world poses onto the Fabric transforms the renderer reads.
 
-        Called when a transform writer scope exits; Newton keeps site poses in Warp state the RTX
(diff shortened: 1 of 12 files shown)


source/isaaclab_physx/isaaclab_physx/sim/views/fabric_frame_view.py
```diff
@@ -27,26 +27,6 @@
 logger = logging.getLogger(__name__)
 
 
-def _parent_path(prim_path: str) -> str:
```
mataylor-nvidia, line 30: `_parent_path` moved rather than disappeared: its only caller was the parent-path computation in `_initialize_fabric`, which is now `FabricXformSelection.__init__` in `isaaclab/sim/views/fabric_xform_selection.py`. It could not stay here, since `isaaclab` cannot import from `isaaclab_physx`, and the Newton site view now shares the same selection code.

The body and contract are unchanged, including the `RuntimeError` for a prim directly under the stage root; only the error message is generalized off `FabricFrameView`, as both backends now hit it.

mataylor-nvidia:
All tests from the upstream issue are passing

https://github.com/isaac-sim/IsaacLab/issues/7676

https://github.com/alexmillane/IsaacLab/pull/4

mataylor-nvidia:
run-ci

mataylor-nvidia:
> ## Isaac Lab Review Bot
> The shared Fabric selection extraction is reasonable, but it changes root-level `FabricFrameView` behavior by replacing an explicit parent validation with pseudoroot handling. Newton frame writes can also compute a local matrix from stale parent Fabric state when a body and its child frame are moved before rendering.
> 
> * **Design and architecture:** Centralizing Fabric tagging, selections, and slot mapping behind `FabricXformSelection` cleanly supports the PhysX ownership and Newton mirroring paths. However, the abstraction must preserve the existing non-pseudoroot parent requirement for the seeding path, and Newton local-transform derivation should use current Newton-side parent/body state rather than render-synchronized Fabric state.
> * **API:** Exporting `FabricXformSelection` creates a documented public surface, while the `FabricFrameView` refactor preserves its existing private accessors. The shared `parent_path` nevertheless relaxes the prior root-level contract: root-level managed prims now resolve to `/` instead of raising the targeted `RuntimeError`. Restore that validation for the Fabric seeding path.
> * **Implementation:** The Newton mirror correctly uses the read-write selection and updates both world and local matrices. However, `update_indexed_local_matrix_from_world` reads the parent world matrix from Fabric, which may lag current Newton body state until render-time synchronization. A body pose write followed by a child-frame pose write before rendering can therefore produce an incorrect local matrix that the next hierarchy pass propagates into an incorrect world pose. Derive the local transform from current Newton-side body-relative state.
> 
> **Minor fixes needed.** Posted 2 actionable findings inline.
> 
> _Automated review; human maintainers own approval decisions._

fixed this

mataylor-nvidia:
run-ci

AntoineRichard (dismissed):
# AI Review — Request changes

This review was generated by an AI coding agent. Human maintainers should validate the findings and decide whether they block merging.

## Findings

1. **Important — `source/isaaclab/isaaclab/sim/views/fabric_xform_selection.py:111-113`: zero-count selections are left partially initialized.**

   `UsdFrameView` permits zero matches, and the previous `FabricFrameView` implementation initialized an empty view. The new early return leaves selections, buffers, and `view_indices` undefined. A subsequent Fabric getter can therefore fail with `AttributeError` when it accesses `view_indices`. Please either fully initialize the empty state or explicitly reject zero-count views at the interface boundary.

2. **Important — `source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py:919-941`: exception exits can leave Newton and Fabric poses inconsistent.**

   The mirror runs only when `exc_type is None`. If `set_poses()` successfully mutates Newton state and later code in the same writer scope raises, Fabric is not updated even though the Newton write remains. The rendered prim can then remain stale indefinitely. Track whether a pose mutation occurred and reconcile Fabric during exception unwinding without masking the original exception.

3. **Important — `source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py:298-300,683-688`: Fabric state and callbacks do not have complete lifecycle cleanup.**

   `close()` does not deregister `_physics_ready_handle`, so a closed pre-model view can initialize later when `PHYSICS_READY` fires. There is also no shutdown-safe garbage-collection fallback comparable to `FabricFrameView`, so an unclosed directly created view can leave its per-view Fabric tags behind. Deregister the callback in `close()` and add best-effort finalization for the Fabric selection.

4. **Important test gap — the no-seeding constraint is not exercised in its risky ordering.**

   The existing test initializes the mirror before moving the body. It does not cover moving and rendering the Newton body first, then performing the first frame write. A regression to `seed_from_usd=True` could reset the already-moved Fabric parent to stale USD state while the body synchronization path is clean and therefore a no-op. Add a regression test with that ordering.

5. **Important — `FabricXformSelection` is exported as public API but is absent from the API documentation and exposes mutable implementation state.**

   It is exported from `source/isaaclab/isaaclab/sim/views/__init__.pyi`, but is not listed in `docs/source/api/lab/isaaclab.sim.views.rst`. Either keep the helper private, or document and type its lifecycle, state invariants, and supported public surface.

## Suggestion

Mirror only after an actual pose mutation. Empty, getter-only, `set_poses(None, None)`, and scale-only writer scopes currently perform full-view Fabric synchronization even though no mirrored pose changed.

## Review coverage

Reviewed the complete `develop...f53d31d617bb532940d0e6aeca1d70d7ea27cf6f` diff across correctness, error handling, tests, API/type design, comments, and simplification. Relevant Newton, PhysX, rendering, pre-commit, and changelog checks pass; the overall PR check suite still contains unrelated or not-yet-triaged failures in `isaaclab_contrib` and `isaaclab_visualizers`.

mataylor-nvidia:
fixed @AntoineRichard's, AI Review change request

kellyguo11 (approved):

mataylor-nvidia:
@AntoineRichard thanks — all five findings and the suggestion are addressed. That review was against `f53d31d`; head is now `3765c67`. Mapping:

1. **Zero-count selections partially initialized** — `53d3d7e`. `FabricXformSelection` now initializes empty selections, buffers and `view_indices` instead of returning early, so a zero-match `UsdFrameView` keeps working through the Fabric getters.
2. **Exception exits leave Newton and Fabric inconsistent** — `7570e13`. The mirror now runs while an exception unwinds too, since the Newton-side write is committed as soon as `set_poses` returns. A mirror that itself fails during unwinding is logged rather than raised, so it cannot mask the original exception — same shape as the PhysX writer's opposite-space recovery. Pinned by `test_frame_view_pose_write_reaches_fabric_when_the_scope_raises`.
3. **Lifecycle cleanup** — `7570e13`. `close()` deregisters `_physics_ready_handle` (cleared first, so a repeat `close()` cannot deregister twice), and a shutdown-safe `__del__` runs the same cleanup best-effort for a directly created view that is never closed.
4. **Test gap on the risky ordering** — `test_first_frame_pose_write_after_body_move_leaves_the_body_rendered` in `test_newton_fabric_body_sync.py` moves and renders the Newton body first, then performs the first frame write, so a regression to `seed_from_usd=True` would show up as the parent snapping back to its spawn pose.
5. **Public API surface** — `2dd2a14`. `FabricXformSelection` is no longer exported from `isaaclab.sim.views`; both backends import it by module path, so it stays an internal cross-backend helper. The Refactor paragraph in the description said otherwise and has been corrected.

**Suggestion (mirror only after an actual pose mutation)** — taken, in `7570e13`. Empty, getter-only, `set_poses(None, None)` and scale-only scopes no longer trigger a full-view synchronization; scale is deliberately not mirrored at all.

Also addressed since your review: the bot's point about deriving the local matrix from render-lagged Fabric parent state — `cdc2d981` syncs body transforms before deriving the mirrored frame locals.

Re-requesting review.

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

source/isaaclab_newton/isaaclab_newton/sim/views/newton_site_frame_view.py
```diff
@@ -494,6 +566,101 @@ def _create_buffers(
...
+
+        The local matrix is written too, else Newton's body sync forward-propagates ``parent * local``
+        over the world matrix for frames under a body.
+        """
+        if self._fabric_sel is None and not self._initialize_fabric_mirror():
+            return
+
+        # Bodies sync at render cadence, so after a ``render=False`` step the local derivation below
+        # would read a stale parent. No-op when clean.
+        NewtonManager.sync_transforms_to_usd()
```
pbarejko: Why do we sync transforms to usd?

mataylor-nvidia: I updated this to be named `sync_transforms_to_fabric` since this method does not do any USD authoring

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -674,13 +674,16 @@ def pre_render(cls) -> None:
...
                 cls._mark_transforms_dirty()
-        cls.sync_transforms_to_usd()
+        cls.sync_transforms_to_fabric()
         cls.sync_cables_to_usd()
         cls.sync_particles_to_usd()
 
     @classmethod
-    def sync_transforms_to_usd(cls) -> None:
-        """Write Newton body_q to USD Fabric world matrices for Kit viewport / RTX rendering.
+    def sync_transforms_to_fabric(cls) -> None:
```
mataylor-nvidia, line 682: No USD is authored only writes to USD Fabric world matrices

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

source/isaaclab_newton/isaaclab_newton/physics/newton_manager.py
```diff
@@ -674,13 +674,16 @@ def pre_render(cls) -> None:
             cls.forward()
             if NewtonManager._transforms_may_change_on_graph_replay:
                 cls._mark_transforms_dirty()
-        cls.sync_transforms_to_usd()
+        cls.sync_transforms_to_fabric()
```
pbarejko, line 677: Sorry for being a bit pedantic @mataylor-nvidia, but it seems like those methods below contain `_usd` suffix as well. Can you correct it as well please?

pbarejko (approved):

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

source/isaaclab_newton/test/physics/test_newton_fabric_body_sync.py
```diff
@@ -474,3 +476,168 @@ def test_cable_points_follow_newton_segments_after_step_and_reset():
...
+        body_target = body_pose[0, :3].cpu()
+        scene["cube"].write_root_link_pose_to_sim_index(root_pose=body_pose)
+        _render(sim, device)
+        _assert_position(_fabric_position(body_path), body_target)
+
+        written_position = body_target + torch.tensor([0.5, 0.0, 0.0])
+        _write_frame_world_position(view, written_position.to(device))
+        _render(sim, device)
+
+        _assert_position(_fabric_position(body_path), body_target)
```
nvsekkin, line 618: question: running this test locally from this PRs HEAD I get an error on this assertion. @mataylor-nvidia do you get this locally?

nvsekkin: here is what I get locally:
```
source/isaaclab_newton/test/physics/test_newton_fabric_body_sync.py:618: AssertionError

actual = tensor([0., 0., 0.]), expected = tensor([ 1.5000, -0.7500,  2.0000])

E       AssertionError: Tensor-likes are not close!
E       Mismatched elements: 3 / 3 (100.0%)
E       Greatest absolute difference: 2.0 at index (2,) (up to 0.0001 allowed)
```

nvsekkin:
Following test fails on CI: https://github.com/isaac-sim/IsaacLab/actions/runs/35017339386/job/104561924842?pr=7691#step:4:711

I tried to run this locally and also got:
```
E           Failed: cartpole (physics=newton, visualizer=kit, mode=tiled) failed:
E             Pixel diff (6.29%) exceeds threshold of 3.00%. SSIM=0.8689.
```

I opened the `tests/comparison-images/cartpole-newton-kit-tiled-actual.png` and this image is completely black

source/isaaclab_newton/test/physics/test_newton_fabric_body_sync.py
nvsekkin: https://github.com/isaac-sim/IsaacLab/pull/7691/changes#diff-a36cf96fbdf4c4d146e5f4b751c977260b5de4c2a3a1deaf1c63c6bba21df2eeR614

Isolated it to one line. Flipping newton_site_frame_view.py:614 to seed_from_usd=False and changing nothing else makes all 5 frame-view tests pass; at True this assertion fails.

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

mataylor-nvidia:
run-ci

PR #7697, opened 2026-09-09
Fix shutdown hang on Ctrl+C with multiple GUI-backed visualizers

matthewtrepte:
# Description

An open-ended session with multiple GUI-backed visualizers active (e.g. `--visualizer
newton,kit,rerun,viser`) did not stop cleanly on Ctrl+C / SIGINT: it stayed alive through the
shutdown grace period and had to be killed. Single-visualizer sessions stopped cleanly on
their own; multi-visualizer combinations were inconsistent.

Two distinct gaps combined to cause this:

1. **`env.close()` only ran on the happy path.** In `random_agent`/`zero_agent` and every
   `train`/`play` RL backend (`rsl_rl`, `rl_games`, `skrl`, `sb3`), the pattern was
   `try: ...loop... ; env.close() except KeyboardInterrupt: pass`. When the interrupt landed
   during the loop, execution jumped straight to `except`, which silently swallowed it
   *without* calling `env.close()`. That call is what reaches
   `SimulationContext.clear_instance()`, which is the only place that calls `.close()` (not
   just `.stop()`) on every active visualizer, tearing down `rerun`'s/`viser`'s server
   threads and the Newton/Kit windows. Skipping it left those resources running, and the
   process hung waiting for them.

2. **A raw `KeyboardInterrupt` doesn't reliably land where you'd expect.** Python delivers
   `SIGINT` asynchronously, interrupting whatever bytecode happens to be executing at the
   moment. With multiple GUI-backed visualizers pumping their own render/event callbacks,
   that's often *inside one of Kit's own internal callbacks* (its viewport texture or event
   dispatch), not inside the script's own loop -- where no `try/except` here could ever catch
   it. This is consistent with the reported shutdown matrix being inconsistent across
   visualizer combinations rather than reliably tied to one frontend.

## Fix

- `env.close()` now runs from a `finally` block on every exit path (normal completion, `break`,
  or interrupt) in all 8 affected entrypoints.
- `random_agent`/`zero_agent` additionally install a `SIGINT` handler that only sets a flag,
  checked at a controlled point at the top of the loop, instead of relying on an async
  exception landing somewhere useful. This is safe regardless of which frame the signal
  actually interrupts.

## Testing

Reproduced with the exact repro command from the report:

```
env DISPLAY=:0 OMNI_KIT_ACCEPT_EULA=yes timeout --signal=INT --kill-after=30 90 \
  uv run isaaclab random_agent --task Isaac-Cartpole --num_envs 64 --max_steps 1000000 \
  physics=newton_mjwarp --visualizer newton,kit,rerun,viser
```

- Before the fix: exit 137 (SIGKILL required after the grace period).
- After the fix: exit 124 (clean shutdown within the grace period), consistent across two
  repeated runs. The log shows the expected clean teardown sequence ("(viser) Server
  stopped", "SimulationContext cleared", "Simulation App Shutting Down").
- `source/isaaclab_rl/test/test_entrypoints.py`: 43 passed.

## Type of change

## Checklist

- [x] I have made corresponding changes to the documentation (none required; no public API changed)
- [x] I have added a changelog fragment under `source/isaaclab_rl/changelog.d/`

### DIFF
--- a/source/isaaclab_rl/changelog.d/mtrepte-fix-shutdown-hang-multi-visualizer.rst
+++ b/source/isaaclab_rl/changelog.d/mtrepte-fix-shutdown-hang-multi-visualizer.rst
@@ -1,0 +1,15 @@
+Fixed
+^^^^^
+
+* Fixed a shutdown hang in ``random_agent``/``zero_agent`` and every ``train``/``play`` RL
+  backend (``rsl_rl``, ``rl_games``, ``skrl``, ``sb3``) when interrupted with Ctrl+C while
+  multiple GUI-backed visualizers were active (for example ``--visualizer newton,kit``).
+  ``env.close()`` previously only ran on the happy path -- ``except KeyboardInterrupt: pass``
+  silently swallowed the interrupt without closing the environment, so the process never
+  reached the visualizer teardown in :meth:`~isaaclab.sim.SimulationContext.clear_instance`
+  and had to be killed. ``env.close()`` now runs from a ``finally`` block on every exit path.
+  In ``random_agent``/``zero_agent``, a raw ``KeyboardInterrupt`` could also be delivered
+  asynchronously inside one of Kit's own internal callbacks (its viewport/render event loop)
+  rather than in the script's own loop; that entrypoint now installs a ``SIGINT`` handler that
+  only sets a flag, checked at a controlled point in the loop, instead of relying on where the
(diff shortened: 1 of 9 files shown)


matthewtrepte:
Closing to reopen with a follow-up fix for a P1 review finding (second Ctrl+C during cleanup could abort teardown early). See the new PR.

